PR #3, opened 2015-08-14
WordLemmaBucket: mehr lemma ist mehr gut

diekmann:
Two additional lemmas about enumerations; only for type word


lsf37:
Thanks for sending in the CLA. Pull request is now rebased and merged!

PR #6, opened 2015-10-09
Consistently use /usr/bin/env to invoke python

cmr:


Smattr:
Looks fine to me. I'm surprised we were hardcoding the Python path previously.

lsf37:
Rebased and applied internally. Will be pushed out when the current ongoing merge is done and has passed tests (1-2 days or so).

Smattr:
@lsf37, did you have an update on this one? I assume the merge you were referring to is notification binding. Has it passed all the internal hurdles?

lsf37:
It's in l4v internally already. The regression server (lemma) has died last week and we're still resolving the aep -> notification rename, so things are not pushed out yet, but this part is just sitting there ready to go.

lsf37:
Now available in d28994d860cc6b5b9608a1b0f22e32827f98afa5. Thanks for cleaning this up.

PR #7, opened 2015-10-09
misc/regression/memusage.py: work around psutil API changes

### LINKED ISSUE #5: psutil methods with get_ prefix are deprecated
psutil as of 2.0.0 renamed all getters to avoid the `get_` prefix, in March 2014, leaving behind a deprecation warning.

The 3.0.0 release in June this year removed them.

Unfortunately, Ubuntu/Debian still ship an ancient psutil version. Other Linux distros are using the new version, and the new versions are what can be obtained from pip. I'm not sure what the best way forward is here.

cmr:
Closes #5

### DIFF
--- a/misc/regression/memusage.py
+++ b/misc/regression/memusage.py
@@ -20,6 +20,11 @@ import subprocess, sys, threading, time
 PSUTIL_NOT_AVAILABLE=False
 try:
     import psutil
+    if not hasattr(psutil.Process, "children") and hasattr(psutil.Process, "get_children"):
+        psutil.Process.children = psutil.Process.get_children
+    if not hasattr(psutil.Process, "memory_maps") and hasattr(psutil.Process, "get_memory_maps"):
+        psutil.Process.memory_maps = psutil.Process.get_memory_maps
+
 except ImportError:
   PSUTIL_NOT_AVAILABLE=True
 
@@ -35,7 +40,7 @@ else:
     to give us a more accurate total usage.'''
     assert isinstance(proc, psutil.Process)
     try:
-        return sum([m.pss for m in proc.get_memory_maps(grouped=True)])
+        return sum([m.pss for m in proc.memory_maps(grouped=True)])
     except psutil.AccessDenied:
         # If we don't have permission to read a particular process,
         # just return 0.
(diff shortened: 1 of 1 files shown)


misc/regression/memusage.py
```diff
@@ -20,6 +20,11 @@
 PSUTIL_NOT_AVAILABLE=False
 try:
     import psutil
+    if not hasattr(psutil.Process, "children") and hasattr(psutil.Process, "get_children"):
+        psutil.Process.children = psutil.Process.get_children
+    if not hasattr(psutil.Process, "memory_maps") and hasattr(psutil.Process, "get_memory_maps"):
+        psutil.Process.memory_maps = psutil.Process.get_get_memory_maps
```
Smattr: Is this correct? I think you mean `get_memory_maps`.

cmr: You're right, typo.

cmr:
Updated.

Smattr:
Thanks for submitting this change. Have you contributed to an seL4 project before? For legal reasons, we need every contributor to sign a [Contributor Licence Agreement](http://sel4.systems/Community/Contributing/). Are you able to fill one of these out and email it to the address on that page? If you've sent one through previously, please let me know.

cmr:
I've emailed a CLA from me and my group a few weeks ago, though I got a notification that the receiver was on a vacation.

lsf37:
Found it and have added it to the database now. Sorry for the mixup.

Smattr:
This change looks OK to me, but we should probably test it with a wider variety of psutil users. My local psutil (1.2.1) has the older get_ functions. The machine "lemma" has 2.1.1 which seems to have both variants. Seems the older API was removed in 3.0, so might be good to test with a user with >=3.0.

cmr:
I'm with 3.2.1. I had to do these changes to not have the script hang when running run_tests.

Smattr:
@lsf37, did you have more to say on this one? I'm happy for it to be merged.

lsf37:
Ok, happy with that. Will apply it internally and see if `lemma` is agreeable to it when resurrected from hdd death.

lsf37:
Rebased and pushed internally. Should show up next time we push to github.

lsf37:
Now available on github, see 1ad00aa1aeede11a74a6fb257b6e9c5739f052a8

PR #8, opened 2016-03-24
Mehr Lemmata ist mehr gut

diekmann:
Word lemmas about casting and bitlists.

gk has the CLA.


lsf37:
Merged internally, if everything goes well, should show up here within 24h.

PR #10, opened 2016-05-03
WordLemmaBucket: NOT_mask_shifted_lenword

diekmann:
A simplification lemma which heavily relies on the above `mask_shift_and_negate` lemma.

gk has the CLA.


lsf37:
Thanks, will pull in! I'm in the process of re-organising the word library in preparation for an AFP submission. I might push this out in a branch in a while so you can have a look, since you seem to be one of the most active users.

diekmann:
By the way, regarding the `Word_Lib` refactoring: I upgraded my git submodule to l4v@45c5aaa (current head of master). Everything works fine for me. Very nice work! :+1:

lsf37:
Great! Your lemma should appear soon, when the internal test is green the next time.

lsf37:
Now finally rebased and pushed here. See aa50dc6858f5dbe2a2b643bcb18cc609afdee15f.

PR #27, opened 2019-04-30
misc: support marginal comments in Vim syntax

Smattr:
This adds Vim syntax highlighting for the newer `— ‹ ... ›` style marginal comments. It could do with a bit of improvement, as the escape sequences aren't properly concealed even at `conceallevel=2`. I tried meddling with `hs` and `ms` on the region start but the best I could manage was conceal only if the comment was not preceded by the start of the line, which seemed worse. If any reader has greater Vim fu and knows how to do this, please feel free to amend.

Is there any interest in upstreaming this into Vim? It probably needs some review from the Vim maintainers, but they're generally pretty accepting of new syntax highlighting.

### DIFF
--- a/misc/vim/isabelle.vim
+++ b/misc/vim/isabelle.vim
@@ -108,6 +108,8 @@ syntax region IsabelleComment matchgroup=IsabelleComment start="(\*" end="\*)" c
 syntax match IsabelleCommentStart "(\*" contained nextgroup=IsabelleCommentContent contains=IsabelleCommentStart
 syntax match IsabelleCommentContent ".*" contained
 
+syntax region IsabelleComment matchgroup=IsabelleComment fold start="\\<comment>\s*\\<open>" end="\\<close>" contains=IsabelleSpecial keepend
+
 " You can use LaTeX within text {* ... *} blocks and friends and sometimes it
 " is useful to have syntax highlighting enabled within these blocks when
 " working on a PDF-destined theory. This is off by default because it can be a
@@ -640,10 +642,11 @@ syn match IsabelleSpecial /\\<s>/ conceal cchar=𝗌
 syn match IsabelleSpecial /\\<Leftrightarrow>/ conceal cchar=⇔
 syn match IsabelleSpecial /\\<heartsuit>/ conceal cchar=♡
 syn match IsabelleSpecial /\\<four>/ conceal cchar=𝟰
-syn match IsabelleSpecial /\\<open>/ conceal cchar=‹
-syn match IsabelleSpecial /\\<close>/ conceal cchar=›
+syn match IsabelleSpecial /\\<open>/ conceal cchar=‹ transparent
+syn match IsabelleSpecial /\\<close>/ conceal cchar=› transparent
+syn match IsabelleSpecial /\\<comment>/ conceal cchar=— transparent
 
-syn cluster IsabelleInnerStuff contains=IsabelleSpecial
+syn cluster IsabelleInnerStuff contains=IsabelleSpecial,IsabelleComment
 
 " Enable folding of proofs and locales. Note that the starting regex needs to
 " match with zero width to preserve syntax highlighting of the opening command.


lsf37:
Nice! I should leave upstreaming comments to the actual vim users amongst us, but I see no problem with it.

Xaphiosis:
I don't use the l4v syntax highlighter, but I do use something similar I hacked up for myself, with a similar need for this update, so thanks.
Not sure what the exact problem you're having is, but it is relatively hard to get all the start/end and highlighting right. Maybe what I ended up with will help solve it for you? Here's the relevant extract.

```
syn region isaComment start="\\<comment>\s*\\<open>" end="\\<close>" keepend contains=isaComment,isarSymbol
HiLink isaComment Comment
syn match isarSymbol '\\<comment>' conceal cchar=— contained transparent
syn match isarSymbol '\\<open>' conceal cchar=‹ contained transparent
syn match isarSymbol '\\<close>' conceal cchar=› contained transparent
```

Nice to hear from you again by the way, and to see you're still using Isabelle for something.

Smattr:
Nice to talk to both of you too :)

@Xaphiosis, the relevant differences look like you also have the comment region containing itself and the symbols are contained. After applying this, I get the ending `\<close>` concealed but not the leading `\<comment>` or `\<open>`. Do these conceal for you? I'm testing with Vim 8.1 on Linux.

If you have a different syntax file, I imagine it's probably more advanced than the l4v one. Is it worthwhile (or do you have any desire) to upstream yours into Vim?

Xaphiosis:
Yeah, they conceal and highlight for me (vim 8.1, ubuntu):
hi_comment

I think my version is both a giant hack and much more limited than the l4v version, i.e. hardly suitable for upstreaming into vim. I stripped it down to the a minimal syntax file, can you look at this and see if it at least this works for you to both highlight the comment and conceal?

```
if !exists("main_syntax")
  if version < 600
    syntax clear
  elseif exists("b:current_syntax")
    finish
  endif
  let main_syntax = 'thy'
endif
syn case ignore

syn region isaComment start="(\*" end="\*)" contains=isaComment,isarSymbol
syn region isaComment start="\\<comment>\s*\\<open>" end="\\<close>" keepend contains=isaComment,isarSymbol
syn region isaAntiQuot contained start="@{" end="}" keepend contains=@NoSpell,isarSymbol

command -nargs=+ HiLink hi def link <args>
HiLink isaTextCmd PreProc
HiLink isaAntiQuot Comment
HiLink isaComment Comment
delcommand HiLink

if main_syntax == 'thy'
    unlet main_syntax
endif

if has("conceal") && &enc == 'utf-8'

  " stripped down from original
  let s:isar_greek=[
    \ ['alpha', 'α'],
    \ ['Omega', 'Ω']]

  " stripped down from original
  let s:isar_misc=[
    \ ['bool', 'ℬ'],
    \ ['int', 'ℤ']]

  " stripped down from original
  let s:isar_symbols=[
    \ ['leftarrow', '←'],
    \ ['rightarrow', '→'],
    \ ['some', 'ϵ']]

  for isartok in s:isar_greek
   exe "syn match isarSymbol '\\\\<".isartok[0].">' conceal cchar=".isartok[1]
  endfor

  for isartok in s:isar_misc
   exe "syn match isarSymbol '\\\\<".isartok[0].">' conceal cchar=".isartok[1]
  endfor

  for isartok in s:isar_symbols
   exe "syn match isarSymbol '\\\\<".isartok[0].">' conceal cchar=".isartok[1]
  endfor

  syn match isarSymbol '\\<comment>' conceal cchar=— contained transparent
  syn match isarSymbol '\\<open>' conceal cchar=‹ contained transparent
  syn match isarSymbol '\\<close>' conceal cchar=› contained transparent

endif

let b:current_syntax = 'thy'
```

Smattr:
Nice! TBH your version looks less hacky than mine. Yours seems to work except for:

1. It doesn't highlight the opening and closing text as part of the comment when it's concealed. Your screenshot shows that to, so I guess this is a known limitation.
2. It doesn't support nested comments which Isabelle seems to allow.

I don't know why nested comments don't work, as I cross referenced Vim's OCaml syntax file which supports them and it seems to be doing the same thing.

What's your preferred next step @Xaphiosis? I would vote for closing this PR and replacing the logic in l4v's syntax with something more like yours.

Xaphiosis:
Concealed text is sadly only ever highlighted with one group: "Conceal", i.e. this is a Vim limitation. Was this what the original issue with your patch was? I'm now not sure what the issue is, it might be the case that your patch is fine and I just made a lot of noise for no reason. If it is a problem still, could I ask for a screenshot?
Thanks for pointing out my version does not support nested comments.

I will have a bit of time to look at this next week, and maybe can reach something that works based on our combined efforts.

Smattr:
Ah I did not realise concealed characters could not be highlighted. In that case, I think yours is probably the best we can do. Do you want to close this PR and push something closer to your logic?

lsf37:
Merged in 18c9f36

PR #33, opened 2020-08-07
lib/riscv refine: move lemma

lsf37:
Minor commit, mostly to check github workflow.

### DIFF
--- a/lib/Corres_UL.thy
+++ b/lib/Corres_UL.thy
@@ -701,6 +701,12 @@ lemma corres_assert_assume:
   by (auto simp: bind_def assert_def fail_def return_def
                  corres_underlying_def)
 
+lemma corres_assert_gen_asm_cross:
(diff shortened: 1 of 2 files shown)


lsf37:
Trying to get the testboard running with github remotes. First one worked, but this one does not: https://bamboo.ts.data61.csiro.au/browse/L4V-TESTBOARD-572

Will keep investigating

mbrcknl:
Not sure what's gone wrong, but at least this line from the failed Bamboo log doesn't look right:
```
error	07-Aug-2020 18:21:43	error: Cannot fetch l4v from ssh://git@bitbucket.ts.data61.csiro.au:7999/sel4/git@github.com:lsf37/l4v
```

lsf37:
I think I figured it out. `repo` wants ssh URLS and treats `git@github.com:lsf37/l4v` as a relative path instead of an URL. I should be able to adjust the testboard script to substitute `ssh://git@github.com/lsf37/l4v` instead.

lsf37:
Yup, it looks like https://bamboo.ts.data61.csiro.au/browse/L4V-TESTBOARD-572 will be succeeding.

PR #34, opened 2020-08-08
more github workflows

lsf37:
This standardises to github workflows to the ones in seL4/ci-actions (which includes a link checker for .md files), and runs the ARM AInvs and Refine sessions on github CI.


mbrcknl:
Is `rt` the intended target branch?

lsf37:
Thanks for spotting that, it used to be, but is not any more (should be updated to master now).

lsf37:
I'll let it run for a few more hours to get an AInvs and Refine cache, but it's looking good so far (testing so far was only up to ExecSpec on my fork).

PR #38, opened 2020-08-10
sync rt with master

lsf37:
This should make `AInvs` and `Refine` tests available on the PRs in the `rt` branch.


lsf37:
(not really looking for a real review, this is more FYI that the merge is incoming)

lsf37:
> I see that the GitHub workflows refer to branches, yet the config files that do this are on branches. How does that work?

It's even worse in some way: the github workflow that is executed on a PR is the one *inside* the PR. But inside that workflow file you can have conditions on when you want things to run (e.g. it could say to not run on PRs to `rt`). So if the workflow in master says test `master` and `rt`, then the idea is that PRs will come off branches of `master` and contain the same workflow file which then has the right conditions (assuming it hasn't been changed). For `rt` this currently means that it will only start checking proofs from this `PR` on, but already inside this `PR`. Once it is merged, all further PRs off `rt` (and then to `rt` again) will contain this workflow file.

mbrcknl:
> the github workflow that is executed on a PR is the one _inside_ the PR.

I guess that works. :-)

PR #40, opened 2020-08-12
mcs: strengthen typ_at_lifts

corlewis:
This extends typ_at_lifts to also handle the negated versions. As part of this it produces two different formulations of the lifting lemmas, using ML_goal instead of manually duplicating all of them.


proof/invariant-abstract/ARM/ArchVSpace_AI.thy
```diff
@@ -1719,18 +1719,10 @@ lemmas set_vm_root_typ_ats [wp] = abs_typ_at_lifts [OF set_vm_root_typ_at]
...
-  apply (insert bool_function_four_cases[where f=P])
-  apply (erule disjE)
-   apply (cases pte)
-     apply (simp add: data_at_def | wp hoare_vcg_const_imp_lift x)+
-  apply (erule disjE)
-   apply (cases pte)
-     apply (simp add: data_at_def | wp hoare_vcg_disj_lift hoare_vcg_const_imp_lift x)+
-  apply (erule disjE)
-   apply (simp | wp)+
+  apply (cases P rule: bool_to_bool_cases; wpsimp; cases pte)
```
mbrcknl, line 1722: Huh, nice to know that this works!

PR #41, opened 2020-08-12
mcs: prove cancelBadgedSends_invs

corlewis:
@michaelmcinerney has been working on a similar lemma, so a review from him would be helpful. Otherwise whoever has time to look at this.

### DIFF
--- a/proof/refine/ARM/IpcCancel_R.thy
+++ b/proof/refine/ARM/IpcCancel_R.thy
@@ -2658,7 +2658,7 @@ proof -
 qed
 
 lemma cancel_all_invs'_helper:
-  "\<lbrace>all_invs_but_sym_refs_ct_not_inQ' and sch_act_sane
+  "\<lbrace>all_invs_but_sym_refs_ct_not_inQ'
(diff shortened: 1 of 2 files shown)


michaelmcinerney:
I like it!

lsf37:
The proof seems to time out on github. Do you have a testboard run?

corlewis:
> 
> The proof seems to time out on github. Do you have a testboard run?

Yep, and it worked fine both there and locally for me. I don't think it's related, but one of the proofs broke after rebasing over some of the recent commits that added a new precondition for `rescheduleRequired`. I think it would have been possible to show that the precondition is preserved, but it turns out that it isn't necessary. I've just added a new commit that fixes this up, maybe that will also fix the test?

lsf37:
Let's see what it does.

lsf37:
Ok, it looks like Refine is passing now, you just need to actually squash the sqash commit to make gitlint happy.

corlewis:
> 
> 
> Ok, it looks like Refine is passing now, you just need to actually squash the sqash commit to make gitlint happy.

Yep, and it's nice that we have a test for that now!

PR #44, opened 2020-08-13
[MCS] revert notification binding changes

MitchellBuckley:
This PR updates the aspec, ainvs, haskell, and refine proofs to properly re-sync the `rt` branch with `master` with respect to `maybe_return_sc`: both in its contents, and where it is called.

### DIFF
--- a/proof/invariant-abstract/DetSchedSchedule_AI.thy
+++ b/proof/invariant-abstract/DetSchedSchedule_AI.thy
@@ -12803,6 +12803,64 @@ lemma set_thread_state_pred_map_eq[wp]:
   "\<lbrace>\<top>\<rbrace> set_thread_state t st \<lbrace>\<lambda>_ s. pred_map_eq st (tcb_sts_of s) t\<rbrace>"
   by (wpsimp wp: valid_sched_wp simp: vs_all_heap_simps)
 
+crunches schedule_tcb
+  for valid_sched[wp]: valid_sched
+  (wp: reschedule_valid_sched_const)
+
+crunches maybe_return_sc
+  for etcbs_of[wp]: "\<lambda>s. P (etcbs_of s)"
+  (simp: wp: hoare_drop_imp hoare_vcg_if_lift2 ignore: set_tcb_obj_ref)
+
+(* FIXME RT: move somewhere appropriate *)
+lemma tcb_sc_refill_cfgs_upd:
+  "tcb_sc_refill_cfgs_2 (\<lambda>a. if a = tptr then Some None else tcb_scps_of s a) (sc_refill_cfgs_of s)
+   = (\<lambda>a. if a = tptr then None else tcb_sc_refill_cfgs_2 (tcb_scps_of s) (sc_refill_cfgs_of s) a)"
+  unfolding tcb_sc_refill_cfgs_2_def
+  by (fastforce simp: map_join_def vs_all_heap_simps opt_map_def)
+
+(* FIXME RT: move somewhere appropriate *)
+lemma sc_ready_times_upd:
+  "sc_ready_times_2 (\<lambda>a. if a = tptr then None else tcb_sc_refill_cfgs_of s a)
+   = (\<lambda>a. if a = tptr then None else sc_ready_times_2 (tcb_sc_refill_cfgs_of s) a)"
+  unfolding sc_ready_times_2_def
+  by (clarsimp simp: map_project_upd)
+
+(* FIXME RT: move somewhere appropriate, maybe rename *)
+lemma sorted_release_q_heap_upd_idem:
+  "not_in_release_q tptr s \<Longrightarrow>
+   sorted_release_q_2 (\<lambda>a. if a = tptr then None else tcb_sc_refill_cfgs_of s a) (release_queue s)
+   = sorted_release_q_2 (tcb_sc_refill_cfgs_of s) (release_queue s)"
+  by (rule sorted_release_q_2_eq_lift, clarsimp simp: in_release_q_def sc_ready_times_upd)
+
+lemma maybe_return_sc_valid_sched[wp]:
+  "\<lbrace> valid_sched and scheduler_act_not tptr and not_queued tptr and not_in_release_q tptr\<rbrace>
+   maybe_return_sc ntfnptr tptr
+   \<lbrace> \<lambda>_. valid_sched\<rbrace>"
+  apply (simp add: maybe_return_sc_def assert_opt_def)
+  apply (wpsimp wp: reschedule_valid_sched_const valid_sched_wp get_tcb_obj_ref_wp
+                    get_sk_obj_ref_wp)
+  apply (clarsimp simp: valid_sched_def)
+  apply (intro conjI)
+      apply (clarsimp simp: valid_ready_qs_def vs_all_heap_simps not_queued_def)
+     apply (clarsimp simp: valid_release_q_def vs_all_heap_simps not_in_release_q_def)
+     apply (intro conjI)
+      apply clarsimp
+     apply (clarsimp simp: tcb_sc_refill_cfgs_upd
+            , subst sorted_release_q_heap_upd_idem
+            , clarsimp simp: not_in_release_q_def
+            , simp)
+    apply (clarsimp simp: valid_sched_action_def weak_valid_sched_action_def vs_all_heap_simps
+                          scheduler_act_not_def)
+   apply (fastforce simp: valid_blocked_defs vs_all_heap_simps)
+  apply (clarsimp simp: released_ipc_queues_defs vs_all_heap_simps)
+  done
+
+lemma maybe_return_sc_released_if_bound_sc_tcb_at[wp]:
+  "maybe_return_sc a b \<lbrace>released_if_bound_sc_tcb_at thread\<rbrace>"
+  unfolding maybe_return_sc_def
+  apply (wpsimp wp: valid_sched_wp get_tcb_obj_ref_wp get_sk_obj_ref_wp)
+  by (fastforce simp: pred_map_def pred_map_eq_def vs_all_heap_simps obj_at_def)
+
 context DetSchedSchedule_AI begin
 
 lemma send_signal_BOR_helper:
@@ -13017,6 +13075,18 @@ crunches receive_ipc_preamble, refill_unblock_check
   and cur_time[wp]: "\<lambda>s. P (cur_time s)"
   (wp: crunch_wps)
 
+lemma maybe_return_sc_not_queued[wp]:
+  "\<lbrace>not_queued thread and scheduler_act_sane\<rbrace>
+   maybe_return_sc ntfnptr thread
+   \<lbrace>\<lambda>rv. not_queued thread\<rbrace>"
+  unfolding maybe_return_sc_def
+  by (wpsimp wp: reschedule_required_not_queued get_tcb_obj_ref_wp get_sk_obj_ref_wp)
+
+crunches maybe_return_sc
+  for scheduler_act_not[wp]: "scheduler_act_not t"
+  and cur_time[wp]: "\<lambda>s. P (cur_time s)"
+  (wp: crunch_wps simp: crunch_simps)
+
 lemma receive_ipc_valid_sched:
   "\<lbrace>\<lambda>s. valid_sched s
         \<and> st_tcb_at active thread s
@@ -13050,6 +13120,18 @@ lemma receive_ipc_valid_sched:
             in hoare_vcg_if_split, wp)
    prefer 2 apply fastforce
   apply (thin_tac "\<not> (_ \<and> _)", clarsimp)
+  apply (rule_tac B="\<lambda>r s. valid_sched s \<and>
+             st_tcb_at active thread s \<and>
+             cur_thread s = thread \<and>
+             not_queued thread s \<and>
+             not_in_release_q thread s \<and>
+             scheduler_act_not thread s \<and>
+             (pred_map_eq None (tcb_scps_of s) thread \<or> released_sc_tcb_at thread s) \<and>
+             receive_ipc_preamble_rv reply_cap reply_opt s \<and>
+             ko_at (Endpoint ep) ep_ptr s \<and> invs s \<and> current_time_bounded (Suc 0) s"
+           in hoare_seq_ext[rotated])
+  apply (wpsimp split: if_split)
+  apply (frule (1) not_idle_thread[OF invs_valid_idle], simp)
   apply (case_tac ep; simp)
   \<comment> \<open>IdleEP, RecvEP\<close>
   apply (wpsimp wp: receive_ipc_blocked_valid_sched[where reply=reply_cap])
@@ -13085,6 +13167,7 @@ lemma receive_ipc_valid_sched:
                     split: endpoint.splits if_splits)
   \<comment> \<open>get_thread_state\<close>
   apply (rule hoare_seq_ext[OF _ gts_sp])
+  apply (rename_tac sender_state)
   apply (case_tac sender_state; clarsimp)
   apply (rename_tac ep_ptr' sender_data)
   \<comment> \<open>do_ipc_transfer, and stash some knowledge for later\<close>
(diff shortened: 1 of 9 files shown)


lsf37:
You can mark PR's that shouldn't be merged/reviewed yet as draft ("convert to draft" under the reviewers). Makes it easier to see where input is required.

spec/haskell/src/SEL4/Kernel/Thread.lhs
```diff
@@ -356,7 +356,7 @@ We check here that the candidate has the highest priority in the system.
 >
 >             idleThread <- getIdleThread
 >             targetPrio <- threadGet tcbPriority candidate
->             curPrio <- threadGet tcbPriority curThread
+>             curPrio <- if (idleThread == curThread) then (return 0) else (threadGet tcbPriority curThread)
```
lsf37, line 359:
```suggestion
>             curPrio <- if idleThread == curThread then return 0 else threadGet tcbPriority curThread
```
(none of the parens are needed)

spec/haskell/src/SEL4/Object/SchedContext.lhs
```diff
@@ -686,6 +686,8 @@ This module uses the C preprocessor to select a target architecture.
 >         scPtr <- return $ fromJust nscOpt
 >         sc <- getSchedContext scPtr
 >         setSchedContext scPtr (sc { scTCB = Nothing })
+>         ct <- getCurThread
+>         when (tcbPtr == ct) $ rescheduleRequired()
```
lsf37, line 690:
```suggestion
>         when (tcbPtr == ct) $ rescheduleRequired
```

proof/invariant-abstract/IpcDet_AI.thy
```diff
@@ -1086,6 +1086,200 @@ lemma reply_push_sender_sc_Some_invs:
                    elim: if_live_then_nonz_capD2)
   done
 
+crunches maybe_return_sc
+  for pspace_aligned[wp]: pspace_aligned
+  and pspace_distinct[wp]: pspace_distinct
+  and arch_state[wp]: "\<lambda>s. P (arch_state s)"
+  (simp: get_sk_obj_ref_def get_simple_ko_def get_object_def wp: weak_if_wp)
+
+lemma maybe_return_sc_valid_objs[wp]:
```
MitchellBuckley, line 1095: Can we crunch many of these?

proof/invariant-abstract/IpcDet_AI.thy
```diff
@@ -1086,6 +1086,200 @@ lemma reply_push_sender_sc_Some_invs:
...
+   \<Longrightarrow> (\<And>a P. f \<lbrace>reply_sc_reply_at P a \<rbrace>)
+   \<Longrightarrow> f \<lbrace>receive_ipc_preamble_rv reply reply_opt\<rbrace>"
+  by (case_tac reply_opt; wpsimp wp: hoare_vcg_conj_lift; assumption)
+
+lemma maybe_return_sc_fault_tcbs_valid_states[wp]:
+  "maybe_return_sc y thread \<lbrace>fault_tcbs_valid_states\<rbrace>"
+  unfolding fault_tcbs_valid_states_def
+  by (wpsimp wp: hoare_vcg_all_lift hoare_vcg_imp_lift' maybe_return_sc_pred_tcb_at)
+
+(* FIXME RT: rename during cleanup *)
```
MitchellBuckley, line 1223: Fix this.

proof/invariant-abstract/IpcDet_AI.thy
```diff
@@ -1086,6 +1086,200 @@ lemma reply_push_sender_sc_Some_invs:
...
+
+lemma reschedule_required_cur_sc_tcb':
+  "\<lbrace>\<lambda>s. sc_tcb_sc_at \<top> (cur_sc s) s\<rbrace> reschedule_required \<lbrace>\<lambda>_. cur_sc_tcb\<rbrace>"
+  supply set_scheduler_action_cur_sc_tcb [wp del]
+  unfolding reschedule_required_def
+  by (wpsimp simp: reschedule_required_def set_scheduler_action_def tcb_sched_action_def
+                   set_tcb_queue_def get_tcb_queue_def thread_get_def is_schedulable_def
+                   cur_sc_tcb_def sc_tcb_sc_at_def obj_at_def)
+
+lemma maybe_return_sc_fault_tcbs_valisdfd_states[wp]:
```
MitchellBuckley, line 1250: rename

proof/invariant-abstract/IpcDet_AI.thy
```diff
@@ -1109,16 +1304,27 @@ lemma ri_invs[wp]:
    apply (wpsimp wp: receive_ipc_preamble_invs receive_ipc_preamble_caps_of_state
                      receive_ipc_preamble_st_tcb_at receive_ipc_preamble_rv
                simp: ex_nonz_cap_to_def cte_wp_at_caps_of_state)
+  apply (fastforce)
```
MitchellBuckley, line 1307: Remove brackets

proof/invariant-abstract/IpcDet_AI.thy
```diff
@@ -2570,20 +2608,18 @@ lemma maybe_return_sc_schedule_tcb_helper2:
...
    do
      _ <- maybe_return_sc ntfn_ptr tcb_ptr;
      schedule_tcb tcb_ptr
    od
    \<lbrace>\<lambda>_. invs\<rbrace>"
   apply (rule hoare_seq_ext[OF schedule_tcb_invs'])
-  apply (simp add: maybe_return_sc_def)
-  apply (rule hoare_seq_ext[OF _ gsc_ntfn_sp])
-  apply (rule hoare_seq_ext[OF _ gsc_sp])
-  apply (wpsimp wp: hoare_vcg_disj_lift maybe_return_sc_schedule_tcb_helper1
```
MitchellBuckley, line 2583: These deleted functions are now not needed?

proof/invariant-abstract/IpcDet_AI.thy
```diff
@@ -1086,6 +1086,200 @@ lemma reply_push_sender_sc_Some_invs:
                    elim: if_live_then_nonz_capD2)
   done
 
+crunches maybe_return_sc
```
MitchellBuckley, line 1089: Large parts of this block were merely moved from lemmas below.

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -12804,6 +12804,67 @@ lemma set_thread_state_pred_map_eq[wp]:
   "\<lbrace>\<top>\<rbrace> set_thread_state t st \<lbrace>\<lambda>_ s. pred_map_eq st (tcb_sts_of s) t\<rbrace>"
   by (wpsimp wp: valid_sched_wp simp: vs_all_heap_simps)
 
+crunches schedule_tcb
+  for valid_sched[wp]: valid_sched
+  (wp: reschedule_valid_sched_const)
+
+crunches maybe_return_sc
```
MitchellBuckley, line 12811: Merge up to IPC crunches

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -21232,6 +21276,8 @@ method cur_sc_in_release_q_imp_zero_consumed_syscall_combined
       , rule invs_strengthen_cur_sc_tcb_are_bound, blast
       , clarsimp simp: vs_all_heap_simps)
 
+find_theorems scheduler_act_sane schact_is_rct
```
MitchellBuckley, line 21279: remove find_theorems

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -21136,7 +21178,9 @@ lemma handle_recv_cur_sc_in_release_q_imp_zero_consumed[wp]:
         \<and> invs s
         \<and> cur_sc_more_than_ready s
         \<and> current_time_bounded (Suc 0) s
-        \<and> ct_not_in_release_q s\<rbrace>
+        \<and> ct_not_in_release_q s \<and>
+              scheduler_act_sane s \<and>
```
MitchellBuckley, line 21182: indentation

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -21089,32 +21138,25 @@ lemma reply_push_cur_sc_in_release_q_imp_zero_consumed[wp]:
...
    \<lbrace>\<lambda>_. cur_sc_in_release_q_imp_zero_consumed :: det_state \<Rightarrow> _\<rbrace>"
   unfolding receive_ipc_def
-  apply wpsimp
+  apply (wpsimp wp: thread_get_wp')
                 apply (wpsimp wp: hoare_vcg_all_lift hoare_drop_imp)
                apply (wpsimp wp: hoare_vcg_all_lift hoare_drop_imp)
-              apply wpsimp
              apply (wpsimp wp: gts_wp)
-            apply (rule_tac Q="\<lambda>_. cur_sc_in_release_q_imp_zero_consumed and valid_release_q"
+            apply (rule_tac Q="\<lambda>_ s. cur_sc_in_release_q_imp_zero_consumed s \<and> valid_release_q s"
```
MitchellBuckley, line 21150: undo this change

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -13018,6 +13079,20 @@ crunches receive_ipc_preamble, refill_unblock_check
...
+lemma fhfhfhhfh[wp]:
+  "\<lbrace>not_queued thread
+    and scheduler_act_sane\<rbrace>
+   maybe_return_sc y thread
+   \<lbrace>\<lambda>rv. not_queued thread\<rbrace>"
+  unfolding maybe_return_sc_def
+  by (wpsimp wp: reschedule_required_not_queued get_tcb_obj_ref_wp get_sk_obj_ref_wp)
+
+crunches maybe_return_sc
+  for release_queue[wp]: "\<lambda>s. P (release_queue s)"
```
MitchellBuckley, line 13091: Potentially move up some these crunches to IPC.

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -13018,6 +13079,20 @@ crunches receive_ipc_preamble, refill_unblock_check
   and cur_time[wp]: "\<lambda>s. P (cur_time s)"
   (wp: crunch_wps)
 
+lemma fhfhfhhfh[wp]:
```
MitchellBuckley, line 13082: rename

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -12804,6 +12804,67 @@ lemma set_thread_state_pred_map_eq[wp]:
...
+   sorted_release_q_2 (\<lambda>a. if a = tptr then None else tcb_sc_refill_cfgs_of s a) (release_queue s)
+   = sorted_release_q_2 (tcb_sc_refill_cfgs_of s) (release_queue s)"
+  by (rule sorted_release_q_2_eq_lift, clarsimp simp: in_release_q_def sc_ready_times_upd)
+
+lemma maybe_return_sc_valid_sched[wp]:
+  "\<lbrace> valid_sched and scheduler_act_not tptr and not_queued tptr and not_in_release_q tptr\<rbrace>
+   maybe_return_sc ntfnptr tptr
+   \<lbrace> \<lambda>_. valid_sched\<rbrace>"
+  apply (simp add: maybe_return_sc_def assert_opt_def)
+find_theorems reschedule_required valid_sched
```
MitchellBuckley, line 12842: Clean up this proofs

proof/invariant-abstract/IpcDet_AI.thy
MitchellBuckley: True, done.

spec/haskell/src/SEL4/Object/SchedContext.lhs
MitchellBuckley: Cheers!

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -12803,6 +12803,64 @@ lemma set_thread_state_pred_map_eq[wp]:
...
 
+crunches schedule_tcb
+  for valid_sched[wp]: valid_sched
+  (wp: reschedule_valid_sched_const)
+
+crunches maybe_return_sc
+  for etcbs_of[wp]: "\<lambda>s. P (etcbs_of s)"
+  (simp: wp: hoare_drop_imp hoare_vcg_if_lift2 ignore: set_tcb_obj_ref)
+
+(* FIXME: move somewhere appropriate *)
```
lsf37, line 12814:
```suggestion
(* FIXME RT: move somewhere appropriate *)
```

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -12803,6 +12803,64 @@ lemma set_thread_state_pred_map_eq[wp]:
...
+lemma tcb_sc_refill_cfgs_upd:
+  "tcb_sc_refill_cfgs_2 (\<lambda>a. if a = tptr then Some None else tcb_scps_of s a) (sc_refill_cfgs_of s)
+   = (\<lambda>a. if a = tptr then None else tcb_sc_refill_cfgs_2 (tcb_scps_of s) (sc_refill_cfgs_of s) a)"
+  unfolding tcb_sc_refill_cfgs_2_def
+  by (fastforce simp: map_join_def vs_all_heap_simps opt_map_def)
+
+(* FIXME: move somewhere appropriate *)
```
lsf37, line 12821:
```suggestion
(* FIXME RT: move somewhere appropriate *)
```

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -12803,6 +12803,64 @@ lemma set_thread_state_pred_map_eq[wp]:
...
+lemma sc_ready_times_upd:
+  "sc_ready_times_2 (\<lambda>a. if a = tptr then None else tcb_sc_refill_cfgs_of s a)
+   = (\<lambda>a. if a = tptr then None else sc_ready_times_2 (tcb_sc_refill_cfgs_of s) a)"
+  unfolding sc_ready_times_2_def
+  by (clarsimp simp: map_project_upd)
+
+(* FIXME: move somewhere appropriate, maybe rename *)
```
lsf37, line 12828:
```suggestion
(* FIXME RT: move somewhere appropriate, maybe rename *)
```

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -21248,9 +21287,10 @@ lemma handle_event_cur_sc_in_release_q_imp_zero_consumed:
   apply (cases e; simp)
 
   subgoal for syscall
-    by (case_tac syscall; simp
+    apply (case_tac syscall; simp
         ; cur_sc_in_release_q_imp_zero_consumed_syscall_single?
```
lsf37, line 21291: The indent doesn't fit any more to the `apply` (previously `by`)

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -21248,9 +21287,10 @@ lemma handle_event_cur_sc_in_release_q_imp_zero_consumed:
...
           , cur_sc_in_release_q_imp_zero_consumed_syscall_combined?) \<comment> \<open>takes 30 seconds or so\<close>
+    by fastforce+
```
lsf37, line 21293: Since this has a `takes long` comment, I'd recommend to use the intial-method/final-method form of `by`, so that interactive proof progress isn't stuck at this point, i.e

```
by (case_tac ...; ..)
     fastforce+
```

spec/abstract/SchedContext_A.thy
```diff
@@ -423,7 +423,9 @@ where
     when (nsc_opt = tsc_opt \<and> nsc_opt \<noteq> None) $ do
       sc_ptr \<leftarrow> assert_opt nsc_opt;
       set_tcb_obj_ref tcb_sched_context_update tcb_ptr None;
-      set_sc_obj_ref sc_tcb_update sc_ptr None
+      set_sc_obj_ref sc_tcb_update sc_ptr None;
+      ct \<leftarrow> gets cur_thread;
+      when (tcb_ptr = ct) $ reschedule_required
```
lsf37, line 428:
```suggestion
      when (tcb_ptr = ct) reschedule_required
```

proof/invariant-abstract/IpcDet_AI.thy
```diff
@@ -1096,7 +1270,8 @@ lemma ri_invs[wp]:
   "\<lbrace>\<lambda>s. invs s \<and> st_tcb_at active thread s
                \<and> ex_nonz_cap_to thread s
                \<and> (\<forall>r\<in>zobj_refs ep_cap. ex_nonz_cap_to r s)
-               \<and> (\<forall>r\<in>zobj_refs reply. ex_nonz_cap_to r s)\<rbrace>
+               \<and> (\<forall>r\<in>zobj_refs reply. ex_nonz_cap_to r s)
+               \<and> thread = cur_thread s\<rbrace>
```
lsf37, line 1274: Since this lemma is `[wp]`: are we sure `thread = cur_thread s` is the only situation in which this function is used?

proof/invariant-abstract/IpcDet_AI.thy
```diff
@@ -1086,6 +1086,180 @@ lemma reply_push_sender_sc_Some_invs:
...
+  supply set_scheduler_action_cur_sc_tcb [wp del]
+  unfolding reschedule_required_def
+  by (wpsimp simp: reschedule_required_def set_scheduler_action_def tcb_sched_action_def
+                   set_tcb_queue_def get_tcb_queue_def thread_get_def is_schedulable_def
+                   cur_sc_tcb_def sc_tcb_sc_at_def obj_at_def)
+
+lemma maybe_return_sc_cur_sc_tcb[wp]:
+  "\<lbrace>cur_sc_tcb and (\<lambda>s. thread = cur_thread s)\<rbrace>
+    maybe_return_sc y thread
+   \<lbrace>\<lambda>_. cur_sc_tcb\<rbrace>"
```
mbrcknl, line 1233: I guess this is where the new `thread = cur_thread s` precondition comes from in `ri_invs`.

I think it might be possible to remove this precondition, and replace it with `sym_refs`:
- If the `scheduler_action` is *not* `resume_cur_thread`, then the postcondition is satisfied as long as there is a scheduling context at `cur_sc`, which you get from the precondition. No problem here.
- If the `scheduler_action` *is* `resume_cur_thread`, and `thread` is `cur_thread`, then the `reschedul_required` means that the postcondition is similarly satisfied. Still no problem.
- But if the `scheduler_action` is `resume_cur_thread`, and `thread` is *not* `cur_thread`, then you would need some extra precondition to establish that the scheduling context that's updated is not the `cur_sc`. I think `sym_refs` would give you that, since you know that the `cur_sc` is bound to `cur_thread`, and `thread` is bound to the scheduling context that's updated, but `thread` is not `cur_thread`.

I can see why you did this, though: `maybe_return_sc` is only called from `receive_signal` and now `receive_ipc`, which are only called on the `cur_thread`. And that's easier to reason about than `sym_refs`.

Even so, it might be worth thinking about whether it's worth replacing the precondition. If not, consider whether `[wp]` is appropriate for such a strong precondition.

Finally, if we're committing to only calling `maybe_return_sc` on the `cur_thread`, why are we testing for this in the implementation? (As opposed to asserting it.)

mbrcknl: Also, I think this is obscure enough to warrant a comment on why this precondition is needed, especially if you don't change the precondition to `sym_refs`.

Also also, meaningful variable names (here and in some other lemmas).

MitchellBuckley:
Thanks for all the comments @lsf37 and @mbrcknl!

proof/invariant-abstract/IpcDet_AI.thy
MitchellBuckley: I like that `sym_refs` seems strong enough but it might be harder to use in proofs that use this wp rule.

I think the best option here is to remove `[wp]`, keep the existing precondition, and add some comments to explain. Sound ok?

mbrcknl: I was going to say you could use `heap_refs_retract_at`, but that won't be available in `IpcDet_AI`. :-(

I'll leave the decision to you, but I'm pretty sure you have all of `invs` at the point you need this, so I don't think `sym_refs` should cause any trouble.

I still think it would also be good to look into why the `reschedule_required` in `maybe_return_sc` is conditional on whether `thread` is the current thread. This lemma raises the question of whether we ever call it on anything other than the current thread. From what I can tell, we don't. In that case, the `thread = cur_thread s` could be an assertion instead of a precondition, and the `reschedule_required` could be unconditional. Refine will force us to (easily) prove the assertion, and we'll get the assurance that it's only called in the way we expect. I'll make a separate comment on that bit of the patch.

spec/abstract/SchedContext_A.thy
```diff
@@ -423,7 +423,9 @@ where
     when (nsc_opt = tsc_opt \<and> nsc_opt \<noteq> None) $ do
       sc_ptr \<leftarrow> assert_opt nsc_opt;
       set_tcb_obj_ref tcb_sched_context_update tcb_ptr None;
-      set_sc_obj_ref sc_tcb_update sc_ptr None
+      set_sc_obj_ref sc_tcb_update sc_ptr None;
+      ct \<leftarrow> gets cur_thread;
+      when (tcb_ptr = ct) reschedule_required
```
mbrcknl, line 428: As far as I can tell, we only ever call this on the current thread. Can we make the `reschedule_required` unconditional, and turn this check into an assertion instead?

mbrcknl (approved):

proof/invariant-abstract/IpcDet_AI.thy
mbrcknl: Oh, and don't forget to give things names that are somewhat more meaningful than `y`! :-)

MitchellBuckley: Thanks! You're right on both counts: the check is not needed at all and replacing the precondition with sym_refs seems to work just fine.

mbrcknl (approved):
👍  to updates.

Also, 👍  to updating pull requests by adding new commits, like I think you did here. It makes it much easier to see what was updated since last review, especially since GitHub has some nice UI for that. It's still fine to squash just before you merge, of course.

PR #46, opened 2020-08-15
sc replies relation

mktnk3:
This commits flips the traversal direction of `sc_replies_relation`, to follow the `replyPrev` links instead of the `replyNext` links.

This means that we have the following:
- `reply_relation reply reply'` gives `reply_sc reply = replySc reply’”
- `sc_replies_relation s s'` gives 1)`hd (sc_replies sc) = scReply sc’`; and 2) `sc_replies sc` equals to the `replyPrev` traversal of the linked list starting from `scReply sc’`
- we need `sym_refs (list_refs_of_replies’ s') to know that the `replyNext` traversal matches the `replyPrev` traversal

This is a minimal and necessary change to `sc_replies_relation` (meaning this should be merged). 

There is also a question of whether it helps or makes it more painful to include the “Next” traversal in the relation or not. My plan is to experiment a bit more as I continue working on `reply_remove_tcb`, and it should go into a separate PR if there should be a further change.

### DIFF
--- a/proof/refine/ARM/ADT_H.thy
+++ b/proof/refine/ARM/ADT_H.thy
@@ -458,7 +458,7 @@ definition absCNode :: "nat \<Rightarrow> (obj_ref \<rightharpoonup> kernel_obje
                                      else None)"
 
 definition scMap :: "(obj_ref \<rightharpoonup> obj_ref) \<Rightarrow> sched_context \<Rightarrow> Structures_A.sched_context" where
-  "scMap replyNexts sc = \<lparr>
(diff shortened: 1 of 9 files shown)


mktnk3:
> For the other direction, if a formulation with heap walk or a direct `x -> next -> prev = x` in the state relation would be useful sometimes, we can probably achieve almost the same directness by encoding this into a lemma like `[| sc_replies_relation s s'; sym_refs ... |] ==> next/prev property`

Yes, something like that would be handy. I’ll do that.

PR #47, opened 2020-08-16
Add tcb arg to replyUnlink etc.

mktnk3:
This updates Haskell replyUnlink, replyRemove, replyPop, and replyClear to take a tcb ref argument, which should be appropriately linked to the first argument.

This corresponds to the C change in seL4 [PR#1808](https://bitbucket.ts.data61.csiro.au/projects/SEL4/repos/sel4/pull-requests/1808/overview) (on the internal bitbucket).

### DIFF
--- a/proof/refine/ARM/IpcCancel_R.thy
+++ b/proof/refine/ARM/IpcCancel_R.thy
@@ -63,8 +63,7 @@ end
 lemma cancelIPC_simple[wp]:
   "\<lbrace>\<top>\<rbrace> cancelIPC t \<lbrace>\<lambda>rv. st_tcb_at' simple' t\<rbrace>"
   unfolding cancelIPC_def
-  apply (wpsimp wp: setThreadState_st_tcb replyRemoveTCB_st_tcb_at'_Inactive cancelSignal_simple
-                    replyUnlink_st_tcb_at'_wp set_reply'.set_wp gts_wp' threadSet_wp
+  apply (wpsimp wp: setThreadState_st_tcb gts_wp' threadSet_wp
               simp: Let_def tcb_obj_at'_pred_tcb'_set_obj'_iff)
   apply (clarsimp simp: st_tcb_at'_def o_def obj_at'_def isBlockedOnReply_def)
   done
@@ -112,6 +111,7 @@ lemma cancelSignal_st_tcb_at':
                     hoare_drop_imp[where R="\<lambda>rv _. \<exists>a b. delete t (f rv) = a # b" for f])
   done
 
+(* FIXME RT: Michael to fix and use in his next PR
 lemma cancelIPC_simple'_not_awaiting_reply:
   "\<lbrace>\<lambda>s. sym_refs (state_refs_of' s)\<rbrace>
    cancelIPC t
@@ -129,7 +129,7 @@ lemma cancelIPC_simple'_not_awaiting_reply:
    apply (frule(1) sym_ref_Receive_or_Reply_replyTCB', fast)
    apply (clarsimp simp: obj_at'_def projectKO_eq project_inject)
   apply (clarsimp simp: pred_tcb_at'_def obj_at'_def)
-  done
+  done*)
 
 context begin interpretation Arch .
 crunch typ_at'[wp]: emptySlot "\<lambda>s. P (typ_at' T p s)"
@@ -165,7 +165,7 @@ lemma gts_wf''[wp]:
 
 lemma replyUnlink_valid_objs':
   "\<lbrace>valid_objs'\<rbrace>
-   replyUnlink rptr
+   replyUnlink rptr tptr
    \<lbrace>\<lambda>_. valid_objs'\<rbrace>"
   apply (clarsimp simp: replyUnlink_def setReplyTCB_def getReplyTCB_def liftM_def)
   apply (wpsimp wp: set_reply_valid_objs' hoare_vcg_all_lift gts_wp'
@@ -198,7 +198,7 @@ lemma reply_unlink_tcb_corres:
      (valid_objs and pspace_aligned and pspace_distinct
        and st_tcb_at ((=) st) t and reply_tcb_reply_at ((=) (Some t)) rp)
           (valid_objs' and valid_release_queue_iff)
-        (reply_unlink_tcb t rp) (replyUnlink rp)" (is "_ \<Longrightarrow> corres _ _ ?conc_guard _ _")
+        (reply_unlink_tcb t rp) (replyUnlink rp t)" (is "_ \<Longrightarrow> corres _ _ ?conc_guard _ _")
   apply (rule_tac Q="?conc_guard
               and st_tcb_at' (\<lambda>st. st = BlockedOnReceive ep (receiver_can_grant pl) (Some rp)
                                  \<or> st = BlockedOnReply (Some rp)) t" in corres_cross_over_guard)
@@ -253,7 +253,7 @@ lemma blocked_cancel_ipc_corres:
                else
                  return $ epQueue_update (%_. (remove1 t (epQueue ep))) ep;
                y \<leftarrow> setEndpoint epPtr ep';
-               case reply_opt of None \<Rightarrow> return () | Some reply \<Rightarrow> replyUnlink reply;
+               case reply_opt of None \<Rightarrow> return () | Some reply \<Rightarrow> replyUnlink reply t;
                setThreadState Structures_H.thread_state.Inactive t
             od)" (is "\<lbrakk> _ ; _ ; _ \<rbrakk> \<Longrightarrow> corres _ (?abs_guard and _) (?conc_guard and _) _ _")
   apply (simp add: blocked_cancel_ipc_def gbep_ret)
@@ -845,7 +845,7 @@ crunches setReply, setSchedContext
   (simp: crunch_simps sch_act_simple_def wp: crunch_wps)
 
 lemma replyUnlink_sch_act_simple[wp]:
-  "replyUnlink t' \<lbrace>sch_act_simple\<rbrace>"
+  "replyUnlink r t' \<lbrace>sch_act_simple\<rbrace>"
   apply (simp add: replyUnlink_def setReplyTCB_def getReplyTCB_def)
   by (wpsimp wp: hoare_drop_imps)
 
@@ -1313,7 +1313,7 @@ lemma setThreadState_inQ[wp]:
   done
 
 lemma replyUnlink_valid_inQ_queues[wp]:
-  "replyUnlink replyPtr \<lbrace>valid_inQ_queues\<rbrace>"
+  "replyUnlink replyPtr tcbPtr \<lbrace>valid_inQ_queues\<rbrace>"
   apply (clarsimp simp: replyUnlink_def setReplyTCB_def getReplyTCB_def)
   apply (wpsimp wp: set_reply'.set_wp gts_wp')
   apply (fastforce simp: valid_inQ_queues_def obj_at'_def projectKOs)
@@ -2191,7 +2191,7 @@ lemma cancelSignal_queues[wp]:
   done
 
 lemma replyUnlink_valid_queues[wp]:
-  "replyUnlink replyPtr \<lbrace>valid_queues\<rbrace>"
+  "replyUnlink replyPtr tcbPtr \<lbrace>valid_queues\<rbrace>"
   apply (clarsimp simp: replyUnlink_def setReplyTCB_def getReplyTCB_def)
   apply (wpsimp wp: sts_valid_queues gts_wp')
   apply (fastforce simp: replyUnlink_assertion_def valid_queues_def valid_queues_no_bitmap_def
@@ -2231,7 +2231,7 @@ lemma cancelIPC_queues[wp]:
    apply (fastforce simp: st_tcb_at'_def obj_at'_def projectKOs objBitsKO_def
                    intro: tcbFault_update_valid_queues)
   apply (wpsimp wp: setThreadState_not_runnable'_valid_queues hoare_vcg_all_lift hoare_drop_imps
-                    replyUnlink_st_tcb_at'_Inactive)
+                    replyUnlink_st_tcb_at')
   apply (fastforce simp: st_tcb_at'_def obj_at'_def)
   done
 
@@ -2664,7 +2664,7 @@ lemma cancel_all_invs'_helper:
                               else state_refs_of' s x)
               \<and>  (\<forall>x \<in> set q. ex_nonz_cap_to' x s))\<rbrace>
    mapM_x (\<lambda>t. do st <- getThreadState t;
-                  y <- case if isReceive st then replyObject st else None of None \<Rightarrow> return () | Some x \<Rightarrow> replyUnlink x;
+                  y <- case if isReceive st then replyObject st else None of None \<Rightarrow> return () | Some x \<Rightarrow> replyUnlink x t;
                   fault <- threadGet tcbFault t;
                   if fault = None then do y <- setThreadState Structures_H.thread_state.Restart t;
                                           possibleSwitchTo t
@@ -2941,7 +2941,7 @@ lemma setThreadState_valid_ep'[wp]:
   done
 
 lemma replyUnlink_valid_ep'[wp]:
-  "replyUnlink replyPtr \<lbrace>valid_ep' ep\<rbrace>"
+  "replyUnlink replyPtr tcbPtr \<lbrace>valid_ep' ep\<rbrace>"
   apply (clarsimp simp: replyUnlink_def setReplyTCB_def getReplyTCB_def)
   apply (wpsimp wp: set_reply'.set_wp gts_wp')
   apply (fastforce simp: valid_ep'_def obj_at'_def projectKOs objBitsKO_def split: endpoint.splits)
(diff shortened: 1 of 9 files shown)


mktnk3:
Given that many of us have been working on replyUnlink related things, and this PR can potentially help us go over the bump, shall we consider merging this first, maybe? Assuming that there is no issue with the C change, which should be ok.

proof/refine/ARM/IpcCancel_R.thy
```diff
@@ -112,6 +112,7 @@ lemma cancelSignal_st_tcb_at':
                     hoare_drop_imp[where R="\<lambda>rv _. \<exists>a b. delete t (f rv) = a # b" for f])
   done
 
+(* Not used?
```
pierzchalski, line 115: @michaelmcinerney did you still need this?

proof/refine/ARM/Reply_R.thy
```diff
@@ -8,31 +8,28 @@ theory Reply_R
...
 
-lemma replyUnlink_st_tcb_at'_wp:
-  "\<lbrace>\<lambda>s. reply_at' rptr s \<longrightarrow>
-          obj_at' (\<lambda>reply.
-                      (replyTCB reply = Some tptr \<longrightarrow> test Inactive) \<and>
-                      (replyTCB reply \<noteq> Some tptr \<longrightarrow> st_tcb_at' test tptr s))
-                  rptr s\<rbrace>
-   replyUnlink rptr
-   \<lbrace>\<lambda>_. st_tcb_at' test tptr\<rbrace>"
+lemma replyUnlink_st_tcb_at':
```
pierzchalski, line 15: 👌

proof/refine/ARM/Reply_R.thy
```diff
@@ -41,25 +38,30 @@ lemma replyUnlink_st_tcb_at'_sym_ref:
...
   done
+*)
 
-lemma replyRemoveTCB_st_tcb_at'_Inactive:
-  "\<lbrace>\<lambda>s. tcb_at' tptr s \<longrightarrow> st_tcb_at' test tptr s \<and> test Inactive\<rbrace>
+lemma replyRemoveTCB_st_tcb_at'_Inactive':
+  "\<lbrace>\<top>\<rbrace>
    replyRemoveTCB tptr
-   \<lbrace>\<lambda>_. st_tcb_at' test tptr\<rbrace>"
+   \<lbrace>\<lambda>_. st_tcb_at' ((=) Inactive) tptr\<rbrace>"
```
pierzchalski, line 46: Isn't this going to cause issues with goals like `st_tcb_at' (Not o runnable')` or whatever?

proof/refine/ARM/TcbAcc_R.thy
```diff
@@ -5017,6 +5017,13 @@ lemma sts_st_tcb_at'_cases:
   apply fastforce
   done
 
+lemma sts_st_tcb_at'_cases_strong:
+  "\<lbrace>\<lambda>s. tcb_at' t s \<longrightarrow> (t = t' \<longrightarrow> P (P' ts)) \<and> (t \<noteq> t' \<longrightarrow> P (st_tcb_at' P' t' s))\<rbrace>
+     setThreadState ts t
```
pierzchalski, line 5022: Indentation.

proof/refine/ARM/TcbAcc_R.thy
```diff
@@ -2789,7 +2789,7 @@ lemma threadSet_queued_sch_act_wf[wp]:
   done
 
 lemma tcbSchedEnqueue_pred_tcb_at'[wp]:
-  "\<lbrace>\<lambda>s. pred_tcb_at' proj P' t' s \<rbrace> tcbSchedEnqueue t \<lbrace>\<lambda>_ s. pred_tcb_at' proj P' t' s\<rbrace>"
+  "\<lbrace>\<lambda>s. P (pred_tcb_at' proj P' t' s) \<rbrace> tcbSchedEnqueue t \<lbrace>\<lambda>_ s. P (pred_tcb_at' proj P' t' s)\<rbrace>"
```
pierzchalski, line 2792: Should this use the invariant syntax?

proof/refine/ARM/Reply_R.thy
```diff
@@ -8,31 +8,28 @@ theory Reply_R
...
 lemma replyUnlink_st_tcb_at'_Inactive:
   "\<lbrace>\<lambda>s. st_tcb_at' test tptr s \<and> test Inactive\<rbrace>
-   replyUnlink rptr
+   replyUnlink rptr tptr'
    \<lbrace>\<lambda>_. st_tcb_at' test tptr\<rbrace>"
-  apply (wpsimp wp: replyUnlink_st_tcb_at'_wp)
+  apply (wpsimp wp: replyUnlink_st_tcb_at')
   done
 
+(* Not used
```
MitchellBuckley, line 32: Maybe remove this.

proof/refine/ARM/Reply_R.thy
MitchellBuckley: Agreed, probably good to leave the more general result in there.

proof/refine/ARM/TcbAcc_R.thy
MitchellBuckley: I would also use `if then else` but that's just a matter of stlye.

spec/haskell/src/SEL4/Object/Reply.lhs
```diff
@@ -155,17 +156,18 @@ This module specifies the behavior of reply objects.
...
+>     replyUnlink rptr tptr
 
-> replyUnlink :: PPtr Reply -> Kernel ()
-> replyUnlink replyPtr = do
+> replyUnlink :: PPtr Reply -> PPtr TCB -> Kernel ()
+> replyUnlink replyPtr tcbPtr = do
 >     tptrOpt <- getReplyTCB replyPtr
 >     tptr <- maybeToMonad tptrOpt
->     state <- getThreadState tptr
+>     assert (tptr == tcbPtr) "replyTCB is equal to tcbPtr"
```
MitchellBuckley, line 165: Is this assertion necessary?

proof/refine/ARM/TcbAcc_R.thy
pierzchalski: I think the rough consensus is to use `(P -> X) /\ (!P - Y)` for syntactically small `P`, since it means you don't need to add another lifting lemma to your `wp` proofs.

spec/haskell/src/SEL4/Object/Reply.lhs
pierzchalski: It would certainly make showing `sym_refs` easier.

proof/refine/ARM/Reply_R.thy
mktnk3: Well, actually, I think it makes sense to leave these commented out. The changes here will affect the ongoing proofs, and depending on how that goes, these might not be needed at all or there could be some use for them. We might want to wait.

proof/refine/ARM/IpcCancel_R.thy
michaelmcinerney: Yes this is needed for `suspend_invs'`. I'll update this proof to use this soon

mktnk3: Ok, so, it would be ok to leave these for you to fix up with your `suspend_invs’` proof, right? @michaelmcinerney ?

michaelmcinerney: Sure. I can make a PR just for this, but I'd prefer to just include it next time I make a PR for something else

mktnk3: Thanks, I’ll add some comments to note that. I think it should be ok if you include these with your `suspend_ivns’` proof.

proof/refine/ARM/Reply_R.thy
pierzchalski: Out of curiosity, in what sense is it fiddly?

mktnk3: I don’t quite remember, sorry. It could have been just being non-trivial, but we don’t really need this form anyway, at least at the moment. And I am not sure if the original lemma is actually more general, honestly.

proof/refine/ARM/TcbAcc_R.thy
lsf37: In general, the implication form is nicer than `if` because Isabelle likes to split `if`s by default and pull out the implication to the largest scope, which may be outside the hoare triple (that's why we see a lot of `split del: if_split` in `wp` proofs). For this one here that is likely to happen, because both `t` and `t'` are in scope outside the precondition. 

Only when `P` becomes so large that it's annoying to repeat I'd opt for `if`.

proof/refine/ARM/Reply_R.thy
```diff
@@ -8,31 +8,28 @@ theory Reply_R
...
-lemma replyUnlink_st_tcb_at'_wp:
-  "\<lbrace>\<lambda>s. reply_at' rptr s \<longrightarrow>
-          obj_at' (\<lambda>reply.
-                      (replyTCB reply = Some tptr \<longrightarrow> test Inactive) \<and>
-                      (replyTCB reply \<noteq> Some tptr \<longrightarrow> st_tcb_at' test tptr s))
-                  rptr s\<rbrace>
-   replyUnlink rptr
-   \<lbrace>\<lambda>_. st_tcb_at' test tptr\<rbrace>"
+lemma replyUnlink_st_tcb_at':
+  "\<lbrace>\<lambda>s. tcb_at' t s \<longrightarrow> (if t'=t then P (P' Inactive) else P (st_tcb_at' P' t' s))\<rbrace>
```
lsf37, line 16: I'd also go for the implication form here instead of the `if`

proof/refine/ARM/Reply_R.thy
lsf37: Would it make sense to leave the statement in, but oops the lemma? Might make it slightly easier to find and at least be type checked (although it's not going to turn up in `find_theorems`, so maybe not that important).

In any case, commented-out blocks should get a `FIXME RT` so that we remember to remove them if we didn't need them in the end.

proof/refine/ARM/Reply_R.thy
lsf37: We could keep the general form, but instantiate and `[wp]` the `Inactive` form with `lemmas`

proof/refine/ARM/Reply_R.thy
```diff
@@ -72,6 +74,7 @@ lemma replyUnlink_tcb_obj_at'_no_change:
               simp: o_def)
   done
 
+(* Not used
```
lsf37, line 77: Add `FIXME RT`

spec/haskell/src/SEL4/Object/Reply.lhs
```diff
@@ -100,25 +99,27 @@ This module specifies the behavior of reply objects.
...
 
-> replyRemove :: PPtr Reply -> Kernel ()
-> replyRemove replyPtr = do
+> replyRemove :: PPtr Reply -> PPtr TCB -> Kernel ()
+> replyRemove replyPtr tcbPtr = do
 >     reply <- getReply replyPtr
->     tcbPtrOpt <- return $ replyTCB reply
->     state <- getThreadState $ fromJust tcbPtrOpt
+>     tptr <- maybeToMonad  $ replyTCB reply
+>     assert (tptr == tcbPtr) "replyRemove: replyTCB is equal to tcbPtr"
```
lsf37, line 106: nitpick:
```suggestion
>     assert (tptr == tcbPtr) "replyRemove: replyTCB is not equal to tcbPtr"
```

spec/haskell/src/SEL4/Object/Reply.lhs
lsf37: If the `assert` is proof-motivated, it looks fine (it does look like it would help with `sym_refs`). If it is just there because there is one in C, we should remove it.

Oh, also the message should be about the failed assert, i.e.
```suggestion
>     assert (tptr == tcbPtr) "replyTCB is not equal to tcbPtr"
```

or

```suggestion
>     assert (tptr == tcbPtr) "replyTCB must be equal to tcbPtr"
```

spec/haskell/src/SEL4/Object/ObjectType.lhs
```diff
@@ -133,7 +133,9 @@ When the last capability to an endpoint is deleted, any IPC operations currently
 > finaliseCap (ReplyCap { capReplyPtr = ptr }) final _ = do
 >     when final $ do
 >         tptrOpt <- getReplyTCB ptr
->         when (tptrOpt /= Nothing) $ replyClear ptr
+>         when (tptrOpt /= Nothing) $ do
+>             let tptr = fromJust tptrOpt
+>             replyClear ptr tptr
```
lsf37, line 138: If it's only used once, I'd probably inline the `let` unless there is some other reason (e.g. to line it up with something in C or abstract), i.e.

```suggestion
>             replyClear ptr (fromJust tptrOpt)
```
(and removing the `let` above)

proof/refine/ARM/IpcCancel_R.thy
lsf37: If you're sure @michaelmcinerney is going to uncomment it again in the next PR I'm fine without a `FIXME RT`, but if there's a chance we might forget, I'd still tag it.

proof/refine/ARM/Reply_R.thy
mktnk3: But the preconditions are different. ..

spec/haskell/src/SEL4/Object/Reply.lhs
mktnk3: > If the `assert` is proof-motivated, it looks fine (it does look like it would help with `sym_refs`). If it is just there because there is one in C, we should remove it.

At this point, I alone can’t conclude either way. Whoever actually is going to use this soon would have better ideas, and we can take it from there. Also, I am the one who added it to the C as well at this point. 

> Oh, also the message should be about the failed assert, i.e.

Right, I was wondering about that, and got confused at some point! I’ll fix those up.

proof/refine/ARM/Reply_R.thy
mktnk3: Yes, I've talked to Michael and will update those with more info.

proof/refine/ARM/TcbAcc_R.thy
mktnk3: I know that it is better to do that, but that’s more for when you are writing a new lemma from scratch, I’d think? I can of course change this, too, but I am not going so strongly in the clean up everything I touch mode, particularly for this one...

proof/refine/ARM/Reply_R.thy
lsf37: > In the old theorem, I suspect the st_tcb_at' part of the precondition is not necessary. Perhaps the best of both worlds might be to keep the test, but remove the st_tcb_at' from the precondition.

I think that is correct, yes.

proof/refine/ARM/TcbAcc_R.thy
lsf37: You're right, it's not a huge issue. I tend to try to do them, but I also don't do all of them. Up to you.

proof/refine/ARM/Reply_R.thy
mktnk3: Well, the lemma I wrote can’t be written as an instantiation of the original one. There is something a bit odd about the original lemma, actually. `replyUnlink t` sets the thread state of `t` to `Inactive` at the end. And `t` is in `BlockedOnReply` at the beginning of the function. So, the original lemma is true only for `test` that happens to be true for both `BlockedOnReply` and `Inactive`. So, it cannot be instantiated to `(=) Inactive`.

This original one is not really general because we do want all `P` that holds for `Inactive` only in the conclusion, regardless of the precondition (without being constrained by `BlockedOnReply`).

And this is only used in
```
lemma cancelIPC_simple[wp]:
   "\<lbrace>\<top>\<rbrace> cancelIPC t \<lbrace>\<lambda>rv. st_tcb_at' simple' t\<rbrace>”
```
which happens to have `simple’` as `test`.

mktnk3: And, really, the `wp` rule Matt is talking about is just the same thing as combining the lemma I wrote with `pred_tcb’_weakenE`.

proof/refine/ARM/Reply_R.thy
```diff
@@ -8,31 +8,28 @@ theory Reply_R
...
   unfolding replyUnlink_def
-  supply if_split[split del]
-  apply (wpsimp wp: sts_st_tcb' hoare_vcg_if_lift2 hoare_vcg_imp_lift
-                    hoare_vcg_disj_lift threadSet_pred_tcb_at_state gts_wp'
-              simp: getReplyTCB_def cong: conj_cong disj_cong)
-  by (fastforce simp: obj_at'_def)
+  apply (wpsimp simp: getReplyTCB_def
+                  wp: sts_st_tcb_at'_cases_strong gts_wp' hoare_vcg_imp_lift
+                cong: conj_cong split: if_split_asm)
+done
```
VictorPhan, line 23: indent

proof/refine/ARM/Reply_R.thy
```diff
@@ -41,25 +31,30 @@ lemma replyUnlink_st_tcb_at'_sym_ref:
...
   unfolding replyRemoveTCB_def
-  apply (wpsimp wp: hoare_drop_imp (* FIXME: dangerous, but necessary to avoid the goal blowing
-                                             up in size *)
-                    hoare_vcg_if_lift hoare_vcg_all_lift replyUnlink_st_tcb_at'_Inactive
-              simp: cleanReply_def
-         split_del: if_split)
+  apply (wpsimp wp: replyUnlink_st_tcb_at')
+  apply (clarsimp simp: pred_tcb_at'_def)
+  done
+
```
mktnk3, line 44: This is the lemma with `(=) Inactive` and `pred_tcb’_weakenE`. When the `t` in the postcondition is the same as the argument to `replyRemoveTCB`, which was the case in the original lemma, this is what we need. And this is I believe the same thing as what Matt was saying.

PR #48, opened 2020-08-18
Prove that the C state relation is empty, then make it non-empty

mbrcknl:
**Reviewers**: You definitely want to review the commits separately, since the second commit reverts the first.

#### crefine: prove that the C state relation is empty

As currently defined, the C state relation is empty, and consequently, `ccorres` is trivially true for any pair of functions. This means that, in a very technical sense, our C refinement proofs are meaningless.

The state relation is empty because several conjuncts in `cstate_relation` form a contradiction:
- Two conjuncts claim that `intStateIRQNode_array_Ptr` points to a heap object within the set of addresses `kernel_data_refs`.
- Another implies that all heap objects are within `domain`.
- Another claims that `kernel_data_refs = -domain`, forming the contradiction.

This commit proves the contradiction, and also proves that `ccorres` is trivially true for any pair of functions.

Fortunately, we never made any essential use of this contradiction, and so the issue can be fixed fairly easily. The issue seems to have arisen out of a conflation of two different concepts:
- `kernel_data_refs` is introduced in the intermediate specification, and is intended to be the set of addresses containing global heap objects that are not covered by capabilities.
- `domain` was introduced for binary verification, and was intended to be the set of all addresses that may be used for heap objects.

The easiest fix seems to be to expand the meaning of `kernel_data_refs` to include all addresses that are not covered by capabilities. If we assert that `kernel_data_refs = -domain`, then this does not allow for heap objects that are not covered by capabilities. If instead, we make a weaker assertion that `-domain <= kernel_data_refs`, we can have heap objects that are not covered by capabilities, such as the one pointed to by `intStateIRQNode_array_Ptr`.

This fix will be performed in a subsequent commit.

#### crefine: make C state relation non-empty

We believe this commit fixes the issue described in the previous commit. It also reverts that commit, since the proofs that the C state relation is empty no longer work.

As the previous commit demonstrated, it is important to demonstrate the non-triviality of properties. In this case, we should exhibit a witness of the non-emptiness of the C state relation. We have not yet done that.


mbrcknl:
For TS reviewers, there are two test runs:
- [L4V-TESTBOARD-588](https://bamboo.ts.data61.csiro.au/browse/L4V-TESTBOARD-588) for the first commit.
- [L4V-TESTBOARD-587](https://bamboo.ts.data61.csiro.au/browse/L4V-TESTBOARD-587) for the second commit.

lsf37:
Can you add a testboard link? Github CI isn't powerful enough yet to run all proofs on master.

mbrcknl:
> Can you add a testboard link? Github CI isn't powerful enough yet to run all proofs on master.

@lsf37 Yeah, I added a couple of testboard links in a comment a few moments before yours.

mbrcknl:
I just noticed a typo in the commit message, so will do a quick force-push.

lsf37:
> > Can you add a testboard link? Github CI isn't powerful enough yet to run all proofs on master.
> 
> @lsf37 Yeah, I added a couple of testboard links in a comment a few moments before yours.

Thanks, can see them now!

mbrcknl:
> It's nice that the proofs actually seem to be a small bit simpler in the fixed version.

Yeah, we really didn't use the extra strength of the equality at all. In every detail, the subset relation works just as well.

mbrcknl:
> * [L4V-TESTBOARD-587](https://bamboo.ts.data61.csiro.au/browse/L4V-TESTBOARD-587) for the second commit.

587 had a timeout, but [L4V-TESTBOARD-589](https://bamboo.ts.data61.csiro.au/browse/L4V-TESTBOARD-589) passed.

PR #2, opened 2014-08-05
ROOT: whitespace fix

machinaut:

PR #49, opened 2020-08-20
Rt sc replies relation #2

mktnk3:
This adds a few lemmas about `replyPrev` and `replyNext`.

This also adds opt_map_SomeD, which is
```
lemma opt_map_SomeD:
   "f x = Some y \<Longrightarrow> (f |> g) x = g y"
```
and it might be good to add this to the simp set. But that requires some work to fix proofs, so I haven’t done that yet.

I think it makes sense because, for `opt_map` and `Some`, it seems that we currently only have lemmas for cases where the whole computation returns `Some`, but even without that knowledge we should be able to proceed with the computation from the inside if we know enough to do so. Currently, if we want to do that, we effectively have to unfold `opt_map` which is often less than ideal typically because you have other `opt_map`s in the subgoal that you’d rather not unfold. Let me know if I am missing something.

Also, I realise that the lemma could do with a better name in the similar way as `get_tcb_SomeD` is not quite right.

### DIFF
--- a/proof/refine/ARM/StateRelation.thy
+++ b/proof/refine/ARM/StateRelation.thy
@@ -846,6 +846,17 @@ lemma pspace_dom_relatedE:
   apply fastforce
   done
 
+lemma ghost_relation_typ_at:
+  "ghost_relation (kheap s) ups cns \<equiv>
+   (\<forall>a sz. data_at sz a s = (ups a = Some sz)) \<and>
+   (\<forall>a n. typ_at (ACapTable n) a s = (cns a = Some n))"
+   apply (rule eq_reflection)
+   apply (clarsimp simp: ghost_relation_def typ_at_eq_kheap_obj data_at_def)
+   apply (intro conjI impI iffI allI; force)
+   done
+
+(* more replyNext/replyPrev related lemmas *)
+
 lemma sc_replies_relation_scReply:
   "\<lbrakk> sc_replies_relation s s';
      kheap s x = Some (kernel_object.SchedContext sc n);
@@ -870,7 +881,7 @@ lemma sc_replies_prevs_walk:
   apply (clarsimp simp: opt_map_def projectKO_opt_sc)
   done
 
-lemma sc_replies_relation_prev_list:
+lemma sc_replies_relation_prevs_list:
   "\<lbrakk> sc_replies_relation s s';
      kheap s x = Some (kernel_object.SchedContext sc n);
      ksPSpace s' x = Some (KOSchedContext sc')\<rbrakk>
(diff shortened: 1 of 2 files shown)


pierzchalski (approved):
Neat!

lsf37:
> This also adds opt_map_SomeD, which is

I'm surprised that we don't have that already, let me double-check.

proof/refine/ARM/StateRelation.thy
```diff
@@ -880,33 +891,66 @@ lemma sc_replies_relation_prev_list:
...
+              split: option.split_asm)
+
+(* FIXME RT: move *)
+lemma opt_map_SomeD:
+  "f x = Some y \<Longrightarrow> (f |> g) x = g y"
+  by (clarsimp simp: opt_map_Some_def)
+
+lemma replyNext_replyNexts_of_opt_map[simp]:
+  "\<lbrakk>ksPSpace s' p = Some (KOReply reply); replyNext reply = Some (Next p')\<rbrakk>
+    \<Longrightarrow> (replyNexts_of s' |> replyPrevs_of s') p = replyPrevs_of s' p'"
```
mbrcknl, line 906: Could both instances of `replyPrevs_of s'` in this theorem be replaced with some fully generalised function `f`?

mbrcknl (approved):
Looks good!

As for whether `opt_map_SomeD` should be in the `[simp]` set, I'm unsure. It's a conditional `simp` rule, so you either want it to match infrequently, or to be able to fail quickly. In projection-heavy code, it might match in quite a lot of places. Each match would require the simplifier to pause rewriting while it looks at the premise. Again, in projection-heavy code, the `f` is highly likely to be another `opt_map`, so the situation would recurse until the simplifier gets the bottom, only to discover that it can't rewrite to `Some y`, unwind the stack, and continue without the rewrite. At least it only has a single premise, and `opt_map` stacks won't be *too* deep, so maybe it's ok.

It's certainly a good rule, and will definitely useful as an explicit `simp` rule. But it might need some thought and/or experimentation before making it the default.

lsf37:
> > This also adds opt_map_SomeD, which is
> 
> I'm surprised that we don't have that already, let me double-check.

It is indeed not, as far as I can see, we seem to have gotten away with `in_opt_map_eq`, `in_opt_map_None_eq`, and `opt_map_def` so far.

As @mbrcknl says, it's a good lemma, we should have it, but it's not a good general `simp` lemma, because it fits every `|>` everywhere and the precondition will not often be provable (only where it can basically be unfolded). So it should remain an on-demand lemma. 

Thinking of the name, from `opt_map_SomeD`, I would expect a lemma of the form `(f |> g) x = Some y ==> ..`.

Maybe `opt_map_left_Some`? Which (correctly) suggests we should have a `opt_map_left_None`, which could be proved with `in_opt_map_None_eq`, but has different `simp` behaviour, and I think we make sense to keep also.

A corresponding `opt_map_right_[Some/None]` could also be written, but makes less sense, so I don't think we need those.

proof/refine/ARM/StateRelation.thy
```diff
@@ -880,33 +891,66 @@ lemma sc_replies_relation_prev_list:
...
+  by (clarsimp simp: opt_map_SomeD projectKO_opt_reply split: option.split)
+
+lemma sym_replies_next_then_prev_id:
+  "\<lbrakk>sym_refs (list_refs_of_replies' s');
+    ksPSpace s' p = Some (KOReply reply); replyNext_of reply = Some p'\<rbrakk>
```
lsf37, line 911: I would have expected a precondition of the form `replyNexts_of s' p = Some p'`. This would avoid unfolding this projection here and would fit in goals where you have the projection itself without forcing those to be unfolded as well (but with `in_opt_map_eq` in the simp set would also work for those where it is unfolded).

proof/refine/ARM/StateRelation.thy
```diff
@@ -880,33 +891,66 @@ lemma sc_replies_relation_prev_list:
...
-    \<Longrightarrow> heap_list (replyNexts_of s') (scReply sc') (rev (sc_replies sc))"
-  apply (frule (2) sc_replies_relation_prev_list)
-  apply (case_tac "sc_replies sc"; simp)
-  apply (clarsimp simp: sc_replies_relation_def sc_replies_of_scs_def scs_of_kh_def map_project_def)
-  apply (drule_tac x=x and y="sc_replies sc" in spec2)
-  apply (clarsimp simp: sc_of_def opt_map_def projectKO_opt_sc)
-  oops
+lemma sym_replies_prev_then_next_id:
+  "\<lbrakk>sym_refs (list_refs_of_replies' s');
+    ksPSpace s' p = Some (KOReply reply); replyPrev reply = Some p'\<rbrakk>
```
lsf37, line 922: Same as above, I'd write `replyPrevs_of s' p = Some p'`

proof/refine/ARM/StateRelation.thy
lsf37: Ah, you might still have to unfold it in this proof here for `sym_refs` (I'd suggest after `simp` with the right lemmas into `replyNexts_of s' |> replyPrevs_of s'`). Would still write it in contracted form in the lemma statement, though.

lsf37 (approved):
Cool. Dont' be discouraged by all the comments, this is good stuff.

michaelmcinerney (approved):

proof/refine/ARM/StateRelation.thy
mktnk3: yes, that’s right. This is all about `replyNext` and `Next`. I’ll generalise the second function.

proof/refine/ARM/StateRelation.thy
mktnk3: I actually wrote both versions. You can find the projection version below. We can probably make use of both versions?

proof/refine/ARM/StateRelation.thy
mktnk3: See the projection version below.

proof/refine/ARM/StateRelation.thy
lsf37: Right, of course, should have realised that, sorry. I'd normalise to and keep only the contracted version, otherwise we'll start duplicating everything that has projections in it and will never we quite sure which version to use.

mktnk3:
I understand what you are saying about simp set now. I guess the lemmas I have here are small and matching, so it’s fine with them, but not generally. Now I think about it, when I tried putting it in the simp set, what I thought as the need to fix proof might have been just Isabelle getting stuck...

> Maybe `opt_map_left_Some`? Which (correctly) suggests we should have a `opt_map_left_None`, which could be proved with `in_opt_map_None_eq`, but has different `simp` behaviour, and I think would make sense to keep also.
> 
> A corresponding `opt_map_right_[Some/None]` could also be written, but makes less sense, so I don't think we need those.

`opt_map_left_Some` sounds good. I’ll rename this one, and I’ll also add the None version, too.

mktnk3:
I now see that the `opt_map_left_Some` is not quite usable in an automated way. It’s a bit of work to match things up in the way you want, and you often need to instantiate `f` or `g` as necessary, it seems.

One such example was the lemma for `replyNext` and `Next`. When it is generalised as Matt suggests, it doesn’t match as it did previously, unless it is instantiated (with `replyPrevs_of` in the case I had). I think it’s still better to generalise it.
 
I also added `opt_map_assoc` and `case_opt_map_distrib` that let you push the outside `|>` in. They are basically the same things.

proof/refine/ARM/StateRelation.thy
```diff
@@ -880,33 +891,57 @@ lemma sc_replies_relation_prev_list:
...
-   apply (rule eq_reflection)
-   apply (clarsimp simp: ghost_relation_def typ_at_eq_kheap_obj data_at_def)
-   apply (intro conjI impI iffI allI; force)
-   done
+lemma list_refs_of_replies'_reftype[simp]:
+  "(p, reftype) \<in> list_refs_of_replies' s' p' \<Longrightarrow> reftype \<in> {ReplyPrev, ReplyNext}"
+  by (clarsimp simp: list_refs_of_replies'_def list_refs_of_reply'_def get_refs_def2
+              split: option.split_asm)
+
+(* FIXME RT: move *)
```
mktnk3, line 899: Move to Lib.thy?

lsf37: They should go into `lib/Monad_WP/OptionMonad.thy`

proof/refine/ARM/StateRelation.thy
```diff
@@ -880,33 +891,57 @@ lemma sc_replies_relation_prev_list:
...
+  "f x = None \<Longrightarrow> (f |> g) x = None"
+  by (clarsimp simp: opt_map_def)
+
+lemma case_opt_map_distrib:
+  "((\<lambda>s. case_option None g (f s)) |> h)
+   = ((\<lambda>s. case_option None (g |> h) (f s)))"
+  by (fastforce simp: opt_map_def split: option.splits)
+
+lemma opt_map_assoc:
+  "f |> g |> h = f |> (g |> h)"
```
lsf37, line 914: Nice! We should put this the other way around, to normalise to the non-parenthesis form.

PR #51, opened 2020-08-22
Remove sym_refs

michaelmcinerney:
This PR removes sym_refs (state_refs_of' s) from invs'. See the commit messages for a quick run-down.

I was not sure whether I have the best approach to the definition of `stateAssertE`. I'm happy to reformulate it. And happy to move things around, too

### DIFF
--- a/proof/refine/ARM/IpcCancel_R.thy
+++ b/proof/refine/ARM/IpcCancel_R.thy
@@ -286,8 +286,7 @@ lemma blocked_cancel_ipc_corres:
      st = Structures_A.BlockedOnSend epPtr p \<longrightarrow> reply_opt = None \<rbrakk> \<Longrightarrow>
    corres dc (valid_objs and pspace_aligned and pspace_distinct
               and st_tcb_at ((=) st) t and (\<lambda>s. sym_refs (state_refs_of s)))
-             (valid_objs' and valid_release_queue_iff
-              and st_tcb_at' ((=) st') t and (\<lambda>s. sym_refs (state_refs_of' s)))
+             (valid_objs' and valid_release_queue_iff and st_tcb_at' ((=) st') t)
            (blocked_cancel_ipc st t reply_opt)
            (do ep \<leftarrow> getEndpoint epPtr;
                y \<leftarrow> assert (\<not> (case ep of IdleEP \<Rightarrow> True | _ \<Rightarrow> False));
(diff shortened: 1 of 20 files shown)


proof/refine/ARM/IpcCancel_R.thy
```diff
@@ -311,8 +312,10 @@ lemma blocked_cancel_ipc_corres:
                          and (\<lambda>_. (st = Structures_A.BlockedOnSend epPtr p
                                       \<longrightarrow> (\<exists>list. rv = Structures_A.SendEP list))
                                 \<and> (st = Structures_A.thread_state.BlockedOnReceive epPtr reply_opt p'
-                                     \<longrightarrow> (\<exists>list. rv = Structures_A.RecvEP list)))" and
-                      P'="?conc_guard and valid_ep' ep" in corres_inst)
+                                     \<longrightarrow> (\<exists>list. rv = Structures_A.RecvEP list)))"
+                  and P'="valid_objs' and valid_release_queue_iff and st_tcb_at' ((=) st') t
```
mktnk3, line 316: ?conc_guard was meant to match with all the concrete guards except sym_refs, and the intention was to make it easier to drop sym_refs here. So, there isn’t a need to make this change, if you had removed `and _` after ?cong_guard in the pattern above. But, of course, there is no need to revert what you did, either.

proof/refine/ARM/IpcCancel_R.thy
```diff
@@ -445,6 +451,8 @@ lemma cancel_signal_corres:
           (invs' and st_tcb_at' ((=) (BlockedOnNotification ntfn)) t)
           (cancel_signal t ntfn)
           (cancelSignal t ntfn)"
+  apply (rule_tac Q="(\<lambda>s. sym_refs (state_refs_of' s))" in corres_cross_add_guard)
```
mktnk3, line 454: It might be worth considering to alias this line or to make it into a method because we will be doing a lot in many places.

mktnk3 (approved):
This is really nice. A great idea implemented beautifully :)

proof/refine/ARM/IpcCancel_R.thy
michaelmcinerney: Personally, I'd prefer to leave it as is, so that there's not another layer of definitions on top. But happy to do this if others would like to see it

lsf37: It'd be this line and the next. Might be indeed be worth a short method, because it's going to introduce a lot of duplication, and if we happen to tweak anything relevant it'll need a lot of updates.

michaelmcinerney: A method for both lines wouldn't apply in all cases that it's used; I add `sym_refs` in `blocked_cancel_ipc_corres`, and then only need the cross rule (as opposed to the other bits unpacking `invs'`)

michaelmcinerney: I've come up with a little method that is able to do both lines. Where should I put it? It needs to see `corres_cross_add_guard`, which is in Move, and also `state_refs_of_cross_eq`, which is in KHeap. It seems like KHeap doesn't import Move. Is that a mistake? If not, the earliest file I can see which sees both is TcbAcc. Should I put it there? If so, where in the file? Or is there a better place? I need to be able to use the method in Detype_R and IpcCancel_R, which seem to be in fairly different branches of the "tree"

mktnk3: I think it would make sense if Move (or ArchMove) is imported by KHeap_R. I am not sure if there is a particular reason for that. Could you please try doing that? And, if it works, you can put your method just after `state_refs_of_cross_eq` I guess.

michaelmcinerney:
I've added the method. It's in KHeap_R if you would like to look.

I squashed the commits too, so that I can test everything here again

PR #52, opened 2020-08-24
The state relation for abstract-to-haskell refinement is non-empty [VER-1342]

MitchellBuckley:
The title says it all.

This is a draft because I think that this needs to be moved or cleaned up before merging, and it could go into `master` (current PR is against `rt`).


mbrcknl (commented):
I think this is a useful thing to have. Some general comments:

There are already some defined states used for similar purposes elsewhere in the proofs. In the abstract spec, there is the state `init_A_st`, and a proof `invs_A` that it satisfies the invariants. In the intermediate spec, there is `initKernel`, which might be a somewhat more realistic initial state, but might not be up-to-date, and might not be well defined on all architectures. I suspect that `init_A_st` would not be related to the state resulting from `initKernel`. I think it would be nice to update this proof to use one of those existing states. If you choose `init_A_st`, I don't think that would be too difficult.

I suggest moving this to a new theory, `Init_R`, and put it near the end, probably after `ADT_H`.

lsf37 (approved):
Cool, this is very good to have. Re placement I agree with Matt, it should into its own theory. It fits with ADT_H, so could be imported there, but doesn't need anything beyond KHeap, so should only import that, I think.

mbrcknl:
One other minor comment: formatting/indentation is inconsistent. For definitions with short types, use:
```
definition name :: type where
  "name == body"
```
If the type is long (probably not in this PR), use:
```
definition name ::
  type
  where
  "name == body"
```

lsf37:
Let me see if Jira integration is working yet. This PR addresses VER-1342 or is the syntax [VER-1342]?

lsf37:
Cool, seems to work. You have to type `[VER-1342]`

lsf37:
Hey @MitchellBuckley, this project seems to have gone stale. Are you planning to return to it?

It'd be nice to have it in the release that is happening (hopefully) this week, but it would have to be on the `master` branch for that to make sense. Might be too much of a change, but if not, let me know.

MitchellBuckley:
> Hey @MitchellBuckley, this project seems to have gone stale. Are you planning to return to it?
> 
> It'd be nice to have it in the release that is happening (hopefully) this week, but it would have to be on the `master` branch for that to make sense. Might be too much of a change, but if not, let me know.

Hey @lsf37 I probably can't have this done this week, but I don't expect it to be hard to move it to master. I'm happy for you to take over if you're interested.

MitchellBuckley:
> Hey @MitchellBuckley, this project seems to have gone stale. Are you planning to return to it?
> 
> It'd be nice to have it in the release that is happening (hopefully) this week, but it would have to be on the `master` branch for that to make sense. Might be too much of a change, but if not, let me know.

Indeed it is stale. I don't plan on returning to it soon. I'm very happy for someone to complete it without me. If no-one does complete it, I may resurrect it when I have a bit more free time.

lsf37:
> Hey @lsf37 I probably can't have this done this week, but I don't expect it to be hard to move it to master. I'm happy for you to take over if you're interested.

Ok, no worries. I don't know yet how the week will pan out meeting-wise, but if there is a bit of air I might give it a shot.

MitchellBuckley:
@lsf37 I've started looking at this again. If we defer reasoning about initial states, then this is easy. I alrealy have a branch that ports this to master and uses zeroed states to show the state relation is not empty. Is that enough for now? (Defer reasonsing about initial states to another time?)

lsf37:
> @lsf37 I've started looking at this again. If we defer reasoning about initial states, then this is easy. I alrealy have a branch that ports this to master and uses zeroed states to show the state relation is not empty. Is that enough for now? (Defer reasonsing about initial states to another time?)

Zeroed state is better than what we have now, so yes, from my side it would be Ok to merge that.

Maybe put in a comment/FIXME (and Jira issue :-)) that explains that we should in the future use the same initial state as we do for the invariants. The main gain from that, apart from avoiding defining two arbitrary states, is that we then have a state that satisfies both invariants and state relation, which is something the main refinement theorem assumes. So that would complete the story.

PR #54, opened 2020-08-25
c-parser: avoid cleaning `umm_heap/ARM_HYP`

mbrcknl:
Initially the `Makefile` copied `umm_heap/ARM_HYP` from `umm_heap/ARM`,
and deleted `umm_heap/ARM_HYP` during `make clean`. However, the
contents of `umm_heap/ARM_HYP` have since been committed, so this is no
longer appropriate.

Reported-by: Michael Norrish <Michael.Norrish@data61.csiro.au>
Signed-off-by: Matthew Brecknell <Matthew.Brecknell@data61.csiro.au>

### DIFF
--- a/tools/c-parser/standalone-parser/Makefile
+++ b/tools/c-parser/standalone-parser/Makefile
@@ -42,14 +42,6 @@ standalone-tokenizer stp_all: $(TOKENIZERS)
 
 include $(STP_PFX)/../Makefile
 
-# This is only used for ARM_HYP and ARM since the former is a copy of the latter
-UMM_HEAP_ARM_HYP=$(realpath $(STP_PFX)/../umm_heap)/ARM_HYP
-UMM_HEAP_ARM=$(realpath $(STP_PFX)/../umm_heap/ARM)
-
-$(UMM_HEAP_ARM_HYP)/% : $(UMM_HEAP_ARM)/%
-	mkdir -p $(UMM_HEAP_ARM_HYP)
-	cp -p $< $@
-
 STP_CLEAN_TARGETS := $(STPARSERS) $(TOKENIZERS) $(STP_PFX)/c-parser.o $(STP_PFX)/table.ML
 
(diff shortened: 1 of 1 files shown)


mn200 (commented):
Looks right to me.

lsf37 (commented):
Also looks fine from here in principle.

Only concern is that we committed this file to the repo in the first place. Is it regenerated when out of date, e.g. when the C changes to include a new toplevel type? It looks like this rule might have taken that action when ARM is regenerated, under the assumption that they will always be out of date at the same time, but I'm not sure.

mbrcknl:
> Only concern is that we committed this file to the repo in the first place. Is it regenerated when out of date, e.g. when the C changes to include a new toplevel type? It looks like this rule might have taken that action when ARM is regenerated, under the assumption that they will always be out of date at the same time, but I'm not sure.

The `ARM` files were never generated. For `ARM`, `X64` and `RISCV64`, they were always checked in from the start. For a little while, it looks like these rules treated `ARM_HYP` specially, by "generating" its files as a simple copy of the `ARM` files. But then the `ARM` files got checked in. Given that they were checked in without deleting these `Makefile` rules, I suspect they were checked in by someone working on `ARM_HYP` who wasn't aware that it had special treatment in the `Makefile`.

I'm not really bothered which way we do it. The files aren't very big, so the cognitive overhead of maintaining some checked-in copies is probably comparable to the overhead of giving `ARM_HYP` special treatment.

mbrcknl:
I should add that these `Makefile` rules only survived for 3 weeks before someone checked the files in. :-)

Xaphiosis:
I agree. Overhead of handling ARM_HYP separately to ARM is minimal compared to the special-cased Makefile rules that clearly did not survive the test of time.

lsf37 (approved):
Ok, I'm happy with that. Let's go for it.

mbrcknl:
Oh, and the copying rules never fire, because:
- We never update the `ARM` files.
- `git` checks out in filesystem sort order, so `ARM_HYP` comes out after `ARM`. :-)

PR #55, opened 2020-08-25
Haskell fix

lsf37:
This fixes the Haskell compile breakage introduced in #51.

It turns out that to do that, we need to teach the Haskell translator how to deal with (and ignore) multi-variable type classes.


lsf37:
Testboard [running here](https://bamboo.ts.data61.csiro.au/browse/L4V-TESTBOARD-592)

mktnk3 (approved):
Thanks for taking care of this, Gerwin. Good to hear that haskell translator can learn this :)
If only we could teach it everything...

mbrcknl (approved):
Nice!

PR #56, opened 2020-08-25
complete github CI check for MCS 

lsf37:
This should now run all tests the internal testboard currently runs on github.

Unfortunately, once we get to CRefine github CI will runners be too small to do that, but until then we can enjoy completeness.

### DIFF
--- a/spec/haskell/Makefile
+++ b/spec/haskell/Makefile
@@ -74,6 +74,7 @@ build-riscv: sandbox $(BOOT_FILES)
 # cabal package database is missing, the build will fail.
 
 .stack-work:
+	mkdir ~/.stack
 	stack --install-ghc build cabal-install
 
 $(CUSTOM_BOOT_FILES):
--- a/.github/workflows/proof.yml
+++ b/.github/workflows/proof.yml
@@ -12,6 +12,20 @@ on:
   pull_request:
 
 jobs:
+  haskell:
+    name: Haskell
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        arch: [ARM]
+    steps:
+    - name: compile Haskell
+      uses: seL4/ci-actions/run-proofs@master
+      with:
+        L4V_ARCH: ${{ matrix.arch }}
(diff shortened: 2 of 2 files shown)


lsf37:
I think this should work now. We'll see tomorrow morning.

spec/haskell/Makefile
```diff
@@ -74,6 +74,7 @@ build-riscv: sandbox $(BOOT_FILES)
 # cabal package database is missing, the build will fail.
 
 .stack-work:
+	mkdir ~/.stack
```
mktnk3, line 77: Just curious, is this needed for the changes in this PR specifically or for something else?

mktnk3 (approved):
Thanks for doing this, Gerwin. I don’t understand the details of the changes in yaml but it looks like it is working fine :)

spec/haskell/Makefile
lsf37: This is needed because on github, the user home (`~`) is not owned by the CI runner, and `stack` will refuse to create the directory `~/.stack` when that is the case.

mbrcknl (approved):
Looks good, even though these GitHub workflow configs still seem somewhat magical to me. I guess I will eventually need to read some docs to understand how the workflow config influences the chain of steps leading up to the `run_tests` in `seL4/ci-actions/blob/master/run-proofs/steps.sh`. :-)

lsf37:
> Looks good, even though these GitHub workflow configs still seem somewhat magical to me. I guess I will eventually need to read some docs to understand how the workflow config influences the chain of steps leading up to the `run_tests` in `seL4/ci-actions/blob/master/run-proofs/steps.sh`. :-)

Yes, it did require quite a bit of reading + trial & error to get it going. And, of course they added the feature that would have made everything much easier just after we were done (composite workflow steps, so that you can reference a common sequence in other workflows). If there is time I'm still planning to simplify and use these instead of the docker/js/shell mix we have now.

Also, main docs are [here](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions) if you are curious. There's always more, of course.

PR #57, opened 2020-08-26
haskell: deal with existing "~/.stack"

lsf37:
Bring happiness to both bamboo and github: create ~/.stack, but don't fail if it's there.

[Testboard running](https://bamboo.ts.data61.csiro.au/browse/L4V-TESTBOARD-593)

### DIFF
--- a/spec/haskell/Makefile
+++ b/spec/haskell/Makefile
@@ -74,7 +74,7 @@ build-riscv: sandbox $(BOOT_FILES)
 # cabal package database is missing, the build will fail.
 
 .stack-work:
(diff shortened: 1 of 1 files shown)


mbrcknl (approved):
I can almost feel your pain from the other coast of Australia.

lsf37:
> I can almost feel your pain from the other coast of Australia.

:-) Where it is 35C and my phone is overheating from wifi hotspot duties..

mktnk3 (approved):
ah, I see. It turned out to be a bit pickier.

PR #58, opened 2020-08-27
some lemmas for handling sched_context size + improved `update_sc_no_reply_stack_update_corres`

mktnk3:
- the main purpose of this commit is to have a more usable
`update_sc_no_reply_stack_update_corres` lemma.

In order to match up `update_sched_context` with `setSchedContext` which is actully `setObject`, we need to consider `set_object` updating the abstract heap with a new scheduling context with some particular `n`. The new `update_sc_no_reply_stack_update_corres` gets this from the concrete side object `sc’`.

- this also led to cleaning up the way we deal with the scheduling
context sizes. This commit adds a few lemmas that provide ways to
handle scheduling context sizes explicitly.
  See `get_sc_corres_size`.

- this also adds `sc_obj_at_cross` and `corres_cross_add_abs_guard`
- also includes some cleanup for `ArchMove_R` import

### DIFF
--- a/lib/Corres_UL.thy
+++ b/lib/Corres_UL.thy
@@ -1178,6 +1178,18 @@ lemma corres_move_asm:
 
 lemmas corres_cross_over_guard = corres_move_asm[rotated]
 
+lemma corres_cross_add_guard:
+  "\<lbrakk> \<And>s s'. \<lbrakk>(s,s') \<in> sr; P s; P' s'\<rbrakk> \<Longrightarrow> Q s';
+     corres_underlying sr nf nf' r P  (P' and Q) f g\<rbrakk>
+    \<Longrightarrow> corres_underlying sr nf nf' r P P' f g"
+  by (fastforce simp: corres_underlying_def)
+
+lemma corres_cross_add_abs_guard:
+  "\<lbrakk> \<And>s s'. \<lbrakk>(s,s') \<in> sr; P s; P' s'\<rbrakk> \<Longrightarrow> Q s;
+     corres_underlying sr nf nf' r (P and Q)  P' f g\<rbrakk>
+    \<Longrightarrow> corres_underlying sr nf nf' r P P' f g"
+  by (fastforce simp: corres_underlying_def)
+
 lemma corres_cross_back:
   "\<lbrakk> \<And>s s'. \<lbrakk> (s,s') \<in> sr; P' s' \<rbrakk> \<Longrightarrow> P s; \<And>s. Q' s \<Longrightarrow> P' s;
      corres_underlying sr nf nf' r (Q and P) Q' f g \<rbrakk>
--- a/proof/refine/ARM/KHeap_R.thy
+++ b/proof/refine/ARM/KHeap_R.thy
@@ -8,7 +8,6 @@ theory KHeap_R
 imports
   "AInvs.ArchDetSchedSchedule_AI"
   Machine_R
-  "../Move_R"
 begin
 
 lemma lookupAround2_known1:
@@ -3390,13 +3389,6 @@ lemma ko_at'_cross:
                      scBits_simps sc_relation_def objBits_simps
               split: Structures_A.kernel_object.split_asm if_split_asm
                      ARM_A.arch_kernel_obj.split_asm)
-
-(* FIXME RT: move *)
-lemma corres_cross_add_abs_guard:
-  "\<lbrakk> \<And>s s'. \<lbrakk>(s,s') \<in> sr; P s; P' s'\<rbrakk> \<Longrightarrow> Q s;
-     corres_underlying sr nf nf' r (P and Q)  P' f g\<rbrakk>
-    \<Longrightarrow> corres_underlying sr nf nf' r P P' f g"
-  by (fastforce simp: corres_underlying_def)
 
 lemma update_sc_no_reply_stack_update_corres:
   "\<lbrakk>\<forall>sc n sc'. sc_relation sc n sc' \<longrightarrow> sc_relation (f sc) n (f' sc');
--- a/proof/refine/ARM/KHeap_R.thy
+++ b/proof/refine/ARM/KHeap_R.thy
@@ -1303,30 +1303,34 @@ lemma set_reply_corres: (* for reply update that doesn't touch the reply stack *
         simp add: typ_at'_def ko_wp_at'_def obj_at'_def project_inject opt_map_def sc_of_def)
   qed
 
-lemma update_sc_no_reply_stack_update_corres:
-  "\<forall>ae. sc_replies (f ae) = sc_replies ae \<Longrightarrow>
-   corres dc
-     (obj_at (\<lambda>ko. \<exists>ae n. ko = kernel_object.SchedContext ae n
-                               \<and> sc_relation (f ae) n ae') ptr)
-     (obj_at' (\<lambda>sc. scReply ae' = scReply sc \<and> objBits ae' = objBits sc) ptr)
-          (update_sched_context ptr f) (setSchedContext ptr ae')"
+lemma setSchedContext_no_stack_update_corres:
+  "\<lbrakk> \<forall>sc n sc'. sc_relation sc n sc' \<longrightarrow> sc_relation (f sc) n (f' sc');
+     \<forall>sc. sc_replies sc = sc_replies (f sc); \<forall>sc'. objBits sc' = objBits (f' sc');
+     scReply sc' = scReply (f' sc')\<rbrakk>
+    \<Longrightarrow> corres dc
+         (\<lambda>s. kheap s ptr = Some (kernel_object.SchedContext sc n))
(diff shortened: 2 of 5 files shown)


proof/refine/ARM/KHeap_R.thy
```diff
@@ -3314,5 +3349,49 @@ lemma refillSufficient_wp:
...
+   apply (rule get_object_sp)
+  apply (clarsimp simp: monadic_rewrite_def obj_at_def is_sc_obj_def)
+  done
+
+lemma update_sc_no_reply_stack_update_corres:
+  "\<lbrakk>\<forall>sc sc'. sc_relation sc n sc' \<and> n + minSchedContextBits = objBits sc'
+                    \<longrightarrow> sc_relation (f sc) n (f' sc');
+    \<forall>sc. sc_replies sc = sc_replies (f sc);
+    \<forall>sc'. scReply (f' sc') = scReply sc' \<rbrakk> \<Longrightarrow>
+   corres dc (sc_obj_at n ptr and pspace_aligned and pspace_distinct) (sc_at' ptr)
```
mktnk3, line 3373: I thought I removed the concrete side guard and used the cross lemma. Maybe I mixed things up a bit while I did some cleanup.

proof/refine/ARM/KHeap_R.thy
```diff
@@ -3314,5 +3349,49 @@ lemma refillSufficient_wp:
...
+          od)"
+  apply (rule corres_guard_imp)
+    apply (rule_tac P="sc_obj_at n ptr" and P'="sc_at' ptr"
+               in monadic_rewrite_corres[OF _ update_sched_context_rewrite[where n=n]])
+    apply (rule corres_guard_imp)
+      apply (rule corres_split[OF _ get_sc_corres_size[where n=n]])
+        apply (rename_tac sc sc')
+        apply (rule_tac P="\<lambda>s. kheap s ptr = Some (kernel_object.SchedContext sc n)"
+                    and P'="ko_at' sc' ptr" in corres_inst)
+        apply (rule setSchedContext_corres[where n=n])
```
mktnk3, line 3386: This lemma here is renamed to `setSchedContext_no_stack_update_corres`

mktnk3:
This is to fit with what Michael is currently doing, and having discussed with him, it seems it could do with further tweak. So, please wait for reviewing. And now I realise I probably should have commented this before removing the review requests...

proof/refine/ARM/KHeap_R.thy
```diff
@@ -1303,47 +1302,50 @@ lemma set_reply_corres: (* for reply update that doesn't touch the reply stack *
...
                    split: Structures_A.kernel_object.splits)
-    apply (rename_tac sc)
-    apply (prop_tac "obj_at (same_caps (kernel_object.SchedContext ae n)) ptr a")
+    apply (prop_tac "obj_at (same_caps (kernel_object.SchedContext sc n)) ptr a")
      apply (clarsimp simp: obj_at_def)
     apply (clarsimp simp: state_relation_def
                           z[simplified obj_at'_def is_sc_obj_def projectKO_eq projectKO_opts_defs, simplified])
-    apply (clarsimp simp add: caps_of_state_after_update cte_wp_at_after_update
+    apply (clarsimp simp: caps_of_state_after_update cte_wp_at_after_update
                               swp_def fun_upd_def obj_at_def)
```
michaelmcinerney, line 1345: Nitpick: indentation

proof/refine/ARM/KHeap_R.thy
```diff
@@ -1473,6 +1475,25 @@ lemma get_sc_corres:
   apply (fastforce simp add: other_obj_relation_def)
 done
 
+lemma get_sc_corres_size:
+  "corres (\<lambda>sc sc'. sc_relation sc n sc' \<and> n + minSchedContextBits = objBits sc')
+     (sc_obj_at n ptr) (sc_at' ptr)
+     (get_sched_context ptr) (getSchedContext ptr)"
+  apply (rule corres_no_failI)
+   apply wp
+  apply (simp add: get_sched_context_def getSchedContext_def get_object_def
```
michaelmcinerney, line 1484: Can some of these simps and clarsimps be combined? Or would it make more sense to leave them as is?

michaelmcinerney (approved):
Looks great! I did try this out in the corres rule for `sched_context_yield_to` and it worked perfectly. Thanks!

proof/refine/ARM/KHeap_R.thy
mktnk3: Putting all together actually breaks the proof. It is possible that some of these are ok to combine, but some attempts I made didn’t work and I don’t think it is worth the effort to explore other possibilities right now. So, I’ll leave it like this for now.

proof/refine/ARM/KHeap_R.thy
```diff
@@ -3314,5 +3348,86 @@ lemma refillSufficient_wp:
...
+  shows "sc_obj_at (objBits sc' - minSchedContextBits) ptr s" using assms
+  apply (clarsimp simp: obj_at'_def projectKOs)
+  apply (erule (1) pspace_dom_relatedE)
+  by (clarsimp simp: obj_relation_cuts_def2 obj_at_def is_sc_obj cte_relation_def
+                     other_obj_relation_def pte_relation_def pde_relation_def
+                     scBits_simps sc_relation_def objBits_simps
+              split: Structures_A.kernel_object.split_asm if_split_asm
+                     ARM_A.arch_kernel_obj.split_asm)
+
+lemma update_sc_no_reply_stack_update_corres:
```
mktnk3, line 3393: This proof is quite a journey in that it goes:
`sc_at` -> `sc_at’` -> symb_exec of `getSchedContext` to fix `sc’` -> `ko_at’` -> `sc_at_n`

It makes sense that this works. But I wonder if there is a more direct way to do something like this.

lib/Corres_UL.thy
```diff
@@ -1178,6 +1178,18 @@ lemma corres_move_asm:
 
 lemmas corres_cross_over_guard = corres_move_asm[rotated]
 
+lemma corres_cross_add_guard:
```
MitchellBuckley, line 1181: Note: these two lemmas can be expressed using cross_rel. But I don't know if that's necessary. Just wanted to make the observation in case it's useful.

proof/refine/ARM/KHeap_R.thy
```diff
@@ -3314,5 +3348,86 @@ lemma refillSufficient_wp:
   apply (wpsimp wp: getRefills_wp)
   by (clarsimp simp: sufficientRefills_def obj_at'_def)
 
+lemma update_sched_context_rewrite:
```
MitchellBuckley, line 3351: This is excellent!!

MitchellBuckley (approved):

lsf37 (commented):
👍

lsf37 (approved):

lib/Corres_UL.thy
mktnk3: It is actually slightly different, I think. `cross_rel` only uses `P` and the state_relation to derive `Q`, but here `Q` also depends on `P’`. This dependency exists in and is copied from the `corres_cross_over_guard` lemma defined just above here. I am not sure, but maybe `cross_rel` is slightly restrictive than what we always want. Or, this `P’` is not often used and hence we could do without most of the time. It might be worth while to investigate, and depending how it goes we could consider converting existing some of the cross lemmas into `cross_rel` statements.

MitchellBuckley: I should have been clearer, and I should check in an isabelle session, but I think the cross_rel expression would be `cross_rel P (%s'. P' s' -> Q' s') `

MitchellBuckley: But this trick won't work on the second lemma. So it's true I haven't completely thought this through.

mktnk3: Ah, you are right. That second one is in the opposite direction. And,

> I should have been clearer, and I should check in an isabelle session, but I think the cross_rel expression would be `cross_rel P (%s'. P' s' -> Q' s')`

yes, that does work! I’d find the expression slightly less intuitive in that Q’ actually depends on both P and P’ (probably more on P), but it’s rare that we need to use P’ anyway. This PR is merged now, but it would be really good to start using `cross_rel` more, I agree. There is [JIRA-1290](https://sel4.atlassian.net/browse/VER-1290) for this which we should get on to when we can.

lsf37 (commented):
It looks like this comment didn't get posted because bad internet connection. No problem, though.

lib/Corres_UL.thy
mktnk3: And, just to mention, this `corres_cross_add_guard` is something I added a while ago, and is used in several places including Michael’s `add_sym_refs` method. What is new is the one in the other direction (the one next to this). I moved both to lib on this occasion.

mbrcknl:
I realise I'm very late to the party, but this looks good!

PR #62, opened 2020-08-28
prove cancelAllSignals_invs', fix suspend_invs'

michaelmcinerney:
This proves `cancelAllSignals_invs'`.

I also fix up `suspend_invs'` to use the lemma Ed wrote. This removes another sorry due to duplicating that lemma statement.

I decided to lump this all into one commit, but happy to split it up if anyone would like

### DIFF
--- a/proof/refine/ARM/CNodeInv_R.thy
+++ b/proof/refine/ARM/CNodeInv_R.thy
@@ -607,7 +607,7 @@ lemma suspend_not_recursive_ctes:
   unfolding updateRestartPC_def
   apply (wp threadSet_ctes_of | simp add: unless_def del: o_apply)+
          apply (fold cteCaps_of_def)
-         apply (wp cancelIPC_cteCaps_of gts_wp' hoare_vcg_all_lift hoare_drop_imps)+
+         apply (wp cancelIPC_cteCaps_of gts_wp' stateAssert_wp hoare_vcg_all_lift hoare_drop_imps)+
   apply (clarsimp elim!: rsubst[where P=P] intro!: set_eqI)
   apply (clarsimp simp: cte_wp_at_ctes_of cteCaps_of_def)
   apply (auto simp: isCap_simps finaliseCap_def Let_def)
--- a/proof/refine/ARM/IpcCancel_R.thy
+++ b/proof/refine/ARM/IpcCancel_R.thy
@@ -111,7 +111,6 @@ lemma cancelSignal_st_tcb_at':
                     hoare_drop_imp[where R="\<lambda>rv _. \<exists>a b. delete t (f rv) = a # b" for f])
   done
 
-(* FIXME RT: Michael to fix and use in his next PR
 lemma cancelIPC_simple'_not_awaiting_reply:
   "\<lbrace>\<lambda>s. sym_refs (state_refs_of' s)\<rbrace>
    cancelIPC t
@@ -129,7 +128,7 @@ lemma cancelIPC_simple'_not_awaiting_reply:
    apply (frule(1) sym_ref_Receive_or_Reply_replyTCB', fast)
    apply (clarsimp simp: obj_at'_def projectKO_eq project_inject)
   apply (clarsimp simp: pred_tcb_at'_def obj_at'_def)
-  done*)
+  done
 
 context begin interpretation Arch .
 crunch typ_at'[wp]: emptySlot "\<lambda>s. P (typ_at' T p s)"
@@ -1389,9 +1388,10 @@ lemma (in delete_one) suspend_corres:
   "corres dc (einvs and tcb_at t) (invs' and tcb_at' t)
         (SchedContext_A.suspend t) (ThreadDecls_H.suspend t)"
   apply (simp add: SchedContext_A.suspend_def Thread_H.suspend_def)
+  apply add_sym_refs
   apply (rule corres_guard_imp)
-    apply (rule corres_split_nor [OF _ cancel_ipc_corres])
   sorry (*
+    apply (rule corres_split_nor [OF _ cancel_ipc_corres])
       apply (rule corres_split [OF _ gts_corres])
         apply (rule corres_split_nor)
            apply (rule corres_split_nor [OF _ sts_corres])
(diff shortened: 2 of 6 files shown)


proof/refine/ARM/CNodeInv_R.thy
```diff
@@ -604,6 +604,7 @@ lemma suspend_not_recursive_ctes:
      suspend t
    \<lbrace>\<lambda>rv s. P (not_recursive_ctes s)\<rbrace>"
   apply (simp only: suspend_def not_recursive_ctes_def cteCaps_of_def)
+  apply (rule hoare_seq_ext[OF _ stateAssert_sp])
```
lsf37, line 607: with a `stateAssert_wp` rule this could likely be folded into the `wp` below

proof/refine/ARM/IpcCancel_R.thy
```diff
@@ -2857,39 +2856,59 @@ lemma cancelAllIPC_invs'[wp]:
...
-  apply (fastforce simp: symreftype_inverse' ntfn_bound_refs'_def tcb_bound_refs'_def)
-  done *)
+  apply (rule hoare_seq_ext[OF _ stateAssert_sp])
+  apply (rule hoare_seq_ext[OF _ get_ntfn_sp'])
+  apply (case_tac "ntfnObj ntfn"; simp)
+    apply wpsimp
+   apply wpsimp
+  apply (wpsimp wp: rescheduleRequired_all_invs_but_ct_not_inQ sts_st_tcb_at'_cases
+                    cancelAllSignals_invs'_helper hoare_vcg_const_Ball_lift
+                    sts_valid_queues hoare_drop_imps hoare_vcg_all_lift)
```
lsf37, line 2901: Could this whole block be automated to just the last `wpsimp`? It should do the case distinction automatically. It might need `wp` rule additions for `get_ntfn` (maybe, should be there by default) and `stateAssert`.

lsf37 (approved):
Looks good from here. It'd be nice to to introduce too many single rule `_sp` steps. They are nice now for incremental proof adjustment (and where I haven't commented, they are fine, not worth the effort to eliminate), but if there is a whole bunch of new proof around them, it's at least worth investigating if they couldn't be rolled into `wp`. The main reason for this is to be slightly more robust against code change, so that not every tiny reordering breaks all of them.

proof/refine/ARM/IpcCancel_R.thy
michaelmcinerney: Rearranging this turned out to be quite nasty. I'm not sure exactly why, but when the wpsimp has to do the case distinction, it struggles with the `all_invs_but_ct_not_inQ` term. This is an abbreviation, and it seems like it does some rewriting. So it'll use `rescheduleRequired_all_invs_but_ct_not_inQ`, do some rewriting, and then won't be able to apply `cancelAllSignals_invs'_helper` immediately. 

One way of solving this is to make `all_invs_but_ct_not_inQ` a definition. I'm not sure whether we want these things (`all_invs_but_blah`) to be abbreviations or definitions (we seem to have both). Another approach is to make `valid_dom_schedule'` a definition. This is listed as a FIXME. I have tried both! Making `all_invs_but_ct_not_inQ` touches far fewer lines. But making `valid_dom_schedule'` might be a better solution. Neither really make for that much of an improvement in this proof though

michaelmcinerney: Just for a bit of a laugh, I've pushed a branch to github (called `michaelm-cancelAllSignals_invs2`) that would be the change if we wanted to make `valid_dom_schedule'` a definition

michaelmcinerney: One way to avoid as much unfolding of the definition as there is would be to crunch more things for this. But considering we already unfold `invs'` and `valid_state'` all over the place, and considering that since it was an abbreviation before, simply unfolding the definition will get us back to the same proof state (in most cases), this is probably still a sensible approach

lsf37: > Just for a bit of a laugh, I've pushed a branch to github (called michaelm-cancelAllSignals_invs2) that would be the change if we wanted to make valid_dom_schedule' a definition

You mean [this bit here](https://github.com/seL4/l4v/blob/4594f5337cf4cb44f9e6688c7e34cb4d2f3f38ce/proof/refine/ARM/IpcCancel_R.thy#L2898-L2901)? 

You're right, it doesn't have much impact on this proof, and arguably makes it a bit worse because of the multiple `wpsimp` invocations without explicit `case`. Maybe let's leave this proof as is then. 

I'm unclear on whether `valid_dom_schedule'` as definition has wider impact that I haven't seen. If we're only ever just unfolding it, I'm not sure it's worth it, but I'm not against. If you feel it makes your life easier, go for it.

michaelmcinerney: Yes, that is the bit. I'm generally very much against multiple wpsimps in a row; a helper lemma is probably the only reason I'd go for it. If you'd prefer me to go back to the original proof (I would probably prefer the original proof too), then having `valid_dom_schedule'` as a definition is not necessary. Shall I just go back to the way things were in the PR originally?

michaelmcinerney: (I didn't get as far as the other change in CNodeInv, but I can look into that too)

mbrcknl (approved):
Looks good!

mbrcknl:
@michaelmcinerney Try to avoid force-pushing while iterating the review-update cycle. Add new commits instead, since this makes it easier for reviewers to see what you updated. When everyone is happy with the content, you can then go through a round of cleaning up the history. If you put the work "squash" in a commit message, the automation will remind you to clean up the history before merging.

michaelmcinerney:
> @michaelmcinerney Try to avoid force-pushing while iterating the review-update cycle. Add new commits instead, since this makes it easier for reviewers to see what you updated. When everyone is happy with the content, you can then go through a round of cleaning up the history. If you put the work "squash" in a commit message, the automation will remind you to clean up the history before merging.

Oh yeah I forgot about that. I was worried that if I squashed later, I'd have to run the tests AGAIN and it would delay things even more. But I suppose I can use your trick of checking the diff after the squash

mbrcknl:
> Try to avoid force-pushing while iterating the review-update cycle.

It occurs to me that following this strictly would prevent rebasing during the review-update cycle, and I don't think that's desirable. So perhaps we should refine the convention to say that force-pushing a rebase is fine whenever you like, as long as the force-push doesn't materially change the *content* of the PR if you're in the middle of the review-update cycle. To change the content, you can still add new commits before or after the rebase. It would perhaps also be nice to add a comment to signal that the force-push is just a rebase.

mktnk3 (approved):
This looks great. Thanks for doing the experiment regarding this wpsimp / abbreviation interaction!

PR #64, opened 2020-09-01
Reprogram timer updates + misc fixes

ryybrr:
I added a few missing reprogram timer updates to some Abstract/Haskell functions.
I also fixed a couple of minor issues which affect refinement proofs.

### DIFF
--- a/proof/invariant-abstract/DetSchedSchedule_AI.thy
+++ b/proof/invariant-abstract/DetSchedSchedule_AI.thy
@@ -413,10 +413,13 @@ lemma tcb_sched_enqueue_not_queued:
 \<comment> \<open>tcb_release_remove\<close>
 
 lemma tcb_release_remove_wp[valid_sched_wp]:
-  "\<lbrace>\<lambda>s. P (release_queue_update (tcb_sched_dequeue thread) s)\<rbrace>
+  "\<lbrace>\<lambda>s. if release_queue s \<noteq> [] \<and> hd (release_queue s) = thread
+        then P (s\<lparr>reprogram_timer := True,
+                  release_queue := tcb_sched_dequeue thread (release_queue s)\<rparr>)
+        else P (s\<lparr>release_queue := tcb_sched_dequeue thread (release_queue s)\<rparr>)\<rbrace>
    tcb_release_remove thread
    \<lbrace>\<lambda>_. P\<rbrace>"
-  by (wpsimp simp: tcb_release_remove_def obj_at_def)
+  by (wpsimp simp: tcb_release_remove_def obj_at_def | intro conjI)+
 
 lemma tcb_release_remove_valid_sched_misc[wp]:
   "tcb_release_remove t \<lbrace>\<lambda>s. P (consumed_time s) (cur_sc s) (cur_time s) (cur_domain s) (cur_thread s) (idle_thread s)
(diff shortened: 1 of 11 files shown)


spec/abstract/SchedContext_A.thy
```diff
@@ -366,8 +368,8 @@ definition
   sched_context_bind_tcb :: "obj_ref \<Rightarrow> obj_ref \<Rightarrow> (unit, 'z::state_ext) s_monad"
 where
   "sched_context_bind_tcb sc_ptr tcb_ptr = do
-    set_sc_obj_ref sc_tcb_update sc_ptr (Some tcb_ptr);
     set_tcb_obj_ref tcb_sched_context_update tcb_ptr (Some sc_ptr);
+    set_sc_obj_ref sc_tcb_update sc_ptr (Some tcb_ptr);
```
michaelmcinerney, line 372: Was there a particular reason this was changed? Just for the refinement proof?

michaelmcinerney (approved):
Looks good!

I know that this would bring the abstract spec in line with the Haskell very nicely. Did you have a look to see whether the Haskell matches well with the C? I took a quick look the other day, but not well enough to say

spec/abstract/SchedContext_A.thy
ryybrr: Yep, the order is the other way around in Haskell/C

proof/invariant-abstract/RealTime_AI.thy
```diff
@@ -463,6 +467,11 @@ lemma cur_tcb_release_queue_update[simp]:
   "cur_tcb (release_queue_update f s) = cur_tcb s"
   by (simp add: cur_tcb_def)
 
+(* FIXME: move to Invariants_AI *)
```
lsf37, line 470:
```suggestion
(* FIXME RT: move to Invariants_AI *)
```
(or actually move it now)

proof/invariant-abstract/RealTime_AI.thy
```diff
@@ -449,6 +449,10 @@ lemma valid_irq_states_release_queue_update[simp]:
   "valid_irq_states (release_queue_update f s) = valid_irq_states s"
   by (simp add: valid_irq_states_def)
 
+lemma valid_irq_states_reprogram_timer_update[simp]:
+  "valid_irq_states (reprogram_timer_update f s) = valid_irq_states s"
+  by (simp add: valid_irq_states_def)
+
 (* FIXME: move to Invariants_AI *)
```
lsf37, line 456:
```suggestion
(* FIXME RT: move to Invariants_AI *)
```
(I know not your fault, but if we're here already we should fix the FIXME tags so we can find things later *)

proof/invariant-abstract/RealTime_AI.thy
```diff
@@ -502,6 +511,11 @@ lemma valid_machine_state_release_queue_update[simp]:
   "valid_machine_state (release_queue_update f s) = valid_machine_state s"
   by (simp add: valid_machine_state_def)
 
+(* FIXME: move to Invariants_AI *)
```
lsf37, line 514:
```suggestion
(* FIXME RT: move to Invariants_AI *)
```

lsf37 (approved):
👍 Looks good, only minor FIXME things.

mktnk3 (approved):
Very nice and neat. Thanks for catching all these spec issues!

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -3149,15 +3152,21 @@ lemma map_fst_filter_zip_map_reduce:
   by (induct xs) auto
 
 lemma tcb_release_enqueue_wp':
-  "\<lbrace>\<lambda>s. tcb_sc_refill_cfgs_of s t \<noteq> None
-         \<longrightarrow> P (release_queue_update (tcb_release_enqueue_upd (tcb_ready_times_of s) t) s)\<rbrace>
+  "\<lbrace>\<lambda>s. tcb_sc_refill_cfgs_of s t \<noteq> None \<longrightarrow>
+          (\<forall>rt. tcb_ready_times_of s t = Some rt \<longrightarrow>
+            (if filter (\<lambda>(_, t'). t' \<le> rt) (zip (release_queue s) (map (the \<circ> tcb_ready_times_of s) (release_queue s))) = []
```
corlewis, line 3157: I think it would be nice to have a definition (or maybe an abbreviation) for this filter, that the spec and proofs could both use. Depending on how it's written it might be able to be reused between both specs as well.
It's not a blocking issue, but if other people agree it could be worth a fixme/jira.

corlewis (approved):
Looks great! I mostly skimmed the proof updates, but I put more time into checking the spec changes and they all make sense to me.
I'm guessing that `tcb_release_enqueue` and `dequeue` will be a pain in crefine, but it should be doable and I can't see any simple ways of changing the specs to make it easier.

proof/invariant-abstract/DetSchedSchedule_AI.thy
michaelmcinerney: Yeah, FIXME RT/JIRA sounds good

mbrcknl (approved):
Looks good!

PR #66, opened 2020-09-02
Update address translation in haskell spec to reflect C

xurtis:
This reflects a set of changes made to the C source some time ago: https://github.com/seL4/seL4/commit/fba7c896b025f0e95a4fc88bd2d32a4e085bd986


xurtis:
I'm not sure if this is all that needs to be done for these changes so would appreciate some guidance on the process for ensuring this change can go through.

lsf37:
It looks like nobody replied to this, sorry. The Haskell update as such is looking good. The next step before we can merge it would be to update the abstract spec and the proofs that use these names. The proof changes as such shouldn't be hard, but it looks like one full pass through everything, so it mostly takes a bunch of time.

Xaphiosis:
I'm looking at it. It's going slow but it's going.

xurtis:
I think that just shows how big a refactor pass like this is for the current proofs.

Xaphiosis:
I think it also shows how much time I've spent in meetings, context switching and being unproductive, and is more indicative of an indictment of my performance on this PR, rather than saying something about our overall approach. I will try speed this up, it's getting embarrassing.

Xaphiosis:
I have completed this update, and ran all tests on the work machine. Working out testboard situation now, but a review from other verification people would be good.

mbrcknl (approved):
I've had a very superficial look, and my initial impression is that it looks fine. I'll try to get back for a more detailed look tomorrow, but feel free to go ahead if you need to.

Xaphiosis:
Testboard is here:
https://bamboo.ts.data61.csiro.au/browse/L4V-TESTBOARD-657

lsf37 (approved):
Nice, looks at lot more consistent now! And thanks for updating the ASCII art as well.

Xaphiosis:
Testboard came back green. Will merge.

mktnk3:
> Testboard came back green. Will merge.

Sorry I didn’t get to review in time, but I suppose it ok it’s got the two approvals already. Will have a look later for my own sake.

PR #65, opened 2020-09-01
asmrefine: add timeouts for debugging

pierzchalski:
Also cleans up some of the debug config setup and makes result reporting
more useful.

Signed-off-by: Edward Pierzchalski <ed.pierzchalski@data61.csiro.au>

PR #69, opened 2020-09-03
asmrefine: add timeouts for debugging (for master)

mbrcknl:
This just cherry-picks a commit from `rt` to `master`, based on https://github.com/seL4/l4v/pull/65.

### DIFF
--- a/proof/asmrefine/SEL4GraphRefine.thy
+++ b/proof/asmrefine/SEL4GraphRefine.thy
@@ -60,16 +60,18 @@ lemma snd_snd_gs_new_frames_new_cnodes[simp]:
   "snd (snd ((if P then f else g) gs)) = (if P then snd (snd (f gs)) else snd (snd (g gs)))"
   by (simp_all add: gs_new_frames_def gs_new_cnodes_def gs_clear_region_def)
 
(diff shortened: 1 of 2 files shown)


mbrcknl:
[L4V-TESTBOARD-599](https://bamboo.ts.data61.csiro.au/browse/L4V-TESTBOARD-599)

lsf37 (approved):
Happy with the cherry-pick when the testboard goes through.

mbrcknl:
Ugh, this actually breaks `master`, because it looks like the AsmRefine session was not tested against #65 before merging into the `rt` branch. I guess the best way to handle this is to add a fixup commit on top of @pierzchalski's commit in this PR. Then for `rt`, either cherry-pick the same fixup, or merge `master` into `rt`.

lsf37:
Yes, the `rt` tests are limited to Refine only at the moment, they don't even include CRefine yet. Good that we caught this here. Fixup commit on this PR sounds good, we can then cherry-pick over or I just do another merge from `master` into `rt`, it's probably about time anyway.

mbrcknl:
Declining this to merge #70 instead.

PR #72, opened 2020-09-04
Prove `cancel_all_ipc_corres`

michaelmcinerney:


proof/refine/ARM/IpcCancel_R.thy
```diff
@@ -2521,6 +2439,445 @@ lemma possibleSwitchTo_all_invs_but_ct_not_inQ':
...
+        apply (rule corres_guard_imp)
+          apply (simp only: )
+            apply (rule P[simplified fun_app_def])
+            apply simp
+           apply simp
+          apply simp
+        apply (intro conjI impI allI; fastforce?)
+        apply (elim disjE)
+         apply (clarsimp simp: ep_relation_def)
+          apply (prop_tac "valid_ep' (Structures_H.endpoint.SendEP queue) s")
```
michaelmcinerney, line 2822: (Indentation must be off here)

proof/refine/ARM/IpcCancel_R.thy
```diff
@@ -2366,69 +2323,29 @@ lemma restart_thread_if_no_fault_corres:
...
-                                \<and> Invariants_H.valid_queues s \<and> valid_queues' s
-                                \<and> valid_release_queue_iff s \<and> valid_objs' s"
-                    in hoare_strengthen_post)
-       apply (rule mapM_x_wp')
-       apply (wpsimp wp: hoare_vcg_ball_lift sts_valid_queues sts_st_tcb_at'_cases
-                   simp: valid_tcb_state'_def)+
-      apply (fastforce simp: valid_ep_def intro!: valid_objs_valid_tcbs)
-     apply (wpsimp wp: hoare_vcg_const_Ball_lift simp: valid_ep'_def)
-     done
+(* FIXME RT Michael: also in Victor's PR *)
```
MitchellBuckley, line 2326: nitpick: remove this comment (depending on whether Victor's PR goes first)

proof/refine/ARM/IpcCancel_R.thy
```diff
@@ -2521,6 +2439,445 @@ lemma possibleSwitchTo_all_invs_but_ct_not_inQ':
...
+                        , fastforce simp: obj_at_def))
+             apply (wpsimp wp: gts_wp)
+            apply (wpsimp wp: gts_wp')
+           apply (clarsimp simp: vs_all_heap_simps obj_at_def is_tcb_def)
+          apply clarsimp
+         apply (intro hoare_vcg_conj_lift_left_fix; (solves \<open>wpsimp wp: gts_wp\<close>)?)
+           apply (wpsimp wp: restart_thread_if_no_fault_tcb_sts_of_other
+                             reply_unlink_tcb_tcb_sts_of_other gts_wp)
+         apply (rule hoare_weaken_pre
+               , rule cancel_all_ipc_loop_body_reply_unlink_ts_pred_other
```
MitchellBuckley, line 2626: nitpick: interesting comma style. I was recently told that this is not what we usually do.

proof/refine/ARM/IpcCancel_R.thy
```diff
@@ -2521,6 +2439,445 @@ lemma possibleSwitchTo_all_invs_but_ct_not_inQ':
...
+
+lemma in_send_ep_queue_TCBBlockedSend:
+  "\<lbrakk>kheap s epptr = Some (Endpoint (Structures_A.endpoint.SendEP queue)); t \<in> set queue; invs s\<rbrakk>
+   \<Longrightarrow> (epptr, TCBBlockedSend) \<in> state_refs_of s t"
+  apply (prop_tac "valid_ep (Structures_A.endpoint.SendEP queue) s")
+   apply (fastforce simp: valid_objs_def valid_obj_def dest!: invs_valid_objs)
+  apply (clarsimp simp: state_refs_of_def valid_ep_def split: option.splits)
+  apply (intro conjI impI allI; (fastforce simp: obj_at_def)?)
+  apply (prop_tac "(t, EPSend) \<in> state_refs_of s epptr")
+   apply (clarsimp simp: state_refs_of_def)
```
MitchellBuckley, line 2661: Nitpick/suggestion. In this case, I like to use a comma to put the one-line proof directly after the prop_tac. I think it reads more clearly.

proof/refine/ARM/IpcCancel_R.thy
```diff
@@ -2521,6 +2439,445 @@ lemma possibleSwitchTo_all_invs_but_ct_not_inQ':
...
+                  split: if_splits thread_state.splits)
+  done
+
+lemma IdleEP_versus_SendEP_and_RecvEP:
+  "\<lbrakk>ep = Structures_A.endpoint.IdleEP \<Longrightarrow> P;
+    \<And>q. ep = Structures_A.endpoint.SendEP q \<or> ep = Structures_A.endpoint.RecvEP q \<Longrightarrow> P\<rbrakk>
+   \<Longrightarrow> P"
+  using Structures_A.endpoint.exhaust by auto
+
+lemma not_IdleEP_case_split:
```
MitchellBuckley, line 2715: Is this needed? I can see how it might make a later proof simpler, but it might also contribute to general cluttering of our theory files. If you could do without these, I think that would be nicer.

MitchellBuckley (approved):
Impressive stuff!! Nice!

proof/refine/ARM/IpcCancel_R.thy
michaelmcinerney: Normally I place the comma one to the right. But I do prefer having the commas and the semi colons on the new line

proof/refine/ARM/IpcCancel_R.thy
michaelmcinerney: The SendEP and RecvEP cases are handled identically, as far as the corres part of the proof goes (they're a bit different later on in `cancel_all_ipc_corres`, which is why you see `elim disjE` pop up a few times). So to handle SendEP and RecvEP together, I wrote these few rules. They're annoying, because they have to be in just the right form, and applying them is a pain too (eg. `subst`), but it significantly cuts down on the length of the proof. Before, the function was simple enough to have a giant `by (blah | blah | blah)+` at the end to take care of it all without this splitting, but now it's quite nasty

michaelmcinerney: Also happy to move these rules elsewhere if others think they're useful. I'm sort of surprised we haven't needed something like that before (or maybe we do have them, and I just don't know where they are)

proof/refine/ARM/IpcCancel_R.thy
lsf37: @Xaphiosis what's the style verdict on this one? I don't think we have commas on the new line anywhere yet. I don't like it for semicolons either, but I don't like any operators on the new line, so I'm just being consistent ;-)

mbrcknl (approved):
Looks good!

There are several Hoare triples on the abstract spec. Are these worth moving up to AInvs?

proof/refine/ARM/IpcCancel_R.thy
```diff
@@ -2366,69 +2323,29 @@ lemma restart_thread_if_no_fault_corres:
...
+  "\<lbrace>\<lambda>s. Q (pred_map P (tcb_sts_of s) t') \<and> t \<noteq> t'\<rbrace>
+   restart_thread_if_no_fault t
+   \<lbrace>\<lambda>_ s. Q (pred_map P (tcb_sts_of s) t')\<rbrace>"
+  apply (clarsimp simp: restart_thread_if_no_fault_def)
+  apply (rule hoare_seq_ext_skip, wpsimp)
+  apply (wpsimp wp: possible_switch_to_pred_tcb_at[unfolded obj_at_kh_kheap_simps]
+                    set_thread_state_pred_map_tcb_sts_of thread_get_inv)
+  done
+
+lemma reply_unlink_tcb_tcb_sts_of_other:
```
lsf37, line 2341:
```suggestion
(* FIXME RT: move to AInvs *)
lemma reply_unlink_tcb_tcb_sts_of_other:
```

proof/refine/ARM/IpcCancel_R.thy
```diff
@@ -2366,69 +2323,29 @@ lemma restart_thread_if_no_fault_corres:
...
-             | clarsimp simp: valid_obj_def valid_ep_def
-                              valid_obj'_def valid_ep'_def
-                              invs_valid_pspace projectKOs
-                              valid_sched_def valid_sched_action_def
-             | erule obj_at_valid_objsE
-             | drule ko_at_valid_objs'
-             | rule conjI | clarsimp simp: invs'_def valid_state'_def)+
-    sorry
-qed
+lemma restart_thread_if_no_fault_tcb_sts_of_other:
```
lsf37, line 2331:
```suggestion
(* FIXME RT: move to AInvs *)
lemma restart_thread_if_no_fault_tcb_sts_of_other:
```

proof/refine/ARM/IpcCancel_R.thy
```diff
@@ -2521,6 +2439,445 @@ lemma possibleSwitchTo_all_invs_but_ct_not_inQ':
...
+
+(* FIXME RT Michael: also in Victor's PR *)
+lemma replyUnlink_weak_sch_act_wf[wp]:
+  "\<lbrace>\<lambda>s. weak_sch_act_wf (ksSchedulerAction s) s \<and> sch_act_not t s\<rbrace>
+   replyUnlink r t
+   \<lbrace>\<lambda>_ s. weak_sch_act_wf (ksSchedulerAction s) s\<rbrace>"
+  unfolding replyUnlink_def setReplyTCB_def getReplyTCB_def
+  by (wpsimp wp: hoare_vcg_imp_lift hoare_vcg_all_lift gts_wp'
+           simp: weak_sch_act_wf_def)
+
```
lsf37, line 2459:
```suggestion

(* FIXME RT: move to AInvs *)
```

proof/refine/ARM/IpcCancel_R.thy
```diff
@@ -2521,6 +2439,445 @@ lemma possibleSwitchTo_all_invs_but_ct_not_inQ':
...
+lemma restart_thread_if_no_fault_reply_tcb_reply_at[wp]:
+  "restart_thread_if_no_fault t  \<lbrace>reply_tcb_reply_at ((=) (Some t')) r_opt\<rbrace>"
+  apply (clarsimp simp: sk_obj_at_pred_def)
+  apply (wpsimp wp: restart_thread_if_no_fault_obj_at_impossible')
+  done
+
+lemma reply_unlink_tcb_reply_tcb_reply_at_other:
```
lsf37, line 2466:
```suggestion
(* FIXME RT: move to AInvs *)
lemma reply_unlink_tcb_reply_tcb_reply_at_other:
```

proof/refine/ARM/IpcCancel_R.thy
```diff
@@ -2521,6 +2439,445 @@ lemma possibleSwitchTo_all_invs_but_ct_not_inQ':
...
+  "cancel_all_ipc_loop_body t
+    \<equiv> do st \<leftarrow> get_thread_state t;
+          reply_opt \<leftarrow> case st of Structures_A.thread_state.BlockedOnReceive x r_opt xa
+                                    \<Rightarrow> return r_opt
+                                | _ \<Rightarrow> return None;
+          y \<leftarrow> when (reply_opt \<noteq> None) $ reply_unlink_tcb t (the reply_opt);
+          restart_thread_if_no_fault t
+       od"
+
+lemma cancel_all_ipc_loop_body_reply_unlink_ts_pred_other:
```
lsf37, line 2491:
```suggestion
(* FIXME RT: move to AInvs *)
lemma cancel_all_ipc_loop_body_reply_unlink_ts_pred_other:
```

proof/refine/ARM/IpcCancel_R.thy
```diff
@@ -2521,6 +2439,445 @@ lemma possibleSwitchTo_all_invs_but_ct_not_inQ':
...
+
+crunches replyUnlink
+  for sch_act_not[wp]: "\<lambda>s. sch_act_not t s"
+  and typ_at'[wp]: "\<lambda>s. P (typ_at' T p s)"
+  (wp: crunch_wps)
+
+lemmas replyUnlink_typ_ats[wp] = typ_at_lifts[OF replyUnlink_typ_at']
+
+(* FIXME RT: is there an equivalent to typ_at_lifts? This fact doesn't seem to be produced
+             by reply_unlink_tcb_typ_at *)
```
lsf37, line 2648: There is, yes: `abs_typ_at_lifts`

lsf37: (and it's better to use that than crunching it, so I'd recommend tweaking this one)

proof/refine/ARM/IpcCancel_R.thy
```diff
@@ -2521,6 +2439,445 @@ lemma possibleSwitchTo_all_invs_but_ct_not_inQ':
...
+  done
+
+lemma IdleEP_versus_SendEP_and_RecvEP:
+  "\<lbrakk>ep = Structures_A.endpoint.IdleEP \<Longrightarrow> P;
+    \<And>q. ep = Structures_A.endpoint.SendEP q \<or> ep = Structures_A.endpoint.RecvEP q \<Longrightarrow> P\<rbrakk>
+   \<Longrightarrow> P"
+  using Structures_A.endpoint.exhaust by auto
+
+lemma not_IdleEP_case_split:
+  "ep \<noteq> Structures_A.endpoint.IdleEP
```
lsf37, line 2716: I think `Structures_A.IdleEP` would be clear enough. Isabelle will expand it out to its longest form when the name is not unique, but we don't have to follow that in input. Short is usually easier to read.

proof/refine/ARM/IpcCancel_R.thy
lsf37: I'm fine with these, they do look like they're useful in case distinctions.

lsf37 (approved):
Nice. I'd mark the abstract-only lemmas (have tried to put in commit suggestions, but might have missed a few). Or, if you're feeling virtuous you can move them now already ;-)

proof/refine/ARM/IpcCancel_R.thy
mbrcknl: Now I'm curious why you say these are a pain to apply (e.g., they need `subst`). If your assumptions tell you that `ep` is not `IdleEP`, I would have hoped that `simp` would work with this.

The way you've arranged the proof below, this won't quite work, but I think it's close. I'll comment further on that part of the proof...

proof/refine/ARM/IpcCancel_R.thy
```diff
@@ -2521,6 +2439,445 @@ lemma possibleSwitchTo_all_invs_but_ct_not_inQ':
...
+     apply (case_tac "ep = Structures_A.endpoint.IdleEP")
+      apply (prop_tac "ep' = Structures_H.endpoint.IdleEP")
+       apply (simp add: ep_relation_def get_ep_queue_def)
+      apply clarsimp
+     apply (rule_tac ep=ep in IdleEP_versus_SendEP_and_RecvEP)
+      apply (simp add: ep_relation_def get_ep_queue_def, rename_tac q)
+     apply (subst not_IdleEP_case_split)
+      apply blast
+     apply (subst not_IdleEP_case_split')
+      apply (fastforce simp: ep_relation_def get_ep_queue_def)
```
mbrcknl, line 2808: If you wanted both `case_split` lemmas to work with `simp` here (rather than needing `subst`), you would need assumptions in your context `ep =/= IdleEP` and `ep' =/= IdleEP`. You don't quite manage that here.

Firstly, your `case_tac` only generates evidence that `ep =/= IdleEP`. You could get it for `ep'` as well with something like this:
```
  apply (case_tac "ep = Structures_A.IdleEP \/ ep' = Structures_H.IdleEP")
   (* Just smash everything out in the trivial case. *)
   apply (case_tac ep; case_tac ep'; simp add: ep_relation_def get_ep_queue_def)
  (* The following clarsimp should produce both ep =/= IdleEP and ep' =/= IdleEP,
     and from that, I would hope the case_split rules would work as conditional simp rules. *)
  apply (clarsimp simp: not_IdleEP_case_split not_IdleEP_case_split')
```
That's just off the top of my head, so there's a fair chance it won't quite work. If it doesn't work, I'd be interested to know why.

I'm not saying that you need to change this. I'm just here to provoke some discussion about how to deal with annoying corners of these kinds of proofs. :-)

mbrcknl (approved):

proof/refine/ARM/IpcCancel_R.thy
```diff
@@ -2521,6 +2439,445 @@ lemma possibleSwitchTo_all_invs_but_ct_not_inQ':
...
+
+abbreviation (input)
+  "cancel_all_ipc_loop_body t
+    \<equiv> do st \<leftarrow> get_thread_state t;
+          reply_opt \<leftarrow> case st of Structures_A.thread_state.BlockedOnReceive x r_opt xa
+                                    \<Rightarrow> return r_opt
+                                | _ \<Rightarrow> return None;
+          y \<leftarrow> when (reply_opt \<noteq> None) $ reply_unlink_tcb t (the reply_opt);
+          restart_thread_if_no_fault t
+       od"
```
mbrcknl, line 2489: In an out-of-band chat, @michaelmcinerney commented that getting helper lemmas to apply can sometimes be fiddly. For example, in the proof of `cancel_all_ipc_corres`, we apply `cancel_all_ipc_corres_helper`. The latter is phrased in terms of this and the following abbreviation. This means that the abbreviations need to be phrased to exactly match the whatever the goal happens to be at the point the helper lemma is applied. The abbreviation is therefore sensitive to the preceding proof steps.

An alternative is to make this a `definition`. This requires extra steps to unfold the definition in the helper lemma proof, and to fold it in the `corres` proof. On the other hand, it allows you to do that folding much earlier than the point where you apply the helper lemma. Then the body of the definition is not longer sensitive to the intervening proof steps. In particular, you can fold before you do any proof steps. Then `cancel_all_ipc_loop_body` can look exactly like the corresponding part of the spec.

If you want, you can then take this a step further, and move `cancel_all_ipc_loop_body` all the way up to the spec, and use it in the definition of `cancel_all_ipc`. That would require you to update some other proofs, but there are ways around that, too, e.g.
```
definition cancel_all_ipc where
  cancel_all_ipc_def2: "cancel_all_ipc epptr == ..."

lemmas cancel_all_ipc_def = cancel_all_ipc_def2[unfolded cancel_all_ipc_loop_body_def]
```

proof/refine/ARM/IpcCancel_R.thy
michaelmcinerney: Oh I didn't realise this. Cool!

proof/refine/ARM/IpcCancel_R.thy
michaelmcinerney: Very good points. The thought of making this a definition did cross my mind (like the aborted `valid_dom_schedule'` experiment) but my mind immediately jumped to putting this in the spec, and I wasn't so sure about that. I think I'll go for making them definitions in Refine (with a FIXME RT: for the abstract version), but I'll mull it over a bit more

proof/refine/ARM/IpcCancel_R.thy
michaelmcinerney: Done, thanks!

proof/refine/ARM/IpcCancel_R.thy
michaelmcinerney: This did work well, thanks! In using simp to perform the splitting, it will want to use `K_bind_def`, which messes up the helper lemma. So here I do `simp del: K_bind_def`. I think this is the right trade-off though; it's now only the only instance where I have to delete something from the simpset to get a helper lemma working, and the splitting is more straightforward

proof/refine/ARM/IpcCancel_R.thy
michaelmcinerney: I've gone with the FIXME RT: move to AInvs for the moment. I'm guessing some of these will be useful in the AInvs proofs, but it'd probably take a bit too much time to figure out exactly how

mktnk3 (approved):
Nice work. A lot of cleanup, too. I am glad you could sort out the loop conditions, etc. Thanks.

PR #75, opened 2020-09-06
Update set_next_interrupt

MitchellBuckley:
It appears that set_next_interrupt was not in line with master MCS kernel. This spec update should bring us back in line with master.

See https://github.com/seL4/seL4/blob/master/src/kernel/thread.c#L566

### DIFF
--- a/proof/invariant-abstract/SchedContext_AI.thy
+++ b/proof/invariant-abstract/SchedContext_AI.thy
@@ -727,7 +727,7 @@ lemma refill_unblock_check_bound_sc_tcb_at [wp]:
   by (wpsimp simp: refill_unblock_check_def get_refills_def is_round_robin_def)
 
 lemma set_next_interrupt_invs[wp]: "\<lbrace>invs\<rbrace> set_next_interrupt \<lbrace>\<lambda>rv. invs\<rbrace>"
-  by (wpsimp wp: hoare_drop_imp get_sched_context_wp set_next_timer_interrupt_invs
+  by (wpsimp wp: hoare_drop_imp get_sched_context_wp dmo_setDeadline
            simp: set_next_interrupt_def)
 
 lemma valid_state_consumed_time_update[iff]:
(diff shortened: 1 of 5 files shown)


spec/abstract/Schedule_A.thy
```diff
@@ -89,17 +89,24 @@ where
...
      rq \<leftarrow> gets release_queue;
-     new_thread_time \<leftarrow> if rq = [] then return new_thread_time else do
+     next_interrupt \<leftarrow> if rq = [] then return next_interrupt else do
        rqsc_opt \<leftarrow> get_tcb_obj_ref tcb_sched_context (hd rq);
        rqsc_ptr \<leftarrow> assert_opt rqsc_opt;
        rqsc \<leftarrow> get_sched_context rqsc_ptr;
-       return $ min (r_time (refill_hd rqsc)) new_thread_time
+       return $ min (r_time (refill_hd rqsc)) next_interrupt
      od;
-     set_next_timer_interrupt new_thread_time
```
mbrcknl, line 100: I think with this line deleted, `set_next_timer_interrupt` is unused.

mbrcknl (approved):
This looks good, but just to check my understanding... Does this actually change the meaning of `set_next_interrupt`? As far as I can tell, it just inlines `set_next_timer_interrupt`, reorders things, and adds a test for `num_domains > 1` (which will always be true in the current spec).

Even so, it probably makes sense to do this, to make the specs more aligned to the C.

spec/abstract/Schedule_A.thy
MitchellBuckley: Good catch. I check this and potentially delete the spec for that function.

MitchellBuckley:
> This looks good, but just to check my understanding... Does this actually change the meaning of `set_next_interrupt`? As far as I can tell, it just inlines `set_next_timer_interrupt`, reorders things, and adds a test for `num_domains > 1` (which will always be true in the current spec).
> 
> Even so, it probably makes sense to do this, to make the specs more aligned to the C.

Yes, I don't think this changes the semantics. It just makes it more closely aligned with the C and Haskell.

proof/invariant-abstract/SchedContext_AI.thy
```diff
@@ -727,7 +727,7 @@ lemma refill_unblock_check_bound_sc_tcb_at [wp]:
   by (wpsimp simp: refill_unblock_check_def get_refills_def is_round_robin_def)
 
 lemma set_next_interrupt_invs[wp]: "\<lbrace>invs\<rbrace> set_next_interrupt \<lbrace>\<lambda>rv. invs\<rbrace>"
-  by (wpsimp wp: hoare_drop_imp get_sched_context_wp set_next_timer_interrupt_invs
```
mktnk3, line 730: I believe you'll have to delete this lemma `set_next_timer_interrupt_invs` from AInvs now that you’ve removed the definition of the function?

mktnk3: It’s in ArchVSpace_AI.

mktnk3 (approved):
I agree that it makes sense to align the abstract spec closer to C as you did. Please check that AInvs passes the test before merging.

proof/invariant-abstract/SchedContext_AI.thy
MitchellBuckley: Cheers, done.

PR #76, opened 2020-09-07
Update switch_sched_context

MitchellBuckley:
This is a very simple improvement that removes some assertions from the abstract spec and tweaks the haskell spec to match the abstract spec.

No proof changes are required because this change is so small.

### DIFF
--- a/spec/abstract/Schedule_A.thy
+++ b/spec/abstract/Schedule_A.thy
@@ -108,15 +108,11 @@ where
     cur_th \<leftarrow> gets cur_thread;
     sc_opt \<leftarrow> get_tcb_obj_ref tcb_sched_context cur_th;
     scp \<leftarrow> assert_opt sc_opt;
-    sc' \<leftarrow> get_sched_context scp;
+    sc' \<leftarrow> get_sched_context cur_sc;
 
     when (scp \<noteq> cur_sc \<and> 0 < sc_refill_max sc') $ do
       modify (\<lambda>s. s\<lparr>reprogram_timer := True\<rparr>);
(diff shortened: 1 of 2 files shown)


lsf37 (approved):
Cool. I'm assuming we haven't reached `switch_sched_context` in Refine yet and that's why there is no impact?

mbrcknl (approved):

MitchellBuckley:
> Cool. I'm assuming we haven't reached `switch_sched_context` in Refine yet and that's why there is no impact?

Exactly. I was going to provide corres proofs along with these spec changes, but I've diverged a lot already and want to converge quickly if I can.

spec/abstract/Schedule_A.thy
```diff
@@ -108,15 +108,11 @@ where
     cur_th \<leftarrow> gets cur_thread;
     sc_opt \<leftarrow> get_tcb_obj_ref tcb_sched_context cur_th;
     scp \<leftarrow> assert_opt sc_opt;
-    sc' \<leftarrow> get_sched_context scp;
+    sc' \<leftarrow> get_sched_context cur_sc;
```
mktnk3, line 111: Good catch, this one. This was actually a bug, wasn’t it? The refill_max check is for cur_sc, not cur_thread.

mktnk3 (approved):

spec/abstract/Schedule_A.thy
MitchellBuckley: I don't know whether it was a bug in the C, but it was wrong in the abstract for sure.

PR #77, opened 2020-09-07
Update commit_time

MitchellBuckley:
This updates both abstract and haskell spec to ensure `commit_time` corresponds with the C.

### DIFF
--- a/proof/invariant-abstract/DetSchedSchedule_AI.thy
+++ b/proof/invariant-abstract/DetSchedSchedule_AI.thy
@@ -9185,6 +9185,27 @@ lemma refill_budget_check_refill_ready_offset_ready_and_sufficient:
   apply (simp add: word_le_nat_alt, blast intro: order_trans unat_plus_gt)
   done
 
+(* FIXME RT: move and rename *)
+lemma MIN_BUDGET_subtraction_helper[elim!]:
+  "2 * MIN_BUDGET \<le> b \<Longrightarrow> MIN_BUDGET \<le> b - MIN_BUDGET"
+  apply (rule word_sub_move)
+  apply (metis mult_2)
+  using MIN_BUDGET_le_MIN_SC_BUDGET
+  by (clarsimp simp: MIN_SC_BUDGET_def)
+
+lemma refill_budget_check_is_refill_sufficient:
+  "\<lbrace>\<lambda>s. if sc_ptr = cur_sc s
+        then valid_refills (cur_sc s) s \<and> current_time_bounded 0 s
+             \<and> cur_sc_offset_ready usage s \<and> cur_sc_offset_sufficient usage s
+         else is_refill_sufficient 0 sc_ptr s\<rbrace>
+   refill_budget_check usage
+   \<lbrace>\<lambda>_ s. is_refill_sufficient 0 sc_ptr s\<rbrace>"
+  unfolding refill_budget_check_def get_sc_refill_ready_def
+  apply (wpsimp wp: set_refills_wp is_round_robin_wp)
+  by (auto simp: pred_map_def obj_at_def vs_all_heap_simps refill_sufficient_def refill_capacity_def
+                 valid_refills_def Let_def
+          split: if_splits)
+
 crunches refill_budget_check
   for etcb_eq[wp]: "etcb_eq p d t"
   (wp: crunch_wps simp: crunch_simps)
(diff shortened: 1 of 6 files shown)


spec/haskell/src/SEL4/Object/SchedContext.lhs
```diff
@@ -643,7 +643,8 @@ This module uses the C preprocessor to select a target architecture.
 >     when (numDomains > 1) $ do
```
mbrcknl, line 643: In one of your other PRs on a different function, you included a test for `num_domains` in the abstract, but you haven't done that here. Even though the test is always true in the current specification, we should try to be consistent about this.

spec/haskell/src/SEL4/Object/SchedContext.lhs
```diff
@@ -643,7 +643,8 @@ This module uses the C preprocessor to select a target architecture.
...
 >       consumed <- getConsumedTime
 >       domainTime <- getDomainTime
->       setDomainTime (domainTime - consumed)
+>       time' <- return $ if (domainTime < consumed) then 0 else (domainTime - consumed)
```
mbrcknl, line 646: I may not have kept up with what's going on here. The C currently doesn't appear to clip this at zero. Is there a C change in the pipeline?

mbrcknl (approved):

spec/haskell/src/SEL4/Object/SchedContext.lhs
MitchellBuckley: Excellent point. I'll chase it down.

spec/haskell/src/SEL4/Object/SchedContext.lhs
MitchellBuckley: Mmmm, not yet. I suppose I should do it all at once. Good point.

MitchellBuckley: This is now in seL4 master branch. Thanks for prompting this.

spec/haskell/src/SEL4/Object/SchedContext.lhs
MitchellBuckley: Done.

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -9459,33 +9495,41 @@ lemma refill_budget_check_round_robin_sc_refills_update_unspecified:
...
   supply if_split[split del]
   apply (simp add: commit_time_def)
   apply (rule hoare_seq_ext[OF _ gets_sp])
-  apply (rule hoare_seq_ext[OF _ gets_sp])
   apply (rule hoare_seq_ext[OF _ get_sched_context_sp])
   apply (rule hoare_seq_ext[where B="\<lambda>_. released_ipc_queues"], wpsimp)
   apply (rule hoare_when_cases, simp)
+  apply (rule hoare_seq_ext[OF _ gets_sp])
+  apply (rename_tac csc sc consumed)
   apply (rule hoare_seq_ext[where B="\<lambda>_. released_ipc_queues"], wpsimp)
```
michaelmcinerney, line 9512: I haven't checked in Isabelle, but if you're going to have this `hoare_seq_ext` with `B` just equal to the original postcondition, do you want to instead change the `gets_sp` line before to a regular `hoare_seq_ext_skip`? You should end up in the same spot. Minor though, of course

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -9459,33 +9495,41 @@ lemma refill_budget_check_round_robin_sc_refills_update_unspecified:
...
-   apply (clarsimp simp: obj_at_kh_kheap_simps vs_all_heap_simps)
-  apply (simp add: released_ipc_queues_defs)
-  apply (erule allEI)
-  by (fastforce simp: heap_upd_def vs_all_heap_simps)
+  apply (rule hoare_weaken_pre)
+   apply (rule_tac R="cur_sc_more_than_ready and current_time_bounded 0
+                      and (\<lambda>s. consumed = consumed_time s) and cur_sc_active
+                      and (\<lambda>s. valid_refills (cur_sc s) s)"
+          in released_ipc_queues_lift_pre_conj)
+          apply (wpsimp+)[6]
```
michaelmcinerney, line 9520: I'm again a little scared of multiples wpsimps in a row. Are you able to instead have the following wpsimp with a +?

michaelmcinerney (approved):
Cool!

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -15506,26 +15551,9 @@ lemma refill_budget_check_round_robin_sc_refill_max_sc_at:
...
-      apply (rule hoare_seq_ext_skip
-             , wpsimp wp: refill_budget_check_round_robin_sc_refill_max_sc_at)
-      apply (wpsimp wp: update_sched_context_wp simp: sc_at_pred_n_def obj_at_def)
-     apply (rule hoare_seq_ext_skip, wpsimp wp: refill_budget_check_sc_refill_max_sc_at)
-     apply (wpsimp wp: update_sched_context_wp simp: sc_at_pred_n_def obj_at_def)
-
-     \<comment> \<open>remaining cases\<close>
-    apply (wpsimp wp: update_sched_context_wp simp: sc_at_pred_n_def obj_at_def)+
-  done
+crunches commit_time
```
mktnk3, line 15554: Nice :)

mktnk3 (approved):
Thanks for chasing this up, Mitch!

mbrcknl (approved):

proof/invariant-abstract/DetSchedSchedule_AI.thy
MitchellBuckley: In this case I'm using `hoare_seq_ext` to skip the last command, not the first. So the gets_sp commands are still needed because the precondition is maintained. But I've added `hoare_seq_ext_skip'` to facilitate skipping the last command, and that's a lot neater!

proof/invariant-abstract/DetSchedSchedule_AI.thy
MitchellBuckley: That did work. I don't love it, but I'll do it that way since it seems to be the style we're going for.

PR #79, opened 2020-09-08
Remove runnable

michaelmcinerney:
This removes the condition on the thread state from `valid_queues` by using asserts instead (similar to the `sym_refs_asrt`)

### DIFF
--- a/proof/refine/ARM/CSpace_R.thy
+++ b/proof/refine/ARM/CSpace_R.thy
@@ -1082,27 +1082,13 @@ lemma bitmapQ_no_L2_orphans_lift:
 
 lemma valid_queues_lift_asm:
   assumes tat1: "\<And>d p tcb. \<lbrace>obj_at' (inQ d p) tcb and Q \<rbrace> f \<lbrace>\<lambda>_. obj_at' (inQ d p) tcb\<rbrace>"
-  and     tat2: "\<And>tcb. \<lbrace>st_tcb_at' runnable' tcb and Q \<rbrace> f \<lbrace>\<lambda>_. st_tcb_at' runnable' tcb\<rbrace>"
   and     prq: "\<And>P. \<lbrace>\<lambda>s. P (ksReadyQueues s) \<rbrace> f \<lbrace>\<lambda>_ s. P (ksReadyQueues s)\<rbrace>"
   and     prqL1: "\<And>P. \<lbrace>\<lambda>s. P (ksReadyQueuesL1Bitmap s)\<rbrace> f \<lbrace>\<lambda>_ s. P (ksReadyQueuesL1Bitmap s)\<rbrace>"
   and     prqL2: "\<And>P. \<lbrace>\<lambda>s. P (ksReadyQueuesL2Bitmap s)\<rbrace> f \<lbrace>\<lambda>_ s. P (ksReadyQueuesL2Bitmap s)\<rbrace>"
-  shows   "\<lbrace>Invariants_H.valid_queues and Q\<rbrace> f \<lbrace>\<lambda>_. Invariants_H.valid_queues\<rbrace>"
-  proof -
-    have tat:  "\<And>d p tcb. \<lbrace>obj_at' (inQ d p) tcb and st_tcb_at' runnable' tcb and Q\<rbrace> f
-                         \<lbrace>\<lambda>_. obj_at' (inQ d p) tcb and st_tcb_at' runnable' tcb\<rbrace>"
-      apply (rule hoare_chain [OF hoare_vcg_conj_lift [OF tat1 tat2]])
-       apply (fastforce)+
-       done
-    have tat_combined: "\<And>d p tcb. \<lbrace>obj_at' (inQ d p and runnable' \<circ> tcbState) tcb and Q\<rbrace> f
-                         \<lbrace>\<lambda>_. obj_at' (inQ d p and runnable' \<circ> tcbState) tcb\<rbrace>"
-      apply (rule hoare_chain [OF tat])
-       apply (fastforce simp add: obj_at'_and pred_tcb_at'_def o_def)+
-      done
-    show ?thesis unfolding valid_queues_def valid_queues_no_bitmap_def
-      by (wp tat_combined prq prqL1 prqL2 valid_bitmapQ_lift bitmapQ_no_L2_orphans_lift
-             bitmapQ_no_L1_orphans_lift hoare_vcg_all_lift hoare_vcg_conj_lift hoare_Ball_helper)
-         simp_all
-  qed
+  shows   "\<lbrace>valid_queues and Q\<rbrace> f \<lbrace>\<lambda>_. valid_queues\<rbrace>"
+  unfolding valid_queues_def valid_queues_no_bitmap_def
+  by (wpsimp wp: tat1 prq prqL1 prqL2 valid_bitmapQ_lift bitmapQ_no_L2_orphans_lift
+                 bitmapQ_no_L1_orphans_lift hoare_vcg_all_lift hoare_Ball_helper)
 
 lemmas valid_queues_lift = valid_queues_lift_asm[where Q="\<lambda>_. True", simplified]
 
--- a/proof/refine/ARM/IpcCancel_R.thy
+++ b/proof/refine/ARM/IpcCancel_R.thy
@@ -448,12 +448,15 @@ lemma blocked_cancel_ipc_corres:
 
 lemma cancel_signal_corres:
   "corres dc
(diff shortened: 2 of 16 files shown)


spec/haskell/src/SEL4/Object/Structures.lhs
```diff
@@ -575,6 +575,4 @@ Various operations on the free index of an Untyped cap.
 >         empty = capFreeIndex cap == maxFreeIndex (capBlockSize cap)
 >         startPtr = getFreeRef (capPtr cap) (capFreeIndex cap)
 >         endPtr = capPtr cap + PPtr (2 ^ capBlockSize cap) - 1
-> untypedZeroRange _ = Nothing
```
michaelmcinerney, line 578: Maybe revert this?

proof/refine/ARM/CSpace_R.thy
```diff
@@ -1082,27 +1082,13 @@ lemma bitmapQ_no_L2_orphans_lift:
...
-      done
-    show ?thesis unfolding valid_queues_def valid_queues_no_bitmap_def
-      by (wp tat_combined prq prqL1 prqL2 valid_bitmapQ_lift bitmapQ_no_L2_orphans_lift
-             bitmapQ_no_L1_orphans_lift hoare_vcg_all_lift hoare_vcg_conj_lift hoare_Ball_helper)
-         simp_all
-  qed
+  shows   "\<lbrace>valid_queues and Q\<rbrace> f \<lbrace>\<lambda>_. valid_queues\<rbrace>"
+  unfolding valid_queues_def valid_queues_no_bitmap_def
+  by (wpsimp wp: tat1 prq prqL1 prqL2 valid_bitmapQ_lift bitmapQ_no_L2_orphans_lift
+                 bitmapQ_no_L1_orphans_lift hoare_vcg_all_lift hoare_Ball_helper)
```
corlewis, line 1091: Great job improving this proof!

proof/refine/ARM/IpcCancel_R.thy
```diff
@@ -2195,11 +2064,17 @@ lemma (in delete_one_conc) suspend_invs'[wp]:
...
    apply (wpsimp simp: updateRestartPC_def)
-  apply (wpsimp wp: sts_tcbSchedDequeue_invs' tcbReleaseRemove_invs'
-                    schedContextCancelYieldTo_invs')
-  apply (fastforce simp: tcb_st_refs_of'_def st_tcb_at'_def obj_at'_def projectKOs)
+  apply (rule_tac B="\<lambda>_ s. invs' s
+                           \<and> sch_act_simple s
+                           \<and> tcb_at' t s
+                           \<and> t \<noteq> ksIdleThread s
+                           \<and> st_tcb_at' ((=) Inactive) t s"
+               in hoare_seq_ext[rotated])
```
corlewis, line 2074: Very minor, but if these manual instantiations are definitely needed then you could reduce some duplication by giving a name to the original precondition in the lemma statement.

proof/refine/ARM/Schedule_R.thy
```diff
@@ -1102,8 +1119,12 @@ lemma tcbSchedDequeue_invs_no_cicd'[wp]:
...
 
 lemma switchToThread_invs_no_cicd':
-  "\<lbrace>invs_no_cicd' and st_tcb_at' runnable' t and tcb_in_cur_domain' t \<rbrace> ThreadDecls_H.switchToThread t \<lbrace>\<lambda>rv. invs' \<rbrace>"
-  apply (simp add: Thread_H.switchToThread_def)
+  "\<lbrace>invs_no_cicd' and st_tcb_at' runnable' t and tcb_in_cur_domain' t \<rbrace>
+   do  y <- ARM_H.switchToThread t;
+       y <- tcbSchedDequeue t;
+       setCurThread t
+    od
+   \<lbrace>\<lambda>rv. invs' \<rbrace>"
```
corlewis, line 1127: Could you give a quick explanation for why this is now a lemma about multiple functions? It's not immediately obvious to me, and if nothing else maybe the name should change.

corlewis (approved):
Nice work! It's great to see so many proofs become a lot simpler or even be deleted completely.

proof/refine/ARM/IpcCancel_R.thy
michaelmcinerney: Good spotting. Done!

mbrcknl (approved):
Very nice!

proof/refine/ARM/Schedule_R.thy
michaelmcinerney: Yes! So I added the `ready_qs_runnable` assert to `switchToThread`. However, if I had `switchToThread` in this lemma, then when we get to the lemma `switchToThread_lookupBitmapPriority_wp` below where this lemma is used, I can't generate the fact from the assert because it's hidden beneath the function definition. So I have this lemma here, then in the lemma below, I unfold `switchToThread`, use an sp rule to generate the fact from the assert, then the proof continues as before. 

But you're right that I should at least change the name. I think I'll just tack `_helper` on the end. Let me know if you have another preference though

michaelmcinerney: Well, actually, this lemma is used in a few other places, so I'm not sure just calling it "helper" is best. Would a comment beforehand be best, then?

michaelmcinerney: Ignore my previous comment; the other times the lemma is mentioned, it's deleted from the wp set. Which is strange because it's not in the wp set. Hmm. I've decided to delete these lines (though maybe if they're useful because they indicate that we certainly do not want that lemma being used in wp, I can add them back). I've decided to just go with `switchToThread_invs'_helper`

spec/haskell/src/SEL4/Object/Structures.lhs
MitchellBuckley: Yes please :)

proof/refine/ARM/Schedule_R.thy
```diff
@@ -712,8 +712,13 @@ lemma tcbSchedDequeue_invs'[wp]:
   done
 
 lemma cur_thread_update_corres:
-  "corres dc \<top> \<top> (modify (cur_thread_update (\<lambda>_. t))) (setCurThread t)"
+  "corres dc (pspace_aligned and pspace_distinct and valid_ready_qs) \<top>
```
MitchellBuckley, line 715: I'd like to test my understanding. Can you confirm?

1. You want to make valid_ready_queues' simpler by removing the runnable condition.
2. Because you were using the runnable condition in some hoare logic proofs, you need to add an assertion to continue using this invariant in those places.
3. Because there is now an extra assertion in the haskell, you need to show it holds in the corresponding corres proofs.

Thus, while you make it easier to do hoare logic with valid_ready_queues', you make some corres proofs a little harder. Is that right?

MitchellBuckley (approved):
I like this! Congrats on actually reducing the line count too!

proof/refine/ARM/Schedule_R.thy
MitchellBuckley: (I think this is a good way forward and really like the approach in general, just trying to understand why this corres proof now requires stronger preconditions.)

michaelmcinerney: Yes, exactly! 

The corres proofs are a little harder. Putting the assert at the beginning of functions makes it easy to use the assert in Hoare logic proofs by using `hoare_seq_ext[OF _ stateAssert_sp]` to grab the assertion. So the Hoare logic poofs generally don't change much (often they are easier in some spots because the postcondition is weaker).

The little method `add_ready_qs_runnable` together with the rule `corres_stateAssert_add_assumption` is really all you need. So 3 lines or so per assertion per corres proof (plus the short cross lemma and the new method)

spec/haskell/src/SEL4/Object/Structures.lhs
michaelmcinerney: Donezo

mktnk3 (approved):
This is very cool, thanks, Michael! I wonder if we’ll have more in this series. Maybe?

michaelmcinerney:
> This is very cool, thanks, Michael! I wonder if we’ll have more in this series. Maybe?

Thanks! I've got this approach down pat, now, so if there are more possibilities, I'll definitely go for it! Pretty soon we'll be doing

`crunches callKernel
  for invs'[wp]: invs'`

proof/refine/ARM/Schedule_R.thy
lsf37: They're likely removed there because it was in the set at some previous time until someone realised that was not a good idea, but did not clean up the other mentions.

PR #80, opened 2020-09-09
asmrefine: support small array index types

mbrcknl:
This is a small preemptive update to asmrefine to support a forthcoming kernel patch that will speed up binary verification for the RISC-V MCS kernel. The kernel patch will compress read-only data, by making `register_t` a `typedef` of `uint8_t`, whereas it was previously `word_t`. Lists of registers used in IPC messages are therefore correspondingly smaller, and cause less overhead in binary verification. This asmrefine patch adds support for indexing into the user context with a variable of type `uint8_t`.

### DIFF
--- a/lib/Word_Lib/Word_Lemmas_Internal.thy
+++ b/lib/Word_Lib/Word_Lemmas_Internal.thy
@@ -400,4 +400,16 @@ lemma le_smaller_mask:
   "\<lbrakk> x \<le> mask n; n \<le> m \<rbrakk> \<Longrightarrow> x \<le> mask m"
   by (erule (1) order.trans[OF _ mask_mono])
 
+(* The strange phrasing and ordering of assumptions is to support using this as a
+   conditional simp rule when y is a concrete value. For example, as a simp rule,
+   it will solve a goal of the form:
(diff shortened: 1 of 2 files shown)


mbrcknl:
[L4V-TESTBOARD-603](https://bamboo.ts.data61.csiro.au/browse/L4V-TESTBOARD-603)

MitchellBuckley (approved):

MitchellBuckley:
This is really neat matt! Great to see that the solution is so simple (or at least that you've made it look simple :))

mbrcknl:
@MitchellBuckley If you're curious how this went down: I was waiting for a decompile, so I thought I'd have a look at this. At [step 5 in TestGraphRefine](https://github.com/seL4/l4v/blob/a424d55e3ef6904ae26ffd0c41f90a87162bc5c5/proof/asmrefine/TestGraphRefine.thy#L202) for `getRegister`, there is a message that shows a goal of the form (like in the comment):
```
UCAST(8 --> 32) x < 0x20 ==> unat x < 32
```
So it was just a matter of crafting a suitable word lemma and inserting into the relevant place in the pipeline (and that was also identified in the same message).

This subgoal is a side-condition generated by a proof of a pointer validity assertion. The `x` is the array index, so the subgoal relates to an array bound check.

PR #81, opened 2020-09-10
haskell: deal with existing "~/.stack"

lsf37:
Bring happiness to both bamboo and github: create ~/.stack, but don't fail if it's there.

(forgot to cherry-pick this one, and it's of course the only one that github tests are not going to pick up)


mbrcknl (approved):

lsf37:
Thanks for the quick approval :-)

mbrcknl:
> Thanks for the quick approval :-)

I'm just in it for the activity graph.

PR #82, opened 2020-09-11
clib: add a `hoarep_rewrite` method

mbrcknl:
This is like `ccorres_rewrite`, but for `hoarep`, and uses the same
infrastructure.

The interaction between the `simpl_rewrite` locale and the
`simpl_rewrite` method was confusing, and didn't work well with multiple
interpretations. We replace it with a simple anonymous context. Since
that puts more things in the global namespace, we rename many of them.
The `simpl_rewrite` method is now parameterised by a `hom` fact which
determines the predicate under which we are rewriting.

This also includes a slight generalisation of `exec_eq_is_valid_eq`,
which allows a similar generalisation of `hoarep_rewrite`.

### DIFF
--- a/lib/clib/Simpl_Rewrite.thy
+++ b/lib/clib/Simpl_Rewrite.thy
@@ -14,6 +14,30 @@ imports
   "Lib.Apply_Debug"
 begin
 
+text \<open>One layer of context around a Simpl program.
+
+      For example, if the current focus is the first branch of a @{term Cond},
+      then the context consists of a constructor @{text CondTC} which indicates
+      this is the context, and carries everything but but the focused branch,
+      i.e. the condition and the second branch.
+
+      The @{text CondFC} and @{text HandlerC} cases each carry an extra bit of
+      information, which indicates whether the left-hand branch @{text never_continues}.
+      The assumption here is that we always traverse Simpl programs left-to-right.
+      @{text Seq2C} does not require an extra bit, because we never enter the second
+      command of a @{term Seq} if the first command @{text never_continues}.\<close>
+
+datatype ('s,'p,'f) com_ctxt
+  = Seq1C "('s,'p,'f) com" \<comment> \<open>first command of a @{term Seq}\<close>
+  | Seq2C "('s,'p,'f) com" \<comment> \<open>second command of a @{term Seq}\<close>
+  | CondTC "'s bexp" "('s,'p,'f) com" \<comment> \<open>first branch of a @{term Cond}\<close>
+  | CondFC "'s bexp" "('s,'p,'f) com" bool \<comment> \<open>second branch of a @{term Cond}\<close>
+  | WhileC "'s bexp" \<comment> \<open>body of a @{term While}\<close>
+  | WhileAnnoC "'s bexp" "'s assn" "('s \<times> 's) assn" \<comment> \<open>body of a @{term whileAnno}\<close>
+  | GuardC "'f" "'s bexp" \<comment> \<open>body of a @{term Guard}\<close>
+  | TryC "('s,'p,'f) com" \<comment> \<open>body of a @{term Catch}\<close>
+  | HandlerC "('s,'p,'f) com" bool \<comment> \<open>handler of a @{term Catch}\<close>
+
 text \<open>Definitions and lemmas for reasoning about equivalence of Simpl programs.\<close>
 
 named_theorems C_simp
@@ -21,7 +45,7 @@ named_theorems C_simp_pre
 named_theorems C_simp_simps
 named_theorems C_simp_throws
 
-locale simpl_rewrite_base =
+context
   fixes \<Gamma> :: "'p \<Rightarrow> ('s,'p,'f) com option"
 begin
 
@@ -46,9 +70,7 @@ lemma com_eq_guard_False:
 
 text \<open>Most @{text com_eq} simplification rules will be unguarded, however.\<close>
 
-abbreviation
-  "com_equiv \<equiv> com_eq True"
-
+abbreviation "com_equiv \<equiv> com_eq True"
 notation com_equiv (infix "\<sim>" 10)
 
 text \<open>@{term "com_eq"} is an equivalence relation.\<close>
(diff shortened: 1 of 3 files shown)


mbrcknl:
[4V-TESTBOARD-606](https://bamboo.ts.data61.csiro.au/browse/L4V-TESTBOARD-606)

mbrcknl:
> [4V-TESTBOARD-606](https://bamboo.ts.data61.csiro.au/browse/L4V-TESTBOARD-606)

Bitbucket was down for backup for that one. Try again:
[L4V-TESTBOARD-607](https://bamboo.ts.data61.csiro.au/browse/L4V-TESTBOARD-607)

mbrcknl:
[L4V-TESTBOARD-607](https://bamboo.ts.data61.csiro.au/browse/L4V-TESTBOARD-607) is doomed.

And again: [L4V-TESTBOARD-608](https://bamboo.ts.data61.csiro.au/browse/L4V-TESTBOARD-608)

mbrcknl:
[L4V-TESTBOARD-609](https://bamboo.ts.data61.csiro.au/browse/L4V-TESTBOARD-609)

mbrcknl:
- [ ] fix commit message: "We replace it" -> "We replace the locale".

MitchellBuckley (approved):
This looks great to me. I can see that half of these changes are generalisations and renamings and cleanups on existing work. Unfortunately I have no experience with SIMPL and ccorres, but I'm happy to approve anyway. I hope that @lsf37 or @Xaphiosis will be able to give more detailed feedback.

mbrcknl:
Here's a quick summary of what's in the PR. There's not actually very much to it:
- It takes the `ccorres_rewrite` method, which can perform basic simplifications of Simpl programs (e.g. `if (true) c else d` --> `c`) when they're under a `ccorres` predicate, and reuses its underlying implementation to do the same thing when the Simpl program is under a Hoare triple (`hoarep`).

The content of PR is:
- Remove the `simpl_rewrite` locale. The method defined in the locale didn't seem to work well with multiple instantiations of the locale, and it was confusing anyway. The PR replaces the locale with an anonymous context block.
- Since converting the locale to an anonymous context block puts more things in the global namespace, we rename some things that had very generic names, to make them more specific and verbose. This renaming noise is the bulk of the content of the PR.
- Since the `simpl_rewrite` method still needs to be specialised to the different predicates we want to rewrite under, we give it a new `hom` parameter that takes a `com_eq_hom` fact corresponding to what was previously the locale assumption.
- We refactor the `ccorres_rewrite` method to work with the new shape of the `simpl_rewrite` method. The old proof of the locale assumption becomes a standalone `com_eq_hom` proof.
- We add a new method `hoarep_rewrite`. This one allows both naked `hoarep` predicates, and ones under a `forall` binder.

mbrcknl:
> This looks great to me. I can see that half of these changes are generalisations and renamings and cleanups on existing work. Unfortunately I have no experience with SIMPL and ccorres, but I'm happy to approve anyway. I hope that @lsf37 or @Xaphiosis will be able to give more detailed feedback.

Thanks! Since I think this is unlikely to be controversial, I'll go ahead and merge to unblock some other work. Force-push incoming...

lsf37:
Just to confirm that this looks very nice. Thanks for adding it @mbrcknl!

lib/clib/Simpl_Rewrite.thy
```diff
@@ -276,225 +300,219 @@ lemmas never_continues_Seqs =
...
+abbreviation "com_equiv_ex \<equiv> com_equiv \<Gamma>"
+notation com_equiv_ex (infix "\<sim>" 10)
+
 lemma
   assumes c3: "c3 \<sim> c"
   assumes c: "(c;;c) \<sim> c"
   shows "P (c;; Guard f UNIV (IF X THEN c ELSE c FI);; Cond {} Skip (Skip;;c2);; Skip;;
             (IF False THEN Skip ELSE SKIP;; TRY THROW CATCH c3 END FI;; SKIP))"
-  apply simpl_rewrite
+  apply (simpl_rewrite hom: hom)
```
Xaphiosis, line 502: I'm confused about why `hom` needs to be added manually now. Are there interesting implications of the need to do this?

lib/clib/Simpl_Rewrite.thy
```diff
@@ -276,225 +300,219 @@ lemmas never_continues_Seqs =
...
-  | GuardC "'f" "'s bexp" \<comment> \<open>body of a @{term Guard}\<close>
-  | TryC "('s,'p,'f) com" \<comment> \<open>body of a @{term Catch}\<close>
-  | HandlerC "('s,'p,'f) com" bool \<comment> \<open>handler of a @{term Catch}\<close>
-
 text \<open>Rewrite Simpl programs under a predicate @{term P} which is preserved by @{term com_eq}.\<close>
 
-locale simpl_rewrite = simpl_rewrite_base \<Gamma>
-  for \<Gamma> :: "'p \<Rightarrow> ('s,'p,'f) com option" +
+definition com_eq_hom :: "(('s,'p,'f) com \<Rightarrow> bool) \<Rightarrow> bool" where
+  "com_eq_hom P \<equiv> \<forall>c c'. (c \<sim> c') \<longrightarrow> P c' \<longrightarrow> P c"
```
Xaphiosis, line 323: what does `com_eq_hom` stand for? com program equivalent ... ?

Xaphiosis:
Thumbs up. Good to consolidate and improve things. Much appreciated.

lib/clib/Simpl_Rewrite.thy
mbrcknl: `simpl_rewrite` is the underlying implementation for both `ccorres_rewrite` and now `hoarep_rewrite`. The infrastructure for `simpl_rewrite` is abstracted over the predicate under which you're rewriting, so the actual rewriting needs to be justified by a theorem (the `com_eq_hom`) that expresses that the predicate allows rewriting up to `com` equivalence.

But `simpl_rewrite` isn't intended for the end-user. The `CCorres_Rewrite` theory creates the `ccorres_rewrite` and `hoarep_rewrite` methods by supplying a specific `com_eq_hom` theorem when they delegate the actual work to the underlying `simpl_rewrite` method.

Here, we're just doing a few examples and tests on some hypothetical predicate `P` that we've assumed for the `experiment` block. I didn't bother making a temporary method for that. But I suppose that potentially makes things a bit unclear. Do you think I should do this differently?

lib/clib/Simpl_Rewrite.thy
mbrcknl: I was thinking it's roughly a homomorphism from `com_eq` to  predicate `P`. That is, if `c` and `c'` are `com`-equivalent, then `P c` iff `P c'`. But perhaps "homomorphism" is a bit obscure. Can you think of something clearer?

mbrcknl: I suppose a comment to that effect would help.

Xaphiosis: Oh, that's obscure but ultimately survivable. I think I didn't get to "homomorphism" from "hom", that's all.

lib/clib/Simpl_Rewrite.thy
Xaphiosis: No, that's fine. Your explanation implies there are no interesting consequences for the user of that change. All good.

mbrcknl: Oh, and the reason this needs to be added manually, when it wasn't necessary before is that I've taken it out of the old `simpl_rewrite` locale. I was getting some weird method behaviour when I tried to instantiate the locale multiple times. I gave up trying to figure out what was going on, and just ripped out the locale. So now, the `simpl_rewrite` method doesn't have the locale context to know theorem justifies the rewriting.

lsf37: It sounds like you might have been hitting [VER-570], which says that you can instantiate Eisbach methods only once.

mbrcknl: Yes, I think that's what I was seeing. I instantiated for `hoarep_rewrite` after `ccorres_rewrite`, and found that `hoarep_rewrite` worked, but `ccorres_rewrite` stopped working.

mbrcknl:
In case anyone is curious about what prompted this: For [proofs about new implementations of `clz` and `ctz`](https://github.com/seL4/l4v/pull/87), I wanted to treat the manually unrolled loops the same way we treat actual loops, i.e. using an invariant. But that required writing out the "body" of the "loop" in a definition so we can write a theorem that it preserves the invariant. Since there's bit-shifting involved, the actual body contains many trivial guards. `hoarep_rewrite` gets rid of those, while otherwise preserving structure, so there's less to write out in the definition.

I can almost see a way to do these proofs without writing out those definitions, but it involves a bit more work that I want to avoid at the moment.

PR #92, opened 2020-09-14
misc lemmas for reply_remove_tcb_corres

mktnk3:
This contains more ready to present part of various lemmas that I came up with in the process of proving `reply_remove_tcb_corres`. The progress with this main corres rule itself is not included here.

This is not meant for merging as is, but I thought I’d better share these now in case anyone can use them.

The lemmas are roughly grouped into these categories:

- `sr_inv` and `corres_symb_exec_r_sr` for symbolic execution with `state_relation` preserving state update

- more `heap_path` lemmas

- `sym_refs (list_refs_of_replies’ s’)` and `replyNext`/`replyPrev`

- the concrete version of `sc_with_reply`, its properties, and crossing information from the abstract one

- `update_sched_context` corres rule for `scReply` update

- setting things up for `cleanReply_corres`

### DIFF
--- a/proof/refine/ARM/IpcCancel_R.thy
+++ b/proof/refine/ARM/IpcCancel_R.thy
@@ -507,18 +507,1460 @@ locale delete_one = delete_one_conc + delete_one_abs +
                (invs' and cte_at' (cte_map ptr))
           (cap_delete_one ptr) (cteDeleteOne (cte_map ptr))"
 
+
+(* FIXME RT: move to Lib *)
+(* for instantiations; only need to specify what you are adding *)
+lemma corres_inst_add:
+  "corres_underlying sr nf nf' r (P and Q) (P' and Q') f g
+   \<Longrightarrow> corres_underlying sr nf nf' r (P and Q) (P' and Q') f g" .
+
+(** sr_inv **)
+thm monadic_rewrite_def corres_underlying_def
+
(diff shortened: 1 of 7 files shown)


proof/refine/ARM/IpcCancel_R.thy
```diff
@@ -547,6 +547,532 @@ locale delete_one = delete_one_conc + delete_one_abs +
...
           (cap_delete_one ptr) (cteDeleteOne (cte_map ptr))"
 
+(** from here **)
+
+(* for instantiations; only need to specify what you are adding *)
+lemma corres_inst_add:
+  "corres_underlying sr nf nf' r (P and Q) (P' and Q') f g
+   \<Longrightarrow> corres_underlying sr nf nf' r (P and Q) (P' and Q') f g" .
+
+(** sr_inv **)
```
mktnk3, line 557: As always, naming is difficult. I’d like something short, but I am not fully convinced with this... any ideas?

mktnk3:
As I said in the MCS channel, I now introduced `sym_heap` and rewrote all the `replyNext/replyPrev` symmetry lemmas using it.
```
abbreviation sym_heap where
   "sym_heap hp hp' \<equiv> \<forall>p p'. hp p = Some p' \<longleftrightarrow> hp' p' = Some p"
```
The essence of this is I think the same as Matt’s one in DetSchedInvs_AI, but this is much much simpler, and this is just exactly what I need here. I think the concept of heaps is separate from that of projections, and it is probably worthwhile to formulate it separately.

PR #93, opened 2020-09-15
rt refine: progress on Finalise_R sorries

corlewis:

PR #95, opened 2020-09-16
Reply_R: Prove `replyRemoveTCB_invs'`

pierzchalski:
Step 1 of 100 towards `cancelIPC_invs'`.
Edit: turned out to be step 1 of 2, and also this now includes `cancelIPC_invs'`.

### DIFF
--- a/proof/refine/ARM/Bits_R.thy
+++ b/proof/refine/ARM/Bits_R.thy
@@ -97,6 +97,11 @@ lemma projectKO_reply:
   "(projectKO_opt ko = Some t) = (ko = KOReply t)"
   by (cases ko) (auto simp: projectKO_opts_defs)
 
+lemma reply_of'_KOReply[simp]:
+  "reply_of' (KOReply reply) = Some reply"
+  apply (clarsimp simp: projectKO_reply)
+  done
+
 lemma projectKO_sc:
   "(projectKO_opt ko = Some t) = (ko = KOSchedContext t)"
   by (cases ko) (auto simp: projectKO_opts_defs)
--- a/proof/refine/ARM/Invariants_H.thy
+++ b/proof/refine/ARM/Invariants_H.thy
@@ -356,8 +356,7 @@ definition live_ntfn' :: "notification \<Rightarrow> bool" where
                       \<or> (\<exists>ts. ntfnObj ntfn = WaitingNtfn ts)"
 
 definition live_reply' :: "reply \<Rightarrow> bool" where
-  "live_reply' reply \<equiv> bound (replyTCB reply) \<or> bound (replySc reply) \<or> bound (replyNext reply)
-                        \<or> bound (replyPrev reply)"
+  "live_reply' reply \<equiv> bound (replyTCB reply) \<or> bound (replyNext reply) \<or> bound (replyPrev reply)"
 
 primrec live' :: "Structures_H.kernel_object \<Rightarrow> bool" where
   "live' (KOTCB tcb) =
--- a/proof/refine/ARM/Invariants_H.thy
+++ b/proof/refine/ARM/Invariants_H.thy
@@ -2506,6 +2506,11 @@ lemma obj_at_ko_at'_eq:
   unfolding obj_at'_def
   by blast
 
+lemma ko_at'_replies_of':
+  "ko_at' reply ptr s \<Longrightarrow> replies_of' s ptr = Some reply"
+  apply (clarsimp simp: obj_at'_def projectKO_eq opt_map_def)
+  done
+
 lemma obj_at_aligned':
   fixes P :: "'a :: pspace_storable \<Rightarrow> bool"
   assumes oat: "obj_at' P p s"
@@ -2654,11 +2659,25 @@ lemma typ_at_lift_page_table_at'_strong:
          | fastforce)+
   done
 
+lemma typ_at_lift_valid_tcb_state'_strong:
+  assumes ep: "\<And>p. f \<lbrace>\<lambda>s. P (typ_at' EndpointT p s)\<rbrace>"
+      and reply: "\<And>p. f \<lbrace>\<lambda>s. P (typ_at' ReplyT p s)\<rbrace>"
+      and ntfn: "\<And>p. f \<lbrace>\<lambda>s. P (typ_at' NotificationT p s)\<rbrace>"
+  shows "f \<lbrace>\<lambda>s. P (valid_tcb_state' st s)\<rbrace>"
+  unfolding valid_tcb_state'_def valid_bound_reply'_def
+  apply (case_tac st
+         ; clarsimp split: option.splits
+         , wpsimp wp: hoare_vcg_imp_lift' hoare_vcg_all_lift hoare_vcg_conj_lift_N[where N=P]
+                           typ_at_lift_ep'_strong[OF ep] typ_at_lift_reply'_strong[OF reply]
+                           typ_at_lift_ntfn'_strong[OF ntfn])
+  done
+
 lemmas typ_at_lifts_strong =
   typ_at_lift_tcb'_strong typ_at_lift_ep'_strong
   typ_at_lift_ntfn'_strong typ_at_lift_cte'_strong
   typ_at_lift_reply'_strong typ_at_lift_sc'_strong
   typ_at_lift_page_directory_at'_strong
+  typ_at_lift_valid_tcb_state'_strong
   typ_at_lift_page_table_at'_strong
 
 lemma ko_wp_typ_at':
@@ -2796,9 +2815,13 @@ local
   val strong_thms = @{thms typ_at_lifts_strong[no_vars]};
   fun abstract_P term = Logic.all (Free ("P", @{typ "bool \<Rightarrow> bool"})) term
   fun abstract thm =
-    @{const Pure.imp} $
-      abstract_P (hd (Thm.prems_of thm)) $
-      Thm.concl_of thm
+    let
+      val prems = List.map abstract_P (Thm.prems_of thm);
+      fun imp [] = Thm.concl_of thm
+        | imp (p :: pms) = @{const Pure.imp} $ p $ imp pms
+    in
+      imp prems
+    end
 in
   val typ_at_lifts_internal_goals = List.map abstract strong_thms
 end
@@ -3867,7 +3890,8 @@ private lemma ko_at_defn_ko_wp_at':
   by (clarsimp simp: ko_at'_defn_def obj_at'_real_def
                      ko_wp_at'_def project_inject)
 
-(* VER-1364: normalise_obj_at' sometimes doesn't normalise ob_at' unless you use it twice. *)
+(* FIXME: normalise_obj_at' sometimes doesn't normalise obj_at' unless you use it twice.
+          See VER-1364 for more details. *)
 private method normalise_obj_at'_step =
   (clarsimp?, elim obj_at_ko_at'[folded ko_at'_defn_def, elim_format],
    clarsimp simp: ko_at_defn_rewr ko_at_defn_pred_tcb_at' ko_at_defn_ko_wp_at',
--- a/proof/refine/ARM/Invariants_H.thy
+++ b/proof/refine/ARM/Invariants_H.thy
@@ -3867,11 +3867,15 @@ private lemma ko_at_defn_ko_wp_at':
   by (clarsimp simp: ko_at'_defn_def obj_at'_real_def
                      ko_wp_at'_def project_inject)
 
-method normalise_obj_at' =
+(* VER-1364: normalise_obj_at' sometimes doesn't normalise ob_at' unless you use it twice. *)
+private method normalise_obj_at'_step =
   (clarsimp?, elim obj_at_ko_at'[folded ko_at'_defn_def, elim_format],
    clarsimp simp: ko_at_defn_rewr ko_at_defn_pred_tcb_at' ko_at_defn_ko_wp_at',
    ((drule(1) ko_at_defn_uniqueD)+)?,
    clarsimp simp: ko_at'_defn_def)
+
+method normalise_obj_at' =
+  normalise_obj_at'_step, normalise_obj_at'_step?
 
 end
 
--- a/proof/refine/ARM/IpcCancel_R.thy
+++ b/proof/refine/ARM/IpcCancel_R.thy
@@ -156,18 +156,6 @@ lemma gts_wf''[wp]:
   apply (fastforce simp: valid_obj'_def valid_tcb'_def)
   done
 
-lemma replyUnlink_valid_objs':
-  "\<lbrace>valid_objs'\<rbrace>
-   replyUnlink rptr tptr
-   \<lbrace>\<lambda>_. valid_objs'\<rbrace>"
-  apply (clarsimp simp: replyUnlink_def setReplyTCB_def getReplyTCB_def liftM_def)
-  apply (wpsimp wp: set_reply_valid_objs' hoare_vcg_all_lift gts_wp'
-              simp: valid_tcb_state'_def)
-  apply (frule (1) ko_at_valid_objs'[where 'a=reply, simplified])
-   apply (clarsimp simp: projectKO_opt_reply split: kernel_object.splits)
-  apply (clarsimp simp: valid_obj'_def valid_reply'_def obj_at'_def)
-  done
-
 lemma setReplyTCB_corres:
   "corres dc (reply_at rp) (reply_at' rp)
             (set_reply_obj_ref reply_tcb_update rp new)
@@ -178,11 +166,6 @@ lemma setReplyTCB_corres:
       apply (rule set_reply_corres)
       apply (simp add: reply_relation_def)
   by (wpsimp simp: obj_at'_def replyPrev_same_def)+
-
-defs replyUnlink_assertion_def:
-  "replyUnlink_assertion
-    \<equiv> \<lambda>replyPtr state s. state = BlockedOnReply (Some replyPtr)
-                          \<or> (\<exists>ep d. state = BlockedOnReceive ep d (Some replyPtr))"
 
 lemma setReply_valid_tcb'[wp]:
   "setReply rp new  \<lbrace>valid_tcb' tcb\<rbrace>"
--- a/proof/refine/ARM/IpcCancel_R.thy
+++ b/proof/refine/ARM/IpcCancel_R.thy
@@ -881,21 +881,22 @@ lemma cancelSignal_st_tcb_at:
    apply clarsimp+
   done
 
+lemma blockedCancelIPC_st_tcb_at:
+  "\<lbrace>\<lambda>s. (t = t' \<longrightarrow> P Inactive) \<and> (t \<noteq> t' \<longrightarrow> st_tcb_at' P t s)\<rbrace>
+   blockedCancelIPC st t' rptr
+   \<lbrace>\<lambda>_. st_tcb_at' P t\<rbrace>"
+  unfolding blockedCancelIPC_def Let_def getBlockingObject_def
+  apply (wpsimp wp: setThreadState_st_tcb_at'_cases replyUnlink_st_tcb_at' hoare_vcg_imp_lift'
+                    getEndpoint_wp)
+  done
+
 lemma cancelIPC_st_tcb_at:
   assumes [simp]: "\<And>st. simple' st \<longrightarrow> P st" shows
-  "\<lbrace>st_tcb_at' P t\<rbrace>
-     cancelIPC t'
-   \<lbrace>\<lambda>rv. st_tcb_at' P t\<rbrace>"
-  apply (simp add: cancelIPC_def Let_def
-             cong: if_cong Structures_H.thread_state.case_cong)
-  sorry (*
-  apply (rule hoare_seq_ext [OF _ gts_sp'])
-  apply (case_tac x, simp_all add: isTS_defs list_case_If)
-         apply (wp sts_st_tcb_at'_cases delete_one_st_tcb_at
-                   threadSet_pred_tcb_no_state
-                   cancelSignal_st_tcb_at hoare_drop_imps
-                | clarsimp simp: o_def if_fun_split)+
-  done *)
+  "cancelIPC t' \<lbrace>st_tcb_at' P t\<rbrace>"
+  unfolding cancelIPC_def
+  apply (wpsimp wp: blockedCancelIPC_st_tcb_at replyRemoveTCB_st_tcb_at'_cases
+                    cancelSignal_st_tcb_at threadSet_pred_tcb_no_state gts_wp')
+  done
 
 lemma weak_sch_act_wf_lift_linear:
   "\<lbrakk> \<And>t. \<lbrace>\<lambda>s. sa s \<noteq> SwitchToThread t\<rbrace> f \<lbrace>\<lambda>rv s. sa s \<noteq> SwitchToThread t\<rbrace>;
@@ -905,12 +906,6 @@ lemma weak_sch_act_wf_lift_linear:
   apply (simp only: weak_sch_act_wf_def imp_conv_disj)
   apply (intro hoare_vcg_all_lift hoare_vcg_disj_lift hoare_vcg_conj_lift)
   apply simp_all
-  done
-
-lemma sts_sch_act_not[wp]:
-  "\<lbrace>sch_act_not t\<rbrace> setThreadState st t' \<lbrace>\<lambda>rv. sch_act_not t\<rbrace>"
-  unfolding setThreadState_def rescheduleRequired_def scheduleTCB_def
-  apply (wpsimp wp: hoare_vcg_if_lift2 hoare_drop_imp isSchedulable_inv)
   done
 
 crunches cancelSignal, setBoundNotification
--- a/proof/refine/ARM/IpcCancel_R.thy
+++ b/proof/refine/ARM/IpcCancel_R.thy
@@ -1005,8 +1005,9 @@ lemma (in delete_one_conc_pre) cancelIPC_tcb_at_runnable':
                   | wpc | simp add: o_def if_fun_split)+
   done *)
 
-crunch ksCurDomain[wp]: cancelSignal "\<lambda>s. P (ksCurDomain s)"
-  (wp: crunch_wps)
+crunches cancelIPC
+  for ksCurDomain[wp]: "\<lambda>s. P (ksCurDomain s)"
+  (wp: crunch_wps simp: crunch_simps pred_tcb_at'_def)
 
 (* FIXME move *)
 lemma setBoundNotification_not_ntfn:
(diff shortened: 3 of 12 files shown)


pierzchalski:
[Testboard](https://bamboo.ts.data61.csiro.au/browse/L4V-TESTBOARD-610)

proof/refine/ARM/Invariants_H.thy
```diff
@@ -356,8 +356,7 @@ definition live_ntfn' :: "notification \<Rightarrow> bool" where
                       \<or> (\<exists>ts. ntfnObj ntfn = WaitingNtfn ts)"
 
 definition live_reply' :: "reply \<Rightarrow> bool" where
```
michaelmcinerney, line 358: Can you tell me a little about why you'd like to change this definition?

proof/refine/ARM/IpcCancel_R.thy
```diff
@@ -1044,15 +1019,34 @@ lemma cancelSignal_tcb_obj_at':
   apply (wpsimp wp: setThreadState_not_st getNotification_wp)
   done
 
-lemma (in delete_one_conc_pre) cancelIPC_tcbDomain_obj_at':
+crunches replyRemoveTCB, cancelSignal, getBlockingObject
+  for obj_at'_only_st_qd_ft: "\<lambda>s. P (obj_at' (Q :: tcb \<Rightarrow> bool) t s)"
```
michaelmcinerney, line 1023: Considering the name you've given to the following lemma, is his wha you'd like to call just this predicate?

proof/refine/ARM/Reply_R.thy
```diff
@@ -96,4 +95,487 @@ lemma replyRemoveTCB_st_tcb_at'_sym_ref:
...
+  unfolding replyUnlink_def setReplyTCB_def getReplyTCB_def replyUnlink_assertion_def
+            updateReply_def
+  apply (wpsimp wp: sts'_valid_pspace'_inv hoare_vcg_imp_lift'
+              simp: valid_tcb_state'_def valid_pspace'_def)
+  apply normalise_obj_at'
+  apply (frule(1) reply_ko_at_valid_objs_valid_reply')
+  apply (clarsimp simp: valid_reply'_def)
+  done
+
+lemma replyUnlink_not_runnable_valid_queues:
```
michaelmcinerney, line 143: This should be unnecessary seeing that we have `replyUnlink_valid_queues` below

proof/refine/ARM/Reply_R.thy
```diff
@@ -96,4 +95,487 @@ lemma replyRemoveTCB_st_tcb_at'_sym_ref:
...
+  done
+
+lemma updateReply_valid_objs'_preserved:
+  "\<lbrace>valid_objs' and K (\<forall>r s. valid_reply' r s \<longrightarrow> valid_reply' (upd r) s)\<rbrace>
+   updateReply rptr upd
+   \<lbrace>\<lambda>_. valid_objs'\<rbrace>"
+  apply (wpsimp wp: updateReply_valid_objs'_preserved_strong)
+  done
+
+lemma replyUnlink_valid_objs':
```
michaelmcinerney, line 276: I know that I keep adding this one to `wpsimp`s. Did we wanna put it in the `wp` set?

proof/refine/ARM/Reply_R.thy
```diff
@@ -96,4 +95,487 @@ lemma replyRemoveTCB_st_tcb_at'_sym_ref:
...
+  apply (prop_tac "replyNexts_of s rptr = replyNext_of reply
+                   \<and> replyPrevs_of s rptr = replyPrev reply ")
+   apply (clarsimp simp: opt_map_def reply_of'_KOReply)
+  apply (case_tac "replyNext_of reply"
+         ; case_tac "replyPrev reply"
+         ; clarsimp simp: sym_refs_replyNext_None sym_refs_replyPrev_None
+                          sym_refs_replyNext_replyPrev_proj)
+  done
+
+\<comment>\<open> Used to "hide" @{term "sym_refs o list_refs_of_replies'"} from simplification. \<close>
```
michaelmcinerney, line 482: Are there not more standard ways of doing the same sort of thing? `thin_tac` maybe? I guess if the use of this is limited it's not so bad

proof/refine/ARM/Reply_R.thy
```diff
@@ -96,4 +95,487 @@ lemma replyRemoveTCB_st_tcb_at'_sym_ref:
...
+  apply (wpsimp wp: cleanReply_valid_pspace' updateReply_valid_pspace' hoare_vcg_if_lift
+                    hoare_vcg_imp_lift' gts_wp' setObject_sc_idle' haskell_assert_wp
+         split_del: if_split)
+  apply (clarsimp simp: valid_reply'_def pred_tcb_at'_def)
+  apply (intro conjI impI allI
+         ; simp?
+         ; normalise_obj_at'?)
+  \<comment>\<open> Weird indentation: around 20 subgoals, almost all of which are trivial/boring facts or
+      contradictions about being a valid SC. \<close>
+  apply (all \<open>(solves \<open>clarsimp simp: valid_idle'_def idle_tcb'_def obj_at'_def isReply_def\<close>)?\<close>)
```
michaelmcinerney, line 538: I know it'd make the previous block a bit more unwieldy, but did you wanna put this `all` bit into it too? Does it actually solve the 20 subgoal problem?

proof/refine/ARM/TcbAcc_R.thy
```diff
@@ -5469,5 +5506,15 @@ lemma tcbSchedEnqueue_obj_at':
   apply simp
   done
 
+lemma setThreadState_not_runnable_valid_queues:
```
michaelmcinerney, line 5509: You can delete this and just use `setThreadState_valid_queues` which I updated with the remove runnable' stuff

michaelmcinerney (approved):
1% of `cancelIPC_invs'`? Lord have mercy

proof/refine/ARM/Reply_R.thy
```diff
@@ -96,4 +95,487 @@ lemma replyRemoveTCB_st_tcb_at'_sym_ref:
   apply (clarsimp simp: obj_at'_def pred_tcb_at'_def)
   done
 
+lemma getReplyTCB_wp:
+  "\<lbrace>\<lambda>s. reply_at' rptr s \<longrightarrow> obj_at' (\<lambda>reply. P (replyTCB reply) s) rptr s\<rbrace>
+   getReplyTCB rptr
+   \<lbrace>P\<rbrace>"
```
corlewis, line 101: This slightly clashes with the version of this lemma I added in my PR. See https://github.com/seL4/l4v/pull/93/files#diff-8f12378cc553b81307fa6f31bbc1b337R2395

proof/refine/ARM/Reply_R.thy
```diff
@@ -96,4 +95,487 @@ lemma replyRemoveTCB_st_tcb_at'_sym_ref:
...
+  done
+
+lemma setReply_valid_pde_mappings'[wp]:
+  "setReply p r \<lbrace>valid_pde_mappings'\<rbrace>"
+  unfolding valid_pde_mappings'_def setReply_def
+  apply (wpsimp wp: hoare_vcg_all_lift)
+  done
+
+lemma replyUnlink_valid_pspace'[wp]:
+  "replyUnlink rptr tptr \<lbrace>valid_pspace'\<rbrace>"
```
corlewis, line 133: Uh oh, I think there might be some duplicated work here. Victor proved a lot of these lemmas about `replyUnlink` in IpcCancel_R. We need to make sure that they're all consistent before continuing.

proof/refine/ARM/Reply_R.thy
```diff
@@ -96,4 +95,487 @@ lemma replyRemoveTCB_st_tcb_at'_sym_ref:
...
+  "replyRemoveTCB tptr \<lbrace>\<lambda>s. sym_refs (list_refs_of_replies' s)\<rbrace>"
+  unfolding replyRemoveTCB_def decompose_list_refs_of_replies'
+  supply if_cong[cong]
+  apply (wpsimp wp: cleanReply_list_refs_of_replies' hoare_vcg_if_lift hoare_vcg_imp_lift' gts_wp'
+                    haskell_assert_wp
+              simp: pred_tcb_at'_def
+         split_del: if_split)
+  unfolding decompose_list_refs_of_replies'[symmetric] protected_sym_refs_def[symmetric]
+  \<comment>\<open> opt_mapE will sometimes destroy the @{term "(|>)"} inside @{term replyNexts_of}
+      and @{term replyPrevs_of}, but we're using those as our local normal form. \<close>
```
corlewis, line 496: Is this something we should have a think about more generally? Presumably this is going to come back whenever we deal with `list_refs_of_replies'`.

proof/refine/ARM/Reply_R.thy
```diff
@@ -96,4 +95,487 @@ lemma replyRemoveTCB_st_tcb_at'_sym_ref:
...
+                   , normalise_obj_at'
+                   , (drule obj_at'_to_ks_obj_at', clarsimp)+\<close>)
+       apply (all \<open>drule(1) ks_reply_at'_repliesD[folded protected_sym_refs_def]
+                   , clarsimp simp: projectKO_reply isHead_to_head\<close>)
+       \<comment>\<open> Now, for each case we can blow open @{term sym_refs}, which will give us enough new
+           @{term "(replyNexts_of, replyPrevs_of)"} facts that we can throw it all at metis. \<close>
+       apply (clarsimp simp: sym_refs_def split_paired_Ball in_get_refs
+              , intro conjI impI allI
+              ; metis sym_refs_replyNext_replyPrev_proj[folded protected_sym_refs_def] option.sel
+                      option.inject)+
```
corlewis, line 512: Fancy!

proof/refine/ARM/KHeap_R.thy
```diff
@@ -24,6 +24,37 @@ abbreviation (input)
 where
   "set_obj' ptr obj s \<equiv> set_ko' ptr (injectKO obj) s"
 
+(* FIXME RT: bikeshedding *)
```
corlewis, line 27: To start off the bikeshedding, can you say a bit about the use case where you found this helped? Was it mainly that you didn't want to deal with `is_aligned` and `ps_clear`?

corlewis (approved):
Looks good!

proof/refine/ARM/Reply_R.thy
michaelmcinerney: I think Miki was wanting to use a slightly different predicate here instead of `sym_refs` for the `list_refs_of_replies'`. It's in her large draft PR for `reply_remove_tcb_corres`

michaelmcinerney: Which is to say, this should then become unnecessary

proof/refine/ARM/Reply_R.thy
pierzchalski: `thin_tac` doesn't have a way to only temporarily hide assumptions: see [VER-1244](https://sel4.atlassian.net/browse/VER-1244), which is exactly about this. Sadly, dodgy temporary definitions seem to be the easiest way to do this sort of thing.

proof/refine/ARM/Reply_R.thy
pierzchalski: It solves all but 4, hence the indentation afterwards. Not sure which is preferable?

proof/refine/ARM/IpcCancel_R.thy
pierzchalski: I'm not sure what you're asking here?

proof/refine/ARM/KHeap_R.thy
pierzchalski: This started with @mktnk3 pointing out that in lots of places, `ksPSpace s p = Some blah` is a weaker assumption _and_ weaker conclusion for various properties (I believe `sym_refs`, in my case). Not sure the use-case really affects the bikeshedding, though.

proof/refine/ARM/Reply_R.thy
pierzchalski: Guh, good catch.

proof/refine/ARM/Bits_R.thy
```diff
@@ -97,6 +97,11 @@ lemma projectKO_reply:
   "(projectKO_opt ko = Some t) = (ko = KOReply t)"
   by (cases ko) (auto simp: projectKO_opts_defs)
 
+lemma reply_of'_KOReply:
```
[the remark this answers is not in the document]
pierzchalski: I'm generally wary of adding things to `[simp]`. Do you think it's worth it here?

proof/refine/ARM/IpcCancel_R.thy
michaelmcinerney: You have `cancelIPC_obj_at'_only_st_qd_ft` just below this. That is for a predicate that doesn't care about state, queued, or fault updates. But here, you're crunching, and calling something `obj_at'_only_st_qd_ft`, but this predicate here is more general

proof/refine/ARM/Reply_R.thy
michaelmcinerney: Get workin' on that JIRA issue! Sounds good

proof/refine/ARM/Reply_R.thy
michaelmcinerney: If it's a choice between another ; and the `all < blah >`, then I'd go with another ;

proof/refine/ARM/Reply_R.thy
pierzchalski: Hmm, [this one?](https://github.com/seL4/l4v/pull/92/files) I don't see anything about "removing" `sym_refs`. @mktnk3 got some forward guidance?

proof/refine/ARM/Reply_R.thy
pierzchalski: Yeah, I've removed a lot of now-redundant stuff in `IpcCancel_R`. Wish I'd caught it earlier though.

proof/refine/ARM/Reply_R.thy
michaelmcinerney: I thought that Miki might have changed the `sym_refs (list_refs_of_resplies' s)` to be phrased in terms of 'sym_heap` (which is at line 784 of IpcCancel_R in that PR), but it seems that she hasn't. Might be worth asking her about it

proof/refine/ARM/Reply_R.thy
pierzchalski: Updated to use yours.

proof/refine/ARM/Reply_R.thy
corlewis: I try to at last skim most PRs that come in, even if I don't have anything to contribute to a review. It helps avoid these sorts of situations

proof/refine/ARM/Bits_R.thy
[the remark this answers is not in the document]
corlewis: I'm not 100% sure, but I haven't been able to think of a case where you wouldn't want to use this simplification.  Does anyone else have an opinion on it? For people not already on this PR, @mbrcknl @lsf37 @Xaphiosis

proof/refine/ARM/IpcCancel_R.thy
pierzchalski: Right; this is making use of a `crunches` feature where it goes looking for lemmas named `${constant_name}_${property_name}`, letting those lemmas add side-conditions (I think I have a comment about this somewhere else in the PR). In short, this line doesn't produce a more general lemma, but that fact isn't visible/obvious, which is unfortunate.

proof/refine/ARM/Reply_R.thy
mktnk3: > I thought that Miki might have changed the `sym_refs (list_refs_of_resplies' s)` to be phrased in terms of 'sym_heap` (which is at line 784 of IpcCancel_R in that PR), but it seems that she hasn't. Might be worth asking her about it

Not quite, Michael. We need `sym_refs (list_refs_of_resplies' s)` just as before. What I have there is some abstraction for the case when we have symmetric heaps, for which our main examples are `replyPrevs_of` and `replyNexts_of`, so that we don’t need to write lemmas in pairs for both directions.

proof/refine/ARM/IpcCancel_R.thy
corlewis: To be pedantic, `crunch` adds side-conditions from any lemmas that it finds and reuses, not just those with the appropriate name. It can also find lemmas that it has already proven in the current command or from the `wp` set.

michaelmcinerney: So is the odd naming here actually beneficial?

pierzchalski:
Alright, outstanding issues:
* There are a few lemmas I don't know where to put:
  * `valid_mdb'_lift`
  * `objBits_sc_only_depends_on_scRefills`
* There are some lemmas I'm using that haven't made it in yet, all from @mktnk3's PR.
* Bikeshedding the `ksPSpace`-only `obj_at'` family.
* Some failing tets.

proof/refine/ARM/KHeap_R.thy
mktnk3: Generally speaking, I am quite hesitant of introducing all these. I think this is something that is not supposed to be introduced in bits like this. If we are to take that approach, maybe doing it more fully and generally would be better. Otherwise we run the risk of diverging. Also, I am rather reluctant in introducing abbreviations for such basic one-liners. To me, these will potentially hide information, and that could potentially mean extra load on everyone’s mind. But this is also just my preference as well. @lsf37 can I ask for your option on this?

mktnk3: And I also should say that, at this point, I myself don’t have a clear picture of where my comments you are referring to was heading to, if it ever does.

proof/refine/ARM/Bits_R.thy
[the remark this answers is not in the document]
MitchellBuckley: Agreed on both counts. If we can safely put this in [simp] then we should. Pinging @lsf37 @mbrcknl @Xaphiosis again.

proof/refine/ARM/Invariants_H.thy
```diff
@@ -2655,11 +2654,25 @@ lemma typ_at_lift_page_table_at'_strong:
...
+      and reply: "\<And>p. f \<lbrace>\<lambda>s. P (typ_at' ReplyT p s)\<rbrace>"
+      and ntfn: "\<And>p. f \<lbrace>\<lambda>s. P (typ_at' NotificationT p s)\<rbrace>"
+  shows "f \<lbrace>\<lambda>s. P (valid_tcb_state' st s)\<rbrace>"
+  unfolding valid_tcb_state'_def valid_bound_reply'_def
+  apply (case_tac st
+         ; clarsimp split: option.splits
+         , wpsimp wp: hoare_vcg_imp_lift' hoare_vcg_all_lift hoare_vcg_conj_lift_N[where N=P]
+                           typ_at_lift_ep'_strong[OF ep] typ_at_lift_reply'_strong[OF reply]
+                           typ_at_lift_ntfn'_strong[OF ntfn])
+  done
```
MitchellBuckley, line 2668: Woah, nice lemma!

proof/refine/ARM/Reply_R.thy
```diff
@@ -96,4 +95,469 @@ lemma replyRemoveTCB_st_tcb_at'_sym_ref:
   apply (clarsimp simp: obj_at'_def pred_tcb_at'_def)
   done
 
+crunches setReplyTCB
+  for ksReadyQueues[wp]: "\<lambda>s. P (ksReadyQueues s)"
+  and ksSchedulerAction[wp]: "\<lambda>s. P (ksSchedulerAction s)"
+  and valid_queues[wp]: "valid_queues"
+
+crunches getReplyTCB
```
MitchellBuckley, line 103: Add "FIXME RT: move"

MitchellBuckley (commented):
This is huge and looks excellent.
The addition of new definitions needs some carefuly consideration when they are connected to very fundamental concepts such as `obj_at'`. But it seems like the 'bike-shedding' is alreaduy occuring, so that's good. Though I think in this case it is not really bike-shedding because that kind of infrastructure is not unimportant.

proof/refine/ARM/Reply_R.thy
pierzchalski: Anywhere it "should" go?

MitchellBuckley: No idea. If there is not a general place for it, then at least put it at the top of the file?

proof/refine/ARM/Bits_R.thy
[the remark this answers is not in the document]
pierzchalski: Went ahead and did the obvious thing of... just seeing what happens. Making it `[simp]` doesn't break anything, so sure, I'll change it.

spec/haskell/src/SEL4/Object/Reply.lhs
```diff
@@ -176,9 +185,8 @@ on the thread state of the replyTCB of the replyPtr
 > replyUnlink_assertion _ _ _ = True
 
 > cleanReply :: PPtr Reply -> Kernel ()
-> cleanReply replyPtr = do
->     reply <- getReply replyPtr
->     setReply replyPtr (reply { replyPrev = Nothing, replyNext = Nothing })
+> cleanReply replyPtr = updateReply replyPtr (\reply -> reply { replyPrev = Nothing,
```
mktnk3, line 188: I am ok with having `updateReply` at the Haskell level, but I would like to have `cleanReply` to be defined as two separate lines of `updateReply`, one for the prev update and another for the next update. They are quite different semantically and a bit of pain to handle when bundled together...

pierzchalski: Sounds good!

proof/refine/ARM/KHeap_R.thy
mktnk3: Pinging @lsf37 again on this. I think I have been using projections instead in most of the situations when I want something like this. For example, `ks_reply_at’ r p s` here is equivalent to `replies_of’ s p = Some r`. So, I am not convinced that we should be making a step in the direction suggested here at this point.

pierzchalski: Alright, I think I'll rip this out for now, maybe switch to projections when appropriate.

pierzchalski: Ha, wow, this was almost entirely unused. Oh well.

pierzchalski:
[testboard](https://bamboo.ts.data61.csiro.au/browse/L4V-TESTBOARD-616)

proof/refine/ARM/IpcCancel_R.thy
```diff
@@ -756,136 +717,178 @@ lemma setEndpoint_pde_mappings'[wp]:
...
+  and valid_release_queue'[wp]: valid_release_queue'
+  and untyped_ranges_zero'[wp]: "untyped_ranges_zero'"
+  (wp: crunch_wps)
 
-lemma replyUnlink_sch_act_simple[wp]:
-  "replyUnlink r t' \<lbrace>sch_act_simple\<rbrace>"
-  apply (simp add: replyUnlink_def setReplyTCB_def getReplyTCB_def)
-  by (wpsimp wp: hoare_drop_imps)
+(* FIXME RT: move to... ?*)
+lemma add_asserts:
```
mktnk3, line 829: I think this already exists: `hoare_strengthen_pre_via_assert_backward`

proof/refine/ARM/IpcCancel_R.thy
```diff
@@ -756,136 +717,178 @@ lemma setEndpoint_pde_mappings'[wp]:
...
+  assumes not: "\<lbrace>Not o P\<rbrace> f \<lbrace>\<lambda>_. \<bottom>\<rbrace>"
+  shows "\<lbrace>\<lambda>s. P s \<and> Q s\<rbrace> f \<lbrace>R\<rbrace> \<Longrightarrow> \<lbrace>Q\<rbrace> f \<lbrace>R\<rbrace>"
+  apply (insert use_valid[OF _ not])
+  apply (fastforce simp: valid_def)
+  done
+
+lemma blockedCancelIPC_asserts:
```
mktnk3, line 836: We probably don’t need to have this as a separate lemma. Is there any particular reason you cannot include this in the main proof below?

proof/refine/ARM/TcbAcc_R.thy
```diff
@@ -3039,6 +3058,7 @@ lemma sts_sch_act':
   apply (wpsimp wp: threadSet_pred_tcb_at_state threadSet_sch_act_wf hoare_vcg_disj_lift)
   done
 
+(* FIXME: sts_sch_act' (above) is stronger, and should be the wp rule. VER-1366 *)
```
mktnk3, line 3061: Is this MCS specific? If so, can you label it with FIXME RT?

proof/refine/ARM/Reply_R.thy
```diff
@@ -95,4 +109,462 @@ lemma replyRemoveTCB_st_tcb_at'_sym_ref:
...
+  "\<lbrakk>replies_of' s rptr = Some reply; sym_refs (list_refs_of_replies' s)\<rbrakk>
+   \<Longrightarrow> replyNexts_of s rptr = replyNext_of reply
+       \<and> (case replyNext_of reply of
+            Some next \<Rightarrow> replyPrevs_of s next = Some rptr
+          | None \<Rightarrow> \<forall>rptr'. replyPrevs_of s rptr' \<noteq> Some rptr)
+       \<and> replyPrevs_of s rptr = replyPrev reply
+       \<and> (case replyPrev reply of
+            Some prev \<Rightarrow> replyNexts_of s prev = Some rptr
+          | None \<Rightarrow> \<forall>rptr'. replyNexts_of s rptr' \<noteq> Some rptr)"
+  \<comment>\<open> opt_mapE will sometimes destroy the @{term "(|>)"} inside @{term replyNexts_of}
```
mktnk3, line 459: I don’t quite understand this comment... are you sure this `supply` line is doing anything?

proof/refine/ARM/KHeap_R.thy
```diff
@@ -365,6 +365,14 @@ lemma non_sc_same_typ_at'_objBits_always_the_same:
 lemmas non_sc_same_typ_at'_ko_wp_at'_set_ko'_iff =
   same_size_ko_wp_at'_set_ko'_iff[OF non_sc_same_typ_at'_objBits_always_the_same]
 
+lemma obj_at'_strip:
```
mktnk3, line 368: Is this used in this PR?

proof/refine/ARM/Invariants_H.thy
```diff
@@ -3868,12 +3890,16 @@ private lemma ko_at_defn_ko_wp_at':
   by (clarsimp simp: ko_at'_defn_def obj_at'_real_def
                      ko_wp_at'_def project_inject)
 
-method normalise_obj_at' =
+(* VER-1364: normalise_obj_at' sometimes doesn't normalise ob_at' unless you use it twice. *)
```
mktnk3, line 3893: If this is a complete issue, we usually don’t leave such comments in the theory file. You can record that it’s completed in the JIRA ticket itself, and you can certainly comment on the PR instead.

pierzchalski: Do you mean "completed" issue? This is a hacky temporary fix - the real fix should be to figure out why `normalise_obj_at'_step` doesn't perform all the required work in one go, hence the outstanding comment.

mktnk3, line 3893: I probably don’t see the point of this comment. Can you be a bit more clear? How does the JIRA issue relate to what is written here?

pierzchalski: The point is so when someone comes along later and sees this weird repeated method, they know _why_ it's a weird repeated method: for some reason it doesn't always do what we want it to do, and it's a known issue, and if they want more details they can look up the issue.

mktnk3: If there is something that still needs to be fixed, could you please label it with “FIXME RT”? If the JIRA issue has more details of the issue, you can also say “see VER-XXX for details” or something like that?

proof/refine/ARM/KHeap_R.thy
pierzchalski: Turns out no :)

proof/refine/ARM/TcbAcc_R.thy
pierzchalski: I don't think this is MCS-specific - this is just a weak rule incorrectly declared `[wp]`.

proof/refine/ARM/Reply_R.thy
pierzchalski: Just reiterating our conversation on mattermost: `opt_mapE` is declared `[elim!]` somewhere, and that declaration messes up the proof state here. Turns out we remove it from `[elim!]` on RISCV64; [VER-1367](https://sel4.atlassian.net/browse/VER-1367) asks if we should just remove it from `[elim!]` everywhere.

Separately, you mentioned you got a copy of this proof to go through without the `supply` line - I'll see if that works for me too.

pierzchalski:
If [this testboard](https://bamboo.ts.data61.csiro.au/browse/L4V-TESTBOARD-ARMVERIFY-628) goes through, I think this is ready to merge.

spec/haskell/src/SEL4/Object/Reply.lhs
```diff
@@ -188,6 +195,12 @@ on the thread state of the replyTCB of the replyPtr
 > getReplyTCB r = liftM replyTCB (getReply r)
 
 > setReplyTCB :: Maybe (PPtr TCB) -> PPtr Reply -> Kernel ()
-> setReplyTCB tptrOpt rptr = do
->     r <- getReply rptr
->     setReply rptr (r { replyTCB = tptrOpt})
+> setReplyTCB tptrOpt rptr = updateReply rptr (\reply -> reply { replyTCB = tptrOpt })
+
+> replyClear :: PPtr Reply -> PPtr TCB -> Kernel ()
```
mktnk3, line 200: Hmm. I think this is a duplicated old version. Maybe something went wrong when you rebased. I think Corey moved this to `ObjectType.lhs` and it should have `cancelIPC` instead of `replyUnlink`.

proof/refine/ARM/KHeap_R.thy
mbrcknl: Meta comment: When selecting code to make a comment on a pull request, it would be good to use a multi-line selection for the comment, selecting as many lines as might be relevant to the discussion that could follow from the initial comment.
In the conversation UI, GitHub never shows context *below* the lines that were selected, despite a [7-year-old feature request](https://github.com/isaacs/github/issues/97). Also, GitHub doesn't give a way to get more context in the conversation UI, so one has to go digging in the diff for the file to find the context. And if the branch has been force-pushed, that context can be extremely difficult to find, if it can be found at all.
Explicitly selecting a more generous amount of context would therefore help people like me who are late to the party and want to try to catch up on all this great PR communication. :-)

mktnk3: > Meta comment: When selecting code to make a comment on a pull request, it would be good to use a multi-line selection for the comment, selecting as many lines as might be relevant to the discussion that could follow from the initial comment.

Right. that makes sense. I didn’t know that that was possible, but it should be, I suppose!

spec/haskell/src/SEL4/Object/Reply.lhs
corlewis: Good spotting! As far as I can tell we should just be able to delete this duplicated version. It's not exported so I don't think it's accidentally used anywhere.

pierzchalski: Alright, as expected, removing this didn't break anything On My Machine™️ . Gonna start a testboard anyway, just in case.

pierzchalski:
hopefully [final testboard](https://bamboo.ts.data61.csiro.au/browse/L4V-TESTBOARD-629), confirming that removing a redundant `replyClear` function didn't mess up the proofs.

spec/haskell/src/SEL4/Object/Reply.lhs
```diff
@@ -95,6 +101,7 @@ This module specifies the behavior of reply objects.
 >         setSchedContext scPtr (sc { scReply = prevReplyPtrOpt })
 >         when (prevReplyPtrOpt /= Nothing) $ do
 >             prevReplyPtr <- return $ fromJust prevReplyPtrOpt
+>             assert (replyPtr /= prevReplyPtr) "replyPop: reply lists must be distinct"
```
mktnk3, line 104: I should have caught this before this PR was merged, but this line of assert, which is also added twice in `replyRemoveTCB`, does not seem to be needed at all. Why are these lines added to start with?

As I have been saying many weeks now, I am working on `reply_remove_tcb_corres`. So I’ve wasted some time on this while rebasing. Generally speaking, we probably want to make sure that all the spec updates are absolutely necessary. We really can’t afford to this kind of things, now for MCS in particular, but anytime, actually.

PR #96, opened 2020-09-16
Bjorn refill logic

michaelmcinerney:


michaelmcinerney:
There are a couple things I'm waiting to hear from Curtis about, and I need to fix up the `MIN_BUDGET_merge` function, but hopefully everything else is OK

michaelmcinerney:
I *think* the definition of `MIN_BUDGET_merge` is now OK, but not 100% sure...

spec/abstract/SchedContext_A.thy
```diff
@@ -195,72 +190,76 @@ where
...
+  MIN_BUDGET_merge :: "refill list \<Rightarrow> refill list"
+where
+  "MIN_BUDGET_merge refills
+    = (if r_amount (hd refills) < MIN_BUDGET
+       then let new_hd = \<lparr>r_time = r_time (hd (tl refills)) - r_amount (hd refills),
+                          r_amount = r_amount (hd refills) + r_amount (hd (tl refills))\<rparr>
+            in MIN_BUDGET_merge (new_hd # (tl (tl refills)))
+       else refills)"
+by pat_completeness auto
+
```
michaelmcinerney, line 203: This function is strictly partial, but every refill list (ignoring word addition overflow issues) which has nonzero sum is an element of the domain. Since `MIN_BUDGET` is nonzero, and we are interested only in refill lists where the (`unat`) sum of the refill amounts is a least `MIN_BUDGET`, this is fine. My current thought is to prove a lemma saying that all refill lists where the `unat` sum of the r_amounts is at least `MIN_BUDGET` are in the domain. Then proving lemmas about this function (for instance, that this function will actually return a list with at least `MIN_BUDGET` in the head) is "straightforward"

xurtis:
You're going to have to excuse the time it takes me to review this, I'm still very new to reading Isabelle.

spec/abstract/SchedContext_A.thy
```diff
@@ -195,72 +190,78 @@ where
...
-                   set_refills sc_ptr (schedule_used full (new_head # (tl refills)) used)
-                 od
-              od
+    usage' \<leftarrow> return $ min usage (r_amount (hd refills));
+
+    if usage' > 0
+    then do used \<leftarrow> return \<lparr>r_time = r_time (hd refills) + period, r_amount = usage'\<rparr>;
+            full \<leftarrow> refill_full sc_ptr;
+            refills' \<leftarrow> return $ schedule_used full refills used;
+            usage'' \<leftarrow> return $ min usage' (r_amount (hd refills'));
```
xurtis, line 225: Why does this need to be recalculated here?

xurtis:
This all looks fine to me. Are there any particular questions you had?

spec/abstract/SchedContext_A.thy
michaelmcinerney: Actually, I think this line should be removed and I should be using `usage'` here. Thanks!

michaelmcinerney:
> This all looks fine to me. Are there any particular questions you had?

Yes, I left a few comments on your PR on Bitbucket. I think one of them I have figured out, and the other is minor, but there was a larger question about which invariants we would like to have on the refills

spec/abstract/SchedContext_A.thy
michaelmcinerney: Wait, I was just using `usage''` as a way to calculate the `min` that's in the C after the `schedule_used`. I could lump this all into the following line if it makes it easier to read or understand

michaelmcinerney: I think I'd at least rename this from `usage''` to `min`, or something, to be a little bit less confusing

xurtis: That looks like an error on my part, only one MIN should be there

xurtis: The second one should be redundant. I'll remove it in the C.

PR #98, opened 2020-09-17
Weaken tcbSchedDequeue preconditions

ryybrr:
Some lemmas about `tcbSchedDequeue` had preconditions involving `tcb_at'` that were too strong due to the use of `threadGet_const_tcb_at` instead of `threadGet_wp`.

### DIFF
--- a/proof/refine/ARM/Schedule_R.thy
+++ b/proof/refine/ARM/Schedule_R.thy
@@ -232,7 +232,7 @@ lemma tcbSchedDequeue_valid_queues_weak:
   "\<lbrace> valid_queues_no_bitmap_except t and valid_bitmapQ and
      bitmapQ_no_L2_orphans and bitmapQ_no_L1_orphans and
      correct_queue t and
-     obj_at' (\<lambda>tcb. tcbDomain tcb \<le> maxDomain \<and> tcbPriority tcb \<le> maxPriority) t \<rbrace>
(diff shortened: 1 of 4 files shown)


proof/refine/ARM/Schedule_R.thy
```diff
@@ -319,13 +318,13 @@ lemma setQueue_ksReadyQueues_lift:
   by (wp, clarsimp simp: fun_upd_def snd_def)
 
 lemma tcbSchedDequeue_valid_queues'[wp]:
-  "\<lbrace>valid_queues' and tcb_at' t\<rbrace>
+  "\<lbrace>valid_queues'\<rbrace>
     tcbSchedDequeue t \<lbrace>\<lambda>_. valid_queues'\<rbrace>"
```
corlewis, line 322: Very low priority, but I'm a fan of fixing up the indentation and layout of lemmas when I'm already touching them. Quite a few of the lemmas in this PR were previously wrong.

corlewis (approved):
Nice!

mktnk3 (approved):
Very nice! This can be merged next when you are ready, I think.

Is it correct that this PR doesn’t have the lemma that is sorried in Corey’s current PR? Does that mean there will be another one coming soon?

mktnk3 (approved):
Looks all good to me! Please go ahead and merge when you can. You should be able to click the “rebase and merge” option.

PR #101, opened 2020-09-18
activate_thread: proofs for corres lemma and invs' lemma.

MitchellBuckley:
This solves `activateThread_invs'` and `activateThread_corres` but requires a number of small spec-changes:
* haskell for activateThread was slightly out-of-date
* Update ipc buffer handling: the code now does a lookup before invocation decoding and is passed all the way down.
* Fix ordering of commands for inlining of cancel_yield_to into complete_yield_to.

### DIFF
--- a/proof/refine/ARM/Tcb_R.thy
+++ b/proof/refine/ARM/Tcb_R.thy
@@ -24,59 +24,6 @@ lemma activate_idle_thread_corres:
     (arch_activate_idle_thread t) (activateIdleThread t)"
   by (simp add: arch_activate_idle_thread_def activateIdleThread_def)
 
-lemma activate_corres:
- "corres dc (invs and ct_in_state activatable) (invs' and ct_in_state' activatable')
-            activate_thread activateThread"
-  apply (simp add: activate_thread_def activateThread_def)
-  apply (rule corres_guard_imp)
-    apply (rule corres_split_eqr [OF _ gct_corres])
-  sorry (*
-      apply (rule_tac R="\<lambda>ts s. valid_tcb_state ts s \<and> (idle ts \<or> runnable ts)
-                                \<and> invs s \<and> st_tcb_at ((=) ts) thread s"
-                  and R'="\<lambda>ts s. valid_tcb_state' ts s \<and> (idle' ts \<or> runnable' ts)
-                                \<and> invs' s \<and> st_tcb_at' (\<lambda>ts'. ts' = ts) thread s"
-                  in  corres_split [OF _ gts_corres])
-        apply (rule_tac F="idle rv \<or> runnable rv" in corres_req, simp)
-        apply (rule_tac F="idle' rv' \<or> runnable' rv'" in corres_req, simp)
-        apply (case_tac rv, simp_all add:
-                  isRunning_def isRestart_def,
-                  safe, simp_all)[1]
-         apply (rule corres_guard_imp)
-           apply (rule corres_split_eqr [OF _ getRestartPCs_corres])
-             apply (rule corres_split_nor [OF _ setNextPCs_corres])
-               apply (rule sts_corres)
-               apply (simp | wp weak_sch_act_wf_lift_linear)+
-          apply (clarsimp simp: st_tcb_at_tcb_at)
-         apply fastforce
-        apply (rule corres_guard_imp)
-          apply (rule activate_idle_thread_corres)
-         apply (clarsimp elim!: st_tcb_weakenE)
-        apply (clarsimp elim!: pred_tcb'_weakenE)
-       apply (wp gts_st_tcb gts_st_tcb' gts_st_tcb_at)+
-   apply (clarsimp simp: ct_in_state_def tcb_at_invs
-                  elim!: st_tcb_weakenE)
-  apply (clarsimp simp: tcb_at_invs' ct_in_state'_def
-                 elim!: pred_tcb'_weakenE)
-  done *)
-
-
-lemma bind_notification_corres:
-  "corres dc
-         (invs and tcb_at t and ntfn_at a) (invs' and tcb_at' t and ntfn_at' a)
-         (bind_notification t a) (bindNotification t a)"
-  apply (simp add: bind_notification_def bindNotification_def update_sk_obj_ref_def bind_assoc)
-  apply (rule corres_guard_imp)
-    apply (rule corres_split[OF _ get_ntfn_corres])
-      apply (rule corres_split[OF _ set_ntfn_corres])
-         apply (rule sbn_corres)
-        apply (clarsimp simp: ntfn_relation_def split: Structures_A.ntfn.splits)
-       apply wp+
-  by auto
-
-
-abbreviation
-  "ct_idle' \<equiv> ct_in_state' idle'"
-
 lemma gts_st_tcb':
   "\<lbrace>tcb_at' t\<rbrace> getThreadState t \<lbrace>\<lambda>rv. st_tcb_at' (\<lambda>st. st = rv) t\<rbrace>"
   apply (rule hoare_vcg_precond_imp)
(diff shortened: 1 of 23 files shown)


spec/abstract/Decode_A.thy
```diff
@@ -486,13 +486,15 @@ section "Scheduling Contexts"
 text \<open>The following definitions decode system calls related to scheduling contexts
 and scheduling control.\<close>
 
+find_consts "_ list \<Rightarrow> _ option"
```
MitchellBuckley, line 489: delete these

spec/abstract/SchedContext_A.thy
```diff
@@ -309,34 +309,40 @@ definition
...
-  set_consumed :: "obj_ref \<Rightarrow> data list \<Rightarrow> (unit, 'z::state_ext) s_monad"
+  set_consumed :: "obj_ref \<Rightarrow> data option \<Rightarrow> (unit, 'z::state_ext) s_monad"
 where
-  "set_consumed sc_ptr args \<equiv>do
+  "set_consumed sc_ptr buf \<equiv> do
       consumed \<leftarrow> sched_context_update_consumed sc_ptr;
       ct \<leftarrow> gets cur_thread;
-      buffer \<leftarrow> return $ data_to_oref $ args ! 0;
-      sent \<leftarrow> set_mrs ct (Some buffer) ((ucast consumed) # [ucast (consumed >> 32)]);
+      sent \<leftarrow> set_mrs ct (map_option data_to_oref buf) ((ucast consumed) # [ucast (consumed >> 32)]);
```
MitchellBuckley, line 330: does this need to be here? I'm not sure if data should have been oref the whole time.

spec/machine/ARM/MachineOps.thy
```diff
@@ -144,6 +144,9 @@ and
 axiomatization
   ticks_to_us :: "64 word \<Rightarrow> 64 word"
 
+axiomatization
+  max_ticks_to_us :: "64 word"
```
MitchellBuckley, line 148: Is this necessary? I suspect that `max_ticks_to_us` already existed.

proof/refine/ARM/Tcb_R.thy
```diff
@@ -24,47 +24,236 @@ lemma activate_idle_thread_corres:
...
+ "corres (=) \<top> \<top>
+            (sched_context_update_consumed scp)
+            (schedContextUpdateConsumed scp)"
+  unfolding sched_context_update_consumed_def schedContextUpdateConsumed_def
+  apply (simp add: maxTicksToUs_def ticksToUs_def)
+  apply (rule corres_guard_imp)
+    apply (rule corres_split [OF _ get_sc_corres])
+      apply (rule corres_if)
+        apply (clarsimp simp: maxTicksToUs_def sc_relation_def)
+  sorry
```
MitchellBuckley, line 119: do not introduce new sorries!

proof/refine/ARM/Tcb_R.thy
```diff
@@ -24,47 +24,262 @@ lemma activate_idle_thread_corres:
...
+        apply wpsimp+
+      apply (rule corres_split[OF corres_return_eq_same scConsumed_update_corres],
+             clarsimp simp: sc_relation_def)
+        apply wpsimp+
+  done
+
+crunches sched_context_update_consumed
+  for in_user_frame[wp]: "in_user_frame x"
+
+lemma fhhfhf[wp]:
```
MitchellBuckley, line 148: rename

proof/refine/ARM/Tcb_R.thy
```diff
@@ -24,47 +24,262 @@ lemma activate_idle_thread_corres:
...
+  "schedContextUpdateConsumed scp \<lbrace>\<lambda>s. tcb_at' (ksCurThread s) s\<rbrace>"
+  unfolding schedContextUpdateConsumed_def
+  by (wpsimp | wps)+
+
+lemma fffhhfhf[wp]:
+  "schedContextUpdateConsumed scp \<lbrace>valid_ipc_buffer_ptr' x\<rbrace>"
+  unfolding schedContextUpdateConsumed_def valid_ipc_buffer_ptr'_def
+  by wpsimp
+
+(* this trivial lemma is used to resolve schematic variables in the case that we want
```
MitchellBuckley, line 158: I would like to remove this if possible, but not sure of the best way to remove it.

proof/refine/ARM/Tcb_R.thy
```diff
@@ -3164,3 +3340,4 @@ lemma inv_tcb_IRQInactive:
 end
 
 end
+
```
MitchellBuckley, line 3343: remove extra newline

mbrcknl (approved):
Looks good!

spec/abstract/SchedContext_A.thy
```diff
@@ -309,34 +309,40 @@ definition
   sched_context_update_consumed :: "obj_ref \<Rightarrow> (time,'z::state_ext) s_monad" where
   "sched_context_update_consumed sc_ptr \<equiv> do
     sc \<leftarrow> get_sched_context sc_ptr;
-    set_sc_obj_ref sc_consumed_update sc_ptr 0;
-    return (sc_consumed sc)
+    consumed <- return (sc_consumed sc);
```
michaelmcinerney, line 312: I updated the spec for `sched_context_update_consumed` in my "smaller ipc sorries" PR. I'm definitely happy for you to go first if you'd like, but there is one change that Curtis has made to the C which needs to be included. We need to add `ticks_to_us` before `max_ticks_to_us` in the "then" branch. The C change is part of Curtis's PR #1813 on the SEL4 Bitbucket

proof/invariant-abstract/KHeap_AI.thy
```diff
@@ -2441,8 +2441,9 @@ lemma set_tcb_obj_ref_obj_at_trivial:
   by (wpsimp wp: set_tcb_obj_ref_wp simp: obj_at_def)
 
 lemma sched_context_update_consumed_obj_at_trivial:
-  "\<lbrace>obj_at P t' and
-    K (\<forall>sc n. P (SchedContext sc n) \<longrightarrow> P (SchedContext (sc_consumed_update (K 0) sc) n))\<rbrace>
+  "\<lbrace>obj_at P t'
```
michaelmcinerney, line 2444: I found in my PR that if you add a "forall f" as well, then you can have an arbitrary f to update the consumed time and everything goes through just fine

proof/refine/ARM/Tcb_R.thy
```diff
@@ -24,47 +24,262 @@ lemma activate_idle_thread_corres:
...
+lemma gts_st_tcb':
+  "\<lbrace>tcb_at' t\<rbrace> getThreadState t \<lbrace>\<lambda>rv. st_tcb_at' (\<lambda>st. st = rv) t\<rbrace>"
+  apply (rule hoare_vcg_precond_imp)
+   apply (rule hoare_post_imp[where Q="\<lambda>rv s. \<exists>rv'. rv = rv' \<and> st_tcb_at' (\<lambda>st. st = rv') t s"])
+    apply simp
+   apply (wp hoare_ex_wp)
+  apply (clarsimp simp add: pred_tcb_at'_def obj_at'_def)
+  done
+
+lemma activateIdle_invs:
```
michaelmcinerney, line 36: `invs'` instead? (Though the consistency with this is currently quite poor in Refine)

proof/refine/ARM/Tcb_R.thy
```diff
@@ -24,47 +24,262 @@ lemma activate_idle_thread_corres:
...
+  "activateIdleThread thread \<lbrace>invs'\<rbrace>"
+  by (simp add: activateIdleThread_def)
+
+lemma invs'_live_sc'_ex_nonz_cap_to':
+  "ko_at' ko scp s \<Longrightarrow> invs' s \<Longrightarrow> live_sc' ko \<longrightarrow> ex_nonz_cap_to' scp s"
+  apply (clarsimp simp: invs'_def valid_state'_def if_live_then_nonz_cap'_def)
+  by (fastforce simp: obj_at'_real_def ko_wp_at'_def projectKO_sc)
+
+lemma invs'_ko_at_valid_sc:
```
michaelmcinerney, line 45: I'm much rather use `invs_valid_objs'` and then use the existing rule `sc_ko_at_valid_objs_valid_sc'`

proof/refine/ARM/Tcb_R.thy
```diff
@@ -24,47 +24,262 @@ lemma activate_idle_thread_corres:
...
+  by (wpsimp wp: schedContextUpdateConsumed_invs'
+      | strengthen tcb_at_invs')+
+
+crunches setConsumed
+  for ksSchedulerAction[wp]: "\<lambda>s. P (ksSchedulerAction s)"
+  and tcb_at'[wp]: "\<lambda>s. Q (tcb_at' t s)"
+
+lemma schedContextCompleteYieldTo_invs':
+  "\<lbrace>invs' and sch_act_simple and tcb_at' thread\<rbrace>
+     schedContextCompleteYieldTo thread
```
michaelmcinerney, line 76: Indentation

proof/refine/ARM/Tcb_R.thy
```diff
@@ -24,47 +24,262 @@ lemma activate_idle_thread_corres:
...
+  by (clarsimp simp: invs'_def valid_state'_def valid_idle'_def pred_tcb_at'_def obj_at'_def
+                     idle_tcb'_def)
+
+(* FIXME RT: move *)
+lemma maybeM_when:
+  "maybeM f x = when (x \<noteq> None) (f (the x))"
+  unfolding maybeM_def
+  by (clarsimp split: option.splits)
+
+lemma corres_assert_opt_assume_l'':
```
michaelmcinerney, line 104: Move to Corres_UL?

proof/refine/ARM/Tcb_R.thy
```diff
@@ -24,47 +24,262 @@ lemma activate_idle_thread_corres:
...
+        apply (case_tac obj; simp add: is_sc_obj_def)
+        apply (rule setSchedContext_no_stack_update_corres[where f'="scConsumed_update (\<lambda>_. y)"])
+           apply (clarsimp simp: sc_relation_def objBits_def objBitsKO_def obj_at_def)+
+     apply (fastforce simp: exs_valid_def get_object_def in_monad obj_at_def)
+    apply (wpsimp wp: get_object_wp)
+   apply (fastforce simp: obj_at_def)
+  apply simp
+  done
+
+lemma schedContextUpdateConsumed_corres:
```
michaelmcinerney, line 128: I proved this too in my PR, but your is much much nicer!

proof/refine/ARM/Tcb_R.thy
```diff
@@ -24,47 +24,262 @@ lemma activate_idle_thread_corres:
...
+  apply (rule corres_guard_imp)
+    apply (rule corres_split [OF _ get_tcb_yield_to_corres], simp)
+      apply (rule corres_when2[OF refl])
+      apply (clarsimp, wpfix)
+      apply (rule corres_split[OF _ lipcb_corres], simp)
+        apply (rule corres_split[OF _ setConsumed_corres])
+          apply (clarsimp simp: schedContextCancelYieldTo_def)
+          apply (rule corres_symb_exec_r'[where Q'=\<top>])
+             apply (rule_tac F="scPtrOpt = Some y" in corres_gen_asm2)
+             apply (simp)
```
michaelmcinerney, line 204: Brackets

michaelmcinerney (approved):
Love it!

proof/refine/ARM/Tcb_R.thy
```diff
@@ -24,47 +24,262 @@ lemma activate_idle_thread_corres:
...
+  by (clarsimp split: option.splits)
+
+lemma corres_assert_opt_assume_l'':
+  "corres_underlying sr nf nf' r P Q (f (the X)) g
+  \<Longrightarrow> corres_underlying sr nf nf' r (P and K (X \<noteq> None)) Q (assert_opt X >>= f) g"
+  by (force simp: corres_underlying_def assert_opt_def return_def bind_def fail_def)
+
+lemma scConsumed_update_corres:
+ "x = y \<Longrightarrow>
+  corres dc (sc_at scp) (ko_at' sc scp)
```
michaelmcinerney, line 111: Maybe investigate using `sc_at_cross` to remove the concrete guard? It'll need `pspace_aligned` and `pspace_distinct` but that's probably fine

spec/abstract/SchedContext_A.thy
MitchellBuckley: Sounds good :)

proof/invariant-abstract/KHeap_AI.thy
MitchellBuckley: Awesome!

proof/refine/ARM/Tcb_R.thy
MitchellBuckley: I think in this case we do need the explicit `ko_at` precondition.

spec/machine/MachineExports.thy
```diff
@@ -58,6 +58,7 @@ requalify_consts
   max_ticks_to_us
   us_to_ticks
   ticks_to_us
+  max_ticks_to_us
```
MitchellBuckley, line 61: Remove this.

spec/abstract/Decode_A.thy
```diff
@@ -755,9 +757,9 @@ invocation is allowed.
 definition
   decode_invocation ::
   "bool \<Rightarrow> data \<Rightarrow> data list \<Rightarrow> cap_ref \<Rightarrow> cslot_ptr \<Rightarrow> cap \<Rightarrow>
-   (cap \<times> cslot_ptr) list \<Rightarrow> (invocation, 'z::state_ext) se_monad"
+   (cap \<times> cslot_ptr) list \<Rightarrow> obj_ref option \<Rightarrow> (invocation, 'z::state_ext) se_monad"
 where
-  "decode_invocation first_phase label args cap_index slot cap excaps \<equiv>
+  "decode_invocation first_phase label args cap_index slot cap excaps buffer \<equiv>
```
mktnk3, line 762: Hmm. I am not sure where this change is coming from. Isn’t this `buffer` just the same thing as the third argument `args`?

mktnk3: In other parts of the spec update it appears that `buffer` is replacing `args` but here this function still takes both. Is this how it’s supposed to be?

MitchellBuckley: Not unless there's something tricky going on. See args and buffer in `handle_invocation`
https://github.com/seL4/l4v/blob/rt/spec/abstract/Syscall_A.thy#L231

MitchellBuckley: See also in the source where buffer is passed through.
https://github.com/seL4/seL4/blob/master/src/api/syscall.c#L416

mktnk3: Another question is if this is MCS specific or not, or, how this goes along with the case of master, as in, the argument `buffer` appears in the master version as well. How is this handled?

mktnk3: in `handle_invocation`, `args` are read out of `buffer`. I don’t see why we need to pass both `args` and `buffer` to `decode_invocation`.

MitchellBuckley: I've asked Matt to have a look. I don't know this is supposed to work.

mbrcknl: Yes, I believe MCS does something new here. In particular, before MCS, there were no invocations other than endpoint IPC that needed to *write* to the IPC buffer. Pre-MCS invocation decodes only need to read the syscall args out of the IPC buffer. That meant that in the spec, we could read syscall args early in the decode, and pass those through to the more specific decode functions instead of the buffer. Endpoint IPC operations were special, in that they did (and still do) their own IPC buffer lookups, so we didn't need to pass the buffer through to endpoint IPC operations. The C doesn't have a nice way to pass a list of syscall args through, so the C passes the `buffer` pointer, and individual decode operations use it to call `getSyscallArgs`. In pre-MCS, I think this is all the `buffer` argument gets used for, so for that purpose it's equivalent to the spec-level `args` parameter.

Now, MCS has two new invocations that need to record consumed time in the IPC buffer. So now the (read-only) `args` are not enough. Those new invocations really need the `buffer` pointer. In the C, these invocations use the `buffer` pointer that was passed in.

So what do we do in the spec? I can see these options:
1. Do a second IPC buffer lookup in the invocation specs. But then in CRefine, we'd have to prove that this second lookup gives the same result as the first one in `handleInvocation`. That would be an annoying proof, and I think it would be confusing that the spec performs a lookup while the C doesn't.
2. Pass the `buffer` pointer through to all the invocations, and remove the `args` parameter from all the existing invocations, since we can get the `args` from the `buffer`. This would make the spec look like the C, but it would be a ton of rework for no particular benefit. Actually, I think it would make the spec worse, since it's no longer clear that for most invocations, all you need are the `args`. And the spec would no longer have a nicely factored out syscall args lookup.
3. Pass the `buffer` through only to the new invocations that need to write to it, and remove the `args` from those invocations only. This is a bit more reasonable that 2, but I think it's still annoying to have the syscall args lookup factored out for most invocations, but not for these two.
4. Pass the `buffer` and the `args` through to the new invocations, and leave everything else alone. This is what @MitchellBuckley has done. Although the `args` are redundant, I still think it's the nicest of these 4 options.

mktnk3: Ok, thanks, Matt. I see. With this explanation, I do agree that 4 (what we have here) seems to be the best option for us. It will give us easier time merging with the master as well.

mktnk3: > MCS has two new invocations that need to record consumed time in the IPC buffer, which need the IPC buffer pointer
>  the `buffer` and the `args` through to the invocations, although the `args` are redundant

Mitch, can you please update the commit message to something a bit more explanatory and include something like the above? Also maybe list what those invocations are?

mktnk3: when you clean up later

MitchellBuckley: Ok, can do.

mbrcknl: Here's my suggestion for a commit message for the relevant commit:

```
rt aspec: correct IPC buffer pointer for sched context invocations

In the MCS kernel, two of the invocations on scheduling context
capabilities (consumed and yield-to) need to write the consumed time to
the current thread's IPC buffer.

Prior to this commit, ASpec was mistakenly using the first syscall
argument as the IPC buffer pointer for those invocations.

In this commit, we provide the correct IPC buffer pointer to those
invocations, by adding an extra `buffer` argument to the relevant
invocation decode functions. The IPC buffer pointer is obtained from the
existing call to `lookup_ipc_buffer` in `handle_invocation`. Syscall
arguments are dropped from decoding functions that no longer need them.

This change highlights differences between the pre-MCS and MCS kernels.
In the pre-MCS abstract specification, the only invocation that might
need to write to an IPC buffer is a IPC send on an endpoint. Endpoint
IPC performs its own IPC buffer lookup in `do_normal_transfer`. For all
other invocations, the IPC buffer was only needed for reading syscall
arguments. In the abstract specification, the syscall argument decoding
was therefore performed early in `handle_invocation`, and the syscall
args rather than the buffer were passed to the more specific decode
functions.

Now that MCS requires the IPC buffer pointer for some invocations, we
pass *both* the IPC buffer pointer and the syscall arguments to a small
number of the invocation decode functions, even though the syscall
arguments could be obtained from the IPC buffer. This allows us to keep
performing the syscall argument decoding early in `handle_invocation`,
and to maintain compatibility with existing proofs.

Note that in the C kernel, the IPC buffer pointer is always passed to
individual invocation decoding functions, and syscall arguments are only
decoded when they are needed.
```

mktnk3 (approved):
Thanks a lot for pushing these changes through, Mitch and Matt.

PR #102, opened 2020-09-20
Three sorries in CNodeInv_R.thy

MitchellBuckley:
Solves three sorries in CNodeInv_R.thy.

Commit messages are currently garbage, but the changes are relatively simple so I will fix those later.

### DIFF
--- a/proof/refine/ARM/CNodeInv_R.thy
+++ b/proof/refine/ARM/CNodeInv_R.thy
@@ -583,16 +583,36 @@ lemma suspend_ctes_of_thread:
   apply (case_tac cte, simp)
   done
 
+lemma schedContextUnbindTCB_ctes_of[wp]:
+  "\<lbrace>\<lambda>s. P (ctes_of s)\<rbrace>
+     schedContextUnbindTCB t
+   \<lbrace>\<lambda>_ s. P (ctes_of s)\<rbrace>"
+  apply (wpsimp simp: schedContextUnbindTCB_def wp: threadSet_ctes_ofT)
+     apply (clarsimp simp: ran_def tcb_cte_cases_def split: if_splits)
+  by wpsimp+
+
+crunches setConsumed, schedContextCompleteYieldTo, unbindNotification
+  for ctes_of[wp]: "\<lambda>s. P (ctes_of s)"
+  (simp: crunch_simps wp: crunch_wps)
+
+lemma schedContextUnbindTCB_ctes_of_thread:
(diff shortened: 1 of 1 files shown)


proof/refine/ARM/CNodeInv_R.thy
```diff
@@ -583,16 +583,36 @@ lemma suspend_ctes_of_thread:
...
+   \<lbrace>\<lambda>_ s. P (ctes_of s)\<rbrace>"
+  apply (wpsimp simp: schedContextUnbindTCB_def wp: threadSet_ctes_ofT)
+     apply (clarsimp simp: ran_def tcb_cte_cases_def split: if_splits)
+  by wpsimp+
+
+crunches setConsumed, schedContextCompleteYieldTo, unbindNotification
+  for ctes_of[wp]: "\<lambda>s. P (ctes_of s)"
+  (simp: crunch_simps wp: crunch_wps)
+
+lemma schedContextUnbindTCB_ctes_of_thread:
```
mktnk3, line 598: Do we need these three lemmas? It looks like the one above (`schedContextUnbindTCB_ctes_of`) would solve all of them.

proof/refine/ARM/CNodeInv_R.thy
```diff
@@ -583,16 +583,36 @@ lemma suspend_ctes_of_thread:
...
+
+lemma schedContextCompleteYieldTo_ctes_of_thread:
+  "schedContextCompleteYieldTo t' \<lbrace>\<lambda>s. \<exists>node. ctes_of s x = Some (CTE (ThreadCap t) node)\<rbrace>"
+  by wp
+
+lemma getSchedContext_ctes_of_thread:
+  "getSchedContext t' \<lbrace>\<lambda>s. \<exists>node. ctes_of s x = Some (CTE (ThreadCap t) node)\<rbrace>"
+  by wpsimp
+
 lemma unbindNotification_ctes_of_thread:
```
mktnk3, line 610: Maybe this one, too? Not by the same one but still.

mktnk3 (approved):
Nice work! The only thing is that some of the rules added seem duplicated. We could just instantiate P for specific uses, I imagine?

mktnk3:
After some more thoughts about this, I now think these lemmas are worth having. It does help for the termination proof and it looks like there are other existing ones of the similar style.

MitchellBuckley:
Cheers Miki. Yes it's not ideal, but its consistent with what's there. Hopefully come cleanup and sort this out at a later stage.

PR #103, opened 2020-09-20
rt refine: fix sc_replies_relation (head)

mktnk3:
- in "heap_list hp st xs ed", st can be in the middle of a walk
- however, the sc should be pointing to the reply at the top (most
  recently pushed) of the reply queue/stack
- this commit adds this condition (there is no reply whose replyPrev is
  pointing to the same reply as scReply if it does) to the definition of sc_replies_relation
- also updates the proofs for this fix

### DIFF
--- a/proof/refine/ARM/Retype_R.thy
+++ b/proof/refine/ARM/Retype_R.thy
@@ -1319,8 +1319,15 @@ proof -
      apply (clarsimp simp: pspace_dom_def split: if_split_asm, fastforce)
     apply (clarsimp split: if_split_asm kernel_object.splits)
     apply (frule dom_same')
-    apply (clarsimp simp: opt_map_def)
-    done
+    apply (clarsimp simp: opt_map_left_Some)
+    apply (clarsimp simp:foldr_upd_app_if[folded data_map_insert_def] projectKO_opt_reply
+                   split: if_split_asm kernel_object.split_asm)
+    using ko
+    by (cases ty;
+        simp add: makeObjectKO_def makeObject_reply
+           split: kernel_object.split_asm arch_kernel_object.split_asm object_type.split_asm
+                  apiobject_type.split_asm if_split_asm)
+       fastforce
 qed
 
 lemma retype_state_relation:
--- a/proof/refine/ARM/StateRelation.thy
+++ b/proof/refine/ARM/StateRelation.thy
@@ -456,7 +456,9 @@ definition
   sc_replies_relation_2 ::
   "(obj_ref \<rightharpoonup> obj_ref list) \<Rightarrow> (obj_ref \<rightharpoonup> obj_ref) \<Rightarrow> (obj_ref \<rightharpoonup> obj_ref) \<Rightarrow> bool" where
   "sc_replies_relation_2 sc_repls scRepl replPrevs \<equiv>
-     \<forall>p replies. sc_repls p = Some replies \<longrightarrow> heap_list replPrevs (scRepl p) replies"
+     \<forall>p replies. sc_repls p = Some replies
+                   \<longrightarrow> heap_list replPrevs (scRepl p) replies
+                       \<and> (\<forall>p'. bound (replPrevs p') \<longrightarrow> replPrevs p' \<noteq> scRepl p)"
 
 abbreviation sc_replies_relation :: "det_state \<Rightarrow> kernel_state \<Rightarrow> bool" where
   "sc_replies_relation s s' \<equiv>
--- a/proof/refine/ARM/KHeap_R.thy
+++ b/proof/refine/ARM/KHeap_R.thy
@@ -1295,12 +1295,18 @@ lemma set_reply_corres: (* for reply update that doesn't touch the reply stack *
      apply clarsimp
       (* sc_replies_relation *)
     apply (simp add: sc_replies_relation_def)
-    apply (clarsimp simp: sc_replies_of_scs_def map_project_def scs_of_kh_def)
-    apply (drule_tac x=p in spec)
-    apply (rule conjI; clarsimp simp: sc_of_def split: Structures_A.kernel_object.split_asm if_split_asm)
-    by (subst replyPrevs_of_replyPrev_same_update[simplified, where ob'=ae', simplified];
-        simp add: typ_at'_def ko_wp_at'_def obj_at'_def project_inject opt_map_def sc_of_def)
-  qed
+    apply (rename_tac reply' reply)
+    apply (clarsimp simp: vs_heap_simps)
+    apply (rename_tac scp sc n)
+    apply (drule_tac x=scp and y="sc_replies sc" in spec2, clarsimp)
+    apply (rule conjI)
+     apply (subst replyPrevs_of_replyPrev_same_update[simplified, where ob'=ae', simplified];
+            simp add: typ_at'_def ko_wp_at'_def obj_at'_def project_inject opt_map_def sc_of_def)
+    apply clarsimp
+    apply (drule_tac x=p' in spec)
+    by (fastforce simp: opt_map_left_Some projectKO_opt_sc projectKO_opt_reply replyPrev_same_def
+                 split: if_split_asm)
+qed
 
 lemma setSchedContext_no_stack_update_corres:
   "\<lbrakk> \<forall>sc n sc'. sc_relation sc n sc' \<longrightarrow> sc_relation (f sc) n (f' sc');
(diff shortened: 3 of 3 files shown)


proof/refine/ARM/Retype_R.thy
```diff
@@ -1319,8 +1319,15 @@ proof -
...
-    apply (clarsimp simp: opt_map_def)
-    done
+    apply (clarsimp simp: opt_map_left_Some)
+    apply (clarsimp simp:foldr_upd_app_if[folded data_map_insert_def] projectKO_opt_reply
+                   split: if_split_asm kernel_object.split_asm)
+    using ko
+    by (cases ty;
+        simp add: makeObjectKO_def makeObject_reply
+           split: kernel_object.split_asm arch_kernel_object.split_asm object_type.split_asm
+                 apiobject_type.split_asm if_split_asm)
```
MitchellBuckley, line 1329: nitpick indentation

MitchellBuckley (approved):
Only a single comment on indentation. It looks good!

proof/refine/ARM/StateRelation.thy
```diff
@@ -456,7 +456,9 @@ definition
   sc_replies_relation_2 ::
   "(obj_ref \<rightharpoonup> obj_ref list) \<Rightarrow> (obj_ref \<rightharpoonup> obj_ref) \<Rightarrow> (obj_ref \<rightharpoonup> obj_ref) \<Rightarrow> bool" where
   "sc_replies_relation_2 sc_repls scRepl replPrevs \<equiv>
-     \<forall>p replies. sc_repls p = Some replies \<longrightarrow> heap_list replPrevs (scRepl p) replies"
+     \<forall>p replies. sc_repls p =
+            Some replies \<longrightarrow> heap_list replPrevs (scRepl p) replies
+                             \<and> (\<forall>p'. bound (replPrevs p') \<longrightarrow> replPrevs p' \<noteq> scRepl p)"
```
corlewis, line 461: I would prefer indentation like
```suggestion
     \<forall>p replies. sc_repls p = Some replies
                   \<longrightarrow> heap_list replPrevs (scRepl p) replies
                       \<and> (\<forall>p'. bound (replPrevs p') \<longrightarrow> replPrevs p' \<noteq> scRepl p)"
```

corlewis (approved):
Looks good, although it's not immediately obvious to me that this is the only/best way of expressing it. You might also be able to write the condition as something like `replyNext (scRepl p) = p`, which is more localised but could require more use of `invs'` to actually be useful.

You're the one actually using this though, so if the way you wrote it fits nicely with your use case then I'm definitely happy with it.

proof/refine/ARM/StateRelation.thy
```diff
@@ -456,7 +456,9 @@ definition
   sc_replies_relation_2 ::
   "(obj_ref \<rightharpoonup> obj_ref list) \<Rightarrow> (obj_ref \<rightharpoonup> obj_ref) \<Rightarrow> (obj_ref \<rightharpoonup> obj_ref) \<Rightarrow> bool" where
   "sc_replies_relation_2 sc_repls scRepl replPrevs \<equiv>
-     \<forall>p replies. sc_repls p = Some replies \<longrightarrow> heap_list replPrevs (scRepl p) replies"
+     \<forall>p replies. sc_repls p =
+            Some replies \<longrightarrow> heap_list replPrevs (scRepl p) replies
```
pierzchalski, line 460: Nitpick: I'm not a fan of splitting the `=` here - I was confused for a whole two seconds when I first saw this.

pierzchalski (approved):

mktnk3:
> Looks good, although it's not immediately obvious to me that this is the only/best way of expressing it. You might also be able to write the condition as something like `replyNext (scRepl p) = p`, which is more localised but could require more use of `invs'` to actually be useful.

This is actually a bit subtle, and I don’t think the condition you suggest is the same as what I have. The `replyNext` in your condition is actually `replySc`, and the symmetry you are saying comes from our usual `sym_refs (state_refs_of’ s’)`, and says nothing about `replyPrev`. This is also out side of what `list_refs_of_replies’` because this one only looks at `replyPrev` that matches with `replyNext_of` (not `Head`). So, the condition I am adding here is somewhat like what has fallen in the subtle crack between the two `sym_refs`, and I am fairly convinced that what I have is pretty concise in capturing it.

proof/refine/ARM/StateRelation.thy
mktnk3: ha, yes, I agree!

mktnk3:
Well, what I said is not entirely true. `sym_refs (list_refs_of_replies’)`  does capture non-existence of `replyPrev`. But we don’t have this symmetry as part of the state_relation, and I believe the `sc_replies_relation` needs to capture this without it.

mktnk3:
Sorry, I think I confused myself and also everyone else. If we have `sym_refs (list_refs_of_replies’ s’)`, `replyNext (scRepl p) = p` will give what I am adding here. In fact, just `~ isNext (replyNext (scRepl p)` would be enough. But this sym_refs condition is not part of state_relation (we had this discussion before), so we need to phrase it without using the symmetry, I suppose?

corlewis:
> 
> 
> Sorry, I think I confused myself and also everyone else. If we have `sym_refs (list_refs_of_replies’ s’)`, `replyNext (scRepl p) = p` will give what I am adding here. In fact, just `~ isNext (replyNext (scRepl p)` would be enough. But this sym_refs condition is not part of state_relation (we had this discussion before), so we need to phrase it without using the symmetry, I suppose?

I think I might have missed the previous discussion, is there somewhere I can see it to try and catch up? It's not clear to me why it all needs to be in the state relation and why we can't additionally use things from the invariants.

mktnk3:
> I think I might have missed the previous discussion, is there somewhere I can see it to try and catch up? It's not clear to me why it all needs to be in the state relation and why we can't additionally use things from the invariants.

It was some discussion somewhere leading to #46. This PR (#103) adds what I missed in #46. I think at the point that #46  was written it was already concluded that `sym_refs (list_refs_of_replies’ s’)` is not part of state_relation, and the reason was I believe Gerwin saying that including it in state_relation will result in more pain when we need to deal with situations where the state_relation is broken, although I didn’t fully quite appreciate what that meant.

corlewis:
> It was some discussion somewhere leading to #46. This PR (#103) adds what I missed in #46. I think at the point that #46 was written it was already concluded that `sym_refs (list_refs_of_replies’ s’)` is not part of state_relation, and the reason was I believe Gerwin saying that including it in state_relation will result in more pain when we need to deal with situations where the state_relation is broken, although I didn’t fully quite appreciate what that meant.

Right, I agree that `sym_refs (list_refs_of_replies’ s’)` shouldn't be in the state relation and what I'm suggesting is actually trying to aim more in that direction. What I was going for is that the condition being added to the relation should be as local as possible, and that we then get the rest by appealing to `invs`/`sym_refs`.

However, I've now realised that even my suggestion follows from `sym_refs` and might not be necessary. I haven't fully thought it through, but wouldn't it be possible to write a lemma that says something along the lines of
```
sym_refs (list_refs_of_replies’ s’) ⟹ sym_refs (state_refs_of’ s)
⟹ (∀p p'. bound (replPrevs p') ⟶ replPrevs p' ≠ scRepl p)
```
If it is then we might be able to get away without changing the relation at all.

lsf37:
Coming a bit late to the discussion here, but I think the same reasoning as in #46 applies here. The condition we're adding appears to be only on the concrete level (does not refer to the abstract variables). That would usually indicate that it's better to make this an explicit invariant instead of part of the state relation definition.

The condition we're adding is also global, it talks about all `p'` anywhere in the `scPrevs` heap, not just `p'` associated with the list. It's definitely a bit on the strong side for the base definition, even if it is the property we'd eventually want to use. 

If Corey's lemma hypothesis 

> `sym_refs (list_refs_of_replies’ s’) ⟹ sym_refs (state_refs_of’ s) ⟹ (∀p p'. bound (replPrevs p') ⟶ replPrevs p' ≠ scRepl p)`

is right, then that would definitely be preferable.

Maybe I'm misreading it (should really load this up in Isabelle), but I think this should already be true even without it. From `heap_list` we'll get a list that starts with the contents of `scRepl` and continues through `scPrevs` pointers until it reaches `0`. This list can't contain `scRepl` in any of the `scPrevs` pointers, because that would make the list circular and the abstract list containing the head pointer infinitely often, but we have a lemma that `heap_list` always only produces `distinct` lists (because the abstract lists are finite).

It's a separate step from there to conclude that no `scRepl` anywhere in the heap is contained in any of the `scPrevs` for any list, but I think that should be true as well if we add sym-refs (without sym-refs it would be possible to have one list as a sub-list of another). 

Does this make sense?

mktnk3:
ok, I think I am seeing this. So, `sc_replies_relation` will give `hd (sc_replies sc) = scReply sc’`, and `scReply sc’` (or `scReplies_of s’ scp` for some `scp`) with `sym_refs` statements (both of them) should give that `scReply sc’` is actually at the start of the linked list.

I think what Corey is saying is true. I didn’t think through about having both sym_refs with what I stumbled upon with this,  but I will have another look through there later.

mktnk3:
yes, it does make sense. Both of you.

lsf37:
Cool. Do feel free to ping me if there are further parts we have overlooked!

mktnk3:
I was a bit in a hurry and I didn’t respond in full.

> It's a separate step from there to conclude that no `scRepl` anywhere in the heap is contained in any of the `scPrevs` for any list, but I think that should be true as well if we add sym-refs (without sym-refs it would be possible to have one list as a sub-list of another).

What I wanted in this PR is this part. `sc_replies_relation` gives `heap_list`, but in some discussion with `sym_refs (list_refs_of_replies’ s’)`, I needed to know that the reversed traversal also ends at the original start point, which is the reply what`scReply` points to (i.e., this heap_list is not a sublist). And, as Corey says, I can have that by adding `sym_refs (state_refs_of’ s’)` as well.

PR #105, opened 2020-09-23
rt aspec: spec update for latest version of refill logic

michaelmcinerney:
To mirror Curtis's PR "Fixes for sporadic logic" #1813. 

This is just the spec at the moment, but I thought it would be good to get some opinions before I go too far with things.

### DIFF
--- a/spec/abstract/SchedContext_A.thy
+++ b/spec/abstract/SchedContext_A.thy
@@ -133,21 +133,8 @@ where
      maybe_add_empty_tail sc_ptr
    od"
 
-fun
-  schedule_used :: "bool \<Rightarrow> refill list \<Rightarrow> refill \<Rightarrow> refill list"
-where
-  "schedule_used full [] new = [new]"
-| "schedule_used full (x#rs) new = (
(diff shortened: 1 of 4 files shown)


spec/abstract/SchedContext_A.thy
```diff
@@ -195,72 +203,67 @@ where
...
+
      sc \<leftarrow> get_sched_context sc_ptr;
-     refill_hd \<leftarrow> return $ refill_hd sc;
-     cur_time \<leftarrow> gets cur_time;
-     ready \<leftarrow> get_sc_refill_ready sc_ptr;
-     new_time \<leftarrow> return $ if ready then cur_time else (r_time refill_hd);
-     refill_hd \<leftarrow> return $ \<lparr>r_time = new_time, r_amount = r_amount refill_hd\<rparr>;
+     refills \<leftarrow> return (sc_refills sc);
+     refill_hd \<leftarrow> return (hd refills);
+
```
michaelmcinerney, line 257: I wasn't sure whether to truncate the refill here, as is done in the C, or just truncate to the head in each branch below

PR #107, opened 2020-09-24
rt haskell: refactor replyRemove to match the c

corlewis:
The [C code](https://github.com/seL4/seL4/blob/master/src/object/reply.c#L101) returns early in the `reply_pop` branch, which leads to different behaviour than the previous Haskell implementation. While refactoring this I also moved the `replyUnlink` calls to the end of the function, which is consistent with `replyRemoveTCB` and I think will be convenient when proving hoare triples for `replyRemove`.

It might not be necessary to completely refactor the C code in the same way, but it would make crefine easier if we did. At the very least I think that we should move the `reply_unlink` calls to the end of `reply_remove` there as well.

### DIFF
--- a/spec/haskell/src/SEL4/Object/Reply.lhs
+++ b/spec/haskell/src/SEL4/Object/Reply.lhs
@@ -110,23 +110,21 @@ This module specifies the behavior of reply objects.
 
 >     nextReplyPtrOpt <- return $ replyNext reply
 >     prevReplyPtrOpt <- return $ replyPrev reply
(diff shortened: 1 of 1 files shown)


pierzchalski (approved):
Looks good. I'm fine with updating the C, even though it results in a weird repeated check that I'm sure someone will complain about.

MitchellBuckley (approved):
Looks good to me!

mktnk3 (approved):
This makes sense and is a nice refactor.

PR #109, opened 2020-10-02
rt: finish most of finaliseCap_replaceable

corlewis:
The remaining subgoal involves showing that a reply that isn't linked to a tcb does not have any prev/next pointers. Unfortunately, we can only get from the abstract invariants that it is not part of a queue connected to a scheduling context. To solve this properly we expect that we will need a new Haskell-level invariant.

### DIFF
--- a/proof/refine/ARM/CSpace_I.thy
+++ b/proof/refine/ARM/CSpace_I.thy
@@ -14,8 +14,8 @@ begin
 
 context begin interpretation Arch . (*FIXME: arch_split*)
 
-lemmas capUntypedPtr_simps[simp] = capUntypedPtr_def[split_simps capability.split]
-lemmas arch_capUntypedPtr_simps[simp] = ARM_H.capUntypedPtr_def[split_simps arch_capability.split]
+lemmas capUntypedPtr_simps[simp] = capUntypedPtr_def[split_simps capability.split, simplified PPtr_def id_def]
+lemmas arch_capUntypedPtr_simps[simp] = ARM_H.capUntypedPtr_def[split_simps arch_capability.split, simplified PPtr_def id_def]
 
 lemma rights_mask_map_UNIV [simp]:
   "rights_mask_map UNIV = allRights"
--- a/proof/refine/ARM/Finalise_R.thy
+++ b/proof/refine/ARM/Finalise_R.thy
@@ -1731,7 +1731,9 @@ where
  "final_matters' cap \<equiv> case cap of
     EndpointCap ref bdg s r g gr \<Rightarrow> True
   | NotificationCap ref bdg s r \<Rightarrow> True
+  | ReplyCap ref gr \<Rightarrow> True
   | ThreadCap ref \<Rightarrow> True
+  | SchedContextCap ref sz \<Rightarrow> True
   | CNodeCap ref bits gd gs \<Rightarrow> True
   | Zombie ptr zb n \<Rightarrow> True
   | IRQHandlerCap irq \<Rightarrow> True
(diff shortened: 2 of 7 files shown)


proof/refine/ARM/CSpace_I.thy
```diff
@@ -14,8 +14,8 @@ begin
 
 context begin interpretation Arch . (*FIXME: arch_split*)
 
-lemmas capUntypedPtr_simps[simp] = capUntypedPtr_def[split_simps capability.split]
-lemmas arch_capUntypedPtr_simps[simp] = ARM_H.capUntypedPtr_def[split_simps arch_capability.split]
+lemmas capUntypedPtr_simps[simp] = capUntypedPtr_def[split_simps capability.split, simplified PPtr_def id_def]
+lemmas arch_capUntypedPtr_simps[simp] = ARM_H.capUntypedPtr_def[split_simps arch_capability.split, simplified PPtr_def id_def]
```
corlewis, line 18: I'm not convinced by these changes, but this makes these lemmas the same as on master and means that I don't need to unfold `PPtr_def` in later proofs.

MitchellBuckley (approved):
Love it!

proof/refine/ARM/Finalise_R.thy
```diff
@@ -3032,6 +3017,118 @@ crunches schedContextCompleteYieldTo, unbindNotification, unbindMaybeNotificatio
   for sch_act_simple[wp]: sch_act_simple
   (simp: crunch_simps wp: crunch_wps rule: sch_act_simple_lift)
 
+lemma schedContextZeroRefillMax_unlive[wp]:
```
michaelmcinerney, line 3020: This isn't the weakest precondition, is it? Can't we just have that there's an sc at scPtr, and `schedContextZeroRefillMax` will make it unlive?

michaelmcinerney (approved):

proof/refine/ARM/Finalise_R.thy
mktnk3: scRefillMax doesn’t change the liveness, I think. So, this should be ok.

proof/refine/ARM/CSpace_I.thy
mktnk3: Hmm. I don’t have a particular view on this. I suppose this should be fine for the current state of Refine.

mktnk3 (approved):
Looks good! I hope the new invariant part works out alright. Let me know how it goes. Do you plan to introduce these functions in the C finaliseCap as well?

proof/refine/ARM/Finalise_R.thy
corlewis: Yeah, what Miki said. `finaliseCap` actually makes scheduling contexts unlive over a couple of steps, which is what `schedContextUnbindYieldFrom_makes_unlive` and the following `obj_at'` lemmas are showing.

corlewis:
> Do you plan to introduce these functions in the C finaliseCap as well?

I wasn't sure, and was planning to leave thinking about it until we get to CRefine. It's not necessary since it's pretty easy to inline by unfolding the definitions as part of a refine proof. It's mainly helpful here because it makes it a lot simpler to write tidy hoare triples about them.

proof/refine/ARM/CSpace_I.thy
lsf37: No problem from my side. It's probably not really consistent currently anyway and if it helps, go for it.

mktnk3:
> > Do you plan to introduce these functions in the C finaliseCap as well?
> 
> I wasn't sure, and was planning to leave thinking about it until we get to CRefine. It's not necessary since it's pretty easy to inline by unfolding the definitions as part of a refine proof. It's mainly helpful here because it makes it a lot simpler to write tidy hoare triples about them.

I see. I agree that it makes sense to wait and see how things go in CRefine, and that, unless there arises some particular needs, we probably don’t need to.

PR #110, opened 2020-10-03
Latest refill logic

michaelmcinerney:
To mirror Curtis's PR "Fix for sporadic logic", #1813 on the seL4 Bitbucket

https://bitbucket.ts.data61.csiro.au/projects/SEL4/repos/sel4/pull-requests/1813/overview

### DIFF
--- a/proof/invariant-abstract/DetSchedSchedule_AI.thy
+++ b/proof/invariant-abstract/DetSchedSchedule_AI.thy
@@ -58,7 +58,7 @@ method overflow_hammer =
     | erule iffD1[OF word_le_nat_alt])+)[1]
 
 lemma cur_time_no_overflow:
-  "current_time_bounded 0 s \<Longrightarrow> cur_time s \<le> cur_time s + kernelWCET_ticks"
+  "current_time_bounded l s \<Longrightarrow> cur_time s \<le> cur_time s + kernelWCET_ticks"
   unfolding current_time_bounded_def
   apply (overflow_hammer, clarsimp)
   done
@@ -7002,25 +7002,19 @@ lemma valid_refills_consumed_time_update[iff]:
   by simp
 
 lemma sc_valid_refills_unat_sum_list_at_most_unat_max_word:
-  "\<lbrakk>sc_valid_refills sc; sc_period sc \<noteq> 0;
-    MIN_SC_BUDGET \<le> sc_budget sc\<rbrakk>
+  "\<lbrakk>sc_valid_refills sc; sc_period sc \<noteq> 0; MIN_BUDGET \<le> sc_budget sc\<rbrakk>
    \<Longrightarrow> sum_list (map unat (map r_amount (sc_refills sc))) \<le> unat max_time"
   apply (rule unat_sum_list_at_most_unat_max_word)
       apply (clarsimp simp: sc_valid_refills_def rr_valid_refills_def split: if_splits)+
-    apply (metis MIN_BUDGET_nonzero unat_gt_0 word_le_0_iff)
-  apply (clarsimp simp: sc_valid_refills_def refills_sum_def rr_valid_refills_def split: if_splits)
-  using MIN_BUDGET_le_MIN_SC_BUDGET by force+
+  done
 
 lemma valid_refills_refills_unat_sum_equals_budget:
-  "\<lbrakk>sc_valid_refills sc; sc_period sc \<noteq> 0;
-    MIN_SC_BUDGET \<le> sc_budget sc\<rbrakk>
+  "\<lbrakk>sc_valid_refills sc; sc_period sc \<noteq> 0; MIN_BUDGET \<le> sc_budget sc\<rbrakk>
    \<Longrightarrow> sum_list (map unat (map r_amount (sc_refills sc))) = unat (sc_budget sc)"
   supply map_map [simp del]
   apply (clarsimp simp: sc_valid_refills_def rr_valid_refills_def split: if_splits)
    apply (rule_tac s="refills_sum (sc_refills sc)" and t="sc_budget sc" in subst; assumption?)
    apply (rule unat_sum_list_equals_budget; fastforce?)
-    apply (metis MIN_BUDGET_nonzero unat_eq_zero word_le_0_iff)
-  using MIN_BUDGET_le_MIN_SC_BUDGET apply force
   done
 
 lemma valid_refills_scheduler_action_update[iff]:
@@ -7124,13 +7118,22 @@ lemma refill_new_valid_sched_misc[wp]:
           (etcbs_of s) (tcb_sts_of s) (tcb_scps_of s) (tcb_faults_of s) (sc_replies_of s)\<rbrace>"
   unfolding refill_new_def by wpsimp
 
+lemma refill_unblock_check_valid_sched_misc[wp]:
+  "refill_unblock_check scptr
+   \<lbrace>\<lambda>s. P (consumed_time s) (cur_sc s) (ep_send_qs_of s) (ep_recv_qs_of s) (sc_tcbs_of s)
+          (cur_time s) (cur_domain s) (cur_thread s) (idle_thread s)
+          (ready_queues s) (release_queue s) (scheduler_action s)
+          (etcbs_of s) (tcb_sts_of s) (tcb_scps_of s) (tcb_faults_of s) (sc_replies_of s)\<rbrace>"
+  unfolding refill_unblock_check_def
+  by (wpsimp wp: hoare_drop_imp)
+
 lemma refill_update_valid_sched_misc[wp]:
   "refill_update sc_ptr period budget mrefills
    \<lbrace>\<lambda>s. P (consumed_time s) (cur_sc s) (ep_send_qs_of s) (ep_recv_qs_of s) (sc_tcbs_of s)
           (cur_time s) (cur_domain s) (cur_thread s) (idle_thread s)
           (ready_queues s) (release_queue s) (scheduler_action s)
           (etcbs_of s) (tcb_sts_of s) (tcb_scps_of s) (tcb_faults_of s) (sc_replies_of s)\<rbrace>"
-  unfolding refill_update_def by wpsimp
+  unfolding refill_update_def by (wpsimp wp: hoare_drop_imps)
 
 lemma maybe_add_empty_tail_is_active_sc[wp]:
   "maybe_add_empty_tail sc_ptr \<lbrace>\<lambda>s. Q (is_active_sc scp s)\<rbrace>"
@@ -7162,6 +7165,7 @@ lemma maybe_add_empty_tail_is_refill_sufficient[wp]:
 
 crunches set_refills, refill_update, refill_new
   for pred_tcb_at[wp]: "\<lambda>s. P (pred_tcb_at p Q t s)"
+  (wp: crunch_wps)
 
 lemma set_sc_obj_ref_refill_cfgs_no_change:
   "\<lbrace>\<lambda>s. Q (pred_map R (sc_refill_cfgs_of s) scp)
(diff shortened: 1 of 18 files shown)


proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -9659,15 +9550,12 @@ lemma cur_sc_bounded_release_helper:
...
-                       r_amount (refill_hd sc) < consumed) \<longrightarrow>
-                      cur_sc_offset_ready consumed s))
-    and current_time_bounded 2\<rbrace>
+    and cur_sc_offset_ready 0
+    and current_time_bounded 3\<rbrace>
    refill_budget_check consumed
    \<lbrace>\<lambda>_. active_sc_valid_refills\<rbrace>"
   unfolding active_sc_valid_refills_def
-  apply (wpsimp wp: hoare_vcg_all_lift hoare_vcg_imp_lift
+  apply_trace (wpsimp wp: hoare_vcg_all_lift hoare_vcg_imp_lift
```
michaelmcinerney, line 9558: Tsk tsk

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -15918,9 +15814,10 @@ lemma invoke_sched_control_configure_valid_sched:
     apply (clarsimp simp: pred_map_eq_def pred_map_def obj_at_kh_kheap_simps)
 
    apply (rule hoare_seq_ext[OF _ gets_sp])
+   apply (simp add: return_bind)
    apply (clarsimp split: if_split)
    apply (intro conjI impI)
-    apply (rule hoare_weaken_pre[where Q=valid_sched])
+    apply (rule_tac Q="valid_sched " in hoare_weaken_pre)
```
michaelmcinerney, line 15820: Extra space

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -16330,8 +16239,8 @@ lemma handle_yield_valid_sched:
    apply (rule valid_refills_r_amount_bounded_max_sc_period)
    apply (clarsimp simp: valid_refills_def2 obj_at_def)
   apply (clarsimp simp: cur_sc_offset_ready_def obj_at_def refill_ready_def)
-  apply (simp add: word_le_nat_alt)
-  apply (subst (asm) unat_add_lem', overflow_hammer)
+
```
michaelmcinerney, line 16242: New line

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -21874,13 +21739,20 @@ crunches commit_time, tcb_sched_action
   for sc_tcbs_of[wp]: "\<lambda>s. P (sc_tcbs_of s)"
   (ignore: update_sched_context wp: crunch_wps)
 
+lemma cur_sc_tcb_bound_ready1:
```
michaelmcinerney, line 21742: Check to see whether this can be removed, or whether this and the `cur_sc_tcb_bound_ready` can be generalised

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -21910,18 +21782,19 @@ lemma invoke_sched_control_configure_ct_ready_if_schedulable[wp]:
...
-                             \<and> MIN_SC_BUDGET \<le> budget"
+                             \<and> MIN_BUDGET \<le> budget"
            in hoare_post_imp)
      apply (subgoal_tac "budget_ready (cur_thread s) s")
-      apply (fastforce simp: ct_ready_if_schedulable_def split: if_split)
+      apply (fastforce simp: ct_ready_if_schedulable_def current_time_bounded_def split: if_split)
      apply (clarsimp simp: tcb_at_kh_simps pred_map_eq_normalise)
-     apply (erule (2) cur_sc_tcb_bound_ready)
+     apply (erule (1) cur_sc_tcb_bound_ready1)
+apply simp
```
michaelmcinerney, line 21797: Indentation

proof/invariant-abstract/SchedContextInv_AI.thy
```diff
@@ -1281,378 +1206,158 @@ lemma butlast_nonempty_length:
...
+  apply (cases list)
+   apply (clarsimp simp: schedule_used_def Let_def ordered_disjoint_def)
+  apply (case_tac "can_merge_refill (last list) new")
+   apply (clarsimp simp: schedule_used_def Let_def)
+   apply (intro conjI impI)
+    apply (clarsimp simp: ordered_disjoint_def)
    apply (rule_tac left="a # butlast lista"
-              and right="[last lista\<lparr>r_amount := r_amount (last lista) - (MIN_BUDGET - r_amount new)\<rparr>,
-                                    \<lparr>r_time = r_time new - (MIN_BUDGET - r_amount new), r_amount = MIN_BUDGET\<rparr>]"
+              and right="[\<lparr>r_time = r_time (last lista),
```
michaelmcinerney, line 1219: Do a rename_tac for lista

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -7209,14 +7213,19 @@ lemma sc_refills_update_refill_ready_sc[wp]:
   apply (wpsimp wp: update_sched_context_wp)
   by (clarsimp simp: pred_map_simps vs_all_heap_simps obj_at_def)
 
+lemma refill_unblock_check_not_active_sc_at[wp]:
```
michaelmcinerney, line 7216: Rename; check other lemmas for `refill_unblock_check` and `is_active_sc`

MitchellBuckley:
@michaelmcinerney Is this a draft PR? I'm going to start reviewing, but stop me if you're still in "draft" stage.

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -22536,21 +22394,26 @@ lemma call_kernel_valid_sched:
...
         apply (clarsimp simp: obj_at_def vs_all_heap_simps active_sc_def cur_sc_more_than_ready_def)
        apply (clarsimp simp: ct_ready_if_schedulable_def)
       apply (fastforce simp: heap_refs_inv_def intro: invs_retract_sc_tcbs invs_retract_tcb_scps)
      apply wpsimp
+    apply wpsimp
     apply (rule_tac Q="\<lambda>_. valid_sched
                            and invs
                            and consumed_time_bounded
-                           and next_time_bounded 2
+                           and next_time_bounded 3
```
MitchellBuckley, line 22416: I'm surprised that this is needed. Perhaps double check this?

But there is no problem if this needs to be larger, as long as it's constant and small.

MitchellBuckley: Although now that I think about it, it's not surprising if changes to refill logic would affect this.

MitchellBuckley (approved):
Perfect indeed!! Only a few nitpicky comments :+1:

proof/invariant-abstract/DetSchedSchedule_AI.thy
michaelmcinerney: The new version of `refill_update` will definitely require us to increase the buffer we have before `max_time`.  I'm not 100% sure that `refill_budget_check` will as well, but `refill_update` needing this is enough to make it necessary globally

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -8972,134 +8987,101 @@ lemma refill_budget_check_valid_ready_qs_not_queued:
...
+                current_time_bounded_strengthen le0 unat_plus_gt_trans unat_plus_simple word_sub_le)
+  apply (clarsimp simp: sc_valid_refills_def split: if_splits)
+   apply (clarsimp simp: obj_at_def)
+  apply (clarsimp simp: obj_at_def)
+  apply (case_tac "0 < usage")
+   using MIN_BUDGET_pos apply simp
+  apply (clarsimp simp: refill_ready_def refill_ready_no_overflow_def)
+  apply (clarsimp simp: current_time_bounded_def word_le_nat_alt)
+  apply (metis add_leE unat_plus_simple unat_sum_bound_equiv)
   done
```
mktnk3, line 9046: As discussed in the MCS meeting today, we divide the reviewing task of this file in two, suggesting here for the division point. For the sake of simplifying the process, I would like to request reviews from:

Corey, Victor
----------< division point
Ed, Matt

For others, please add names to either side if you are joining the review. I’ll look though the file anyway. Thanks.

michaelmcinerney:
It'd be great if some reviewers, especially those with CRefine experience, could check the Haskell and the abstract spec against the C. I think it should be fine, but I'm not 100% sure that I've got everything right

mktnk3 (commented):
I looked through the spec, both abstract and haskell, against the C code, and I don’t notice any issues apart from only a couple of comments, mostly more about the C code. There are lots of details to follow and I hope I didn’t miss any things in the details. I will continue to look at the proofs.

spec/abstract/Structures_A.thy
```diff
@@ -480,9 +480,7 @@ record sched_context =
 
 definition "MIN_REFILLS = 2"
 definition "MIN_BUDGET = 2 * kernelWCET_ticks"
-definition "MIN_SC_BUDGET = 2 * MIN_BUDGET"
```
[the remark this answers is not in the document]
michaelmcinerney: I think you're right

spec/haskell/src/SEL4/Object/SchedContext.lhs
```diff
@@ -256,96 +256,79 @@ This module uses the C preprocessor to select a target architecture.
...
->         scheduleUsed scPtr new
+>         let new = Refill { rTime = rTime head + newPeriod, rAmount = unused }
+>         refillAddTail scPtr new
 
 > scheduleUsed :: PPtr SchedContext -> Refill -> Kernel ()
 > scheduleUsed scPtr new = do
 >     sc <- getSchedContext scPtr
 >     assert (scRefillMax sc > 0) "scPtr should be active"
 >     empty <- refillEmpty scPtr
 >     if empty
```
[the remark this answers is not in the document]
michaelmcinerney: I definitely did want a branch in this function for the empty case, but I forget the exactly reason. I'll ask Curtis whether he is happy to have a branch in the C for this too. We should never call this function where the sc has empty refills (it's now called only in `refill_budget_check`, by the way), but if it makes verification easier, then I would think it would be better to have the branch there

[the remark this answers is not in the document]
michaelmcinerney: I asked Curtis, and he said that if the branch if not required, then he will have a hard time convincing Anna to keep it there. I would think that we would very much like to define `schedule_used` to be a total function, so I can't see how we could get around having the branch in the abstract spec. And in the C, in order to properly access the `REFILL_TAIL`, the sc must not be empty. So I can't see a good way around this. Should an assert be added to the C? Or at least a comment saying that it shouldn't be called on an empty list?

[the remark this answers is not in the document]
michaelmcinerney: OK so Curtis still does not want a branch to be added to the C, but he is happy to move the assert that the refill list is nonempty to the beginning of `schedule_used`. So I think the best approach would be to leave the abstract and Haskell as they are. When we prove refinement, we can assume that the refill list is nonempty (we would call `refill_budget_check` only on a scheduling context that satisfies `valid_refills`, and so in particular is not empty). Does that sounds OK?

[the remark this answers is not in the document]
mktnk3: Yes, that sounds great. Thanks.

mktnk3 (approved):
This is a huge chunk of amazing work. Thanks so much for this, Michael. I fail to find anything even to nitpick. I have an impression that the proofs are shorter and nicer overall, although I suspect the complexity in the details are just as hard as before.

michaelmcinerney:
> This is a huge chunk of amazing work. Thanks so much for this, Michael. I fail to find anything even to nitpick. I have an impression that the proofs are shorter and nicer overall, although I suspect the complexity in the details are just as hard as before.

Thanks Miki. I think you're right that the details are just as difficult as before. And with `MIN_BUDGET_merge`, there are more details than before. But at least `refill_budget_check` having the one branch means that things are more modular, and therefore slightly easier to understand and maintain

mktnk3:
It seems that the Isabelle website is down since last night, which appears to have caused the error in Haskell test.

michaelmcinerney:
The HaskellKernel test was failing before due to the Isabelle website being down, but I have tested it locally, and it passed

PR #112, opened 2020-10-07
rt: cancel IPC when finalising reply caps 

corlewis:
Deleting a reply object that is linked to a BlockedOnReceive thread is unusual behaviour and it's not clear what the expected behaviour should be. Previously replyUnlink was used, but it sets the thread to Inactive while leaving it in the endpoint's queue which breaks the invariants. Calling cancelIPC at least guarantees that the system ends up in a consistent state.

While updating the proofs I was able to delete or prove a few sorries and some crunches became more powerful, which unfortunately might end up conflicting with other ongoing work.

### DIFF
--- a/spec/haskell/src/SEL4/Object/ObjectType.lhs
+++ b/spec/haskell/src/SEL4/Object/ObjectType.lhs
@@ -573,3 +573,10 @@ The following two functions returns the base and size of the object a capability
 > capUntypedSize (IRQHandlerCap {})
 >     = 1 -- error in haskell
 
+> replyClear :: PPtr Reply -> PPtr TCB -> Kernel ()
+> replyClear rptr tptr = do
+>     state <- getThreadState $ tptr
+>     case state of
+>         BlockedOnReply _ -> replyRemove rptr tptr
+>         BlockedOnReceive {} -> cancelIPC tptr
+>         _ -> fail "replyClear: invalid state of replyTCB"
--- a/proof/refine/ARM/CNodeInv_R.thy
+++ b/proof/refine/ARM/CNodeInv_R.thy
@@ -631,10 +631,8 @@ lemma suspend_not_recursive_ctes:
   unfolding updateRestartPC_def
   apply (wp threadSet_ctes_of | simp add: unless_def del: o_apply)+
          apply (fold cteCaps_of_def)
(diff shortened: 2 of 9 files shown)


spec/haskell/src/SEL4/Object/ObjectType.lhs
```diff
@@ -573,3 +573,10 @@ The following two functions returns the base and size of the object a capability
 > capUntypedSize (IRQHandlerCap {})
 >     = 1 -- error in haskell
 
+> replyClear :: PPtr Reply -> PPtr TCB -> Kernel ()
+> replyClear rptr tptr = do
+>     state <- getThreadState $ tptr
+>     case state of
+>         BlockedOnReply _ -> replyRemove rptr tptr
+>         BlockedOnReceive {} -> cancelIPC tptr
+>         _ -> fail "replyClear: invalid state of replyTCB"
```
corlewis, line 582: This definition had to be moved out of Reply.lhs to resolve a circular dependency.

mktnk3 (approved):
Nice :)
I am very glad that this spec change turned out to be quite painless.

mbrcknl (approved):
Looks good to me.

I suppose the only other reasonable choice might be to remove the reply from the thread while leaving the thread `BlockedOnReceive`, effectively making the receiver only willing to receive ordinary sends, not calls. That might in some sense be "less policy", but as you say, there's no obvious precedent to tell us what to do here. `cancelIPC` seems a reasonable choice.

mbrcknl:
I presume there is (or will be) a corresponding patch to the C?

corlewis:
> I presume there is (or will be) a corresponding patch to the C?

Yep, I'm just waiting until I actually manage to finish these proofs so that I can be a bit more confident that I won't need to change anything else.

corlewis:
> I suppose the only other reasonable choice might be to remove the reply from the thread while leaving the thread `BlockedOnReceive`, effectively making the receiver only willing to receive ordinary sends, not calls. That might in some sense be "less policy", but as you say, there's no obvious precedent to tell us what to do here. `cancelIPC` seems a reasonable choice.

I agree that that would be slightly more general, but there isn't an existing function that does that and it would require a lot of work to update the existing abstract invariant proofs. If we come across a specific use case then we might want to reconsider this decision, but for now we have all been happy to go with `cancelIPC`.

PR #113, opened 2020-10-08
rt haskell + refine: prove si_invs', sfi_invs_plus' and hf_invs'

ryybrr:
Proved si_invs' and updated the haskell spec in a few places.

### DIFF
--- a/proof/refine/ARM/IpcCancel_R.thy
+++ b/proof/refine/ARM/IpcCancel_R.thy
@@ -2742,6 +2742,59 @@ proof -
   done
 qed
 
+lemma replyUnlink_sch_act[wp]:
+  "replyUnlink r t \<lbrace>\<lambda>s. sch_act_wf (ksSchedulerAction s) s\<rbrace>"
+  apply (simp only: replyUnlink_def setReplyTCB_def getReplyTCB_def liftM_def)
+  apply (wpsimp wp: sts_sch_act' gts_wp')
+  apply (fastforce simp: replyUnlink_assertion_def st_tcb_at'_def obj_at'_def)
+  done
+
+lemma replyUnlink_list_refs_of_replies'[wp]:
+  "replyUnlink r t \<lbrace>\<lambda>s. P (list_refs_of_replies' s)\<rbrace>"
+  unfolding replyUnlink_def setReplyTCB_def updateReply_def getReplyTCB_def
+  apply (wpsimp simp: updateObject_default_def setObject_def split_def wp: gts_wp')
+  apply (erule arg_cong[where f=P, THEN iffD1, rotated])
+  apply (rule ext)
+  apply (clarsimp simp: opt_map_def sym_refs_def fun_upd_def list_refs_of_reply'_def
+                        map_set_def projectKO_opt_reply obj_at'_real_def ko_wp_at'_def
+                 split: option.split)
+  done
+
+lemma replyUnlink_valid_pspace'[wp]:
+  "\<lbrace>\<lambda>s. valid_pspace' s\<rbrace>
+   replyUnlink r t
+   \<lbrace>\<lambda>_ s. valid_pspace' s\<rbrace>"
+  unfolding replyUnlink_def setReplyTCB_def updateReply_def getReplyTCB_def
+  apply (wpsimp wp: gts_wp' simp: valid_tcb_state'_def)
+  apply (frule valid_pspace_valid_objs')
+  apply (frule(1) reply_ko_at_valid_objs_valid_reply')
+  apply (drule(1) ko_at'_inj)
+  apply (clarsimp simp: valid_reply'_def)
+  done
+
+lemma replyUnlink_if_live_then_nonz_cap'[wp]:
+  "\<lbrace>\<lambda>s. if_live_then_nonz_cap' s\<rbrace>
+   replyUnlink r t
+   \<lbrace>\<lambda>_ s. if_live_then_nonz_cap' s\<rbrace>"
+  unfolding replyUnlink_def setReplyTCB_def updateReply_def getReplyTCB_def
+  apply (wpsimp wp: gts_wp')
+  apply (erule if_live_then_nonz_capE')
+  apply normalise_obj_at'
+  apply (clarsimp simp: live_reply'_def)
+  done
+
+lemma replyUnlink_valid_idle'[wp]:
+  "\<lbrace>\<lambda>s. valid_idle' s \<and> valid_pspace' s \<and> t \<noteq> ksIdleThread s\<rbrace>
+   replyUnlink r t
+   \<lbrace>\<lambda>_ s. valid_idle' s\<rbrace>"
+  unfolding replyUnlink_def setReplyTCB_def updateReply_def getReplyTCB_def
+  apply (wpsimp wp: gts_wp' simp: valid_reply'_def)
+  apply (frule valid_pspace_valid_objs')
+  apply (frule(1) reply_ko_at_valid_objs_valid_reply')
+  apply (drule(1) ko_at'_inj)
+  apply (clarsimp simp: valid_reply'_def)
+  done
+
 lemma replyUnlink_valid_irq_node'[wp]:
   "replyUnlink r t \<lbrace>\<lambda> s. valid_irq_node' (irq_node' s) s\<rbrace>"
   unfolding replyUnlink_def setReplyTCB_def getReplyTCB_def
--- a/proof/refine/ARM/Ipc_R.thy
+++ b/proof/refine/ARM/Ipc_R.thy
@@ -1911,15 +1911,15 @@ lemma doIPCTransfer_sch_act_simple [wp]:
   by (simp add: sch_act_simple_def, wp)
 
 lemma possibleSwitchTo_invs'[wp]:
-  "\<lbrace>invs' and st_tcb_at' runnable' t
-          and (\<lambda>s. ksSchedulerAction s = ResumeCurrentThread \<longrightarrow> ksCurThread s \<noteq> t)\<rbrace>
-   possibleSwitchTo t \<lbrace>\<lambda>_. invs'\<rbrace>"
-  apply (simp add: possibleSwitchTo_def curDomain_def inReleaseQueue_def)
-  apply (wp tcbSchedEnqueue_invs' ssa_invs')
-       apply (rule hoare_post_imp[OF _ rescheduleRequired_sa_cnt])
-      apply (wpsimp wp: ssa_invs' threadGet_wp)+
-  apply (clarsimp dest!: obj_at_ko_at' simp: tcb_in_cur_domain'_def obj_at'_def)
-  by fastforce
+  "\<lbrace>invs' and st_tcb_at' runnable' tptr\<rbrace>
+   possibleSwitchTo tptr
+   \<lbrace>\<lambda>_. invs'\<rbrace>"
+  apply (simp add: possibleSwitchTo_def)
+  apply (wpsimp wp: hoare_vcg_imp_lift threadGet_wp inReleaseQueue_wp ssa_invs')
+  apply (clarsimp simp: invs'_def valid_state'_def valid_idle'_def
+                        idle_tcb'_def pred_tcb_at'_def obj_at'_def
+                        ct_idle_or_in_cur_domain'_def tcb_in_cur_domain'_def)
+  done
 
 crunches isFinalCapability
   for cur' [wp]: "\<lambda>s. P (cur_tcb' s)"
@@ -2087,6 +2087,7 @@ lemma possibleSwitchTo_valid_queues[wp]:
 lemma schedContextDonate_valid_queues':
   "schedContextDonate sc t \<lbrace> valid_queues' \<rbrace>"
   apply (clarsimp simp: schedContextDonate_def)
+  apply (rule hoare_seq_ext_skip, solves wpsimp)
   apply (rule hoare_seq_ext_skip, solves wpsimp)
   apply (rule hoare_seq_ext_skip)
    apply (rule hoare_when_cases, simp)
@@ -2299,6 +2300,7 @@ lemma send_ipc_corres:
              (send_ipc bl call bg cg cgr cd t ep) (sendIPC bl call bg cg cgr cd t ep)"
 proof -
   show ?thesis
+  sorry (*
   apply (insert assms)
   apply (unfold send_ipc_def sendIPC_def Let_def)
   apply (case_tac bl)
@@ -2321,7 +2323,6 @@ proof -
              apply (simp add: fault_rel_optionation_def)
             apply wp+
           apply (clarsimp simp: st_tcb_at_tcb_at valid_tcb_state_def)
-  sorry (*
          apply clarsimp
          \<comment> \<open>concludes IdleEP if bl branch\<close>
         apply (simp add: ep_relation_def)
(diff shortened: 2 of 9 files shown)


proof/refine/ARM/IpcCancel_R.thy
```diff
@@ -2868,11 +2868,11 @@ proof -
 qed
 
 lemma replyUnlink_sch_act[wp]:
-  "\<lbrace>\<lambda>s. sch_act_wf (ksSchedulerAction s) s \<and> sch_act_not t s\<rbrace>
-   replyUnlink r t
-   \<lbrace>\<lambda>_ s. sch_act_wf (ksSchedulerAction s) s\<rbrace>"
-  apply (clarsimp simp: replyUnlink_def setReplyTCB_def getReplyTCB_def liftM_def)
-  by (wpsimp wp: sts_sch_act' hoare_drop_imp)
+  "replyUnlink r t \<lbrace>\<lambda>s. sch_act_wf (ksSchedulerAction s) s\<rbrace>"
```
michaelmcinerney, line 2871: Sweeet

proof/refine/ARM/Ipc_R.thy
```diff
@@ -3560,160 +3560,622 @@ lemma cteInsert_ct'[wp]:
...
+  apply (auto simp: obj_at'_def projectKO_eq projectKO_sc valid_idle'_def)
+  done
+
+lemma schedContextDonate_valid_release_queue'[wp]:
+  "schedContextDonate scPtr tcbPtr \<lbrace>valid_release_queue'\<rbrace>"
+  unfolding schedContextDonate_def
+  by (wpsimp wp: threadSet_valid_release_queue' simp: obj_at_ko_at'_eq valid_release_queue'_def
+      | rule hoare_strengthen_post[where Q="\<lambda>_. valid_release_queue'"])+
+
+lemma schedContextDonate_assume_asserts:
```
michaelmcinerney, line 3640: A little rule to introduce the `sc_at'` fact is nice, but I'd prefer to not include the `sym_refs` assert here; we may add more asserts in the future, which would mean updating this too, and it's easy enough to use `stateAssert_sp` to add the assert to the preconditions

proof/refine/ARM/Ipc_R.thy
```diff
@@ -3560,160 +3560,622 @@ lemma cteInsert_ct'[wp]:
...
+  apply (simp only: obj_at'_real_def)
+  apply (wp set_reply'.ko_wp_at)
+  apply (auto simp: obj_at'_def ko_wp_at'_def projectKO_eq projectKO_reply)
+  done
+
+lemma setReplyTCB_obj_at':
+  "\<lbrace>\<lambda>s. reply_at' rptr s \<longrightarrow>
+          P (obj_at' (\<lambda>ko. if rptr = p then Q (replyTCB_update (\<lambda>_. tptrOpt) ko) else Q ko) p s)\<rbrace>
+   setReplyTCB tptrOpt rptr
+  \<lbrace>\<lambda>rv s. P (obj_at' (Q :: reply \<Rightarrow> bool) p s)\<rbrace>"
```
michaelmcinerney, line 3702: Nitpick: indentation?

proof/refine/ARM/Ipc_R.thy
```diff
@@ -3560,160 +3560,622 @@ lemma cteInsert_ct'[wp]:
...
+     (fastforce simp: valid_tcb_state'_def valid_bound_obj'_def obj_at'_def projectKO_eq
+                      projectKO_ep projectKO_reply projectKO_ntfn projectKO_tcb
+               split: thread_state.splits option.splits)
+
+crunches setThreadState
+  for sc_at'_n[wp]: "sc_at'_n n p"
+  and sc_ko_at'[wp]: "\<lambda>s. P (ko_at' (sc :: sched_context) p s)"
+  (wp: crunch_wps simp: crunch_simps)
+
+definition replyPush' where
```
michaelmcinerney, line 3794: Maybe it'll become evident as I keep reviewing, but could you give me a little explanation as to why you want this new definition?

proof/refine/ARM/Ipc_R.thy
```diff
@@ -3560,160 +3560,622 @@ lemma cteInsert_ct'[wp]:
...
+       replyPush' callerPtr calleePtr replyPtr canDonate scPtrOptDonated scPtrOptCallee;
+       when (scPtrOptDonated \<noteq> Nothing \<and> (if scPtrOptCallee \<noteq> Nothing then False else canDonate)) $
+           schedContextDonate (fromJust scPtrOptDonated) calleePtr od)
+   (replyPush callerPtr calleePtr replyPtr canDonate)"
+  unfolding replyPush_def replyPush'_def
+  apply (clarsimp simp only: K_bind_def bind_assoc)
+  apply (rule monadic_rewrite_gen_asm[simplified pred_conj_def]
+              monadic_rewrite_bind_tail[OF _ return_sp]
+              monadic_rewrite_bind_tail[OF _ hoare_post_taut])+
+  apply (case_tac "(scPtrOptDonated \<noteq> Nothing \<and> x)")
```
michaelmcinerney, line 3846: What's x? Could you do a rename_tac?

proof/refine/ARM/Ipc_R.thy
```diff
@@ -3560,160 +3560,622 @@ lemma cteInsert_ct'[wp]:
...
+           simp: getReplyTCB_def obj_at'_def split_del: if_split)
+
+lemma replyPush_sch_act_wf:
+  "\<lbrace>\<lambda>s. sch_act_wf (ksSchedulerAction s) s \<and> sch_act_not callerPtr s\<rbrace>
+   replyPush callerPtr calleePtr replyPtr canDonate
+   \<lbrace>\<lambda>_ s. sch_act_wf (ksSchedulerAction s) s\<rbrace>"
+  unfolding replyPush_def getReplyTCB_def
+  by (wpsimp wp: sts_sch_act' hoare_vcg_all_lift hoare_vcg_if_lift hoare_drop_imps)
+
+lemma replyPush_sym_refs_helper:
```
michaelmcinerney, line 3883: Could you put `list_refs_of_replies'` in the name of this one and the following lemma to distinguish it from regular old `sym_refs (state_refs_of' s)`?

proof/refine/ARM/Ipc_R.thy
```diff
@@ -3560,160 +3560,622 @@ lemma cteInsert_ct'[wp]:
...
+        apply (wp hoare_vcg_all_lift hoare_vcg_const_imp_lift)
+        apply (simp (no_asm_use) add: replyPush'_def getReplyTCB_def split del: if_split
+               | wp hoare_vcg_all_lift hoare_vcg_disj_lift hoare_vcg_imp_lift'
+                    hoare_vcg_if_lift setReply_obj_at' setReplyTCB_obj_at'
+               | rule threadGet_wp)+
+  apply (clarsimp simp: obj_at'_def projectKO_eq projectKO_tcb projectKO_sc projectKO_reply)
+  apply (rule context_conjI; clarsimp simp: sym_refs_asrt_def)
+   apply (erule if_live_then_nonz_capE')
+   apply (drule_tac ko=obj in sym_refs_ko_atD'[rotated])
+    apply (auto simp: refs_of_rev' live_sc'_def idle_tcb'_def valid_idle'_def
```
michaelmcinerney, line 3939: Did we ever agree on a standard when using auto in the middle of a proof? Was it "subgoal by" that we wanted? (And check to see if a fastforce or something else is enough here, if you haven't already)

proof/refine/ARM/Ipc_R.thy
```diff
@@ -3560,160 +3560,622 @@ lemma cteInsert_ct'[wp]:
...
+  "\<lbrace>\<lambda>s. invs' s \<and> sch_act_not callerPtr s \<and> st_tcb_at' active' callerPtr s \<and>
+        ex_nonz_cap_to' callerPtr s \<and> ex_nonz_cap_to' calleePtr s \<and> ex_nonz_cap_to' replyPtr s\<rbrace>
+   replyPush callerPtr calleePtr replyPtr canDonate
+   \<lbrace>\<lambda>_ s. invs' s\<rbrace>"
+  unfolding invs'_def valid_state'_def valid_pspace'_def
+  by (wpsimp wp: replyPush_valid_objs' replyPush_sch_act_wf replyPush_valid_queues
+                 replyPush_if_live_then_nonz_cap' replyPush_valid_idle'
+           simp: valid_pspace'_def)
+
+lemma setThreadState_Running_invs':
```
michaelmcinerney, line 4004: Can you generalise this to the case where the thread state is restart, or any other thread state? Is it because it's runnable? I'm not sure...

proof/refine/ARM/Ipc_R.thy
```diff
@@ -3560,160 +3560,622 @@ lemma cteInsert_ct'[wp]:
...
-  sendIPC bl call ba cg cgr cd t ep
-  \<lbrace>\<lambda>rv. invs'\<rbrace>"
+   sendIPC bl call ba cg cgr cd t ep
+   \<lbrace>\<lambda>rv. invs'\<rbrace>"
   supply if_split[split del]
   apply (simp add: sendIPC_def)
+  apply (rule hoare_seq_ext[OF _ stateAssert_sp])
   apply (rule hoare_seq_ext [OF _ get_ep_sp'])
-  apply (case_tac epa)
+  apply (case_tac x)
```
michaelmcinerney, line 4097: Rename_tac

proof/refine/ARM/Ipc_R.thy
```diff
@@ -3560,160 +3560,622 @@ lemma cteInsert_ct'[wp]:
...
+                                   (\<exists>sc. bound_sc_tcb_at' ((=) sc) t s \<and>
+                                         ex_nonz_cap_to' (the sc) s))"
+                     in hoare_strengthen_post[rotated])
+         apply (fastforce simp: obj_at'_def projectKO_eq projectKO_tcb
+                                pred_tcb_at'_def invs'_def valid_state'_def
+                         dest!: global'_no_ex_cap)
+        apply (wpsimp simp: invs'_def valid_state'_def valid_pspace'_def comp_def sym_refs_asrt_def
+                        wp: hoare_vcg_all_lift hoare_vcg_ex_lift hoare_vcg_imp_lift' gts_wp')+
+    apply (intro conjI; clarsimp?)
+        apply (auto simp: obj_at'_def projectKO_eq projectKO_ep valid_obj'_def valid_ep'_def
```
michaelmcinerney, line 4117: Auto here too

proof/refine/ARM/TcbAcc_R.thy
```diff
@@ -5224,20 +5220,19 @@ lemma sts_invs_minor':
                    \<and> (st \<noteq> Inactive \<and> \<not> idle' st \<longrightarrow>
                       st' \<noteq> Inactive \<and> \<not> idle' st')) t
       and (\<lambda>s. t = ksIdleThread s \<longrightarrow> idle' st)
-      and sch_act_simple
+      and sch_act_not t
```
michaelmcinerney, line 5223: Awesome!

michaelmcinerney (approved):
Excellent!

proof/refine/ARM/Ipc_R.thy
ryybrr: Some wp rules require valid_objs' as a precondition to schedContextDonate. Having a monadic rewrite rule for replyPush means we only need to prove that valid_objs' is invariant over replyPush'.

proof/refine/ARM/Ipc_R.thy
```diff
@@ -3560,160 +3560,622 @@ lemma cteInsert_ct'[wp]:
...
+  "valid_tcb_state' ts (s\<lparr>ksReadyQueues := qs\<rparr>) = valid_tcb_state' ts s"
+  "valid_tcb_state' ts (s\<lparr>ksReadyQueuesL1Bitmap := l1b\<rparr>) = valid_tcb_state' ts s"
+  "valid_tcb_state' ts (s\<lparr>ksReadyQueuesL2Bitmap := l2b\<rparr>) = valid_tcb_state' ts s"
+  "valid_tcb_state' ts (s\<lparr>ksSchedulerAction := sa\<rparr>) = valid_tcb_state' ts s"
+  by (auto simp: valid_tcb_state'_def valid_bound_obj'_def split: thread_state.splits option.splits)
+
+crunches addToBitmap, setQueue
+  for valid_tcb_state'[wp]: "valid_tcb_state' ts"
+
+lemma setSchedContext_obj_at':
```
corlewis, line 3680: Are you able to move this and the following lemmas to the `simple_ko'` locale? If you can it would let you avoid duplicating it for the various different types and constants.

proof/refine/ARM/Ipc_R.thy
corlewis: I don't think that we have a strong rule on this, but the preference is to avoid `auto` in the middle of proofs. In this case it's a bit more complicated since the auto is solving two goals, and we also don't want to just copy and paste it if we can avoid it.

In my mind, the optimal but possibly hard than it's worth solution would be to somehow handle all three of the final subgoals together, in one terminating `auto`. The other options that I'm equally happy with are leaving it as is, or using `prefer` to handle the `fastforce` first, before finishing with the `auto`.

@mbrcknl @Xaphiosis @lsf37 @mktnk3 Do any of you have an opinion?

proof/refine/ARM/Ipc_R.thy
```diff
@@ -3560,160 +3560,622 @@ lemma cteInsert_ct'[wp]:
...
+
+lemma replyPush_valid_idle':
+  "\<lbrace>valid_idle' and valid_pspace' and st_tcb_at' active' callerPtr\<rbrace>
+   replyPush callerPtr calleePtr replyPtr canDonate
+   \<lbrace>\<lambda>_. valid_idle'\<rbrace>"
+  apply (simp only: replyPush_def)
+  apply (simp split del: if_split
+         | wp hoare_vcg_if_lift[where f="setReply _ _"] hoare_vcg_imp_lift'
+              hoare_vcg_all_lift setReply_obj_at' schedContextDonate_valid_idle'
+         | (rule hoare_lift_Pf3[where f=ksIdleThread,rotated]; wp setSchedContext_obj_at'))+
```
corlewis, line 3953: I'm fine with this if it's necessary, but this is a bit of a messy apply step. It definitely doesn't work with some combination of `wpsimp` and `wps`?

pierzchalski (approved):

corlewis (approved):
Great job with this, both with the new proofs and also with the existing preconditions that you've managed to simplify!

proof/refine/ARM/Ipc_R.thy
mktnk3: This does seem interesting. I would like to have a closer look at this...

proof/refine/ARM/Ipc_R.thy
mktnk3: again, I think this one is ok, too, since it’s limited to one subgoal. You can do `subgoal`, too, of course, but I don’t think there is particular advantage in choosing that over `[1]` here. I might be wrong, though...

mktnk3 (approved):
Nicely done! This is a big chunk of work - thanks, Ryan.

proof/refine/ARM/Ipc_R.thy
mktnk3: > A little rule to introduce the `sc_at'` fact is nice, but I'd prefer to not include the `sym_refs` assert here; we may add more asserts in the future, which would mean updating this too, and it's easy enough to use `stateAssert_sp` to add the assert to the preconditions

Actually, I am not sure what exactly mean, Michael. Which assert do you prefer not including?

michaelmcinerney: I'd rather not have `sym_refs (state_refs_of' s)` in the precondition here. I think it would be better to have a rule to add the `sc_at'` bit to the preconditions, and just use `stateAssert_sp` separately to get the `sym_refs` bits. That way, if we add further asserts to `schedContextDonate` later on, we won't need to update this proof too. Not that this is a big deal

mktnk3: Right. I see. We might want to look again why Ryan wants `sym_refs` in the later proofs.

ryybrr: > I'd rather not have `sym_refs (state_refs_of' s)` in the precondition here. I think it would be better to have a rule to add the `sc_at'` bit to the preconditions, and just use `stateAssert_sp` separately to get the `sym_refs` bits. That way, if we add further asserts to `schedContextDonate` later on, we won't need to update this proof too. Not that this is a big deal
It's not in the precondition, this is just a rule to add it to the precondition. To make it clearer, I've removed this lemma and added the proofs to `schedContextDonate_invs'` instead.

proof/refine/ARM/Ipc_R.thy
ryybrr: > What's x? Could you do a rename_tac?

It's a bit hard to rename this one; I don't know what name I could give to `(if scPtrOptCallee ≠ Nothing then False else canDonate)`, and there are 9 other variables after `x` that I'd have to rename too unless there's a way to target one variable in particular.

michaelmcinerney: Ah. In that case, don't worry about it

proof/refine/ARM/Ipc_R.thy
ryybrr: > I'm fine with this if it's necessary, but this is a bit of a messy apply step. It definitely doesn't work with some combination of `wpsimp` and `wps`?

This is about as simple as I can make it, unfortunately. The simp step needs to take precedence over the wp step, and the application of hoare_lift_Pf3 causes looping when I use it with the other wp rules.

proof/refine/ARM/Ipc_R.thy
ryybrr: Resolved by changing the auto instance into two separate fastforces

proof/refine/ARM/Ipc_R.thy
ryybrr: Changed to force since fastforce didn't work

proof/refine/ARM/Ipc_R.thy
pierzchalski: This looks like a missing variant of `simple_ko'.obj_at'_strongest`. There are "nice" "simple" variants for `P (ko_at' ko p s)` but not for `P (obj_at' Q p s)`, so it makes sense to add a `simple_ko'.obj_at'` which has the form of your `setSchedContext_obj_at'`: this is probably the form people find most useful/familiar.

PR #114, opened 2020-10-08
rt haskell refine: make reply functions more consistent

corlewis:
This involves small refactors to replyPop, replyRemove and replyRemoveTCB, and also makes these functions slightly easier to prove properties about. The existing proofs about them are also updated.

### DIFF
--- a/proof/refine/ARM/Reply_R.thy
+++ b/proof/refine/ARM/Reply_R.thy
@@ -32,13 +32,16 @@ lemma replyUnlink_st_tcb_at'_sym_ref:
   apply (fastforce simp: obj_at'_def projectKOs)
   done
 
+crunches cleanReply
+  for st_tcb_at'[wp]: "st_tcb_at' P t"
+  (wp: crunch_wps simp: crunch_simps ignore: threadSet)
+
 lemma replyRemoveTCB_st_tcb_at'_Inactive':
   "\<lbrace>\<top>\<rbrace>
    replyRemoveTCB tptr
    \<lbrace>\<lambda>_. st_tcb_at' ((=) Inactive) tptr\<rbrace>"
   unfolding replyRemoveTCB_def
-  apply (wpsimp wp: replyUnlink_st_tcb_at')
-  apply (clarsimp simp: pred_tcb_at'_def)
+  apply (wpsimp wp: replyUnlink_st_tcb_at' hoare_vcg_if_lift split_del: if_split)
   done
 
 lemma replyRemoveTCB_st_tcb_at'[wp]:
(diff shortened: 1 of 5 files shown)


proof/refine/ARM/Reply_R.thy
```diff
@@ -66,33 +69,28 @@ lemma replyUnlink_tcb_obj_at'_no_change:
...
+                            set_reply'.set_no_update[where upd="\<lambda>r. (replyNext_update Map.empty r)"]
+                            set_reply'.get_ko_at')
+         apply (rule_tac Q="\<lambda>_. obj_at' (\<lambda>r. replyTCB r = Some tptr) rptr" in hoare_post_imp,
+                clarsimp)
+         apply (wpsimp wp: hoare_vcg_imp_lift'
+                           set_reply'.set_no_update[where upd="\<lambda>r. (replyPrev_update Map.empty r)"]
+                           set_reply'.get_ko_at')
+        apply (rule_tac Q="\<lambda>_. obj_at' (\<lambda>r. replyTCB r = Some tptr) rptr" in hoare_post_imp,
+               clarsimp)
+        apply (wpsimp wp: gts_wp')+
```
corlewis, line 93: I'm not happy with this proof but I think a nicer version requires the `updateReply` changes from #95.

pierzchalski (approved):

mktnk3:
I am not sure if we really want this - the `replyNext` update in `cleanReply` corresponds to setting `reply_sc` to None when the reply is at the head of the reply stack. Swapping `replyUnlink` and `cleanReply` makes the corres proof harder, I think.

mktnk3:
Maybe it is okay with `replyRemove` but probably not for `replyRemoveTCB` and `replyPop`.

corlewis:
> I am not sure if we really want this - the `replyNext` update in `cleanReply` corresponds to setting `reply_sc` to None when the reply is at the head of the reply stack. Swapping `replyUnlink` and `cleanReply` makes the corres proof harder, I think.

Ah, I think you might be right. The main thing that I wanted was for them to all be consistent, it was then just slightly easier for me to prove a few hoare triples if they were in this order. The other order should be fine for me, and it looks like consistently having the replyUnlink at the end of the functions would also be closer to the abstract.

mktnk3:
> Would you be happy with this change if I swapped the order?

Haha, I had to look again to see what would be left if we move `cleanReply` back to where it was ...
So, the only change that remains would be moving down `replyUnlink` to the end of `replyPop`, is this right? Sure, I have no problem with that.

corlewis:
> So, the only change that remains would be moving `replyUnlink` down to the end of `replyPop`, is this right? Sure, I have no problem with that.

Yep, that's right. I'll also update the C, since the order there doesn't always match.

corlewis:
I'll probably end up combining the remaining changes from this with my next PR.

PR #117, opened 2020-10-13
rt refine: finish replyClear_makes_unlive

corlewis:
Doing this requires introducing a new invariant called `valid_replies'`. Adding this to `invs'` and proving that it is preserved still needs to be done.

### DIFF
--- a/proof/refine/ARM/Finalise_R.thy
+++ b/proof/refine/ARM/Finalise_R.thy
@@ -3605,7 +3605,7 @@ lemma replyPop_valid_objs'[wp]:
                in hoare_weaken_pre[rotated])
    apply (fastforce intro!: reply_ko_at_valid_objs_valid_reply')
   apply (intro hoare_seq_ext[OF _ assert_sp])
-  apply (rule hoare_seq_ext_skip, wpsimp wp: replyUnlink_valid_objs')
+  apply (rule hoare_seq_ext, wpsimp wp: replyUnlink_valid_objs')
   apply (rule hoare_when_cases, clarsimp)
   apply (rule hoare_seq_ext[OF _ assert_sp])
   apply (rule hoare_seq_ext_skip; wp)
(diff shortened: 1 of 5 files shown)


proof/refine/ARM/Finalise_R.thy
```diff
@@ -2945,32 +2955,99 @@ lemma schedContextMaybeUnbindNtfn_obj_at'_ntfnSc:
...
+   frule (2) sym_ref_replyTCB_Receive_or_Reply,
+   fastforce simp: weak_sch_act_wf_def pred_tcb_at'_def obj_at'_def projectKOs)
+
+lemma cancelIPC_makes_unlive:
+  "\<lbrace>\<lambda>s. obj_at' (\<lambda>reply. replyTCB reply = Some tptr) rptr s
+        \<and> weak_sch_act_wf (ksSchedulerAction s) s \<and> valid_replies' s\<rbrace>
+   cancelIPC  tptr
+   \<lbrace>\<lambda>_. ko_wp_at' (Not \<circ> live') rptr\<rbrace>"
+  apply (simp add: cancelIPC_def Let_def getBlockingObject_def sym_refs_asrt_def)
+  apply (rule hoare_seq_ext[OF _ stateAssert_sp])
```
michaelmcinerney, line 3024: Maybe worth replacing these two lines with an `apply (intro hoare_seq_ext[OF _ stateAssert_sp]`? You could lump the following sp rule into that `intro` too

michaelmcinerney (approved):

proof/refine/ARM/Invariants_H.thy
```diff
@@ -997,6 +997,12 @@ definition "vs_valid_duplicates' \<equiv> \<lambda>h.
               x && ~~ mask (vs_ptr_align ko) = y && ~~ mask (vs_ptr_align ko) \<longrightarrow>
               h x = h y"
 
+definition valid_replies' :: "kernel_state \<Rightarrow> bool" where
```
mktnk3, line 1000: This name already exists in AInvs, actually. But maybe the one in AInvs could do a better name...hmm.

mktnk3: And I am wondering how this relates to `valid_replies` in AInvs. I would hope crossing from abstract to concrete to hold. Crossing `sc_with_reply` will give you the `replyPrev/replyNext` assumption, I think, so probably it will work.

mktnk3 (approved):
Nice to see all these new pieces coming together :)
I hope proving the `valid_replies’` invariant goes smoothly!

PR #119, opened 2020-10-19
regression: use `python3` for tests-xml-correct

mbrcknl:
PEP 394 expects that Python 3 installations provide a `python3` command,
but does not require a `python` command. Some distributions (including
Debian) are no longer providing a `python` command, but do provide
`python3`.

In this change, the `python3` interpreter is invoked via the existing
`#!` line in the `testspec.py` script.


mbrcknl:
[L4V-TESTBOARD-621](https://bamboo.ts.data61.csiro.au/browse/L4V-TESTBOARD-621)

MitchellBuckley (approved):

LukeMondy (approved):

PR #122, opened 2020-10-23
mcs: improve typ_at_props locales and typ_at lifting

corlewis:
This makes the locales explicitly discharge all `typ_at_lifts` assumptions and updates the proofs to use the locales instead of manually instantiating `typ_at_lifts` with a lemmas command.

To properly discharge all of the assumptions I added a `REPEAT` attribute, that applies the given attributes one or more times. This should successfully handle almost all standard uses of attributes that I have been able to think of, with the only exception being `symmetric` which will loop. We could guard against this by passing along all seen thms and checking that we haven't looped to any previous one, but this is a bit unpleasant and wouldn't perform well in complicated cases. Opinions and suggestions welcome!

As side distractions, while doing this I also added wp rules for several Haskell assert functions and moved some proofs outside of the anonymous locale that we use to access `Arch` facts. I think the first change is definitely helpful and will lead to more consistent behaviour, although it did require some annoying updates to existing proofs. Moving proofs out of the locale is less beneficial, and I mainly did it because I was already needing to drop out of and back in to the anonymous locale. I guess it might make a future arch split slightly easier?

### DIFF
--- a/lib/Repeat_Attribute.thy
+++ b/lib/Repeat_Attribute.thy
@@ -1,0 +1,78 @@
+(*
+ * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ *)
+
+theory Repeat_Attribute
+imports Main
+begin
+
+ML \<open>
+local
+
+val attribute_generic = Context.cases Attrib.attribute_global Attrib.attribute
+
+fun apply_attributes attrs thm ctxt =
+  let
+    val (thm', ctxt') = fold (uncurry o Thm.apply_attribute) attrs (thm, ctxt)
+  in if Thm.eq_thm (thm, thm')
+     then (SOME ctxt', SOME thm)
+     else
+       apply_attributes attrs thm' ctxt'
+       handle e =>
+         (if Exn.is_interrupt e then Exn.reraise e else ();
+          (SOME ctxt', SOME thm'))
+  end
+
+fun repeat_attribute_cmd attr_srcs (ctxt, thm) =
+  let val attrs = map (attribute_generic ctxt) attr_srcs
+  in apply_attributes attrs thm ctxt end
+
+in
+
+val _ = Theory.setup
+  (Attrib.setup @{binding REPEAT}
+    (Attrib.attribs >> repeat_attribute_cmd)
+    "higher order attribute combinator to repeatedly apply other attributes one or more times")
+
+end
+\<close>
+
+text \<open>
+
+  The @{attribute REPEAT} attribute is an attribute combinator that repeatedly applies
+  other attributes one or more times. It will stop applying the attributes once they can
+  either no longer be applied, or if applying them would not change the theorem being
+  modified.
+
+  Usage:
+    thm foo[REPEAT [<attributes>]]
+
+\<close>
+
+section \<open>Examples\<close>
+
+experiment begin
+  lemma test1: "True \<Longrightarrow> True" .
+  lemma test2: "True \<Longrightarrow> True \<Longrightarrow> True" .
+  lemmas tests = test1 test2
+  text \<open>
+    `tests[OF TrueI]` would only discharge one of the assumptions of @{thm test2}, but
+    @{attribute REPEAT} handles both cases.
+  \<close>
+  thm tests[REPEAT [OF TrueI]]
+
+  text \<open>
+    @{attribute REPEAT} succesfully terminates when applying an attribute that could loop,
+    such as @{attribute simplified} and @{attribute simp}. Importantly, it still updates
+    the context if necessary, in this case by adding P to the simp set.
+  \<close>
+  lemma
+    assumes P[REPEAT[simplified], REPEAT [simp]]: "P \<or> False"
+    shows P
+    by simp
+
+end
+
+end
--- a/proof/refine/ARM/ArchAcc_R.thy
+++ b/proof/refine/ARM/ArchAcc_R.thy
@@ -1173,10 +1173,16 @@ lemma lookup_pt_slot_corres [@lift_corres_args, corres]:
 
 declare in_set_zip_refl[simp]
 
-crunch typ_at'[wp]: copyGlobalMappings "\<lambda>s. P (typ_at' T p s)"
+crunches copyGlobalMappings
+  for typ_at'[wp]: "\<lambda>s. P (typ_at' T p s)"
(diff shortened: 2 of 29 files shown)


corlewis:
I think the most relevant people to look at this are @lsf37, @pierzchalski and @mbrcknl, but reviews from everyone else would definitely be welcome.

lib/Repeat_Attribute.thy
```diff
@@ -0,0 +1,78 @@
...
+    "higher order attribute combinator to repeatedly apply other attributes one or more times")
+
+end
+\<close>
+
+text \<open>
+
+  The @{attribute REPEAT} attribute is an attribute combinator that repeatedly applies
+  other attributes one or more times. It will stop applying the attributes once they can
+  either no long be applied, or if applying them would not change the theorem being
```
lsf37, line 46:
```suggestion
  either no longer be applied, or if applying them would not change the theorem being
```

proof/refine/ARM/ArchAcc_R.thy
```diff
@@ -1173,10 +1173,16 @@ lemma lookup_pt_slot_corres [@lift_corres_args, corres]:
...
 declare in_set_zip_refl[simp]
 
-crunch typ_at'[wp]: copyGlobalMappings "\<lambda>s. P (typ_at' T p s)"
+crunches copyGlobalMappings
+  for typ_at'[wp]: "\<lambda>s. P (typ_at' T p s)"
+  and sc_at'_n[wp]: "\<lambda>s. P (sc_at'_n n p s)"
   (wp: mapM_x_wp')
+end
+
+sublocale Arch < copyGlobalMappings: typ_at_all_props' "copyGlobalMappings newPD"
```
lsf37, line 1182: That looks like a nice way to do it!

proof/refine/ARM/CSpace1_R.thy
```diff
@@ -512,8 +513,6 @@ lemma cap_table_at_gsCNodes:
   apply blast
   done
 
-declare resolve_address_bits'.simps[simp del]
```
lsf37, line 515: I think we want this one globally -- although best declared directly after the definition of `resolve_address_bits'`, not in a much later theory here. Any specific reason to move this into the lemma context?

proof/refine/ARM/Invariants_H.thy
```diff
@@ -2844,6 +2790,119 @@ lemmas typ_at_lifts = typ_at_lifts_internal
...
+  assumes scs: "f \<lbrace>\<lambda>s. Q (sc_at'_n n p s)\<rbrace>"
+begin
+
+lemmas typ_at_sc_at'_n_lifts'[wp] = typ_at_sc_at'_n_lifts[OF typ' scs]
+
+context begin
+(* We want to enforce that typ_at_sc_at'_n_lifts' only contains lemmas that have no
+   assumptions. The following thm statements should fail if this is not true. *)
+private lemmas check_valid_internal = iffD1[OF refl, where P="valid p g q" for p g q]
+thm typ_at_lifts'[atomized, THEN check_valid_internal]
```
lsf37, line 2897: Nice!

proof/refine/ARM/TcbAcc_R.thy
```diff
@@ -4637,27 +4649,24 @@ lemma sbn_st_tcb'[wp]:
...
 
-interpretation setBoundNotification: typ_at_props' "setBoundNotification v p"
-  by (unfold_locales) (rule setBoundNotification_typ_at')
+global_interpretation threadSet: typ_at_all_props' "threadSet f p"
+  by typ_at_props'
 
+global_interpretation setThreadState: typ_at_all_props' "setThreadState st p"
+  by typ_at_props'
 
+global_interpretation setBoundNotification: typ_at_all_props' "setBoundNotification v p"
```
lsf37, line 4668: After seeing so many of these, I'm starting to think it'd be nice to have a short-hand command doing many interpretations at the same time (a bit like `crunches`). Problem for some other time, but it might actually be fairly easy to write.

lsf37 (approved):
Nice! Looks like a lot more work than I anticipated, but it makes things much more uniform.

The CI check for Refine failed, I haven't looked at that in detail.

corlewis:
Please ignore the Refine failure, I made these changes on top of #117 and it should pass once that one is merged.

proof/refine/ARM/CSpace1_R.thy
corlewis: I don't remember if I had a specific reason for this beyond noticing it and not liking the random declare in the middle of some unrelated proofs.  If you think we want to keep it globally I'm happy to move it to the definition as part of this PR. It seems like it won't affect anything else.

proof/refine/ARM/TcbAcc_R.thy
corlewis: Yeah, I thought the same and something like that would definitely have been nice while I was making these changes. This was already enough of a side project though, so I resisted the temptation to implement it.

proof/refine/ARM/CSpace1_R.thy
lsf37: Staring at the definition again, it should actually be fine to do this locally. I can't see a situation where you would want it automatically unfolded, but the case I was trying to avoid (non-terminating `simp` wherever `resolve_address_bits'` occurs) doesn't actually apply, because the first thing on the rhs of it is a big case statement.

It'd be slightly cleaner to move it to the definition site in ASpec, but I'm Ok either way.

pierzchalski (approved):

proof/refine/ARM/CSpace1_R.thy
```diff
@@ -583,7 +582,7 @@ proof (induct a arbitrary: c' cref' bits rule: resolve_address_bits'.induct)
       hence [simp]: "((cbits + length guard = 0) = False) \<and>
                      ((cbits = 0 \<and> guard = []) = False) \<and>
                     (0 < cbits \<or> guard \<noteq> []) " by simp
-      note if_split [split del] drop_append[simp del]
+      note if_split [split del] drop_append[simp del] resolve_address_bits'.simps[simp del]
```
mktnk3, line 585: Since I see that you’ve added this in ASpec, we could remove this one?

mktnk3 (approved):
This is very nice and neat! Thanks for doing this, Corey.

We could do with a few JIRA issues related to this PR:
- a shorthand command for simultaneous locale interpretations
- backport REPEAT attribute to master
- do we want typ_at_props locale in master, too? One day?

PR #123, opened 2020-10-24
Preemption should not be linked with an irq (VER-1368)

mktnk3:
In the abstract spec, the preemption monad `p_monad` is defined to carry around an irq that is supposedly coming from the interrupt that caused the exemption. However, in C, apparently the kernel does not assume that an exemption to be associated to an irq.

This gap presents an issue in verifying the MCS kernel because we want to have scheduling context/time related exemptions that do not have any associated irqs. This PR removes the dependency on irqs from the definition of `p_monad` (on master).

I am not too sure the way I fixed the proofs is the best one, but it is fairly minimum, so I hope this is good enough.

The PR for MCS: #125.

### DIFF
--- a/spec/abstract/Exceptions_A.thy
+++ b/spec/abstract/Exceptions_A.thy
@@ -38,7 +38,7 @@ the current thread or to its fault handler.
 type_synonym ('a,'z) lf_monad = "(lookup_failure + 'a,'z) s_monad"
 
 text \<open>The preemption monad. May throw an interrupt exception.\<close>
-type_synonym ('a,'z) p_monad = "(interrupt + 'a,'z) s_monad"
+type_synonym ('a,'z) p_monad = "(unit + 'a,'z) s_monad"
 
 
 text \<open>
@@ -49,7 +49,7 @@ translations
   (type) "'a f_monad" <= (type) "(fault + 'a) s_monad"
   (type) "'a se_monad" <= (type) "(syscall_error + 'a) s_monad"
   (type) "'a lf_monad" <= (type) "(lookup_failure + 'a) s_monad"
-  (type) "'a p_monad" <=(type) "(interrupt + 'a) s_monad"
+  (type) "'a p_monad" <=(type) "(unit + 'a) s_monad"
 
 text \<open>Perform non-preemptible operations within preemptible blocks.\<close>
 definition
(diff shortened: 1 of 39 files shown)


spec/abstract/Exceptions_A.thy
```diff
@@ -49,7 +49,7 @@ translations
   (type) "'a f_monad" <= (type) "(fault + 'a) s_monad"
   (type) "'a se_monad" <= (type) "(syscall_error + 'a) s_monad"
   (type) "'a lf_monad" <= (type) "(lookup_failure + 'a) s_monad"
-  (type) "'a p_monad" <=(type) "(interrupt + 'a) s_monad"
+  (type) "'a p_monad" <=(type) "(unit + 'a) s_monad"
```
michaelmcinerney, line 52: I don't really know anything about preemption or about these monads, but doesn't this change remove the ability to associate an irq with an exemption? If we were wanting to increase the flexibility of the preemption monad to allow it to be associated with things other than an irq, then wouldn't we want to introduce a `preemption` type, have `interrupt` as one of the possibilities, and then, say for MCS, allow others as necessary?

mktnk3: Well, so, the point is, we don’t need to be able to associate an irq with a preemption. Or, at least, that is what we have in the C code, as I understand it. In the C code, preemption is just a label that conveys the information that preemption occurred, and that is what the change here is trying to capture.

michaelmcinerney: Ah OK. I'm not sure why we bothered to have the `interrupt` type there in the first place, then. Maybe it's historical

lsf37: It's historical, yes. They way the kernel implemented preemption changed a few times. Initially it followed Haskell in looking up the IRQ number inside the preemption point and passing it out, but then the mechanism changed and it was easier to look it up again outside the preemption point. From that point on, the IRQ number was actually ignored in Haskell, which didn't break anything, so we didn't see immediately that it can/should be removed.

Since MCS now adds a path where there is no IRQ number to associate with the preemption, it makes sense to sync up with C.

lsf37 (approved):
Nice! This is larger than I thought it would be, but thankfully most of the volume is just replacing `intr` with `dc`, which makes sense and confirms that we did indeed ignore the returned IRQ.

lsf37:
The timeout failure in the Github test can be ignored. Could you post a testboard link though, please? Just to make sure.

mktnk3:
https://bamboo.ts.data61.csiro.au/browse/L4V-TESTBOARD-633
https://bamboo.ts.data61.csiro.au/browse/L4V-TESTBOARD-634

X64 Refine had timeout in the first run. So, the second one is a rerun of only X64, which passed.
Would these put together do or should I do another run?

lsf37:
Together they are fine, no need to start a new one. I've upped the time for Refine in the Isabelle2020 PR, so as soon as that one is merged (after this), we shouldn't see timeouts on that any more.

mktnk3:
Ok, then I will merge this soon. This seems to be missing Matt’s most recent commit which changes the test script slightly. So, it should ok to rebase and then just merge.

PR #126, opened 2020-10-25
preemption_point for MCS

mktnk3:
update preemption_point for MCS
    
    - in MCS, preemption_point calls update_time_stamp and check_budget,
    which are only available after Ipc_R
    - this commit introduces a new file InvocationFuns_A, moves the
    definition of preemption_point there
    - also shuffles around the file dependency a little
    - there are some invocation functions that depend on preemption_point;
    they are also moved to InvocationFuns_A

This PR contains only the abstract spec update (so not for merging as is). This is for Michael to take over for the rest of the update, but I would like others to have a look at this point, just to make sure everyone is ok with the change :)

This is done on top of the preemption fix (p_monad fix). It looks like I can’t target this to a branch on my fork... please just look at the last commit.


spec/abstract/InvocationFuns_A.thy
```diff
@@ -0,0 +1,379 @@
...
+ "preemption_point \<equiv> doE liftE $ do_extended_op update_work_units;
+                         OR_choiceE (work_units_limit_reached)
+                           (doE liftE $ do_extended_op reset_work_units;
+                                irq_opt \<leftarrow> liftE $ do_machine_op (getActiveIRQ True);
+                                case_option (do
+                                               update_time_stamp;
+                                               result \<leftarrow> check_budget;
+                                               if (\<not> result)
+                                               then returnOk ()
+                                               else (throwError $ ())
```
michaelmcinerney, line 29: Shouldn't the "then" and "else" branches be swapped?

mktnk3: Oops, yes, I think you are right. Actually, I meant to add the "cur_sc is active" check as well.

michaelmcinerney: Oh, I thought you were wanting me to do that when you said that I would be carrying on this work. Happy for you to do that, of course!

spec/abstract/InvocationFuns_A.thy
```diff
@@ -0,0 +1,385 @@
...
+begin
+
+section \<open>Allow preemption at this point.\<close>
+definition
+  preemption_point :: "(unit,'z::state_ext) p_monad" where
+ "preemption_point \<equiv> doE liftE $ do_extended_op update_work_units;
+                         OR_choiceE (work_units_limit_reached)
+                           (doE liftE $ do_extended_op reset_work_units;
+                                irq_opt \<leftarrow> liftE $ do_machine_op (getActiveIRQ True);
+                                case_option (do
```
michaelmcinerney, line 24: Nitpick: were you intending to put the first line of the `do` block on a new line? I'd prefer it on the same line with the `do` and `od` indented the same number of spaces, but I know there are other thoughts on this

michaelmcinerney (approved):
Cool!

spec/abstract/InvocationFuns_A.thy
mktnk3: Hmm. I think I do like what you are saying, too, but when the block surrounded by `do` and `od` is small, it feels to spread a bit too thinly sometimes, and also when the block is in the branches of `if-then-else`, I would prefer to see the `if-then-else` structure clearly than the `do`-block (also because the `do`-`od` demarcation is redundant in this case here). Something like that is what I was thinking in the background here. But not with any strong opinion, really.

mktnk3: So, this is now ready for you to start the proof :)

PR #127, opened 2020-10-26
rt ainvs: rename valid_replies' to valid_replies_2

corlewis:


MitchellBuckley (approved):
If this is purely a rename (as far as I can see) then this looks good!

mktnk3 (approved):
Thanks, Corey. That was quick :)

corlewis:
> If this is purely a rename (as far as I can see) then this looks good!

Yep, it's just a rename

PR #131, opened 2020-10-30
lib: A tutorial and some 'modify' monad rules for Lib.EquivValid

robs-cse:
Thanks to Toby Murray (@tobycmurray) for early feedback.

Signed-off-by: Robert Sison <robert.sison@unimelb.edu.au>

### DIFF
--- a/lib/EVTutorial/ROOT
+++ b/lib/EVTutorial/ROOT
@@ -1,0 +1,15 @@
+(*
+ * Copyright 2020, The University of Melbourne (ABN 84 002 705 224)
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ *)
+
+chapter Lib
+
+session EVTutorial = Lib +
+  options [document = pdf]
+  theories
+    EquivValidTutorial
+  document_files
+    "root.bib"
+    "root.tex"
--- a/.github/workflows/proof.yml
+++ b/.github/workflows/proof.yml
@@ -51,7 +51,7 @@ jobs:
       with:
         L4V_ARCH: ${{ matrix.arch }}
         isa_branch: ts-2020
-        session: ExecSpec ASpec AInvs
+        session: ExecSpec ASpec AInvs EVTutorial
 
   refine:
     name: Refine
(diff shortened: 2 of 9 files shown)


robs-cse:
The tutorial can be built to PDF by running `./isabelle/bin/isabelle build -d . EVTutorial` from the l4v directory.

lsf37:
Ok to ignore the link checker failures, these are hopefully temporary, and in any case unrelated. Also fine to ignore the `Refine` failures. They're running out of store and I should probably remove them from the test, because that happens too frequently now.

lsf37 (changes requested):
The tutorial itself looks very nice from my side!

One small change request before I can approve: can you please add it to the regression test infrastructure, i.e. to the appropriate `tests.xml` and make sure it is also called in the corresponding github workflow. I think adding it to `.github/workflows/proof.yml `  in the run that calls `ASpec` and `AInvs` would make sense (this one also builds `Lib` as a dependency, IIRC).

lib/EVTutorial/ROOT
```diff
@@ -0,0 +1,15 @@
+(*
+ * Copyright 2020, The University of Melbourne (ABN 84 002 705 224)
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ *)
+
+chapter Lib
+
+session EVTutorial = Lib +
+  options [document = pdf, document_output = "output"]
```
lsf37, line 10: Unless you really want the user to inspect the generated sources, I would usually not add an `output` option.

```suggestion
  options [document = pdf]
```

robs-cse:
The force-push to commit 1ee13dd is just a rebase, with some minor fixes to the last commit message.

mbrcknl (approved):
I've not read every detail of the tutorial, but the bits I looked at were great! I look forward to working through the rest...

Thanks for doing this!

lsf37 (approved):
👍

robs-cse:
(Force-push to commit 3bd44af is another rebase.)

robs-cse:
@lsf37 has advised that the two reviews are sufficient to approve the merge. The force push to commit e2194d4 is a rebase to the current mainline head, squashing all the fixup commits.

lsf37:
Ok to ignore the failed link test. Apparently the nicta.com.au certificates have expired.

robs-cse:
The force push to commit 19312b5 rebased to the mainline head, and passed [this testboard run](https://bamboo.ts.data61.csiro.au/browse/L4V-TESTBOARD-658). I'll "Rebase and Merge" it now - thank you again @lsf37 and @mbrcknl.

PR #132, opened 2020-10-31
Strengthen tcb_inv_wf, prove tcbinv_corres

pierzchalski:


proof/refine/ARM/Tcb_R.thy
```diff
@@ -2346,62 +2346,64 @@ lemma setSchedulerAction_invs'[wp]:
...
-         apply (rule corres_guard_imp [OF writereg_corres], simp+)[1]
-        apply (rule corres_guard_imp [OF readreg_corres], simp+)[1]
-       apply (rule corres_guard_imp [OF copyreg_corres], simp+)[1]
+          apply (rule corres_guard_imp[OF writereg_corres], simp+)[1]
+         apply (rule corres_guard_imp[OF readreg_corres], simp+)[1]
+        apply (rule corres_guard_imp[OF copyreg_corres], simp+)[1]
+       apply (clarsimp simp del: invoke_tcb.simps)
+       apply (rename_tac word a b fault_h time_h croot vroot ipcb sl' fault_h' time_h' croot'
+                         vroot' ipcb')
+       apply (rule_tac F="is_aligned word 5" in corres_req)
```
michaelmcinerney, line 2375: 5. Is there some special significance to this number? If the proof broke later on could we figure out what it was?

michaelmcinerney (approved):

proof/refine/ARM/Tcb_R.thy
pierzchalski: Ah, right. The unsatisfying answer is "that's what the number was previously", but I might dig into it and see where it comes from.

corlewis (approved):
Looks good to me! My only suggestion would be that the commit message would make more sense to me as "Adds the constraint that if _any_ tcb cte field is set, ...". The original message reads to me like it only holds when fault handers are being set, which isn't true.

proof/refine/ARM/Tcb_R.thy
pierzchalski: We will never know the true significance of 5, for it turns out it was unnecessary for the proof.

pierzchalski:
[testboard](https://bamboo.ts.data61.csiro.au/browse/L4V-TESTBOARD-645)

pierzchalski:
[Post-rebase testboard](https://bamboo.ts.data61.csiro.au/browse/L4V-TESTBOARD-647)

proof/refine/ARM/Tcb_R.thy
mktnk3: good that this was sorted in that way.

mktnk3:
> Looks good to me! My only suggestion would be that the commit message would make more sense to me as "Adds the constraint that if _any_ tcb cte field is set, ...". The original message reads to me like it only holds when fault handers are being set, which isn't true.

has this been addressed?

pierzchalski:
> 
> 
> > Looks good to me! My only suggestion would be that the commit message would make more sense to me as "Adds the constraint that if _any_ tcb cte field is set, ...". The original message reads to me like it only holds when fault handers are being set, which isn't true.
> 
> has this been addressed?

I think so? @corlewis does [this wording](https://github.com/seL4/l4v/pull/132/commits/b82a718a07627a815a8f67dd488133c12dd79adf) address your concern?

corlewis:
> 
> 
> > > Looks good to me! My only suggestion would be that the commit message would make more sense to me as "Adds the constraint that if _any_ tcb cte field is set, ...". The original message reads to me like it only holds when fault handers are being set, which isn't true.
> > 
> > 
> > has this been addressed?
> 
> I think so? @corlewis does [this wording](https://github.com/seL4/l4v/pull/132/commits/b82a718a07627a815a8f67dd488133c12dd79adf) address your concern?

Yep!

mktnk3:
great, all good then!

PR #133, opened 2020-11-01
all: remove theory import path references

lsf37:
In Isabelle2020, when isabelle jedit is started without a session
context, e.g. `isabelle jedit -l ASpec`, theory imports with path
references cause the isabelle process to hang.

Since sessions now declare directories, Isabelle can find those files
without path reference and we therefore remove all such path references
from import statements. With this, `jedit` and `build` should work with
and without explicit session context as before.


corlewis (approved):
I'm fine with ignoring the failing tests here, they're unrelated and you had a green testboard run.

mbrcknl (approved):
Impressive. :-)

Thinking about this some more, it seems strange that this change is necessary for the case where a target (`-R`) session is *not* specified, but is not necessary when one *is* specified. In either case, Isabelle needs to use its knowledge of sessions from the in-scope `ROOT` files to figure out which session the current theory is in, and therefore which directories to search for unqualified imports. If anything, not having a target (`-R`) session requires Isabelle to do more work to figure out the session for the current theory.

But regardless of my confusion, it seems clear that import paths don't make much sense any more, and we should do what works.

lsf37:
> I'm fine with ignoring the failing tests here, they're unrelated and you had a green testboard run.

Thanks for spotting the testboard run, I should have linked that. But yes, it was green :-)

lsf37:
>  If anything, not having a target (-R) session requires Isabelle to do more work to figure out the session for the current theory.

Yes it smells like a bug, but I ended up with the same conclusion as you.

PR #139, opened 2020-11-06
Add REPEAT attribute and remove resolve_address_bits'.simps

corlewis:
This cherry-picks two commits from the rt branch. See https://github.com/seL4/l4v/pull/122 for previous comments.

### DIFF
--- a/lib/Repeat_Attribute.thy
+++ b/lib/Repeat_Attribute.thy
@@ -1,0 +1,78 @@
+(*
+ * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ *)
+
+theory Repeat_Attribute
+imports Main
+begin
+
+ML \<open>
+local
+
+val attribute_generic = Context.cases Attrib.attribute_global Attrib.attribute
+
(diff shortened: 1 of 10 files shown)


lsf37 (approved):
👍 Thanks for moving that `[simp del]`.

lib/Repeat_Attribute.thy
```diff
@@ -0,0 +1,78 @@
...
+  \<close>
+  thm tests[REPEAT [OF TrueI]]
+
+  text \<open>
+    @{attribute REPEAT} succesfully terminates when applying an attribute that could loop,
+    such as @{attribute simplified} and @{attribute simp}. Importantly, it still updates
+    the context if necessary, in this case by adding P to the simp set.
+  \<close>
+  lemma
+    assumes P[REPEAT[simplified], REPEAT [simp]]: "P \<or> False"
```
Xaphiosis, line 72: This is really nitpicky, but do you want people putting a space after `REPEAT` or not? It should be consistent in the docs/examples as least.

corlewis: That's a very good point. I'll stay consistent with the preexisting `TRY` and use spaces everywhere.

PR #140, opened 2020-11-09
corres_whileM

michaelmcinerney:


lib/ExtraCorres.thy
```diff
@@ -251,4 +251,117 @@ lemma wp_from_corres_unit:
...
+      apply (fastforce simp: valid_def)
+     apply (clarsimp simp: valid_def)
+    apply wpsimp
+   apply (fastforce intro: no_fail_pre)
+  apply (fastforce intro: wf_subset[OF whileLoop_terminates_wf[where C=C' and B=B']])
+  done
+
+lemma corres_whileLoop:
+  assumes cond: "\<And>r r' s s'. \<lbrakk>rrel r r'; (s, s') \<in> srel\<rbrakk> \<Longrightarrow> C r s = C' r' s'"
+  assumes corres: "\<And>r r'. rrel r r' \<Longrightarrow> corres_underlying srel False nf' rrel P P' (B r) (B' r')"
```
corlewis, line 314: Hopefully it doesn't take too much work, but you should be able to have `C r` and `C' r'` in the preconditions here.

corlewis (approved):
These rules look good to me!

lib/ExtraCorres.thy
michaelmcinerney: Thanks Corey. I should have closed this PR when I made my latest PR. The latest one does include an updated rule for `whileLoop` which has better assumptions and preconditions

PR #141, opened 2020-11-09
Finalise_R reply operation valid lemmas progress review

VictorPhan:
PR for progress review. Main lemma proved so far is `replyPop_list_refs_of_replies`. These are lemmas working towards showing that reply operations (clear, unlink, pop, etc.) will ultimately preserve `invs'`.

### DIFF
--- a/proof/refine/ARM/Finalise_R.thy
+++ b/proof/refine/ARM/Finalise_R.thy
@@ -2403,19 +2403,19 @@ replyPop replyPtr tcbPtr
   apply (rule hoare_seq_ext_skip, solves \<open>wpsimp\<close>, simp?)+
   apply (rule hoare_seq_ext)
    apply (wpsimp wp: cleanReply_list_refs_of_replies')
+  apply (rule hoare_when_cases)
+   apply wpsimp
+   apply (erule delta_sym_refs
+          ; clarsimp simp: isHead_def obj_at'_def projectKOs list_refs_of_reply'_def
+                           list_refs_of_replies'_def
+                    split: reply_next.split_asm if_splits option.splits)
+  apply (rule hoare_seq_ext[OF _ assert_sp])
+  apply (rule hoare_seq_ext_skip, solves \<open>wpsimp simp: comp_def\<close>, simp?)+
+  apply (rule hoare_seq_ext)
+   apply (rule hoare_seq_ext_skip, solves \<open>wpsimp\<close>, simp?)
+   apply (wpsimp wp: hoare_when_cases)
   apply (rule hoare_when_cases[rotated])
-   apply (rule hoare_seq_ext[OF _ assert_sp])
-   apply (rule hoare_seq_ext_skip, solves \<open>wpsimp simp: comp_def\<close>, simp?)+
-   apply (rule hoare_seq_ext)
-    apply wpsimp
-   apply (rule hoare_when_cases)
-    apply (subst if_cancel, clarsimp)
-    apply (erule delta_sym_refs
-           ; clarsimp simp: isHead_def obj_at'_def projectKOs list_refs_of_reply'_def
-                            list_refs_of_replies'_def
-                     split: reply_next.split_asm if_splits option.splits)
-   apply (rule hoare_seq_ext[OF _ get_reply_sp'])
-   apply (wpsimp wp: hoare_vcg_if_lift)
+apply (wpsimp wp: updateReply_list_refs_of_replies'_inv)
    apply (clarsimp simp: isHead_def)
    apply (frule_tac reply=prevReply in ko_at'_replies_of')
    apply (frule (3) sym_refs_replyNext)
@@ -2424,7 +2424,7 @@ replyPop replyPtr tcbPtr
                             list_refs_of_replies'_def  get_refs_def
                      split: if_splits option.splits | rule conjI)+)
   apply (clarsimp simp: obj_at'_def)
-  done
+  sorry
 
 lemma updateReply_list_refs_of_replies':
   assumes "\<And>reply. list_refs_of_reply' (f reply) = k (list_refs_of_reply' reply)"
@@ -2435,12 +2435,6 @@ lemma updateReply_list_refs_of_replies':
   apply (erule subst[rotated])
   apply (rule arg_cong[where f=P, OF all_ext], clarsimp)
   by (clarsimp simp: map_set_def ko_at'_replies_of' opt_map_left_Some)
-
-lemma bar:
-  "list_refs_of_reply' (reply \<lparr> replyPrev := Nothing \<rparr>)
-   = Set.filter (\<lambda>ref. snd ref \<noteq> ReplyPrev) (list_refs_of_reply' reply)"
-  apply (simp add: Set.filter_def list_refs_of_reply'_def get_refs_def split: option.splits)
-  sorry
 
 crunches updateReply, cleanReply
   for valid_irq_node'[wp]: "\<lambda>s. valid_irq_node' (irq_node' s) s"
--- a/proof/refine/ARM/Finalise_R.thy
+++ b/proof/refine/ARM/Finalise_R.thy
@@ -2742,14 +2742,49 @@ lemma replyPop_list_refs_of_replies'[wp]:
 
 lemma replyPop_iflive:
   "\<lbrace>\<lambda>s. if_live_then_nonz_cap' s \<and> valid_objs' s \<and>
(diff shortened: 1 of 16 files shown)


proof/refine/ARM/Finalise_R.thy
```diff
@@ -2329,6 +2329,118 @@ lemma replyUnlink_invs'[wp]:
   unfolding invs'_def valid_state'_def
   by wpsimp
 
+lemma setReply_list_refs_of_replies'[wp]:
```
VictorPhan, line 2332: Ignore this lemma as it wasn't supposed to be added to PR

proof/refine/ARM/Finalise_R.thy
```diff
@@ -2329,6 +2329,118 @@ lemma replyUnlink_invs'[wp]:
...
+  "ko_at' reply replyPtr s \<Longrightarrow> list_refs_of_replies' s replyPtr = list_refs_of_reply' reply"
+  by (clarsimp simp: obj_at'_def projectKOs list_refs_of_replies'_def opt_map_def)
+
+lemma in_list_refs_of_replies'_E:
+  assumes in_list: "(p, tp) \<in> list_refs_of_replies' s replyPtr"
+  assumes nextR: "\<And>reply. tp = ReplyNext \<Longrightarrow> replyNext reply = Some (Next p)
+                            \<Longrightarrow> replies_of' s replyPtr = Some reply \<Longrightarrow> R"
+  assumes prevR: "\<And>reply. tp = ReplyPrev \<Longrightarrow> replyPrev reply = Some p
+                            \<Longrightarrow> replies_of' s replyPtr = Some reply \<Longrightarrow> R"
+  shows "R"
```
corlewis, line 2368: This is a great lemma!
It's very nitpicky, but I would suggest reversing the order of some of the assumptions. It's equivalent, but where possible I like to introduce new variables before using them, to make it easier for humans to read.
```suggestion
lemma in_list_refs_of_replies'_E:
  assumes in_list: "(p, tp) \<in> list_refs_of_replies' s replyPtr"
  assumes nextR: "\<And>reply. tp = ReplyNext \<Longrightarrow> replies_of' s replyPtr = Some reply
                            \<Longrightarrow> replyNext reply = Some (Next p) \<Longrightarrow> R"
  assumes prevR: "\<And>reply. tp = ReplyPrev \<Longrightarrow> replies_of' s replyPtr = Some reply
                            \<Longrightarrow> replyPrev reply = Some p \<Longrightarrow> R"
  shows "R"
```

proof/refine/ARM/Finalise_R.thy
```diff
@@ -2329,6 +2329,118 @@ lemma replyUnlink_invs'[wp]:
...
+         ; clarsimp simp: list_refs_of_reply'_def in_get_refs replyPrev_list_refs_of_replies'
+                   split: if_splits)
+   apply (rule conjI | clarsimp simp: obj_at'_def projectKOs | erule in_list_refs_of_replies'_E)+
+  done
+
+lemma replyUnlink_list_refs_of_replies'[wp]:
+  "\<lbrace>\<lambda>s. P ((list_refs_of_replies' s)(replyPtr := list_refs_of_reply' (upd reply)))
+        \<and> ko_at' reply replyPtr s\<rbrace>
+   updateReply replyPtr upd
+   \<lbrace>\<lambda>rv s. P (list_refs_of_replies' s)\<rbrace>"
```
corlewis, line 2432: I think changing this lemma to something like the following should help you with the issue of `reply` being a schematic.
```suggestion
lemma replyUnlink_list_refs_of_replies'[wp]:
  "\<lbrace>\<lambda>s. \<forall>reply. ko_at' reply replyPtr s \<longrightarrow> P ((list_refs_of_replies' s)(replyPtr := list_refs_of_reply' (upd reply)))\<rbrace>
   updateReply replyPtr upd
   \<lbrace>\<lambda>rv s. P (list_refs_of_replies' s)\<rbrace>"
```

corlewis (approved):
Really good job getting through this `sym_refs` proof!

proof/refine/ARM/Finalise_R.thy
VictorPhan: This makes sense as it also orders the assumptions from general to more specific.

proof/refine/ARM/Finalise_R.thy
VictorPhan: Yep! This is how I want it to look like in the end, but I'm going to work on structuring the lemma like `update_sched_context_valid_sched_pred'` so we have a more general lemma that we can instantiate to this one.

corlewis: Ooh, that would be great!

proof/refine/ARM/Finalise_R.thy
```diff
@@ -2329,12 +2329,162 @@ lemma replyUnlink_invs'[wp]:
...
   by wpsimp
 
+crunches schedContextDonate
+  for replies_of'[wp]: "\<lambda>s. P (replies_of' s)"
+
+lemma isHeadSome:
+ "isHead (Some h) \<longleftrightarrow> (\<exists>p. h = Head p)"
+  by (clarsimp simp: isHead_def split: reply_next.split)
+
+lemma replyPrev_list_refs_of_replies':
```
mktnk3, line 2339: For this one, and some the lemmas below, we probably don’t want to ask for alignment conditions in the assumption. What I mean is,  unlike `state_refs_of’`, `list_refs_of_replies’` does not require `is_aligned` and `ps_clear` for its elements. So, we don’t need to have those in the assumptions. So, in that sense, `ko_at’` is asking too much.

In fact, we already have this lemma in the form of projections. `replyPrevs_of_refs` in StateRelation.thy includes this one in one direction, and we don’t need to know the reply object specifically.

proof/refine/ARM/Finalise_R.thy
```diff
@@ -2329,12 +2329,162 @@ lemma replyUnlink_invs'[wp]:
...
+  apply (clarsimp simp: list_refs_of_replies'_def list_refs_of_reply'_def in_get_refs
+                 split: option.splits)
+  by (auto elim: nextR prevR simp: opt_map_def)
+
+lemma sym_refs_replyNext:
+  "\<lbrakk>sym_refs (list_refs_of_replies' s); replyPrev reply = Some prevReplyPtr;
+    ko_at' (replyTCB_update Map.empty reply) replyPtr s; ko_at' prevReply prevReplyPtr s\<rbrakk>
+     \<Longrightarrow> replyNext prevReply = Some (Next replyPtr)"
+  apply (drule sym_refs_replyNext_replyPrev_proj[where rp=prevReplyPtr and rp'=replyPtr])
+  by (clarsimp simp: opt_map_def obj_at'_def projectKOs split: option.splits)
```
mktnk3, line 2367: This also has extra things in the assumptions. The `replyTCB_update` part is irrelevant to the proof. And we don’t need to have `ko_at’`, either. The lemma is essentially a non-projection version of `sym_refs_replyNext_replyPrev_proj`, but we don’t need both versions. It should be possible to use the projection version instead of this one.

VictorPhan: Fixed, using non-projective version now. Removing this lemma from the PR.

proof/refine/ARM/Finalise_R.thy
```diff
@@ -2329,12 +2329,162 @@ lemma replyUnlink_invs'[wp]:
...
+lemma bar:
+  "list_refs_of_reply' (reply \<lparr> replyPrev := Nothing \<rparr>)
+   = Set.filter (\<lambda>ref. snd ref \<noteq> ReplyPrev) (list_refs_of_reply' reply)"
+  apply (simp add: Set.filter_def list_refs_of_reply'_def get_refs_def split: option.splits)
+  sorry
+
+lemma cleanReply_list_refs_of_replies':
+  "\<lbrace>\<lambda>s. P (map_set (heap_upd k replyPtr (list_refs_of_replies_opt' s)))\<rbrace>
+   cleanReply replyPtr
+   \<lbrace>\<lambda>rv s. P (list_refs_of_replies' s)\<rbrace>"
```
mktnk3, line 2435: Can you prove this lemma using `updateReply_list_refs_of_replies’` above, instead of unfolding `updateReply`"

VictorPhan:
Putting up a new PR with cleanup, and separating valid_dom_schedule' converted to definition changes.

PR #144, opened 2020-11-10
rt: make valid_replies' easier to prove

corlewis:
This updates valid_replies' to only hold for replies with pointers to other replies. Replies connected to an SC are instead handled by an assertion and a cross lemma, using the abstract valid_replies invariant.


corlewis:
@mktnk3 @mbrcknl Does this match what our plans were for the `valid_replies'` invariant?

mbrcknl (approved):
This looks amazing!

mktnk3 (approved):
Very cool!
So you separated out the alignment for the reply object. Nice to see the Haskell assert working well :)

PR #147, opened 2020-11-19
corres_whileLoop and awaken_corres

michaelmcinerney:
This introduces a `corres` rule for `whileLoop` (as found in NonDetMonad.thy).

`awaken` in the abstract spec is defined in terms of `whileLoop`.

`awaken` in the Haskell is defined in terms of a function (perhaps confusingly, also called `whileLoop`) which emulates the `whileLoop` from NonDetMonad.thy. This is then thrown away and the design spec uses a definition directly in terms of the `whileLoop` from NonDetMonad.thy

A proof for `awaken_corres` is included.

There are quite a few lemmas which might be better elsewhere. I'm happy to move things. In particular, `one_branching` might be good to have in `lib`


spec/haskell/src/SEL4/Object/TCB.lhs
```diff
@@ -1064,33 +1064,31 @@ On some architectures, the thread context may include registers that may be modi
...
->         possibleSwitchTo awakened
->         setReprogramTimer True
+>     awakened <- tcbReleaseDequeue
+>     runnable <- isRunnable awakened
+>     assert runnable "the awakened thread must be runnable"
+>     possibleSwitchTo awakened
+
+> awaken :: Kernel ()
+> awaken = do
+>     whileLoop (\r s -> funOfM releaseQNonEmptyAndReady s) (\r -> awaken_body) ()
```
Xaphiosis, line 1091: no need for `do` for one statement

spec/haskell/src/SEL4/Model/StateData.lhs
```diff
@@ -332,6 +332,17 @@ The function "whileM" allows writing while loops where the loop body and the con
...
+The function "whileLoop" emulates the definition whileLoop from NonDetMonad.thy in lib.
+
+> funOfM :: MonadState s m => m a -> s -> a
+> funOfM f s = let (rv,s') = runState f s in rv
+
+> condition :: MonadState s m => (s -> Bool) -> m a -> m a -> m a
+> condition cond left right = get >>= \s -> if (cond s) then left else right
+
+> whileLoop :: Monad m => (a -> b -> bool) -> (a -> m b) -> a -> m b
+> whileLoop cond body r = (condition (cond r) (body r) (return r)) >> whileLoop C B
```
Xaphiosis, line 344: Does this go through Haskell translator? `Monad m` is very generic, I'm interested if this will work...

michaelmcinerney: It's ignored.

Though I've just remembered that I need to test the HaskellKernel, and that fails. Great. I hate Haskell

lib/ExtraCorres.thy
```diff
@@ -251,4 +251,88 @@ lemma wp_from_corres_unit:
...
+  apply (clarsimp simp: pred_conj_def)
+  apply (prop_tac "P t \<and> P' t'")
+   apply (metis body_inv pred_andI use_valid)
+  apply (drule_tac x=t in spec)
+  apply (drule_tac x=rv in spec)
+  apply clarsimp
+  apply (rename_tac rv' t'')
+  apply (rule_tac x=rv' in exI)
+  apply (rule_tac x=t'' in exI)
+  apply (simp add: cond whileLoop_def whileLoop_results.intros)
```
Xaphiosis, line 293: this looks like fastforce or something can take care of this instantiation block and subsequent simp

Xaphiosis:
I approve of the non-MCS bits, but can't really approve of MCS stuff as I have no idea what's good there. I like the improvement to abstraction. Someone should also look over style. @corlewis maybe?

lib/ExtraCorres.thy
michaelmcinerney: A plain old fastforce or something with the rules I've got here doesn't work. sledgehammer is able to solve it a few lines earlier, but uses some rules that are not so pretty. So I think leaving it as is would be the best option

spec/haskell/src/SEL4/Model/StateData.lhs
```diff
@@ -332,6 +332,17 @@ The function "whileM" allows writing while loops where the loop body and the con
...
+
+> funOfM :: MonadState s m => m a -> s -> a
+> funOfM f state = let (rv,state') = runState f state in rv
+
+> condition :: MonadState s m => (s -> Bool) -> m a -> m a -> m a
+> condition cond left right = get >>= \s -> if (cond s) then left else right
+
+> whileLoop :: Monad m => (a -> b -> Bool) -> (a -> m b) -> a -> m b
+> whileLoop cond body r = (condition (cond r) (body r) (return r)) >> whileLoop cond body
+
```
lsf37, line 345: Nicely generic.

spec/design/skel/Thread_H.thy
```diff
@@ -43,6 +43,9 @@ requalify_consts
 
 end
 
+definition awaken_def:
+"awaken \<equiv> whileLoop (\<lambda>_ s. the (fun_of_m releaseQNonEmptyAndReady s)) (\<lambda>_. awaken_body) ()"
```
lsf37, line 47: Can we line up the type of `funOfM` in Haskell with the one in Isabelle such that they match? Then we wouldn't need the manual translation here, but could restrict it to `whileLoop` and `funOfM` only.

(If `funOfM` is used in other places in Haskell as well, a different function could work as well)

spec/design/skel/TCB_H.thy
```diff
@@ -47,7 +47,7 @@ fun tcbEPFindIndex where
 
 declare tcbEPFindIndex.simps[simp del]
 
-#INCLUDE_HASKELL SEL4/Object/TCB.lhs Arch= bodies_only NOT liftFnMaybe assertDerived archThreadGet archThreadSet asUser sanitiseRegister getSanitiseRegisterInfo takeWhileM sort_key tcbEPFindIndex
+#INCLUDE_HASKELL SEL4/Object/TCB.lhs Arch= bodies_only NOT liftFnMaybe assertDerived archThreadGet archThreadSet asUser sanitiseRegister getSanitiseRegisterInfo takeWhileM sort_key tcbEPFindIndex awaken
```
lsf37, line 50: It might be possible to let it include `awaken` as well (see below). The fewer special cases the better..

proof/refine/ARM/Schedule_R.thy
```diff
@@ -3368,5 +3317,840 @@ lemma possible_switch_to_corres:
...
+        apply (rule getTCBRefillReady_corres)
+       apply fastforce
+      apply simp
+     apply wpsimp
+    apply wpsimp
+   apply (clarsimp simp: valid_release_q_def)
+  apply simp
+  done
+
+(* FIXME RT: move? NonDetMonad? *)
```
lsf37, line 3451: Yes, this should go into NonDetMonad (together with the corresponding lemmas, some which might be fine in NonDetMonadLemmas)

proof/refine/ARM/Schedule_R.thy
```diff
@@ -3368,5 +3317,840 @@ lemma possible_switch_to_corres:
...
+  apply (wpsimp wp: get_tcb_refill_ready_no_fail)
+  apply (fastforce simp: valid_release_q_def obj_at_kh_kheap_simps vs_all_heap_simps)
+  done
+
+lemma reprogram_timer_update_release_queue_update_monad_commute:
+  "monad_commute \<top> (modify (reprogram_timer_update (\<lambda>_. b))) (modify (release_queue_update q))"
+  apply (clarsimp simp: monad_commute_def modify_def get_def put_def bind_def return_def)
+  done
+
+(* FIXME RT: move to NonDetMonadLemmaBucket, next to monad_commute_simple?*)
```
lsf37, line 3673: Yes

proof/refine/ARM/Schedule_R.thy
```diff
@@ -3368,5 +3317,840 @@ lemma possible_switch_to_corres:
...
+  apply (clarsimp simp: state_relation_def cdt_relation_def)
+  apply (clarsimp simp: release_queue_relation_def)
+  apply (prop_tac "ksReleaseQueue s' \<noteq> []", fastforce)
+  apply (rule filter_hd_equals_tl)
+   apply (prop_tac "distinct (release_queue s)")
+    apply (clarsimp simp: valid_release_q_def)
+   apply simp+
+  done
+
+(* FIXME RT: move? Is this useful?*)
```
lsf37, line 3707: Do you have any specific use for it? Otherwise it does look a bit too specialised.

proof/refine/ARM/Schedule_R.thy
```diff
@@ -3368,5 +3317,840 @@ lemma possible_switch_to_corres:
...
+             (\<lambda>s'. ksReleaseQueue s' \<noteq> [])"
+  apply (clarsimp simp: cross_rel_def)
+   apply (fastforce dest: state_relation_release_queue_relation
+                  intro!: tcb_at_cross
+                    simp: valid_release_q_def release_queue_relation_def vs_all_heap_simps
+                          obj_at_def is_tcb_def)
+  done
+
+(* FIXME RT: move? *)
+lemma assert_drop:
```
lsf37, line 3829: You'd usually want to either remove Q from the precondition or have it as implication. Putting it in as something to prove does not look useful (creates unnecessary work).

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -3310,14 +3310,14 @@ lemma possible_switch_to_valid_sched_strong:
   by (auto simp: obj_at_kh_kheap_simps valid_sched_valid_sched_except_blocked vs_all_heap_simps
                  in_cur_domain_def etcb_at_def valid_sched_def not_cur_thread_def ct_in_cur_domain_def)
 
-crunch etcb_at[wp]: awaken "etcb_at P t"
```
lsf37, line 3313: I wonder if we could make `awaken`/`whileLoop` crunchable by instantiating the `whileLoop_wp` rule with one invariant predicate for all parts such that it works without further instantiation (and then giving that rule to crunch as `wp`).

Have you tried something like that?

lsf37:
Sorry for taking so long, I hope the PR is still useful. This looks generally very good, just a few small fixup-comments.

lib/ExtraCorres.thy
lsf37: Yes, if fastforce/force don't work, then that's fine from my side.

spec/haskell/src/SEL4/Model/StateData.lhs
lsf37: Ah, yes, that one is also still open. If you're at it, see if you can make the type of `funOfM` conform to the one in Isabelle, then you don't have to exclude `awaken` from the translation.

lsf37: If you need help with that one, let me know!

proof/invariant-abstract/DetSchedSchedule_AI.thy
corlewis: Agreed, that's what I was thinking as well. Something like the following seems like it should work for this.
```
lemma whileLoop_wp':
  "(⋀r. ⦃ λs. I r s ∧ C r s ⦄ B r ⦃ I ⦄) ⟹
  ⦃ I r ⦄ whileLoop C B r ⦃ I ⦄"

File: proof/invariant-abstract/DetSchedSchedule_AI.thy [hunk 12]
```diff
@@ -8893,52 +8818,98 @@ lemma active_sc_tcb_at_budget_sufficient:
...
+    apply (wpsimp simp: valid_release_q_def obj_at_kh_kheap_simps)
+   apply wpsimp
+   apply (fastforce intro: active_sc_tcb_at_budget_sufficient
+                     simp: released_sc_tcb_at_def valid_release_q_def)
+  apply (wpsimp simp: obj_at_kh_kheap_simps)
+  done
+
+lemma fun_of_m_get_tcb_refill_budget_ready_valid_release_q:
+  "\<lbrakk>valid_release_q s; t \<in> set (release_queue s)\<rbrakk>
+   \<Longrightarrow> budget_ready t s = the (fun_of_m (get_tcb_refill_ready t) s)"
```
corlewis, line 8860: It's hard to tell without actually going through the proofs, but my guess would be that this equality is more useful as
```
fun_of_m (get_tcb_refill_ready t) s = Some (budget_ready t s)

File: proof/invariant-abstract/DetSchedSchedule_AI.thy [hunk 13]
```diff
@@ -8893,52 +8818,98 @@ lemma active_sc_tcb_at_budget_sufficient:
...
+  apply (fastforce simp: fun_of_m_get_tcb_refill_budget_ready valid_release_q_def vs_all_heap_simps)
+  done
+
+lemma release_q_non_empty_and_ready_fun_of_m:
+  "\<lbrakk>valid_release_q s; the (fun_of_m release_q_non_empty_and_ready s)\<rbrakk>
+   \<Longrightarrow> release_queue s \<noteq> [] \<and> budget_ready (hd (release_queue s)) s"
```
corlewis, line 8866: Same with this lemma, maybe something like
```
fun_of_m release_q_non_empty_and_ready s = Some (release_queue s ≠ [] ∧ budget_ready (hd (release_queue s)) s)

File: proof/refine/ARM/Schedule_R.thy (hunk 10 above)
Comment (corlewis) [reply proof/refine/ARM/Schedule_R.thy:3829]: Agreed! The default action is to use `assert_wp`, and if the implication is annoying to deal with then you can use `assert_inv` instead.

**Review by corlewis — APPROVED** [review]

**Comment by lsf37** [comment]
I'll have a look at the Haskell failure and push any fixes to this branch.

**Comment by lsf37** [comment]
Ok, so it looks like what we were trying to do with `funOfM` is not possible. I had forgotten that in Haskell the `Kernel` monad stack contains the `IO` monad at the bottom for interacting with the actual machine. There is no escaping from that side-effecting monad back into pure functions: imagine the monad doing a number of machine operations which alter the state of the "external" world. There is no way to execute these and then ignore the side effect, and just return the result of the computation. You're going to have to live with the side effects. In logic, you can easily postulate this of course, but in actual computations you cannot.

I'll think about what we can do instead. Possibly convert `releaseQNonEmptyAndReady` into a pure function, which should be possible (I don't think it changes the state or does machine ops).

**Comment by mktnk3** [comment]
> Ok, so it looks like what we were trying to do with `funOfM` is not possible. I had forgotten that in Haskell the `Kernel` monad stack contains the `IO` monad at the bottom for interacting with the actual machine. There is no escaping from that side-effecting monad back into pure functions: imagine the monad doing a number of machine operations which alter the state of the "external" world. There is no way to execute these and then ignore the side effect, and just return the result of the computation. You're going to have to live with the side effects. In logic, you can easily postulate this of course, but in actual computations you cannot.
> 
> I'll think about what we can do instead. Possibly convert `releaseQNonEmptyAndReady` into a pure function, which should be possible (I don't think it changes the state or does machine ops).

I spent some time thinking about this, too. A "pure" (non-IO monad) version of `releaseQNonEmptyAndReady` would require a pure version of `getTCBRefillReady`, which will ask for the same of `threadGet`, which seemed non-trivial (or do we have such a thing? I didn't think so, but I am not entirely sure). Or is there any other way?

**Comment by lsf37** [comment]
> A "pure" (non-IO monad) version of `releaseQNonEmptyAndReady` would require a pure version of `getTCBRefillReady`, which will ask for the same of `threadGet`, which seemed non-trivial (or do we have such a thing? I didn't think so, but I am not entirely sure). 

Yes, I think that's basically what we'll have to do. It'll be a bit of work, but maybe not too bad. I'm still thinking about whether it would make sense to start a `Reader Maybe Kernel` monad like we have in Isabelle (which we misnomed OptionMonad). I might experiment with that a little, because it would be useful for other places as well. This monad would be restricted to kernel state only (no machine state) in Haskell, because the machine state lives in the IO monad which creates the problem here. The problem with that new monad would be to teach the Haskell translator about it, but maybe that is less problematic than it sounds.

**Comment by lsf37** [comment]
I pushed the parts that fix the types of the old definitions, but haven't started yet on the new ideas, so this will still fail.

**Comment by lsf37** [comment]
The `awaken_body` -> `awakenBody` will break things that referred to the old name, will fix these up when I have the rest.

**Comment by mktnk3** [comment]
> > A "pure" (non-IO monad) version of `releaseQNonEmptyAndReady` would require a pure version of `getTCBRefillReady`, which will ask for the same of `threadGet`, which seemed non-trivial (or do we have such a thing? I didn't think so, but I am not entirely sure).
> 
> Yes, I think that's basically what we'll have to do. It'll be a bit of work, but maybe not too bad. I'm still thinking about whether it would make sense to start a `Reader Maybe Kernel` monad like we have in Isabelle (which we misnomed OptionMonad). I might experiment with that a little, because it would be useful for other places as well. This monad would be restricted to kernel state only (no machine state) in Haskell, because the machine state lives in the IO monad which creates the problem here. The problem with that new monad would be to teach the Haskell translator about it, but maybe that is less problematic than it sounds.

That sounds interesting, except for the Haskell translator part, maybe :thinking:

**Comment by lsf37** [comment]
I've pushed a version with the `Reader` monad, which will break all the proofs for now. I'll experiment with something a bit closer to the Isabelle monad that we have, which is not very idiomatic Haskell, but should make our life in the proofs easier. Let's see how that goes.

File: proof/refine/ARM/Schedule_R.thy [hunk 14]
```diff
@@ -3368,5 +3317,850 @@ lemma possible_switch_to_corres:
...
+lemma isRunnable_no_fail[wp]:
+  "no_fail (tcb_at' tcbPtr) (isRunnable tcbPtr)"
+  apply (wpsimp simp: isRunnable_def)
+  done
+
+lemma awaken_body_corres:
+  "corres dc (pspace_aligned and pspace_distinct and valid_objs and valid_sched_action
+              and valid_tcbs and valid_release_q  and (\<lambda>s. Structures_A.release_queue s \<noteq> []))
+             (valid_objs' and valid_queues and valid_queues' and valid_release_queue_iff)
+             Schedule_A.awaken_body awaken_body"
```
mktnk3, line 3852: It took me a while to realise that the function was renamed on the concrete side but the proofs are left stale. A mystery solved... 😅

PR #148, opened 2020-11-20
rt refine: add missing sc_replies_relation in state_relationD/E

mktnk3:


michaelmcinerney:
Neither of these have anything for `release_queue_relation`. I added it to one of these in my latest PR, but maybe it would be best to add them here

mktnk3:
> Neither of these have anything for `release_queue_relation`. I added it to one of these in my latest PR, but maybe it would be best to add them here

Oh right, that does make sense. I can certainly do that if you don’t mind.

michaelmcinerney (approved):

corlewis (approved):

PR #149, opened 2020-11-21
c-parser: improve asm support in modifies proofs

mbrcknl:
This commit adds support for inline assembly whose `lhs` updates a
global variable (such as the heap).

Prior to this commit, the modifies prover assumed that the `lhs` update
of an `asm_spec` only updated local variables. Specifically, the use of
`asm_store_eq_helper[OF globals.surjective globals.surjective]` as a
rewrite rule assumes that `globals (lhs v s)` simplifies to `globals s`,
exposing the `asm_store` inside `s` to the rewrite rule.

This commit avoids the assumption by using `globals.equality` as an
introduction rule. This produces more subgoals, but the subgoals are
relatively simple, so the perfomance is essentially unchanged.

### DIFF
--- a/tools/c-parser/CProof.thy
+++ b/tools/c-parser/CProof.thy
@@ -422,12 +422,21 @@ lemma asm_specE:
 
 lemmas state_eqE = arg_cong[where f="\<lambda>s. (globals s, state.more s)", elim_format]
 
-lemmas asm_store_eq_helper
-    = arg_cong2[where f="(=)" and a="asm_store f v s"]
-      arg_cong2[where f="(=)" and c="asm_store f v s"] for f v s
-
 definition asm_semantics_ok_to_ignore :: "'a itself \<Rightarrow> bool \<Rightarrow> string \<Rightarrow> bool" where
   "asm_semantics_ok_to_ignore ti volatile specifier
     = (\<forall>xs gl. snd ` asm_semantics specifier xs (gl :: (heap_mem \<times> 'a)) = {gl})"
+
+lemma asm_spec_preserves:
+  assumes spec: "\<And>t v' (gl'::heap_mem \<times> 'a).
+                 \<lbrakk> gdata (asm_store gdata gl' (globals \<sigma>)) = gdata (globals \<sigma>);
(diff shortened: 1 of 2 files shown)


mbrcknl:
I've tested this on my machine. I don't have a testboard run yet, since the testboard script doesn't like the way I've set up my remotes! I'll post a testboard as soon as I figure out a workaround...

lsf37 (approved):
Looks good from my side if all C tests pass.

lsf37:
> I don't have a testboard run yet, since the testboard script doesn't like the way I've set up my remotes!

You may have to `repo sync` or `git pull` in the internal repo that contains the `testboardpush` script.

mbrcknl:
> > I don't have a testboard run yet, since the testboard script doesn't like the way I've set up my remotes!
> 
> You may have to `repo sync` or `git pull` in the internal repo that contains the `testboardpush` script.

The `testboardpush` script doesn't like having the same remote name in different repositories unless they have the same parent.

For example, it's fine if I have a `github` remote in each of `seL4` and `l4v` pointing to `git@github.com:seL4/seL4.git` and `git@github.com:seL4/l4v.git` respectively, because the parent in both cases is `git@github.com:seL4` (or the `ssh://...` encoded version).

But if I also add a `github` remote to `HOL4` pointing to `git@github.com:HOL-Theorem-Prover/HOL.git`, then I get the error:
```
Remote name "github" refers to multiple URLs:
...
```
For now, I'll just rename the offending remotes (e.g. `github-hol4`).

Xaphiosis:
There is no way I would have sorted this out without you in time. Many thanks!

Xaphiosis (approved):

lsf37:
> The `testboardpush` script doesn't like having the same remote name in different repositories unless they have the same parent.

Ah, yes, that is indeed a thing I've come across as well. It mostly expects a `repo` checkout and that wouldn't occur there.

mbrcknl:
[L4V-TESTBOARD-659](https://bamboo.ts.data61.csiro.au/browse/L4V-TESTBOARD-659)

PR #150, opened 2020-11-23
autocorres: fix code inclusion in quickstart doc

mbrcknl:
The AutoCorres quickstart document includes code from `*.c` files from a
given line number, to avoid including license headers and other details
that aren't useful in the document. This updates the line numbers for
the current license headers.


mbrcknl:
[L4V-TESTBOARD-660](https://bamboo.ts.data61.csiro.au/browse/L4V-TESTBOARD-660)

lsf37 (approved):
Thanks for spotting that. Might be worth doing a point release for after your C parser PR is merged.

PR #153, opened 2020-11-23
lib: avoid internal index names

lsf37:
The lemma set `exception_set_finite` contained the members
`exception_set_finite_1` and `exception_set_finite_2`. The `_1`/`_2`
suffix clashes with the internal `(1)` suffix for lemma set references,
which in some code paths is internally represented as `_1`, leading to
an error message.

Curiously this error message only occurs when the proof is run
single-threaded, so it has gone unnoticed for quite some time.

### DIFF
--- a/lib/Lib.thy
+++ b/lib/Lib.thy
@@ -1997,15 +1997,15 @@ lemma dom_eqD:
   "\<lbrakk> f x = Some v; dom f = S \<rbrakk> \<Longrightarrow> x \<in> S"
   by clarsimp
 
-lemma exception_set_finite_1:
+lemma exception_set_finite1:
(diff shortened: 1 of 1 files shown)


lsf37:
This was discovered by a COMP4161 student trying to run AutoCorres in a VM. Searching isabelle-users, I suspect this was also encountered maybe in different from by Norbert.

lsf37:
Testboard should pop yup [here](https://bamboo.ts.data61.csiro.au/browse/L4V-TESTBOARD-661)

mbrcknl (approved):
Wow, that's pretty obscure. :-)

lsf37:
> Wow, that's pretty obscure. :-)

It is, right? And what's more, the student actually solved it (using `--o threads=2`). I was impressed.

Xaphiosis (approved):

PR #154, opened 2020-11-24
rt lib+aspec+ainvs: replace sort_queue

ryybrr:
Replace the implementation of sort_queue with tcb_ep_dequeue and tcb_ep_append to match the Haskell and C implementations. Keeping sort_queue would require us to prove a new invariant instead and would also make the refinement more difficult.

### DIFF
--- a/lib/Lib.thy
+++ b/lib/Lib.thy
@@ -2729,4 +2729,28 @@ text \<open>Prevent clarsimp and others from creating Some from not None by fold
 definition
   "not_None x = (x \<noteq> None)"
 
+primrec findIndex' :: "nat \<Rightarrow> ('a \<Rightarrow> bool) \<Rightarrow> 'a list \<Rightarrow> nat option" where
+  "findIndex' _ _ [] = None"
+| "findIndex' n P (x#xs) = (if P x then Some n else findIndex' (Suc n) P xs)"
+
+definition
+  findIndex :: "('a \<Rightarrow> bool) \<Rightarrow> 'a list \<Rightarrow> nat option" where
+  "findIndex = findIndex' 0"
+
+lemma findIndex'_app:
+  "findIndex' n P (xs @ ys) = (if \<exists>x \<in> set xs. P x then findIndex' n P xs
+                                                   else findIndex' (n + length xs) P ys)"
+  by (induct xs arbitrary: n; clarsimp)
+
+lemma findIndex_member:
+  "tptr \<in> set xs \<Longrightarrow>
+   \<exists>ys zs. xs = ys @ tptr # zs \<and> tptr \<notin> set ys \<and> findIndex (\<lambda>x. x = tptr) xs = Some (length ys)"
+  apply (induct xs; clarsimp)
+  apply (case_tac "tptr=a"; clarsimp)
+   apply (rule_tac x="[]" in exI)
+   apply (clarsimp simp: findIndex_def)
+  apply (rule_tac x="a#ys" in exI)
+  apply (clarsimp simp: findIndex_def findIndex'_app)
+  done
+
 end
--- a/proof/invariant-abstract/DetSchedSchedule_AI.thy
+++ b/proof/invariant-abstract/DetSchedSchedule_AI.thy
@@ -6650,13 +6650,13 @@ lemma set_mcpriority_bound_sc_tcb_at_cur_thread[wp]:
 
 lemma reorder_ntfn_bound_sc_tcb_at_cur_thread[wp]:
   "\<lbrace>\<lambda>s. bound_sc_tcb_at P (cur_thread s) s\<rbrace>
-   reorder_ntfn a
+   reorder_ntfn ntfn_ptr tptr
    \<lbrace>\<lambda>_. \<lambda>s. bound_sc_tcb_at P (cur_thread s) s\<rbrace>"
   by (rule_tac f="cur_thread" in hoare_lift_Pf; wpsimp)
 
 lemma reorder_ep_bound_sc_tcb_at_cur_thread[wp]:
   "\<lbrace>\<lambda>s. bound_sc_tcb_at P (cur_thread s) s\<rbrace>
-   reorder_ep a
+   reorder_ep ep_ptr tptr
    \<lbrace>\<lambda>_. \<lambda>s. bound_sc_tcb_at P (cur_thread s) s\<rbrace>"
   by (rule_tac f="cur_thread" in hoare_lift_Pf; wpsimp)
 
--- a/proof/invariant-abstract/Finalise_AI.thy
+++ b/proof/invariant-abstract/Finalise_AI.thy
@@ -1579,4 +1579,11 @@ crunches test_reschedule, tcb_release_remove
   and obj_at[wp]: "\<lambda>s. P (obj_at Q p s)"
   (wp: crunch_wps simp: crunch_simps)
 
+lemma tcb_ep_find_index_inv[wp]:
+  "tcb_ep_find_index tptr qs curindex \<lbrace>P\<rbrace>"
+  by (induct curindex) (wpsimp | simp add: tcb_ep_find_index.simps)+
+
+crunches tcb_ep_dequeue, tcb_ep_append
+  for inv[wp]: P
+
 end
--- a/proof/invariant-abstract/IpcDet_AI.thy
+++ b/proof/invariant-abstract/IpcDet_AI.thy
@@ -87,112 +87,108 @@ lemma sfi_tcb_at [wp]:
     \<lbrace>\<lambda>_. tcb_at t\<rbrace>"
   by (wpsimp simp: send_fault_ipc_def)
 
-lemma sort_queue_valid_ep_helper:
-  "distinct list \<Longrightarrow> (a, b) \<in> set (zip list' list) \<Longrightarrow> (a', b) \<in> set (zip list' list) \<Longrightarrow> a = a'"
-  apply (induct list arbitrary: list')
-   apply clarsimp
-  apply (clarsimp simp: zip_Cons)
-  apply (erule disjE, fastforce dest!: in_set_zipE)
-  apply (erule disjE, fastforce dest!: in_set_zipE, clarsimp)
+lemma tcb_ep_find_index_wp:
+  "\<lbrace>\<lambda>s. (\<forall>i j. 0 \<le> i \<and> i \<le> Suc sz \<longrightarrow>
+               (\<forall>tcb tcba. ko_at (TCB tcb) tptr s \<and> ko_at (TCB tcba) (queue ! j) s \<longrightarrow>
+                           (Suc j = i \<longrightarrow> tcb_priority tcba \<ge> tcb_priority tcb) \<longrightarrow>
+                           (i < j \<and> j \<le> sz \<longrightarrow> tcb_priority tcba < tcb_priority tcb) \<longrightarrow> Q i s))\<rbrace>
+   tcb_ep_find_index tptr queue sz \<lbrace>Q\<rbrace>"
+  by (induct sz) (wp thread_get_wp' | simp add: tcb_ep_find_index.simps obj_at_def le_Suc_eq)+
+
+lemma tcb_ep_append_valid_SendEP:
+  "\<lbrace>valid_ep (SendEP (t#q)) and K (t \<notin> set q)\<rbrace> tcb_ep_append t q \<lbrace>\<lambda>q'. valid_ep (SendEP q')\<rbrace>"
+  apply (simp only: tcb_ep_append_def)
+  apply (case_tac q; wpsimp wp: tcb_ep_find_index_wp)
+  apply (fastforce simp: valid_ep_def set_take_disj_set_drop_if_distinct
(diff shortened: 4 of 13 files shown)


lib/Lib.thy
```diff
@@ -2729,4 +2729,12 @@ text \<open>Prevent clarsimp and others from creating Some from not None by fold
...
   "not_None x = (x \<noteq> None)"
 
+primrec findIndex' :: "nat \<Rightarrow> ('a \<Rightarrow> bool) \<Rightarrow> 'a list \<Rightarrow> nat option" where
+  "findIndex' _ _ [] = None"
+| "findIndex' n P (x#xs) = (if P x then Some n else findIndex' (Suc n) P xs)"
+
+definition
+  findIndex :: "('a \<Rightarrow> bool) \<Rightarrow> 'a list \<Rightarrow> nat option" where
+  "findIndex = findIndex' 0"
+
```
mktnk3, line 2739: I am not sure if we want to keep camel case for these, now that they are in Lib... @lsf37 ? Any preference?

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -6650,13 +6650,13 @@ lemma set_mcpriority_bound_sc_tcb_at_cur_thread[wp]:
 
 lemma reorder_ntfn_bound_sc_tcb_at_cur_thread[wp]:
   "\<lbrace>\<lambda>s. bound_sc_tcb_at P (cur_thread s) s\<rbrace>
-   reorder_ntfn a
+   reorder_ntfn a b
```
michaelmcinerney, line 6653: Could you use the variables names from the definition of `reorder_ntfn` (and for the other uses of this function and `reorder_ep`)?

proof/invariant-abstract/Finalise_AI.thy
```diff
@@ -1579,4 +1579,11 @@ crunches test_reschedule, tcb_release_remove
   and obj_at[wp]: "\<lambda>s. P (obj_at Q p s)"
   (wp: crunch_wps simp: crunch_simps)
 
+lemma tcb_ep_find_index_inv [wp]:
```
michaelmcinerney, line 1582: Nitpick: not sure if there are any strict rules on this, but almost all attributes immediately follow the name of the lemma

proof/invariant-abstract/IpcDet_AI.thy
```diff
@@ -87,111 +87,84 @@ lemma sfi_tcb_at [wp]:
...
-  "distinct list \<Longrightarrow> (a, b) \<in> set (zip list' list) \<Longrightarrow> (a', b) \<in> set (zip list' list) \<Longrightarrow> a = a'"
-  apply (induct list arbitrary: list')
-   apply clarsimp
-  apply (clarsimp simp: zip_Cons)
-  apply (erule disjE, fastforce dest!: in_set_zipE)
-  apply (erule disjE, fastforce dest!: in_set_zipE, clarsimp)
+lemma findIndex'_app:
+  "findIndex' n P (xs @ ys) = (if \<exists>x \<in> set xs. P x then findIndex' n P xs
+                                                   else findIndex' (n + length xs) P ys)"
+   by (induct xs arbitrary: n; clarsimp)
```
michaelmcinerney, line 93: Extra space

proof/invariant-abstract/IpcDet_AI.thy
```diff
@@ -87,111 +87,84 @@ lemma sfi_tcb_at [wp]:
...
+
+lemma findIndex_member:
+  "tptr \<in> set xs \<Longrightarrow>
+   \<exists>ys zs. xs = ys @ tptr # zs \<and> tptr \<notin> set ys \<and> findIndex (\<lambda>x. x = tptr) xs = Some (length ys)"
+  apply(induct xs; clarsimp)
+  apply (case_tac "tptr=a"; clarsimp)
+   apply(rule_tac x="[]" in exI)
```
michaelmcinerney, line 100: Spaces after `apply`

proof/invariant-abstract/Tcb_AI.thy
```diff
@@ -1654,68 +1650,159 @@ lemma mapM_priorities:
...
-                        if_live_then_nonz_cap_invs live_def)
-  apply (rule conjI; clarsimp simp: obj_at_def distinct_map distinct_zipI2 inj_on_snd_set_zip_map)
-   apply (fastforce dest!: not_idle_tcb_in_SendEp)
-  apply (fastforce dest!: not_idle_tcb_in_RecvEp)
+  "\<lbrace>invs and st_tcb_at (\<lambda>st. ep_blocked st = Some nptr) tptr\<rbrace> reorder_ep nptr tptr \<lbrace>\<lambda>rv. invs\<rbrace>"
+  apply (simp only: reorder_ep_def get_ep_queue_def)
+  apply (wp set_ep_minor_invs hoare_drop_imps tcb_ep_dequeue_rv_wf'
+            tcb_ep_dequeue_valid_ep tcb_ep_dequeue_inv tcb_ep_dequeue_rv_wf'''
+            tcb_ep_append_valid_ep tcb_ep_append_inv tcb_ep_append_rv_wf''')
+    apply (rule monadic_rewrite_refine_valid[OF monadic_rewrite_queue_of,where P''=\<top>,simplified])
```
michaelmcinerney, line 1776: Nitpick: space after the comma

proof/invariant-abstract/Tcb_AI.thy
```diff
@@ -1654,68 +1650,159 @@ lemma mapM_priorities:
...
+  apply (clarsimp cong: conj_cong)
+  apply (rename_tac ep)
+  apply (frule valid_objs_ko_at[rotated], fastforce)
+  apply (clarsimp simp: valid_obj_def valid_ep_def' pred_tcb_at_def obj_at_def cong: conj_cong)
+  apply (subgoal_tac "tptr \<in> set (queue_of ep)")
+   apply (rule context_conjI, clarsimp simp: queue_of_def)
+   apply (clarsimp simp: update_ep_queue_triv)
+   apply (intro conjI allI)
+    apply (erule_tac x="idle_thread s" in ballE)
+     apply (frule_tac p="nptr" in sym_refs_ko_atD[OF _ invs_sym_refs, rotated])
```
michaelmcinerney, line 1787: Shouldn't need the quotation marks aound `nptr`

michaelmcinerney (approved):
Very cool

spec/abstract/IpcCancel_A.thy
```diff
@@ -316,28 +316,45 @@ where
...
+  tcb_ep_find_index :: "obj_ref \<Rightarrow> obj_ref list \<Rightarrow> nat \<Rightarrow> (nat, 'z::state_ext) s_monad"
+where
+  "tcb_ep_find_index tptr qs curindex = do
+     prio \<leftarrow> thread_get tcb_priority tptr;
+     curprio \<leftarrow> thread_get tcb_priority (qs ! curindex);
+     if prio > curprio then
+       if curindex = 0 then return 0
+       else tcb_ep_find_index tptr qs (curindex - 1)
+     else return (curindex + 1)
+   od"
```
mktnk3, line 329: A couple of things about this. One is the naming. This finds an index but quite a different one from the one `findIndex` finds. Maybe this one can have a slightly different name so that no one has a wrong expectation? Maybe something like `tcb_ep_find_prio_index`?

Another thing is, we could do without the first `thread_get` line in the definition, if we pass the `prio` as an argument? It remains constant through this function, so we don’t need to get it every time. We can get the value of `prio` before the call to this function, and pass it as an argument. Then, for the proofs, we’d want to have some projection in the preconditions that says the priority of the tcb at `tptr` is `prio`. Not sure if this helps in any way, though. But also wondering if there is any advantage of doing thread_get every time.

mktnk3 (approved):
Nicely done. Looking forward to seeing the refinement proof :)

spec/abstract/IpcCancel_A.thy
```diff
@@ -316,28 +316,47 @@ where
...
-  "sort_queue qs = do
-     prios \<leftarrow> mapM (thread_get tcb_priority) qs;
-     return $ map snd $ sort_key (\<lambda>x. 255 - (fst x)) (zip prios qs) \<comment> \<open>@{text \<open>0 \<le> priority < 256\<close>}\<close>
+  "tcb_ep_dequeue tptr qs = do
+     index \<leftarrow> return $ the $ findIndex (\<lambda>x. x = tptr) qs;
+     return $ take index qs @ drop (index + 1) qs
    od"
 
+definition
+  tcb_ep_append :: "obj_ref \<Rightarrow> obj_ref list \<Rightarrow> (obj_ref list, 'z::state_ext) s_monad"
```
corlewis, line 342: We should change the name of this function since it's no longer doing an append. Maybe `tcb_ep_insert`? The same change in the C and Haskell would also help.

proof/invariant-abstract/IpcDet_AI.thy
```diff
@@ -87,111 +87,68 @@ lemma sfi_tcb_at [wp]:
...
-   apply clarsimp
-  apply (clarsimp simp: zip_Cons)
-  apply (erule disjE, fastforce dest!: in_set_zipE)
-  apply (erule disjE, fastforce dest!: in_set_zipE, clarsimp)
+lemma tcb_ep_find_index_wp:
+  "\<lbrace>\<lambda>s. (\<forall>i j. 0 \<le> i \<and> i \<le> Suc sz \<longrightarrow>
+               (\<forall>tcb tcba. ko_at (TCB tcb) tptr s \<and> ko_at (TCB tcba) (queue ! j) s \<longrightarrow>
+                           (0 \<le> j \<and> j < i \<longrightarrow> tcb_priority tcba \<ge> tcb_priority tcb) \<longrightarrow>
+                           (i < j \<and> j \<le> sz \<longrightarrow> tcb_priority tcba < tcb_priority tcb) \<longrightarrow> Q i s))\<rbrace>
+   tcb_ep_find_index tptr queue sz \<lbrace>Q\<rbrace>"
```
corlewis, line 95: I'm a bit confused, doesn't this lemma only make sense if you know that the queue was previously sorted? I thought that was the invariant we were trying to avoid?

proof/invariant-abstract/Tcb_AI.thy
```diff
@@ -1654,68 +1650,159 @@ lemma mapM_priorities:
...
+   set_endpoint ptr val
    \<lbrace>\<lambda>rv. invs\<rbrace>"
-  apply (wpsimp simp: invs_def valid_state_def valid_pspace_def
-          wp: valid_irq_node_typ valid_ioports_lift simp_del: fun_upd_apply)
+  apply (wpsimp wp: valid_irq_node_typ valid_ioports_lift
+              simp: invs_def valid_state_def valid_pspace_def)
   apply (clarsimp simp: state_refs_of_def obj_at_def ext elim!: rsubst[where P = sym_refs])
   done
 
+lemma tcb_ep_dequeue_valid_SendEP:
```
corlewis, line 1723: Very minor, but is it worth moving these `tcb_eq_dequeue` lemmas to be with the similar `tcb_ep_append` lemmas in `IpcDet`?

corlewis (approved):
This is some great work! I'm a bit concerned with the definition of `tcb_ep_find_index` but I don't think that blocks this PR, especially since what you have matches the current Haskell.

proof/invariant-abstract/IpcDet_AI.thy
```diff
@@ -87,113 +87,109 @@ lemma sfi_tcb_at [wp]:
...
-  apply (clarsimp simp: zip_Cons)
-  apply (erule disjE, fastforce dest!: in_set_zipE)
-  apply (erule disjE, fastforce dest!: in_set_zipE, clarsimp)
+lemma tcb_ep_find_index_wp:
+  "\<lbrace>\<lambda>s. (\<forall>i j. 0 \<le> i \<and> i \<le> Suc sz \<longrightarrow>
+               (\<forall>tcb tcba. ko_at (TCB tcb) tptr s \<and> ko_at (TCB tcba) (queue ! j) s \<longrightarrow>
+                           (Suc j = i \<longrightarrow> tcb_priority tcba \<ge> tcb_priority tcb) \<longrightarrow>
+                           (i < j \<and> j \<le> sz \<longrightarrow> tcb_priority tcba < tcb_priority tcb) \<longrightarrow> Q i s))\<rbrace>
+   tcb_ep_find_index tptr queue sz \<lbrace>Q\<rbrace>"
+  by (induct sz) (wp thread_get_wp' | simp add: tcb_ep_find_index.simps obj_at_def le_Suc_eq)+
```
MitchellBuckley, line 96: nitpick: it's nicer to use `wpsimp wp: foo simp: bar` instead of `(wp foo | simp add: bar)+` when possible.

MitchellBuckley (approved):
Looks like Corey and Miki have already looked closely. I had a quick look over for style and think that it looks good!

proof/invariant-abstract/IpcDet_AI.thy
ryybrr: I did try that, not possible unfortunately.

PR #155, opened 2020-11-25
regression: increase CRefine timeout

mbrcknl:
CRefine takes longer with Isabelle2020, so this gives more headroom.

Yes, the "proof performance" label is meant to be ironic. :-)


lsf37 (approved):
Sad, but necessary..

Xaphiosis (approved):

LukeMondy (approved):

PR #156, opened 2020-11-25
github: remove Refine sessions from github CI

lsf37:
With Isabelle2020 the Refine sessions are too close to memory
boundaries on github runners, the tests randomly fail with out-of-store
exceptions in polyml (but also randomly succeed without change).

Removing the session here until we have a better solution.


lsf37:
Refine still works fine (and is tested on github) on the `rt` branch, which currently is still on Isabelle2019.

mbrcknl (approved):
:(

PR #160, opened 2020-11-30
rt sr_inv_ul

mktnk3:
In symbolic execution on the concrete side, the state sometimes changes but in such a way that the state relation is still preserved. We encounter this situation quite often in the MCS verification.

There is existing corres lemmas that cater for this situation, namely, `corres_noop` (corres with `return ()` on the abstract side, which is rarely used so far), but only in combination with something like `corres_split_noop_rhs` (which inserts `return` on the abstract side). Since we have many situations that we want to do this, it is nicer to write a rule that does these steps as one symbolic execution.

Also, in `corres_noop`, the state relation preservation condition is integrated in a Hoare triple together with normal preconditions and postconditions. It is often the case that there are existing wp rules that we can use for these normal pre- and post-conditions, while we do need to show that the state relation is preserved, and they do not interact (although we do have some situations that they do interact). `sr_inv_ul` is for packaging up the proof for the preservation of state relation so that we can reuse it for such cases.

The rule `corres_symb_exec_r_sr` is a symbolic execution rule for the concrete side where the concrete state is modified, and defined in terms of `sr_inv_ul`.

So, this PR contains:
- `sr_inv_ul` and related basic lemmas
- `corres_symb_exec_r_sr`
- proof of `replyNext_Next_update_sr_inv`: an example of `sr_inv`

(one thing: I am not too sure is the naming of `sr_inv_ul`...)

### DIFF
--- a/lib/Corres_UL.thy
+++ b/lib/Corres_UL.thy
@@ -1519,18 +1519,15 @@ section \<open> Preservation of state_relation for the concrete side symbolic ex
 
 (** sr_inv **)
 
-definition sr_inv_ul :: "('a \<times> 'b) set \<Rightarrow> ('a \<Rightarrow> bool) \<Rightarrow> ('b \<Rightarrow> bool) \<Rightarrow> ('b \<Rightarrow> (unit \<times> 'b) set \<times> 'c) \<Rightarrow> bool" where
+definition
+  sr_inv_ul :: "('a \<times> 'b) set \<Rightarrow> ('a \<Rightarrow> bool) \<Rightarrow> ('b \<Rightarrow> bool) \<Rightarrow> ('b \<Rightarrow> (unit \<times> 'b) set \<times> 'c) \<Rightarrow> bool" where
   "sr_inv_ul sr P P' f \<equiv>
-       \<forall>s s' t'. (s, s') \<in> sr \<longrightarrow> P s \<longrightarrow> P' s'
-              \<longrightarrow> ((), t') \<in> fst (f s') \<longrightarrow> (s, t') \<in> sr"
+       \<forall>s s' t' x. (s, s') \<in> sr \<longrightarrow> P s \<longrightarrow> P' s'
+              \<longrightarrow> (x, t') \<in> fst (f s') \<longrightarrow> (s, t') \<in> sr"
 
 lemma corres_noop_assm: (* check equivalence *)
   "(\<forall>s. P s \<longrightarrow> \<lbrace>P' and (\<lambda>s'. (s, s') \<in> sr)\<rbrace> m \<lbrace>\<lambda>_ s'. (s, s') \<in> sr\<rbrace>) \<longleftrightarrow> sr_inv_ul sr P P' m"
   by (fastforce simp: valid_def sr_inv_ul_def)
-
-lemma sr_inv_ul_imp:
-  "\<lbrakk>sr_inv_ul sr Q Q' f; \<And>s. P s \<Longrightarrow> Q s; \<And>s. P' s \<Longrightarrow> Q' s\<rbrakk> \<Longrightarrow> sr_inv_ul sr P P' f "
-  by (clarsimp simp: sr_inv_ul_def)
 
 lemma sr_inv_stronger_imp:
   "\<lbrakk>sr_inv_ul sr Q Q' f;
@@ -1539,12 +1536,8 @@ lemma sr_inv_stronger_imp:
    \<Longrightarrow> sr_inv_ul sr P P' f"
   by (clarsimp simp: sr_inv_ul_def)
 
-lemma sr_inv_ul_cross:
-  "\<lbrakk>sr_inv_ul sr P (P' and Q') f; \<And>s s'. P s \<Longrightarrow> P' s' \<Longrightarrow> Q' s'\<rbrakk> \<Longrightarrow> sr_inv_ul sr P P' f "
-  by (clarsimp simp: sr_inv_stronger_imp)
-
-lemma sr_inv_ul_cross_str:
-  "\<lbrakk>sr_inv_ul sr P (P' and Q') f; \<And>s s'. (s, s') \<in> sr \<Longrightarrow> P s \<Longrightarrow> P' s' \<Longrightarrow> Q' s'\<rbrakk> \<Longrightarrow> sr_inv_ul sr P P' f "
+lemma sr_inv_ul_imp:
+  "\<lbrakk>sr_inv_ul sr Q Q' f; \<And>s. P s \<Longrightarrow> Q s; \<And>s. P' s \<Longrightarrow> Q' s\<rbrakk> \<Longrightarrow> sr_inv_ul sr P P' f "
   by (clarsimp simp: sr_inv_stronger_imp)
 
 lemma return_sr_inv[simp]:
@@ -1552,7 +1545,7 @@ lemma return_sr_inv[simp]:
   by (clarsimp simp: sr_inv_ul_def return_def)
 
 lemma sr_inv_ul_bind:
-  "\<lbrakk> sr_inv_ul sr P P' g; \<lbrace>P'\<rbrace> g \<lbrace> \<lambda>rv. Q' rv\<rbrace>; \<And>rv. sr_inv_ul sr P (Q' rv) (f rv) \<rbrakk>
+  "\<lbrakk> \<And>rv. sr_inv_ul sr P (Q' rv) (f rv); sr_inv_ul sr P P' g; \<lbrace>P'\<rbrace> g \<lbrace> \<lambda>rv. Q' rv\<rbrace> \<rbrakk>
    \<Longrightarrow> sr_inv_ul sr P P' (g >>= f)"
   apply (unfold sr_inv_ul_def valid_def)
   apply clarify
@@ -1562,7 +1555,6 @@ lemma sr_inv_ul_bind:
   apply clarify
   apply (drule_tac x=s and y=s' in spec2)
   apply (drule_tac x=s'' in spec)
-  apply clarify
   apply (drule_tac x=x in meta_spec)
   apply (drule_tac x=s and y=s'' in spec2)
   apply (drule_tac x=t' in spec)
@@ -1570,17 +1562,15 @@ lemma sr_inv_ul_bind:
   apply simp
   done
 
-lemma corres_noop_sr_inv:
-  assumes P: "sr_inv_ul sr P P' f"
+lemma corres_noop_sr:
+  assumes sr: "sr_inv_ul sr P P' f"
+  assumes ex: "\<lbrace> P' \<rbrace> f \<lbrace>\<lambda>rv _. r x rv\<rbrace>"
   assumes nf': "\<And>s. \<lbrakk> P s; nf' \<rbrakk> \<Longrightarrow> no_fail (\<lambda>s'. (s, s') \<in> sr \<and> P' s') f"
-  assumes r: "r () x"
   shows "corres_underlying sr nf nf' r P P' (return x) f"
-  using P r
-  apply (simp add: corres_underlying_def return_def split_def)
-  apply (clarsimp simp: sr_inv_ul_def)
-  apply (drule_tac x=a and y=b in spec2, clarsimp)
-  apply (insert nf')
-  apply (clarsimp simp: valid_def no_fail_def)
+  using sr ex
+  apply (clarsimp simp: sr_inv_ul_def corres_underlying_def return_def)
+  apply (clarsimp simp: no_failD[OF nf'[simplified]] valid_def)
+  apply (drule_tac x=b in spec, clarsimp)
   done
 
 (* corres_symb_exec_r_sr basically combines the two lemmas corres_split_noop_rhs and corres_noop
@@ -1589,7 +1579,7 @@ lemma corres_symb_exec_r_sr:
   assumes z: "corres_underlying sr nf nf' r P Q' x y"
   assumes sr: "sr_inv_ul sr P P' m"
   assumes ex: "\<lbrace> P' \<rbrace> m \<lbrace>\<lambda>_. Q'\<rbrace>"
-  assumes nf: "nf' \<Longrightarrow> no_fail P' m"
+  assumes nf: "\<And>s. \<lbrakk> P s; nf' \<rbrakk> \<Longrightarrow> no_fail (\<lambda>s'. (s, s') \<in> sr \<and> P' s') m"
   shows "corres_underlying sr nf nf' r P P' x (m >>= (\<lambda>_. y))"
   using sr z ex nf
   unfolding corres_underlying_def sr_inv_ul_def
@@ -1604,24 +1594,13 @@ lemma corres_symb_exec_r_sr:
    apply clarsimp
    apply (drule_tac x="(rv', s'')" in bspec, simp)
    apply simp
-  apply (drule_tac x=s and y=s' in spec2, clarsimp simp: no_failD valid_def)
+  apply (drule_tac x=s and y=s' in spec2, clarsimp simp: no_failD[OF nf] valid_def)
   apply (rename_tac t')
-  apply (drule_tac x=t' in spec, clarsimp)
+  apply (drule_tac x=s in meta_spec)
   apply (drule_tac x=s' in spec, clarsimp)
   apply (drule bspec[of "fst _"], simp)
   apply clarsimp
-  apply ( drule_tac x="(s, t')" in bspec, simp)
+  apply (drule_tac x="(s, t')" in bspec, simp)
   by clarsimp
-
-lemma corres_noop_sr:
-  assumes sr: "sr_inv_ul sr P P' f"
-  assumes ex: "\<lbrace> P' \<rbrace> f \<lbrace>\<lambda>rv _. r x rv\<rbrace>"
-  assumes nf': "\<And>s. \<lbrakk> P s; nf' \<rbrakk> \<Longrightarrow> no_fail (\<lambda>s'. (s, s') \<in> sr \<and> P' s') f"
-  shows "corres_underlying sr nf nf' r P P' (return x) f"
-  using sr ex
-  apply (clarsimp simp: sr_inv_ul_def corres_underlying_def return_def)
-  apply (clarsimp simp: no_failD[OF nf'[simplified]] valid_def)
-  apply (drule_tac x=b in spec, clarsimp)
-  done
 
 end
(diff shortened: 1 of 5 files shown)


lib/Corres_UL.thy
```diff
@@ -1515,4 +1515,113 @@ lemma corres_assert_opt_assume_l:
...
+lemma sr_inv_ul_cross:
+  "\<lbrakk>sr_inv_ul sr P (P' and Q') f; \<And>s s'. P s \<Longrightarrow> P' s' \<Longrightarrow> Q' s'\<rbrakk> \<Longrightarrow> sr_inv_ul sr P P' f "
+  by (clarsimp simp: sr_inv_stronger_imp)
+
+lemma sr_inv_ul_cross_str:
+  "\<lbrakk>sr_inv_ul sr P (P' and Q') f; \<And>s s'. (s, s') \<in> sr \<Longrightarrow> P s \<Longrightarrow> P' s' \<Longrightarrow> Q' s'\<rbrakk> \<Longrightarrow> sr_inv_ul sr P P' f "
+  by (clarsimp simp: sr_inv_stronger_imp)
+
+lemma return_sr_inv[simp]:
+  "sr_inv_ul sr P P' (return x)"
```
MitchellBuckley, line 1551: This confuses me. Based on the definition of `sr_inv_ul` I thought that it could only work with non-returning functions. Am I mistaken that `(z,t') \in fst (f s')` means `z` is the return value of `f s'`? If I'm correct, then `sr_inv_ul` will be trivially true for any function which returns anything other than `()`.

This is not a problem but it might become misleading since it would then be true for functions which do disrupt the state relation, but also return some value.

lib/Corres_UL.thy
```diff
@@ -1515,4 +1515,113 @@ lemma corres_assert_opt_assume_l:
...
+  assumes nf: "nf' \<Longrightarrow> no_fail P' m"
+  shows "corres_underlying sr nf nf' r P P' x (m >>= (\<lambda>_. y))"
+  using sr z ex nf
+  unfolding corres_underlying_def sr_inv_ul_def
+  apply (clarsimp simp: corres_underlying_def bind_def)
+  apply (rename_tac s s')
+  apply (rule conjI; clarsimp simp: in_monad)
+   apply (drule_tac x=s and y=s' in spec2, clarsimp)
+   apply (rename_tac t' rv' s'')
+   apply (drule_tac x="(s, t')" in bspec, simp)
```
MitchellBuckley, line 1601: Based on indentation, you could replace ` apply (X, simp) apply clarsimp` with `apply (X; clarsimp)`

proof/refine/ARM/Reply_R.thy
```diff
@@ -574,4 +574,77 @@ lemma replyRemoveTCB_invs':
...
+
+lemma sc_replies_relation_replyNext_update:
+  "\<lbrakk>sc_replies_relation s s'; ko_at' reply' rp s'\<rbrakk>
+     \<Longrightarrow> sc_replies_relation s(s'\<lparr>ksPSpace := (ksPSpace s')(rp \<mapsto>
+                                           KOReply (reply' \<lparr> replyNext := v \<rparr>))\<rparr>)"
+  by (clarsimp simp: scs_of'_replyNext_update[simplified]
+                     replyPrevs_of_replyNext_update[simplified])
+
+(* an example of an sr_inv lemma; to be used in reply_remove_tcb_corres *)
+lemma replyNext_Next_update_sr_inv:
```
MitchellBuckley, line 627: This is nice. I've had to use the older method to resolve this issue recently. Your new approach looks better.

lib/Corres_UL.thy
```diff
@@ -1515,4 +1515,113 @@ lemma corres_assert_opt_assume_l:
...
+  apply simp
+  done
+
+lemma corres_noop_sr_inv:
+  assumes P: "sr_inv_ul sr P P' f"
+  assumes nf': "\<And>s. \<lbrakk> P s; nf' \<rbrakk> \<Longrightarrow> no_fail (\<lambda>s'. (s, s') \<in> sr \<and> P' s') f"
+  assumes r: "r () x"
+  shows "corres_underlying sr nf nf' r P P' (return x) f"
+  using P r
+  apply (simp add: corres_underlying_def return_def split_def)
```
MitchellBuckley, line 1579: Combine simp and clarsimp?

MitchellBuckley (approved):
I think that if the issue with non-null returning functions can be corrected or explained to me, then this looks great. It looks like quite a good general approach that could help with a variety of other problems where the state-relation is really non-trivial (not a bijection).

lib/Corres_UL.thy
```diff
@@ -1515,4 +1515,113 @@ lemma corres_assert_opt_assume_l:
   \<Longrightarrow> corres_underlying sr nf nf' dc (P and K (X \<noteq> None)) Q (assert_opt X >>= f) g"
   by (force simp: corres_underlying_def assert_opt_def return_def bind_def fail_def)
 
+section \<open> Preservation of state_relation for the concrete side symbolic execution \<close>
+
+(** sr_inv **)
+
+definition sr_inv_ul :: "('a \<times> 'b) set \<Rightarrow> ('a \<Rightarrow> bool) \<Rightarrow> ('b \<Rightarrow> bool) \<Rightarrow> ('b \<Rightarrow> (unit \<times> 'b) set \<times> 'c) \<Rightarrow> bool" where
+  "sr_inv_ul sr P P' f \<equiv>
+       \<forall>s s' t'. (s, s') \<in> sr \<longrightarrow> P s \<longrightarrow> P' s'
```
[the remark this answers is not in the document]
mktnk3: The implication binds to the right, as you know. What are you suggesting to do here?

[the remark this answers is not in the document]
mktnk3: Can you say a bit more what you have in your mind?

lib/Corres_UL.thy
```diff
@@ -1515,4 +1515,113 @@ lemma corres_assert_opt_assume_l:
...
+
+lemma corres_noop_assm: (* check equivalence *)
+  "(\<forall>s. P s \<longrightarrow> \<lbrace>P' and (\<lambda>s'. (s, s') \<in> sr)\<rbrace> m \<lbrace>\<lambda>_ s'. (s, s') \<in> sr\<rbrace>) \<longleftrightarrow> sr_inv_ul sr P P' m"
+  by (fastforce simp: valid_def sr_inv_ul_def)
+
+lemma sr_inv_ul_imp:
+  "\<lbrakk>sr_inv_ul sr Q Q' f; \<And>s. P s \<Longrightarrow> Q s; \<And>s. P' s \<Longrightarrow> Q' s\<rbrakk> \<Longrightarrow> sr_inv_ul sr P P' f "
+  by (clarsimp simp: sr_inv_ul_def)
+
+lemma sr_inv_stronger_imp:
```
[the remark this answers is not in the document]
mktnk3: I’d say these are for convenience, just like various corres lemmas of slightly different versions, or equivalent rewriting, etc.

lib/Corres_UL.thy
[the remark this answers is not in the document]
MitchellBuckley: Now that I think about, the right-associativity means that this really is a different statement. So just ignore my previous point.

lib/Corres_UL.thy
MitchellBuckley: As I think about this some more. It could you replace `((), t') ∈ fst (f s')` with `(x, t') ∈ fst (f s')` where x is quantified alongside `t'`?

lib/Corres_UL.thy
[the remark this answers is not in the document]
mktnk3: I don’t see how it relates immediately.

lib/Corres_UL.thy
MitchellBuckley: Hmmm, or maybe `t' \in snd \` fst (f s')`

mktnk3:
I also have a bunch more of `sr_inv` lemmas around `cleanReply` and `updateReply` with prev/next updates of prev/next pointers, although some of them could do with some cleanup. If any of these seems useful, let me know.

lib/Corres_UL.thy
corlewis: To make things confusing in a slightly different way, I think `sr_inv_ul` is defined to only apply to monadic functions with a unit return type. That means that `x` in this lemma is actually of the unit type, and hence equal to `()`.

lib/Corres_UL.thy
```diff
@@ -1515,4 +1515,113 @@ lemma corres_assert_opt_assume_l:
...
+  apply (clarsimp simp: valid_def no_fail_def)
+  done
+
+(* corres_symb_exec_r_sr basically combines the two lemmas corres_split_noop_rhs and corres_noop
+   in one and separates out the sr_inv part *)
+lemma corres_symb_exec_r_sr:
+  assumes z: "corres_underlying sr nf nf' r P Q' x y"
+  assumes sr: "sr_inv_ul sr P P' m"
+  assumes ex: "\<lbrace> P' \<rbrace> m \<lbrace>\<lambda>_. Q'\<rbrace>"
+  assumes nf: "nf' \<Longrightarrow> no_fail P' m"
```
corlewis, line 1592: nitpick: I don't know if it would ever be helpful, but you could do the same trick here that you did in the previous lemma and assume
```
  nf: "⋀s. ⟦ P s; nf' ⟧ ⟹ no_fail (λs'. (s, s') ∈ sr ∧ P' s') m"

File: proof/refine/ARM/Reply_R.thy [hunk 8]
```diff
@@ -574,4 +574,77 @@ lemma replyRemoveTCB_invs':
               simp: cteCaps_of_def)
   done
 
+context begin interpretation Arch .
+
+lemma pspace_relation_reply_update_conc_only:
+  "\<lbrakk> pspace_relation s s'; s x = Some (Structures_A.Reply reply); s' x = Some (KOReply reply');
+     reply_relation reply new\<rbrakk>
+       \<Longrightarrow> pspace_relation s (s'(x \<mapsto> (KOReply new)))"
```
corlewis, line 582: Maybe use `ps` instead of `s` in this lemma? It can be a bit confusing otherwise, since `s` normally refers to a complete state.

Also, it's very minor but the final line is normally indented less than this. The other lemmas following this also have similarly inconsistent indentation

proof/refine/ARM/Reply_R.thy
```diff
@@ -574,4 +574,77 @@ lemma replyRemoveTCB_invs':
...
+      scs_of' (s'\<lparr>ksPSpace := ksPSpace s'(rp \<mapsto>
+                            KOReply (reply' \<lparr> replyNext := v \<rparr>))\<rparr>) = scs_of' s'"
+  apply (clarsimp simp: obj_at'_def projectKOs isNext_def
+                 split: option.split_asm reply_next.split_asm)
+  by (fastforce simp: projectKO_opt_sc opt_map_def)
+
+lemma scs_of'_replyPrev_update:
+  "ko_at' reply' rp s' \<Longrightarrow>
+      scs_of' (s'\<lparr>ksPSpace := ksPSpace s'(rp \<mapsto>
+                            KOReply (reply' \<lparr> replyPrev := v \<rparr>))\<rparr>) = scs_of' s'"
```
corlewis, line 614: I think it should be possible to combine this lemma with the previous one into something more generic. `scs_of'` should be maintained no matter how the reply is modified. My guess is that something like the following would work.
```
  "reply_at' rp s' ⟹ scs_of' (s'⦇ksPSpace := ksPSpace s'(rp ↦KOReply reply)⦈) = scs_of' s'"

**Review by corlewis — APPROVED** [review]
This is a very nice idea, and some great work! I've just got some minor comments and suggestions.

File: lib/Corres_UL.thy [hunk 10]
```diff
@@ -1515,4 +1515,113 @@ lemma corres_assert_opt_assume_l:
   \<Longrightarrow> corres_underlying sr nf nf' dc (P and K (X \<noteq> None)) Q (assert_opt X >>= f) g"
   by (force simp: corres_underlying_def assert_opt_def return_def bind_def fail_def)
 
+section \<open> Preservation of state_relation for the concrete side symbolic execution \<close>
+
+(** sr_inv **)
+
+definition sr_inv_ul :: "('a \<times> 'b) set \<Rightarrow> ('a \<Rightarrow> bool) \<Rightarrow> ('b \<Rightarrow> bool) \<Rightarrow> ('b \<Rightarrow> (unit \<times> 'b) set \<times> 'c) \<Rightarrow> bool" where
```
michaelmcinerney, line 1522: Nitpick: move the name and the type declaration to a new line, with the `where` on the next line too? It spills over a bit as it is

michaelmcinerney (approved):
I'm guessing most of the time we would want to use this machinery, we would have a function that would return just `()`. But I'm curious whether there is a way to generalise this to functions which return an arbitrary type. I had a quick look and the first few lemmas work just fine, but the corres rule break in ways that I didn't really investigate. 

This does look cool. Looking forward to working on something which could use it!

lib/Corres_UL.thy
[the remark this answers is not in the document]
mktnk3: So, we have `corres_guard_imp` and `stronger_corres_guard_imp` (which `sr_inv_stronger_imp` is based on), but we almost always use the first unless it is required to use the other one, I think. But the stronger one is used to prove the first. So, I can do that, surely, but I think we'd want to keep the simpler one as the main thing to use.

lib/Corres_UL.thy
mktnk3: Hmm. This is something left from the initial exploration, and I think it is ok to delete. (sorry...)

proof/refine/ARM/Reply_R.thy
mktnk3: > Maybe use `ps` instead of `s` in this lemma? It can be a bit confusing otherwise, since `s` normally refers to a complete state.

Ah, yes, you are right! I think I typed `sc_replies_relation` a bit too often :D

mktnk3: Will fix the indentation, too.

proof/refine/ARM/Reply_R.thy
mktnk3: Yes, that's right, what you suggest definitely works. Maybe it was too much copy-pasting :D

lib/Corres_UL.thy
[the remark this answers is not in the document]
mktnk3: So, I will prove the simpler one using the stronger one, as both of you suggest, but for the rest and also for the comment below for the `_cross` lemmas, I am not sure.

For the corres rules, we use `corres_guard_imp`, `corres_cross_over_guard`, `corres_cross_add_guard`, and the like, and all are more or less immediate consequences of `stronger_corres_guard_imp`. And, each of these is used on specific situations that it caters for, and we currently use `stronger_corres_guard_imp` itself very rarely. Do you think we should give a shortest name to `stronger_corres_guard_imp` and use it for everything?

I wouldn't think so...because it is rare enough that we actually need the stronger version to complete the proof, and also because the corres subgoals are generally quite complex, we can make use of simpler versions, when we can, which is quite often, to have simpler and clearer proof structure. And that is what we have been doing, I believe.

For `sr_inv`, in particular, it is often possible to extract information from the state relation and the abstract side at the corres level before getting to the symbolic execution. So, it is also likely that the simpler version suffices most of the time. In the proof of `reply_remove_tcb_corres`, there is one subcase that I squeezed all the case distinctions inside one symbolic execution step, and that is the only instance that I needed the stronger version `sr_inv_stronger_imp`.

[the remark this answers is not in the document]
mktnk3: Having said that, I imagine `sr_inv` goals are generally smaller than the corres ones, and we won't have so many use cases of `sr_inv` anyway, so it probably doesn't' really make much difference either way.

lib/Corres_UL.thy
mktnk3: Well, actually, I do need this. And, as Mitch suggests, the return type in `sr_inv_ul` doesn't need to be of the unit type, it can be any universally quantified x, but it will be just ignored anyway.

proof/refine/ARM/Reply_R.thy
mktnk3: I ended up not using this one in my proof for the main corres rule. So, it looks like this lemma will be discarded. Do you think you can make use of it?

lib/Corres_UL.thy
mktnk3: This `simp` is for the set membership assumption of `bspec`. It cannot be a `clarsimp` (due to something to do with unification, I think).

lib/Corres_UL.thy
mktnk3: I am not sure if it helps or not, but it is possible and it didn't affect the proof at all, so why not maybe?
I think `Q'` here will be fairly large most of the time and we'd be wanting to use `no_fail_pre` anyway.

lib/Corres_UL.thy
mktnk3: yes!

mktnk3:
Hmm. I did not mean to close this PR. So, apparently, on GitHub, if you target a PR to a branch of another PR, the PR gets closed when the target PR is merged. This seems a bit inconvenient, because, on bitbucket, it would be retargeted to the (main) branch that the target PR was merged into, and then you could continue with merging the remaining PR to the main branch.

Here it just says there are unmerged commits, and suggests to delete them, and won't even manually retarget the PR... :sweat_smile:

lsf37:
Not sure if that is it, but what might be happening is that merging a PR deletes the PR branch (unless you explicitly change that). If another PR was targeted to that branch, it that deletion might close it.

lsf37:
If you still have the branch in your local copy, re-pushing it to github might enable reopening the PR.

mktnk3:
> If you still have the branch in your local copy, re-pushing it to github might enable reopening the PR.

Ok, I’ll try that.

mktnk3:
> > If you still have the branch in your local copy, re-pushing it to github might enable reopening the PR.
> 
> Ok, I’ll try that.

It seems that the PR is not linked to the branch anymore. I’ve done a force-push (after a rebase) but it doesn’t show up here.

lsf37:
Hm, that is annoying. I guess we'll have to open another PR manually. Do you want me to do that or are you fine doing that yourself?

mktnk3:
That’s ok, no need for you to act. I might open another PR or maybe just include it in the main `reply_remove_tcb_corres` PR.

PR #161, opened 2020-12-02
docs: factor out dependency specs

lsf37:
There is another (out-of-date) dependency description for l4v on the
docs site. To avoid this duplication, this commit factors out the
dependency part of the README, so that it be included directly on the
docs site without going stale.

Also, the README was getting way too long.


lsf37:
(apologies for the force-push, just fixed a link)

Xaphiosis (approved):

PR #165, opened 2020-12-09
Improve `sts_invs_minor'`

MitchellBuckley:
Previously, sym_refs' was inside invs' and so setThreadState needed to be careful about modifying references coming from tcb thread states. Now that sym_refs' has been removed, it is easier to reason about how setThreadState interacts with invs'.

This is a small improvement that should make the lemma much more useable.


corlewis (approved):
You're right, that should make it much easier to use. Great job spotting it!

michaelmcinerney (approved):
Nice!

mktnk3 (approved):
Nice work! Thanks, Mitch!

PR #166, opened 2020-12-09
arm-hyp: proof updates for seL4 c381c7e14c

lsf37:
seL4 commit seL4/seL4@c381c7e14c (PR seL4/seL4#289) changes cache flushing behaviour for the
verified ARM_HYP configuration. This commit adjusts accordingly.

### DIFF
--- a/proof/invariant-abstract/ARM_HYP/Machine_AI.thy
+++ b/proof/invariant-abstract/ARM_HYP/Machine_AI.thy
@@ -274,7 +274,7 @@ lemma no_fail_cleanCacheRange_RAM[simp, wp]:
 
 lemma no_fail_invalidateCacheRange_I[simp, wp]:
   "no_fail \<top> (invalidateCacheRange_I s e p)"
-  by (simp add: invalidateCacheRange_I_def, wp no_fail_invalidateByVA_I)
+  by (simp add: invalidateCacheRange_I_def, wp no_fail_invalidate_I_PoU)
 
 lemma no_fail_invalidateCacheRange_RAM[simp, wp]:
   "no_fail \<top> (invalidateCacheRange_RAM s e p)"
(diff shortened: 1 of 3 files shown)


lsf37:
Test board running [here](https://bamboo.ts.data61.csiro.au/browse/L4V-TESTBOARD-671)

proof/invariant-abstract/ARM_HYP/Machine_AI.thy
```diff
@@ -274,7 +274,7 @@ lemma no_fail_cleanCacheRange_RAM[simp, wp]:
 
 lemma no_fail_invalidateCacheRange_I[simp, wp]:
   "no_fail \<top> (invalidateCacheRange_I s e p)"
-  by (simp add: invalidateCacheRange_I_def, wp no_fail_invalidateByVA_I)
+  by (simp add: invalidateCacheRange_I_def, wp no_fail_invalidate_I_PoU)
```
corlewis, line 277: Nothing to do with this PR, but we're pretty inconsistent with marking the lemmas in this file with `[simp]` and `[wp]`. If that was more consistent then this would have been an even simpler change.

corlewis (approved):
Looks good to me!

proof/invariant-abstract/ARM_HYP/Machine_AI.thy
lsf37: True, that would've made it even smaller.

Xaphiosis:
Looks good. Did you take a look at the Haskell to see if that makes any sense as well?

Xaphiosis (approved):

lsf37:
> Looks good. Did you take a look at the Haskell to see if that makes any sense as well?

It's behind the machine interface for Haskell, and it looks like Haskell ARM_HYP never actually implemented what was in C and abstract for those machine ops. I left it the way it was since it's ignored in any case.

PR #168, opened 2020-12-10
Remove setReplyTCB and getReplyTCB

MitchellBuckley:
Because setReplyTCB is only called in 4 places, and it is just a wrapper for updateReply, having it as a separate definition means writing double the lemmas, or just unfolding the definition in many places. It is simpler just to remove the function altogether. This should streamline a few proofs and reduce duplicated work.

I have also done the same with getReplyTCB.

### DIFF
--- a/proof/refine/ARM/Finalise_R.thy
+++ b/proof/refine/ARM/Finalise_R.thy
@@ -2273,9 +2273,6 @@ lemma unbindMaybeNotification_invs[wp]:
           elim!: obj_atE' if_live_then_nonz_capE'
           split: option.splits ntfn.splits)
 
-crunches getReplyTCB
-  for inv[wp]: P
-
 lemma setNotification_invs':
   "\<lbrace>invs'
     and (\<lambda>s. live_ntfn' ntfn \<longrightarrow> ex_nonz_cap_to' ntfnPtr s)
@@ -2343,7 +2340,7 @@ lemma replyClear_invs'[wp]:
 lemma finaliseCap_True_invs'[wp]:
   "\<lbrace>invs' and sch_act_simple\<rbrace> finaliseCap cap final True \<lbrace>\<lambda>rv. invs'\<rbrace>"
   unfolding finaliseCap_def sym_refs_asrt_def
-  apply (wpsimp wp: irqs_masked_lift getReplyTCB_wp simp: Let_def split_del: if_split)
+  apply (wpsimp wp: irqs_masked_lift simp: Let_def split_del: if_split)
   apply clarsimp
   apply (subgoal_tac "ex_nonz_cap_to' (ksIdleThread s) s")
    apply (fastforce simp: invs'_def valid_state'_def global'_no_ex_cap)
@@ -2965,8 +2962,8 @@ lemma replyUnlink_obj_at_tcb_none:
   "\<lbrace>K (rptr' = rptr)\<rbrace>
    replyUnlink rptr tptr
    \<lbrace>\<lambda>_. obj_at' (\<lambda>reply. replyTCB reply = None) rptr'\<rbrace>"
-  apply (simp add: replyUnlink_def setReplyTCB_def)
-  apply (wpsimp wp: updateReply_wp_all gts_wp' getReplyTCB_wp)
+  apply (simp add: replyUnlink_def)
+  apply (wpsimp wp: updateReply_wp_all gts_wp')
   by (auto simp: obj_at'_def projectKOs objBitsKO_def)
 
 lemma replyUnlink_makes_unlive:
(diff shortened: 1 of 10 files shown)


proof/refine/ARM/Finalise_R.thy
```diff
@@ -3225,52 +3222,50 @@ lemma (in delete_one_conc_pre) finaliseCap_replaceable:
...
-    (* ThreadCap *)
-    apply (frule capAligned_capUntypedPtr [OF valid_capAligned], simp)
-    apply (clarsimp simp: valid_cap'_def)
-    apply (drule valid_globals_cte_wpD'[rotated], clarsimp)
-    apply (clarsimp simp: invs'_def valid_state'_def valid_pspace'_def)
-   (* ArchObjectCap *)
-   apply (clarsimp simp: obj_at'_def)
-  (* ReplyCap *)
-  apply clarsimp
+       (* ThreadCap *)
```
corlewis, line 3260: nitpick: I prefer the original indentation of this comment, where it matches the line after it

proof/refine/ARM/Finalise_R.thy
```diff
@@ -3225,52 +3222,50 @@ lemma (in delete_one_conc_pre) finaliseCap_replaceable:
...
+     apply (frule capAligned_capUntypedPtr [OF valid_capAligned], simp)
+     apply (clarsimp simp: valid_cap'_def)
+     apply (drule valid_globals_cte_wpD'[rotated], clarsimp)
+     apply (fastforce simp: invs'_def valid_state'_def valid_pspace'_def)
+    apply fastforce
```
corlewis, line 3265: very minor nitpick: the indentation suggests that this `fastforce` is for a new case, and to match the previous style of this proof could use a comment saying which case that is.

proof/refine/ARM/Ipc_R.thy
```diff
@@ -4203,7 +4265,7 @@ lemma replyPush_valid_queues:
    replyPush callerPtr calleePtr replyPtr canDonate
    \<lbrace>\<lambda>_. valid_queues\<rbrace>"
   by (rule monadic_rewrite_refine_valid[OF monadic_rewrite_replyPush, where P''=\<top>, simplified])
-     (wpsimp wp: weak_if_wp schedContextDonate_valid_queues hoare_drop_imps simp: getReplyTCB_def)
+     (wpsimp wp: weak_if_wp schedContextDonate_valid_queues hoare_drop_imps simp:)
```
corlewis, line 4268: nitpick: remove the `simp:`

corlewis (approved):
Nice! I think it would be possible to keep the terms as an abbreviation if we would want that, but otherwise this looks great to me!

I do have to admit that I didn't really look at the changes to the `replyPush` proofs, since they'll all change again with your other PR.

spec/haskell/src/SEL4/Object/Reply.lhs
```diff
@@ -187,9 +187,3 @@ on the thread state of the replyTCB of the replyPtr
 
 > setReply :: PPtr Reply -> Reply -> Kernel ()
 > setReply rptr r = setObject rptr r
-
-> getReplyTCB :: PPtr Reply -> Kernel (Maybe (PPtr TCB))
-> getReplyTCB r = liftM replyTCB (getReply r)
-
-> setReplyTCB :: Maybe (PPtr TCB) -> PPtr Reply -> Kernel ()
-> setReplyTCB tptrOpt rptr = updateReply rptr (\reply -> reply { replyTCB = tptrOpt })
```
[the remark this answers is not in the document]
mktnk3: What this makes me think is that there is not much point in defining `setReplyTCB` via `updateReply` to start with. It has an extra getting of the reply object that is just discarded.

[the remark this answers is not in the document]
mktnk3: Well, it seems that is not much of a point either. `setReplyTCB` was defined in that way originally. I can't decide if keeping abbreviation would be good or not.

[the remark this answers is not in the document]
mktnk3: Apologies for rambling. I think it is ok to proceed as in this PR, without abbreviations. There are only a few occurrences of this, and we don't have such things for `replyPrev` and `replyNext`, which we use more. What do you think, Mitch?

mktnk3 (approved):

spec/haskell/src/SEL4/Object/Reply.lhs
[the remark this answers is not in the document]
MitchellBuckley: I think it's best to continue without abbreviations. The abbreviations are nice, but sometimes they hide useful information.

proof/refine/ARM/Reply_R.thy
```diff
@@ -13,33 +13,25 @@ defs replyUnlink_assertion_def:
...
                           \<or> (\<exists>ep d. state = BlockedOnReceive ep d (Some replyPtr))"
 
-crunches setReplyTCB
-  for pred_tcb_at'[wp]: "\<lambda>s. P (pred_tcb_at' proj test t s)"
-  and tcb_at'[wp]: "\<lambda>s. P (tcb_at' t s)"
-  and ksReadyQueues[wp]: "\<lambda>s. P (ksReadyQueues s)"
-  and ksSchedulerAction[wp]: "\<lambda>s. P (ksSchedulerAction s)"
-  and valid_queues[wp]: "valid_queues"
-  and reply_at'[wp]: "\<lambda>s. P (reply_at' rp s)"
-
```
mktnk3, line 23: This used to generate a bunch more of `updateReply` lemmas that are not included in the merged version below. My proof uses one of those and now I want to revive these crunches for `updateReply`. I am just wondering if other people might experience the same while rebasing.

MitchellBuckley: Sure thing. Go for it! Just make a Pull Request?

mktnk3: Yes, the PR is up now :)

PR #169, opened 2020-12-10
Prove schedContextDonate_corres

ryybrr:
I updated `schedContextDonate` to align with the Abstract and C specs. I then proved `schedContextDonate_corres`.

### DIFF
--- a/proof/refine/ARM/Finalise_R.thy
+++ b/proof/refine/ARM/Finalise_R.thy
@@ -2902,13 +2902,21 @@ crunch invs[wp]: prepareThreadDelete "invs'"
 
 end
 
+lemma weak_sch_act_wf_updates[simp]:
+  "weak_sch_act_wf sa (s\<lparr>ksReprogramTimer := a\<rparr>) = weak_sch_act_wf sa s"
+  "weak_sch_act_wf sa (s\<lparr>ksReleaseQueue := b\<rparr>) = weak_sch_act_wf sa s"
+  by (auto simp: weak_sch_act_wf_def tcb_in_cur_domain'_def)
+
+crunches tcbReleaseRemove
+  for weak_sch_act_wf[wp]: "\<lambda>s. weak_sch_act_wf (ksSchedulerAction s) s"
+
 lemma schedContextDonate_weak_sch_act_wf[wp]:
   "schedContextDonate scPtr tcbPtr \<lbrace>\<lambda>s. weak_sch_act_wf (ksSchedulerAction s) s\<rbrace>"
   apply (simp only: schedContextDonate_def)
   apply (wpsimp wp: threadSet_weak_sch_act_wf setSchedContext_weak_sch_act_wf rescheduleRequired_weak_sch_act_wf)
-        apply (rule_tac Q="\<lambda>_ s. weak_sch_act_wf (ksSchedulerAction s) s" in hoare_strengthen_post[rotated], fastforce)
-        apply (wpsimp wp: threadSet_weak_sch_act_wf)
-       apply wpsimp+
+         apply (rule_tac Q="\<lambda>_ s. weak_sch_act_wf (ksSchedulerAction s) s" in hoare_strengthen_post[rotated], fastforce)
+         apply (wpsimp wp: threadSet_weak_sch_act_wf)
+        apply wpsimp+
   done
 
 lemma ntfnSc_sym_refsD:
@@ -3531,6 +3539,8 @@ lemma schedContextDonate_valid_queues:
     apply (wpsimp wp: tcbSchedDequeue_valid_queues)
     apply (fastforce intro: valid_objs'_maxDomain valid_objs'_maxPriority)
    apply (rule hoare_seq_ext_skip)
+    apply (wpsimp wp: tcbReleaseRemove_valid_queues)
+   apply (rule hoare_seq_ext_skip)
     apply (wpsimp wp: threadSet_valid_queues_new threadSet_valid_objs')
     apply (clarsimp simp: obj_at'_def inQ_def valid_tcb'_def tcb_cte_cases_def)
    apply (wpsimp wp: rescheduleRequired_valid_queues)
@@ -3578,6 +3588,14 @@ lemma rescheduleRequired_valid_sched_context'[wp]:
 global_interpretation schedContextDonate: typ_at_all_props' "schedContextDonate scPtr tcbPtr"
   by typ_at_props'
 
+lemma valid_sched_context'_updates[simp]:
+  "valid_sched_context' sc' (s\<lparr>ksReprogramTimer := a\<rparr>) = valid_sched_context' sc' s"
+  "valid_sched_context' sc' (s\<lparr>ksReleaseQueue := b\<rparr>) = valid_sched_context' sc' s"
+  by (auto simp: valid_sched_context'_def valid_bound_obj'_def split: option.splits)
+
+crunches tcbReleaseRemove
+  for valid_sched_context'[wp]: "\<lambda>s. valid_sched_context' sc' s"
+
 lemma schedContextDonate_valid_objs':
   "\<lbrace>valid_objs' and tcb_at' tcbPtr\<rbrace>
    schedContextDonate scPtr tcbPtr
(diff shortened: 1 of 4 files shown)


proof/refine/ARM/TcbAcc_R.thy
```diff
@@ -2413,15 +2426,21 @@ lemma rescheduleRequired_corres:
...
+   apply (clarsimp simp: no_fail_def return_def vs_all_heap_simps
+                         obj_at_def pred_tcb_at_def weaker_valid_sched_action_def)
   apply (corressimp corres: tcbSchedEnqueue_corres
                       simp: obj_at_def is_tcb_def weak_sch_act_wf_def)
   done
 
+lemma rescheduleRequired_corres:
+  "corres dc (valid_tcbs and weak_valid_sched_action and pspace_aligned and pspace_distinct)
+             (valid_tcbs' and valid_queues and valid_queues' and valid_release_queue_iff)
+             (reschedule_required) rescheduleRequired"
```
MitchellBuckley, line 2440: nitpick: unneeded parentheses

proof/refine/ARM/TcbAcc_R.thy
```diff
@@ -2413,15 +2426,21 @@ lemma rescheduleRequired_corres:
...
-
+    apply (clarsimp simp: exs_valid_def return_def vs_all_heap_simps
+                          obj_at_def pred_tcb_at_def weaker_valid_sched_action_def)
+   apply (clarsimp simp: no_fail_def return_def vs_all_heap_simps
+                         obj_at_def pred_tcb_at_def weaker_valid_sched_action_def)
   apply (corressimp corres: tcbSchedEnqueue_corres
                       simp: obj_at_def is_tcb_def weak_sch_act_wf_def)
   done
 
+lemma rescheduleRequired_corres:
```
MitchellBuckley, line 2437: Nice! I'm surprised this didn't already exist.

proof/refine/ARM/TcbAcc_R.thy
```diff
@@ -2367,8 +2367,20 @@ lemma get_sc_refill_ready_sp:
            \<and> P s\<rbrace>"
   by (wpsimp simp: obj_at_def)
 
-lemma rescheduleRequired_corres:
-  "corres dc (valid_tcbs and weak_valid_sched_action and pspace_aligned and pspace_distinct)
+definition weaker_valid_sched_action where
```
MitchellBuckley, line 2370: If this is really needed, consider moving it somewhere like `Invariants_H.thy`

proof/refine/ARM/Ipc_R.thy
```diff
@@ -3049,13 +3049,15 @@ lemma thread_set_valid_tcb[wp]:
   by (wpsimp wp: get_object_wp simp: thread_set_def)
 
 lemma thread_set_valid_tcbs[wp]:
-  "\<lbrace>valid_tcbs and tcb_at tptr and
```
MitchellBuckley, line 3052: Nice :+1:

MitchellBuckley (approved):
This is a really neat peice of work!

Can you explain why we need `weaker_valid_sched_action`? It's good to make sure that new definitions have a definite reason for existing, otherwise we really end up cluttering the code and making things a bit harder to understand in the long run.

proof/refine/ARM/TcbAcc_R.thy
ryybrr: It actually did, but I needed to weaken the precondition because `weak_valid_sched_action` can't be propagated backwards over the `thread_set` instance in `sched_context_donate`. Hence, I've introduced a weaker variant of the corres rule to use in this particular case and moved the proofs to that rule to avoid boilerplating.

proof/refine/ARM/TcbAcc_R.thy
ryybrr: It's needed, but only for this proof. Does anyone else have an opinion on this?

MitchellBuckley: If it's only needed for local proofs, then I think it's ok to leave the definition here. Thanks for clarifying :+1:

corlewis (approved):

proof/refine/ARM/Finalise_R.thy
```diff
@@ -2902,13 +2902,21 @@ crunch invs[wp]: prepareThreadDelete "invs'"
 
 end
 
+lemma weak_sch_act_wf_updates[simp]:
+  "weak_sch_act_wf sa (s\<lparr>ksReprogramTimer := a\<rparr>) = weak_sch_act_wf sa s"
+  "weak_sch_act_wf sa (s\<lparr>ksReleaseQueue := b\<rparr>) = weak_sch_act_wf sa s"
+  by (auto simp: weak_sch_act_wf_def tcb_in_cur_domain'_def)
+
+crunches tcbReleaseRemove
+  for weak_sch_act_wf[wp]: "\<lambda>s. weak_sch_act_wf (ksSchedulerAction s) s"
```
mktnk3, line 2911: would it be possible to combine some of these crunches together? I imagine they are placed as they are needed, so it might not be possible, but if we can have a bigger crunches block where possible, that would be easier to maintain.

proof/refine/ARM/TcbAcc_R.thy
mktnk3: Maybe a line of comment explaining what this is would be good to have :)

mktnk3 (approved):
Nice work! Glad that we could catch the missed spec update.

proof/refine/ARM/Finalise_R.thy
```diff
@@ -3530,6 +3538,8 @@ lemma schedContextDonate_valid_queues:
    apply (rule_tac B="\<lambda>_. ?pre" in hoare_seq_ext[rotated])
     apply (wpsimp wp: tcbSchedDequeue_valid_queues)
     apply (fastforce intro: valid_objs'_maxDomain valid_objs'_maxPriority)
+   apply (rule hoare_seq_ext_skip)
```
michaelmcinerney, line 3541: Slowly converting you to forward reasoning, I see

proof/refine/ARM/Finalise_R.thy
```diff
@@ -3831,13 +3855,30 @@ lemma tcbSchedDequeue_valid_inQ_queues:
   done
 
 lemma threadSet_tcbSchedContext_update_valid_inQ_queues[wp]:
-  "\<lbrace>valid_inQ_queues and tcb_at' t\<rbrace>
+  "\<lbrace>valid_inQ_queues\<rbrace>
```
michaelmcinerney, line 3858: Cool. And maybe now that it's "invariant", could you use that notation?

proof/refine/ARM/Ipc_R.thy
```diff
@@ -3840,6 +3841,15 @@ crunches schedContextDonate
      wp: threadSet_not_inQ hoare_vcg_imp_lift' valid_irq_node_lift valid_irq_handlers_lift''
          setQueue_cur untyped_ranges_zero_lift threadSet_ifunsafe'T threadSet_cur crunch_wps)
 
+lemma tcbReleaseRemove_list_refs_of_replies'[wp]:
+  "tcbReleaseRemove tcbPtr \<lbrace>\<lambda>s. P (list_refs_of_replies' s)\<rbrace>"
+  by (wpsimp simp: tcbReleaseRemove_def)
+
+lemma schedContextDonate_list_refs_of_replies' [wp]:
+"schedContextDonate scPtr tcbPtr \<lbrace>\<lambda>s. P (list_refs_of_replies' s)\<rbrace>"
```
michaelmcinerney, line 3849: Indentation

proof/refine/ARM/TcbAcc_R.thy
```diff
@@ -2367,10 +2367,22 @@ lemma get_sc_refill_ready_sp:
...
+  "weaker_valid_sched_action s \<equiv>
+   \<forall>t. scheduler_action s = switch_thread t \<longrightarrow>
+       tcb_at t s \<and> (bound_sc_tcb_at ((\<noteq>) None) t s \<longrightarrow> released_sc_tcb_at t s)"
+
+lemma weak_valid_sched_action_strg:
+  "weak_valid_sched_action s \<longrightarrow> weaker_valid_sched_action s"
+  by (fastforce simp: weak_valid_sched_action_def weaker_valid_sched_action_def
```
michaelmcinerney, line 2377: Maybe `vs_all_heap_simps` or `obj_at_kh_kheap_simps` or both would do the trick here, or at least reduce the number of things you have to give to simp

michaelmcinerney (approved):
Very nice, as always

proof/refine/ARM/Finalise_R.thy
ryybrr: They are a little bit strategic, it's difficult to combine them together without things breaking.

mktnk3: I see. Yes, that is very much likely. Thanks for having a try anyway.

PR #170, opened 2020-12-11
rt refine: make more use of reply projections

corlewis:
This was factored out of my ongoing `valid_replies'` work so that it could be used in other PRs. It updates various lemmas that are connected to `valid_replies'` to use projections instead of `obj_at'`.

### DIFF
--- a/proof/refine/ARM/Invariants_H.thy
+++ b/proof/refine/ARM/Invariants_H.thy
@@ -210,6 +210,12 @@ lemma getHeadScPtr_None_iff:
   "(getHeadScPtr x) = None \<longleftrightarrow> (\<forall>rn. x \<noteq> Some (Head rn))"
   by (cases x; clarsimp simp: getHeadScPtr_def split: reply_next.split)
 
+lemma replyNext_None_iff:
+  "replyNext r = None \<longleftrightarrow> replyNext_of r = None \<and> replySc r = None"
+  apply (cases "replyNext r"; clarsimp)
+  apply (case_tac a; clarsimp)
+  done
+
 lemma getReplyNextPtr_Head_None[simp]:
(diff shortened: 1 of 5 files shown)


michaelmcinerney (approved):

proof/refine/ARM/Invariants_H.thy
```diff
@@ -1010,15 +1016,18 @@ definition "vs_valid_duplicates' \<equiv> \<lambda>h.
               x && ~~ mask (vs_ptr_align ko) = y && ~~ mask (vs_ptr_align ko) \<longrightarrow>
               h x = h y"
 
+abbreviation
+  "in_reply_queue rptr s \<equiv> replyNexts_of s rptr \<noteq> None \<or> replyPrevs_of s rptr \<noteq> None"
+
```
mktnk3, line 1021: I am happy to have this abbreviation, but I feel that the name might be somewhat misleading.
This is meant to cover the case when there is no SC linked to the queue, right? I.e., the case when the reply at `rptr` is not in the "reply stack", which we sometimes call also as "reply queue". 
Since this statement actually looks only at the adjacent replies, how about something like `is_reply_linked`? (of course, you can then ask linked to what, but...) Or maybe `in_reply_path` as in it is in a `heap_path`.

mktnk3 (approved):
Nice!

proof/refine/ARM/Invariants_H.thy
mktnk3: then again, even a stand along reply can be in a `heap_path`. Hmm.

corlewis: Good point with the name. The intention is to cover the case where the reply at `rptr` is linked to other replies, regardless of whether there is an `SC` eventually linked to. I think with that in mind I'm happy to use `is_reply_linked`.

PR #171, opened 2020-12-14
rt refine: weaken some simple_ko assumptions

mktnk3:
This is something little that I noticed and needed to tweak for my main proof.

- we don't need to know specific ko to prove exs_valid (=) lemmas for
  simple_ko; using exsitential instead
- and for no_fail lemmas, it is enough to have ntfn_at/ep_at/reply_at


michaelmcinerney (approved):

VictorPhan (approved):

mktnk3:
Merged as part of #176

PR #146, opened 2020-11-12
ReplyPush_corres

MitchellBuckley:
Leaving aside a sorried corres lemma for `donate_sched_context`, this proves `ReplyPush_corres`.

Ideally I would prove that sorried lemma, but I think it's more important to periodically converge.

PR #172, opened 2020-12-14
Shorten replyPush proofs

ryybrr:


corlewis:
How does this interact with the changes to `replyPush` in https://github.com/seL4/l4v/pull/146?

MitchellBuckley:
The changes to `replyPush` in #146 will need to be rebased over #168 anyway. So it's not clear. There will be some slight inefficiencies here where some new proofs will be immediately adjusted, but I think that this approach is the right one anyway: solving one problem at a time to avoid confusion.

PR #173, opened 2020-12-15
rt refine: resurrect some crunches for updateReply

mktnk3:

### DIFF
--- a/proof/refine/ARM/Reply_R.thy
+++ b/proof/refine/ARM/Reply_R.thy
@@ -15,11 +15,12 @@ defs replyUnlink_assertion_def:
 
 
 crunches updateReply
-  for tcb_at'[wp]: "\<lambda>s. P (tcb_at' t s)"
-  and st_tcb_at'[wp]: "\<lambda>s. P (st_tcb_at' P' t' s)"
-  and ksReadyQueues[wp]: "\<lambda>s. P (ksReadyQueues s p) t"
+  for pred_tcb_at'[wp]: "\<lambda>s. P (pred_tcb_at' proj test t s)"
+  and tcb_at'[wp]: "\<lambda>s. P (tcb_at' t s)"
+  and ksReadyQueues[wp]: "\<lambda>s. P (ksReadyQueues s)"
   and ksSchedulerAction[wp]: "\<lambda>s. P (ksSchedulerAction s)"
   and valid_queues[wp]: "valid_queues"
+  and reply_at'[wp]: "\<lambda>s. P (reply_at' rp s)"
 
 lemma replyTCB_update_reply_projs[wp]:
   "\<lbrace>\<lambda>s. P (replyNexts_of s) (replyPrevs_of s)


proof/refine/ARM/Reply_R.thy
```diff
@@ -14,9 +14,12 @@ defs replyUnlink_assertion_def:
                           \<or> (\<exists>ep d. state = BlockedOnReceive ep d (Some replyPtr))"
 
 crunches updateReply
-  for tcb_at'[wp]: "\<lambda>s. P (tcb_at' t s)"
-  and st_tcb_at'[wp]: "\<lambda>s. P (st_tcb_at' P' t' s)"
-  and ksReadyQueues[wp]: "\<lambda>s. P (ksReadyQueues s p) t"
+  for pred_tcb_at'[wp]: "\<lambda>s. P (pred_tcb_at' proj test t s)"
+  and tcb_at'[wp]: "\<lambda>s. P (tcb_at' t s)"
```
michaelmcinerney, line 18: Are we able to do a `typ_at'` and then lift them to get `tcb_at'` and `reply_at'`?

mktnk3: I suppose so, but not as a shorter proof than this crunch, perhaps?
I am using `updateReply_reply_at'` with `use_valid` (not with wp/wpsimp), so it seemed quickest to revive these crunches. And better before other people find their proofs potentially fail after rebasing, assuming that other missing ones are also used somewhere.

corlewis: Later on in this file there's a line that does this lifting:
```
global_interpretation updateReply: typ_at_all_props' "updateReply p f"
  by typ_at_props'
```
My strong preference would be to prove `P (typ_at' p s)` here and then use that, to minimise duplicated lemmas.

corlewis (changes requested):

proof/refine/ARM/Reply_R.thy
mktnk3: > Later on in this file there's a line that does this lifting:
> 
> ```
> global_interpretation updateReply: typ_at_all_props' "updateReply p f"
>   by typ_at_props'
> ```
> 
> My strong preference would be to prove `P (typ_at' p s)` here and then use that, to minimise duplicated lemmas.

There is a `global_interpretation` for `cleanReply` later in the file but not for `updateReply`.

There are a few things here. I think there’s something that are not quite clarified:

- `tcb_at’ crunch here existed from long before, even before the `typ_at_props’` update
-  no one said anything when `reply_at’` was added at the end of these crunches, which was after the `typ_at_props’` update
- my proof was working fine until these crunches were removed
- I am just getting things back to how it was before the last merge

So, if you are willing to add the global_interpretation for updateReply and incorporate it to remove part of these crunches, you are welcome to do that. But it probably doesn’t need to be now, maybe. And I suspect some of these need to be resurrected anyway.

mktnk3: So, there is no duplication as it currently is because there is no global interpretation for updateReply. Just to clarify.

corlewis: Oh, sorry. It turns out that I added the interpretation for `update_reply` at some point in my own work and then forgot about it.

All good then, as you suggested I'll update this later on when I finish of my changes.

corlewis (approved):

proof/refine/ARM/Reply_R.thy
mktnk3: Thanks, Corey. I will leave that to you, then!

PR #174, opened 2020-12-15
rt: heap list rename

mktnk3:
The name heap_list is already used in AutoCorres.
Renaming the one in lib/Heap_List.thy to heap_ls to avoid clash.


michaelmcinerney (approved):
I guess you have further uses of this in your upcoming `reply_remove_tcb_corres` proof, but if this is everything so far, cool

mktnk3:
Yes, there are more in there than you see here, but it's not too bad.

corlewis (approved):
That's an unfortunate name collision. Good job spotting it though!

MitchellBuckley (approved):
Nice!

lsf37 (approved):

mktnk3:
Merged as part of #176

PR #175, opened 2020-12-17
Reply pop spec update

VictorPhan:
Updating replyPop to match the abstract in that the `sched_context_donate` block is called later.


VictorPhan:
Replaced by Miki's PR: https://github.com/seL4/l4v/pull/185

PR #178, opened 2020-12-19
rt_replyPush

MitchellBuckley:
This proves `replyPush_corres` and adds proofs about a new function `bindScReply`.
* A large number of reply-related lemmas are moved to Reply_R and some new ones added.
* Tweaks to `donateSchedContext_corres` only weaken the preconditions slightly.
* Moving `donateSchedContext_corres` is isolated to a separate commit to make reviewing easier
* Some `FIXME RT: move` comments remain. I would rather not resolve them now, but I can if needed.

This PR replaces the previous one (https://github.com/seL4/l4v/pull/146) which I will decline since so much has changed.

### DIFF
--- a/proof/invariant-abstract/IpcDet_AI.thy
+++ b/proof/invariant-abstract/IpcDet_AI.thy
@@ -929,7 +929,7 @@ lemma reply_push_sender_sc_Some_invs:
              \<not> (case fault of None \<Rightarrow> False
                  | Some x \<Rightarrow> is_timeout_fault x))
    \<lbrace>\<lambda>r. invs\<rbrace>"
-  apply (clarsimp simp: reply_push_def)
+  apply (clarsimp simp: reply_push_def bind_sc_reply_def bind_assoc)
   apply (rule hoare_seq_ext[OF _ gsc_sp])
   apply (rule hoare_seq_ext[OF _ gsc_sp])
   apply (rule_tac S="sender_sc = sc_caller \<and> sender \<noteq> thread \<and> sender \<noteq> reply_ptr \<and> thread \<noteq> reply_ptr"
@@ -1005,6 +1005,7 @@ lemma reply_push_sender_sc_Some_invs:
                          sym_refs_bound_sc_tcb_iff_sc_tcb_sc_at)
                        (* sc_caller has non-empty sc_replies *)
   apply (clarsimp simp: invs_def valid_state_def valid_pspace_def)
+  apply (rename_tac sc_caller s x xa)
   apply (prop_tac "reply_at reply_ptr s",
          clarsimp simp: reply_sc_reply_at_def obj_at_def is_reply_def, simp)
   apply (prop_tac "reply_at x s",
@@ -1059,7 +1060,7 @@ lemma reply_push_sender_sc_Some_invs:
          apply (fastforce dest: reply_reftypes reply_at_ppred_reply_at)
         apply (fastforce dest: SCReply_ref_fst_replies_with_sc)
        apply (case_tac "sc_caller = x"; clarsimp)
-        apply_trace (fastforce dest: SCReply_ref_fst_replies_with_sc)
+        apply (fastforce dest: SCReply_ref_fst_replies_with_sc)
        apply (case_tac "y = x"; clarsimp)
         apply (fastforce dest: SCReply_ref_fst_replies_with_sc)
        apply (fastforce dest: SCReply_ref_fst_replies_with_sc)
@@ -1460,7 +1461,7 @@ lemma reply_push_st_tcb_at_Inactive:
   "\<lbrace>st_tcb_at ((=) Inactive) callee and K (callee \<noteq> caller)\<rbrace>
    reply_push caller callee reply_ptr can_donate
    \<lbrace>\<lambda>rv. st_tcb_at ((=) Inactive) callee\<rbrace>"
-  unfolding reply_push_def update_sk_obj_ref_def comp_def
+  unfolding reply_push_def bind_sc_reply_def update_sk_obj_ref_def comp_def
   by (wpsimp wp: get_simple_ko_wp get_tcb_obj_ref_wp hoare_vcg_if_lift hoare_vcg_all_lift
                  sts_st_tcb_at_cases
      | wp (once) hoare_drop_imp)+
@@ -1552,35 +1553,35 @@ lemma reply_push_invs_helper:
    apply (wpsimp simp: get_simple_ko_def get_object_def
                    wp: valid_sc_typ_list_all_reply valid_ioports_lift)
   apply (subgoal_tac "list_all (\<lambda>r. reply_at r s) (a # list) \<and> reply_ptr \<notin> set (a # list) \<and> distinct (a # list)")
-  apply (clarsimp simp: invs_def valid_pspace_def valid_state_def)
-  apply (intro conjI)
-  apply (rule replies_with_sc_upd_replies_valid_replies_add_one, simp)
-     apply (clarsimp simp:replies_blocked_def image_def, fastforce)
-    apply simp
-   apply (clarsimp simp:sc_replies_sc_at_def obj_at_def)
-   apply (erule delta_sym_refs)
-     apply (clarsimp split: if_splits
-                     elim!: delta_sym_refs)
-    apply (clarsimp simp: reply_sc_reply_at_def obj_at_def state_refs_of_def get_refs_def2
-                          pred_tcb_at_def is_tcb is_reply is_sc_obj sc_at_pred_n_def
-                   split: if_splits
-                   elim!: delta_sym_refs)
-   apply (safe; clarsimp?)
-   apply (subgoal_tac "(a,y) \<in> replies_with_sc s \<and> (a,sc_ptr) \<in> replies_with_sc s")
-    apply (clarsimp dest!: valid_replies_2_inj_onD )
+   apply (clarsimp simp: invs_def valid_pspace_def valid_state_def)
    apply (intro conjI)
-    apply (subgoal_tac "valid_reply reply s")
-     apply (clarsimp simp: valid_reply_def refs_of_def obj_at_def is_sc_obj_def
-                    split: option.splits)
-     apply (case_tac x2; clarsimp simp: get_refs_def)
-     apply (erule disjE, clarsimp split: option.splits)+
-     apply (clarsimp simp: replies_with_sc_def sc_replies_sc_at_def obj_at_def split: option.splits)
-    apply (erule valid_objs_valid_reply, assumption)
-   apply(clarsimp simp: replies_with_sc_def sc_replies_sc_at_def obj_at_def)
-  apply (metis cons_set_intro)
-  apply (fastforce simp: idle_sc_no_ex_cap tcb_at_def is_tcb_def
-                   dest: pred_tcb_at_tcb_at get_tcb_SomeD)
-  apply (clarsimp simp: pred_tcb_at_def obj_at_def is_tcb)
+      apply (rule replies_with_sc_upd_replies_valid_replies_add_one, simp)
+        apply (clarsimp simp:replies_blocked_def image_def, fastforce)
+       apply simp
+      apply (clarsimp simp:sc_replies_sc_at_def obj_at_def)
+     apply (erule delta_sym_refs)
+      apply (clarsimp split: if_splits
+                      elim!: delta_sym_refs)
+     apply (clarsimp simp: reply_sc_reply_at_def obj_at_def state_refs_of_def get_refs_def2
+                           pred_tcb_at_def is_tcb is_reply is_sc_obj sc_at_pred_n_def
+                    split: if_splits
+                    elim!: delta_sym_refs)
+     apply (safe; clarsimp?)
+     apply (subgoal_tac "(a,y) \<in> replies_with_sc s \<and> (a,sc_ptr) \<in> replies_with_sc s")
+      apply (clarsimp dest!: valid_replies_2_inj_onD )
+     apply (intro conjI)
+      apply (subgoal_tac "valid_reply reply s")
+       apply (clarsimp simp: valid_reply_def refs_of_def obj_at_def is_sc_obj_def
+                      split: option.splits)
+       apply (case_tac x2; clarsimp simp: get_refs_def)
+       apply (erule disjE, clarsimp split: option.splits)+
+       apply (clarsimp simp: replies_with_sc_def sc_replies_sc_at_def obj_at_def split: option.splits)
+      apply (erule valid_objs_valid_reply, assumption)
+     apply (clarsimp simp: replies_with_sc_def sc_replies_sc_at_def obj_at_def)
+     apply (metis cons_set_intro)
+    apply (fastforce simp: idle_sc_no_ex_cap tcb_at_def is_tcb_def
+                     dest: pred_tcb_at_tcb_at get_tcb_SomeD)
+   apply (clarsimp simp: pred_tcb_at_def obj_at_def is_tcb)
   apply (clarsimp simp del: distinct.simps list.pred_inject insert_iff)
   apply (frule sc_replies_sc_at_subset_fst_replies_with_sc)
   apply (frule invs_valid_objs)
@@ -1618,7 +1619,7 @@ lemma reply_push_invs':
    reply_push caller callee reply_ptr can_donate
    \<lbrace>\<lambda>rv. invs\<rbrace>"
   supply if_weak_cong[cong del]
-  apply (simp add: reply_push_def)
+  apply (simp add: reply_push_def bind_sc_reply_def bind_assoc)
   apply (rule hoare_seq_ext[OF _ gsc_sp])
   apply (rule hoare_seq_ext[OF _ gsc_sp])
   apply (case_tac sc_caller; simp)
@@ -1641,7 +1642,6 @@ lemma reply_push_invs':
                           tcb_st_refs_of_def reply_tcb_reply_at_def
                    split: if_splits thread_state.splits)
    apply (clarsimp simp: idle_no_ex_cap)
-
   apply (wpsimp wp: reply_push_invs_helper)
     apply (wpsimp simp: invs_def valid_state_def valid_pspace_def
                     wp: sts_only_idle valid_irq_node_typ sts_valid_replies
--- a/proof/invariant-abstract/IpcDet_AI.thy
+++ b/proof/invariant-abstract/IpcDet_AI.thy
@@ -932,13 +932,6 @@ lemma reply_push_sender_sc_Some_invs:
   apply (clarsimp simp: reply_push_def)
   apply (rule hoare_seq_ext[OF _ gsc_sp])
   apply (rule hoare_seq_ext[OF _ gsc_sp])
-  apply (rule hoare_seq_ext[OF _ liftM_wp[where P=P and Q="\<lambda>r. P" for P, simplified,
-                                          OF get_simple_ko_inv]])
-  apply (rule hoare_seq_ext[OF _ assert_inv])
-  apply (rule hoare_seq_ext[OF _ no_reply_in_ts_inv])
-  apply (rule hoare_seq_ext[OF _ assert_inv])
-  apply (rule hoare_seq_ext[OF _ no_reply_in_ts_inv])
-  apply (rule hoare_seq_ext[OF _ assert_inv])
   apply (rule_tac S="sender_sc = sc_caller \<and> sender \<noteq> thread \<and> sender \<noteq> reply_ptr \<and> thread \<noteq> reply_ptr"
            in hoare_gen_asm'', fastforce simp: sk_obj_at_pred_def pred_tcb_at_def obj_at_def)
   apply (rule subst[of "do _ <- do _ <- a; b od; c od"
@@ -1012,8 +1005,10 @@ lemma reply_push_sender_sc_Some_invs:
                          sym_refs_bound_sc_tcb_iff_sc_tcb_sc_at)
                        (* sc_caller has non-empty sc_replies *)
   apply (clarsimp simp: invs_def valid_state_def valid_pspace_def)
-  apply (rule conjI, clarsimp simp: reply_sc_reply_at_def obj_at_def is_reply_def)
-  apply (rule conjI, clarsimp simp:  obj_at_def is_reply_def)
+  apply (prop_tac "reply_at reply_ptr s",
+         clarsimp simp: reply_sc_reply_at_def obj_at_def is_reply_def, simp)
+  apply (prop_tac "reply_at x s",
+         erule (1) valid_objs_sc_replies_reply_at, simp, simp)
   apply (rule conjI, rule replies_blocked_list_all_reply_at, assumption)
    apply (drule valid_replies_2D1)
    apply (rule_tac B="fst ` replies_with_sc s" in subset_trans; simp?)
@@ -1022,7 +1017,8 @@ lemma reply_push_sender_sc_Some_invs:
   apply (rule conjI, clarsimp)
    apply (drule sc_replies_sc_at_hd_SCReply_ref)
    apply (drule sym_refsD, assumption)
-   apply (clarsimp simp: state_refs_of_def reply_sc_reply_at_def obj_at_def)
+   apply (clarsimp simp: state_refs_of_def reply_sc_reply_at_def obj_at_def get_refs_def
+                  split: option.splits)
   apply (rule conjI, clarsimp)
    apply (subgoal_tac "(reply_ptr, sc_caller) \<in> replies_with_sc s", fastforce simp: image_def)
    apply (clarsimp simp: replies_with_sc_def sc_replies_sc_at_def obj_at_def)
@@ -1063,22 +1059,21 @@ lemma reply_push_sender_sc_Some_invs:
          apply (fastforce dest: reply_reftypes reply_at_ppred_reply_at)
         apply (fastforce dest: SCReply_ref_fst_replies_with_sc)
        apply (case_tac "sc_caller = x"; clarsimp)
-        apply (fastforce dest: SCReply_ref_fst_replies_with_sc)
+        apply_trace (fastforce dest: SCReply_ref_fst_replies_with_sc)
        apply (case_tac "y = x"; clarsimp)
         apply (fastforce dest: SCReply_ref_fst_replies_with_sc)
        apply (fastforce dest: SCReply_ref_fst_replies_with_sc)
-      apply (case_tac "y = reply_ptr"; clarsimp)
-       apply (fastforce dest: reply_reftypes reply_at_ppred_reply_at)
-      apply (safe; simp)[1]
-       apply (fastforce simp: valid_reply_def obj_at_def is_sc_obj_def dest: valid_objs_valid_reply)
-      apply (fastforce simp: valid_reply_def obj_at_def is_sc_obj_def dest: valid_objs_valid_reply)
+      apply (fastforce simp: sc_at_pred_n_def obj_at_def reply_sc_reply_at_def)
      apply (case_tac "reply_ptr = x"; clarsimp)
       apply (clarsimp simp: reply_sc_reply_at_def obj_at_def)
+      apply (prop_tac "(reply_ptr, sc_caller) \<in> replies_with_sc s")
+       apply (clarsimp simp: replies_with_sc_def sc_at_pred_n_def obj_at_def)
+       apply (metis cons_set_intro)
+      apply fastforce
      apply (clarsimp simp: reply_sc_reply_at_def obj_at_def)
      apply (fastforce dest: sc_replies_sc_at_state_refs_of[OF eq_commute, THEN iffD2])
     apply (safe; simp?)[1]
-      apply (fastforce simp: valid_reply_def obj_at_def reply_sc_reply_at_def is_sc_obj_def
-                       dest: valid_objs_valid_reply)
+      apply (fastforce simp: sc_at_pred_n_def obj_at_def reply_sc_reply_at_def)
      apply (fastforce dest: reply_reftypes reply_at_ppred_reply_at)
     apply (drule SCReply_ref_replies_with_sc)
     apply (subgoal_tac "(x, sc_caller) \<in> replies_with_sc s")
@@ -1511,11 +1506,8 @@ lemma reply_push_invs_helper:
          \<and> reply_ptr \<notin> fst ` replies_with_sc s \<rbrace>
     when (sc_callee = None \<and> donate) (do
       sc_replies <- liftM sc_replies (get_sched_context sc_ptr);
-      y <- case sc_replies of [] \<Rightarrow> assert True
-              | r # x \<Rightarrow> do reply <- get_reply r;
-                            assert (reply_sc reply = sc_caller);
-                            set_reply_obj_ref reply_sc_update r None
-                         od;
+      y <- case sc_replies of [] \<Rightarrow> return ()
+              | r # x \<Rightarrow> set_reply_obj_ref reply_sc_update r None;
       y <- set_sc_obj_ref sc_replies_update sc_ptr (reply_ptr # sc_replies);
       y <- set_reply_obj_ref reply_sc_update reply_ptr (Some sc_ptr);
       sched_context_donate sc_ptr callee
@@ -1559,34 +1551,20 @@ lemma reply_push_invs_helper:
     apply (wpsimp wp: set_sc_replies_valid_replies update_sched_context_valid_idle)
    apply (wpsimp simp: get_simple_ko_def get_object_def
                    wp: valid_sc_typ_list_all_reply valid_ioports_lift)
-  apply (clarsimp split: if_splits, intro conjI impI)
-                apply (clarsimp simp: is_reply obj_at_def is_sc_obj_def)+
-        apply (subgoal_tac "reply_sc_reply_at (\<lambda>sp. sp = Some sc_ptr) reply_ptr s")
-         apply (clarsimp simp:reply_sc_reply_at_def obj_at_def)
-        apply (fastforce simp: sym_refs_reply_sc_reply_at invs_sym_refs reply_at_ppred_reply_at)
-       apply (clarsimp simp: is_reply obj_at_def is_sc_obj_def)+
-  apply (clarsimp simp: invs_def valid_state_def valid_pspace_def sc_replies_sc_at_def
-                        valid_obj_def valid_sched_context_def
-                 elim!: obj_at_valid_objsE)
-  apply (drule_tac s="a # list" in sym, simp)
-  apply (rule conjI, clarsimp simp: obj_at_def is_reply)
-  apply (rule conjI, clarsimp simp: replies_with_sc_def image_def sc_replies_sc_at_def obj_at_def)
-   apply (subgoal_tac "reply_ptr \<in>  set (sc_replies sc)")
-    apply (drule_tac x=sc_ptr in spec, fastforce)
-   apply fastforce
-  apply (rule conjI, rule replies_with_sc_upd_replies_valid_replies_add_one, clarsimp)
+  apply (subgoal_tac "list_all (\<lambda>r. reply_at r s) (a # list) \<and> reply_ptr \<notin> set (a # list) \<and> distinct (a # list)")
+  apply (clarsimp simp: invs_def valid_pspace_def valid_state_def)
+  apply (intro conjI)
+  apply (rule replies_with_sc_upd_replies_valid_replies_add_one, simp)
      apply (clarsimp simp:replies_blocked_def image_def, fastforce)
     apply simp
    apply (clarsimp simp:sc_replies_sc_at_def obj_at_def)
-  apply (rule conjI)
    apply (erule delta_sym_refs)
-    apply (clarsimp split: if_splits
-                    elim!: delta_sym_refs)
-   apply (clarsimp simp: reply_sc_reply_at_def obj_at_def state_refs_of_def get_refs_def2
-                          pred_tcb_at_def is_tcb
+     apply (clarsimp split: if_splits
+                     elim!: delta_sym_refs)
+    apply (clarsimp simp: reply_sc_reply_at_def obj_at_def state_refs_of_def get_refs_def2
+                          pred_tcb_at_def is_tcb is_reply is_sc_obj sc_at_pred_n_def
                    split: if_splits
                    elim!: delta_sym_refs)
-    apply (safe; clarsimp?)
    apply (safe; clarsimp?)
    apply (subgoal_tac "(a,y) \<in> replies_with_sc s \<and> (a,sc_ptr) \<in> replies_with_sc s")
     apply (clarsimp dest!: valid_replies_2_inj_onD )
@@ -1599,8 +1577,18 @@ lemma reply_push_invs_helper:
      apply (clarsimp simp: replies_with_sc_def sc_replies_sc_at_def obj_at_def split: option.splits)
     apply (erule valid_objs_valid_reply, assumption)
    apply(clarsimp simp: replies_with_sc_def sc_replies_sc_at_def obj_at_def)
+  apply (metis cons_set_intro)
   apply (fastforce simp: idle_sc_no_ex_cap tcb_at_def is_tcb_def
                    dest: pred_tcb_at_tcb_at get_tcb_SomeD)
+  apply (clarsimp simp: pred_tcb_at_def obj_at_def is_tcb)
+  apply (clarsimp simp del: distinct.simps list.pred_inject insert_iff)
+  apply (frule sc_replies_sc_at_subset_fst_replies_with_sc)
+  apply (frule invs_valid_objs)
+  apply (intro conjI)
+    apply (erule replies_blocked_list_all_reply_at)
+    apply (meson dual_order.trans invs_valid_replies valid_replies_defs(1))
+   apply fastforce
+  apply (erule (1) valid_objs_sc_replies_distinct)
   done
 
 lemma sts_in_replies_blocked:
@@ -1626,19 +1614,13 @@ lemma reply_push_invs':
   "\<lbrace>all_invs_but_fault_tcbs and fault_tcbs_valid_states_except_set {caller} and
     ex_nonz_cap_to callee and ex_nonz_cap_to caller and ex_nonz_cap_to reply_ptr and
     st_tcb_at active caller and st_tcb_at ((=) Inactive) callee and
-    reply_sc_reply_at (\<lambda>sc. sc = None) reply_ptr\<rbrace>
+    reply_sc_reply_at (\<lambda>sc. sc = None) reply_ptr and reply_tcb_reply_at (\<lambda>tptr. tptr = None) reply_ptr\<rbrace>
    reply_push caller callee reply_ptr can_donate
    \<lbrace>\<lambda>rv. invs\<rbrace>"
   supply if_weak_cong[cong del]
   apply (simp add: reply_push_def)
   apply (rule hoare_seq_ext[OF _ gsc_sp])
   apply (rule hoare_seq_ext[OF _ gsc_sp])
-  apply (rule hoare_seq_ext[OF _ grt_sp])
-  apply (rule hoare_seq_ext[OF _ assert_sp])
-  apply (rule hoare_seq_ext[OF _ no_reply_in_ts_inv])
-  apply (rule hoare_seq_ext[OF _ assert_inv])
-  apply (rule hoare_seq_ext[OF _ no_reply_in_ts_inv])
-  apply (rule hoare_seq_ext[OF _ assert_inv])
   apply (case_tac sc_caller; simp)
    apply (wpsimp simp: invs_def valid_state_def valid_pspace_def
                    wp: sts_only_idle valid_irq_node_typ set_reply_tcb_valid_tcb_state
@@ -1659,6 +1641,7 @@ lemma reply_push_invs':
                           tcb_st_refs_of_def reply_tcb_reply_at_def
                    split: if_splits thread_state.splits)
    apply (clarsimp simp: idle_no_ex_cap)
+
   apply (wpsimp wp: reply_push_invs_helper)
     apply (wpsimp simp: invs_def valid_state_def valid_pspace_def
                     wp: sts_only_idle valid_irq_node_typ sts_valid_replies
@@ -1704,7 +1687,8 @@ lemma reply_push_invs':
 lemma reply_push_invs:
   "\<lbrace>invs and ex_nonz_cap_to callee and ex_nonz_cap_to caller and ex_nonz_cap_to reply_ptr and
     st_tcb_at active caller and st_tcb_at ((=) Inactive) callee and
-    reply_sc_reply_at (\<lambda>sc. sc = None) reply_ptr\<rbrace>
+    reply_sc_reply_at (\<lambda>sc. sc = None) reply_ptr and
+    reply_tcb_reply_at (\<lambda>tptr. tptr = None) reply_ptr\<rbrace>
    reply_push caller callee reply_ptr can_donate
    \<lbrace>\<lambda>rv. invs\<rbrace>"
   by (wpsimp wp: reply_push_invs' simp: invs_def valid_state_def valid_pspace_def) auto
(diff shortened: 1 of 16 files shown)


proof/refine/ARM/Ipc_R.thy
```diff
@@ -3161,6 +3161,150 @@ lemma maybeReturnSc_corres:
...
+lemma schedContextDonate_corres_helper:
+  "(case rv' of SwitchToThread x \<Rightarrow> when (x = t \<or> t = cur) rescheduleRequired
+                             | _ \<Rightarrow> when (t = cur) rescheduleRequired) =
+   (when (t = cur \<or> (case rv' of SwitchToThread x \<Rightarrow> t = x | _ \<Rightarrow> False)) rescheduleRequired)"
+  by (case_tac rv'; clarsimp simp: when_def)
+
+lemma schedContextDonate_corres:
+  "corres dc (invs and sc_at scp and tcb_at thread and weak_valid_sched_action) invs'
+             (sched_context_donate scp thread)
+             (schedContextDonate scp thread)"
```
mktnk3: Can we have only the invariants that we really need in the preconditions, instead of having `invs` and `invs’`?

MitchellBuckley: Yes, this is done in the following commit.

proof/invariant-abstract/IpcDet_AI.thy
```diff
@@ -1550,34 +1543,20 @@ lemma reply_push_invs_helper:
...
-                 elim!: obj_at_valid_objsE)
-  apply (drule_tac s="a # list" in sym, simp)
-  apply (rule conjI, clarsimp simp: obj_at_def is_reply)
-  apply (rule conjI, clarsimp simp: replies_with_sc_def image_def sc_replies_sc_at_def obj_at_def)
-   apply (subgoal_tac "reply_ptr \<in>  set (sc_replies sc)")
-    apply (drule_tac x=sc_ptr in spec, fastforce)
-   apply fastforce
-  apply (rule conjI, rule replies_with_sc_upd_replies_valid_replies_add_one, clarsimp)
+  apply (subgoal_tac "list_all (\<lambda>r. reply_at r s) (a # list) \<and> reply_ptr \<notin> set (a # list) \<and> distinct (a # list)")
+  apply (clarsimp simp: invs_def valid_pspace_def valid_state_def)
```
michaelmcinerney, line 1547: Doesn't the `subgoal_tac` generate another subgoal?

proof/refine/ARM/Ipc_R.thy
```diff
@@ -3161,6 +3161,340 @@ lemma maybeReturnSc_corres:
...
+   apply (clarsimp simp: idle_sc_no_ex_cap)
+  apply wpsimp
+     apply (wpsimp simp: invs_def valid_state_def valid_pspace_def
+                     wp: valid_irq_node_typ set_reply_sc_valid_replies_already_BlockedOnReply
+                         valid_ioports_lift valid_sc_typ_list_all_reply)
+    apply (wpsimp wp: set_sc_replies_valid_replies update_sched_context_valid_idle)
+   apply (wpsimp simp: get_simple_ko_def get_object_def
+                   wp: valid_sc_typ_list_all_reply valid_ioports_lift)
+  apply (subgoal_tac "list_all (\<lambda>r. reply_at r s) (a # list) \<and> reply_ptr \<notin> set (a # list) \<and> distinct (a # list)")
+  apply (clarsimp simp: invs_def valid_pspace_def valid_state_def)
```
michaelmcinerney, line 3362: Indentation?

proof/refine/ARM/Ipc_R.thy
```diff
@@ -3161,6 +3161,340 @@ lemma maybeReturnSc_corres:
...
+  apply (intro conjI)
```
michaelmcinerney, line 3363: Doesn't the `intro conjI` produce new subgoals?

proof/refine/ARM/Ipc_R.thy
```diff
@@ -3161,6 +3161,340 @@ lemma maybeReturnSc_corres:
...
+                   split: if_splits
+                   elim!: delta_sym_refs)
+   apply (safe; clarsimp?)
+   apply (subgoal_tac "(a,y) \<in> replies_with_sc s \<and> (a,sc_ptr) \<in> replies_with_sc s")
+    apply (clarsimp dest!: valid_replies_2_inj_onD )
+   apply (intro conjI)
+    apply (subgoal_tac "valid_reply reply s")
+     apply (clarsimp simp: valid_reply_def refs_of_def obj_at_def is_sc_obj_def
+                    split: option.splits)
+     apply (case_tac x2; clarsimp simp: get_refs_def)
```
michaelmcinerney, line 3382: Could you `rename_tac` this `x2`?

proof/refine/ARM/Ipc_R.thy
```diff
@@ -3161,6 +3161,340 @@ lemma maybeReturnSc_corres:
...
+
+lemma replyPush_corres:
+  "can_donate = can_donate' \<Longrightarrow>
+   corres dc (invs and tcb_at caller and tcb_at callee and reply_at reply_ptr
+              and ex_nonz_cap_to reply_ptr
+              and st_tcb_at active caller
+              and reply_sc_reply_at (\<lambda>tptr. tptr = None) reply_ptr
+              and reply_tcb_reply_at (\<lambda>tptr. tptr = None) reply_ptr
+              and weak_valid_sched_action and scheduler_act_not caller)
+             (valid_release_queue and valid_release_queue'
```
michaelmcinerney, line 3408: You can use the `valid_release_queue_iff` abbreviation here is you'd like

proof/refine/ARM/Ipc_R.thy
```diff
@@ -4085,27 +4275,7 @@ lemma valid_tcb_state'_update[simp]:
...
-   \<lbrace>\<lambda>rv s. P (obj_at' Q p' s)\<rbrace>"
-  apply (simp only: obj_at'_real_def)
-  apply (wp set_reply'.ko_wp_at)
-  apply (auto simp: obj_at'_def ko_wp_at'_def projectKO_eq projectKO_reply)
-  done
-
-lemma setObject_reply_pde_mappings'[wp]:
-  "setObject ptr (val :: reply) \<lbrace>valid_pde_mappings'\<rbrace>"
-  by (wp valid_pde_mappings_lift')
+crunches bindScReply for irq_node'[wp]: "\<lambda>s. P (irq_node' s)"
```
michaelmcinerney, line 4278: Don't we normally put the `for blah` bit on a new line? I'd consult the style guide, IF THERE WERE ONE

proof/refine/ARM/Ipc_R.thy
```diff
@@ -4137,240 +4307,196 @@ lemma tcbSchedEnqueue_valid_tcb_state'[wp]:
...
+  apply (clarsimp simp: obj_at'_real_def ko_wp_at'_def opt_map_def projectKO_reply)
+  apply (case_tac "replyNext obj"; simp)
+  apply (case_tac a; simp)
+  done
+
+lemma obj_at'_SCReply_ref:
+  "obj_at' (\<lambda>a. scReply a = Some rp) scp s
+   \<Longrightarrow> (rp, SCReply) \<in> state_refs_of' s scp"
+   by (clarsimp simp: obj_at'_real_def ko_wp_at'_def state_refs_of'_def
+      projectKO_sc get_refs_def)
```
michaelmcinerney, line 4343: Indenation

proof/refine/ARM/Ipc_R.thy
```diff
@@ -4137,240 +4307,196 @@ lemma tcbSchedEnqueue_valid_tcb_state'[wp]:
...
+    and obj_at' (\<lambda>ko. replyTCB ko = None) replyPtr\<rbrace>
+   replyPush callerPtr calleePtr replyPtr canDonate
+   \<lbrace>\<lambda>_ s. sym_refs (list_refs_of_replies' s)\<rbrace>"
+  supply if_split [split del]
+  unfolding replyPush_def
+  apply wpsimp
+         apply (wpsimp wp: bindScReply_sym_refs_list_refs_of_replies'
+                           hoare_vcg_if_lift hoare_vcg_imp_lift' hoare_vcg_all_lift)
+        apply (rule_tac Q="(\<lambda>_ s.
+                sym_refs (list_refs_of_replies' s) \<and>
```
michaelmcinerney, line 4365: Reformat

proof/refine/ARM/Ipc_R.thy
```diff
@@ -4137,240 +4307,196 @@ lemma tcbSchedEnqueue_valid_tcb_state'[wp]:
...
+         apply (fastforce split: if_splits simp del: comp_apply)[1]
+        apply (wpsimp wp: hoare_vcg_all_lift hoare_vcg_imp_lift'
+                          updateReply_obj_at'_inv threadGet_wp)+
+  apply (intro conjI)
+     apply (clarsimp simp: obj_at'_real_def ko_wp_at'_def)
+    apply (clarsimp simp: obj_at'_real_def ko_wp_at'_def)
+   apply (clarsimp simp: sym_refs_asrt_def)
+   apply (rule ReplySchedContext_obj_at'_ref)
+    apply (fastforce simp: sym_refs_def dest: obj_at'_SCReply_ref)
+   apply (frule_tac p=xa in  obj_at_ko_at', clarsimp)
```
michaelmcinerney, line 4378: rename_tac this xa

proof/refine/ARM/Ipc_R.thy
```diff
@@ -4137,240 +4307,196 @@ lemma tcbSchedEnqueue_valid_tcb_state'[wp]:
...
-   apply (drule_tac ko=obj in sym_refs_ko_atD'[rotated])
-    apply (fastforce simp: obj_at'_def projectKO_eq projectKO_tcb)
-   apply (fastforce simp: live_sc'_def idle_tcb'_def valid_idle'_def
-                          refs_of_rev' pred_tcb_at'_def ko_wp_at'_def obj_at'_def)
-  apply (fastforce elim: if_live_then_nonz_capE' dest: sym_refs_ko_atD'[rotated]
-                   simp: ko_wp_at'_def obj_at'_def projectKO_eq refs_of_rev' live_reply'_def)
+   apply (intro conjI)
+    apply (clarsimp simp: obj_at'_def projectKO_eq projectKO_tcb projectKO_sc projectKO_reply)
+    apply (erule if_live_then_nonz_capE')
+    apply (drule_tac ko=obja in sym_refs_ko_atD'[rotated])
```
michaelmcinerney, line 4409: rename_tac

proof/refine/ARM/KHeap_R.thy
```diff
@@ -1375,38 +1375,113 @@ lemma set_reply_corres: (* for reply update that doesn't touch the reply stack *
         simp add: typ_at'_def ko_wp_at'_def obj_at'_def project_inject opt_map_def sc_of_def)
   qed
 
-lemma setSchedContext_no_stack_update_corres:
-  "\<lbrakk>sc_relation sc n sc' \<longrightarrow> sc_relation (f sc) n (f' sc');
-     \<forall>sc. sc_replies sc = sc_replies (f sc); \<forall>sc'. objBits sc' = objBits (f' sc');
-     scReply sc' = scReply (f' sc')\<rbrakk>
-    \<Longrightarrow> corres dc
+lemma setReply_corres: (* for reply updates on replies not in fst ` replies_with_sc *)
```
michaelmcinerney, line 1378: Some parts of this look a bit similar to some of the `setSchedContext` corres rules that @mktnk3  has. Is there any way to refactor things to reduce duplication?

proof/refine/ARM/KHeap_R.thy
```diff
@@ -3641,5 +3747,61 @@ lemma update_sc_no_reply_stack_update_corres:
...
+lemma ko_at'_inj:
+  "ko_at' ko ptr  s \<Longrightarrow> ko_at' ko' ptr s \<Longrightarrow> ko' = ko"
+  by (clarsimp simp: obj_at'_real_def ko_wp_at'_def)
+
+(* FIXME RT: Move these to AInvs where possible *)
+definition injective_ref where
+  "injective_ref ref heap \<equiv> (\<forall>q p1 p2. (p1, ref) \<in> heap q \<and> (p2, ref) \<in> heap q \<longrightarrow> p1 = p2)"
+
+lemma sym_refs_inj:
+  "sym_refs heap \<Longrightarrow>
```
michaelmcinerney, line 3759: Use [| ...; ...|] instead (and in the following proof)

proof/refine/ARM/KHeap_R.thy
```diff
@@ -3641,5 +3747,61 @@ lemma update_sc_no_reply_stack_update_corres:
...
+  apply (clarsimp simp: refs_of'_def)
+  apply (prop_tac "\<exists>z. x2 = KOSchedContext z")
+   apply (clarsimp split: kernel_object.splits)
+      apply (clarsimp split: Structures_H.endpoint.splits simp: ep_q_refs_of'_def)
+     apply (clarsimp split: Structures_H.ntfn.splits option.splits
+                      simp: ntfn_q_refs_of'_def get_refs_def)
+    apply (clarsimp simp: tcb_st_refs_of'_def tcb_bound_refs'_def get_refs_def
+                   split: Structures_H.thread_state.splits if_splits option.splits)
+   apply (clarsimp simp: get_refs_def split: option.splits)
+  apply clarsimp
```
michaelmcinerney, line 3791: Are both clarsimps needed? Also rename_tac above

proof/refine/ARM/Reply_R.thy
```diff
@@ -634,4 +618,486 @@ lemma replyRemoveTCB_invs':
...
+  "corres dc (reply_at rp) (reply_at' rp
+              and obj_at' (\<lambda>ko. replyPrev_same (f (\<lambda>_. new) ko) ko) rp)
+            (set_reply_obj_ref g rp new)
+            (updateReply rp (f (\<lambda>_. new)))"
+  apply (simp add: update_sk_obj_ref_def updateReply_def)
+  apply (rule corres_guard_imp)
+    apply (rule corres_split[OF _ get_reply_corres])
+      apply (rule set_reply_corres)
+      apply (simp add: rr)
+  apply wpsimp+
```
michaelmcinerney, line 657: Indentation much be off here. And the formatting of the conclusion of the lemma is off

proof/refine/ARM/Reply_R.thy
```diff
@@ -634,4 +618,486 @@ lemma replyRemoveTCB_invs':
...
+  done
+
+lemma setObject_reply_pde_mappings'[wp]:
+  "setObject ptr (val :: reply) \<lbrace>valid_pde_mappings'\<rbrace>"
+  by (wp valid_pde_mappings_lift')
+
+crunches bindScReply
+  for pspace_aligned'[wp]: pspace_aligned'
+  and pspace_distinct'[wp]: pspace_distinct'
+  and if_unsafe_then_cap'[wp]: "if_unsafe_then_cap'"
```
michaelmcinerney, line 861: Many of these quotation marks are unnecessary

proof/refine/ARM/Reply_R.thy
```diff
@@ -634,4 +618,486 @@ lemma replyRemoveTCB_invs':
...
+      and z: "p2refs = rfs' p2"
+      and symrefrel: "tp' = symreftype tp"
+  shows "sym_refs (\<lambda>p. if p = p1 then (insert (p2, tp) p1refs)
+                  else if p = p2 then (insert (p1, tp') p2refs)
+                  else rfs' p)"
+  apply (rule delta_sym_refs[OF x])
+  using diff by (auto simp: y z symrefrel split: if_splits)
+
+lemma no_tcb_not_in_replies_with_sc:
+  "valid_replies s \<Longrightarrow>
```
michaelmcinerney, line 965: Use [| ...; ... |]

proof/refine/ARM/Reply_R.thy
```diff
@@ -634,4 +618,486 @@ lemma replyRemoveTCB_invs':
...
+   sym_refs (state_refs_of s) \<Longrightarrow>
+   reply_tcb_reply_at (\<lambda>tptr. tptr = None) reply_ptr s \<Longrightarrow>
+   reply_ptr \<notin> fst ` replies_with_sc s"
+  apply (intro notI)
+  unfolding valid_replies_defs
+  apply (erule conjE)
+  apply (drule (1) set_mp)
+  apply clarsimp
+  apply (frule_tac thread = ba in  st_tcb_reply_state_refs[rotated])
```
michaelmcinerney, line 974: Extra space, rename_tac

proof/refine/ARM/Reply_R.thy
```diff
@@ -634,4 +618,486 @@ lemma replyRemoveTCB_invs':
...
+  "\<lbrace>\<lambda>s. \<forall>ko. ko_at' ko rptr s \<longrightarrow> P ((list_refs_of_replies' s)(rptr := list_refs_of_reply' (f ko)))\<rbrace>
+   updateReply rptr f
+   \<lbrace>\<lambda>rv s. P (list_refs_of_replies' s)\<rbrace>"
+  unfolding updateReply_def by wp
+
+lemma replyTCB_update_list_refs_of_replies'[wp]:
+  "updateReply rptr (replyTCB_update f) \<lbrace>\<lambda>s. P (list_refs_of_replies' s)\<rbrace>"
+  apply (wpsimp wp: updateReply_list_refs_of_replies')
+  apply (subst arg_cong[where f=P])
+  prefer 2 apply assumption
```
michaelmcinerney, line 989: Indentation

proof/refine/ARM/Reply_R.thy
```diff
@@ -634,4 +618,486 @@ lemma replyRemoveTCB_invs':
...
+  by (wpsimp wp: updateReply_iflive'_strong)
+
+lemma bindScReply_valid_objs'[wp]:
+  "\<lbrace>valid_objs' and reply_at' replyPtr and sc_at' scp\<rbrace>
+   bindScReply scp replyPtr
+   \<lbrace>\<lambda>_. valid_objs'\<rbrace>"
+  apply (solves wp | simp (no_asm_use) add: bindScReply_def split del: if_split cong: conj_cong |
+           wp hoare_when_wp haskell_assert_wp hoare_vcg_if_lift2 hoare_vcg_all_lift
+              hoare_vcg_disj_lift hoare_vcg_imp_lift' set_sc'.valid_objs'
+              set_reply'.valid_objs' setReply_obj_at' )+
```
michaelmcinerney, line 1016: Extra space

proof/refine/ARM/Reply_R.thy
```diff
@@ -634,4 +618,486 @@ lemma replyRemoveTCB_invs':
...
+lemma bindScReply_sym_refs_list_refs_of_replies':
+  "\<lbrace>\<lambda>s. sym_refs (list_refs_of_replies' s) \<and>
+        obj_at' (\<lambda>ko. replyPrev ko = None \<and> replyNext ko = None) replyPtr s
+       \<and> (\<forall>oldReplyPtr. obj_at' (\<lambda>ko. scReply ko = Some oldReplyPtr) scPtr s \<longrightarrow>
+           obj_at' (\<lambda>ko. replyNext ko = Some (Head scPtr)) oldReplyPtr s)\<rbrace>
+   bindScReply scPtr replyPtr
+   \<lbrace>\<lambda>_ s. sym_refs (list_refs_of_replies' s)\<rbrace>"
+  supply if_split [split del]
+  unfolding bindScReply_def
+  apply (wpsimp wp: hoare_vcg_if_lift2 updateReply_list_refs_of_replies'
```
michaelmcinerney, line 1046: Indentation here, and in the following line. And the auto at the end doesn't need the `[1]`

proof/refine/ARM/Reply_R.thy
```diff
@@ -634,4 +618,486 @@ lemma replyRemoveTCB_invs':
...
+  "setReply rp new  \<lbrace>valid_tcb' tcb\<rbrace>"
+  apply (clarsimp simp: setReply_def)
+  apply (rule setObject_valid_tcb')
+  done
+
+lemma setReply_valid_tcbs'[wp]:
+  "setReply rp new  \<lbrace>valid_tcbs'\<rbrace>"
+  unfolding valid_tcbs'_def
+  apply (wpsimp wp: hoare_vcg_all_lift hoare_vcg_imp_lift')
+    apply (wpsimp simp: setReply_def )+
```
michaelmcinerney, line 1096: Extra space

spec/haskell/src/SEL4/Object/Reply.lhs
```diff
@@ -34,51 +34,31 @@ This module specifies the behavior of reply objects.
 >     reply <- getReply replyPtr
 >     setReply replyPtr (upd reply)
 
+> bindScReply :: PPtr SchedContext -> PPtr Reply -> Kernel ()
+> bindScReply scPtr replyPtr = do
+>     sc <- getSchedContext scPtr
+>     scReplyOpt <- return $ scReply sc
+>     when (scReplyOpt /= Nothing) $ do
+>         scReplyPtr <- return $ fromJust scReplyOpt
+>         updateReply scReplyPtr (\reply -> reply { replyNext = Just (Next replyPtr)  })
```
michaelmcinerney, line 43: Extra space

michaelmcinerney (approved):

proof/invariant-abstract/IpcDet_AI.thy
lsf37: Yes, there should be and additional indent here

proof/refine/ARM/TcbAcc_R.thy
```diff
@@ -315,6 +315,16 @@ lemma valid_tcbs'_maxPriority:
 
 lemmas valid_objs'_maxPriority = valid_tcbs'_maxPriority[OF valid_objs'_valid_tcbs']
 
+lemma valid_tcbs_valid_tcbE':
```
lsf37, line 318: Under that name, I'd expect a rule that eliminates `valid_tcbs'` and gives you `valid_tcb'`. Maybe `valid_tcbs_obj_at'` instead?

spec/abstract/IpcCancel_A.thy
```diff
@@ -239,6 +239,21 @@ where
...
 
+definition
+  bind_sc_reply :: "obj_ref \<Rightarrow> obj_ref \<Rightarrow> (unit, 'z::state_ext) s_monad"
+where
+  "bind_sc_reply sc_ptr reply_ptr  = do
+      sc_replies \<leftarrow> liftM sc_replies $ get_sched_context (sc_ptr);
+      \<comment> \<open>unlink head reply and sc before pushing\<close>
+      case sc_replies of
+          [] \<Rightarrow> return ()
+        | (r#_) \<Rightarrow> set_reply_obj_ref reply_sc_update r None;
```
lsf37, line 250:
```suggestion
        | r#_ \<Rightarrow> set_reply_obj_ref reply_sc_update r None;
```
avoid unnecessary parentheses

lsf37: It feels like there should be a neater way to express this case on option, but I'd have to fire up Isabelle to find it.

proof/refine/ARM/Ipc_R.thy
```diff
@@ -3161,6 +3161,341 @@ lemma maybeReturnSc_corres:
...
+lemma thread_set_empty_tcb_sched_context_valid_tcbs[wp]:
+  "thread_set (tcb_sched_context_update Map.empty) t \<lbrace>valid_tcbs\<rbrace>"
+  by (wp thread_set_valid_tcbs) (simp add: valid_tcbs_def valid_tcb_def tcb_cap_cases_def)
+
+lemma threadSet_empty_tcbSchedContext_valid_tcbs'[wp]:
+  "threadSet (tcbSchedContext_update Map.empty) t \<lbrace>valid_tcbs'\<rbrace>"
+  by (wp threadSet_valid_tcbs') (simp add: valid_tcb'_def valid_tcbs'_def tcb_cte_cases_def)
+
+lemma update_valid_tcb[simp]:
+  "valid_tcb ptr tcb (s\<lparr>release_queue := q\<rparr>) = valid_tcb ptr tcb s"
```
lsf37, line 3173: Isn't there some locale that provides these rewrites in Refine if you instantiate `valid_tcb` to it? I can't remember if we did these only for AInvs or not. (Something that says the predicate depends only on the heap). Same for `valid_tcbs'` below.

proof/refine/ARM/Reply_R.thy
```diff
@@ -634,4 +618,486 @@ lemma replyRemoveTCB_invs':
...
+      | r # x \<Rightarrow> f r x)
+   = (when (l \<noteq> []) $ f (hd l) (tl l))"
+  by (clarsimp simp: list_case_If2)
+
+lemma update_sc_reply_stack_update_ko_at'_corres:
+  "corres dc
+     (sc_at ptr)
+     (ko_at' sc' ptr and (\<lambda>s. heap_ls (replyPrevs_of s) replies reply_ptr))
+     (update_sched_context ptr (sc_replies_update (\<lambda>_. reply_ptr)))
+     (setSchedContext ptr (scReply_update (\<lambda>_. replies) sc'))"
```
mktnk3, line 729: I don't quite understand the type of `reply_ptr` and `replies` here. Probably they are meant to be swapped?
Also, having `(\<lambda>s. heap_ls (replyPrevs_of s) replies reply_ptr)` in the concrete assumption is probably a bit odd.  It should be an assumption to the corres rule. This is in a sense part of the state_relation, involving both the abstract and the concrete sides. What it says is that the new head and the new sc_replies queue shoulld satisfy `sc_replies_relation`.

proof/refine/ARM/Ipc_R.thy
```diff
@@ -3884,8 +4219,8 @@ lemma schedContextDonate_if_live_then_nonz_cap':
            simp: conj_ac cong: conj_cong | wp hoare_drop_imps | fastforce simp: tcb_cte_cases_def)+
 
 lemma schedContextDonate_valid_idle':
-  "\<lbrace>\<lambda>s. valid_idle' s \<and> tcbPtr \<noteq> ksIdleThread s \<and>
-        obj_at' (\<lambda>x. scTCB x \<noteq> Some (ksIdleThread s)) scPtr s\<rbrace>
+  "\<lbrace>\<lambda>s. valid_idle' s \<and> tcbPtr \<noteq> idle_thread_ptr \<and>
+        obj_at' (\<lambda>x. scTCB x \<noteq> Some (idle_thread_ptr)) scPtr s\<rbrace>
```
lsf37, line 4223:
```suggestion
        obj_at' (\<lambda>x. scTCB x \<noteq> Some idle_thread_ptr) scPtr s\<rbrace>
```
(remove parentheses)

lsf37 (approved):
Looks good, only minor nitpicks.

proof/refine/ARM/Reply_R.thy
MitchellBuckley: Yes, I've accidentally swapped the names. I'll fix that :+1:

MitchellBuckley: I believe that if we move that predicate to an assumption on the whole rule, it will generate a goal which states that this particular update satisfies `heap_ls` _for all heaps_. The current formulation will instead generate a precondition which only requires that `heap_ls` holds on `replyPrevs_of s`. This makes the assumptions weaker, and thus the lemma stronger. If I move the predicate into the assumption, then this lemma will work in cases where the update is quite harmless, but it will fail to work when proving `heap_ls` really requires knowing something about this particular `replyPrevs_of s`.

proof/refine/ARM/Reply_R.thy
```diff
@@ -634,4 +618,479 @@ lemma replyRemoveTCB_invs':
...
+                     is_reply_def2 set_object_def exec_get put_id_return)
+
+lemma updateReply_corres:
+  assumes
+    rr: "\<And>r1 r2. reply_relation r1 r2 \<Longrightarrow>
+           reply_relation (g (\<lambda>y. new) r1) (f (\<lambda>_. new) r2)"
+  shows
+  "corres dc (reply_at rp) (reply_at' rp and obj_at' (\<lambda>ko. replyPrev_same (f (\<lambda>_. new) ko) ko) rp)
+     (set_reply_obj_ref g rp new)
+     (updateReply rp (f (\<lambda>_. new)))"
```
mktnk3, line 650: I am not sure if we want to call this lemma `updateReply_corres`. This name sounds too generic.
Also, would it be possible to clean up the constraints on the concrete side update? It seems a bit confusing what we want for `new` and what we want for `f`.

proof/refine/ARM/Reply_R.thy
MitchellBuckley: I'll try to find the case where I use it which should hopeful clarify why it is needed.

mktnk3: well, you can have `replyPrevs_of s` as the heap with `\forall s`, so it doesn't need to be for *all_heaps*, right?

mktnk3: but yes, if the way you proved things does require it to be there, I suppose it won't harm to leave it in that way.

mktnk3:
This is great, and it's so good to be done with the corres rule for `reply_push`!

I haven't quite finished reviewing yet, but it feels there's a bit of proliferation of various corres rules with not very descriptive names. Can you please consider cleaning things up around there a bit more?

Also, one simple question - I thought that the basic rule for corres rule names is to use the name of the abstract side function when they exist for both. Isn't this the case?

proof/refine/ARM/Reply_R.thy
MitchellBuckley: The place where I use this is for binding sched contexts and reply objects in `bindScReply` and `bind_sc_reply`. In that case, the particular update is `setSchedContext scPtr (sc { scReply = Just replyPtr })` which is immediately preceded by `updateReply replyPtr (\reply -> reply { replyPrev = scReplyOpt })`. 

Now the state relation states that `∀p replies. sc_replies_of s p = Some replies ⟶ heap_ls (replyPrevs_of s') (scReplies_of s' p) replies` and this relates the concrete-side linked list with abstract-side lists. So fundamentally, updating the `scReply` field of some SchedContext is not safe, unless replyPtr is the head of an appropriate linked list. But then the assumption `\forall s'. heap_ls (replyPrevs_of s')  replyPtr replies` amounts to saying that `replyPtr` is the head of a linked list in all possible universes, which is not true. But it is true in the context of `bindScReply` because the preceding line does the appropriate update.

Does that clarify things at all?

corlewis (approved):
Great job! I just have a few minor comments that haven't already been covered by others.

MitchellBuckley:
> Also, one simple question - I thought that the basic rule for corres rule names is to use the name of the abstract side function when they exist for both. Isn't this the case?

I thought we were naming them after the concrete side. Is there a standard @lsf37 ? Another thing for the style guide.

michaelmcinerney:
> > Also, one simple question - I thought that the basic rule for corres rule names is to use the name of the abstract side function when they exist for both. Isn't this the case?
> 
> I thought we were naming them after the concrete side. Is there a standard @lsf37 ? Another thing for the style guide.

In the conventions.md file (in `l4v/docs`) it does say to use the concrete name, but says this is currently too random and should converge more. I at first misread this and used the abstract name a few times myself

lsf37:
> > > Also, one simple question - I thought that the basic rule for corres rule names is to use the name of the abstract side function when they exist for both. Isn't this the case?
> > 
> > 
> > I thought we were naming them after the concrete side. Is there a standard @lsf37 ? Another thing for the style guide.
> 
> In the conventions.md file (in `l4v/docs`) it does say to use the concrete name, but says this is currently too random and should converge more. I at first misread this and used the abstract name a few times myself

Yes, it's covered in [conventions.md](https://github.com/seL4/l4v/blob/master/docs/conventions.md#corres-rules) and it should be the concrete name.

mktnk3:
> > > > Also, one simple question - I thought that the basic rule for corres rule names is to use the name of the abstract side function when they exist for both. Isn't this the case?
> > > 
> > > 
> > > I thought we were naming them after the concrete side. Is there a standard @lsf37 ? Another thing for the style guide.
> > 
> > 
> > In the conventions.md file (in `l4v/docs`) it does say to use the concrete name, but says this is currently too random and should converge more. I at first misread this and used the abstract name a few times myself
> 
> Yes, it's covered in [conventions.md](https://github.com/seL4/l4v/blob/master/docs/conventions.md#corres-rules) and it should be the concrete name.

Oops, sorry. I think I just lived so far with the impression that there are more lemmas with abstract ones. Then we can at least be consistent with the new functions in MCS with the concrete names.

spec/abstract/IpcCancel_A.thy
MitchellBuckley: It would be like `maybeM` but for lists. Or otherwise using `list_option`.

proof/refine/ARM/Reply_R.thy
mktnk3: Ok, I think I see it, this is not to be considered as directly constituting `sc_replies_relation`. It is just the matter of the particular `replyPrevs_of` heap of the particular instance on the concrete side.

Well, you get this concrete side precondition from the fact that what `replyPtr` is newly linked to previously satisfied `sc_replies_relation`. But this is all in the context, outside of this lemma, I guess.

MitchellBuckley: That's it :+1:

proof/refine/ARM/KHeap_R.thy
```diff
@@ -3641,5 +3747,55 @@ lemma update_sc_no_reply_stack_update_corres:
    apply (wpsimp simp: obj_at_def is_sc_obj_def)+
   done
 
+lemma ko_at'_inj:
+  "ko_at' ko ptr  s \<Longrightarrow> ko_at' ko' ptr s \<Longrightarrow> ko' = ko"
+  by (clarsimp simp: obj_at'_real_def ko_wp_at'_def)
+
+(* FIXME RT: Move these to AInvs where possible *)
+definition injective_ref where
+  "injective_ref ref heap \<equiv> (\<forall>q p1 p2. (p1, ref) \<in> heap q \<and> (p2, ref) \<in> heap q \<longrightarrow> p1 = p2)"
```
mktnk3, line 3756: I think this `injective_ref` is more or less the same concept as `heap_refs_inj`  or similar related ones. There is `sym_refs_inj_sc_tcbs`, which might be quite close to `injective_ref_SCTcb` below. But maybe this one here is easy to use in any case. Could you create a JIRA ticket to investigate this and add that info to the FIXME?

proof/refine/ARM/KHeap_R.thy
```diff
@@ -1444,9 +1519,40 @@ lemma setSchedContext_no_stack_update_corres:
...
+            (setSchedContext ptr (f' sc'))"
+  apply (rule stronger_corres_guard_imp)
+    apply (erule setSchedContext_sc_relation_corres; clarsimp)
+   apply clarsimp
+  apply (clarsimp simp: state_relation_def sc_replies_relation_def)
+  apply (drule_tac x=ptr and y="sc_replies sc" in spec2)
+  apply (clarsimp simp: vs_all_heap_simps obj_at'_real_def ko_wp_at'_def opt_map_def)
+  done
+
+lemma scReply_update_corres:
```
mktnk3, line 1543: This is used only in once place, and, looking at how it is proved, it seems that we could do without this lemma and just use `setSchedContext_sc_relation_corres` directly instead, maybe?

proof/refine/ARM/Ipc_R.thy
```diff
@@ -3161,6 +3161,341 @@ lemma maybeReturnSc_corres:
...
+             split: option.splits Structures_A.kernel_object.splits)
+  done
+
+lemma schedContextDonate_corres_helper:
+  "(case rv' of SwitchToThread x \<Rightarrow> when (x = t \<or> t = cur) rescheduleRequired
+                             | _ \<Rightarrow> when (t = cur) rescheduleRequired) =
+   (when (t = cur \<or> (case rv' of SwitchToThread x \<Rightarrow> t = x | _ \<Rightarrow> False)) rescheduleRequired)"
+  by (case_tac rv'; clarsimp simp: when_def)
+
+lemma schedContextDonate_corres:
```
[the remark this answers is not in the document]
mktnk3: yes, I'll be moving this to Reply_R, I think.

proof/refine/ARM/KHeap_R.thy
MitchellBuckley: Yes, this acheives the same as the heap_refs version, but that version depends on heaps which are not defined for all references. It's definitely possible that I could replace `injective_ref_SCTcb` with `sym_refs_inj_sc_tcbs` . What kind of JIRA issue did you have in mind? Just doing that replacement?

proof/refine/ARM/KHeap_R.thy
MitchellBuckley: I spent a few hours on this, but it's a lot of work to unify things. The reason they look similar is that they're both based on the generic `set_object` corres proofs. But in each case because the state-relation is slightly non-trivial on these particular kernel objects, the generic argument doesn't quite work. In the future it would be nice to unify these somehow, but it would require quite a lot of time to get it working.

proof/refine/ARM/Ipc_R.thy
MitchellBuckley: I gave it a shot but could not immediately see how it should be done (I'm not expert with locales yet). I've added a `FIXME RT` instead.

spec/haskell/src/SEL4/Object/Reply.lhs
```diff
@@ -34,51 +34,31 @@ This module specifies the behavior of reply objects.
...
+> bindScReply :: PPtr SchedContext -> PPtr Reply -> Kernel ()
+> bindScReply scPtr replyPtr = do
+>     sc <- getSchedContext scPtr
+>     scReplyOpt <- return $ scReply sc
+>     when (scReplyOpt /= Nothing) $ do
+>         scReplyPtr <- return $ fromJust scReplyOpt
+>         updateReply scReplyPtr (\reply -> reply { replyNext = Just (Next replyPtr) })
+>     updateReply replyPtr (\reply -> reply { replyPrev = scReplyOpt })
+>     setSchedContext scPtr (sc { scReply = Just replyPtr })
+>     updateReply replyPtr (\reply -> reply { replyNext = Just (Head $ scPtr) })
```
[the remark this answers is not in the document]
MitchellBuckley: My plan with this was to just sort out the differences in CRefine. I haven't worked in CRefine, but I was under the impression that slight reorderings were not too hard to accomodate because the proofs are close to symbolic executions anyway.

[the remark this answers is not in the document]
MitchellBuckley: https://sel4.atlassian.net/browse/VER-1380

proof/refine/ARM/Ipc_R.thy
mktnk3: `valid_tcbs` is added since we started Refine, and it hasn't made its way into AInvs locale yet. In AInvs, we do have such a locale for `valid_objs`, which is very nice, so we should definitely incorporate `valid_tcbs` in there somehow at some point, but maybe not now.

Still, Ipc_R is probably not the right place for this, and there are a few `valid_tcbs` update listed in `TcbAcc_R` (where `valid_tcbs` is defined). So, can you move this there and clean things up a bit while you are here? There is VER1343 that talks about sorting out the `valid_objs/valid_tcbs` interaction, and it would be also good if we add something regarding locale there.

proof/refine/ARM/Ipc_R.thy
```diff
@@ -3161,6 +3161,341 @@ lemma maybeReturnSc_corres:
...
+  "threadSet (tcbSchedContext_update Map.empty) t \<lbrace>valid_tcbs'\<rbrace>"
+  by (wp threadSet_valid_tcbs') (simp add: valid_tcb'_def valid_tcbs'_def tcb_cte_cases_def)
+
+lemma update_valid_tcb[simp]:
+  "valid_tcb ptr tcb (s\<lparr>release_queue := q\<rparr>) = valid_tcb ptr tcb s"
+  "valid_tcb ptr tcb (s\<lparr>reprogram_timer := rt\<rparr>) = valid_tcb ptr tcb s"
+  by (auto simp: valid_tcb_def valid_tcb_state_def valid_bound_obj_def valid_arch_tcb_def
+          split: Structures_A.thread_state.splits option.splits)
+
+lemma update_valid_tcbs[simp]:
```
mktnk3, line 3178: These `valid_tcbs'` updates can go in TcbAcc_R, too.

proof/refine/ARM/KHeap_R.thy
mktnk3: Doing some investigation and finding a most clean and minimal solution would be good.

proof/refine/ARM/Reply_R.thy
```diff
@@ -634,4 +618,479 @@ lemma replyRemoveTCB_invs':
...
+  "\<lbrace>\<lambda>s. reply_at' rptr s \<longrightarrow>
+          P (obj_at' (\<lambda>ko. if rptr = p then Q (f ko) else Q ko) p s)\<rbrace>
+   updateReply rptr f
+   \<lbrace>\<lambda>rv s. P (obj_at' Q p s)\<rbrace>"
+  apply (simp only: obj_at'_real_def updateReply_def)
+  apply (wp set_reply'.ko_wp_at)
+  apply (auto simp: ko_wp_at'_def obj_at'_def projectKO_eq projectKO_reply split: if_splits)
+  done
+
+(* RT MOVE: to go with delta_sym_refs *)
```
mktnk3, line 949: Please move

proof/refine/ARM/Ipc_R.thy
```diff
@@ -4141,240 +4281,198 @@ lemma tcbSchedEnqueue_valid_tcb_state'[wp]:
...
-crunches rescheduleRequired, scheduleTCB
-  for valid_tcb_state'[wp]: "valid_tcb_state' ts"
-  (wp: crunch_wps)
-
-crunches setThreadState
-  for sc_at'_n[wp]: "sc_at'_n n p"
-  and sc_ko_at'[wp]: "\<lambda>s. P (ko_at' (sc :: sched_context) p s)"
-  (wp: crunch_wps simp: crunch_simps)
-
-definition replyPush' where
```
mktnk3, line 4153: I have been cleaning things up with this PR (and mine to remove overlaps), and noticed that there are a fair bit of unnecessary changes introduced in this PR. I think we want to revert and cleanup those.

Firstly, I don’t see why we need to remove `replyPush’`. This is Ryan’s work that led to a nice wp rule for `valid_objs’` and others. Consequently, in the below, these wp rules have added preconditions that we shouldn’t need. Please re-establish Ryan’s results for these wp rules, and for that, I believe resurrecting `replyPush’` is the shortest path.

proof/refine/ARM/Ipc_R.thy
```diff
@@ -4141,240 +4281,198 @@ lemma tcbSchedEnqueue_valid_tcb_state'[wp]:
...
+   bindScReply scPtr replyPtr
+   \<lbrace>\<lambda>_. valid_idle'\<rbrace>"
+  unfolding bindScReply_def
+  by (wpsimp wp: hoare_vcg_imp_lift' hoare_vcg_all_lift set_reply'.obj_at')
+
 lemma replyPush_valid_idle':
-  "\<lbrace>valid_idle' and valid_pspace' and st_tcb_at' active' callerPtr\<rbrace>
+  "\<lbrace>valid_idle'
+    and valid_pspace'
+    and valid_objs'
```
mktnk3, line 4415: we don’t need `valid_objs’` here.

proof/refine/ARM/Ipc_R.thy
mktnk3: I thought that the spec change in this PR is semantically equivalent, but I can now see that there is not just the `bindScReply` part but also the part where asserts are removed. I understand that this does mean that we might now need `reply_at’ replyPtr` in the precondition. But I think it should be just that and we shouldn't need `tcb_at’` for `caller` and `callee`. The `list_refs_of_replies’` lemma below also doesn’t need any of the added preconditions, I believe.
And, at this point, I am not entirely sure how much `replyPush’` can help in these. It is certainly possible to update it with the spec change for `replyPush`, but I now think that it is most likely that the issue of redundant preconditions can be sorted out without it.

MitchellBuckley: Ok, cool I'll check it out and maybe we can chat more later today.

corlewis: I think I agree with Miki's second comment. I admit that I didn't understand the `replyPush'` proof very well, but the changes in this PR made other work much easier. In particular, I was finding it extremely hard to prove `replyPush_valid_replies'` before this PR was merged, and it was nice and straightforward after.
If we could strengthen the lemmas by weakening the preconditions then that would clearly be nice, but I'd be very sad to see this reverted, and without reverting I think resurrecting `replyPush'` might be tricky.

proof/refine/ARM/Ipc_R.thy
MitchellBuckley: You're right! Fixed this.

proof/refine/ARM/Reply_R.thy
```diff
@@ -634,4 +618,453 @@ lemma replyRemoveTCB_invs':
...
+
+lemma updateReply_replyTCB_valid_objs'[wp]:
+  "\<lbrace>valid_objs' and case_option \<top> (\<lambda>t. tcb_at' t) p\<rbrace>
+   updateReply rptr (replyTCB_update (\<lambda>_. p))
+   \<lbrace>\<lambda>_. valid_objs'\<rbrace>"
+  by (wpsimp wp: set_reply'.valid_objs' simp: updateReply_def)
+     (auto simp: valid_obj'_def valid_reply'_def obj_at'_def projectKO_eq projectKO_reply
+          split: option.splits)
+
+lemma updateReply_valid_objs':
```
mktnk3, line 888: We don’t really need this lemma. There is already `updateReply_valid_objs’_preserved_strong` and `updateReply_valid_objs’_preserved` earlier in this file. They should serve us enough. This one here has `reply_at’` in the implication of the precondition, bundled up with the requirement of preserving `valid_obj’`, which would be fiddly to deal with.

proof/refine/ARM/Reply_R.thy
```diff
@@ -634,4 +618,453 @@ lemma replyRemoveTCB_invs':
...
+  "\<lbrace>valid_objs'
+    and (\<lambda>s. \<forall>ko. ko_at' ko rptr s \<longrightarrow> valid_reply' (f ko) s)\<rbrace>
+   updateReply rptr f
+   \<lbrace>\<lambda>_. valid_objs'\<rbrace>"
+  apply (wpsimp wp: set_reply'.valid_objs' simp: updateReply_def)
+  by (fastforce simp: valid_obj'_def)
+
+lemma replyNext_update_Head_valid_objs'[wp]:
```
mktnk3, line 896: This one and a few below which are instantiations for specific cases of the above `updateReply_valid_objs’` are also something we could do without, if we use the `xxx_preserved` lemmas. I don’t think it is a great idea to chucking all these in the wp set, either.

proof/refine/ARM/Reply_R.thy
mktnk3: We should only need `updateReply_valid_objs’_preserved_strong`, and it’s probably good to rename it to `updateReply_valid_objs’` and remove the other.

MitchellBuckley: Sorry about this. I think that Reply_R was previously not imported into IPC_R? And so when I was first writing these lemmas `find_theorems` could not find them. And by the time I moved things to `Reply_R` and created the import, I missed the overlap.

MitchellBuckley: Removed it.

proof/refine/ARM/Reply_R.thy
mktnk3: These cases come up as nuisance I think because `set_sc_valid_objs’` and the friends are in the wp set. We should take them out of the wp set.

MitchellBuckley: Happy to remove them if needed, but can you explain a little more about why they're a nuisance? That surprises me because they look almost like weakest preconditions to me.

mktnk3: I am just guessing that you wrote these because you had to stop at these points while working on your main proof (like `bindScReply_valid_objs’ below), rather than just resolving the subgoals in place. But it is probably not because of what I said. It is more because of the `ko_at’` in your version of `updateReply_valid_objs’` that require a bit of work (but not much). With `updateReply_valid_objs’_preserved_strong`, these are just instantiations of it with a bit of unfolding, and I don’t think we need to list all these specific lemmas when we have a more abstract one.

Still, I think we still want to remove `set_sc_valid_objs’` and others from the wp set. They unfold `valid_obj’` unnecessarily, when wpsimp can make more progress with `valid_obj’`.

mktnk3: Mitch and I agreed that, for these lemmas, leaving them as they are now would be the best option for now (least work required).

PR #181, opened 2020-12-28
whileLoops for refill_budget_check and refill_unblock_check

michaelmcinerney:
To make refinement possible (or at least tractable), I have used `whileLoop` in the definition of `refill_budget_check` and `refill_unblock_check` in both the abstract spec and the Haskell. 

There wasn't any change in the semantics for these functions, so there isn't anything terribly interesting going on in the proofs. The proofs in AInvs are much more verbose, because we cannot simply use `set_refills_wp` in many of the proofs as before. I have made some attempt to shorten the proofs (for example, the `foo_simple` methods) but any other ideas that would help with this would be great. In Refine, I have unbundled the proofs for `invs'`. This should make the other PR I have which adds `refill_unblock_check` to many places much more straightforward.

This relies on some of the work I did for `awaken`. I'll need to rebase that branch, and then this branch. And there will be a failure for compiling the Haskell kernel as in the `awaken` PR. Hopefully Gerwin and I can look at this soon


spec/abstract/document/git-root.tex
```diff
@@ -0,0 +1 @@
+/home/mci186/v/l4v
```
michaelmcinerney, line 1: Oops this file shouldn't be here

spec/abstract/SchedContext_A.thy
```diff
@@ -157,31 +157,55 @@ where
...
 where
-  "refills_merge_prefix [] = []"
-| "refills_merge_prefix [r] = [r]"
-| "refills_merge_prefix (r1 # r2 # rs) =
-     (if can_merge_refill r1 r2
-      then refills_merge_prefix (merge_refill r1 r2 # rs)
-      else r1 # r2 # rs)"
+  "refill_head_overlapping_loop sc_ptr
+     = whileLoop (\<lambda>_ s. the (fun_of_m (refill_head_overlapping sc_ptr) s))
+                 (\<lambda>_. merge_refills sc_ptr) ()"
```
MitchellBuckley, line 194: This definition is really neat!! So compact and clear!

spec/abstract/SchedContext_A.thy
```diff
@@ -195,17 +219,30 @@ where
                         \<lparr>r_time = r_time (hd (tl refills)), r_amount = r_amount (hd (tl refills)) + usage\<rparr>]
    od"
 
-fun
-  MIN_BUDGET_merge :: "refill list \<Rightarrow> refill list"
+definition head_insufficient :: "obj_ref \<Rightarrow> (bool, 'z::state_ext) s_monad"
+where
+  "head_insufficient sc_ptr \<equiv> do
+     refills \<leftarrow> get_refills sc_ptr;
+     if r_amount (hd refills) < MIN_BUDGET then return True else return False
```
MitchellBuckley, line 226: same note as above about `if _ then True else False`

proof/invariant-abstract/RealTime_AI.thy
```diff
@@ -1054,12 +1054,19 @@ crunches
   (wp_del: get_sc_active_wp get_sc_refill_ready_wp get_sc_refill_sufficient_wp
            get_sc_released_wp refill_full_wp)
 
+(* FIXME RT Michael: move to lib *)
+lemma whileLoop_wp':
+  "(\<And>r. \<lbrace> \<lambda>s. I r s \<and> C r s \<rbrace> B r \<lbrace> I \<rbrace>) \<Longrightarrow>
+  \<lbrace> I r \<rbrace> whileLoop C B r \<lbrace> I \<rbrace>"
+  apply (fastforce intro: whileLoop_wp)
+  done
+
```
MitchellBuckley, line 1063: Since `whileLoop_wp'` is used in so many crunches proofs, is it sensible to add it to `wp` or `crunch_wps` instead of adding it in so many places? In the cases were you don't want to use it, which would be few, you could `supply whileLoop_wp' [wp del]`. What do you think?

proof/refine/ARM/IpcCancel_R.thy
```diff
@@ -1821,8 +1821,7 @@ lemma tcbReleaseRemove_valid_release_queue'[wp]:
   done
 
 crunches setReprogramTimer
-  for valid_objs'[wp]: valid_objs'
```
MitchellBuckley, line 1824: nitpick: Was this line removed on purpose?

MitchellBuckley: Ah, looks like it was moved.

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -9664,8 +10529,8 @@ lemma refill_unblock_check_valid_blocked_except_set[wp]:
    \<lbrace>\<lambda>rv. valid_blocked_except_set S\<rbrace>"
   unfolding refill_unblock_check_def
   apply (wpsimp wp: set_refills_wp get_refills_wp is_round_robin_wp)
-  apply (clarsimp simp: vs_all_heap_simps valid_blocked_except_set_2_def valid_blocked_thread_def)
-  apply (drule_tac x=t in spec, fastforce simp: obj_at_def)
+  apply (fastforce simp: vs_all_heap_simps valid_blocked_except_set_2_def valid_blocked_thread_def
+                        obj_at_def)
```
MitchellBuckley, line 10533: nitpick: one more space needed

MitchellBuckley (approved):
This looks like lots of work! I didn't read every detail of every proof, but I really like the approach and I think it looks good! I especially like the `foo_simple` methods.

spec/abstract/SchedContext_A.thy
```diff
@@ -157,31 +157,55 @@ where
...
                       in (butlast original) @ [new_last])"
 
-fun
-  refills_merge_prefix :: "refill list \<Rightarrow> refill list"
+definition
+  refill_pop_head :: "obj_ref \<Rightarrow> (refill, 'z::state_ext) s_monad"
+where
+  "refill_pop_head sc_ptr \<equiv> do
+     refills \<leftarrow> get_refills sc_ptr;
+     assert (0 < length refills);
```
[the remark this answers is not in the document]
michaelmcinerney: I know that this assert is very useful in the Refine version, but yeah, maybe I can get away without this in the aspec. I'll check it out

proof/invariant-abstract/RealTime_AI.thy
michaelmcinerney: I'm not even sure how to add it to `crunch_wps`. @corlewis can you let me know what you think of this, and how to do it if it's a good idea?

corlewis: Hmm, at the moment the only way to add things to `crunch_wps` is by manually adding it when it is first created, which is currently in Bits_AI. This would require you following through with the move to lib.

Separately, I definitely don't think it should go in`wp` but it might be ok in `crunch_wps`. In general we really only want safe rules in `crunch_wps` and especially in `wp`. I haven't looked at this PR enough yet to really tell how safe this rule is, but my guess would have been that this can only be used in cases where the body doesn't really interact with the condition. More complicated cases will need `whileLoop_wp` and I think it's too confusing to expect people to know to delete `whileLoop_wp'` first.

The reason that I think it might be ok for `crunch_wps` is that you're probably only going to be able to crunch functions with loops when it's for a property that's independent of the loop. Anything else seems like it would already be too complicated for crunch without some specific helper lemmas. @lsf37 what do you think?

lsf37: Yes, that makes sense to me. We probably don't want loop rules in general `wp`, but for invariants in `crunch_wps` it would make sense.

PR #182, opened 2021-01-03
MCS AInvs for RISCV

lsf37:
This PR adds `ASpec`, `Haskell`, `design-spec`, and `AInvs` to MCS. `Refine` is still to do.

PR #184, opened 2021-01-04
rt haskell: revert accidental edit

lsf37:
This should have been part of #182, but I somehow managed to drop the commit.


mktnk3 (approved):
thanks for the quick fix - it must be just because of how busy you are, probably having multiple things going on simultaneously :)

lsf37:
> thanks for the quick fix - it must be just because of how busy you are, probably having multiple things going on simultaneously :)

No worries. I thought I was taking it slow yesterday, but I rushed the squash part, which I shouldn't have done.

PR #189, opened 2021-01-06
rt haskell fixes (KernelP, etc.)

mktnk3:
almost trivial changes

- use `()` instead of `irq` in  KernelP
- add `wordBitsCase` for `replySizeBits`

No need for proof update. The `KernelP` part should go into the master as well at some point.


michaelmcinerney (approved):
Thanks!

corlewis (approved):

lsf37 (approved):
Looks good!

PR #190, opened 2021-01-06
rt: more consistent thread_state discriminators

corlewis:
It's not clear that this is much nicer, but it at least makes the names and locations of these functions more consistent.

Unfortunately, using the standard discriminators and selectors that the datatype package can generate led to confusing issues with unrelated proofs breaking. I wasn't able to fix this and ended up manually defining them.

The actual goal that started me down this path was wanting a nice way to define `is_reply_state`.

### DIFF
--- a/proof/invariant-abstract/Invariants_AI.thy
+++ b/proof/invariant-abstract/Invariants_AI.thy
@@ -1064,8 +1064,7 @@ lemmas valid_replies_defs = valid_replies_2_def replies_with_sc_def replies_bloc
 
 abbreviation
   "fault_tcb_states st \<equiv>
-     (\<exists>ep data. st = BlockedOnSend ep data) \<or> (\<exists>ep. st = BlockedOnReply ep) \<or>
-     st = Inactive"
+     is_blocked_on_send st \<or> is_blocked_on_reply st \<or> st = Inactive"
 
 definition
   "fault_tcbs_valid_states_except_set TS \<equiv>
@@ -1340,6 +1339,9 @@ abbreviation
                  st = Running \<or>
                  st = Restart \<or>
                  idle st \<or> awaiting_reply st"
+
+abbreviation
+  "is_reply_state st \<equiv> is_blocked_on_reply st \<or> is_blocked_on_receive st"
 
 abbreviation
   "ct_active \<equiv> ct_in_state active"
@@ -1618,7 +1620,6 @@ lemma tcb_st_refs_of_simps[simp]: (* ARMHYP add TCBHypRef? *)
  "tcb_st_refs_of (Running)               = {}"
  "tcb_st_refs_of (Inactive)              = {}"
  "tcb_st_refs_of (Restart)               = {}"
-(* "tcb_st_refs_of (YieldTo t)             = {(t, TCBYieldTo)}"*)
  "tcb_st_refs_of (BlockedOnReply r')      = {(r', TCBReply)}"
  "tcb_st_refs_of (IdleThreadState)       = {}"
  "\<And>x. tcb_st_refs_of (BlockedOnReceive x r data)  = (if bound r then {(x, TCBBlockedRecv), (the r, TCBReply)} else {(x, TCBBlockedRecv)})"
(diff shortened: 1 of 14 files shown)


spec/abstract/Structures_A.thy
```diff
@@ -384,13 +384,71 @@ datatype thread_state
...
   | Restart
-(*  | YieldTo obj_ref (* sc ref *)*)
   | BlockedOnReceive obj_ref "obj_ref option" receiver_payload
   | BlockedOnSend obj_ref sender_payload
   | BlockedOnReply obj_ref
   | BlockedOnNotification obj_ref
   | IdleThreadState
 
+(* FIXME RT: generating the following discriminators and selectors automatically breaks
+             unrelated proofs in strange ways *)
```
lsf37, line 394: How strange? I like the new consistency, but getting that from the datatype package would definitely be a bonus.

lsf37 (approved):
I think it's an improvement. It'd be nice to eventually get the definitions from the datatype package, but I'm happy with merging this as a step on the way.

spec/abstract/Structures_A.thy
corlewis: It was before the break so I don't quite remember the details, but it was something to do with unification. Some proofs could be fixed by changing `force` to `auto`, but there was one in IpcDet_AI that I couldn't get working.

The strange part was that there were no obvious `thread_states` in the proof, and `apply_trace` didn't show any relevant rules being used.

lsf37: Once this is merged, we could try again with the updates you already have and compare what the datatype package generates to what you have here. Maybe that'll give us a hint.

MitchellBuckley:
> Unfortunately, using the standard discriminators and selectors that the datatype package can generate led to confusing issues with unrelated proofs breaking. I wasn't able to fix this and ended up manually defining them.

I had this problem too. Very odd. I expect the explanation will be simple when someone has the time to figure it out.

proof/invariant-abstract/Invariants_AI.thy
```diff
@@ -1064,8 +1064,7 @@ lemmas valid_replies_defs = valid_replies_2_def replies_with_sc_def replies_bloc
 
 abbreviation
   "fault_tcb_states st \<equiv>
-     (\<exists>ep data. st = BlockedOnSend ep data) \<or> (\<exists>ep. st = BlockedOnReply ep) \<or>
-     st = Inactive"
+     is_blocked_on_send st \<or> is_blocked_on_reply st \<or> st = Inactive"
```
MitchellBuckley, line 1067: smallest-nitpick-ever: I think that `inactive st` is an abbreviation for `st = Inactive` which would make this abbreviation ever so slightly neater. But don't bother changing this unless you want to.

MitchellBuckley (approved):
Love it!

mktnk3 (approved):
This is very nice. Once this is merged, we should use these abbreviation in other existing PRs.
It would be good to create a JIRA issue stating that we should investigate the interaction with the datatype package? And I suppose you've update AInvs thoroughly, right, Corey?

VictorPhan (approved):

corlewis:
> 
> 
> This is very nice. Once this is merged, we should use these abbreviation in other existing PRs.
> It would be good to create a JIRA issue stating that we should investigate the interaction with the datatype package? And I suppose you've update AInvs thoroughly, right, Corey?

A JIRA issue has been created at https://sel4.atlassian.net/browse/VER-1381.

I went pretty thoroughly through AInvs but I might have missed some uses and there are other existing discriminators that I didn't touch. For example, we might be able to replace `awaiting_reply`, or turn it into a rule giving an alternative definition for `is_blocked_on_reply`.

proof/invariant-abstract/Invariants_AI.thy
corlewis: I decided not to do this, mainly because `inactive` is currently very rarely used. I'll leave changing that to a future round of consolidating these discriminators.

PR #191, opened 2021-01-07
Remove send_ipc assertion

MitchellBuckley:
Remove an assertion from send_ipc (ASpec)

### DIFF
--- a/proof/invariant-abstract/DetSchedSchedule_AI.thy
+++ b/proof/invariant-abstract/DetSchedSchedule_AI.thy
@@ -11276,9 +11276,6 @@ lemma send_ipc_valid_ready_qs:
   apply (rule hoare_seq_ext[OF _ gsc_sp], simp add: pred_conj_def obj_at_kh_kheap_simps pred_map_eq_normalise)
   apply (rule hoare_seq_ext[OF _ thread_get_sp])
   apply (rule_tac B="\<lambda>_ s. valid_ready_qs s \<and> released_if_bound_sc_tcb_at dest s" in hoare_seq_ext)
-   apply (rule hoare_seq_ext[OF _ gbn_inv])
-   apply (rule hoare_seq_ext_skip, wpsimp)
-   apply (rule hoare_seq_ext[OF _ assert_inv])
    apply (wpsimp wp: possible_switch_to_valid_ready_qs'' set_thread_state_valid_ready_qs
                      set_thread_state_pred_map_tcb_sts_of)
   apply (simp add: pred_conj_def obj_at_def)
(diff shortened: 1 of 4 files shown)


michaelmcinerney (approved):
Niiiiiiiiice

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -17298,8 +17269,10 @@ lemma send_ipc_ct_in_state:
    \<lbrace>\<lambda>_. ct_in_state P :: 'state_ext state \<Rightarrow> _\<rbrace>"
   supply if_split [split del]
   unfolding send_ipc_def
-  apply (wpsimp wp: sts_ctis_neq maybeM_inv)
-            apply (wpsimp wp: hoare_vcg_imp_lift' hoare_vcg_all_lift get_tcb_obj_ref_wp)
+  apply wpsimp
+       apply ((wpsimp wp: sts_ctis_neq)+)[3]
+    apply wpsimp
+            apply (wpsimp wp: sts_ctis_neq)
```
corlewis, line 17275: That looks a bit unpleasant. What does the middle wpsimp use instead of `sts_ctis_neq`?

proof/invariant-abstract/IpcDet_AI.thy
```diff
@@ -1900,6 +1893,10 @@ lemma si_invs'_helper_no_reply:
   apply (fastforce simp: live_def elim!: if_live_then_nonz_capD2)
   done
 
+lemma active_st_tcb_at_awaiting_replyD:
+  "st_tcb_at active t s \<Longrightarrow> st_tcb_at (\<lambda>x. \<not> awaiting_reply x) t s"
+  by (fastforce simp: pred_tcb_at_def obj_at_def split: thread_state.splits)
```
corlewis, line 1898: I'm generally not a fan of such specific thread_state lemmas. What's the use case, and is it possible to use `pred_tcb_weakenE` instead?

corlewis (approved):
Nice, that should make things simpler!

proof/invariant-abstract/IpcDet_AI.thy
MitchellBuckley: Good point, fixed it :+1:

proof/invariant-abstract/DetSchedSchedule_AI.thy
MitchellBuckley: It was there to break the wpsimp loop at a specific point, but a little investigation has revealed a simpler proof :) Thanks for pointing it out!

mktnk3 (approved):
Nice :)

lsf37 (approved):
👍

PR #192, opened 2021-01-08
readme: use -R instead of -l

corlewis:


lsf37 (approved):
Cool, thanks for updating that.

PR #195, opened 2021-01-11
rt: update unbindFromSC

corlewis:
This updates the Haskell so that `unbindFromSC` checks the `scYieldFrom` field before accessing it and also makes it better match the abstract.


VictorPhan (approved):

mktnk3 (approved):
Thanks for chasing this up. So, this will be followed by a C spec change, right?

corlewis:
> 
> 
> Thanks for chasing this up. So, this will be followed by a C spec change, right?

Yep, at some point. The C actually runs correctly at the moment, so strictly speaking it isn't needed, but it's probably worth making it more closely match the Haskell and abstract.

PR #196, opened 2021-01-11
rt: disable sorry words in thylint action

corlewis:


.github/workflows/pr.yml
```diff
@@ -43,4 +43,5 @@ jobs:
         repo-token: "${{ secrets.GITHUB_TOKEN }}"
         title: 'File annotations for theory linter'
         input: './annotations.json'
+        disable: sorry
```
lsf37, line 46: The line has to go one higher into the previous step. I.e.
```
- uses: seL4/ci-actions/thylint@master
  with:
    disable: sorry
```

lsf37 (approved):
I'll approve now so you can merge directly when you've pushed the change.

.github/workflows/pr.yml
corlewis: Oh right, of course

PR #197, opened 2021-01-12
Small decode sorries

michaelmcinerney:
Signed-off-by: Michael McInerney <Michael.McInerney@data61.csiro.au>

### DIFF
--- a/proof/refine/ARM/Tcb_R.thy
+++ b/proof/refine/ARM/Tcb_R.thy
@@ -3143,6 +3143,13 @@ lemma checkPrio_lt_ct_weak:
   apply (clarsimp simp: pred_tcb_at'_def obj_at'_def)
   by (rule le_ucast_ucast_le) simp
 
+lemma checkPrio_lt_ct_weak':
+  "\<lbrace>P\<rbrace> checkPrio prio auth \<lbrace>\<lambda>rv s. P s \<and> mcpriority_tcb_at' (\<lambda>mcp. ucast prio \<le> mcp) auth s\<rbrace>, -"
+  apply (wpsimp wp: hoare_vcg_conj_liftE1)
+    apply (wpsimp wp: checkPrio_inv)
+   apply (wpsimp wp: checkPrio_lt_ct_weak)+
+  done
+
 crunches checkPrio
   for tcb_at'[wp]: "tcb_at' t"
   and ex_nonz_cap_to'[wp]: "ex_nonz_cap_to' t"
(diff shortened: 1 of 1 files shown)


proof/refine/ARM/Tcb_R.thy
```diff
@@ -3143,6 +3143,13 @@ lemma checkPrio_lt_ct_weak:
   apply (clarsimp simp: pred_tcb_at'_def obj_at'_def)
   by (rule le_ucast_ucast_le) simp
 
+lemma checkPrio_lt_ct_weak':
+  "\<lbrace>P\<rbrace> checkPrio prio auth \<lbrace>\<lambda>rv s. P s \<and> mcpriority_tcb_at' (\<lambda>mcp. ucast prio \<le> mcp) auth s\<rbrace>, -"
+  apply (wpsimp wp: hoare_vcg_conj_liftE1)
+    apply (wpsimp wp: checkPrio_inv)
+   apply (wpsimp wp: checkPrio_lt_ct_weak)+
+  done
```
MitchellBuckley, line 3151: nitpick: merge into one wpsimp if possible (but I know it's not always possible).

MitchellBuckley (approved):
Noice. You're a machine!

proof/refine/ARM/Tcb_R.thy
michaelmcinerney: No dice, unfortunately

mktnk3 (approved):
Nice. This will put us below 80 sorries :)

proof/refine/ARM/Tcb_R.thy
```diff
@@ -3143,6 +3143,13 @@ lemma checkPrio_lt_ct_weak:
   apply (clarsimp simp: pred_tcb_at'_def obj_at'_def)
   by (rule le_ucast_ucast_le) simp
 
+lemma checkPrio_lt_ct_weak':
+  "\<lbrace>P\<rbrace> checkPrio prio auth \<lbrace>\<lambda>rv s. P s \<and> mcpriority_tcb_at' (\<lambda>mcp. ucast prio \<le> mcp) auth s\<rbrace>, -"
```
corlewis, line 3147: Ooh, that's a nice way of solving the problem we discussed. I guess that means that the idea of replacing `checkPrio_inv` with several more specific lemmas didn't work though?

corlewis (approved):

proof/refine/ARM/Tcb_R.thy
michaelmcinerney: Yeah that quickly became annoying. The `wp` produced some postconditions where the preemption case was not simply `True`, and I thought the approach to fixing it was worse than the slightly more manual forwards proof. This proof does take quite a while to get through the case spliting in the first few lines, but I think it's much clearer what's going on

PR #200, opened 2021-01-13
Simple page table duplicate sorries

michaelmcinerney:
Signed-off-by: Michael McInerney <Michael.McInerney@data61.csiro.au>

### DIFF
--- a/proof/refine/ARM/PageTableDuplicates.thy
+++ b/proof/refine/ARM/PageTableDuplicates.thy
@@ -374,9 +374,8 @@ lemma mapM_x_storePTE_update_helper:
       apply (drule pspace_alignedD')
        apply simp
       apply (simp add:objBits_simps' archObjSize_def pteBits_def
-        is_aligned_weaken[where y = 2] pageBits_def pdeBits_def
+        is_aligned_weaken[where y = 2] pageBits_def pdeBits_def vs_ptr_align_def
         split:kernel_object.splits arch_kernel_object.splits)
-  sorry (*
      apply (simp add:mask_lower_twice)
      apply (drule mask_out_first_mask_some[where m = ptBits])
       apply (simp add:vs_ptr_align_def
@@ -403,7 +402,7 @@ lemma mapM_x_storePTE_update_helper:
   apply (clarsimp split:option.splits)
   apply (drule_tac p' = y in valid_duplicates'_D)
      apply simp+
-  done *)
+  done
 
 lemma page_directory_at_pde_atD':
   "\<lbrakk>page_directory_at' p s;is_aligned p' 2; p' && ~~ mask pdBits = p\<rbrakk> \<Longrightarrow> pde_at' p' s"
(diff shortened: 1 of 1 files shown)


proof/refine/ARM/PageTableDuplicates.thy
```diff
@@ -1168,38 +1166,46 @@ lemma createObject_valid_duplicates'[wp]:
...
-                 word_bits_conv archObjSize_def is_aligned_mask
-          split: ARM_H.object_type.splits)+)[1]
-  apply (clarsimp simp: word_bits_def)*)
- sorry
-
+     apply fastforce
+    apply (simp add: vs_entry_align_def)
+   apply simp
+  apply clarsimp
+  apply (drule(2) valid_duplicates'_update) prefer 3
```
MitchellBuckley, line 1206: nitpick: This prefer 3 is not necessary?

proof/refine/ARM/PageTableDuplicates.thy
```diff
@@ -1293,48 +1299,49 @@ lemma valid_duplicates_deleteObjects_helper:
...
-   apply fastforce+
+           apply fastforce+
   apply (rename_tac arch_kernel_object)
   apply (case_tac arch_kernel_object)
-    apply fastforce+
-   apply (clarsimp split:ARM_H.pte.splits)
-   apply auto[1]
+      apply fastforce+
+     apply (clarsimp split:ARM_H.pte.splits)
+     apply auto[1]
```
MitchellBuckley, line 1310: nitpick: remove this auto and others like it.

MitchellBuckley (approved):
Noice.

proof/refine/ARM/PageTableDuplicates.thy
michaelmcinerney: No bueno

proof/refine/ARM/PageTableDuplicates.thy
michaelmcinerney: Ur, this file is kinda messed up. Horrible style. Autos all over the shop. Real ugly to look at, to boot. So I've kind of made a minimal effort to make it nice.

I'd make a JIRA issue to say that we should clean it up, but I'm sure it's not the only file that's like it...

MitchellBuckley: Mmmm, is it ok if we at least try not to increase the number of autos?

proof/refine/ARM/PageTableDuplicates.thy
MitchellBuckley: Interesting! Perhaps there is some kind of schematic unification that has to happen in a particular order. Cool, no worries.

proof/refine/ARM/PageTableDuplicates.thy
michaelmcinerney: Having a closer look at this proof, I fixed the indentation of some of the lines (which is where one of the "new" autos is coming from), and elsewhere, it was mainly a rearrangement of the existing proof. So I haven't added any autos. 

I admit that I do not fully understand what's going on in some of these proofs, so it might take a fair while to figure out how to remove these autos. @mktnk3 do you have any preference for what I should do here?

mktnk3: All these autos are pre-existing ones, and I bet they are in the master, too. So, no need to go removing them right now. We should remove other sorries instead.

You can change your settings to hide white space changes. Then you can't even see these comments if you do that on the diff.

PageTableDuplicates is a pretty messy file to start with. It would be good to do a good clean up when we can afford to. I wouldn't be surprised if there is an existing JIRA issue for it.

proof/refine/ARM/PageTableDuplicates.thy
mktnk3: Actually, `apply (fastforce elim!: valid_duplicates'_update simp: vs_entry_align_def)` works fine instead.

corlewis (approved):
Looks good!

PR #201, opened 2021-01-13
More decode sorries

michaelmcinerney:
This should bring the Haskell in line with the abstract and C. It would be great if someone could closely check the Haskell to make sure it's OK, but given that the corres rules are working, I'm fairly sure it's fine

### DIFF
--- a/proof/refine/ARM/Tcb_R.thy
+++ b/proof/refine/ARM/Tcb_R.thy
@@ -3584,23 +3584,23 @@ lemma cap_CNode_case_throw:
      = (doE unlessE (isCNodeCap cap) (throw x); m odE)"
   by (cases cap, simp_all add: isCap_simps unlessE_def)
 
-lemma decode_set_space_corres:
+lemma decodeCVSpace_corres:
   notes if_cong [cong]
   shows
- "\<lbrakk> cap_relation cap cap'; list_all2 (\<lambda>(c, sl) (c', sl'). cap_relation c c' \<and> sl' = cte_map sl) extras extras';
-      is_thread_cap cap \<rbrakk> \<Longrightarrow>
-  corres (ser \<oplus> tcbinv_relation)
-         (invs and valid_cap cap and (\<lambda>s. \<forall>x \<in> set extras. cte_at (snd x) s))
-         (invs' and valid_cap' cap'  and (\<lambda>s. \<forall>x \<in> set extras'. cte_at' (snd x) s))
-      (decode_set_space args cap slot extras)
-      (decodeSetSpace args cap' (cte_map slot) extras')"
-  apply (simp    add: decode_set_space_def decodeSetSpace_def
+ "\<lbrakk>cap_relation cap cap';
+   list_all2 (\<lambda>(c, sl) (c', sl'). cap_relation c c' \<and> sl' = cte_map sl) extras extras';
+   is_thread_cap cap\<rbrakk> \<Longrightarrow>
+  corres (ser \<oplus> (\<lambda>abs_inv conc_inv. tcbinv_relation abs_inv conc_inv))
+      (invs and valid_cap cap and (\<lambda>s. \<forall>x \<in> set extras. cte_at (snd x) s))
+      (invs' and valid_cap' cap' and (\<lambda>s. \<forall>x \<in> set extras'. cte_at' (snd x) s))
+      (decode_cv_space args cap slot extras)
+      (decodeCVSpace args cap' (cte_map slot) extras')"
+  apply (simp    add: decode_cv_space_def decodeCVSpace_def
                       Let_def
            split del: if_split)
-  apply (cases "3 \<le> length args \<and> 2 \<le> length extras'")
+  apply (cases "2 \<le> length args \<and> 2 \<le> length extras'")
    apply (clarsimp simp: val_le_length_Cons list_all2_Cons2
               split del: if_split)
-  sorry (*
    apply (simp add: liftE_bindE liftM_def
                     getThreadCSpaceRoot getThreadVSpaceRoot
                  split del: if_split)
@@ -3618,10 +3618,10 @@ lemma decode_set_space_corres:
                             apply (rule corres_trivial)
                             apply (clarsimp simp: returnOk_def newroot_rel_def is_cap_simps
                                                   list_all2_conv_all_nth split_def)
-                           apply (unfold unlessE_whenE)
+                            apply (unfold unlessE_whenE)
                            apply (rule corres_whenE)
                              apply (case_tac vroot_cap', simp_all add:
-                                              is_valid_vtable_root_def isValidVTableRoot_def
+                                              is_valid_vtable_root_def
                                               ARM_H.isValidVTableRoot_def)[1]
                              apply (rename_tac arch_cap)
                              apply (clarsimp, case_tac arch_cap, simp_all)[1]
@@ -3645,7 +3645,7 @@ lemma decode_set_space_corres:
               apply (rule corres_trivial, simp)
              apply simp
             apply (unfold whenE_def, wp+)[2]
-          apply (clarsimp simp: is_cap_simps get_tcb_vtable_ptr_def cte_map_tcb_1[simplified] objBits_defs)
+          apply (clarsimp simp: is_cap_simps get_tcb_vtable_ptr_def cte_map_tcb_1[simplified])
          apply simp
          apply (wp hoare_drop_imps)+
        apply (clarsimp simp: is_cap_simps get_tcb_ctable_ptr_def cte_map_tcb_0)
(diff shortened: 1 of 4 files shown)


proof/refine/ARM/Tcb_R.thy
```diff
@@ -3547,14 +3549,66 @@ lemma decode_set_space_corres:
...
+     apply wp+
+   apply fastforce+
+  done
+
+lemma decodeCVSpace_wf[wp]:
+  "\<lbrace>invs' and tcb_at' t and ex_nonz_cap_to' t and cte_at' slot
+      and (\<lambda>s. \<forall>x \<in> set extras. s \<turnstile>' fst x \<and> cte_at' (snd x) s \<and> t \<noteq> snd x \<and> t + 16 \<noteq> snd x)\<rbrace>
+     decodeCVSpace args (ThreadCap t) slot extras
+   \<lbrace>tcb_inv_wf'\<rbrace>,-"
+  apply (simp       add: decodeCVSpace_def Let_def split_def
```
MitchellBuckley, line 3591: nitpick: check indentation in this lemma. Seems a little odd, but maybe it's just a style that you prefer.

proof/refine/ARM/Tcb_R.thy
```diff
@@ -3475,23 +3475,25 @@ lemma cap_CNode_case_throw:
...
-         (invs and valid_cap cap and (\<lambda>s. \<forall>x \<in> set extras. cte_at (snd x) s))
-         (invs' and valid_cap' cap'  and (\<lambda>s. \<forall>x \<in> set extras'. cte_at' (snd x) s))
-      (decode_set_space args cap slot extras)
-      (decodeSetSpace args cap' (cte_map slot) extras')"
-  apply (simp    add: decode_set_space_def decodeSetSpace_def
+ "\<lbrakk>cap_relation cap cap';
+   list_all2 (\<lambda>(c, sl) (c', sl'). cap_relation c c' \<and> sl' = cte_map sl) extras extras';
+   is_thread_cap cap\<rbrakk> \<Longrightarrow>
+  corres (ser \<oplus> (\<lambda>abs_inv conc_inv. tcbinv_relation abs_inv conc_inv
+                                     \<and> Invocations_A.is_ThreadControlCaps abs_inv
```
MitchellBuckley, line 3485: nitpick: indentation

proof/refine/ARM/Tcb_R.thy
```diff
@@ -3547,14 +3549,66 @@ lemma decode_set_space_corres:
...
+         (invs and valid_cap cap and (\<lambda>s. \<forall>x \<in> set extras. cte_at (snd x) s))
+         (invs' and valid_cap' cap' and (\<lambda>s. \<forall>x \<in> set extras'. cte_at' (snd x) s))
+      (decode_set_space args cap slot extras)
+      (decodeSetSpace args cap' (cte_map slot) extras')"
+  apply (simp add: decode_set_space_def decodeSetSpace_def check_handler_ep_def unlessE_whenE)
+  apply (cases "\<not> (2 \<le> length args \<and> 3 \<le> length extras')")
+   apply (clarsimp dest!: list_all2_lengthD split: list.split)
+   apply fastforce
+  apply (clarsimp simp: val_le_length_Cons list_all2_Cons2
+              split del: if_split)
```
MitchellBuckley, line 3568: nitpick: indentation

proof/refine/ARM/Tcb_R.thy
```diff
@@ -3569,9 +3623,20 @@ lemma decodeSetSpace_wf[wp]:
   apply (simp del: length_greater_0_conv add: valid_updateCapDataI)
   done
 
+lemma decodeCVpace_inv[wp]:
+  "\<lbrace>P\<rbrace> decodeCVSpace args cap slot extras \<lbrace>\<lambda>rv. P\<rbrace>"
```
MitchellBuckley, line 3627: Nitpick: use `my_function {P}` notation.
Also, can this be done with crunch?

MitchellBuckley (approved):
Very nice! I trust you on the Haskell spec.

proof/refine/ARM/Tcb_R.thy
```diff
@@ -3547,14 +3549,66 @@ lemma decode_set_space_corres:
...
+         apply (clarsimp simp: returnOk_def newroot_rel_def is_cap_simps list_all2_conv_all_nth)
+        apply blast
+       apply fastforce+
+     apply wp+
+   apply fastforce+
+  done
+
+lemma decodeCVSpace_wf[wp]:
+  "\<lbrace>invs' and tcb_at' t and ex_nonz_cap_to' t and cte_at' slot
+      and (\<lambda>s. \<forall>x \<in> set extras. s \<turnstile>' fst x \<and> cte_at' (snd x) s \<and> t \<noteq> snd x \<and> t + 16 \<noteq> snd x)\<rbrace>
```
mktnk3, line 3588: I am just wondering if we can replace `16` with something more abstract. I can see `16` also in `decodeSetSpace_wf` below, which is an existing lemma, and this number is different (32) for X64 and RISCV. My guess is this is `2^slot_bits` but can someone maybe confirm/check, please?

mktnk3 (approved):
Thanks for looking into this. I am glad that everything worked out nicely.

I thought I had reviewed this long time ago, but apparently I forgot to press submit...

proof/refine/ARM/Tcb_R.thy
```diff
@@ -3614,42 +3713,54 @@ lemma decodeSetSpace_tc_slot[wp]:
   apply simp
   done
 
-lemma decode_tcb_conf_corres:
+lemma decodeTCBConfigure_corres:
```
mktnk3, line 3716: We probably want to defer updating the names of existing lemmas (those that are also in master) for now, just to save the extra work it might generate while merging. We’ll clean them up altogether at some point.

But we can be consistent and proper when we name the rules for the functions we are adding newly for MCS.

proof/refine/ARM/Tcb_R.thy
michaelmcinerney: If I can generalise this, do I also update `decodeSetSpace_wf`, or is that bad for the merge?

proof/refine/ARM/Tcb_R.thy
michaelmcinerney: Well, the merge is going to be horrific anyway, right? My vote is to change whatever we like now and deal with an ever so slightly worse merge later. I've found the cost of not following the conventions is actually quite high; if you think there should be a lemma called `some_lemma`, but it's actually called `some_Lemma_haha` then it's quite annoying

spec/haskell/src/SEL4/Object/TCB.lhs
```diff
@@ -341,17 +341,31 @@ This is to ensure that the source capability is not made invalid by the deletion
...
+>         tcCapsFaultHandler = Nothing,
+>         tcCapsTimeoutHandler = Nothing,
+>         tcCapsCRoot = Just cRoot,
+>         tcCapsVRoot = Just vRoot,
+>         tcCapsBuffer = Nothing }
+> decodeCVSpace _ _ _ _ = throw TruncatedMessage
+
+> decodeSetSpace :: [Word] -> Capability -> PPtr CTE ->
+>         [(Capability, PPtr CTE)] -> KernelF SyscallError TCBInvocation
+> decodeSetSpace (cRootData:vRootData:_) cap slot (cRootArg:vRootArg:fhArg:_) = do
```
corlewis, line 356: The C code has the `fhArg` in the original first extra cap slot. I can see a pretty strong argument for why this order is preferred, so that it's more consistent with the non MCS code, but is there a corresponding change being made to the C?

proof/refine/ARM/Tcb_R.thy
```diff
@@ -3475,23 +3475,25 @@ lemma cap_CNode_case_throw:
...
-         (invs' and valid_cap' cap'  and (\<lambda>s. \<forall>x \<in> set extras'. cte_at' (snd x) s))
-      (decode_set_space args cap slot extras)
-      (decodeSetSpace args cap' (cte_map slot) extras')"
-  apply (simp    add: decode_set_space_def decodeSetSpace_def
+ "\<lbrakk>cap_relation cap cap';
+   list_all2 (\<lambda>(c, sl) (c', sl'). cap_relation c c' \<and> sl' = cte_map sl) extras extras';
+   is_thread_cap cap\<rbrakk> \<Longrightarrow>
+  corres (ser \<oplus> (\<lambda>abs_inv conc_inv. tcbinv_relation abs_inv conc_inv
+                                     \<and> Invocations_A.is_ThreadControlCaps abs_inv
+                                     \<and> is_ThreadControlCaps conc_inv))
```
corlewis, line 3486: I might be wrong, but I think it's a bit unusual to have these sort of conditions in the return relation of a corres statement. Would it be possible to prove them as post conditions of a separate hoare triple?
This is pretty minor, so I'm fine with leaving it as is if having them here is more useful.

corlewis (approved):
Awesome work as always! Apart from my one comment about the corresponding C code, the Haskell looks good to me.

spec/haskell/src/SEL4/Object/TCB.lhs
michaelmcinerney: Ah, good catch. I suppose for this part, I was happily following along with `decode_set_space` in the abstract. But I think that's wrong. So I'll change both

mktnk3: That is a good catch. Thanks Corey. I can see that C has the FH first, not sure how abstract came to have that order.

proof/refine/ARM/Tcb_R.thy
VictorPhan: Tracking this issue here: [VER-1385](https://sel4.atlassian.net/browse/VER-1385)

proof/refine/ARM/Tcb_R.thy
VictorPhan: Tracking issue: [VER-1386](https://sel4.atlassian.net/browse/VER-1386)

VictorPhan (approved):

PR #185, opened 2021-01-05
rt reply pop tweaks

mktnk3:
This PR contains some spec updates for `repplyPop` and the proof fixes

- use `updateReply` where possible
- contains Victor’s commit from #175: move `schedContextDonate` to just before `cleanReply`
- move `threadGet` to just before `schedContextDonate`

PR #203, opened 2021-01-15
rt: use updateReply everywhere

corlewis:
This changes all of the remaining `getReply; setReply` patterns to use `updateReply`, except for `replyPop` which is left for https://github.com/seL4/l4v/pull/185.

### DIFF
--- a/spec/haskell/src/SEL4/Object/Reply.lhs
+++ b/spec/haskell/src/SEL4/Object/Reply.lhs
@@ -14,7 +14,7 @@ This module specifies the behavior of reply objects.
 \begin{impdetails}
 
 % {-# BOOT-IMPORTS: SEL4.Machine SEL4.Model SEL4.Object.Structures #-}
-% {-# BOOT-EXPORTS: replyRemove replyRemoveTCB replyPush replyUnlink getReply setReply #-}
+% {-# BOOT-EXPORTS: replyRemove replyRemoveTCB replyPush replyUnlink updateReply #-}
 
 > import {-# SOURCE #-} SEL4.Kernel.Thread(getThreadState, setThreadState)
 > import SEL4.Machine.RegisterSet(PPtr)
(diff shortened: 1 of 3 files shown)


spec/haskell/src/SEL4/Object/Reply.lhs
```diff
@@ -28,6 +28,11 @@ This module specifies the behavior of reply objects.
 
 \end{impdetails}
 
+> getReply :: PPtr Reply -> Kernel Reply
+> getReply rptr = getObject rptr
+
+> setReply :: PPtr Reply -> Reply -> Kernel ()
+> setReply rptr r = setObject rptr r
```
mktnk3, line 35: Do we need to move this?

mktnk3 (approved):
#185 has the same spec change for `replyRemove`, but in this case, the rebase should work cleanly, hopefully. There’s some overlap in `replyRemove_valid_objs`, but I suppose it won’t be too bad. Please keep in mind that both Victor and myself are working on top of #185 at the moment.

corlewis:
> 
> 
> #185 has the same spec change for `replyRemove`, but in this case, the rebase should work cleanly, hopefully. There’s some overlap in `replyRemove_valid_objs'`, but I suppose it won’t be too bad. Please keep in mind that both Victor and myself are working on top of #185 at the moment.

Sorry, I missed that #185 also updated `replyRemove`. I've removed it from this PR, which now just has the SchedContext changes I was originally thinking about.

spec/haskell/src/SEL4/Object/Reply.lhs
corlewis: Yeah, it's not needed but it made sense to me to have them next to `updateReply`, since that's now the main place that they're used. I'm happy to undo it if you'd prefer not.

mktnk3: Ah, I see. That makes sense! No need to move, happy for them to stay here.

VictorPhan (approved):

PR #205, opened 2021-01-18
ainvs cleanup: requalify some arch lemmas proved in ArchRetype_AI

mktnk3:
This PR is for the master branch.

While working on the preemption_point spec update for MCS, I noticed some odd requalification work-around at the end of ArchRetype_AI, when it is perfectly fine to do requalification normally.

For MCS, it makes sense to clean this up on this occasion, and I thought I might as well do it for the master, too.

I am running the testboard here: [testboard](https://bamboo.ts.data61.csiro.au/browse/L4V-TESTBOARD3)
but it looks like something is not quite right with it. I think the testing will be operating in a useful way, but it is running on its own new branch. Is this what is expected??

### DIFF
--- a/proof/invariant-abstract/ARM/ArchRetype_AI.thy
+++ b/proof/invariant-abstract/ARM/ArchRetype_AI.thy
@@ -1354,24 +1354,13 @@ lemma valid_arch_mdb_detype:
          (\<lambda>p. if fst p \<in> untyped_range cap then None else caps_of_state s p)"
   by auto
 
-end
-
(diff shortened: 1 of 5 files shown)


lsf37 (approved):
Nice, thanks for cleaning that up.

Since the the test board has moved to git (instead of hg) it now creates a new branch for every new test. That part is fine. It looks like the `send email` part of it is still trying hg commands, though and is failing because of that. I'll have a look at this, but it's Ok to ignore for now.

From my side the PR is ready to merge.

PR #207, opened 2021-01-20
decodeSchedContext lemmas

michaelmcinerney:
See the commit messages for a quick rundown of what's here

### DIFF
--- a/proof/refine/ARM/SchedContextInv_R.thy
+++ b/proof/refine/ARM/SchedContextInv_R.thy
@@ -5,7 +5,7 @@
  *)
 
 theory SchedContextInv_R
-imports Ipc_R
+imports Invocations_R Tcb_R
 begin
 
 context begin interpretation Arch . (*FIXME: arch_split*)
@@ -54,30 +54,76 @@ primrec valid_sc_ctrl_inv' :: "sched_control_invocation \<Rightarrow> kernel_sta
 
 primrec sc_inv_rel :: "Invocations_A.sched_context_invocation \<Rightarrow> sched_context_invocation \<Rightarrow> bool"
   where
-  "sc_inv_rel (Invocations_A.InvokeSchedContextConsumed sc bf) sci' =
-   (sci' = InvokeSchedContextConsumed sc bf)"
-| "sc_inv_rel (Invocations_A.InvokeSchedContextBind sc cap) sci' =
-   (\<exists>cap'. cap_relation cap cap' \<and> sci' = InvokeSchedContextBind sc cap')"
-| "sc_inv_rel (Invocations_A.InvokeSchedContextUnbindObject sc cap) sci' =
-   (\<exists>cap'. cap_relation cap cap' \<and> sci' = InvokeSchedContextUnbindObject sc cap')"
-| "sc_inv_rel (Invocations_A.InvokeSchedContextUnbind sc cap) sci' =
-   (sci' = InvokeSchedContextUnbind sc)" \<comment> \<open>FIXME RT: remove cap in abstract InvokeSchedContextUnbind\<close>
-| "sc_inv_rel (Invocations_A.InvokeSchedContextYieldTo sc bf) sci' =
-   (sci' = InvokeSchedContextYieldTo sc bf)"
+  "sc_inv_rel (Invocations_A.InvokeSchedContextConsumed sc_ptr bf) sci' =
+   (sci' = InvokeSchedContextConsumed sc_ptr bf)"
+| "sc_inv_rel (Invocations_A.InvokeSchedContextBind sc_ptr cap) sci' =
+   (\<exists>cap'. cap_relation cap cap' \<and> sci' = InvokeSchedContextBind sc_ptr cap')"
+| "sc_inv_rel (Invocations_A.InvokeSchedContextUnbindObject sc_ptr cap) sci' =
+   (\<exists>cap'. cap_relation cap cap' \<and> sci' = InvokeSchedContextUnbindObject sc_ptr cap')"
+| "sc_inv_rel (Invocations_A.InvokeSchedContextUnbind sc_ptr) sci' =
+   (sci' = InvokeSchedContextUnbind sc_ptr)"
+| "sc_inv_rel (Invocations_A.InvokeSchedContextYieldTo sc_ptr bf) sci' =
+   (sci' = InvokeSchedContextYieldTo sc_ptr bf)"
 
 primrec sc_ctrl_inv_rel ::
   "Invocations_A.sched_control_invocation \<Rightarrow> sched_control_invocation \<Rightarrow> bool" where
-  "sc_ctrl_inv_rel (Invocations_A.InvokeSchedControlConfigure sc budget period refills badge) sci' =
-    (sci' = InvokeSchedControlConfigure sc budget period refills badge)"
+  "sc_ctrl_inv_rel (Invocations_A.InvokeSchedControlConfigure sc_ptr budget period refills badge) sci' =
+    (sci' = InvokeSchedControlConfigure sc_ptr budget period refills badge)"
 
-(* FIXME RT: preconditions can be reduced, this is what is available at the call site: *)
-lemma decodeSchedContextInvocation_wf:
-  "\<lbrace> valid_cap' (SchedContextCap sc n) and invs' and sch_act_simple and ex_nonz_cap_to' sc and
-     (\<lambda>s. \<forall>cap\<in>set excaps. \<forall>r\<in>zobj_refs' cap. ex_nonz_cap_to' r s) and
-     (\<lambda>s. \<forall>x\<in>set excaps. valid_cap' x s) \<rbrace>
-   decodeSchedContextInvocation label sc excaps args
+lemma decodeSchedContext_Bind_wf:
+  "\<lbrace>\<lambda>s. \<exists>n. valid_cap' (SchedContextCap sc_ptr n) s
+        \<and> ex_nonz_cap_to' sc_ptr s
+        \<and> (\<forall>cap\<in>set excaps. \<forall>r\<in>zobj_refs' cap. ex_nonz_cap_to' r s)
+        \<and> (\<forall>x\<in>set excaps. valid_cap' x s)\<rbrace>
+   decodeSchedContext_Bind sc_ptr excaps
    \<lbrace>valid_sc_inv'\<rbrace>, -"
-  sorry
+  apply (clarsimp simp: decodeSchedContext_Bind_def)
+  apply (wpsimp wp: gts_wp' threadGet_wp getNotification_wp
+              simp: scReleased_def scActive_def isBlocked_def refillReady_def)
+  apply (clarsimp simp: valid_cap'_def)
+  apply (drule_tac x="hd excaps" in bspec, fastforce dest: hd_in_set)+
+  apply (fastforce simp: pred_tcb_at'_def obj_at'_def)
+  done
+
+lemma decodeSchedContext_UnbindObject_wf:
+  "\<lbrace>\<lambda>s. \<exists>n. valid_cap' (SchedContextCap sc_ptr n) s
+        \<and> ex_nonz_cap_to' sc_ptr s
+        \<and> (\<forall>cap\<in>set excaps. \<forall>r\<in>zobj_refs' cap. ex_nonz_cap_to' r s)
+        \<and> (\<forall>x\<in>set excaps. valid_cap' x s)\<rbrace>
+   decodeSchedContext_UnbindObject sc_ptr excaps
+   \<lbrace>valid_sc_inv'\<rbrace>, -"
+  apply (clarsimp simp: decodeSchedContext_UnbindObject_def)
+  apply (wpsimp wp: gts_wp' threadGet_wp getNotification_wp
+              simp: scReleased_def scActive_def isBlocked_def refillReady_def)
+  apply (clarsimp simp: valid_cap'_def)
+  apply (drule_tac x="hd excaps" in bspec, fastforce dest: hd_in_set)+
+  apply (fastforce simp: pred_tcb_at'_def obj_at'_def)
+  done
+
+lemma decodeSchedContext_YieldTo_wf:
+  "\<lbrace>\<lambda>s. \<exists>n. valid_cap' (SchedContextCap sc_ptr n) s \<and> ex_nonz_cap_to' sc_ptr s\<rbrace>
+   decodeSchedContext_YieldTo sc_ptr args
+   \<lbrace>valid_sc_inv'\<rbrace>, -"
+  apply (clarsimp simp: decodeSchedContext_YieldTo_def)
+  apply (wpsimp wp: gts_wp' threadGet_wp getNotification_wp getTCB_wp
+              simp: scReleased_def scActive_def isBlocked_def refillReady_def)
+  apply (clarsimp simp: valid_cap'_def)
+  apply (clarsimp simp: pred_tcb_at'_def obj_at'_def projectKOs)
+  done
+
+lemma decodeSchedContextInvocation_wf:
+  "\<lbrace>\<lambda>s. \<exists>n. valid_cap' (SchedContextCap sc_ptr n) s
+        \<and> ex_nonz_cap_to' sc_ptr s
+        \<and> (\<forall>cap\<in>set excaps. \<forall>r\<in>zobj_refs' cap. ex_nonz_cap_to' r s)
+        \<and> (\<forall>x\<in>set excaps. valid_cap' x s)\<rbrace>
+   decodeSchedContextInvocation label sc_ptr excaps args
+   \<lbrace>valid_sc_inv'\<rbrace>, -"
+  apply (simp add: decodeSchedContextInvocation_def)
+  apply (wpsimp wp: decodeSchedContext_Bind_wf
+                    decodeSchedContext_UnbindObject_wf
+                    decodeSchedContext_YieldTo_wf)
+  apply (fastforce dest: valid_SchedContextCap_sc_at')
+  done
 
 (* FIXME RT: preconditions can be reduced, this is what is available at the call site: *)
 lemma decodeSchedControlInvocation_wf:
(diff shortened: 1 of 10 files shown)


proof/refine/ARM/SchedContextInv_R.thy
```diff
@@ -70,14 +70,65 @@ primrec sc_ctrl_inv_rel ::
...
+   decodeSchedContext_YieldTo sc args
+   \<lbrace>valid_sc_inv'\<rbrace>, -"
+  apply (clarsimp simp: decodeSchedContext_YieldTo_def)
+  apply (wpsimp wp: gts_wp' threadGet_wp getNotification_wp getTCB_wp
+              simp: scReleased_def scActive_def isBlocked_def refillReady_def)
+  apply (clarsimp simp: valid_cap'_def)
+  apply (clarsimp simp: pred_tcb_at'_def obj_at'_def projectKOs)
+  done
+
+lemma valid_SchedContextCap_sc_at':
```
MitchellBuckley, line 112: nitpick: perhaps move this to Invariants_H or somewhere else higher up from here.

proof/refine/ARM/SchedContextInv_R.thy
```diff
@@ -88,14 +139,187 @@ lemma decodeSchedControlInvocation_wf:
...
+     apply (clarsimp simp: valid_cap'_def)
+    apply (rule liftE_validE[THEN iffD2, OF thread_get_sp])
+   apply (rule liftE_validE[THEN iffD2, OF threadGet_sp])
+  apply (rule corres_splitEE_skip; (solves \<open>wpsimp simp: valid_cap'_def obj_at'_def\<close>)?)
+   apply (corressimp corres: get_ntfn_corres
+                       simp: get_sk_obj_ref_def sc_relation_def)
+  apply (rule corres_guard_imp)
+    apply (rule corres_split_eqrE)
+       apply (rule corres_splitEE)
+          apply (rule whenE_throwError_corres)
```
MitchellBuckley, line 186: nitpick: consider `apply (rule xxx; simp)` here to shorten?

proof/refine/ARM/SchedContextInv_R.thy
```diff
@@ -88,14 +139,187 @@ lemma decodeSchedControlInvocation_wf:
...
+            apply simp
+           apply simp
+          apply (clarsimp simp: returnOk_def)
+         apply (subst corres_liftE_rel_sum)
+         apply (rule corres_rel_imp)
```
MitchellBuckley, line 191: nitpick: instead of two rules and simp, consider `apply (rule xxx[OF yyy], simp)` or similar.

MitchellBuckley (approved):
Great work! Very neat!

spec/abstract/Decode_A.thy
```diff
@@ -515,47 +509,62 @@ where
...
-          ct_ptr \<leftarrow> liftE $ gets cur_thread;
-          whenE (tcb_ptr = ct_ptr) $ throwError IllegalOperation;
-          prios \<leftarrow> liftE $ thread_get tcb_priority tcb_ptr;
-          ct_mcp \<leftarrow> liftE $ thread_get tcb_mcpriority ct_ptr;
-          whenE (prios > ct_mcp) $ throwError IllegalOperation
-      odE;
-\<^cancel>\<open>     whenE (sc_yield_from sc \<noteq> None) $ throwError IllegalOperation;\<close>
       ct_ptr \<leftarrow> liftE $ gets cur_thread;
+      whenE (the (sc_tcb sc) = ct_ptr) $ throwError IllegalOperation;
+      whenE (sc_tcb sc = None) $ throwError IllegalOperation;
```
corlewis, line 544: It's a bit confusing but I think that we want to either reverse these lines so that we know that we can safely use `the (sc_tcb sc)`, or change the first check to `sc_tcb sc = Some ct_ptr`. Otherwise it's possible to get some unintentional behaviour with `undefined`.

My guess is that as it is currently we might run in to problems with this in crefine.

spec/abstract/Decode_A.thy
```diff
@@ -515,47 +509,62 @@ where
...
+where
+  "decode_sched_context_invocation label sc_ptr excaps buffer \<equiv>
+  case gen_invocation_type label of
+    SchedContextConsumed \<Rightarrow> returnOk $ InvokeSchedContextConsumed sc_ptr buffer
+  | SchedContextBind \<Rightarrow> decode_sched_context_bind sc_ptr excaps
+  | SchedContextUnbindObject \<Rightarrow> decode_sched_context_unbind_object sc_ptr excaps
+  | SchedContextUnbind \<Rightarrow> doE
+      sc \<leftarrow> liftE $ get_sched_context sc_ptr;
+      ct_ptr \<leftarrow> liftE $ gets cur_thread;
+      whenE (the (sc_tcb sc) = ct_ptr) $ throwError IllegalOperation;
```
corlewis, line 565: Similar comment here, I would use `sc_tcb sc = Some ct_ptr`.

spec/abstract/Decode_A.thy
corlewis: If I'm correct in wanting to make this change, we also need to update the Haskell.

spec/haskell/src/SEL4/Object/TCB.lhs
```diff
@@ -689,12 +689,14 @@ The domain cap is invoked to set the domain of a given TCB object to a given val
...
 >     ctPtr <- withoutFailure $ getCurThread
 >     when (fromJust (scTCB sc) == ctPtr) $ throw IllegalOperation
->     ct <- withoutFailure $ getObject ctPtr
+>     when (scTCB sc == Nothing) $ throw IllegalOperation
 >     priority <- withoutFailure $ threadGet tcbPriority $ fromJust $ scTCB sc
->     when (priority > tcbMCP ct) $ throw IllegalOperation
+>     ct_mcp <- withoutFailure $ threadGet tcbMCP ctPtr
+>     when (priority > ct_mcp) $ throw IllegalOperation
+>     yt_ptr <- withoutFailure $ threadGet tcbYieldTo ctPtr
+>     when (yt_ptr /= Nothing) $ throw IllegalOperation
```
corlewis, line 699: Is this a new condition to the C as well? I don't see it at the moment, but I might have missed it in a branch or PR.

proof/refine/ARM/SchedContextInv_R.thy
```diff
@@ -70,14 +70,65 @@ primrec sc_ctrl_inv_rel ::
   "sc_ctrl_inv_rel (Invocations_A.InvokeSchedControlConfigure sc budget period refills badge) sci' =
     (sci' = InvokeSchedControlConfigure sc budget period refills badge)"
 
-(* FIXME RT: preconditions can be reduced, this is what is available at the call site: *)
+lemma decodeSchedContext_Bind_wf:
+  "\<lbrace>valid_cap' (SchedContextCap sc n) and ex_nonz_cap_to' sc
+    and (\<lambda>s. \<forall>cap\<in>set excaps. \<forall>r\<in>zobj_refs' cap. ex_nonz_cap_to' r s)
+    and (\<lambda>s. \<forall>x\<in>set excaps. valid_cap' x s)\<rbrace>
+   decodeSchedContext_Bind sc excaps
+   \<lbrace>valid_sc_inv'\<rbrace>, -"
```
corlewis, line 78: Do you mind changing the `sc`'s in these lemmas to something like `scptr`? It's a minor thing, but I find it much easier to follow when it's immediately clear what's a pointer and what's an object.

proof/refine/ARM/SchedContextInv_R.thy
```diff
@@ -70,14 +70,65 @@ primrec sc_ctrl_inv_rel ::
...
+lemma valid_SchedContextCap_sc_at':
+  "\<lbrakk>valid_cap' (capability.SchedContextCap sc n) s\<rbrakk> \<Longrightarrow> sc_at' sc s"
+  apply (clarsimp simp: valid_cap'_def obj_at'_real_def)
+  apply (rule ko_wp_at'_weakenE)
+   apply (fastforce simp: objBits_simps
+                   split: kernel_object.splits)+
+  done
+
 lemma decodeSchedContextInvocation_wf:
-  "\<lbrace> valid_cap' (SchedContextCap sc n) and invs' and sch_act_simple and ex_nonz_cap_to' sc and
```
corlewis, line 75: Nice job simplifying the precondition!

proof/refine/ARM/SchedContextInv_R.thy
```diff
@@ -70,14 +70,65 @@ primrec sc_ctrl_inv_rel ::
...
+  "\<lbrace>valid_cap' (SchedContextCap sc n) and ex_nonz_cap_to' sc
+    and (\<lambda>s. \<forall>cap\<in>set excaps. \<forall>r\<in>zobj_refs' cap. ex_nonz_cap_to' r s)
+    and (\<lambda>s. \<forall>x\<in>set excaps. valid_cap' x s)\<rbrace>
    decodeSchedContextInvocation label sc excaps args
    \<lbrace>valid_sc_inv'\<rbrace>, -"
-  sorry
+  apply (simp add: decodeSchedContextInvocation_def)
+  apply (wpsimp wp: decodeSchedContext_Bind_wf[where n=n]
+                    decodeSchedContext_UnbindObject_wf[where n=n]
+                    decodeSchedContext_YieldTo_wf[where n=n])
```
corlewis, line 129: It sometimes ends up causing more problems than it helps, but you can avoid having to specify `n` here by changing these lemmas to be `∃n. valid_cap' (SchedContextCap sc n) ...`

corlewis (approved):
Great job fixing these functions up! I have a couple of comments that I think we want to consider for the future, but none of them are blocking for this PR.

spec/abstract/Decode_A.thy
```diff
@@ -515,47 +509,62 @@ where
...
+where
+  "decode_sched_context_unbind_object sc_ptr excaps \<equiv> doE
       whenE (length excaps = 0) $ throwError TruncatedMessage;
       cap \<leftarrow> returnOk $ hd excaps;
       case cap of
         ThreadCap tcb_ptr \<Rightarrow> doE
-          sc_tcb_opt \<leftarrow> liftE $ get_sc_obj_ref sc_tcb sc_ptr;
-          whenE (sc_tcb_opt \<noteq> Some tcb_ptr) $ throwError IllegalOperation;
+          sc \<leftarrow> liftE $ get_sched_context sc_ptr;
+          whenE (sc_tcb sc \<noteq> Some tcb_ptr) $ throwError IllegalOperation;
```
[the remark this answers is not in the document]
michaelmcinerney: I think I just preferred using the `corres` rule for `get_sched_context`. But I'll have another look at these things and give it another shot

spec/haskell/src/SEL4/Object/TCB.lhs
michaelmcinerney: Yes, this is on Curtis's list of things to do

proof/refine/ARM/SchedContextInv_R.thy
michaelmcinerney: Oh yeah, I normally do do this. I'll change these

michaelmcinerney:
I've asked Curtis if he can rearrange `decodeSchedContext_YieldTo` so that the check that the `scTCB` is not `NULL` comes first

spec/haskell/src/SEL4/Object/TCB.lhs
mktnk3: Can you please include here the C change information when it becomes available? And when you do that, can you paste in the commit message as well? If it is a PR of its own, that would be good. The way things are done on the C side at the moment is very fluid and hard to keep track of. It would be good to find a better way of organising things, but that is a separate issue, I suppose.

spec/abstract/Decode_A.thy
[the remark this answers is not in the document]
mktnk3: I think you can just unfold `get_sc_obj_ref` to see `get_sched_context`, maybe with simp or something.

spec/abstract/Decode_A.thy
```diff
@@ -515,47 +509,62 @@ where
...
-      cap \<leftarrow> returnOk $ hd excaps;
-      tcb_ptr \<leftarrow> returnOk $ obj_ref_of cap;
-      ct_ptr \<leftarrow> liftE $ gets cur_thread;
-      whenE (tcb_ptr = ct_ptr) $ throwError IllegalOperation;
-      returnOk $ InvokeSchedContextUnbind sc_ptr cap odE
-  | SchedContextYieldTo \<Rightarrow> doE
+    odE"
+
+definition
+  decode_sched_context_yield_to ::
```
mktnk3, line 537: Nice to have this.

spec/abstract/Schedule_A.thy
```diff
@@ -295,8 +295,8 @@ where
         set_sc_obj_ref sc_yield_from_update sc_ptr (Some ct_ptr);
         set_tcb_obj_ref tcb_yield_to_update ct_ptr (Some sc_ptr);
         tcb_sched_action tcb_sched_dequeue tcb_ptr;
-        tcb_sched_action tcb_sched_enqueue tcb_ptr;
         tcb_sched_action tcb_sched_enqueue ct_ptr;
+        tcb_sched_action tcb_sched_enqueue tcb_ptr;
```
mktnk3, line 299: Thanks for catching this.

proof/refine/ARM/Tcb_R.thy
```diff
@@ -4067,7 +4079,7 @@ lemma tcb_real_cte_16:
   "\<lbrakk> real_cte_at' (t+2^cteSizeBits) s; tcb_at' t s \<rbrakk> \<Longrightarrow> False"
   by (clarsimp simp: obj_at'_def projectKOs objBitsKO_def ps_clear_16)
 
-lemma corres_splitEE':
+lemma corres_splitEE_pair:
```
mktnk3, line 4082: Move to Corres_UL?

proof/refine/ARM/Tcb_R.thy
```diff
@@ -1154,6 +1154,18 @@ lemma gts_isRunnable_corres:
    apply auto
   done
 
+lemma gts_isBlocked_corres:
```
mktnk3, line 1157: Do we want to move this up somewhere? Maybe?

mktnk3 (approved):
I am glad all the spec issues were sorted in a good way. It’s nice to see this file growing and getting completed.

spec/abstract/Decode_A.thy
```diff
@@ -532,30 +531,40 @@ where
...
+  "obj_ref \<Rightarrow> obj_ref option \<Rightarrow> (sched_context_invocation, 'z::state_ext) se_monad"
+where
+  "decode_sched_context_yield_to sc_ptr buffer \<equiv> doE
       sc \<leftarrow> liftE $ get_sched_context sc_ptr;
```
mktnk3, line 541: Seeing that `sc` always appears with `sc_tcb` in this function, we can use `get_sc_obj_ref sc_tcb` here also.

michaelmcinerney: Would you like me to update this?

mktnk3: That would be nice, I think. I hope it’s not too much more work.

michaelmcinerney: Well, it's in the abstract spec, so I'll have to check all layers and then test all layers again.

I did not modify this line originally. Are we wanting to do this for just this function, or all the functions in the abstract spec?

mktnk3: I thought this is the function you added newly. Maybe you copied this bit from somewhere else? What I am talking about is what I can see inside this function. Not sure about others, but I don’t think it applies to others.

michaelmcinerney: Yeah this was all in one massive `decode_sched_context_invocation` function. 

OK

mktnk3: I thought that, if you are going to move lemmas to Lib, you are likely to need to rebuild anyway.

michaelmcinerney: The lemma you mentioned moving to `Corres_UL` was preexisting. I don't want to move that

mktnk3: Ah I see. Well, in that case, we should probably not rename that one but rename the other one.

michaelmcinerney: The preexisting one had the stupid name. It had maybe only one or maybe two uses in Refine. The lemma I added is named in analogy with `corres_split` and `corres_splits'`

spec/abstract/Decode_A.thy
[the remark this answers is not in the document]
mbrcknl: @corlewis summoned me here, but it looks like this is already changed back to `get_sc_obj_ref`. :-)

@michaelmcinerney, can you clarify what was preferable about `get_sc_corres`? Is it just that it existed, but `get_sc_ob_ref_corres` did not?

In the long run, I think the optimum would be for every access to be just `get` with a lens (and preferably the most focussed lens possible). Then you only need `corres_get`. For now, we at least have that `get_sc_obj_ref` is better than (hypothetical) `get_sc_tcb`, since the former needs only `get_sc_obj_ref_corres` whereas the latter would need `get_sc_tcb_corres`, `get_sc_yield_to_corres`, etc. And along the lines of @corlewis' comment, `get_sc_obj_ref sc_tcb` is usually preferable to `get_sched_context`, because the latter sometimes means you end up explaining things about fields you didn't actually look at.

And @mktnk3 is also right that in proof contexts where `get_sched_context` does not force you to explain things about other fields, you can just unfold, leaving the spec in the more specific form.

spec/abstract/Decode_A.thy
mktnk3: As we have seen repeatedly recently, we don’t do that kind of cleanup in MCS. We do it on master. You are welcome to do it on master. You can add your lemmas, too.

spec/abstract/Decode_A.thy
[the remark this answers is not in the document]
michaelmcinerney: It was just that at the time, I knew off the top of my head that there was `get_sc_corres`, but I couldn't remember a similar one for getting a particular field in the sc. And seeing that it was used elsewhere and lined up nicely with the Haskell, I just went for it. I've gone with unfolding `get_sc_obj_ref` and using another rule to rewrite it so that I can then use `get_sc_corres`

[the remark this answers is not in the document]
mbrcknl: Right, makes sense.

proof/refine/ARM/Tcb_R.thy
```diff
@@ -4079,7 +4079,7 @@ lemma tcb_real_cte_16:
   "\<lbrakk> real_cte_at' (t+2^cteSizeBits) s; tcb_at' t s \<rbrakk> \<Longrightarrow> False"
   by (clarsimp simp: obj_at'_def projectKOs objBitsKO_def ps_clear_16)
 
-lemma corres_splitEE_pair:
+lemma corres_splitEE':
```
mktnk3:
```suggestion
(* FIXME: rename this to corres_splitEE’_pair on master, then rename cores_splitEE’’ in Corres_UL to corres_splitEE’ *)
lemma corres_splitEE':
```

mktnk3 (approved):
Thanks a lot for the changes, Michael. Much appreciated.

proof/refine/ARM/Tcb_R.thy
mktnk3: Just recording this as comment.

PR #210, opened 2021-01-21
receiveSignal_corres

michaelmcinerney:


proof/refine/ARM/Ipc_R.thy
```diff
@@ -3617,61 +3617,311 @@ lemma receive_ipc_corres:
...
+             (do_nbrecv_failed_transfer thread) (doNBRecvFailedTransfer thread)"
+  apply (rule corres_cross[where Q' = "tcb_at' thread", OF tcb_at'_cross_rel], simp)
+  apply (clarsimp simp: do_nbrecv_failed_transfer_def doNBRecvFailedTransfer_def)
+  apply (rule corres_guard_imp)
+    apply (clarsimp simp: badge_register_def badgeRegister_def)
+    apply (rule user_setreg_corres)
+   apply clarsimp+
+  done
+
+lemma tcb_ep_find_index_corres:
```
michaelmcinerney, line 3734: The lemmas about `tcbEPFindIndex` and `tcbEPAppend` just moved from Tcb_R.thy

MitchellBuckley (approved):
Amazing! If possible, I think we should start moving these extra little lemmas much higher in the dependency tree because it's becoming very common to move lemmas and it might soon become quite a nuisance.

michaelmcinerney:
> Amazing! If possible, I think we should start moving these extra little lemmas much higher in the dependency tree because it's becoming very common to move lemmas and it might soon become quite a nuisance.

The `asUser` and `as_user` stuff might be useful elsewhere. Happy to move those if you'd like. The other functions like `maybe_return_sc`, `schedule_tcb`, and `do_nbrecv_failed_transfer` are really only for these IPC related functions, so they should be in this file. Not sure exactly where. Maybe once your `sendSignal_corres` PR is merged I can take another looky

michaelmcinerney:
After some discussion with Curtis, we decided to move the call to `maybeReturnSc` within `receiveSignal` to the end of the block in which it occurs. The PR for SEL4 is here

https://bitbucket.ts.data61.csiro.au/projects/SEL4/repos/sel4/pull-requests/1878/overview

I've updated the abstract spec, fixed the proofs in AInvs, updated the Haskell (again) and fixed the proof in Refine

corlewis (commented):
The proofs look great, but I'm a bit confused by the removal of `schedule_tcb`. Have I missed a different seL4 PR? The one you linked doesn't make the same change.

michaelmcinerney:
> The proofs look great, but I'm a bit confused by the removal of `schedule_tcb`. Have I missed a different seL4 PR? The one you linked doesn't make the same change.

It is a bit confusing. So, here

https://github.com/seL4/l4v/blob/master/proof/crefine/ARM/Ipc_C.thy#L5954

we have a lemma which does all the heavy lifting. The `schedule_tcb` isn't necessary in the abstract - it's incorporated into `set_thread_state`. I'm not sure whether @lsf37 checked this at the time, but I think he agreed that this was the correct approach

mktnk3:
> > The proofs look great, but I'm a bit confused by the removal of `schedule_tcb`. Have I missed a different seL4 PR? The one you linked doesn't make the same change.
> 
> It is a bit confusing. So, here
> 
> https://github.com/seL4/l4v/blob/master/proof/crefine/ARM/Ipc_C.thy#L5954
> 
> we have a lemma which does all the heavy lifting. The `schedule_tcb` isn't necessary in the abstract - it's incorporated into `set_thread_state`. I'm not sure whether @lsf37 checked this at the time, but I think he agreed that this was the correct approach

I haven’t been following the C changes closely yet, but I can say that it is also my understanding that `schedule_tcb` is unfolded and simplified in the abstract `set_thread_state`, and that’s how things have (always? or long enough) been, I believe.

corlewis:
> 
> 
> > > The proofs look great, but I'm a bit confused by the removal of `schedule_tcb`. Have I missed a different seL4 PR? The one you linked doesn't make the same change.
> > 
> > 
> > It is a bit confusing. So, here
> > https://github.com/seL4/l4v/blob/master/proof/crefine/ARM/Ipc_C.thy#L5954
> > we have a lemma which does all the heavy lifting. The `schedule_tcb` isn't necessary in the abstract - it's incorporated into `set_thread_state`. I'm not sure whether @lsf37 checked this at the time, but I think he agreed that this was the correct approach
> 
> I haven’t been following the C changes closely yet, but I can say that it is also my understanding that `schedule_tcb` is unfolded and simplified in the abstract `set_thread_state`, and that’s how things have (always? or long enough) been, I believe.

Right, that all makes sense. Thanks for the explanation!

corlewis (approved):

lsf37:
> The schedule_tcb isn't necessary in the abstract - it's incorporated into set_thread_state. I'm not sure whether @lsf37 checked this at the time, but I think he agreed that this was the correct approach.

Just confirming: I did, and agree.

mktnk3 (approved):
Nice! Looks good to go, after rebasing.

PR #211, opened 2021-01-22
Prove updateReply_replyNext_Nothing_invs'

ryybrr:


mktnk3 (approved):
Nice! For `valid_objs’_reply_update`, there is the `ep` version of the whole lot in KHeap_R, so we might as well move it and what it depends on there. I am not sure if it is possible to apply some abstraction and put the `ep` ones and `reply` ones together. Maybe not. Also, there is no need to put everything in the arch context, I think. Only the `valid_arch_obj’` one and maybe the `valid_cap’, maybe`?

corlewis (approved):
Look good, thanks for proving this!

ryybrr:
> For `valid_objs’_reply_update`, there is the `ep` version of the whole lot in KHeap_R, so we might as well move it and what it depends on there.
I did copy those lemmas, so I agree.

PR #214, opened 2021-01-28
Prove send_fault_ipc_corres, hf_corres and handleTimeout_corres

ryybrr:


mktnk3 (approved):
Nice and neat. Much shorter proofs than the original ones, too.

proof/invariant-abstract/DetSchedInvs_AI.thy
```diff
@@ -3054,7 +3054,7 @@ lemma active_reply_scsE:
   using assms by (simp add: active_reply_scs_2_def active_if_reply_sc_at_2_def)
 
 definition valid_sched_2 where
-  "valid_sched_2 wk_vsa vbl ctime cdom ct it queues rlq sa etcbs tcb_sts tcb_scps tcb_faults sc_refill_cfgs sc_reps \<equiv>
+  "valid_sched_2 wk_vsa vbl riq ctime cdom ct it queues rlq sa etcbs tcb_sts tcb_scps tcb_faults sc_refill_cfgs sc_reps \<equiv>
```
MitchellBuckley, line 3057: This is a neat solution and quite effective, but I'm slightly worried about the number of these boolean flags getting out of hand. It will be good to confirm with the rest of the team before merging this change, just to be safe.

MitchellBuckley (approved):
Very neat! I like it! It could even be merged ahead of mine if that doesn't disrupt anyone. Then I could fix any breakages when rebasing.

proof/invariant-abstract/DetSchedInvs_AI.thy
MitchellBuckley: @mbrcknl in particular who might not be seeing all these PRs.

mbrcknl: I'm fine with this for now. This is another bundling problem where it would be nice to have a systematic way to patch or (de)select components of bundles. But that will take quite a lot of thought and disruption, so this is probably the best we can do right now.

mbrcknl (approved):

mktnk3:
Thanks for adding `handleTimeout_corres`, Ryan. Looks great to me, ready to go (after rebase).

proof/invariant-abstract/SchedContextInv_AI.thy
```diff
@@ -1566,10 +1566,12 @@ lemma handle_timeout_Timeout_invs:
    apply (clarsimp simp: pred_tcb_at_def obj_at_def dest!: get_tcb_SomeD)
    apply (drule invs_iflive)
    apply (drule (1) if_live_then_nonz_capD2)
-    apply (fastforce simp: live_def split: )
+    apply (fastforce simp: live_def)
    apply clarsimp
-  apply (rule_tac x=4 in exI)
-  apply (auto simp: tcb_cnode_map_def caps_of_state_tcb_index_trans)
+  apply (rule_tac P="\<exists>n. P n" and P'="P 4" and Q=Q and Q'=Q for P Q in conj_forward)
```
corlewis, line 1571: Not a blocking issue at all, but out of curiosity do you remember where the `4` comes from in this proof?

proof/refine/ARM/Ipc_R.thy
```diff
@@ -4751,56 +4586,42 @@ lemma tg_sp':
 
 declare lookup_cap_valid' [wp]
 
+lemma thread_set_fault_valid_sched_except_blocked_except_released_ipc_qs[wp]:
```
corlewis, line 4589: FIXME: move?

corlewis (approved):
Very nice!

proof/invariant-abstract/SchedContextInv_AI.thy
ryybrr: Refers to the timeout handler, which is in the 4th slot

PR #217, opened 2021-02-02
four small sorries in Refine_R.thy

MitchellBuckley:
Some of the small added lemmas should probably be moved. Either to AInvs, or to move.thy.

### DIFF
--- a/proof/refine/ARM/Refine.thy
+++ b/proof/refine/ARM/Refine.thy
@@ -86,7 +86,7 @@ lemma typ_at_UserDataI:
               split: Structures_A.kernel_object.split_asm
                      Structures_H.kernel_object.split_asm
                      if_split_asm arch_kernel_obj.split_asm)
-  sorry (* schedContext
+  apply (case_tac ko; simp)
   apply (rename_tac vmpage_size n)
   apply (rule_tac x = vmpage_size in exI)
   apply (subst conjunct2 [OF is_aligned_add_helper])
@@ -95,7 +95,7 @@ lemma typ_at_UserDataI:
    apply (erule word_less_power_trans2 [OF _ pbfs_atleast_pageBits])
    apply (case_tac vmpage_size, simp_all add: word_bits_conv)[1]
   apply (simp add: obj_at_def  a_type_def)
-  done *)
+  done
 
 lemma typ_at_DeviceDataI:
   "\<lbrakk> typ_at' UserDataDeviceT (p && ~~ mask pageBits) s';
@@ -118,7 +118,7 @@ lemma typ_at_DeviceDataI:
               split: Structures_A.kernel_object.split_asm
                      Structures_H.kernel_object.split_asm
                      if_split_asm arch_kernel_obj.split_asm)
-  sorry (* schedContext
+  apply (case_tac ko; simp)
   apply (rename_tac vmpage_size n)
   apply (rule_tac x = vmpage_size in exI)
   apply (subst conjunct2 [OF is_aligned_add_helper])
@@ -127,7 +127,7 @@ lemma typ_at_DeviceDataI:
    apply (erule word_less_power_trans2 [OF _ pbfs_atleast_pageBits])
    apply (case_tac vmpage_size, simp_all add: word_bits_conv)[1]
   apply (simp add: obj_at_def  a_type_def)
-  done *)
+  done
 
 lemma pointerInUserData_relation:
   "\<lbrakk> (s,s') \<in> state_relation; valid_state' s'; valid_state s\<rbrakk>
@@ -227,21 +227,6 @@ lemma absKState_correct:
   done
 
 text \<open>The top-level invariance\<close>
-
-lemma activate_thread_sched_act:
-  "\<lbrace>ct_in_state activatable and
-    (\<lambda>s. is_schedulable_bool (cur_thread s) s \<and> P (scheduler_action s))\<rbrace>
-   activate_thread
-   \<lbrace>\<lambda>_ s. P (scheduler_action s)\<rbrace>"
-  unfolding activate_thread_def
-  sorry (*
-  by (simp add: activate_thread_def set_thread_state_def arch_activate_idle_thread_def
-      | (wp set_thread_state_sched_act gts_wp)+ | wpc)+ *)
-
-lemma schedule_sched_act_rct[wp]:
-  "\<lbrace>\<top>\<rbrace> Schedule_A.schedule \<lbrace>\<lambda>_ s. scheduler_action s = resume_cur_thread\<rbrace>"
-  unfolding Schedule_A.schedule_def
-  sorry
 
 lemma call_kernel_sched_act_rct[wp]:
   "\<lbrace> einvs and (\<lambda>s. e \<noteq> Interrupt \<longrightarrow> ct_running s) and
@@ -590,7 +575,7 @@ lemma kernel_corres:
        apply (simp add: kernelExitAssertions_def state_relation_def)
       apply (simp only: bind_assoc)
       apply (rule kernel_corres')
-  sorry (* FIXME RT: call_kernel_domain_time_inv_det_ext
+  sorry (* FIXME RT: call_kernel_domain_time_inv_det_ext in DetSchedDomainTime_AI
      apply (wp call_kernel_domain_time_inv_det_ext call_kernel_domain_list_inv_det_ext)
     apply wp
    apply clarsimp
--- a/proof/refine/Move_R.thy
+++ b/proof/refine/Move_R.thy
@@ -386,4 +386,48 @@ lemma thread_set_valid_tcb[wp]:
   "thread_set f tptr \<lbrace>valid_tcb ptr tcb\<rbrace>"
   by (wpsimp wp: get_object_wp simp: thread_set_def)
 
+(* BEGIN: scheduler_action lemmas needed in Refine.thy *)
+
+(* FIXME RT: improve the existing sts_schedulable_scheduler_action *)
+lemma sts_schedulable_scheduler_action2:
+  "\<lbrace>\<lambda>s. P (scheduler_action s) \<and> is_schedulable_bool thread s \<and> runnable st\<rbrace>
+   set_thread_state thread st
+  \<lbrace>\<lambda>_ s. P (scheduler_action s)\<rbrace>"
+  apply (wpsimp wp: stsa_schedulable_scheduler_action set_object_wp simp: set_thread_state_def)
+  apply (clarsimp simp: is_schedulable_bool_def is_sc_active_def get_tcb_def
+                        in_release_queue_def
+                 split: option.splits kernel_object.splits)
+  apply (rename_tac ko ko' ko'')
+  apply (case_tac ko; clarsimp)
+  by fastforce
+
+lemma activate_thread_sched_act:
+  "\<lbrace>(\<lambda>s. P (scheduler_action s)) and
+    ct_in_state activatable and
+    (\<lambda>s. is_schedulable_bool (cur_thread s) s) \<rbrace>
+   activate_thread
+   \<lbrace>\<lambda>_. \<lambda>s. P (scheduler_action s)\<rbrace>"
+  unfolding activate_thread_def
+  by (wpsimp wp: sts_schedulable_scheduler_action2 gts_wp hoare_drop_imp hoare_vcg_all_lift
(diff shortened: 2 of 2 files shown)


proof/refine/ARM/Refine.thy
```diff
@@ -229,19 +229,29 @@ lemma absKState_correct:
...
-  by (simp add: activate_thread_def set_thread_state_def arch_activate_idle_thread_def
-      | (wp set_thread_state_sched_act gts_wp)+ | wpc)+ *)
+  apply (wpsimp wp: set_thread_state_schact_is_rct_strong gts_wp hoare_drop_imp hoare_vcg_all_lift)
+  by (clarsimp simp: is_schedulable_bool_def2)
+
+crunches sc_and_timer
+  for scheduler_action[wp]: "\<lambda>s. P (scheduler_action s)"
+  (wp: crunch_wps)
+
+lemmas ssa_schact_is_rct_obvious[wp] = set_scheduler_action_obvious [
```
michaelmcinerney, line 243: Nitpick: did you mean to have the square bracket at the end of the line?

proof/refine/ARM/Refine.thy
```diff
@@ -229,19 +229,29 @@ lemma absKState_correct:
...
+lemma schact_is_rct_trivial[simp]:
+  "schact_is_rct_2 resume_cur_thread"
+  by (simp add: schact_is_rct_def)
 
 lemma schedule_sched_act_rct[wp]:
-  "\<lbrace>\<top>\<rbrace> Schedule_A.schedule \<lbrace>\<lambda>_ s. scheduler_action s = resume_cur_thread\<rbrace>"
+  "\<lbrace>\<top>\<rbrace> Schedule_A.schedule \<lbrace>\<lambda>_. schact_is_rct\<rbrace>"
   unfolding Schedule_A.schedule_def
-  sorry
+  by wpsimp
```
michaelmcinerney, line 254: Another nitpicky thing: what are the reasons for you preferring using `by` as the final line in a proof? I like it when it's the final line in a proof (or the final bit in a `subgoal` block) that takes a while to process, and when it's for proofs that are almost surely never going to break. But I've come across an absolute tonne of proofs that have broken recently, and it's a bit annoying to not be able to move the cursor to the final line and see what the proof state is. Minor, I know, but if we are going to write a style guide soon, it might be good to hash this out now

michaelmcinerney (approved):
That sorry count stands no chance

VictorPhan (approved):

proof/refine/ARM/Refine.thy
MitchellBuckley: I did, but I can change it if that looks wrong.

proof/refine/ARM/Refine.thy
MitchellBuckley: I do it for purely aesthetic reasons, so I'm happy to use apply-done if that's better. Has this been discussed before @Xaphiosis @mbrcknl @lsf37 ?

Xaphiosis: My personal opinion is with @michaelmcinerney on this one and his reasons seem sound. I'd use `apply/done` if I can't get rid of it in one go (well, `by` can take two commands, so two goes).

Apart from the obvious reason of "saves a vertical line", there is another reason to consider: interactive performance. People like Dan used `by` on the last line of the proof as surprisngly often the last statement of a lemma is a "hammer", unfolding a bunch of stuff and then throwing everything at it. Those "hammer" steps are slow, and `by` is able to fork off the work to another process so the interactivity keeps going. `apply` does not do this.
Dan is also the guy who developed "defer proofs", so the effect of that last-line by is largely mitigated, but it might still matter to people (how much though?).

mbrcknl: I don't recall any discussions about this. I know I've had a tendency to favour `by`, for no particularly principled reason. But like @michaelmcinerney, I've gradually come to the view that this is sometimes an anti-pattern, for the same reason @michaelmcinerney describes. All else being equal, we should do whatever makes life even a tiny bit easier for future maintainers.

And so for a method like `wpsimp`, which can make progress and then effectively stop waiting for input regarding which rule to apply next, I think `apply` is a good choice. If a broken `wpsimp` just needs one extra rule, then this saves you having to change `by` to `apply` to see what's going on, and then back to `by` again once you've fixed it. Instead, you just point and then add the new rule.

For methods like `fastforce`, I think `by` is fine, since it allows an expensive proof to fork into the background, and a broken `fastforce` won't tell you anything regardless.

`auto` is a bit more interesting. In this case, I think all else is not equal. For `auto`, we want to know that we only use it to completely solve all subgoals exposed to it. And so I think we still always want `by auto` to make this clear to reviewers.

mbrcknl (approved):

proof/refine/ARM/Refine.thy
michaelmcinerney: Yeah `by` is a good option when we want to fork the proof off. That's what I meant when I said I like it for proofs that take a while to process.

`auto` is an interesting one. I think if we're doing `auto` part way through a proof, `subgoal by auto` would be best (or at least if we're going to use `auto` part way through a proof, using `by` is a good option). I suppose if the `auto` is the final line of a proof, then it's pretty clear that it's meant to solve everything, so we're back in the original case where it's an advantage only if it takes a while to process

mbrcknl: That's a fair point. While I'm sure we don't want to relax the rule that every use of `auto` must completely solve any subgoals exposed to it, we can perhaps relax the rule about how this is expressed.
My motivations for wanting a hard rule that `auto` must always be used with `by` (whether or not in a `subgoal` focus) were:
- Make it easy for reviewers (and perhaps eventually CI checks) to spot transgressions of the rule that `auto` must always solve all subgoals exposed to it.
- Make it hard for maintainers to accidentally break that rule by adding new steps to solve unsolved subgoals without noticing that those subgoals came out of an `auto`.
If there's consensus that `apply (auto ...)\n done` achieves those same goals, then I'd be happy to go with that.
I'm still dead against `apply (auto ...)[1]`, though. :-)

michaelmcinerney: No that sounds good, especially the idea of making a CI check (which should be straightforward, and hopefully can be done soonish!)

lsf37: I agree with the discussion so far. I do personally slightly prefer `by` if it is a single-line proof (not the end of a larger proof) usually just because it is shorter and looks neater, and also because it still gives you progress when there is non-termination when you are on a sorry- or proof-fixing run, esp if you have a few cores. You still have to go back and fix it, but it doesn't force you to do this specific one immediately.

In the current Isabelle update, I'm usually grateful for the proofs that use `by` for anything longer than about a half second, because I'm on a fast machine with sufficient cores, and there is a large difference between being stuck on a single "takes long" `fastforce` at the end somewhere with `apply` vs one that ends in `by`. Even one that takes about a second can block 3-4 smaller proofs down the line and be the difference between seeing if there is a lot more breakage or none.

You often still get to see the failing final proof state when you put the cursor on the `by` (it's usually in the error message, although maybe not with all tactics). It is true though, that then, to fix it, you usually do have to unfold it into apply/done. It'd be nice if that could be automated with a jEdit macro, so that it becomes one key stroke. Then we could have both.

mbrcknl: Ha ha, while writing my comments, I did have a vague sense that I got to see the unfinished goal state even with a `by`, but @michaelmcinerney seemed quite sure that he didn't, so I assumed I was misremembering. :-)

lsf37: I think it depends on the error message the tactic produces. I do remember being annoyed at having to unfold to see something, but I have also been using the error output quite a bit in the current update.

michaelmcinerney: I probably just instictinvely do not look in the output pane for the proof state. Hmm. I guess you can put me down as "generally hates `by` in most cases for no good reason"

mbrcknl: > I think it depends on the error message the tactic produces.

Right, that makes a lot of sense.

MitchellBuckley: Ok, I'm gonna leave my particular case unchanged. But good discussion!

proof/refine/ARM/Refine.thy
```diff
@@ -268,7 +278,7 @@ lemma kernel_entry_invs:
             in hoare_post_imp)
    apply clarsimp
   apply (simp add: kernel_entry_def)
-  sorry (* FIXME RT: some of these are commented out and still to prove in AInvs: (call_kernel_domain_time)
+  sorry (* FIXME RT: some of these are commented out andOk,I'm still to prove in AInvs: (call_kernel_domain_time)
```
corlewis, line 281: Is this an accidental change?

corlewis (approved):

proof/refine/ARM/Refine.thy
MitchellBuckley: Yup! Will revert.

proof/refine/ARM/Refine.thy
```diff
@@ -229,19 +229,29 @@ lemma absKState_correct:
 text \<open>The top-level invariance\<close>
 
 lemma activate_thread_sched_act:
-  "\<lbrace>ct_in_state activatable and
-    (\<lambda>s. is_schedulable_bool (cur_thread s) s \<and> P (scheduler_action s))\<rbrace>
+  "\<lbrace>ct_in_state activatable and (\<lambda>s. is_schedulable_bool (cur_thread s) s) and schact_is_rct\<rbrace>
    activate_thread
-   \<lbrace>\<lambda>_ s. P (scheduler_action s)\<rbrace>"
+   \<lbrace>\<lambda>_. schact_is_rct\<rbrace>"
```
mktnk3, line 234: Just wondering, is this change fully justified?

mktnk3: Never mind. It’s better if you can show this, of course.

mktnk3 (approved):
Nice!

proof/refine/ARM/Refine.thy
michaelmcinerney: Oooooh I've just found another reason to support my hatred of `by` in most cases: a broken proof ending in `by` will actually still produce the fact. Maybe that's the wrong terminology, but you will still get something when you go `thm broken_lemma`. I've been bitten by this quite a few times (it's especially bad when a `crunches` fails and still produces facts, but apparently that behaviour is quite hard to get rid of). Again, not a big deal, but something to think about

MitchellBuckley:
I've updated this a little with:
* moved these proofs up to `Move.thy` because they can be moved there.
* Strengthened that single lemma that Miki commented on.
I hope those are safe/small changes.

proof/refine/ARM/Refine.thy
corlewis: That specific behaviour is again a divisive one where I think it really depends on people's workflows. Personally, this is one which I've found useful more often than it's caused issues, but I can also easily see it going in the opposite direction for other people.

I'm the same as Gerwin in that I find it useful to still be able to make progress even with an earlier broken (or non-terminating) proof. This is particularly the case when the broken lemmas are a bit unrelated to what I'm currently focused on  and I don't want to split my attention.

Separately, from a technical perspective my guess is that getting rid of this behaviour is not a fight that you're going to win, even if there was consensus to do it. If you want to still be able to fork off proofs for performance reasons then any alternative behaviour that I've been able think of would be much worse.

michaelmcinerney: I’ve always been in support of using by when it comes to proofs that take a long time to check. And I’m not expecting to convince anyone, necessarily. I thought I’d just make a note of it

corlewis: Yeah, I was trying to say that I don't think it's technically feasible to use by to improve performance like that and avoid having broken proofs still produce the fact.

michaelmcinerney: Ah, I getcha.

Maybe there's a way to make us both happy: it would be cool if there were a warning of some kind when you've completed a proof that relies on something that's failed or has been sorried. I'm not sure whether that's technically feasible. I know that Ed whipped up a few lines of ML code that would say whether a theorem used any sort of oracle (I forget the details, though I could probably find the code if anyone's interested) but I think it would really only be useful if it were baked into Isabelle and jEdit.

PR #218, opened 2021-02-03
Modify corres_split rule

corlewis:
This change reorders the `corres_split` assumptions so that the first resulting subgoal concerns the first functions of the composite monad. That is, if we are proving `corres ... (a >>= b) (c >>= d)` then we should prove `corres ... a c` before `corres ... b d`. I find this much more natural than the reverse order and so far it has made my proofs slightly simpler.

At this stage I have kept the old version of the lemma as a deprecated rule to avoid having to update all of the proofs. From some quick experiments I don't expect any of the proofs to be hard to update, but there's a lot of them and I think it would be difficult to automate.

I haven't reordered any of the other corres split rules that are derived from `corres_split`. I am undecided whether I should reorder them now and make deprecated versions of them as well or leave it until we update the proofs.

Testing is ongoing but it has passed for ARM and ARM_HYP and I don't expect issues from the other architectures.

For reviewing purposes, the only interesting changes are in `lib/Corres_UL.thy`. Everything else is the result of a `sed` command.


corlewis:
Ignore the Theory Linter failure, it found an existing use of `thm`  in documentation of a test file.

lsf37:
I'm happy with the reorder as such, but I'm not sure I like the fact that we now have `corres_split_old` pretty much everywhere. Do you have concrete plans of updating things after this fairly soon, or are we going to be in this state for the next few years? Asking from my own laziness ;-)

I'm hoping to get `Refine` next on the list for the Isabelle update (currently in `DRefine` with surprisingly much breakage after `AInvs` went smoother than I had feared). After that I might be able to give an impression how much this will conflict.

lib/Corres_UL.thy
```diff
@@ -246,10 +246,10 @@ lemma corres_split':
 text \<open>Derivative splitting rules\<close>
 
 lemma corres_split:
-  assumes y: "\<And>rv rv'. r' rv rv' \<Longrightarrow> corres_underlying sr nf nf' r (R rv) (R' rv') (b rv) (d rv')"
   assumes x: "corres_underlying sr nf nf' r' P P' a c"
+  assumes y: "\<And>rv rv'. r' rv rv' \<Longrightarrow> corres_underlying sr nf nf' r (R rv) (R' rv') (b rv) (d rv')"
   assumes    "\<lbrace>Q\<rbrace> a \<lbrace>R\<rbrace>" "\<lbrace>Q'\<rbrace> c \<lbrace>R'\<rbrace>"
-  shows      "corres_underlying sr nf nf' r (P and Q) (P' and Q') (a >>= (\<lambda>rv. b rv)) (c >>= (\<lambda>rv'. d rv'))"
+  shows      "corres_underlying sr nf nf' r (P and Q) (P' and Q') (a >>= b) (c >>= d)"
```
mbrcknl, line 252: These `%rv. b rv` patterns are often intentional. Even though the internal term representation primarily uses de Bruijn indices, it remembers bound variable names to use as hints to use in printing. These often end up being visible in the goal. (Every variable name that shows as green is one of these bound variables, and its name is derived from this hint.)
When the higher-order unifier has to generate a lambda abstraction, I guess it sometimes generates such a hint from one of the patterns it's matching. I don't know enough details to know how that works, or whether that would happen here, or under what circumstances that `rv` might end up being visible in a goal. But I think the intention is that if it does end up being visible, it shows as `rv` rather than `x7`. I think @Xaphiosis is quite keen on this, so perhaps he knows more.

mbrcknl:
> `corres_split_old` pretty much everywhere.

I suspect a further search and replace from `corres_split_old *\[ *OF +_ +` to `corres_split[OF ` would be safe, and would remove quite a lot of cases. (It would be unsafe in cases with two facts after the `_`, but I didn't notice any of those.) If you can figure out how to do a multi-line replacement, you might also be able to handle the `prefer 2` cases, though there don't seem to be very many of those.

As for naming, perhaps `corres_split_deprecated`?

lib/Corres_UL.thy
Xaphiosis: I missed this. Yes, this should definitely be left in. I thought it's also so that if you have:

```
do
  tptr <- f x;
  tcb <- g tptr;
  h
od
```

and you split these with split rules, then you get `tptr` and `tcb` as your variables, instead of what's  used *inside* the split rule itself or some other rando as @mbrcknl mentioned, or usually `rv`, `rva`, 'rvaa` etc. It's lame to do a split, immediately followed by rename_tac when the names we want are already in the program. You'll see a bunch of this in fastpath proofs. @lsf37 will have the Isabelle internals angle here, I only have the pragmatic attitude. Still vote to keep this schema alive.

lsf37: Yes, also has my vote. I have forgotten the internals, but exactly that was the intention behind the eta expansion here.

corlewis: Oh, I didn't know that this sort of expansion helped with names like that. I definitely agree that we want to keep it in then, especially since I don't even remember removing this in the first place.

corlewis:
> 
> 
> > `corres_split_old` pretty much everywhere.
> 
> I suspect a further search and replace from `corres_split_old *\[ *OF +_ +` to `corres_split[OF ` would be safe, and would remove quite a lot of cases. (It would be unsafe in cases with two facts after the `_`, but I didn't notice any of those.) If you can figure out how to do a multi-line replacement, you might also be able to handle the `prefer 2` cases, though there don't seem to be very many of those.
> 
> As for naming, perhaps `corres_split_deprecated`?

I agree that that would resolve a lot of the cases, but it might cause a few breakages where the following steps need to be reordered. The complication comes from when the additional rule being supplied has assumptions of its own that create new subgoals.

In general I don't think updating any of the proofs would be hard, it's just a lot of work and I wanted this merged soon since I used this version of the split rule in my mcs work. It's also nice to get it sooner rather than later so that people can start using it and changing over.

What sort of conflicts are you worried about @lsf37? In my mind this was going to be relatively easy to merge over, any new work would hopefully use `corres_split` and anything untouched would stay as `corres_split_deprecated` (for now).

mbrcknl:
> The complication comes from when the additional rule being supplied has assumptions of its own that create new subgoals.

You're right. I did not think of that.

mbrcknl:
If @lsf37 can be convinced that merge conflicts will be manageable, then I'm in favour of this.

BTW, I thought of a way to do this incrementally: Use context blocks to bring back the bad `corres_split` with its old name:
```
context
  notes corres_split = corres_split_deprecated
begin
...
end
```
However, this is terrible, so I suggest we don't do this. It's terrible because it's the worst of both worlds: you have a bunch of work to do to find all the places you need to make these context blocks, but you don't get the benefit of forcing people to start using the better version.

lsf37:
> What sort of conflicts are you worried about @lsf37? In my mind this was going to be relatively easy to merge over, any new work would hopefully use corres_split and anything untouched would stay as corres_split_deprecated (for now).

Just conflicts with the Isabelle update that might touch the same lines of code for no particular reason. I think you're right, it's unlikely and we should just try.

> BTW, I thought of a way to do this incrementally: Wrap things in context blocks to bring back the bad corres_split with its old name:

I agree, let's not do this one. Nice hack to remember for other situations, though!

corlewis:
@lsf37 Are you happy for this to go in now? Would it help to have it merged before you get up to `Refine` with the Isabelle update?

lsf37:
Yes, I'm happy with it and it would help to have it merged soon -- I'm about to run out of other sessions before Refine ;-)

corlewis:
Great, any approvals from anyone then? The eta expansion change was reverted and the name updated to `corres_split_deprecated`, which I think were the concrete suggestions from previous reviews.

The `DRefine` failure is just due to running out of memory, it works locally for me.

mbrcknl (approved):

mktnk3 (approved):
Nice. Thanks all for the interesting discussion.
I think we could do with a JIRA ticket on the cleanup, with a link to this PR.

lsf37 (approved):
Oops, sorry, I thought I already had approved :-)

corlewis:
> I think we could do with a JIRA ticket on the cleanup, with a link to this PR.

Good point, I've made https://sel4.atlassian.net/browse/VER-1390 for this.

PR #222, opened 2021-02-09
Reader Monad update

mktnk3:
This PR contains the reader monad changes separated out from the awaken PR, plus the proof fixes for Refine.

Rough descriptions of the changes are as follows: 

Haskell:
- adding `KernelR`, using ReaderM (from Control.Monad.Reader), which takes a KernelState and returns a Maybe value
- new definition of `getObject` via `readObject`
- some getter functions are redefined via `readObject` (due to the awaken PR) but not all

Haskell translator:
- updates for handling the reader monad

Design:
- new definitions of alignment related functions, `magnitudeCheck`, etc., via the reader versions of them
- `loadObject` is now in `kernel_r`, while `updateObject` is still in `kernel`. So, alignment checks in `loadObject_default` now uses the reader version of the functions.

Refine:
- proof fixes for the `getObject` change and others
- definitions of `no_fail_r` and `ovalid`

--------

A few comments:
- there are things that need to be moved to better places. For instance,`no_fail_r`. 
- `ovalid` is currently not very useful, but concept-wise it is quite clear and it is rather hard to read without at least an abbreviation.
- most of the Haskell updates are by Gerwin. Also the initial lib and haskell-translator setup.

### DIFF
--- a/lib/HaskellLib_H.thy
+++ b/lib/HaskellLib_H.thy
@@ -34,6 +34,12 @@ definition
  "runState f s \<equiv> THE x. x \<in> fst (f s)"
 
 definition
+  "runReaderT \<equiv> id"
+
+abbreviation (input)
+  "getsJust \<equiv> gets_the"
+
+definition
   sassert :: "bool \<Rightarrow> 'a \<Rightarrow> 'a" where
  "sassert P \<equiv> if P then id else (\<lambda>x. undefined)"
 
--- a/lib/HaskellLib_H.thy
+++ b/lib/HaskellLib_H.thy
@@ -15,13 +15,14 @@ imports
   NatBitwise
   "More_Numeral_Type"
   "Monad_WP/WhileLoopRules"
+  "Monad_WP/OptionMonadWP"
 begin
 
 abbreviation (input) "flip \<equiv> swp"
 
 abbreviation(input) bind_drop :: "('a, 'c) nondet_monad \<Rightarrow> ('a, 'b) nondet_monad
                       \<Rightarrow> ('a, 'b) nondet_monad" (infixl ">>'_" 60)
-  where "bind_drop \<equiv> (\<lambda>x y. bind x (K_bind y))"
+  where "bind_drop \<equiv> (\<lambda>x y. NonDetMonad.bind x (K_bind y))"
 
 lemma bind_drop_test:
   "foldr bind_drop x (return ()) = sequence_x x"
@@ -602,5 +603,14 @@ lemma whileM_inv:
   shows "whileM P f \<lbrace>Q\<rbrace>"
   unfolding whileM_def
   by (wpsimp wp: whileLoop_wp[where I="\<lambda>_. Q"])
+
+definition ohaskell_fail :: "unit list \<Rightarrow> ('s, 'a) lookup" where
+  "ohaskell_fail = K ofail"
+
+definition ohaskell_assert :: "bool \<Rightarrow> unit list \<Rightarrow> ('s, unit) lookup" where
+  "ohaskell_assert P ls \<equiv> if P then oreturn () else ofail"
+
+lemmas omonad_defs = ofail_def oreturn_def oassert_def oassert_opt_def asks_def
+                     ohaskell_assert_def ohaskell_fail_def ounless_def owhen_def
 
 end
--- a/lib/Monad_WP/OptionMonad.thy
+++ b/lib/Monad_WP/OptionMonad.thy
@@ -91,9 +91,22 @@ definition
 definition
   "oassert P \<equiv> if P then oreturn () else ofail"
 
-definition oapply :: "'a \<Rightarrow> ('a \<Rightarrow> 'b option) \<Rightarrow> 'b option"
-  where
+definition
+  "oassert_opt r \<equiv> case r of None \<Rightarrow> ofail | Some x \<Rightarrow> oreturn x"
+
+definition oapply :: "'a \<Rightarrow> ('a \<Rightarrow> 'b option) \<Rightarrow> 'b option" where
   "oapply x \<equiv> \<lambda>s. s x"
+
+definition oliftM :: "('a \<Rightarrow> 'b) \<Rightarrow> ('s,'a) lookup \<Rightarrow> ('s,'b) lookup" where
+  "oliftM f m \<equiv> do { x \<leftarrow> m; oreturn (f x) }"
+
+(* Reader monad interface: *)
+abbreviation (input)
+  "ask \<equiv> Some"
+
+definition
+  "asks f = do { v <- ask; oreturn (f v) }"
+
 
 text \<open>
   If the result can be an exception.
--- a/lib/Monad_WP/OptionMonad.thy
+++ b/lib/Monad_WP/OptionMonad.thy
@@ -113,6 +113,8 @@ abbreviation (input)
 definition
   "asks f = do { v <- ask; oreturn (f v) }"
 
+abbreviation
+  "ogets \<equiv> asks"
 
 text \<open>
   If the result can be an exception.
--- a/lib/Monad_WP/OptionMonad.thy
+++ b/lib/Monad_WP/OptionMonad.thy
@@ -100,6 +100,12 @@ definition oapply :: "'a \<Rightarrow> ('a \<Rightarrow> 'b option) \<Rightarrow
 definition oliftM :: "('a \<Rightarrow> 'b) \<Rightarrow> ('s,'a) lookup \<Rightarrow> ('s,'b) lookup" where
   "oliftM f m \<equiv> do { x \<leftarrow> m; oreturn (f x) }"
 
+definition ounless :: "bool \<Rightarrow> ('s, unit) lookup \<Rightarrow> ('s, unit) lookup" where
+  "ounless P f \<equiv> if P then oreturn () else f"
+
+definition owhen :: "bool \<Rightarrow> ('s, unit) lookup \<Rightarrow> ('s, unit) lookup" where
+  "owhen P f \<equiv> if P then f else oreturn ()"
+
 (* Reader monad interface: *)
 abbreviation (input)
   "ask \<equiv> Some"
--- a/proof/refine/ARM/ArchAcc_R.thy
+++ b/proof/refine/ARM/ArchAcc_R.thy
@@ -79,6 +79,54 @@ end
 
 context begin interpretation Arch . (*FIXME: arch_split*)
 
+method readObject_arch_obj_at'_method
+  =  clarsimp simp: readObject_def obind_def omonad_defs split_def loadObject_default_def obj_at'_def
+                    objBits_simps archObjSize_def projectKOs bit_simps' typ_at_to_obj_at_arches
+             split: option.splits if_split_asm
+
+lemma readObject_misc_arch_ko_at'[simp, dest!]:
+  shows
+  readObject_ko_at'_pte: "readObject p s = Some (pte :: pte) \<Longrightarrow> ko_at' pte p s" and
+  readObject_ko_at'_pde: "readObject p s = Some (pde :: pde) \<Longrightarrow> ko_at' pde p s" and
+  readObject_ko_at'_asidpool: "readObject p s = Some (asidp :: asidpool) \<Longrightarrow> ko_at' asidp p s"
+   by readObject_arch_obj_at'_method+
+
+
+lemma readObject_misc_arch_obj_at'[simp]:
+  shows
+  readObject_pte_at'[simplified]: "bound (readObject p s ::pte option) \<Longrightarrow> pte_at' p s" and
+  readObject_pde_at'[simplified]: "bound (readObject p s ::pde option) \<Longrightarrow> pde_at' p s" and
+  readObject_asid_pool_at'[simplified]: "bound (readObject p s ::asidpool option) \<Longrightarrow> asid_pool_at' p s"
+   by readObject_arch_obj_at'_method+
+
+method no_ofail_readObject_method =
+  clarsimp simp: obj_at'_def readObject_def obind_def omonad_defs split_def projectKO_eq no_ofail_def,
+  rule ps_clear_lookupAround2, assumption+, simp,
+  blast intro: is_aligned_no_overflow,
+  clarsimp simp: bit_simps' project_inject obj_at_simps lookupAround2_known1 split: option.splits
+
+lemma no_ofail_arch_obj_at'_readObject_pte[simp]:
+  "no_ofail (obj_at' (P::pte \<Rightarrow> bool) p) (readObject p::pte kernel_r)"
+  by no_ofail_readObject_method
+
+lemma no_ofail_arch_obj_at'_readObject_pde[simp]:
+  "no_ofail (obj_at' (P::pde \<Rightarrow> bool) p) (readObject p::pde kernel_r)"
+  by no_ofail_readObject_method
+
+lemma no_ofail_arch_obj_at'_readObject_asidpool[simp]:
+  "no_ofail (obj_at' (P::asidpool \<Rightarrow> bool) p) (readObject p::asidpool kernel_r)"
+  by no_ofail_readObject_method
+
+lemma no_ofail_arch_misc_readObject:
+  shows
+  no_ofail_pte_at'_readObject[simp]: "no_ofail (pte_at' p) (readObject p::pte kernel_r)" and
+  no_ofail_pde_at'_readObject[simp]: "no_ofail (pde_at' p) (readObject p::pde kernel_r)" and
+  no_ofail_asidpool_at'_readObject[simp]: "no_ofail (asid_pool_at' p) (readObject p::asidpool kernel_r)"
+  by (clarsimp simp: typ_at_to_obj_at_arches no_ofail_def
+              dest!: no_ofailD[OF no_ofail_arch_obj_at'_readObject_pte]
+                     no_ofailD[OF no_ofail_arch_obj_at'_readObject_pde]
+                     no_ofailD[OF no_ofail_arch_obj_at'_readObject_asidpool])+
+
 (* aliases for compatibility with master *)
 
 lemmas getPTE_wp = setObject_pte.get_wp
@@ -285,32 +333,17 @@ lemma get_asid_pool_corres [corres]:
           (get_asid_pool p) (getObject p')"
   apply (simp add: getObject_def get_asid_pool_def get_object_def split_def)
   apply (rule corres_no_failI)
-   apply (rule no_fail_pre, wp)
-   apply (clarsimp simp: typ_at'_def ko_wp_at'_def)
-   apply (case_tac ko; simp)
-   apply (rename_tac arch_kernel_object)
-   apply (case_tac arch_kernel_object, simp_all)[1]
-   apply (clarsimp simp: lookupAround2_known1
-                         projectKOs)
-   apply (clarsimp simp: obj_at'_def projectKOs objBits_simps
-                         archObjSize_def)
-   apply (erule (1) ps_clear_lookupAround2)
-     apply simp
-    apply (erule is_aligned_no_overflow)
-   apply simp
-   apply (clarsimp simp add: objBits_simps archObjSize_def
-                      split: option.split)
+   apply wp
+   apply (fastforce simp: typ_at_to_obj_at_arches
+                    dest: no_ofailD[OF no_ofail_asidpool_at'_readObject])
   apply (clarsimp simp: in_monad loadObject_default_def projectKOs)
   apply (simp add: bind_assoc exec_gets)
   apply (drule asid_pool_at_ko)
   apply (clarsimp simp: obj_at_def)
   apply (simp add: return_def)
-  apply (simp add: in_magnitude_check objBits_simps
-                   archObjSize_def pageBits_def)
-  apply clarsimp
   apply (clarsimp simp: state_relation_def pspace_relation_def)
   apply (drule bspec, blast)
-  apply (clarsimp simp: other_obj_relation_def asid_pool_relation_def)
+  apply (clarsimp simp: other_obj_relation_def asid_pool_relation_def obj_at'_def projectKOs)
   done
 
 lemma aligned_distinct_relation_asid_pool_atI'[elim]:
@@ -384,25 +417,16 @@ lemma get_pde_corres [corres]:
      (get_pde p) (getObject p')"
   apply (simp add: getObject_def get_pde_def get_pd_def get_object_def split_def bind_assoc)
   apply (rule corres_no_failI)
-   apply (rule no_fail_pre, wp)
-   apply (clarsimp simp: ko_wp_at'_def typ_at'_def lookupAround2_known1)
-   apply (case_tac ko; simp)
-   apply (rename_tac arch_kernel_object)
-   apply (case_tac arch_kernel_object; simp add: projectKOs)
-   apply (clarsimp simp: objBits_def cong: option.case_cong)
-   apply (erule (1) ps_clear_lookupAround2)
-     apply simp
-    apply (erule is_aligned_no_overflow)
-    apply (simp add: objBits_simps archObjSize_def word_bits_def)
+  apply wp
+   apply (fastforce simp: typ_at_to_obj_at_arches dest: no_ofailD[OF no_ofail_pde_at'_readObject])
    apply simp
   apply (clarsimp simp: in_monad loadObject_default_def projectKOs)
   apply (simp add: bind_assoc exec_gets)
   apply (clarsimp simp: pde_at_def obj_at_def)
   apply (clarsimp simp add: a_type_def return_def
                   split: if_split_asm Structures_A.kernel_object.splits arch_kernel_obj.splits)
-  apply (clarsimp simp: typ_at'_def ko_wp_at'_def)
+  apply (clarsimp simp: obj_at'_def projectKOs)
   apply (simp add: in_magnitude_check objBits_simps archObjSize_def pageBits_def pdeBits_def)
-  apply (clarsimp simp: bind_def)
   apply (clarsimp simp: state_relation_def pspace_relation_def)
   apply (drule bspec, blast)
   apply (clarsimp simp: other_obj_relation_def pde_relation_def)
@@ -473,19 +497,10 @@ lemma get_master_pde_corres [@lift_corres_args, corres]:
     apply (simp add: getObject_def get_pde_def get_pd_def  get_object_def
       split_def bind_assoc pde_relation_aligned_def get_master_pde_def)
     apply (rule corres_no_failI)
-     apply (rule no_fail_pre, wp)
-     apply (clarsimp simp: ko_wp_at'_def typ_at'_def lookupAround2_known1)
-     apply (case_tac ko, simp_all)[1]
-     apply (rename_tac arch_kernel_object)
-     apply (case_tac arch_kernel_object, simp_all add: projectKOs)[1]
-     apply (clarsimp simp: objBits_def cong: option.case_cong)
-     apply (erule (1) ps_clear_lookupAround2)
-       apply simp
-      apply (erule is_aligned_no_overflow)
-     apply (simp add: objBits_simps archObjSize_def word_bits_def)
-    apply simp
+     apply wp
+   apply (fastforce simp: typ_at_to_obj_at_arches dest: no_ofailD[OF no_ofail_pde_at'_readObject])
     apply (clarsimp simp: in_monad loadObject_default_def
-      projectKOs and_not_mask_twice)
+                          projectKOs and_not_mask_twice)
     apply (simp add: bind_assoc exec_gets)
     apply (clarsimp simp: pde_at_def obj_at_def)
     apply (clarsimp split:ARM_A.pde.splits)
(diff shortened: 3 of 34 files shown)


proof/refine/ARM/Invariants_H.thy
```diff
@@ -2293,17 +2296,19 @@ lemma magnitudeCheck_wp:
         \<longrightarrow> P s\<rbrace>
    magnitudeCheck ptr next bits
    \<lbrace>\<lambda>_. P\<rbrace>"
-  unfolding magnitudeCheck_def
-  apply wpsimp
+  unfolding magnitudeCheck_def read_magnitudeCheck_def
+  apply (simp add: gets_the_def exec_gets assert_opt_def valid_def
+                  return_def split_def fail_def
```
michaelmcinerney, line 2301: Indentation

proof/refine/ARM/KHeap_R.thy
```diff
@@ -74,65 +74,187 @@ abbreviation (input)
...
+  "o\<lbrace>P\<rbrace> f \<lbrace>Q\<rbrace> \<equiv> \<forall>s. P s \<longrightarrow> (\<exists>y. f s = Some y) \<longrightarrow>  Q (the (f s)) s"
+
+abbreviation (input)
+  oinvariant :: "('s,'a) lookup \<Rightarrow> ('s \<Rightarrow> bool) \<Rightarrow> bool" ("_ o\<lbrace>_\<rbrace>" [59,0] 60)
+where
+  "oinvariant f P \<equiv> o\<lbrace>P\<rbrace> f \<lbrace>\<lambda>_. P\<rbrace>"
+
+lemma asks_wp[rule_format, rotated, simp]:
+  "o\<lbrace>\<lambda>s. P (f s) s\<rbrace> asks f \<lbrace>P\<rbrace>"
+  by(simp add:  split_def asks_def oreturn_def obind_def)
```
michaelmcinerney, line 94: Extra space after `:`

proof/refine/ARM/KHeap_R.thy
```diff
@@ -147,60 +269,57 @@ lemma objBitsKO_bounded[simp]:
...
+   by readObject_obj_at'_method+
+
+lemmas readObject_ko_at'_tcb  [simp, dest!] = readObject_misc_ko_at'(1)
+lemmas readObject_ko_at'_ep   [simp, dest!] = readObject_misc_ko_at'(2)
+lemmas readObject_ko_at'_ntfn [simp, dest!] = readObject_misc_ko_at'(3)
+lemmas readObject_ko_at'_reply[simp, dest!] = readObject_misc_ko_at'(4)
+lemmas readObject_ko_at'_sc   [simp, dest!] = readObject_misc_ko_at'(5)
+
+lemma readObject_misc_obj_at':
+  "bound (readObject p s ::tcb option) \<Longrightarrow> tcb_at' p s"
```
michaelmcinerney, line 307: Would it be best to eithe add a space after `::` or remove the one before?

proof/refine/ARM/KHeap_R.thy
```diff
@@ -147,60 +269,57 @@ lemma objBitsKO_bounded[simp]:
...
+lemmas readObject_ntfn_at'[simp] = readObject_misc_obj_at'(3)[simplified]
+lemmas readObject_reply_at'[simp] = readObject_misc_obj_at'(4)[simplified]
+lemmas readObject_sc_at'[simp] = readObject_misc_obj_at'(5)[simplified]
 
 lemma getObject_tcb_at':
   "\<lbrace> \<top> \<rbrace> getObject t \<lbrace>\<lambda>r::tcb. tcb_at' t\<rbrace>"
-  by (clarsimp simp: valid_def getObject_def in_monad
-                     loadObject_default_def obj_at'_def projectKOs split_def
-                     in_magnitude_check objBits_simps')
+  unfolding getObject_def by wpsimp (clarsimp elim!:  ko_at_obj_at')
```
michaelmcinerney, line 322: Extra space after `:`

tools/haskell-translator/lhs_pars.py
```diff
@@ -128,7 +128,7 @@ def get_defs(filename):
 
 def top_transform(input, isLhs):
     """Top level transform, deals with lhs artefacts, divides
-        the code up into a series of seperate definitions, and
+        the code up into a series of separate definitions, and
```
michaelmcinerney, line 131: Thank you

lib/Monad_WP/OptionMonad.thy
```diff
@@ -91,10 +91,23 @@ definition
...
 
+definition oliftM :: "('a \<Rightarrow> 'b) \<Rightarrow> ('s,'a) lookup \<Rightarrow> ('s,'b) lookup" where
+  "oliftM f m \<equiv> do { x \<leftarrow> m; oreturn (f x) }"
+
+(* Reader monad interface: *)
+abbreviation (input)
+  "ask \<equiv> Some"
+
+definition
+  "asks f = do { v <- ask; oreturn (f v) }"
```
michaelmcinerney, line 108: Ahh this is some fancy notation trickery. I was very confused at first

michaelmcinerney (approved):
Very cool. I was at first a little worried that there would be a lot more to learn and become famliar with, but it's been made quite straightforward and nicely integrated with everything.

I have no idea about the Haskell translator, and I'm no expert with the very low level `getObject` and `setObject` stuff, but from what I do know, it looks great

lib/Monad_WP/OptionMonad.thy
lsf37: Got this from the Haskell Reader monad world

proof/refine/ARM/KHeap_R.thy
lsf37: I think we should standardise on space before and after, at least if it is a longer type name like this or something with a type parameter. We could have exceptions for single-letter type variables, i.e. `x::'a` still reads fine, but `x::'a word` visually suggests the wrong grouping `(x::'a) (word)`, whereas `x :: 'a word` is at least neutral.

There is also the unicode symbol that has the two colons close together. That one could be used without extra spaces, because the symbol basically has them, but I'd prefer the ASCII version of the symbol usually.

lib/Monad_WP/OptionMonad.thy
lsf37: (it's like `get`/`gets`)

spec/design/skel/PSpaceStorable_H.thy
```diff
@@ -54,23 +54,59 @@ where
 | "koTypeOf (KOKernelData) = KernelDataT"
 | "koTypeOf (KOArch e) = ArchT (archTypeOf e)"
 
+(* FIXME: move *)
```
lsf37, line 57: Please move to OptionMonad before merge

proof/refine/ARM/TcbAcc_R.thy
```diff
@@ -548,13 +548,17 @@ lemma assert_get_tcb_corres:
   apply assumption
   done
 
+schematic_goal threadGet_getObject:
+  "threadGet f t = ?x"
```
lsf37, line 552: Please spell out the `?x`, otherwise if we later change any definition below this we won't know if we need to make a change here or not.

proof/refine/ARM/KHeap_R.thy
```diff
@@ -74,65 +74,187 @@ abbreviation (input)
 where
   "set_obj' ptr obj s \<equiv> set_ko' ptr (injectKO obj) s"
 
+(* FIXME: move *)
+(** reader monad **)
+
+(* if state s satisfies P, and if f s returns a value, the value satisfies Q *)
+abbreviation
+  ovalid :: "('s \<Rightarrow> bool) \<Rightarrow> ('s,'a) lookup \<Rightarrow> ('a \<Rightarrow> 's \<Rightarrow>  bool) \<Rightarrow> bool"
```
lsf37, line 82: There is already a definition of `ovalid` in `lib/Monad_WP/OptionMonadWP.thy`. We should use that one if we need it.

I'd probably state lemmas about the option monad directly in the `[| P s; f s = Some v |] ==> Q s` form, because it's almost zero notational overhead over `ovalid` and more directly applicable as rule. Although we should probably think about this a bit more carefully before we start it everywhere, because we do lose the syntactic hooks that `ovalid` would provide.  (We did use it the "lightweight" way in RISCV).

proof/refine/ARM/KHeap_R.thy
```diff
@@ -74,65 +74,187 @@ abbreviation (input)
...
+  ("o\<lbrace>_\<rbrace>/ _ /\<lbrace>_\<rbrace>")
```
lsf37, line 83: The existing `ovalid` doesn't have the syntactic sugar. It might be nice to add that, including the `oinvariant` bit.

proof/refine/ARM/KHeap_R.thy
```diff
@@ -74,65 +74,187 @@ abbreviation (input)
...
+  oinvariant :: "('s,'a) lookup \<Rightarrow> ('s \<Rightarrow> bool) \<Rightarrow> bool" ("_ o\<lbrace>_\<rbrace>" [59,0] 60)
+where
+  "oinvariant f P \<equiv> o\<lbrace>P\<rbrace> f \<lbrace>\<lambda>_. P\<rbrace>"
+
+lemma asks_wp[rule_format, rotated, simp]:
+  "o\<lbrace>\<lambda>s. P (f s) s\<rbrace> asks f \<lbrace>P\<rbrace>"
+  by(simp add:  split_def asks_def oreturn_def obind_def)
+
+(* no fail for reader monad *)
+definition
```
lsf37, line 97: Please move to `lib/Monad_WP/OptionMonadWP` or `lib/Monad_WP/OptionMonad` before merge.

proof/refine/ARM/KHeap_R.thy
```diff
@@ -74,65 +74,187 @@ abbreviation (input)
...
+  ("o\<lbrace>_\<rbrace>/ _ /\<lbrace>_\<rbrace>")
+where
+  "o\<lbrace>P\<rbrace> f \<lbrace>Q\<rbrace> \<equiv> \<forall>s. P s \<longrightarrow> (\<exists>y. f s = Some y) \<longrightarrow>  Q (the (f s)) s"
+
+abbreviation (input)
+  oinvariant :: "('s,'a) lookup \<Rightarrow> ('s \<Rightarrow> bool) \<Rightarrow> bool" ("_ o\<lbrace>_\<rbrace>" [59,0] 60)
+where
+  "oinvariant f P \<equiv> o\<lbrace>P\<rbrace> f \<lbrace>\<lambda>_. P\<rbrace>"
+
+lemma asks_wp[rule_format, rotated, simp]:
```
lsf37, line 92: Following my arguments above, I'd probably state this directly without `ovalid` (to have the explicit form), and then add a separate `asks_wp` rule that folds it into `ovalid`. (The existing `ovalid` is a definition, not abbreviation)

proof/refine/ARM/KHeap_R.thy
```diff
@@ -74,65 +74,187 @@ abbreviation (input)
...
+  by (clarsimp simp: no_fail_r_def)
+
+method no_fail_r_readObject_method =
+  clarsimp simp: obj_at'_def readObject_def obind_def in_omonad split_def projectKO_eq no_fail_r_def,
+  rule ps_clear_lookupAround2, assumption+, simp,
+  blast intro: is_aligned_no_overflow,
+  clarsimp simp: objBits_simps' project_inject split: option.splits
+
+lemma no_fail_r_obj_at'_readObject_tcb[simp]:
+  "no_fail_r (obj_at' (P::tcb \<Rightarrow> bool) p) (readObject p::tcb kernel_r)"
```
lsf37, line 226: I can never quite decide if it is better to have `P` here or `\top`. `P` matches more often, but if used as intro rule will leave schematics in the final goal if the goal is produced `wp`-style. I think we may want `P` with `[simp]`, and `\top` with `intro!`. @mbrcknl and @Xaphiosis what is your take on these?

proof/refine/ARM/KHeap_R.thy
```diff
@@ -147,60 +269,57 @@ lemma objBitsKO_bounded[simp]:
...
+
+lemma readObject_misc_ko_at':
+  "readObject p s = Some (tcb :: tcb) \<Longrightarrow> ko_at' tcb p s"
+  "readObject p s = Some (ep :: endpoint) \<Longrightarrow> ko_at' ep p s"
+  "readObject p s = Some (ntfn :: notification) \<Longrightarrow> ko_at' ntfn p s"
+  "readObject p s = Some (reply :: reply) \<Longrightarrow> ko_at' reply p s"
+  "readObject p s = Some (sc :: sched_context) \<Longrightarrow> ko_at' sc p s"
+   by readObject_obj_at'_method+
+
+lemmas readObject_ko_at'_tcb  [simp, dest!] = readObject_misc_ko_at'(1)
```
lsf37, line 300: Better to move the `[simp, dest!]` to `readObject_misc_ko_at'`.

We can also give the different names directly in the lemma statement, with `and`. I forget if the attribute then has to be repeated, though.

In general I think it's better to make a collection from several named lemmas than to make a list and take it apart, because if you edit the list later (e.g. add something in the middle), the indices will shift and all the names are off, but you will not get any error messages until the lemma is used later, and it might be hard to track down that it is due to a different name.

In the other way around, you get an error message if something is renamed or removed, the only thing that will not be checked is if a new lemma is added, but not added to the collection, but existing proofs should not be affected.

proof/refine/ARM/Invariants_H.thy
```diff
@@ -1616,26 +1616,28 @@ lemma ps_clear_def2:
   apply (drule plus_one_helper, simp)
   done
 
+lemmas in_omonad = ofail_def oreturn_def oassert_def oassert_opt_def asks_def
```
lsf37, line 1619: There is an existing `in_omonad` in `lib/Monad_WP/OptionMonadND` -- please extend there instead of shadowing it, because it's a set used in a lot of places (AutoCorres, AInvs etc).

But: I don't think these definitions should be just unfolded -- `in_monad` and `in_omonad` are for the situation `(s', r) : monad s` or `monad s = Some r`. Only the rules that take apart these constructs into smaller components should be in that set.

We could make a set `omonad_defs` that contains the definitions you have here. That would then be use to always unfold all primitive monad operators. Then again, we should not need to do that often apart from when we're developing the monad framework (i.e. the equivalent of wp rules, which is what the `in_omonad` set is largely supposed to accomplish).

Still fine to define that `omonad_defs` set, we should just be aware that using it is a sign that we're not done with the framework yet. The only thing that is a bit annoying is `ohaskell_fail_def` and `ohaskell_assert_def`, which both should be defined only in `Haskell_Lib`, not in `OptionMonad*`. It'd be Ok to extend the collection there, though.

lsf37 (commented):
This is looking quite nice and I have the impression that the proofs are getting simpler too. Is that impression correct when you're doing them?

There is a bit of cleanup left to do with integration into the existing `OptionMonad` definitions. I can help with that if you want.

lsf37:
I thought I had run the python style formatter on the Haskell translator, but it seems there are still a few tweaks it wants. I can push a commit to correct these if you want.

proof/refine/ARM/KHeap_R.thy
lsf37: Actually, `no_ofail` already exists as well in `OptionMonadWP`, so we should use that and prove the additional rules there.

mktnk3: oh, great, will do that. I probably should have studied those files a bit more before getting started 😄

mktnk3:
> This is looking quite nice and I have the impression that the proofs are getting simpler too. Is that impression correct when you're doing them?

I would think so. Things are generally simpler, although part of it (a fair bit) might have been just that those lemmas could do with good cleanup.

I believe we can rewrite all the getter corres rules as equalities.

> There is a bit of cleanup left to do with integration into the existing `OptionMonad` definitions. I can help with that if you want.

Thanks, Gerwin. I am continuing with the awaken PR, so I might as well just do those cleanups myself along the way. But I would like to ask you to run the python formatter.

proof/refine/ARM/KHeap_R.thy
mktnk3: I agree with what you say.

> We can also give the different names directly in the lemma statement, with `and`. I forget if the attribute then has to be repeated, though.

This was what I was missing ... :) I'll rearrange things quickly.

mktnk3:
> I believe we can rewrite all the getter corres rules as equalities if we want. Not that that makes great difference, but some, for sure.

Well, that’s via `fun_of_m` and we probably don’t want that most of the time, so this was a bit too much to say. (my excuse is that there is quite a few of that in awaken Schedule_R... :D )

proof/refine/ARM/ArchAcc_R.thy
```diff
@@ -79,6 +79,53 @@ end
...
+                    obj_at_simps bit_simps' typ_at_to_obj_at_arches
+             split: option.splits if_split_asm
+
+lemma readObject_misc_arch_ko_at':
+  "readObject p s = Some (pte :: pte) \<Longrightarrow> ko_at' pte p s"
+  "readObject p s = Some (pde :: pde) \<Longrightarrow> ko_at' pde p s"
+  "readObject p s = Some (asidp :: asidpool) \<Longrightarrow> ko_at' asidp p s"
+   by readObject_arch_obj_at'_method+
+
+lemmas readObject_ko_at'_pte   [simp, dest!] = readObject_misc_arch_ko_at'(1)
```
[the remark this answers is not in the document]
mbrcknl: If you particularly want names for individual lemmas in addition to a name for the bundle, you can also do:
```
lemma readObject_misc_arch_ko_at'[simp, dest!]:
  shows readObject_ko_at'_pte: "readObject p s = Some (pte :: pte) \<Longrightarrow> ko_at' pte p s"
  and readObject_ko_at'_pde: "readObject p s = Some (pde :: pde) \<Longrightarrow> ko_at' pde p s"
  and readObject_ko_at'_asidp: "readObject p s = Some (asidp :: asidpool) \<Longrightarrow> ko_at' asidp p s"
  by readObject_arch_obj_at'_method+
```
Though I find it hard to decide whether this is better or worse.

proof/refine/ARM/KHeap_R.thy
mbrcknl: That's a reasonable assessment. So keep this lemma as is, but potentially add a second one?
```
lemmas no_fail_r_obj_at'_readObject_tcb_top[wp]
  = no_fail_r_obj_at'_readObject_tcb[where P="\top"]
```
The alternative would be to only have the rule with `\top`, but then you might need to add weakening rules for both `no_fail_r` and `obj_at'` to get it to work with `simp`. I don't think you want those weakening rules in the default `simp` set.

mbrcknl (commented):
I've done only a very quick scan, so I'm sure there are lots of details I've missed. But the general idea looks great!

lsf37:
> Thanks, Gerwin. I am continuing with the awaken PR, so I might as well just do those cleanups myself along the way. But I would like to ask you to run the python formatter.

No worries, will have a look at that later today.

lsf37:
> The alternative would be to only have the rule with \top, but then you might need to add weakening rules for both no_fail_r and obj_at' to get it to work with simp. I don't think you want those weakening rules in the default simp set.

I agree, weakening rules wouldn't be good to have in the simpset.

proof/refine/ARM/KHeap_R.thy
mktnk3: It seems that definitions in `OptionMonadWP` are not accessible from Refine files, while those in `OptionMonad` are. I am not sure why it is that way.

lsf37: Ah, that explains why they didn't show up.

That's entirely possibly, because AInvs on ARM doesn't use the option monad much or at all, so only the most basic bits would have been imported. We should include OptionMonadWP somewhere in the base AInvs inclusions, I think, so it is available by default. This might theoretically break something, so needs to be checked, but it should be fine if we didn't use the option monad.

proof/refine/ARM/Invariants_H.thy
mktnk3: In `Haskell_Lib`, `lookup` is not available, while it seems to fine to put `ohaskell_fail` and the other in `OptionMonad` along with others (they just take one extra argument that is discarded). Are you saying they shouldn’t be there because it is about haskell, or some other reasons?

I’ll remove `in_omonad` from here and add `omonad_defs` (in OptionMonad maybe).

proof/refine/ARM/KHeap_R.thy
mktnk3: Ok, I’ll chase it up.

proof/refine/ARM/Invariants_H.thy
lsf37: > Are you saying they shouldn’t be there because they are about haskell, or some other reasons?

Yes. Since we're now using the option monad in Haskell, it'd be perfectly fine to include those monads there already, i.e. add the `OptionMonad` theory you need to the `import` statements there. This would also make it naturally available to the rest of `Refine`.

proof/refine/ARM/KHeap_R.thy
mktnk3: So, ASpec imports `OptionMonad` only, via `Machine_A`: `OptionMonad` -> `OptionMonadND` -> `ExecSpec.MachineTypes` -> `Machine_A`

`OptionMonadND` imports `OptionMonad`, and `OptionMonadWP` imports `OptionMonadND`.

We could change `MachineTypes` to import `OptionMonadWP` instead of `OptionMonadND`.

But from what you are suggesting the other comment, we could import `OptionMonadWP`  in `HaskellLib`, which seems like a better option.

lsf37: Yes, I think that's good. The monad constants should be defined in `OptionMonad`, which is included in `ASpec`, and validity, lemmas, etc, should be in `OptionMonadWP/ND`, which makes sense in `Haskell_Lib`.

mktnk3: > We could change `MachineTypes` to import `OptionMonadWP` instead of `OptionMonadND`.
> 
> But from what you are suggesting the other comment, we could import `OptionMonadWP` in `HaskellLib`, which seems like a better option.

This gives an issue coming from the syntax overloading of `bind`, which seems obscure.

lsf37: I can have a look at that if you like. If the general `bind` is in scope, usually you can disambiguate with `NonDetMonad.bind` or things like that, or by annotating the type.

proof/refine/ARM/KHeap_R.thy
```diff
@@ -74,65 +74,187 @@ abbreviation (input)
...
+  by (simp add: no_fail_r_def ofail_def)
+
+lemma non_fail_r_asks_simp[simp]:
+  "no_fail_r P (asks f)"
+  unfolding asks_def get_def oreturn_def obind_def no_fail_r_def
+  by simp
+
+lemma non_fail_r_asks[wp]:
+  "no_fail_r \<top> (asks f)"
+  by simp
```
mktnk3, line 128: I just noticed that these go along with what was discussed between Matt and Gerwin about `P` vs `\<top>`.

proof/refine/ARM/KHeap_R.thy
mktnk3: That did work. Thanks, Gerwin!

mktnk3: Right, now that we have imported `OptionMonadWP` in Refine, I can see that there is `ogets` defined in `OptionMonadND` which is equivalent to `asks`. Shall we replace `asks` with `ogets`?

proof/refine/ARM/KHeap_R.thy
mktnk3: Using the existing `ovalid` instead of what is defined here works fine.

I am wondering if the second part is something we need to discuss further, or something we should try now. I can push an updated version after integrating the `OptionMonadWP` definition, and we can see what it looks like, maybe.

proof/refine/ARM/KHeap_R.thy
lsf37: Oh, well spotted, I overlooked that one. Should have been defined in `OptionMonad` of course. I blame DavidG :-)

I guess we have to keep the old one, because it is exported to AutoCorres, but `asks` is the more standard name for people familiar with that monad. I think we should make one of them a translation. Maybe the following:

 - keep the definition of `ask` and `asks` (in `OptionMonad`)
 - make `ogets` a (full) abbreviation for `asks`
 - provide a lemma `ogets_def` in in `OptionMonadWP` that corresponds to the old definition

This should keep the name people know in AutoCorres, as well as the name for people how are familiar with the reader monad.

proof/refine/ARM/KHeap_R.thy
lsf37: It'd be fine to do that later. Maybe make a Jira issue so we don't forget about it.

proof/refine/ARM/KHeap_R.thy
mktnk3: ok, I’ll do that.

proof/refine/ARM/KHeap_R.thy
mktnk3: Ok. I guess the point is how to make the definitions nicely work with what `wp` will leave us with. The definition of `ovalid` in `OptionMonadWP` or here make sense conceptually, but they would always need certain steps before being able to apply to what you get after applying `wp`.

My solution for now is to use this one
```
lemma use_ovalid:
  "[| ovalid P f Q; f s = Some r; P s |] ==> Q r s”
```
which seems to work fine.

lsf37: Yes, that looks good.

proof/refine/ARM/KHeap_R.thy
mktnk3: actually, "\top” versions are listed further below. I will make them `wp` rules rather than simp.

lib/HaskellLib_H.thy
```diff
@@ -15,13 +15,14 @@ imports
   NatBitwise
   "More_Numeral_Type"
   "Monad_WP/WhileLoopRules"
+  "Monad_WP/OptionMonadWP"
```
mktnk3, line 18: It seems that this has the side effect of `opt_mapE` reintroduced in `elim!` at some point before ArchDetype_AI, and that is breaking RISCV AInvs. It is removed from `elim!` in ArchInvariants_AI for RISCV, but it is reintroduced via an import of MonadicRewrite in SchedContext_AI, it seems (a late import with a FIXME). This needs to be sorted properly at some point, but, for now, I’ll remove `opt_mapE` from `elim!` for  RISCV at the beginning of ArchSchedule_A.

mktnk3:
I believe I’ve addressed all the comments so far. This will be ready for merging after a rebase, if there is no further comments.

lib/HaskellLib_H.thy
lsf37: Yes, that's fine. It's probably a sign that it shouldn't be `[elim!]` in the first place, but that is an even larger change.

lsf37 (approved):
Content looks good from my side, you only need to tweak some to the commit messages to make gitlint happy.

mktnk3:
> Content looks good from my side, you only need to tweak some to the commit messages to make gitlint happy.

Ah yes. Rebase and squash 😄

lsf37:
In fact, there's already a JIRA item about the `[elim!]` thing: VER-1367

But no resolution what to do..

proof/refine/ARM/ArchAcc_R.thy
```diff
@@ -79,6 +79,54 @@ end
...
+                    objBits_simps archObjSize_def projectKOs bit_simps' typ_at_to_obj_at_arches
+             split: option.splits if_split_asm
+
+lemma readObject_misc_arch_ko_at'[simp, dest!]:
+  shows
+  readObject_ko_at'_pte: "readObject p s = Some (pte :: pte) \<Longrightarrow> ko_at' pte p s" and
+  readObject_ko_at'_pde: "readObject p s = Some (pde :: pde) \<Longrightarrow> ko_at' pde p s" and
+  readObject_ko_at'_asidpool: "readObject p s = Some (asidp :: asidpool) \<Longrightarrow> ko_at' asidp p s"
+   by readObject_arch_obj_at'_method+
+
```
MitchellBuckley, line 93: nitpick: extra newline here

spec/design/skel/PSpaceStorable_H.thy
```diff
@@ -54,23 +54,43 @@ where
...
 
+definition
+  read_alignCheck :: "machine_word \<Rightarrow> nat \<Rightarrow> unit kernel_r" where
+  "read_alignCheck x n \<equiv> ounless ((x && mask n) = 0) $ read_alignError n"
+
 definition
   alignCheck :: "machine_word \<Rightarrow> nat \<Rightarrow> unit kernel" where
-  "alignCheck x n \<equiv> unless ((x && mask n) = 0) $ alignError n"
+  "alignCheck x n \<equiv> gets_the $ read_alignCheck x n"
+
```
MitchellBuckley, line 80: nitpick: extra newline here

MitchellBuckley (approved):
I haven't looked in much detail, but this looks good overall!!

It's confusing that this is referred to as option monad when it's a reader monad, but there's already a FIXME to address that.

proof/refine/ARM/ArchAcc_R.thy
```diff
@@ -802,8 +800,8 @@ lemma get_master_pte_corres [@lift_corres_args, corres]:
                bind_def get_pte_def get_object_def gets_def get_def
             split: if_split_asm Structures_A.kernel_object.splits
                    arch_kernel_obj.splits)
-   apply (clarsimp simp:typ_at'_def ko_wp_at'_def)
-   apply (clarsimp simp: in_magnitude_check objBits_simps
+      apply (clarsimp simp: obj_at'_def projectKOs)
```
VictorPhan, line 803: nitpick: indentation for subgoal count

proof/refine/ARM/Detype_R.thy
```diff
@@ -2709,9 +2705,9 @@ lemma locateCTE_ko_wp_at':
    \<lbrace>\<lambda>rv. ko_wp_at' \<top> rv \<rbrace>"
   apply (clarsimp simp:locateCTE_def split_def)
   apply wp
-  apply (clarsimp simp:cte_wp_at'_def getObject_def
-    gets_def split_def get_def bind_def return_def
-    ko_wp_at'_def lookupAround2_char1 assert_opt_def)
+  apply (clarsimp simp: cte_wp_at'_def getObject_def gets_the_def  obind_def omonad_defs
```
VictorPhan, line 2708: nitpick: remove extra space after `gets_the_def`

VictorPhan (approved):

PR #224, opened 2021-02-18
rt: make more use of reply projections

corlewis:


proof/refine/ARM/Ipc_R.thy
```diff
@@ -5096,19 +5096,32 @@ crunches setThreadState
...
+                      projectKO_sc get_refs_def)
+
+lemma ReplySchedContext_obj_at'_ref:
+  "(scp, ReplySchedContext) \<in> state_refs_of' s rptr
+   \<Longrightarrow> reply_at' rptr s
+   \<Longrightarrow> replySCs_of s rptr = Some scp"
+   by (clarsimp simp: obj_at'_real_def ko_wp_at'_def state_refs_of'_def
+                      projectKOs get_refs_def opt_map_def
+               split: option.splits)
+
```
mktnk3, line 5121: I removed these in my clean up PR that was merged last week. I believe these are not used anymore. Maybe something has gone a bit funny while rebasing.

proof/refine/ARM/Ipc_R.thy
```diff
@@ -5117,21 +5130,22 @@ lemma replyPush_sym_refs_list_refs_of_replies'[wp]:
          apply (wpsimp wp: bindScReply_sym_refs_list_refs_of_replies'
                            hoare_vcg_if_lift hoare_vcg_imp_lift' hoare_vcg_all_lift)
         apply (rule_tac Q="(\<lambda>_ s. sym_refs (list_refs_of_replies' s) \<and>
-                 (\<forall>rptr scp. obj_at' (\<lambda>ko. scReply ko = Some rptr) scp s \<longrightarrow>
-                  obj_at' (\<lambda>ko. replyNext ko = Some (Head scp)) rptr s) \<and>
-                 obj_at' (\<lambda>ko. replyPrev ko = None \<and> replyNext ko = None) replyPtr s)"
+                 (\<forall>rptr scp. obj_at' (\<lambda>ko. scReply ko = Some rptr) scp s
+                              \<longrightarrow> replySCs_of s rptr = Some scp) \<and>
+                 \<not> is_reply_linked replyPtr s \<and> replySCs_of s replyPtr = None)"
```
mktnk3, line 5135: I might be wrong here, but do you think we can replace `obj_at' (\<lambda>ko. scReply ko = Some rptr) scp s` with `scReplies_of` as well?
If we actually need `obj_at’`, can we decompose it into `sc_at’` and a projection?

proof/refine/ARM/Ipc_R.thy
```diff
@@ -5264,17 +5278,16 @@ lemma replyPush_sch_act_wf:
   by (wpsimp wp: sts_sch_act' hoare_vcg_all_lift hoare_vcg_if_lift hoare_drop_imps)
 
 lemma replyPush_invs':
-  "\<lbrace>\<lambda>s. invs' s \<and> sch_act_not callerPtr s \<and> obj_at' (\<lambda>ko. replyTCB ko = None) replyPtr s \<and>
+  "\<lbrace>\<lambda>s. invs' s \<and> sch_act_not callerPtr s \<and> reply_at' replyPtr s \<and>
         st_tcb_at' (\<lambda>st. st \<noteq> Inactive \<and> st \<noteq> IdleThreadState) callerPtr s \<and>
-        ex_nonz_cap_to' callerPtr s \<and> ex_nonz_cap_to' calleePtr s \<and> ex_nonz_cap_to' replyPtr s\<rbrace>
+        ex_nonz_cap_to' callerPtr s \<and> ex_nonz_cap_to' calleePtr s \<and> ex_nonz_cap_to' replyPtr s \<and>
+        replyTCBs_of s replyPtr = None\<rbrace>
```
mktnk3, line 5284: I agree with this change. When you actually need `obj_at’` part in the expression, separating it into `reply_at’`, etc., and the projection.

proof/refine/ARM/Reply_R.thy
```diff
@@ -33,6 +54,17 @@ lemma replyTCB_update_reply_projs[wp]:
      apply (clarsimp simp: ext opt_map_def obj_at'_def projectKO_eq)+
   done
 
+lemma updateReply_reply_projs:
+  "\<lbrace>\<lambda>s. \<forall>ko. ko_at' ko rptr s \<longrightarrow>
+       P (\<lambda>a. if a = rptr then replyNext_of (f ko) else replyNexts_of s a)
+         (\<lambda>a. if a = rptr then replyPrev (f ko) else replyPrevs_of s a)
+         (\<lambda>a. if a = rptr then replyTCB (f ko) else replyTCBs_of s a)
+         (\<lambda>a. if a = rptr then replySc (f ko) else replySCs_of s a)\<rbrace>
```
mktnk3, line 62: I am not entirely sure, but do we want to write this with implications or with `if _ then _ else _`?

proof/refine/ARM/Reply_R.thy
```diff
@@ -953,34 +946,35 @@ lemma bindScReply_sch_act_wf[wp]:
   by (wpsimp wp: sts_sch_act' hoare_vcg_all_lift hoare_vcg_if_lift hoare_drop_imps)
 
 lemma bindScReply_sym_refs_list_refs_of_replies':
-  "\<lbrace>\<lambda>s. sym_refs (list_refs_of_replies' s) \<and>
-        obj_at' (\<lambda>ko. replyPrev ko = None \<and> replyNext ko = None) replyPtr s \<and>
-        (\<forall>oldReplyPtr. obj_at' (\<lambda>ko. scReply ko = Some oldReplyPtr) scPtr s \<longrightarrow>
-         obj_at' (\<lambda>ko. replyNext ko = Some (Head scPtr)) oldReplyPtr s)\<rbrace>
+  "\<lbrace>\<lambda>s. sym_refs (list_refs_of_replies' s)
+        \<and> \<not> is_reply_linked replyPtr s \<and> replySCs_of s replyPtr = None
+        \<and> (\<forall>oldReplyPtr. obj_at' (\<lambda>ko. scReply ko = Some oldReplyPtr) scPtr s
```
mktnk3, line 951: I guess this `obj_at’` is needed. Can we change this to `sc_at’` and a projection, too?

proof/refine/ARM/Reply_R.thy
```diff
@@ -953,34 +946,35 @@ lemma bindScReply_sch_act_wf[wp]:
...
-                 opt_map_Some_def obj_at'_def projectKO_eq get_refs_def
-           elim: delta_sym_refs split: if_splits)
+                 opt_map_Some_def obj_at'_def projectKO_eq
+           elim: delta_sym_refs split: if_splits)[1]
 
 lemma bindScReply_if_live_then_nonz_cap':
   "\<lbrace>if_live_then_nonz_cap'
     and ex_nonz_cap_to' scPtr and ex_nonz_cap_to' replyPtr
-    and (\<lambda>s. \<forall>rp. obj_at' (\<lambda>ko. scReply ko = Some rp) scPtr s \<longrightarrow> obj_at' (\<lambda>ko. replyNext ko = Some (Head scPtr)) rp s )\<rbrace>
+    and (\<lambda>s. \<forall>rp. obj_at' (\<lambda>ko. scReply ko = Some rp) scPtr s
```
mktnk3, line 966: The same here.

mktnk3 (approved):
Very nice! It would be nice to do something like this through AInvs, too. One day.

proof/refine/ARM/Ipc_R.thy
corlewis: We almost definitely could, but I've mainly focused on the reply projections so far. I don't think much has been done with the SC projections yet, so if we started to use them more I expect that we would then need to prove that various functions preserve them.

My feeling is that this would be nice to do, but I don't know if it's a priority at the moment.

proof/refine/ARM/Reply_R.thy
corlewis: I'm not sure either, but I just moved this lemma and extended it to all of the projections.

proof/refine/ARM/Ipc_R.thy
mktnk3: Hmm. I agree that we don’t want to go much further with SC projections at this point, but `scReplies_of` comes up a lot because of `sc_replies_relation`. So, it might work nicely if we make more use of it, and in a sense it does relate to replies as well. Having said that I am not sure how much more work it will require to do so.

proof/refine/ARM/Reply_R.thy
MitchellBuckley: I think it's good this way: it's very easy to read (easier than implications and conjunctions) and I don't think it's any harder to manage otherwise.

MitchellBuckley (approved):
Great stuff, I really like this!

proof/refine/ARM/Reply_R.thy
mktnk3: I remember some discussions that went like wpsimp being too eager to split if-statements, leading to preferring implication over if-then-else when expressions are small (I might be misremembering the details).
Having said that, this is almost similar to `fun_upd`, so probably we don’t need to worry too much anyway.

mktnk3: @lsf37  might remember more, maybe?

mbrcknl (approved):
Looks good!

PR #226, opened 2021-02-20
rt Reader Monad update (abstract)

mktnk3:
This adds the reader monad to the abstract spec.

I’ve tweaked the definitions so that the changes with the proofs are more or less minimum. This might mean that we are not quite making use of the reader monad in full, but I think this is what we want for now.

The main thing to note is: `read_object p s = kheap s p`

### DIFF
--- a/proof/invariant-abstract/ARM/ArchVSpace_AI.thy
+++ b/proof/invariant-abstract/ARM/ArchVSpace_AI.thy
@@ -1844,9 +1844,9 @@ lemma simpler_set_pd_def:
               False)
         else ({}, True))"
   apply (rule ext)
-  apply (auto simp: set_pd_def get_object_def simpler_gets_def assert_def
+  apply (auto simp: set_pd_def get_object_def simpler_gets_def assert_def gets_the_def assert_opt_def
                     return_def fail_def set_object_def get_def put_def bind_def a_type_def
-          split: Structures_A.kernel_object.split arch_kernel_obj.split)
+             split: Structures_A.kernel_object.split arch_kernel_obj.split option.splits)
   done
 
 lemma set_pd_valid_vs_lookup_map:
@@ -2817,7 +2817,8 @@ lemma simpler_store_pde_def:
         | _ => ({}, True))"
   apply     (auto simp: store_pde_def simpler_set_pd_def get_object_def simpler_gets_def assert_def
                         return_def fail_def set_object_def get_def put_def bind_def get_pd_def
-                  split: Structures_A.kernel_object.splits option.splits arch_kernel_obj.splits if_split_asm)
+                        gets_the_def assert_opt_def
+                 split: Structures_A.kernel_object.splits option.splits arch_kernel_obj.splits if_split_asm)
   done
 
 lemma pde_update_valid_vspace_objs:
--- a/proof/invariant-abstract/BCorres_AI.thy
+++ b/proof/invariant-abstract/BCorres_AI.thy
@@ -14,6 +14,14 @@ abbreviation "bcorres \<equiv> bcorres_underlying truncate_state"
 
 abbreviation "s_bcorres \<equiv> s_bcorres_underlying truncate_state"
 
+lemma get_object_bcorres[wp]:
+  "bcorres (get_object p) (get_object p)"
+  apply (simp add: get_object_def)
+  apply (simp add: gets_the_def gets_def bind_def get_def return_def assert_opt_def read_object_def)
+  apply (simp add: bcorres_underlying_def s_bcorres_underlying_def fail_def return_def
+            split: option.splits)
+  done
+
 lemma dxo_bcorres[wp]:
   "bcorres (do_extended_op f) (do_extended_op f)"
   apply (simp add: do_extended_op_def)
@@ -76,15 +84,15 @@ crunch (bcorres) bcorres[wp]:
 lemma get_cap_det:
   "(r,s') \<in> fst (get_cap p s) \<Longrightarrow> get_cap p s = ({(r,s)}, False)"
   apply (cases p)
-  apply (clarsimp simp add: in_monad get_cap_def get_object_def
-                     split: kernel_object.split_asm)
-   apply (clarsimp simp add: bind_def return_def assert_opt_def simpler_gets_def)
-  apply (simp add: bind_def simpler_gets_def return_def assert_opt_def)
-  done
+  apply (clarsimp simp: in_monad get_cap_def get_object_def read_object_def)
+  apply (rename_tac ko caps; case_tac ko; simp add: fail_def)
+  by (clarsimp simp: in_monad gets_the_def gets_def get_def bind_assoc assert_opt_def
+                     fail_def return_def bind_def
+              split: option.splits)+
 
 lemma get_object_bcorres_any[wp]:
   "bcorres_underlying (trans_state e) (get_object a) (get_object a)"
-  by (wpsimp simp: get_object_def)
+  by (wpsimp simp: get_object_def read_object_def)
 
 lemma get_cap_bcorres_any:
   "bcorres_underlying (trans_state e) (get_cap x) (get_cap x)"
@@ -111,6 +119,9 @@ lemma get_cap_helper:
 lemma is_final_cap_bcorres[wp]:
   "bcorres (is_final_cap a) (is_final_cap a)"
   by (simp add: is_final_cap_def is_final_cap'_def gets_def get_cap_helper | wp)+
+
+lemma read_object_truncate[simp]: "read_object a (truncate_state s) = read_object a s"
+  by (clarsimp simp: read_object_def)
 
 lemma get_tcb_truncate[simp]: "get_tcb a (truncate_state s) = get_tcb a s"
   by (simp add: get_tcb_def)
@@ -166,9 +177,5 @@ lemma preemption_point_bcorres[wp]:
   unfolding preemption_point_def by wpsimp
 
 crunch (bcorres) bcorres[wp]: cap_swap_for_delete truncate_state
-
-lemma gets_the_get_tcb_bcorres[wp]:
-  "bcorres (gets_the (get_tcb a)) (gets_the (get_tcb a)) "
-  by (wpsimp simp: gets_the_def bcorres_underlying_def assert_opt_def split: option.splits|rule conjI)+
 
 end
--- a/proof/invariant-abstract/CSpace_AI.thy
+++ b/proof/invariant-abstract/CSpace_AI.thy
@@ -160,7 +160,7 @@ lemma unpleasant_helper:
 
 lemma get_object_det:
   "(r,s') \<in> fst (get_object p s) \<Longrightarrow> get_object p s = ({(r,s)}, False)"
-  by (auto simp: in_monad get_object_def bind_def gets_def get_def return_def)
+  by (auto simp: in_monad get_object_def bind_def gets_def get_def return_def gets_the_def)
 
 
 lemma get_object_at_obj:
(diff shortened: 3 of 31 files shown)


proof/invariant-abstract/ARM/ArchVSpace_AI.thy
```diff
@@ -1844,9 +1844,9 @@ lemma simpler_set_pd_def:
               False)
         else ({}, True))"
   apply (rule ext)
-  apply (auto simp: set_pd_def get_object_def simpler_gets_def assert_def
+  apply (auto simp: set_pd_def get_object_def simpler_gets_def assert_def gets_the_def assert_opt_def
                     return_def fail_def set_object_def get_def put_def bind_def a_type_def
-          split: Structures_A.kernel_object.split arch_kernel_obj.split)
+          split: Structures_A.kernel_object.split arch_kernel_obj.split option.splits)
```
michaelmcinerney, line 1849: Nipick: indentation (though it was like this originally, so not too fussed)

proof/invariant-abstract/BCorres_AI.thy
```diff
@@ -14,6 +14,14 @@ abbreviation "bcorres \<equiv> bcorres_underlying truncate_state"
 
 abbreviation "s_bcorres \<equiv> s_bcorres_underlying truncate_state"
 
+lemma get_object_bcorres[wp]:
```
michaelmcinerney, line 17: I guess we'd like this for when this commit goes into `master`, but while I remember to ask, do we know whether we'd like `bcorres` for MCS?

proof/invariant-abstract/CSpace_AI.thy
```diff
@@ -483,12 +483,11 @@ lemma set_cap_cte_eq:
   cte_at p' s \<and> cte_wp_at P p t = (if p = p' then P c else cte_wp_at P p s)"
   apply (cases p)
   apply (cases p')
-  apply (auto simp: set_cap_def2 split_def in_monad cte_wp_at_cases
-                    get_object_def set_object_def wf_cs_upd
-             split: Structures_A.kernel_object.splits if_split_asm
-                    option.splits,
-         auto simp: tcb_cap_cases_def split: if_split_asm)
-  done
+ apply (clarsimp simp: set_cap_def2 split_def in_monad set_object_def
```
michaelmcinerney, line 486: Needs one more space I think

proof/invariant-abstract/KHeap_AI.thy
```diff
@@ -446,10 +446,11 @@ lemma get_simple_ko_actual_ko[wp]:
...
    \<lbrace> \<lambda>rv. obj_at (\<lambda>k. k = f rv) ep \<rbrace>"
-   by (fastforce simp: get_simple_ko_def get_object_def bind_def partial_inv_def
-                             valid_def gets_def get_def return_def in_fail
-                             assert_def obj_at_def split_def the_equality
-                   split: kernel_object.splits option.splits)
+    by (clarsimp simp: get_simple_ko_def get_object_def bind_def partial_inv_def
+                       valid_def gets_def get_def return_def in_fail gets_the_def fail_def
+                       assert_def obj_at_def split_def the_equality assert_opt_def
+                split: kernel_object.splits option.splits)
+
```
michaelmcinerney, line 454: Did we want another line here?

proof/invariant-abstract/KHeap_AI.thy
```diff
@@ -954,39 +955,12 @@ lemma dmo_invs:
...
-  apply clarsimp
-  apply (case_tac "kheap s' t = None")
-   apply fastforce
-  apply clarsimp
-  apply (rename_tac ko)
-  apply (case_tac ko; (solves auto)?)
-  apply clarsimp
-  apply blast
-  done
+  by (force simp: as_user_def bind_assoc bind_select_f_bind[symmetric] split_def
```
michaelmcinerney, line 958: Nice! I'm normally sceptical of large blocks like this, but the previous proof really did look unpleasant

proof/invariant-abstract/LevityCatch_AI.thy
```diff
@@ -39,10 +39,12 @@ lemma const_on_failure_wp :
...
   "(v, s') \<in> fst (get_cap p s) \<Longrightarrow> (s' = s)"
-  by (clarsimp simp: get_cap_def get_object_def in_monad
-                     split_def
-              split: Structures_A.kernel_object.splits)
-
+  by (clarsimp simp: get_cap_def get_object_def read_object_def gets_the_def
+                     split_def gets_def get_def assert_opt_def bind_def
+                     return_def fail_def assert_def
+              split: option.splits)
+     (rename_tac ko a; case_tac ko;
```
michaelmcinerney, line 46: Just a thought: I've seen quite a few cases where we have to make up variable names if we'd like to `rename_tac` something that's not last in the list of green variables. I guess we'd normally want to avoid a situation where we're `rename_tac`ing the, say, 9th of 16 new green variables, but it would be cool if we could use `_` here (I'm pretty sure that doesn't work currently)

spec/abstract/KHeap_A.thy
```diff
@@ -233,6 +257,14 @@ where
      return $ f sc
    od"
 
+definition (* update only the schedcontext in place, keeping the size *)
```
michaelmcinerney, line 260: Is this function not exactly equal to `update_sched_context`?

spec/abstract/KHeap_A.thy
```diff
@@ -19,14 +19,20 @@ for kernel objects.\<close>
 
 section "General Object Access"
 
+text \<open>The kernel reaer option monad\<close>
```
michaelmcinerney, line 22: "reader" misspelt

michaelmcinerney (approved):
Awesome!

There was a commented bit asking about moving lemmas and there was also a `declare`. I'm probably not the best to comment on those, but otherwise, looks great.

Just out of curiosity, I did take a quick look in jedit at the KHeap_A.thy file and saw that there are a few proofs there. These are checked by the bamboo tests, right? How about GitHub? What about lib?

spec/abstract/KHeap_A.thy
mktnk3: It is not. But this is still not final and not needed for this PR, so I will remove.

michaelmcinerney: Hmmm. 

```
lemma asdf:
  "read_update_sched_context ptr f = update_sched_context ptr f"
  apply (clarsimp simp: read_update_sched_context_def update_sched_context_def get_object_def)
  done
  ```

works for me. Or is there some other reason we'd like both at some stage? I guess we can come back to it later

Comment (mktnk3) [reply spec/abstract/KHeap_A.thy:260]: Ah I thought you meant literally. This is just an experiment that I forgot to come back and is not meant for this PR. So, as I said, it will be removed.

File: proof/invariant-abstract/KHeap_AI.thy (hunk 5 above)
Comment (MitchellBuckley) [reply proof/invariant-abstract/KHeap_AI.thy:957]: Agreed.

File: proof/invariant-abstract/KHeapPre_AI.thy [hunk 9]
```diff
@@ -133,10 +133,20 @@ lemma set_object_neg_ko:
...
-  apply (case_tac "kheap s t", simp_all add: get_tcb_def)
+  apply (case_tac "kheap s t", simp_all add: get_tcb_def read_object_def)
   apply (case_tac a, simp_all)
   done
 
+lemma get_tcb_read_object_Some:
+  "get_tcb t s = Some tcb \<longleftrightarrow> read_object t s = Some (TCB tcb)"
+  by (rule iffI;
+      clarsimp simp: get_tcb_def split: option.splits kernel_object.split_asm)
+
```
MitchellBuckley, line 144: nit: extra newline here.

MitchellBuckley (approved):
Fixups look fine! Do these changes only effect Abstract?

proof/invariant-abstract/BCorres_AI.thy
lsf37: I'm pretty sure it is necessary, unless we have fully made everything deterministic, including the cdt etc.

proof/invariant-abstract/LevityCatch_AI.thy
lsf37: `rename_tac` doesn't accept `_` arguments, yes. If someone wants to dig into it, that could probably be changed by treating `_` specially and only doing the underlying rename for the remaining arguments. The `subgoal` command does support the `_` syntax.  The change would have to be made upstream, otherwise we'd be incompatible with standard Isabelle.

spec/abstract/KHeap_A.thy
```diff
@@ -19,14 +19,20 @@ for kernel objects.\<close>
...
 definition
   get_object :: "obj_ref \<Rightarrow> (kernel_object,'z::state_ext) s_monad"
 where
-  "get_object ptr \<equiv> do
-     kh \<leftarrow> gets kheap;
-     assert (kh ptr \<noteq> None);
-     return $ the $ kh ptr
-   od"
+  "get_object ptr \<equiv> gets_the $ read_object ptr"
+(* this is equivalent to "get_object ptr \<equiv> gets_map kheap ptr" *)
```
lsf37, line 35: It'd be nice to state the comment as a lemma instead, i.e. `lemma get_object_gets_map: "get_object = gets_map kheap"`

spec/abstract/KHeap_A.thy
```diff
@@ -42,15 +48,19 @@ where
 section "TCBs"
 
 definition
-  get_tcb :: "obj_ref \<Rightarrow> 'z state \<Rightarrow> tcb option"
+  get_tcb :: "obj_ref \<Rightarrow> (tcb, 'z::state_ext ) r_monad"
 where
   "get_tcb tcb_ref state \<equiv>
-   case kheap state tcb_ref of
+   case read_object tcb_ref state of
```
lsf37, line 54: I would change this into a projection and separate reader like for `read_object`. 

Something like
```
definition
tcb_of obj == case obj of TCB tcb => Some tcb | _ => None

abbreviation
tcbs_of s == kheap s |> tcb_of

definition
get_tcb = swp tcbs_of
```

To not have to update all of the proofs, you could still define a lemma `get_tcb_def` as before, and maybe give the actual `get_tcb` definition the name `get_tcb_tcbs_of`.

We would then have the projection available for preconditions etc and have a direct connection to the definition.

spec/abstract/KHeap_A.thy
```diff
@@ -216,6 +226,15 @@ abbreviation
 
 section \<open>Scheduling Contexts\<close>
 
+definition
+  read_sched_context :: "obj_ref \<Rightarrow> (sched_context,'z::state_ext) r_monad"
+where
+  "read_sched_context ptr \<equiv> do {
+     kobj \<leftarrow> read_object ptr;
+     case kobj of SchedContext sc n \<Rightarrow> oreturn sc
```
lsf37, line 234: Like for tcbs, I'd factor out the `case`, which enables both projection and reader, i.e.

`sc_of obj = case obj of SchedContext sc n => Some sc | _ => None`

This one is a bit more non-standard, because we are discarding the `n`, but I think the version where we'd want to keep that would be `sc_n_of`, so that's fine.

With that you can have the same pattern as above:
```
abbreviation
scs_of s = kheap s |> sc_of

read_sched_context = swp scs_of
```

lsf37 (approved):
This looks good from my side. It'd be nice to add the projections, because it makes it more uniform, but it'd be Ok to derive the old definitions as lemmas to not disturb the proofs.

Can also be done in a separate PR.

lsf37:
@michaelmcinerney 
> Just out of curiosity, I did take a quick look in jedit at the KHeap_A.thy file and saw that there are a few proofs there. These are checked by the bamboo tests, right? How about GitHub? What about lib?

ASpec and Lib are both checked by Bamboo and by GitHub

mktnk3:
Thanks everyone for the reviews.

> This looks good from my side. It'd be nice to add the projections, because it makes it more uniform, but it'd be Ok to derive the old definitions as lemmas to not disturb the proofs.
> 
> Can also be done in a separate PR.

Right, yes, that is what I was wondering about. I’ll follow this up as a separate PR, but soon. In the meantime, we can use this one to make progress with the awaken PR.

VictorPhan (approved):

mktnk3:
> Do these changes only effect Abstract?

It is the updates for the abstract spec that are added in this PR. The design part (#222 ) was merged last week. The proof fixes are needed in both AInvs and Refine, as you can see.

PR #230, opened 2021-02-26
c-parser: use fresh names for temporaries

mbrcknl:
Prior to rendering an expression to SIMPL, the C parser extracts
function calls from the expression and reinserts them as new statements
placed just before the statement containing the expression. The result
of each such function call is assigned to a temporary variable which
takes the place of the function call in the original expression.

Prior to this commit, the C parser would not always generate fresh
temporary variable names when multiple temporaries were needed. In
particular, when the left-hand side of an assignment contained a
function call returning the same type as a function call in the
right-hand side expression, the extracted function calls would be
assigned to the *same* temporary variable.

This commit addresses the issue by carrying name generation state across
all expressions in each statement. It implements a state monad as an
abstract data type for this purpose.

Fixes https://sel4.atlassian.net/browse/VER-1389.

### DIFF
--- a/tools/c-parser/syntax_transforms.ML
+++ b/tools/c-parser/syntax_transforms.ML
@@ -4,6 +4,141 @@
  * SPDX-License-Identifier: BSD-2-Clause
  *)
 
+signature FUN_RM =
+sig
+  (* Statement lists. *)
+  type stmt_list
+  val stmt_list_of : Absyn.statement list -> stmt_list
+  val of_stmt_list : stmt_list -> Absyn.statement list
+  val @@ : stmt_list * stmt_list -> stmt_list
+  val @> : stmt_list * Absyn.statement list -> Absyn.statement list
+  val empty : stmt_list -> bool
+  (* A monad for extracting function calls that are embedded in expressions. *)
+  type 'rv fun_rm
+  val run : 'a fun_rm -> stmt_list * 'a
+  val run' : unit fun_rm -> Absyn.statement list
+  val scope_pre : 'a fun_rm -> ('a -> 'b fun_rm) -> (stmt_list * 'b) fun_rm
+  val scope_stmts : unit fun_rm -> Absyn.statement list fun_rm
+  val scope_rv : 'a fun_rm -> 'a fun_rm
+  val add_stmts : stmt_list -> unit fun_rm
+  val add_stmts_fst : (stmt_list * 'a) -> 'a fun_rm
+  val add_stmt : Absyn.statement -> unit fun_rm
+  val next_sym : string -> int fun_rm
+  val return : 'a -> 'a fun_rm
+  val pass : ('a -> unit fun_rm) -> 'a -> 'a fun_rm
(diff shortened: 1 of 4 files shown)


mbrcknl:
Testboard: [L4V-TESTBOARD11-ARMVERIFY-1](https://bamboo.ts.data61.csiro.au/browse/L4V-TESTBOARD11-ARMVERIFY-1)

Note, this PR is currently rebased over https://github.com/seL4/l4v/pull/231, which is the proof update for https://github.com/seL4/seL4/pull/310. This PR therefore shows some l4v proof updates included in https://github.com/seL4/l4v/pull/231. Once https://github.com/seL4/l4v/pull/231 is merged, this PR won't require an l4v proof update.

To explain what's going on here: `decodeARMMMUInvocation` previously contained a mistake in a branch hint which hit the C parser issue addressed in this PR, but just happened to do so in a way that was harmless. Rather than keep the dodgy branch hint and do a proof update to go with this C parser change, I decided to fix the branch hint. Added bonus: the proof update for the branch hint patch was smaller than what would have been required with this PR if we'd kept the old C code.

lsf37 (approved):
This looks fine as far as I can figure it out. I'm glad we have good test coverage for the parser for the standard cases ;-)

mn200 (approved):

tools/c-parser/syntax_transforms.ML
```diff
@@ -4,6 +4,131 @@
...
+fun f <$ g = return f <*> g
+fun f $> g = g <$ f
+fun f <:> g = pair <$ f <*> g
+fun f >> g = snd <$ (f <:> g)
+fun f >>- g = f >- (fn x => g x >> return x)
+
+fun mmap _ [] = return []
+  | mmap f (h::t) = cons <$ f h <*> mmap f t
+
+(* A simple difference list. *)
```
zaklogician, line 58: As we discussed on Mmost earlier: while the code will rarely see lists long enough to make naive concatenation suboptimal, I'm happy for the difflist to remain in place. However, given the use of ` option`, one should replace the highlighted comment with something more descriptive.

zaklogician (commented):
See remarks on "simple difference list" comment.

mbrcknl:
I made some tweaks to improve compatibility with older proofs. It should now be quite rare that this C parser produces a spec that is different from specs produced by earlier C parser versions, unless the earlier specs were fundamentally broken by the issue addressed here.

In particular, this PR is now completely independent of https://github.com/seL4/seL4/pull/310 and https://github.com/seL4/l4v/pull/231. Nevertheless, to make it easier to review the updates, I'll leave this PR rebased over https://github.com/seL4/l4v/pull/231 until I'm just about ready to merge.

I also added some more test cases covering the updates to short-circuit logical operators and conditional operators.

Testboard for a version of this PR that is *not* rebased over https://github.com/seL4/l4v/pull/231:
[L4V-TESTBOARD13-1](https://bamboo.ts.data61.csiro.au/browse/L4V-TESTBOARD13-1).

I'll merge after I've also tested with binary verification, likely late today or early tomorrow.

PR #233, opened 2021-03-01
Move invariant field update lemmas

corlewis:
This creates a new file early in the dependency tree to contain all of the lemmas showing that various invariants are preserved by different updates to the kernel state. There are no guarantees that I found and moved all of these lemmas, and suggestions are welcome for the name of this file (it is currently `InvariantUpdates_H.thy`).

This is just a first step towards cleaning these lemmas up. There are multiple locales in `Invariants_H` that could be updated to generate these lemmas (e.g. `PSpace_update_eq`), and there is also the experimental `add_upd_simps` that we could be improving and using more often.

The diff didn't apply neatly to the other architectures, so at the moment this is just for arm.

### DIFF
--- a/proof/crefine/ARM/ArchMove_C.thy
+++ b/proof/crefine/ARM/ArchMove_C.thy
@@ -77,7 +77,7 @@ lemma setCTE_asidpool':
    apply (simp add: updateObject_cte)
    apply (clarsimp simp: updateObject_cte typeError_def magnitudeCheck_def in_monad
                    split: kernel_object.splits if_splits option.splits)
-  apply (clarsimp simp: ps_clear_upd' lookupAround2_char1)
+  apply (clarsimp simp: ps_clear_upd lookupAround2_char1)
   done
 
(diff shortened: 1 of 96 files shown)


lsf37 (approved):
Awesome! Thanks for doing this.

mbrcknl (approved):
:+1:

corlewis:
Sorry, I missed a push before putting this PR up. There was also a `ps_clear_upd'` that had to be removed in `crefine/ARM/ArchMove_C.thy`.

corlewis:
I've added to this so that the first commit updates all architectures and the second will be for ARM and RISCV64.

MitchellBuckley (approved):
This looks great!!!

corlewis:
Ok, I ended up doing this for the other architectures as well. I'm pretty sure that's actually everything, so feel free to have a look at the new parts and otherwise I'll give it a final squash and cleanup the commits before merging.

PR #234, opened 2021-03-02
Fixes to send_ipc and sendIPC

MitchellBuckley:
**I recommend reviewing commit-by-commit.**

In this PR:

1. Update `send_ipc` (abstract) to match source. Fix AInvs.
2. Update `sendIPC` to be slightly easier for refinement.
3. Remove `sym_ref` assertions from `replyPush` and `schedContextDonate`
   a. These assertions can actually fail in executable code.
4. Fix up `refine` for these changes in Haskell
5. Some slight improvements to `replyPush` lemmas. 

As a side effect:
* 1 new heap
* many new `sym_refs` lemmas based on heaps.

### DIFF
--- a/proof/refine/ARM/Invariants_H.thy
+++ b/proof/refine/ARM/Invariants_H.thy
@@ -260,6 +260,33 @@ abbreviation tcbs_of' :: "kernel_state \<Rightarrow> obj_ref \<Rightarrow> tcb o
 abbreviation tcb_scs_of' :: "kernel_state \<Rightarrow> obj_ref \<Rightarrow> obj_ref option" where
   "tcb_scs_of' s \<equiv> tcbs_of' s |> tcbSchedContext"
 
+abbreviation scTCBs_of :: "kernel_state \<Rightarrow> obj_ref \<Rightarrow> obj_ref option" where
+  "scTCBs_of s \<equiv> scs_of' s |> scTCB"
+
+(* 
+  sym_heapd is a definition version of sym_heap (abbreviation).
+  This is temporary for now, it should be combined with sym_heap properly,
+  either within the current PR, or later.
+*)
+
+definition sym_heapd :: "('a \<rightharpoonup> 'b) \<Rightarrow> ('b \<rightharpoonup> 'a) \<Rightarrow> bool" where
+  "sym_heapd h1 h2 \<equiv> \<forall>p q. pred_map_eq q h1 p \<longleftrightarrow> pred_map_eq p h2 q"
+
+lemma sym_heapd_def2:
+  "sym_heapd h1 h2 = 
+  ((\<forall>p q. pred_map_eq q h1 p \<longrightarrow> pred_map_eq p h2 q) \<and> (\<forall>p q. pred_map_eq q h2 p \<longrightarrow> pred_map_eq p h1 q))"
+  unfolding sym_heapd_def by fastforce
+
+lemma sym_heapd_sym:
+  "sym_heapd h1 h2 = sym_heapd h2 h1"
+  unfolding sym_heapd_def by fastforce
+
+abbreviation tcbs_scs_sym_refs where
+  "tcbs_scs_sym_refs s \<equiv> sym_heapd (tcb_scs_of' s) (scTCBs_of s)"
+
+abbreviation replies_scs_sym_refs where
+  "replies_scs_sym_refs s \<equiv> sym_heapd (scReplies_of s) (replySCs_of s)"
+
 definition
   tcb_cte_cases :: "word32 \<rightharpoonup> ((tcb \<Rightarrow> cte) \<times> ((cte \<Rightarrow> cte) \<Rightarrow> tcb \<Rightarrow> tcb))" where
  "tcb_cte_cases \<equiv> [    0 \<mapsto> (tcbCTable, tcbCTable_update),
--- a/proof/refine/ARM/Invariants_H.thy
+++ b/proof/refine/ARM/Invariants_H.thy
@@ -256,14 +256,14 @@ abbreviation tcb_of' :: "kernel_object \<Rightarrow> tcb option" where
 abbreviation tcbs_of' :: "kernel_state \<Rightarrow> obj_ref \<Rightarrow> tcb option" where
   "tcbs_of' s \<equiv> ksPSpace s |> tcb_of'"
 
-abbreviation tcb_scs_of' :: "kernel_state \<Rightarrow> obj_ref \<Rightarrow> obj_ref option" where
-  "tcb_scs_of' s \<equiv> tcbs_of' s |> tcbSchedContext"
+abbreviation tcbSCs_of' :: "kernel_state \<Rightarrow> obj_ref \<Rightarrow> obj_ref option" where
+  "tcbSCs_of' s \<equiv> tcbs_of' s |> tcbSchedContext"
 
 abbreviation scTCBs_of :: "kernel_state \<Rightarrow> obj_ref \<Rightarrow> obj_ref option" where
   "scTCBs_of s \<equiv> scs_of' s |> scTCB"
 
 abbreviation scTCB_sym_refs where
-  "scTCB_sym_refs s \<equiv> sym_heap (tcb_scs_of' s) (scTCBs_of s)"
+  "scTCB_sym_refs s \<equiv> sym_heap (tcbSCs_of' s) (scTCBs_of s)"
 
 abbreviation scReply_sym_refs where
   "scReply_sym_refs s \<equiv> sym_heap (scReplies_of s) (replySCs_of s)"
--- a/proof/refine/ARM/Invariants_H.thy
+++ b/proof/refine/ARM/Invariants_H.thy
@@ -250,9 +250,8 @@ abbreviation scs_of' :: "kernel_state \<Rightarrow> obj_ref \<Rightarrow> sched_
 abbreviation scReplies_of :: "kernel_state \<Rightarrow> obj_ref \<Rightarrow> obj_ref option" where
   "scReplies_of s \<equiv> scs_of' s |> scReply"
 
-definition tcb_of' :: "kernel_object \<Rightarrow> tcb option" where
-  "tcb_of' kobj \<equiv> (case kobj of KOTCB tcb \<Rightarrow> Some tcb
-                               | _         \<Rightarrow> None)"
+abbreviation tcb_of' :: "kernel_object \<Rightarrow> tcb option" where
+  "tcb_of' \<equiv> projectKO_opt"
 
 abbreviation tcbs_of' :: "kernel_state \<Rightarrow> obj_ref \<Rightarrow> tcb option" where
   "tcbs_of' s \<equiv> ksPSpace s |> tcb_of'"
@@ -263,29 +262,11 @@ abbreviation tcb_scs_of' :: "kernel_state \<Rightarrow> obj_ref \<Rightarrow> ob
 abbreviation scTCBs_of :: "kernel_state \<Rightarrow> obj_ref \<Rightarrow> obj_ref option" where
   "scTCBs_of s \<equiv> scs_of' s |> scTCB"
 
-(* 
-  sym_heapd is a definition version of sym_heap (abbreviation).
-  This is temporary for now, it should be combined with sym_heap properly,
-  either within the current PR, or later.
-*)
+abbreviation scTCB_sym_refs where
+  "scTCB_sym_refs s \<equiv> sym_heap (tcb_scs_of' s) (scTCBs_of s)"
 
-definition sym_heapd :: "('a \<rightharpoonup> 'b) \<Rightarrow> ('b \<rightharpoonup> 'a) \<Rightarrow> bool" where
-  "sym_heapd h1 h2 \<equiv> \<forall>p q. pred_map_eq q h1 p \<longleftrightarrow> pred_map_eq p h2 q"
-
-lemma sym_heapd_def2:
-  "sym_heapd h1 h2 = 
-  ((\<forall>p q. pred_map_eq q h1 p \<longrightarrow> pred_map_eq p h2 q) \<and> (\<forall>p q. pred_map_eq q h2 p \<longrightarrow> pred_map_eq p h1 q))"
-  unfolding sym_heapd_def by fastforce
-
-lemma sym_heapd_sym:
-  "sym_heapd h1 h2 = sym_heapd h2 h1"
-  unfolding sym_heapd_def by fastforce
-
-abbreviation tcbs_scs_sym_refs where
-  "tcbs_scs_sym_refs s \<equiv> sym_heapd (tcb_scs_of' s) (scTCBs_of s)"
-
-abbreviation replies_scs_sym_refs where
-  "replies_scs_sym_refs s \<equiv> sym_heapd (scReplies_of s) (replySCs_of s)"
+abbreviation scReply_sym_refs where
+  "scReply_sym_refs s \<equiv> sym_heap (scReplies_of s) (replySCs_of s)"
 
 definition
   tcb_cte_cases :: "word32 \<rightharpoonup> ((tcb \<Rightarrow> cte) \<times> ((cte \<Rightarrow> cte) \<Rightarrow> tcb \<Rightarrow> tcb))" where
@@ -2018,7 +1999,7 @@ lemma valid_pspaceE' [elim]:
 lemma idle'_only_sc_refs:
   "valid_idle' s \<Longrightarrow> state_refs_of' s (ksIdleThread s) = {(idle_sc_ptr, TCBSchedContext)}"
   by (clarsimp simp: valid_idle'_def pred_tcb_at'_def obj_at'_def tcb_ntfn_is_bound'_def
-                     projectKO_eq project_inject state_refs_of'_def idle_tcb'_def tcb_of'_def)
+                     projectKO_eq project_inject state_refs_of'_def idle_tcb'_def)
 
 lemma idle'_not_queued':
   "\<lbrakk>valid_idle' s; sym_refs (state_refs_of' s);
@@ -4133,7 +4114,7 @@ lemma sym_refs_replyNext_replyPrev_sym:
 
 lemma reply_sym_heap_Next_Prev:
   "sym_refs (list_refs_of_replies' s') \<Longrightarrow> sym_heap (replyNexts_of s') (replyPrevs_of s')"
-  using sym_refs_replyNext_replyPrev_sym by clarsimp
+  using sym_refs_replyNext_replyPrev_sym by (clarsimp simp: sym_heap_def)
 
 lemmas reply_sym_heap_Prev_Next
   = sym_heap_symmetric[THEN iffD1, OF reply_sym_heap_Next_Prev]
--- a/proof/refine/ARM/Ipc_R.thy
+++ b/proof/refine/ARM/Ipc_R.thy
@@ -1970,6 +1970,197 @@ crunches cteDeleteOne
 crunches handleFaultReply
   for valid_objs'[wp]: valid_objs'
 
+lemma bind_sc_reply_weak_valid_sched_action[wp]:
+  "bind_sc_reply a b \<lbrace>weak_valid_sched_action\<rbrace>"
+  unfolding bind_sc_reply_def by wpsimp
+
+lemma bind_sc_reply_invs[wp]:
+  "\<lbrace> \<lambda>s. invs s
+         \<and> reply_at reply_ptr s
+         \<and> sc_at sc_ptr s
+         \<and> ex_nonz_cap_to reply_ptr s
+         \<and> ex_nonz_cap_to sc_ptr s
+         \<and> reply_sc_reply_at (\<lambda>sc_ptr'. sc_ptr' = None) reply_ptr s
+         \<and> reply_ptr \<in> fst ` replies_blocked s
+         \<and> reply_ptr \<notin> fst ` replies_with_sc s \<rbrace>
+    bind_sc_reply sc_ptr reply_ptr
+   \<lbrace> \<lambda>rv. invs \<rbrace>"
+  unfolding bind_sc_reply_def
+  supply if_weak_cong[cong del] if_split[split del]
+  apply (rule hoare_seq_ext[OF _ gscrpls_sp])
+  apply (rename_tac sc_replies')
+  apply (case_tac sc_replies'; simp)
+   apply (wpsimp wp: sched_context_donate_invs)
+     apply (wpsimp simp: invs_def valid_state_def valid_pspace_def
+                     wp: valid_irq_node_typ set_reply_sc_valid_replies_already_BlockedOnReply
+                         valid_ioports_lift)
+    apply (wpsimp wp: set_sc_replies_valid_replies update_sched_context_valid_idle)
+   apply clarsimp
+   apply (clarsimp simp: invs_def valid_state_def valid_pspace_def
+                          reply_sc_reply_at_def obj_at_def state_refs_of_def get_refs_def2
+                          sc_replies_sc_at_def pred_tcb_at_def is_tcb is_reply is_sc_obj_def
+                  split: if_splits
+                  elim!: delta_sym_refs)
+   apply safe
+        apply fastforce
+       apply fastforce
+      apply (clarsimp simp: valid_idle_def)
+     apply (rule replies_with_sc_upd_replies_new_valid_replies)
+       apply fastforce
+      apply (clarsimp simp: image_def)
+      apply (rule_tac x="(reply_ptr, b)" in bexI; fastforce)
+     apply (clarsimp simp: image_def)
+    apply (fastforce simp: invs_def valid_state_def valid_pspace_def
+                           reply_sc_reply_at_def obj_at_def state_refs_of_def get_refs_def2
+                           sc_replies_sc_at_def pred_tcb_at_def is_tcb is_reply is_sc_obj_def
+                    split: if_splits
+                    elim!: delta_sym_refs)
+   apply (clarsimp simp: idle_sc_no_ex_cap)
+  apply wpsimp
+     apply (wpsimp simp: invs_def valid_state_def valid_pspace_def
+                     wp: valid_irq_node_typ set_reply_sc_valid_replies_already_BlockedOnReply
+                         valid_ioports_lift valid_sc_typ_list_all_reply)
+    apply (wpsimp wp: set_sc_replies_valid_replies update_sched_context_valid_idle)
+   apply (wpsimp simp: get_simple_ko_def get_object_def
+                   wp: valid_sc_typ_list_all_reply valid_ioports_lift)
+  apply (subgoal_tac "list_all (\<lambda>r. reply_at r s) (a # list) \<and> reply_ptr \<notin> set (a # list) \<and> distinct (a # list)")
+   apply (clarsimp simp: invs_def valid_pspace_def valid_state_def)
+   apply (intro conjI)
+     apply (rule replies_with_sc_upd_replies_valid_replies_add_one, simp)
+       apply (clarsimp simp:replies_blocked_def image_def, fastforce)
+      apply simp
+     apply (clarsimp simp:sc_replies_sc_at_def obj_at_def)
+    apply (erule delta_sym_refs)
+     apply (clarsimp split: if_splits
+                     elim!: delta_sym_refs)
+    apply (clarsimp simp: reply_sc_reply_at_def obj_at_def state_refs_of_def get_refs_def2
+                          pred_tcb_at_def is_tcb is_reply is_sc_obj sc_at_pred_n_def
+                   split: if_splits
+                   elim!: delta_sym_refs)
+    apply (safe; clarsimp?)
+    apply (rename_tac rp1 tl s tptr scp sc r1 r2 n1)
+    apply (subgoal_tac "(rp1,scp) \<in> replies_with_sc s \<and> (rp1,sc_ptr) \<in> replies_with_sc s")
+     apply (clarsimp dest!: valid_replies_2_inj_onD )
+    apply (intro conjI)
+     apply (subgoal_tac "valid_reply r1 s")
+      apply (clarsimp simp: valid_reply_def refs_of_def obj_at_def is_sc_obj_def
+                     split: option.splits)
+      apply (rename_tac ko n2)
+      apply (case_tac ko; clarsimp simp: get_refs_def)
+      apply (erule disjE, clarsimp split: option.splits)+
+      apply (clarsimp simp: replies_with_sc_def sc_replies_sc_at_def obj_at_def split: option.splits)
+     apply (erule valid_objs_valid_reply, assumption)
+    apply(clarsimp simp: replies_with_sc_def sc_replies_sc_at_def obj_at_def)
+    apply (metis cons_set_intro)
+   apply (fastforce simp: idle_sc_no_ex_cap tcb_at_def is_tcb_def
+                    dest: pred_tcb_at_tcb_at get_tcb_SomeD)
+  apply (clarsimp simp del: distinct.simps list.pred_inject insert_iff)
+  apply (frule sc_replies_sc_at_subset_fst_replies_with_sc)
+  apply (frule invs_valid_objs)
+  apply (intro conjI)
+    apply (erule replies_blocked_list_all_reply_at)
+    apply (meson dual_order.trans invs_valid_replies valid_replies_defs(1))
+   apply fastforce
+  apply (erule (1) valid_objs_sc_replies_distinct)
+  done
+
+lemma replyPush_corres:
+  "can_donate = can_donate' \<Longrightarrow>
+   corres dc (invs and tcb_at caller and tcb_at callee and reply_at reply_ptr
+              and ex_nonz_cap_to reply_ptr
+              and st_tcb_at active caller
+              and reply_sc_reply_at (\<lambda>tptr. tptr = None) reply_ptr
+              and reply_tcb_reply_at (\<lambda>tptr. tptr = None) reply_ptr
+              and weak_valid_sched_action and scheduler_act_not caller)
+             (valid_release_queue_iff and valid_objs' and valid_queues and valid_queues')
+   (reply_push caller callee reply_ptr can_donate)
+   (replyPush caller callee reply_ptr can_donate')"
+  unfolding reply_push_def replyPush_def
+  apply clarsimp
+  apply add_sym_refs
+  apply (rule_tac Q="\<lambda>s. valid_replies'_sc_asrt reply_ptr s" in corres_cross_add_guard)
+   apply (fastforce elim: valid_replies_sc_cross)
+  apply (rule corres_stateAssert_implied[where P'=\<top>, simplified])
+   apply (rule corres_stateAssert_implied[where P'=\<top>, simplified])
+    apply (rule stronger_corres_guard_imp)
+      apply (simp add: get_tcb_obj_ref_def)
+      apply (rule corres_split_deprecated [OF _ threadget_corres[where r="(=)"]])
+         apply (rule corres_split_deprecated [OF _ threadget_corres[where r="(=)"]])
+            apply (rule corres_split_deprecated [OF _ replyTCB_update_corres])
+              apply (rule corres_split_deprecated[OF _ sts_corres])
+                 apply (rule corres_when2, clarsimp)
+                 apply simp
+                 apply (rule corres_split_deprecated [OF schedContextDonate_corres bindReplySc_corres])
+                  apply (wpsimp wp: sc_at_typ_at)
+                 apply wpsimp
+                apply simp
+               apply (wpsimp wp: set_thread_state_invs hoare_vcg_imp_lift'
+                                 hoare_vcg_all_lift sts_in_replies_blocked
+                                 set_thread_state_weak_valid_sched_action)
+              apply (wpsimp wp: hoare_vcg_imp_lift' sts_invs_minor')
+             apply clarsimp
+             apply (wpsimp wp: hoare_vcg_imp_lift' hoare_vcg_all_lift)
+            apply (clarsimp simp: valid_tcb_state'_def cong: conj_cong)
+            apply (wpsimp wp: hoare_vcg_imp_lift' hoare_vcg_all_lift updateReply_valid_objs')
+           apply (clarsimp simp: tcb_relation_def)
+          apply (wpsimp wp: thread_get_wp')
+          apply assumption
+         apply (wpsimp wp: threadGet_wp)
+        apply (clarsimp simp: tcb_relation_def)
+       apply (wpsimp wp: thread_get_wp')
+      apply (wpsimp wp: threadGet_wp)
+     apply (subgoal_tac "caller \<noteq> reply_ptr")
+      apply (subgoal_tac "caller \<noteq> idle_thread_ptr")
+       apply (clarsimp simp: invs_def valid_state_def
+                             valid_pspace_def cong: conj_cong)
+       apply (frule valid_objs_valid_tcbs, clarsimp)
+       apply (frule (1) valid_objs_ko_at[where ptr=caller])
+       apply (subgoal_tac "ex_nonz_cap_to caller s", clarsimp)
+        apply (frule (1) idle_no_ex_cap)
+        apply (frule (2) no_tcb_not_in_replies_with_sc, clarsimp)
+        apply (intro conjI)
+            apply (clarsimp simp: valid_obj_def valid_tcb_def)
+           apply clarsimp
+          apply (clarsimp simp: replies_blocked_def pred_tcb_at_def obj_at_def)
+         apply (erule delta_sym_refs_insert_only, simp)
+           apply (subst set_eq_subset, intro conjI)
+            apply (clarsimp simp: state_refs_of_def)
+           apply (clarsimp simp: state_refs_of_def obj_at_def is_tcb get_refs_def2)
+           apply (subgoal_tac "tcb_st_refs_of (tcb_state tcb) = {}", simp)
+            apply fastforce
+           apply (fastforce simp: tcb_st_refs_of_def pred_tcb_at_def obj_at_def)
+          apply (subst set_eq_subset, intro conjI)
+           apply (clarsimp simp: state_refs_of_def)
+          apply (clarsimp simp: state_refs_of_def obj_at_def is_reply get_refs_def2)
+          apply (clarsimp simp: obj_at_def sk_obj_at_pred_def)
+         apply simp
+        apply (subgoal_tac "sc_tcb_sc_at ((=) (Some caller)) y s")
+         apply (clarsimp simp: sc_at_pred_n_def)
+         apply (erule (1) if_live_then_nonz_capD)
+         apply (clarsimp simp: live_def live_sc_def, fastforce)
+        apply (subst sym_refs_bound_sc_tcb_iff_sc_tcb_sc_at[symmetric, OF refl eq_commute])
+         apply assumption
+        apply (clarsimp simp: pred_tcb_at_def obj_at_def)
+       apply (erule (1) if_live_then_nonz_capD)
+       apply (clarsimp simp: live_def  is_tcb obj_at_def)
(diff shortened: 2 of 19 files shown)


MitchellBuckley:
Since @VictorPhan has already reviewed this once, can I ask someone else to review this please? There is not a rush, but there are two other peices of work which rely on this.

VictorPhan (approved):
if it's not different than the version I reviewed, then I'll stay on approve

proof/refine/ARM/KHeap_R.thy
```diff
@@ -3921,4 +3935,315 @@ lemma threadGet_getObject:
...
   done
 
+lemma obj_at'_typ_at'[elim!]:
+  "obj_at' (P :: ('a :: pspace_storable) \<Rightarrow> bool) p s \<Longrightarrow>
+   obj_at' (\<top> :: ('a :: pspace_storable) \<Rightarrow> bool) p s"
+  by (clarsimp simp: obj_at'_real_def ko_wp_at'_def)
+
+lemma sc_tcbs_of_pred_map_equiv:
+  "sc_at' p s \<Longrightarrow>
+   obj_at' (\<lambda>x. scTCB x = Some t) p s = pred_map_eq t (scTCBs_of s) p"
```
corlewis, line 3945: Nice, I think we've needed lemmas like these ones for a while, that link `obj_at'` and projections.

proof/refine/ARM/KHeap_R.thy
```diff
@@ -3921,4 +3935,315 @@ lemma threadGet_getObject:
...
+  apply (clarsimp simp: valid_tcb'_def)
+  done
+
+lemma not_idle_scTCB:
+  "tcbs_scs_sym_refs s \<Longrightarrow>
+   valid_objs' s \<Longrightarrow>
+   valid_idle' s \<Longrightarrow>
+   p \<noteq> idle_sc_ptr \<Longrightarrow>
+   sc_at' p s \<Longrightarrow>
+   obj_at' (\<lambda>x. scTCB x \<noteq> Some idle_thread_ptr) p s"
```
corlewis, line 3993: These lemmas are really nice as well, I think there's been quite a few times where I've had to recreate something similar on the spot.

proof/refine/ARM/KHeap_R.thy
```diff
@@ -3921,4 +3935,315 @@ lemma threadGet_getObject:
...
+  "setObject c (n::notification) \<lbrace>\<lambda>s. P' (tcbs_of' s)\<rbrace>"
+  "setObject c (sc::sched_context) \<lbrace>\<lambda>s. P' (tcbs_of' s)\<rbrace>"
+  by setObject_easy_cases+
+
+lemma setObject_cte_tcb_scs_of'[wp]:
+  "setObject c (reply::cte) \<lbrace>\<lambda>s. P' (tcb_scs_of' s)\<rbrace>"
+  by setObject_easy_cases
+
+lemma setObject_tcb_scs_of'[wp]:
+  "setObject c (tcb::tcb) \<lbrace>\<lambda>s. P' (scs_of' s)\<rbrace>"
```
corlewis, line 4034: Is it worth doing this for more object types, like you did for `tcbs_of'`? Or is this a case of only adding different types as necessary.

proof/refine/ARM/Ipc_R.thy
```diff
@@ -4946,9 +4940,96 @@ lemma valid_replies'_no_tcb_not_linked:
...
+crunches schedContextUpdateConsumed
+  for tcb_scs_of'[wp]: "\<lambda>s. P (tcb_scs_of' s)"
+
+crunches doIPCTransfer
+  for replySCs_of[wp]: "\<lambda>s. P (replySCs_of s)"
+  (wp: crunch_wps simp: crunch_simps)
+
+lemma schedContextUpdateConsumed_misc_heaps[wp]:
+  "schedContextUpdateConsumed scPtr \<lbrace>\<lambda>s. P (scReplies_of s) (replySCs_of s)\<rbrace>"
+  "schedContextUpdateConsumed scPtr \<lbrace>\<lambda>s. P (tcb_scs_of' s) (scTCBs_of s)\<rbrace>"
```
corlewis, line 4996: Wouldn't it be slightly stronger to combine these into one condition that handles all of the heaps? I don't have anything in mind that it would actually help with, but I also don't see a benefit in splitting them apart.

proof/refine/ARM/Ipc_R.thy
```diff
@@ -4946,9 +4940,96 @@ lemma valid_replies'_no_tcb_not_linked:
...
+  and scs_tcbs_of[wp]: "\<lambda>s. P (tcb_scs_of' s) (scTCBs_of s)"
+  (wp: crunch_wps)
+
+(* END *)
+
+lemma replySCs_of_obj_at'_retract:
+  "valid_objs' s \<Longrightarrow>
+   replies_scs_sym_refs s \<Longrightarrow>
+     \<forall>rPtr scPtr. obj_at' (\<lambda>a. scReply a = Some rPtr) scPtr s \<longrightarrow>
+            obj_at' (\<lambda>a. replyNext a = Some (Head scPtr)) rPtr s"
```
corlewis, line 5018: I'm not sure, but this might be useful earlier.

proof/refine/ARM/Invariants_H.thy
```diff
@@ -260,6 +260,30 @@ abbreviation tcbs_of' :: "kernel_state \<Rightarrow> obj_ref \<Rightarrow> tcb o
...
+  "sqheap_refs_retract heap symheap \<equiv> \<forall>p. sqheap_refs_retract_at heap symheap p"
+
+definition sqheap_refs_inv :: "('a \<rightharpoonup> 'b) \<Rightarrow> ('b \<rightharpoonup> 'a) \<Rightarrow> bool" where
+  "sqheap_refs_inv heap symheap \<equiv> sqheap_refs_retract heap symheap \<and> sqheap_refs_retract symheap heap"
+
+lemmas sqheap_refs_inv_defs = sqheap_refs_inv_def sqheap_refs_retract_def sqheap_refs_retract_at_def
+
+lemma sqheap_refs_inv_def2:
+  "sqheap_refs_inv heap symheap \<equiv> \<forall>p q. pred_map_eq q heap p \<longleftrightarrow> pred_map_eq p symheap q"
+  by (auto simp: atomize_eq sqheap_refs_inv_defs)
```
mbrcknl, line 279: I guess an `sqheap` is a *squashed* heap, as per our previous discussions on this? Might be worth a comment.

proof/refine/ARM/Ipc_R.thy
```diff
@@ -5199,12 +5265,180 @@ lemma cancelIPC_notin_epQueue:
...
+   \<lbrace>\<lambda>_. tcbs_scs_sym_refs\<rbrace>"
+  unfolding maybeReturnSc_def
+  apply (simp add: liftM_def)
+  apply (rule hoare_seq_ext[OF _ stateAssert_sp])
+  apply (rule hoare_seq_ext[OF _ get_ntfn_sp'])
+  apply wpsimp
+      apply (wps, wpsimp wp: setSchedContext_scTCBs_of)
+     apply wpsimp+
+    apply (wps, wpsimp wp: threadSet_tcb_scs_of')
+   apply (wpsimp wp: threadGet_wp)
```
corlewis, line 5284: Is there much interesting happening in this sequence? You might be able to do some sort of `(wpsimp wp: .. | wps)+` here and in the next lemma.

proof/refine/ARM/Ipc_R.thy
```diff
@@ -5199,12 +5265,180 @@ lemma cancelIPC_notin_epQueue:
...
+
+crunches cancelIPC
+  for tcbs_scs_sym_refs[wp]: tcbs_scs_sym_refs
+  (wp: crunch_wps threadSet_tcb_scs_of'_inv
+ ignore: setSchedContext threadSet)
+
+lemma cleanReply_replies_scs_sym_refs :
+  "\<lbrace>\<lambda>s. sqheap_refs_inv (scReplies_of s) (\<lambda>a. if a = rptr then None else replySCs_of s a)\<rbrace>
+  cleanReply rptr
+  \<lbrace>\<lambda>_. replies_scs_sym_refs\<rbrace>" (is "\<lbrace>?pre\<rbrace> _ \<lbrace>_\<rbrace>")
```
corlewis, line 5346: nitpick: indentation, and I guess you ended up not using the `?pre`

proof/refine/ARM/Ipc_R.thy
```diff
@@ -5199,12 +5265,180 @@ lemma cancelIPC_notin_epQueue:
...
+  apply (clarsimp simp: bind_assoc updateReply_def)
+  apply wpsimp
+         apply (wpsimp wp: cleanReply_replies_scs_sym_refs)
+        apply wp
+         apply wps
+         apply wpsimp
+        apply (rule_tac Q="\<lambda>_ s. (replySCs_of s) (the (replyPrev reply)) = None \<and>
+                 sqheap_refs_inv (scReplies_of s)
+                  (\<lambda>a. if a = rptr then None else replySCs_of s a)"
+       in hoare_strengthen_post[rotated], clarsimp)
```
corlewis, line 5367: nitpick: indentation here and in the next two `hoare_strengthen_posts`

mbrcknl (approved):
Looks good!

Perhaps squash the last two fixup commits in the preceding refine proofs commit.

proof/refine/ARM/Ipc_R.thy
```diff
@@ -5199,12 +5265,180 @@ lemma cancelIPC_notin_epQueue:
...
+  done
+
+crunches blockedCancelIPC, cancelSignal
+  for replies_scs_sym_refs[wp]: replies_scs_sym_refs
+  (wp: crunch_wps  ignore: setSchedContext setReply updateReply)
+
+lemma cancelIPC_replies_scs_sym_refs [wp]:
+  "\<lbrace>replies_scs_sym_refs and (\<lambda>s. sym_refs (list_refs_of_replies' s))\<rbrace>
+  cancelIPC t
+  \<lbrace>\<lambda>_. replies_scs_sym_refs\<rbrace>"
```
corlewis, line 5432: nitpick: indentation

corlewis (approved):
Awesome work! It's nice to see that the way that you have removed `sym_refs` has led to shorter proofs if anything.

The only problem that I might have is that in my `valid_replies'` work I've been using a `valid_replies'_sc_asrt` assertion in `bindScReply`, which requires `sym_refs` in the corres proof. Hopefully this won't actually be an issue though, the `replies_scs_sym_refs` you've introduced could be sufficient.

proof/refine/ARM/Invariants_H.thy
```diff
@@ -260,6 +260,30 @@ abbreviation tcbs_of' :: "kernel_state \<Rightarrow> obj_ref \<Rightarrow> tcb o
...
+
+abbreviation tcbs_scs_sym_refs where
+  "tcbs_scs_sym_refs s \<equiv> sqheap_refs_inv (tcb_scs_of' s) (scTCBs_of s)"
+
+abbreviation replies_scs_sym_refs where
+  "replies_scs_sym_refs s \<equiv> sqheap_refs_inv (scReplies_of s) (replySCs_of s)"
```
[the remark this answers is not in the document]
mbrcknl: They are the same.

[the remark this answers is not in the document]
mbrcknl: ...except for the usual disagreements over definitions vs abbreviations, etc. :-)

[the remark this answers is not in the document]
mktnk3: I think the difference goes far beyond that, and in my opinion what we have here as well as the related part of at the end of AInvs have excessive layers of definitions that we don’t benefit from and could do without.

The heap symmetry is a simple notion. It is perfectly fine to define it directly. Generally speaking, a definition (or abbreviation for that matter) should capture the concept succinctly, without coercing anyone into a particular way of thinking when there are other ways to arrive at that concept. Coercing people in that way will be counterproductive because it will impose extra mental load on them.

If some of the subconcepts are useful, it can be derived or given as alternatives, rather than forced building blocks.

But I also would like to say that I don’t see much point in redefining simple things such as forall and implication under different names. These are simple enough things that it is easy to capture the meaning just by glancing the statements. Why do we want to encapsulate and rename them to obscure their meaning? I don’t see the point of having `pred_map_eq` in the heap context to start with. It is equivalent to `m x = Some y` and, to be honest, I don’t quite understand why one would want to hide that.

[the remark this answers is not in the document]
mktnk3: Well, there might be some cases where a certain combination of simple components is very much recurrent and worthwhile to package it up, but I don’t think that is the case for what we have here. This is quite an overkill for a fairy simple concept and we could do with some streamlining.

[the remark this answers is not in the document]
MitchellBuckley: I didn't know that `sym_heap` existed, so I'm happy to look into use that. However, please keep in mind that I didn't pull these definitions out of thin air, I copied them from DetSchedSchedule so they are consistent with much of the existing proofs.

proof/refine/ARM/Ipc_R.thy
MitchellBuckley: Great point!

proof/refine/ARM/Invariants_H.thy
```diff
@@ -260,6 +260,30 @@ abbreviation tcbs_of' :: "kernel_state \<Rightarrow> obj_ref \<Rightarrow> tcb o
...
+definition sqheap_refs_retract_at :: "('a \<rightharpoonup> 'b) \<Rightarrow> ('b \<rightharpoonup> 'a) \<Rightarrow> 'a \<Rightarrow> bool" where
+  "sqheap_refs_retract_at heap symheap p \<equiv> \<forall>r. pred_map_eq r heap p \<longrightarrow> pred_map_eq p symheap r"
+
+definition sqheap_refs_retract :: "('a \<rightharpoonup> 'b) \<Rightarrow> ('b \<rightharpoonup> 'a) \<Rightarrow> bool" where
+  "sqheap_refs_retract heap symheap \<equiv> \<forall>p. sqheap_refs_retract_at heap symheap p"
+
+definition sqheap_refs_inv :: "('a \<rightharpoonup> 'b) \<Rightarrow> ('b \<rightharpoonup> 'a) \<Rightarrow> bool" where
+  "sqheap_refs_inv heap symheap \<equiv> sqheap_refs_retract heap symheap \<and> sqheap_refs_retract symheap heap"
+
+lemmas sqheap_refs_inv_defs = sqheap_refs_inv_def sqheap_refs_retract_def sqheap_refs_retract_at_def
```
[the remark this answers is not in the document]
MitchellBuckley: It stands for "squashed heap". I'm not sure about pronounciation. I'll put a comment in :)

proof/refine/ARM/Invariants_H.thy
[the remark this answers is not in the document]
mbrcknl: The point of making definitions is to control rewriting, and to build abstractions together with machinery that works at a higher level. It's very possible I failed to do that with the definitions that Mitch copied here. But in general, it's not sustainable to expose every low-level logical connective or quantifier to the simplifier and classical prover, so I would often prefer even a flawed attempt at building abstractions over exposing everything to the simplifier.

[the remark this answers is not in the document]
mktnk3: From what I have seen so far, `sym_heap` and projections play quite nicely with simplifier/automation. The only thing you might want to keep in mind is to remove `opt_mapE` from the rule set. That is usually the culprit when the simplifier is too aggressive.

Frankly, I am skeptical of the gain in choosing automation friendly definition at the cost of human brain efficiency for these particular constructs. Also, I really struggle to understand why you’d want to hide away projections under definitions when we have been moving towards using projections more extensively.

proof/refine/ARM/Ipc_R.thy
```diff
@@ -10,6 +10,20 @@ begin
 
 context begin interpretation Arch . (*FIXME: arch_split*)
 
+lemma replySCs_of_obj_at'_retract:
+  "valid_objs' s \<Longrightarrow>
+   replies_scs_sym_refs s \<Longrightarrow>
+     \<forall>rPtr scPtr. obj_at' (\<lambda>a. scReply a = Some rPtr) scPtr s \<longrightarrow>
+            obj_at' (\<lambda>a. replyNext a = Some (Head scPtr)) rPtr s"
```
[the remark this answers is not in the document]
MitchellBuckley: I agree. I always felt a bit uncomfortable with this lemma for exactly those reasons. I'll see if it's hard to replace.

proof/refine/ARM/Invariants_H.thy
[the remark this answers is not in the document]
mktnk3: I also have a comment from another angle. There is version 1 in AInvs, which is not the same thing as `sym_heap`. So, `sym_heap` was added, serving its purpose perfectly well. Then, now you changed it to version 2 to be the same thing as `sym_heap`. Add this will calls for some considerations on what to do with the version 1 and `sym_heap`, I would think.  So, could you tell me your plan on this if you are pushing for merging this version 2  (Mitch and Matt?) ?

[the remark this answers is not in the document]
MitchellBuckley: I'm not pushing for merging this version 2. I said above that I did not know that `sym_heap` existed and that I would look into using it. I have already partially updated this PR to use `sym_heap` (not pushed yet). If you are concerned about using `pred_map_eq` then we can have further discussion, but that is not something new that I am trying to introduce and so I did not consider that there was any problem using it.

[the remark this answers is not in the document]
mbrcknl: @mktnk3, since version 1 came before `sym_heap`, can you tell us what your plan was? I know that the stuff in AInvs is far from ideal, so I'd be happy if you have ways to improve it. W.r.t to this PR, I am ok with what @MitchellBuckley has proposed, but I'm also happy for you to give him some clear direction on what changes you'd like.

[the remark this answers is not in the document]
mktnk3: Version 1 as is does not achieve what `sym_heap` does. They are not the same in purpose, and they are defined totally different way (`sym_heap` is much simpler). So, I added `sym_heap` as something new and different. I have always had in mind that we need some cleanup in AInvs DetSched files, but I consider it as a separate thing.

But now, if this version 2 is the same in terms of semantics as `sym_heap`, and is a modification or improvement of version 1, that would call for some justification/adjustment, is what I meant. You can’t be just adding another thing on top of these, I’d say.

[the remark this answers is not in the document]
mktnk3: Ok, we can see how Mitch goes with the change for now, I suppose.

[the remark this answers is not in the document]
mktnk3: To be honest, I don’t see the need of introducing`pred_map_eq`, either. It not used in Refine so far, except for the very few places in some proofs for AInvs invariants. We have been doing fine with all the Refine proper stuff without using this definition, and we do use the projection extensively.

mktnk3:
I have one question: this PR removes sym_refs asserts, but would it be worth considering keeping asserts of sym_refs for subsets of refs (i.e., heaps)? Can these `replies_scs_sym_refs` be asserted and maybe crossed? Then we could probably simply the overall proof, maybe? Just a thought.

proof/refine/ARM/KHeap_R.thy
MitchellBuckley: I looked for this and could not find it, but it does exist in two places with reply objects. I like the direct equality that this pattern uses. We can discuss `pred_map` separately.

mktnk3: Well, we have `pred_map_eq t (scTCBs_of s) p = scTCBs_of s p = Some t`. And we also know `obj_at' (\<lambda>x. scTCB x = Some t) p s = sc_at’ p s & scTCBs_of s p = Some t`. So, this equality is pretty immediate if we can assume `sc_at’ p s`.

The question is, why do we need to go via `pred_map_eq` when we could easily have equality for `scTCBs_of s p = Some t` directly?

mktnk3: I just realised that this `obj_at (xxx)` pattern must have been copied from AInvs.

I would like you all to note the difference between `obj_at` and `obj_at’`. The latter asks for the alignment conditions, which we often do not need. If the information you actually need is only projections (object field values), we shouldn’t have `obj_at’`.

And, when we do need alignment conditions, it is better if we separate the `obj_at’` and the projection information because, by mixing those, we run the risk of having to prove alignment when we don’t need to.

The existing Refine proof is indeed a total mix of these, and we are forced to prove alignment again and again, which we could do without. So, for the new part of the proof we are adding, it would be good to be ensuring this distinction.

MitchellBuckley: In answer to your first question: `scTCBs_of s p = Some t` exposes heap evaluation to the simplifier via `opt_mapE`.
In this example
```
lemma test_lemma1:
  "⟦scTCBs_of s p = Some t; scs_of' s p = Some j⟧ ⟹ foo"
  apply (clarsimp)
  oops
```
the hypothesis will be immediately simplified to `scTCB sc = Some t; sc_of' v'a = Some sc; ksPSpace s p = Some v'a`. While this may provide detailed information about the state of the proof, it prevent the application of any drule or erule that would have matched on `scTCBs_of s p = Some t`. Yes, I can remove `opt_mapE` from [elim], but you could alternatively add `pred_map_eq` to [simp].

In this case, going via `pred_map_eq` preserves this common pattern and allows for repeated combination of other matching simp rules, erules and drules. I don't think using heaps directly will preserve this information.

MitchellBuckley: Addressing your second comment. I'm well aware of alignment information, it's somewhat patronising to have you explain it to me.

While this lemma is not optimal (I prefer the pattern you have provided in a separate comment), the entire point of the lemma is to remove `obj_at'` phrases from proofs in order to reason about heap values and ignore alignment. This lemma is close to the kind of replacement you highlighted elsewhere and so I'm not sure what point you're trying to make.

MitchellBuckley:
> I have one question: this PR removes sym_refs asserts, but would it be worth considering keeping asserts of sym_refs for subsets of refs (i.e., heaps)? Can these `replies_scs_sym_refs` be asserted and maybe crossed? Then we could probably simplify the overall proof somewhat, maybe? Just a thought.

I'm happy to consider this if it really does simplify the proofs, but I suspect it does not. My understanding is that a cross lemma will provide `sym_refs` on the concrete side when it exists on the abstract side. But in that case we must show that partial sym_refs is known on the abstract side and it will need to be lifted through abstract-side functions up to a point where `invs` holds in its entirety. So as far as I can see, we would be trading lifting-partial-sym-refs-on-concrete for lifting-partial-sym-refs-on-abstract. Perhaps I'm missed something, are there advantages that I'm missing?

proof/refine/ARM/KHeap_R.thy
mktnk3: >Yes, I can remove `opt_mapE` from [elim], but you could alternatively add `pred_map_eq` to [simp].

Removing `opt_mapE` from the rule set is what we have been doing for the whole of Refine proof so far. So, please stick with what we have been doing rather than introducing something different unless you are going to change the whole thing in that way.

MitchellBuckley: A further note, which supports for "hiding things behind definitions" is any proof of the form
```
lemma test_lemma2:
  "⟦sym_heap h1 h2'; foo⟧ ⟹ sym_heap h1 h2"
  apply_trace (clarsimp)
  oops
```
Any call to clarsimp (and not simp) will immediately introduce `\forall p p'.` into the context and leave a goal of the form `(h1 p = Some p') = (h2 p' = Some p)` which is not abbreviated to `sym_heap h1 h2` and so there is no way to use any rules or erules concerning `sym_heap`. In particular, `delta_sym_refs` is an elim rule of this form.

mktnk3: > In this case, going via `pred_map_eq` preserves this common pattern and allows for repeated combination of other matching simp rules, erules and drules. I don't think using heaps directly will preserve this ~information~ pattern.

I don’t think I understand fully what you are trying to say here, but having this definition will leave us with numerous lemmas that we have written so far that are not in the `pred_map_eq` pattern that won’t match with then. Please consider that we have been doing fine with our approach so far.

MitchellBuckley: > Removing opt_mapE from the rule set is what we have been doing for the whole of Refine proof so far. So, please stick with what we have been doing rather than introducing something different unless you are going to change the whole thing in that way.
I'm really not introducing anything different Miki, there are 900+ instances of `pred_map` in Ainvs. I was never told that proofs in refine needed to avoid these patterns and this PR is the first time that we've had any conflict on the issue. If you want to make a new decision going forward, that's fine, but it's simply not true that I'm introducing something different.

mktnk3: > A further note, which supports for "hiding things behind definitions" is any proof of the form
> 
> ```
> lemma test_lemma2:
>   "⟦sym_heap h1 h2'; foo⟧ ⟹ sym_heap h1 h2"
>   apply_trace (clarsimp)
>   oops
> ```
> 
> Any call to clarsimp (and not simp) will immediately introduce `\forall p p'.` into the context and leave a goal of the form `(h1 p = Some p') = (h2 p' = Some p)` which is not abbreviated to `sym_heap h1 h2` and so there is no way to use any rules or erules concerning `sym_heap`. In particular, `delta_sym_refs` is an elim rule which we might want to use.

For this, I have made this suggestion as a DM on mattermost to you:
"by the way, if all these argument can be resolved by having sym_heap as definition, please do consider that”
so please consider that - and I think that would be the only definition layer that we need.

MitchellBuckley: > For this, I have made this suggestion as a DM on mattermost to you:
"by the way, if all these argument can be resolved by having sym_heap as definition, please do consider that”
so please consider that - and I think that would be the only definition layer that we need.

Fair enough.

corlewis: I don't think this conversation is making much progress and personally I think there are benefits and negatives in both directions. Can I suggest that we get together offline to properly work through the options and decide on a path forwards?

mktnk3: > > Removing opt_mapE from the rule set is what we have been doing for the whole of Refine proof so far. So, please stick with what we have been doing rather than introducing something different unless you are going to change the whole thing in that way.
> 
> I'm really not introducing anything different Miki, there are 900+ instances of `pred_map` in Ainvs. I was never told that proofs in refine needed to avoid these patterns and this PR is the first time that we've had any conflict on the issue. If you want to make a new decision going forward, that's fine, but it's simply not true that I'm introducing something different.

`pred_map` instances in AInvs have very little bearing on projections in Refine, I would think? And we are talking about `pred_map_eq`, right? We don’t have `pred_map_eq` in Refine, and we should keep it that way, is what I am saying.

Besides, just think about this. You are suggesting to introduce this `pred_map_eq` to package up `m x = Some y', and at the same time suggesting to put `pred_map_eq y m x = (m x = Some y)` in the simp set. I would think this is indicative that we could do without this roundtrip at all.

mktnk3:
I was assuming that the AInvs counterpart exists. Isn’t this layer of definitions adapted from AInvs’s side partial sym_refs constructs? And wasn’t it used for similar arguments?

MitchellBuckley:
> I was assuming that the AInvs counterpart exists. Isn’t this layer of definitions adapted from AInvs’s side partial sym_refs constructs? And wasn’t it used for similar arguments?

That's a good point, I'll check.

proof/refine/ARM/KHeap_R.thy
mktnk3: > I don't think this conversation is making much progress and personally I think there are benefits and negatives in both directions. Can I suggest that we get together offline to properly work through the options and decide on a path forwards?

I’m happy to meet but not sure when...

mktnk3: Sorry if I sounded patronising, that was not my intention. I just wanted to make sure that everyone is on the same page.

> While this lemma is not optimal (I prefer the pattern you have provided in a separate comment), the entire point of the lemma is to remove `obj_at'` phrases from proofs in order to reason about heap values and ignore alignment. This lemma is close to the kind of replacement you highlighted elsewhere and so I'm not sure what point you're trying to make.

My point is that this might not be the best pattern for Refine proofs. As I said earlier, it looks like this pattern is introduced in several places in the proof by instantiation, but this might not be the best approach in Refine.

MitchellBuckley: > Sorry if I sounded patronising, that was not my intention. I just wanted to make sure that everyone is on the same page.
> 
> > While this lemma is not optimal (I prefer the pattern you have provided in a separate comment), the entire point of the lemma is to remove `obj_at'` phrases from proofs in order to reason about heap values and ignore alignment. This lemma is close to the kind of replacement you highlighted elsewhere and so I'm not sure what point you're trying to make.
> 
> My point is that this might not be the best pattern for Refine proofs. As I said earlier, it looks like this pattern is introduced in several places in the proof by instantiation, but this might not be the best approach in Refine.

Thank you. Ok, if it's the `pred_map_eq` you'd prefer to avoid then I understand what you mean.

proof/refine/ARM/Ipc_R.thy
```diff
@@ -5199,12 +5265,180 @@ lemma cancelIPC_notin_epQueue:
...
+  apply wpsimp
+      apply (wps, wpsimp wp: setSchedContext_scTCBs_of)
+     apply wpsimp+
+    apply (wps, wpsimp wp: threadSet_tcb_scs_of')
+   apply (wpsimp wp: threadGet_wp)
+  apply (clarsimp simp: tcb_at'_ex_eq_all)
+  apply (drule sym, simp)
+  apply (subgoal_tac "pred_map_eq (the (tcbSchedContext tcb)) (tcb_scs_of' s) t")
+   apply (clarsimp simp: sqheap_refs_inv_defs pred_map_eq_upd)
+   apply (auto)[1]
```
michaelmcinerney: Does something else work here? If not, maybe change to `subgoal by auto`? (I think it'd be good if we could decide on what to do with these `auto`s)

proof/refine/ARM/Ipc_R.thy
```diff
@@ -5199,12 +5265,180 @@ lemma cancelIPC_notin_epQueue:
...
+  by (wpsimp wp: setSchedContext_scTCBs_of gts_wp')
+
+lemma replyRemoveTCB_tcbs_scs_sym_refs[wp]:
+  "replyRemoveTCB param_a \<lbrace>tcbs_scs_sym_refs\<rbrace>"
+  by (rule hoare_weaken_pre, wps, wpsimp, simp)
+
+crunches cancelIPC
+  for tcbs_scs_sym_refs[wp]: tcbs_scs_sym_refs
+  (wp: crunch_wps threadSet_tcb_scs_of'_inv
+ ignore: setSchedContext threadSet)
```
michaelmcinerney, line 5323: Indentation

proof/refine/ARM/Ipc_R.thy
```diff
@@ -5199,12 +5265,180 @@ lemma cancelIPC_notin_epQueue:
...
+  cleanReply rptr
+  \<lbrace>\<lambda>_. replies_scs_sym_refs\<rbrace>" (is "\<lbrace>?pre\<rbrace> _ \<lbrace>_\<rbrace>")
+  unfolding cleanReply_def
+  apply (clarsimp simp: bind_assoc updateReply_def)
+  apply (wpsimp wp: hoare_drop_imp | wps)+
+  done
+
+lemma replyRemoveTCB_replies_scs_sym_refs [wp]:
+  "\<lbrace>replies_scs_sym_refs and (\<lambda>s. sym_refs (list_refs_of_replies' s))\<rbrace>
+  replyRemoveTCB t
```
michaelmcinerney: Another space here

proof/refine/ARM/Ipc_R.thy
```diff
@@ -5199,12 +5265,180 @@ lemma cancelIPC_notin_epQueue:
...
+        apply wpsimp
+       apply wp
+        apply wps
+        apply wpsimp
+       apply wpsimp
+      apply wpsimp
+     apply (rule_tac Q="\<lambda>replya s. replies_scs_sym_refs s \<and> sym_refs (list_refs_of_replies' s)"
+    in hoare_strengthen_post[rotated])
+      apply (clarsimp split: if_splits)
+    apply (safe)
```
michaelmcinerney, line 5368: Is `safe` generally preferred? I think I remember @Xaphiosis preferring `inro conjI impI` or something similar (if there is no better option)

proof/refine/ARM/KHeap_R.thy
```diff
@@ -3921,4 +3935,315 @@ lemma threadGet_getObject:
...
+   apply (clarsimp simp: sc_tcbs_of_pred_map_equiv tcb_scs_of_pred_map_equiv)
+   apply (clarsimp simp: sqheap_refs_inv_defs tcb_scs_of_pred_map_equiv)
+  apply (frule obj_at'_typ_at', simp)
+  apply (frule obj_at_ko_at', clarsimp)
+  apply (frule (1) tcb_ko_at_valid_objs_valid_tcb')
+  apply (clarsimp simp: valid_tcb'_def)
+  done
+
+lemma not_idle_scTCB:
+  "tcbs_scs_sym_refs s \<Longrightarrow>
```
michaelmcinerney, line 3980: Use `[| ... ; ... |] ==> ...` format

michaelmcinerney (approved):
Nice!

I think now is the right time to pick an approach to the partial `sym_refs` problem, but that aside, this is looking good

proof/refine/ARM/KHeap_R.thy
```diff
@@ -3921,4 +3927,296 @@ lemma threadGet_getObject:
...
+  by (intro iffI; clarsimp simp: obj_at'_real_def ko_wp_at'_def pred_map_eq
+                                 projectKOs opt_map_def)
+
+lemma scReplies_of_pred_map_equiv:
+  "obj_at' (\<lambda>a. scReply a = Some sc) p s = 
+   (sc_at' p s \<and> pred_map_eq sc (scReplies_of s) p)"
+  by (intro iffI; clarsimp simp: obj_at'_real_def ko_wp_at'_def pred_map_eq
+                                 projectKOs opt_map_def)
+
+lemma tcbs_scs_sym_refs_equiv:
```
MitchellBuckley, line 3959: This is unused, will delete.

MitchellBuckley:
There are 3 major issues which we need to sort out for this PR. Some offline discussion seems to be in order.
1. `pred_map_eq`
2. `sym_heap` abbreviations and definitions.
3. Crossing partial `sym_heaps` over to abstract side using assertions.

Are there any further issues which should be added to this list?

MitchellBuckley:
> > I was assuming that the AInvs counterpart exists. Isn’t this layer of definitions adapted from AInvs’s side partial sym_refs constructs? And wasn’t it used for similar arguments?
> 
> That's a good point, I'll check.

I think that this definitely possible. If we proceeded down that line we would need to
1. implement two partial sym refs assertions in Haskell (or one assertion covering both partial sym refs invariants)
2. Possibly lift one of those partial sym refs assumptions through the abstract (as far as I can see, only the sc<->tcb sym_refs are already lifted through the abstract)
3. Balance the significance of this improvement against the time taken to implement it.

On point 3 I'm agnostic and happy to hear some arguments for this approach in offline discussion.

proof/refine/ARM/KHeap_R.thy
```diff
@@ -3921,4 +3927,259 @@ lemma threadGet_getObject:
...
+   \<Longrightarrow> (scp, ReplySchedContext) \<in> state_refs_of' s r"
+  apply (subgoal_tac "reply_at' r s")
+   apply (clarsimp simp: state_refs_of'_def obj_at'_real_def ko_wp_at'_def projectKO_reply)
+  apply (erule (1) KOReply_reply_at'[rotated])
+  apply (clarsimp simp: projectKO_reply)
+  done
+
+lemma invs'_tcbs_scs_sym_refs:
+  "\<lbrakk>sym_refs (state_refs_of' s); pspace_aligned' s; pspace_distinct' s; valid_objs' s\<rbrakk>
+   \<Longrightarrow> tcbs_scs_sym_refs s"
```
mktnk3, line 4118: I might be missing something, but we need `valid_objs’` here? Also, it would be nice to have some consistent names for this lemma and the next one.

proof/refine/ARM/KHeap_R.thy
```diff
@@ -3921,4 +3927,259 @@ lemma threadGet_getObject:
...
+  unfolding threadSet_def
+  apply (wpsimp wp: setObject_tcb_wp getObject_tcb_wp)
+  apply (clarsimp simp: tcb_at'_ex_eq_all)
+  apply (erule back_subst[where P=P], rule ext)
+  apply (clarsimp simp: opt_map_def obj_at'_real_def ko_wp_at'_def projectKOs)
+  done
+
+lemma replyNexts_Some_replySCs_None:
+  "replyNexts_of s rPtr \<noteq> None \<Longrightarrow> replySCs_of s rPtr = None"
+   by (clarsimp simp: obj_at'_real_def ko_wp_at'_def projectKOs opt_map_def split: option.split)
```
mktnk3, line 4175: Something very similar (and have the same name) exists in Reply_R, and that one is in a pair with the opposite. This one is actually a better statement. Can you please add the opposite version here, too? And maybe replace the one in Reply_R with this one?

proof/refine/ARM/KHeap_R.thy
```diff
@@ -3921,4 +3927,259 @@ lemma threadGet_getObject:
...
+  apply (intro iffI conjI impI allI)
+   apply (frule (2) TCBSC_pred_map_state_refs_of')
+   apply (frule (1) sym_refsD, simp add: SCTcb_state_refs_of' obj_at'_sc_tcbs_of_equiv)
+  apply (frule (2) SCTcb_pred_map_state_refs_of')
+   apply (frule (1) sym_refsD, simp add: TCBSchedContext_state_refs_of' obj_at'_tcb_scs_of_equiv)
+  done
+
+lemma sym_refs_replies_scs:
+  "\<lbrakk>sym_refs (state_refs_of' s); pspace_aligned' s; pspace_distinct' s; valid_objs' s\<rbrakk>
+   \<Longrightarrow> replies_scs_sym_refs s"
```
mktnk3, line 4129: We already have:
```
lemma sym_refs_replySCs_of_scReplies_of:
 "⟦sym_refs (state_refs_of' s'); pspace_aligned' s'; pspace_distinct' s'⟧
  ⟹ replySCs_of s' rp = Some scp ⟷ scReplies_of s' scp = Some rp"
  ```
which says the same thing just not phrased in terms of `sym_heap`, and the proof is pretty much direct. I think we can do the similar for the `tcbs_scs_sym_refs` one above, too.

File: proof/refine/ARM/KHeap_R.thy (hunk 20 above)
Comment (mktnk3) [reply proof/refine/ARM/KHeap_R.thy:4118]: See also the comment below.

**Review by mktnk3 — COMMENTED** [review]
Thanks for the updates, Mitch. I’ve added a few comments, I will continue with the rest later.

File: proof/refine/ARM/KHeap_R.thy [hunk 23]
```diff
@@ -3921,4 +3927,259 @@ lemma threadGet_getObject:
...
+  done
+
+lemma KOTCB_tcb_at':
+  "\<exists>v'. ksPSpace s p = Some (KOTCB v') \<Longrightarrow> pspace_aligned' s \<Longrightarrow> pspace_distinct' s \<Longrightarrow> tcb_at' p s"
+  apply (clarsimp simp: pred_map_def state_refs_of'_def obj_at'_real_def ko_wp_at'_def projectKO_tcb
+                        tcb_of'_Some)
+  apply (intro conjI)
+   apply (erule (1) pspace_alignedD')
+  apply (erule (1) pspace_distinctD')
+  done
```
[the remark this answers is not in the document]
mktnk3: We actually have this generalised version already: `aligned’_distinct’_obj_at’I`.

And the `clarsimp` lines in the proofs have a bunch of unused things. Just having `obj_at’_def` and `projectKOs` should be enough.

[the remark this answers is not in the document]
MitchellBuckley: Yup, can prove an `injectKO` version and it's not really any less usable than the original.

proof/refine/ARM/Ipc_R.thy
```diff
@@ -2007,6 +2006,287 @@ crunches handleFaultReply
...
+  apply (erule (1) valid_objs_sc_replies_distinct)
+  done
+
+lemma replyPush_corres:
+  "can_donate = can_donate' \<Longrightarrow>
+   corres dc (invs and tcb_at caller and tcb_at callee and reply_at reply_ptr
+              and ex_nonz_cap_to reply_ptr
+              and st_tcb_at active caller
+              and reply_sc_reply_at (\<lambda>tptr. tptr = None) reply_ptr
+              and reply_tcb_reply_at (\<lambda>tptr. tptr = None) reply_ptr
```
mktnk3, line 2200: So, since we already have `reply_at reply_ptr` in the preconditions, we could drop the alignment from these two and just use projections: `%s. replySCs_of s reply_ptr = None` and `%s. replyTCBs_of s reply_ptr = None`.

But, given the mixed style of existing proofs, it is probably not clear if this always helps. But if we are keeping these two lines, we probably want to change the `tptr` in the `reply_sc_reply_at` line to something else.

mktnk3: hmm. I think I am confusing the abstract and the concrete. Anyway, the point is that this is something possible, if we want.

mktnk3 (commented):
Do we want to make `tcbs_scs_sym_refs` and `replies_scs_sym_refs` look a bit more “concrete”? By using capitalisation or primes? Maybe?

proof/refine/ARM/Ipc_R.thy
MitchellBuckley: I agree it will be nicer to use projections. In this case I'm mainly just moving the lemma and it was already like that, so because this PR is already quite big I'd like to leave that kind of change for some later cleanup if that's ok.

MitchellBuckley: (I'll change the `tcb` to `sc` though)

MitchellBuckley:
> Do we want to make `tcbs_scs_sym_refs` and `replies_scs_sym_refs` look a bit more “concrete”? By using capitalisation or primes? Maybe?

Will do. Currently going with `scReply_sym_refs` and `scTCB_sym_refs`

proof/refine/ARM/Invariants_H.thy
```diff
@@ -250,16 +250,24 @@ abbreviation scs_of' :: "kernel_state \<Rightarrow> obj_ref \<Rightarrow> sched_
...
-  "tcb_of' kobj \<equiv> (case kobj of KOTCB tcb \<Rightarrow> Some tcb
-                               | _         \<Rightarrow> None)"
+abbreviation tcb_of' :: "kernel_object \<Rightarrow> tcb option" where
+  "tcb_of' \<equiv> projectKO_opt"
 
 abbreviation tcbs_of' :: "kernel_state \<Rightarrow> obj_ref \<Rightarrow> tcb option" where
   "tcbs_of' s \<equiv> ksPSpace s |> tcb_of'"
 
 abbreviation tcb_scs_of' :: "kernel_state \<Rightarrow> obj_ref \<Rightarrow> obj_ref option" where
   "tcb_scs_of' s \<equiv> tcbs_of' s |> tcbSchedContext"
```
mktnk3, line 260: This is something existing, but I think we want to rename this to `tcbSCs_of`.

MitchellBuckley: Agreed, but can we leave that for another time? I would like to get this merged so I can continue with `sendIPC_corres`

mktnk3: That is understandable, but, if you actually see that we currently have about five occurrences of this in the MCS Refine, where as this PR will add 30+ of them. So, I think it would make sense to change this here....

MitchellBuckley:
Closed in favour of a fresh PR at https://github.com/seL4/l4v/pull/256

PR #240, opened 2021-03-08
rt: move invariant field update lemmas

corlewis:
This cherry-picks the two commits from https://github.com/seL4/l4v/pull/233 and extends it with the various lemmas that have been added in rt. I also removed a duplicated section of proofs that I noticed.

The first two commits have already been reviewed in the original PR while the last two are new but relatively boring.


mktnk3 (approved):
Nice cleanup!

MitchellBuckley (approved):
This is excellent! I've not examined every line, but what I've seen looks great.

PR #245, opened 2021-03-12
rt: move schedContextDonate and related lemmas

corlewis:
I'm not quite sure what the best approach to reviewing this will be. It's almost all lemmas that have been moved with a couple of minor updates to proofs being required in a few places.

Most lemmas have ended up in TcbAcc_R, SchedContext_R or Schedule_R, so you might want to focus on where lemmas now are and if you agree with that file.

I have also reordered the dependency graph slightly so that Reply_R imports Schedule_R which imports SchedContext_R. This led to much less lemmas needing to be moved and I think results in all of the required results being available where we expect them to be needed.

### DIFF
--- a/proof/refine/ARM/Schedule_R.thy
+++ b/proof/refine/ARM/Schedule_R.thy
@@ -2604,139 +2604,6 @@ lemma possibleSwitchTo_invs':
   apply (clarsimp simp: obj_at'_def tcb_in_cur_domain'_def)
   done
 
-lemma setReleaseQueue_tcb_at'[wp]:
- "setReleaseQueue qs \<lbrace>typ_at' T tcbPtr\<rbrace>"
-  unfolding setReleaseQueue_def
-  by wpsimp
-
-lemma setReleaseQueue_ksReleaseQueue[wp]:
-  "\<lbrace>\<lambda>_. P qs\<rbrace> setReleaseQueue qs \<lbrace>\<lambda>_ s. P (ksReleaseQueue s)\<rbrace>"
-  by (wpsimp simp: setReleaseQueue_def)
-
-lemma setReleaseQueue_pred_tcb_at'[wp]:
- "setReleaseQueue qs \<lbrace>\<lambda>s. P (pred_tcb_at' proj P' t' s)\<rbrace>"
-  by (wpsimp simp: setReleaseQueue_def)
-
-crunches tcbReleaseDequeue
-  for valid_pspace'[wp]: valid_pspace'
-  and state_refs_of'[wp]: "\<lambda>s. P (state_refs_of' s)"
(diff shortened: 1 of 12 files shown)


MitchellBuckley (approved):
Love it :)

corlewis:
Whoops, I missed `schedContextDonate_corres`. I'll hopefully have that done as well and in this PR soon.

proof/refine/ARM/Schedule_R.thy
```diff
@@ -3041,4 +2908,336 @@ lemma ct_released_cross_weak:
...
+            schedContextDonate_valid_idle' schedContextDonate_if_live_then_nonz_cap')
+  apply (clarsimp simp: obj_at'_def projectKO_eq projectKO_sc)
+  apply (drule_tac ko=obj in sym_refs_ko_atD'[rotated, where p=scPtr])
+   apply (auto dest!: global'_sc_no_ex_cap
+               simp: ko_wp_at'_def obj_at'_def projectKO_eq projectKO_tcb
+                     pred_tcb_at'_def valid_idle'_def idle_tcb'_def refs_of_rev')
+  done
+
+
+
```
michaelmcinerney, line 3027: Couple extra lines

proof/refine/ARM/Schedule_R.thy
```diff
@@ -3041,4 +2908,336 @@ lemma ct_released_cross_weak:
...
+lemma tcbSchedDequeue_nonq:
+  "\<lbrace>valid_queues and tcb_at' t and K (t = t')\<rbrace>
+   tcbSchedDequeue t
+   \<lbrace>\<lambda>_ s. t' \<notin> set (ksReadyQueues s (d, p))\<rbrace>"
+  apply (rule hoare_gen_asm)
+  apply (simp add: tcbSchedDequeue_def)
+  apply (wpsimp wp: threadGet_wp)
+  apply (fastforce simp: valid_queues_def valid_queues_no_bitmap_def obj_at'_def inQ_def)
+  done
+
```
michaelmcinerney, line 3045: New lines

michaelmcinerney (approved):

proof/refine/ARM/Schedule_R.thy
corlewis: Good catch. I wasn't sure whether to put these `tcbSchedDequeue` lemmas here or even earlier, and forgot to come back and make a decision. I'll have a look at them and either move them to `TcbAcc_R`, or remove the extra lines surrounding them.

mktnk3 (approved):
Thanks for doing this Corey. It looks like more work than I thought, sorting out all the dependencies.

VictorPhan (approved):

PR #247, opened 2021-03-15
replyPop lemmas in Finalise_R

VictorPhan:
Rebased and cleaned up work for `replyPop_invs'`.
Ready to be reviewed, but yet to add fixes to broken proofs in later theory files due to these changes.


VictorPhan:
Oops. Merging to rt not master.

PR #249, opened 2021-03-17
testboardpush script for github testboard

lsf37:
This should work with the public GitHub testboard ([seL4/gh-testboard](https://github.com/seL4/gh-testboard)).

Use the same as before, should work with verification-manifest as well as verification-manifest-internal (remotes and repo names of verification-manifest-internal will appear publicly, though).

Sub-repo commands are not yet supported.

### DIFF
--- a/misc/testboard/testboardpush
+++ b/misc/testboard/testboardpush
@@ -1,0 +1,307 @@
+#!/usr/bin/env python3
+#
+# Copyright 2021, Data61, CSIRO (ABN 41 687 119 230)
+#
+# SPDX-License-Identifier: BSD-2-Clause
+#
+
+# End-user interface to internal L4.verified test board. The internal test board
+# tracks a Google Repo manifest file that points to bleeding-edge versions of
+# component repos.
+#
+# This script creates a suitable manifest from the current local repository and
+# uploads it to the test board. It also performs some error checking.
+
+import argparse
+import os
+import shutil
+import subprocess
+import sys
+import json
+import tempfile
+import getpass
+import time
+import lxml.etree as etree
+
+parser = argparse.ArgumentParser(description="L4.verified test board script")
+
+# as git commit -m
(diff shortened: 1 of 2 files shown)


lsf37:
The tests themselves are still running on bamboo -- I'll try one with this PR here.

lsf37:
Yup, this seems to have worked nicely. I'll push one more tweak to the README, and then this is ready to merge from my side.

misc/testboard/testboardpush
```diff
@@ -0,0 +1,307 @@
...
+        # This may fail if someone else has pushed a commit in the meantime.
+        try:
+            subprocess.check_call(['git', 'push', '--set-upstream', 'origin', args.testboard_branch], cwd=testboard_repo)
+        except subprocess.CalledProcessError:
+            fatal("Failed to push to test board.")
+
+        testboard_oneline = subprocess_output(['git', 'log', '--oneline'], cwd=testboard_repo).strip()
+        log("Done. Your test board commit is:\n  " + testboard_oneline)
+    finally:
+          shutil.rmtree(testboard_repo)
```
mbrcknl, line 304: Nit: indentation.

```suggestion
        shutil.rmtree(testboard_repo)
```

mbrcknl (approved):
Looks good! Liberating this from `l4v-internal` is an added bonus.

misc/testboard/testboardpush
lsf37: Interesting that the style check didn't catch this! Will have a look.

mbrcknl: I only noticed it when I diffed against the old script in l4v-internal.

lsf37: It was probably tabs vs spaces or something like that (the style checker accepts both version). I had commented out the `rmtree` for testing and who knows what fancy things my editor did that ;-). 

Will force-push the tweak.

PR #253, opened 2021-03-18
Fix refs_valid_procedures definition

ThreeFx:
The current formulation allows an interface of an instance to be mentioned multiple times in the from side of the same connection.

There is the possibility that I have overlooked/misunderstood how this is intended to work, but this looks like a bug to me.

### DIFF
--- a/camkes/adl-spec/Wellformed_CAMKES.thy
+++ b/camkes/adl-spec/Wellformed_CAMKES.thy
@@ -147,7 +147,7 @@ where
   "refs_valid_procedures component_instance procedures conns \<equiv>
     \<forall>(name, (req, proc)) \<in> set procedures.
       req = InterfaceRequired \<longrightarrow> proc \<noteq> [] \<longrightarrow>
-      (\<exists>1 y \<in> conns. \<exists>1 z \<in> conn_from (snd y). z = (component_instance, name))"
+      (\<exists>1 z \<in> (concat (map (\<lambda>y. conn_from (snd y)) conns)). z = (component_instance, name))"
 
 text \<open>
   For events and dataports, an interface can be left unconnected in a system with no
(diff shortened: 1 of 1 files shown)


lsf37:
Nice that someone is actually using these proofs :-). Can I ask what the project is?

TBH we have not been using these particular checks in the past few years. Not that they are not useful, we just haven't needed it for any of our current projects.

The changes as such are fine, but I'll look into what CAmkES would actually do with such a spec. You are most likely right that it should be excluded.

lsf37:
(Ignore the Haskell failures on Github, it's failing in the GHC install phase already, which I hope is a temporary problem only)

A full regression testboard should show up in the checks soon.

lsf37:
Ok, so our CAmkES wizards tell me that this situation is potentially valid, depending on the connector. I.e. it would be unusual, but not necessarily wrong for a component to have one connection that attaches to the same interface twice. If the connector supports that, it should generate valid code.

Is this a specific problem in your application that you need to avoid? I don't think it would be catastrophic to exclude in the formal treatment (if documented), but if it can be supported, then we probably should.

ThreeFx:
> Nice that someone is actually using these proofs :-). Can I ask what the project is?

Of course :) It's foundations for a verified Baseboard Management Controller.

> Ok, so our CAmkES wizards tell me that this situation is potentially valid, depending on the connector. I.e. it would be unusual, but not necessarily wrong for a component to have one connection that attaches to the same interface twice. If the connector supports that, it should generate valid code.

Hmm, in that case should the definition be stricken alltogether? Since if the connector supports such a case the `refs_valid_procedures` definition applies a nonsensical limitation anyway. Adding a regression test for detecting this is probably also helpful then.

lsf37:
Yes, you're right the existing test is already restrictive. Ok, in that case, I think this does make sense to do this test properly instead of just half. If we get to systems we want to verify where we do need such connectors, we can still relax it then.

lsf37 (approved):
The test board has passed, and I'm happy with the restriction now.

PR #254, opened 2021-03-22
riscv: use `uint8_t` for `register_t`

MitchellBuckley:
This PR verifies the changes found in the corresponding pull-request here:
https://bitbucket.ts.data61.csiro.au/projects/SEL4/repos/sel4/pull-requests/1817/overview

### DIFF
--- a/proof/crefine/RISCV64/StateRelation_C.thy
+++ b/proof/crefine/RISCV64/StateRelation_C.thy
@@ -207,7 +207,7 @@ definition
   \<and> tcb_queue_relation getNext getPrev hp queue NULL qhead"
 
 fun
-  register_from_H :: "register \<Rightarrow> machine_word"
+  register_from_H :: "register \<Rightarrow> register_idx"
   where
     "register_from_H RISCV64.LR = scast Kernel_C.LR"
 |   "register_from_H RISCV64.SP = scast Kernel_C.SP"
--- a/proof/crefine/RISCV64/SyscallArgs_C.thy
+++ b/proof/crefine/RISCV64/SyscallArgs_C.thy
@@ -602,11 +602,6 @@ lemma nat_less_4_cases:
   "(x::nat) < 4 \<Longrightarrow> x=0 \<or> x=1 \<or> x=2 \<or> x=3"
   by clarsimp
 
-lemma msgRegisters_scast:
-  "n < unat (scast n_msgRegisters :: machine_word) \<Longrightarrow>
-  unat (scast (index msgRegistersC n)::machine_word) = unat (index msgRegistersC n)"
-  by (simp add: kernel_all_global_addresses.msgRegisters_def fupdate_def update_def n_msgRegisters_def)
-
 lemma asUser_cur_obj_at':
   assumes f: "\<lbrace>P\<rbrace> f \<lbrace>Q\<rbrace>"
   shows "\<lbrace>\<lambda>s. obj_at' (\<lambda>tcb. P (atcbContextGet (tcbArch tcb))) (ksCurThread s) s \<and> t = ksCurThread s\<rbrace>
@@ -1058,7 +1053,6 @@ lemma index_msgRegisters_less':
 
 lemma index_msgRegisters_less:
   "n < 4 \<Longrightarrow> index msgRegistersC n <s 35"
-  "n < 4 \<Longrightarrow> index msgRegistersC n < 35"
   using index_msgRegisters_less'
   by (simp_all add: word_sless_msb_less)
 
@@ -1131,15 +1125,15 @@ lemma getSyscallArg_ccorres_foo:
        apply (rule conseqPre, vcg)
        apply (clarsimp simp: rf_sr_ksCurThread)
        apply (drule (1) obj_at_cslift_tcb)
-       apply (clarsimp simp: typ_heap_simps' msgRegisters_scast)
+       apply (clarsimp simp: typ_heap_simps')
        apply (clarsimp simp: ctcb_relation_def ccontext_relation_def
                              msgRegisters_ccorres atcbContextGet_def
                              carch_tcb_relation_def cregs_relation_def)
        apply (subst (asm) msgRegisters_ccorres)
         apply (clarsimp simp: n_msgRegisters_def)
-       apply (simp add: n_msgRegisters_def word_less_nat_alt)
-       apply (simp add: index_msgRegisters_less unat_less_helper)
-      apply wp[1]
+        apply (clarsimp simp: n_msgRegisters_def word_less_nat_alt word_upcast_0_sle)
+       apply (clarsimp simp: index_msgRegisters_less' ucast_up_less_bounded_iff_less_ucast_down')
+       apply wp[1]
      apply (wp getMRs_tcbContext)
     apply simp
    apply (rule ccorres_seq_skip [THEN iffD2])
--- a/proof/crefine/RISCV64/TcbAcc_C.thy
+++ b/proof/crefine/RISCV64/TcbAcc_C.thy
@@ -120,13 +120,12 @@ lemma register_from_H_less:
   by (cases hr, simp_all add: "StrictC'_register_defs")
 
 lemma register_from_H_sless:
-  "register_from_H hr <s 35"
+  "UCAST(register_idx_len \<rightarrow> int_literal_len) (register_from_H hr) <s 0x23"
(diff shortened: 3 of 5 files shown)


axel-h:
Is this PR on github also? Just wondering, aren't machine registers using 32 or 64 bit, so using an uint8_t there look quite odd.

mbrcknl:
> Is this PR on github also? Just wondering, aren't machine registers using 32 or 64 bit, so using an uint8_t there look quite odd.

I don't think there's a GitHub PR, but there's a copy of the branch here:
https://github.com/seL4/seL4/compare/mbrcknl/riscv-uint8-register-t...mbrcknl:mbrcknl/riscv-uint8-register-t

In the commit messages, I tried to convey that `register_t` is intended for *indexing* into the register set, not for storing register contents, and therefore needs only a few dozen distinct values. Let me know if you think it needs more clarification.

Also happy to make a GitHub PR if you'd like that.

axel-h:
> In the commit messages, I tried to convey that `register_t` is intended for _indexing_ into the register set, not for storing register contents,

OK, I see. To me using `register_t` is a bit counter-intuitive then because C has the keyword `register` and the usual cases I've seen so far in low level code use `register_t` usually in the content context. Maybe using `register_id_t` should be considered to clearly separate that.

proof/crefine/RISCV64/StateRelation_C.thy
```diff
@@ -207,7 +207,7 @@ definition
   \<and> tcb_queue_relation getNext getPrev hp queue NULL qhead"
 
 fun
-  register_from_H :: "register \<Rightarrow> machine_word"
+  register_from_H :: "register \<Rightarrow> 8 word"
```
lsf37, line 210: It'd be good to give the `8 word` a name here, to avoid magic numbers.

proof/crefine/RISCV64/SyscallArgs_C.thy
```diff
@@ -1058,10 +1053,31 @@ lemma index_msgRegisters_less':
 
 lemma index_msgRegisters_less:
   "n < 4 \<Longrightarrow> index msgRegistersC n <s 35"
-  "n < 4 \<Longrightarrow> index msgRegistersC n < 35"
   using index_msgRegisters_less'
   by (simp_all add: word_sless_msb_less)
 
+lemma less_ucast_ucast_less':
+  "x < UCAST('b \<rightarrow> 'a) y \<Longrightarrow> UCAST('a \<rightarrow> 'b) x < y"
```
lsf37, line 1060: These look more generally useful. Can you please move them up to `Word_Lemmas_Internal`, so I can merge them into the AFP `Word_Lib` over time?

proof/crefine/RISCV64/TcbAcc_C.thy
```diff
@@ -120,13 +120,12 @@ lemma register_from_H_less:
   by (cases hr, simp_all add: "StrictC'_register_defs")
 
 lemma register_from_H_sless:
-  "register_from_H hr <s 35"
+  "UCAST(8 \<rightarrow> 32 signed) (register_from_H hr) <s 0x23"
```
lsf37, line 123: To avoid ate least one magic number in this one, I'd just use `ucast` without length annotation, but give the number on the right a type (32 signed from what it looks like). From that Isabelle, should be able to infer the concrete `ucast` type directly.

proof/crefine/RISCV64/TcbAcc_C.thy
```diff
@@ -120,13 +120,12 @@ lemma register_from_H_less:
...
   by (cases hr, simp_all add: "StrictC'_register_defs" word_sless_def word_sle_def)
 
-lemma register_from_H_0_sle[simp]:
-  "0 <=s register_from_H hr"
-  using word_0_sle_from_less[OF order_less_le_trans] register_from_H_less
-  by fastforce
+lemma register_from_H_0_sle'[simp]:
+  "0 <=s UCAST(8 \<rightarrow> 32 signed) (register_from_H hr)"
```
lsf37, line 127: Same as above, I'd give `0` the type, and only write lower-case `ucast` on the right.

lsf37 (approved):
Looks good, just small cleanup comments.

proof/crefine/RISCV64/TcbAcc_C.thy
mbrcknl: I think `UCAST(_ -> 32 signed)` would also avoid one magic number, but I can't tell which style I prefer.

But perhaps even better would be to make some type synonyms:
```
type_synonym register_idx_len = 8
type_synonym register_idx = "register_idx_len word"
type_synonum int_literal_len = "32 signed"
```
Then you can write `UCAST(register_idx_len -> int_lteral_len)`, and also `register_idx` in the type of `register_from_H`.

mbrcknl (approved):
Looks good!

proof/crefine/RISCV64/TcbAcc_C.thy
lsf37: That'd be even better, yes

mbrcknl:
> OK, I see. To me using `register_t` is a bit counter-intuitive then because C has the keyword `register` and the usual cases I've seen so far in low level code use `register_t` usually in the content context. Maybe using `register_id_t` should be considered to clearly separate that.

I agree, but note that fixing the name of `register_t` would be a somewhat larger change, since we'd want to change the name across all architectures for consistency. Here, we're only changing the typedef on RISC-V, and removing some more recent uses of `register_t` on RISC-V that were inconsistent with previous usage (even though they were more consistent with your very reasonable interpretation of the name `register_t`).

@axel-h Would you be willing to allow us to proceed with this smaller change for now? We could consider changing the name separately. Note that aside from the few cases mentioned in RISC-V kernel, none of the uses of `register_t` in the kernel are for register *contents*, so we're already inconsistent with the obvious interpretation.

proof/crefine/RISCV64/TcbAcc_C.thy
mbrcknl: I guess somewhere near the top of `Wellformed_C.thy` would be a reasonable place for these type synonyms?

axel-h:
@mbrcknl thanks for the info, I agree that renaming is out of scope of this PR.

proof/crefine/RISCV64/StateRelation_C.thy
MitchellBuckley: How about
```
type_synonym register_index = "8 word"
```
?

proof/crefine/RISCV64/SyscallArgs_C.thy
MitchellBuckley: The one below it is even nicer. `ucast_up_less_bounded_iff_less_ucast_down'` essentially says that upcast and downcast form a Galois connection.

MitchellBuckley:
Testboard is running here: https://bamboo.ts.data61.csiro.au/browse/L4V-GHTESTBOARD7

proof/crefine/Move_C.thy
```diff
@@ -1347,4 +1347,26 @@ lemma tcbSchedEnqueue_queued_queues_inv:
...
+
+lemma ucast_up_less_bounded_implies_less_ucast_down':
+  assumes len: "LENGTH('a::len) < LENGTH('b::len)"
+  assumes bound: "y < 2 ^ LENGTH('a)"
+  assumes less: "UCAST('a \<rightarrow> 'b) x < y"
+  shows "x < UCAST('b \<rightarrow> 'a) y"
+  apply (rule le_less_trans[OF _ ucast_mono[OF less bound]])
+  using len by (simp add: is_down ucast_down_ucast_id)
+
+lemma ucast_up_less_bounded_iff_less_ucast_down':
```
zaklogician, line 1363: Should these be kept here, or moved to `lib/Word_Lib/Word_Lemmas_Internal.thy` as Gerwin suggested?

zaklogician (changes requested):
Looks jolly good. Just one question to address about the final resting place of the word lemmata.

zaklogician (approved):

proof/crefine/Move_C.thy
mbrcknl: I moved these to `Word_Lemmas_Internal` while rebasing. Fresh testboard below.

mbrcknl:
Rebased: [L4V-GHTESTBOARD15-1](https://bamboo.ts.data61.csiro.au/browse/L4V-GHTESTBOARD15-1)

PR #255, opened 2021-03-23
VER-1391

michaelmcinerney:
This addresses the points listed in JIRA issue VER-1391.

PR #261, opened 2021-03-30
rt refine: sendIPC_corres

MitchellBuckley:
This proves `sendIPC_corres` and attempts to do so with minimal other impact on proofs, aside from some strengthening of lemmas for `replyPush` and `schedContextDonate`.

### DIFF
--- a/proof/refine/ARM/IpcCancel_R.thy
+++ b/proof/refine/ARM/IpcCancel_R.thy
@@ -188,17 +188,18 @@ lemma replyTCB_update_corres:
       apply (simp add: reply_relation_def)
   by (wpsimp simp: obj_at'_def replyPrev_same_def)+
 
-lemma reply_unlink_tcb_corres:
-  "\<lbrakk>st = Structures_A.BlockedOnReceive ep (Some rp) pl
-    \<or> st = Structures_A.BlockedOnReply rp\<rbrakk> \<Longrightarrow>
-   corres dc
+lemma replyUnlinkTcb_corres:
+  "corres dc
      (valid_tcbs and pspace_aligned and pspace_distinct
-       and st_tcb_at ((=) st) t and reply_tcb_reply_at ((=) (Some t)) rp)
-          (valid_tcbs' and valid_release_queue_iff)
-        (reply_unlink_tcb t rp) (replyUnlink rp t)" (is "_ \<Longrightarrow> corres _ _ ?conc_guard _ _")
+       and st_tcb_at (\<lambda>st. \<exists>ep pl. st = Structures_A.BlockedOnReceive ep (Some rp) pl
+                           \<or> st = Structures_A.BlockedOnReply rp) t
+       and reply_tcb_reply_at ((=) (Some t)) rp)
+        (valid_tcbs' and valid_release_queue_iff)
+        (reply_unlink_tcb t rp) (replyUnlink rp t)" (is "corres _ _ ?conc_guard _ _")
   apply (rule_tac Q="?conc_guard
-              and st_tcb_at' (\<lambda>st. st = BlockedOnReceive ep (receiver_can_grant pl) (Some rp)
-                                 \<or> st = BlockedOnReply (Some rp)) t" in corres_cross_over_guard)
+         and st_tcb_at' (\<lambda>st. (\<exists>ep pl. st = BlockedOnReceive ep (receiver_can_grant pl) (Some rp))
+                               \<or> st = BlockedOnReply (Some rp)) t"
+         in corres_cross_over_guard)
    apply clarsimp
    apply (drule (1) st_tcb_at_coerce_concrete; clarsimp simp: state_relation_def)
    apply (fastforce simp: pred_tcb_at'_def obj_at'_def)
@@ -333,10 +334,8 @@ lemma blocked_cancel_ipc_corres:
                                  and st_tcb_at' ((=) st') t"
                       in corres_split_deprecated [OF _ set_ep_corres])
             apply (rule corres_guard_imp)
-              apply (rule corres_split_deprecated [OF _ reply_unlink_tcb_corres])
-                 apply (rule sts_corres)
-                 apply simp
-                apply fastforce
+              apply (rule corres_split_deprecated [OF _ replyUnlinkTcb_corres])
+                apply (rule sts_corres, simp)
                apply wpsimp
               apply (wpsimp wp: replyUnlink_valid_objs')
              apply (fastforce simp: pred_tcb_at_def obj_at_def is_tcb)
@@ -391,17 +390,17 @@ lemma blocked_cancel_ipc_corres:
       apply (frule (1) Receive_or_Send_ep_at'[rotated], blast)
       apply (fastforce simp: valid_ep'_def)
      apply (wpsimp wp: getEndpoint_wp hoare_vcg_conj_lift get_simple_ko_wp)+
-   apply (frule (2) Receive_or_Send_ep_at)
-   apply (clarsimp simp: obj_at_def is_ep pred_tcb_at_def)
+   apply (frule (2) Receive_or_Send_ep_at, clarsimp)
    apply (rule conjI, clarsimp)
-    apply (drule st_tcb_recv_reply_state_refs)
-     apply (fastforce simp: pred_tcb_at_def obj_at_def)
+    apply (drule (1) st_tcb_recv_reply_state_refs)
     apply (clarsimp simp: sk_obj_at_pred_def obj_at_def)
    apply (rule conjI)
-    apply (erule (1) valid_objsE[where x=epPtr], clarsimp simp: valid_obj_def)
-   apply (erule disjE)
-    apply (fastforce dest!: sym_ref_BlockedOnReceive_RecvEP)
-   apply (fastforce dest!: sym_ref_BlockedOnSend_SendEP)
+    apply (clarsimp simp: obj_at_def)
+    apply (erule (1) valid_objsE[where x=epPtr])
+    apply (clarsimp simp: valid_obj_def)
+   apply (erule disjE; clarsimp simp: obj_at_def pred_tcb_at_def)
+    apply (frule (2) sym_ref_BlockedOnReceive_RecvEP[OF _ _ sym], simp)
+   apply (frule (2) sym_ref_BlockedOnSend_SendEP[OF _ _ sym], simp)
   apply clarsimp
   apply (rule context_conjI)
    apply (erule (1) Receive_or_Send_ep_at'[rotated])
@@ -689,15 +688,13 @@ lemma no_fail_sc_wtih_reply_None_helper:
   done
 
 lemma replyRemoveTCB_corres:
-  "\<lbrakk> st = Structures_A.thread_state.BlockedOnReply rp;
-     thread_state_relation st st' \<rbrakk> \<Longrightarrow>
-   corres dc (valid_objs and pspace_aligned and pspace_distinct and valid_replies
-              and st_tcb_at ((=) st) t and (\<lambda>s. sym_refs (state_refs_of s)))
+  "corres dc (valid_objs and pspace_aligned and pspace_distinct and valid_replies
+              and st_tcb_at ((=) (Structures_A.thread_state.BlockedOnReply rp)) t and (\<lambda>s. sym_refs (state_refs_of s)))
              (valid_objs' and valid_release_queue_iff and (\<lambda>s'. sym_refs (list_refs_of_replies' s')))
                 (reply_remove_tcb t rp) (replyRemoveTCB t)"
-  (is "\<lbrakk> _ ; _ \<rbrakk> \<Longrightarrow> corres _ ?abs_guard ?conc_guard _ _")
+  (is "corres _ ?abs_guard ?conc_guard _ _")
   apply add_sym_refs
-  apply (rule_tac Q="st_tcb_at' ((=) st') t" in corres_cross_add_guard)
+  apply (rule_tac Q="st_tcb_at' ((=) (thread_state.BlockedOnReply (Some rp))) t" in corres_cross_add_guard)
    apply (fastforce dest!: st_tcb_at_coerce_concrete elim!: pred_tcb'_weakenE)
   apply (clarsimp simp: reply_remove_tcb_def replyRemoveTCB_def isReply_def)
   apply (rule corres_guard_imp)
@@ -730,10 +727,10 @@ lemma replyRemoveTCB_corres:
     apply (simp only: bind_assoc[symmetric])
     apply (rule corres_symb_exec_r_sr)
        apply (rule corres_guard_imp)
-         apply (rule reply_unlink_tcb_corres[simplified dc_def])
-         apply (rule disjI2, simp)
-        apply (clarsimp dest!: valid_objs_valid_tcbs st_tcb_reply_state_refs
-                         simp: obj_at_def is_reply reply_tcb_reply_at_def)
+         apply (rule replyUnlinkTcb_corres[simplified dc_def])
+        apply (clarsimp dest!: valid_objs_valid_tcbs)
+        apply (frule (1) st_tcb_reply_state_refs)
+        apply (clarsimp simp: pred_tcb_at_def obj_at_def is_tcb is_reply reply_tcb_reply_at_def)
        apply simp
       apply (erule sr_inv_sc_with_reply_None_helper)
      apply (wpsimp wp: updateReply_valid_objs' simp: valid_reply'_def obj_at'_def)
@@ -815,8 +812,7 @@ lemma replyRemoveTCB_corres:
                          apply (rule corres_symb_exec_r_sr)
                             apply (rule corres_guard_imp)
                              apply (rule corres_split[OF cleanReply_sc_with_reply_None_corres])
-                                apply (rule reply_unlink_tcb_corres[simplified dc_def])
-                                apply (rule disjI2, simp)
+                                apply (rule replyUnlinkTcb_corres[simplified dc_def])
                                apply wpsimp
                               apply wpsimp
                              apply simp
@@ -849,11 +845,14 @@ lemma replyRemoveTCB_corres:
                       apply (metis list.sel(1) list.sel(3) list.set_cases)
                      apply (clarsimp simp: getHeadScPtr_def reply_sc_reply_at_def obj_at_def is_reply
                                     split: reply_next.splits)
-                     apply (clarsimp dest!: valid_objs_valid_tcbs st_tcb_reply_state_refs
+                     apply (frule (1) st_tcb_reply_state_refs)
+                     apply (clarsimp dest!: valid_objs_valid_tcbs
                                       simp: obj_at_def is_reply reply_tcb_reply_at_def)
                      apply (clarsimp simp: opt_map_left_Some opt_map_def split: option.splits)
                      apply (rule context_conjI; clarsimp simp: vs_heap_simps obj_at_def)
-                     apply (metis list.sel(1) list.set_cases)
+                     apply (intro conjI)
+                      apply (metis list.sel(1) list.set_cases)
+                     apply (clarsimp simp: pred_tcb_at_def obj_at_def)
                     apply clarsimp
                     apply (rule conjI)
                      apply (clarsimp dest!: sc_ko_at_valid_objs_valid_sc'
@@ -894,12 +893,12 @@ lemma replyRemoveTCB_corres:
                            apply (rule corres_symb_exec_r_sr)
                               apply (rule corres_guard_imp)
                                 apply (rule corres_split[OF cleanReply_sc_with_reply_None_corres])
-                                  apply (rule reply_unlink_tcb_corres[simplified dc_def])
-                                  apply (rule disjI2, simp)
+                                  apply (rule replyUnlinkTcb_corres[simplified dc_def])
                                  apply wpsimp
                                 apply wpsimp
-                               apply (fastforce dest: valid_objs_valid_tcbs st_tcb_reply_state_refs
-                                                simp: obj_at_def is_reply reply_tcb_reply_at_def)
+                               apply clarsimp
+                               apply (frule (1) st_tcb_reply_state_refs, frule valid_objs_valid_tcbs)
+                               apply (fastforce simp: obj_at_def is_reply reply_tcb_reply_at_def pred_tcb_at_def)
                               apply simp
                              apply (clarsimp cong: conj_cong)
                              apply (case_tac "replyPrev reply'"; simp)
@@ -1030,7 +1029,6 @@ lemma cancel_ipc_corres:
                       apply (rule replyRemoveTCB_corres)
                        apply simp
                       apply (clarsimp simp: thread_state_relation_def)
-                      apply simp
                      apply (clarsimp simp: invs_implies)
                     apply (clarsimp simp: invs'_implies)
                    apply (rule corres_guard_imp)
@@ -2297,12 +2295,11 @@ lemma cancel_all_ipc_corres_helper:
                 apply (corressimp corres: restart_thread_if_no_fault_corres)
                 apply (fastforce simp: obj_at_def)
                apply (rule corres_guard_imp)
-                 apply (rule corres_split_deprecated[OF _ reply_unlink_tcb_corres])
+                 apply (rule corres_split_deprecated[OF _ replyUnlinkTcb_corres])
                     apply (rule corres_guard_imp)
                       apply (rule restart_thread_if_no_fault_corres)
                      apply simp
                     apply simp
-                   apply fastforce
                   apply (wpsimp wp: reply_unlink_tcb_valid_sched_action)
                  apply wpsimp
                 apply (fastforce simp: vs_all_heap_simps pred_tcb_at_def obj_at_def
@@ -2538,8 +2535,8 @@ proof -
         apply (rule not_idle_tcb_in_RecvEp; (fastforce simp: obj_at_def)?)
        apply (clarsimp simp: vs_all_heap_simps reply_unlink_ts_pred_def)
        apply (subst identity_eq)
-       apply (fastforce dest!: st_tcb_recv_reply_state_refs invs_sym_refs
-                         simp: pred_tcb_at_def obj_at_def reply_at_ppred_def)
+       apply (erule st_tcb_recv_reply_state_refs[OF _ invs_sym_refs, rotated])
+       apply (fastforce simp: pred_tcb_at_def obj_at_def reply_at_ppred_def)
       apply (clarsimp simp: valid_ep_def)
      apply (clarsimp simp: ep_relation_def)
     apply (wpsimp simp: get_ep_queue_def)
--- a/proof/refine/ARM/Ipc_R.thy
+++ b/proof/refine/ARM/Ipc_R.thy
@@ -1693,8 +1693,15 @@ crunches doIPCTransfer
   and sc_at'_n[wp]: "\<lambda>s. P (sc_at'_n n p s)"
   and irq_node'[wp]: "\<lambda>s. P (irq_node' s)"
   and valid_arch_state'[wp]: "valid_arch_state'"
+  and vrq[wp]: valid_release_queue
   (wp: crunch_wps
-   simp: zipWithM_x_mapM ball_conj_distrib )
+   simp: zipWithM_x_mapM ball_conj_distrib)
+
+crunches doIPCTransfer
+  for vrq'[wp]: valid_release_queue'
+  (wp: crunch_wps threadSet_vrq'_inv
+   simp: zipWithM_x_mapM ball_conj_distrib
+   ignore: threadSet)
 
 end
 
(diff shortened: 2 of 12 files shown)


corlewis (approved):
Looks great to me! All of my comments are pretty minor.

proof/refine/ARM/Ipc_R.thy
```diff
@@ -2304,20 +2296,478 @@ crunches doIPCTransfer, possibleSwitchTo
...
+lemma no_replySC_valid_replies'_sc_asrt:
+  "replySCs_of s r = None \<Longrightarrow> valid_replies'_sc_asrt r s"
+  unfolding valid_replies'_sc_asrt_def
+  by (simp)
+
+crunches doIPCTransfer
+  for replySCs_of[wp]: "\<lambda>s. P (replySCs_of s)"
+  (wp: crunch_wps simp: crunch_simps)
+
+(* FIXME RT: replace get_ep_valid_ep with this better version *)
```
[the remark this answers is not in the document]
MitchellBuckley: It probably lives in AInvs, so it would also involve changes to AInvs. By default I try to restrict these PRs to lib+Refine only, but don't need to. Do you think it's worth updating AInvs while I'm here?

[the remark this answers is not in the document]
corlewis: Maybe just to `Move_R` then? If nothing else, that makes it accessible to the RISCV64 proofs as well.

proof/refine/ARM/Ipc_R.thy
```diff
@@ -2304,20 +2296,478 @@ crunches doIPCTransfer, possibleSwitchTo
...
-  sorry
+proof -
+  show ?thesis
+    apply (insert assms)
+    apply add_sym_refs
+    apply (unfold send_ipc_def sendIPC_def Let_def)
+    apply (rule corres_stateAssert_assume)
+     apply (clarsimp split del: if_split)
+     apply (rule corres_guard_imp)
+       apply (rule corres_split_deprecated [OF _ get_ep_corres, where
```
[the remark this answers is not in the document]
MitchellBuckley: Ah, yes, I reused lots of the old proof, but I'll correct it as you've suggested.

proof/refine/ARM/IpcCancel_R.thy
```diff
@@ -689,15 +688,13 @@ lemma no_fail_sc_wtih_reply_None_helper:
...
-              and st_tcb_at ((=) st) t and (\<lambda>s. sym_refs (state_refs_of s)))
+  "corres dc (valid_objs and pspace_aligned and pspace_distinct and valid_replies
+              and st_tcb_at ((=) (Structures_A.thread_state.BlockedOnReply rp)) t and (\<lambda>s. sym_refs (state_refs_of s)))
              (valid_objs' and valid_release_queue_iff and (\<lambda>s'. sym_refs (list_refs_of_replies' s')))
                 (reply_remove_tcb t rp) (replyRemoveTCB t)"
-  (is "\<lbrakk> _ ; _ \<rbrakk> \<Longrightarrow> corres _ ?abs_guard ?conc_guard _ _")
+  (is "corres _ ?abs_guard ?conc_guard _ _")
   apply add_sym_refs
-  apply (rule_tac Q="st_tcb_at' ((=) st') t" in corres_cross_add_guard)
+  apply (rule_tac Q="st_tcb_at' ((=) (Structures_H.thread_state.BlockedOnReply (Some rp))) t" in corres_cross_add_guard)
```
michaelmcinerney, line 697: I don't think we need the `Structures_H.thread_state` here

proof/refine/ARM/IpcCancel_R.thy
```diff
@@ -731,9 +728,10 @@ lemma replyRemoveTCB_corres:
     apply (rule corres_symb_exec_r_sr)
        apply (rule corres_guard_imp)
          apply (rule reply_unlink_tcb_corres[simplified dc_def])
-         apply (rule disjI2, simp)
-        apply (clarsimp dest!: valid_objs_valid_tcbs st_tcb_reply_state_refs
-                         simp: obj_at_def is_reply reply_tcb_reply_at_def)
+        apply (clarsimp dest!: valid_objs_valid_tcbs)
+        apply (intro conjI)
```
michaelmcinerney, line 732: Can you solve this and the next two lines with just one command? It might be nicer than using `intro conjI` and then relying on the order being as it is now

proof/refine/ARM/IpcCancel_R.thy
```diff
@@ -849,11 +846,14 @@ lemma replyRemoveTCB_corres:
...
                      apply (clarsimp simp: getHeadScPtr_def reply_sc_reply_at_def obj_at_def is_reply
                                     split: reply_next.splits)
-                     apply (clarsimp dest!: valid_objs_valid_tcbs st_tcb_reply_state_refs
+                     apply (frule (1) st_tcb_reply_state_refs)
+                     apply (clarsimp dest!: valid_objs_valid_tcbs
                                       simp: obj_at_def is_reply reply_tcb_reply_at_def)
                      apply (clarsimp simp: opt_map_left_Some opt_map_def split: option.splits)
                      apply (rule context_conjI; clarsimp simp: vs_heap_simps obj_at_def)
+                     apply (intro conjI)
                      apply (metis list.sel(1) list.set_cases)
```
michaelmcinerney, line 855: The indentation must be off if `intro conjI` succeeds on the previous line

proof/refine/ARM/Ipc_R.thy
```diff
@@ -2053,23 +2060,36 @@ lemma bind_sc_reply_invs[wp]:
...
   "can_donate = can_donate' \<Longrightarrow>
-   corres dc (invs and tcb_at caller and tcb_at callee and reply_at reply_ptr
-              and ex_nonz_cap_to reply_ptr
-              and st_tcb_at active caller
-              and reply_sc_reply_at (\<lambda>scp. scp = None) reply_ptr
-              and reply_tcb_reply_at (\<lambda>tptr. tptr = None) reply_ptr
-              and weak_valid_sched_action and scheduler_act_not caller)
-             (valid_release_queue_iff and valid_objs' and valid_queues and valid_queues')
+   corres dc (valid_replies and pspace_aligned and pspace_distinct and valid_objs and
+              (\<lambda>_. caller \<noteq> idle_thread_ptr) and tcb_at caller and tcb_at callee and
```
michaelmcinerney, line 2080: Since we're discussing style guide stuff, is there a preference for having connectives at the end of the line versus the beginning of the next line? I prefer having them at the beginning of the line so I don't have to hunt back to the end of the previous line to double check what the connective is

proof/refine/ARM/Ipc_R.thy
```diff
@@ -2053,23 +2060,36 @@ lemma bind_sc_reply_invs[wp]:
...
+    valid_replies'_sc_asrt reply_ptr)
    (reply_push caller callee reply_ptr can_donate)
    (replyPush caller callee reply_ptr can_donate')"
   unfolding reply_push_def replyPush_def
   apply clarsimp
-  apply add_sym_refs
-  apply (rule_tac Q="\<lambda>s. valid_replies'_sc_asrt reply_ptr s" in corres_cross_add_guard)
-   apply (fastforce elim: valid_replies_sc_cross)
-  apply (rule corres_stateAssert_implied[where P'=\<top>, simplified])
+   apply (rule corres_stateAssert_implied[where P'=\<top>, simplified])
```
michaelmcinerney, line 2092: I think there is one space too many at the beginning of this line

proof/refine/ARM/Ipc_R.thy
michaelmcinerney, line 2080: `K ( caller ~= idle_thread_ptr)` instead?

proof/refine/ARM/Ipc_R.thy
```diff
@@ -2082,60 +2102,30 @@ lemma replyPush_corres:
...
+           apply (wpsimp wp: thread_get_wp, assumption)
+          apply (wpsimp wp: threadGet_wp)
+         apply (clarsimp simp: tcb_relation_def)
+        apply (wpsimp wp: thread_get_wp')
       apply (wpsimp wp: threadGet_wp)
      apply (subgoal_tac "caller \<noteq> reply_ptr")
       apply (subgoal_tac "caller \<noteq> idle_thread_ptr")
-       apply (clarsimp simp: invs_def valid_state_def
-                             valid_pspace_def cong: conj_cong)
+       apply (clarsimp simp: cong: conj_cong)
```
michaelmcinerney, line 2121: Unnecessary `simp:`

proof/refine/ARM/Ipc_R.thy
```diff
@@ -2082,60 +2102,30 @@ lemma replyPush_corres:
...
-         apply assumption
-        apply (clarsimp simp: pred_tcb_at_def obj_at_def)
-       apply (erule (1) if_live_then_nonz_capD)
-       apply (clarsimp simp: live_def  is_tcb obj_at_def)
-       apply (frule (1) bound_sc_tcb_at_idle_sc_idle_thread[where t=caller])
-        apply (clarsimp simp: pred_tcb_at_def obj_at_def)
-       apply simp
-      apply clarsimp
-      apply (frule invs_valid_idle)
+           apply (subst sc_at_ppred_exm; clarsimp)
```
michaelmcinerney, line 2126: I would have thought this would solve at most one subgoal, so maybe indentation is off

proof/refine/ARM/Ipc_R.thy
```diff
@@ -2304,20 +2296,478 @@ crunches doIPCTransfer, possibleSwitchTo
...
+  by (wpsimp simp: tcbSchedEnqueue_def)
+
+lemma replyPush_valid_objs'[wp]:
+  "replyPush callerPtr calleePtr replyPtr canDonate \<lbrace>valid_objs'\<rbrace>"
+  supply if_split [split del]
+  unfolding replyPush_def updateReply_def bind_assoc
+  apply (wpsimp wp: schedContextDonate_valid_objs' updateReply_valid_objs'
+                    hoare_vcg_if_lift2 threadGet_wp hoare_vcg_imp_lift')
+  apply (clarsimp simp: obj_at'_def projectKOs)
+  apply (safe; (fastforce simp: obj_at'_def projectKOs valid_tcb_state'_def)?)
```
michaelmcinerney, line 2437: `intro conjI impI` or something instead of `safe`?

proof/refine/ARM/Ipc_R.thy
```diff
@@ -2304,20 +2296,478 @@ crunches doIPCTransfer, possibleSwitchTo
...
+
+definition receive_reply :: "Structures_A.thread_state \<Rightarrow> obj_ref option" where
+  "receive_reply st \<equiv> case st of
+           (Structures_A.thread_state.BlockedOnReceive ep reply_opt pl) \<Rightarrow> reply_opt"
+
+lemma receive_reply_sel:
+  "receive_reply (Structures_A.thread_state.BlockedOnReceive ep reply_opt pl) = reply_opt"
+  by (simp add: receive_reply_def)
+
+lemma BlockedOnReceive_reply_tcb_reply_at:
```
michaelmcinerney, line 2511: Use `[| ...; ...|] ==> _` format

proof/refine/ARM/Ipc_R.thy
```diff
@@ -2304,20 +2296,478 @@ crunches doIPCTransfer, possibleSwitchTo
...
+  "\<exists>epptr pl. st_tcb_at ((=) (Structures_A.BlockedOnReceive epptr (Some rptr) pl)) tptr s
+   \<Longrightarrow> sym_refs (state_refs_of s) \<Longrightarrow> valid_objs s
+   \<Longrightarrow> reply_tcb_reply_at (\<lambda>x. x = Some tptr) rptr s"
+  apply (subgoal_tac "reply_at rptr s")
+  apply (clarsimp simp: pred_tcb_at_eq_commute)
```
michaelmcinerney, line 2516: Indentation

proof/refine/ARM/Ipc_R.thy
```diff
@@ -2304,20 +2296,478 @@ crunches doIPCTransfer, possibleSwitchTo
...
+           (Structures_A.thread_state.BlockedOnReceive ep reply_opt pl) \<Rightarrow> reply_opt"
+
+lemma receive_reply_sel:
+  "receive_reply (Structures_A.thread_state.BlockedOnReceive ep reply_opt pl) = reply_opt"
+  by (simp add: receive_reply_def)
+
+lemma BlockedOnReceive_reply_tcb_reply_at:
+  "\<exists>epptr pl. st_tcb_at ((=) (Structures_A.BlockedOnReceive epptr (Some rptr) pl)) tptr s
+   \<Longrightarrow> sym_refs (state_refs_of s) \<Longrightarrow> valid_objs s
+   \<Longrightarrow> reply_tcb_reply_at (\<lambda>x. x = Some tptr) rptr s"
```
michaelmcinerney, line 2514: Could you use a more descriptive bound variable than `x`?

proof/refine/ARM/Ipc_R.thy
```diff
@@ -2304,20 +2296,478 @@ crunches doIPCTransfer, possibleSwitchTo
...
-                                      and valid_sched_except_blocked_except_released_ipc_qs
-                                      and st_tcb_at active t and ep_at ep
-                                      and ex_nonz_cap_to t and scheduler_act_not t
-                                      and (\<lambda>s. cd \<longrightarrow> bound_sc_tcb_at (\<lambda>a. \<exists>y. a = Some y) t s))
+              and valid_sched_action and ep_at ep and ex_nonz_cap_to t and tcb_at t
+              and scheduler_act_not t and (\<lambda>s. cd \<longrightarrow> bound_sc_tcb_at (\<lambda>a. \<exists>y. a = Some y) t s))
              (invs' and sch_act_not t and tcb_at' t and ep_at' ep)
              (send_ipc bl call bg cg cgr cd t ep) (sendIPC bl call bg cg cgr cd t ep)"
-  sorry
+proof -
```
michaelmcinerney, line 2535: Is there a reason to start this off as an Isar proof?

proof/refine/ARM/Ipc_R.thy
```diff
@@ -2304,20 +2296,478 @@ crunches doIPCTransfer, possibleSwitchTo
...
+                        valid_sched_action and
+                        valid_ep rv and obj_at (\<lambda>ob. ob = Endpoint rv) ep
+                        and ex_nonz_cap_to t and scheduler_act_not t
+                        and (\<lambda>s. cd \<longrightarrow> bound_sc_tcb_at (\<lambda>a. \<exists>y. a = Some y) t s)"
+                and
+                R'="\<lambda>rv'. invs' and tcb_at' t and sch_act_not t
+                                and ep_at' ep and valid_ep' rv'
+                                and ko_at' (rv' :: endpoint) ep
+                                and (\<lambda>s'. sym_refs (state_refs_of' s'))"])
+         apply (case_tac epa)
```
michaelmcinerney, line 2554: `rename_tac` this `epa` and `bl` in the next line?

proof/refine/ARM/Ipc_R.thy
```diff
@@ -2304,20 +2296,478 @@ crunches doIPCTransfer, possibleSwitchTo
...
+                            simp del: dc_simp split del: if_split cong: if_cong)
+                apply (rule corres_split_deprecated [OF _ dit_corres])
+                  apply (rule corres_split_deprecated[where r'=dc])
+                     apply (simp only: get_tcb_obj_ref_def)
+                     apply (rule corres_split_deprecated [OF _ threadget_corres[where r="(=)"]])
+                        apply (rule corres_split_deprecated[OF _ threadget_corres[where r=fault_rel_optionation]])
+                           apply (rule corres_split_deprecated[OF _ corres_if[where r=dc], where r=dc])
+                                apply (rule corres_split_deprecated [OF possibleSwitchTo_corres sts_corres])
+                                  apply clarsimp
+                                 prefer 4 \<comment> \<open>deferring Hoare triples\<close>
```
michaelmcinerney, line 2603: Is there a nicer way than using `prefer 4` here? Maybe even a `rotated n` would be slightly nicer

proof/refine/ARM/Ipc_R.thy
```diff
@@ -2304,20 +2296,478 @@ crunches doIPCTransfer, possibleSwitchTo
...
+                                  apply (clarsimp simp: tcb_relation_def)
+                                 \<comment> \<open>starting Hoare triples\<close>
+                                 apply (wpsimp wp: thread_get_wp')
+                                apply (wpsimp wp: threadGet_wp)
+                               apply (wpsimp wp: set_thread_state_valid_sched_action)
+                              apply wpsimp
+                             apply (clarsimp simp: fault_rel_optionation_def)
+                           apply (rule_tac Q="\<lambda>_. valid_objs and pspace_aligned and pspace_distinct
+                                    and tcb_at a and valid_sched_action"
+                                    in hoare_strengthen_post[rotated])
```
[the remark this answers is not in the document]
michaelmcinerney: And maybe `rename_tac` this `a` too

proof/refine/ARM/Ipc_R.thy
```diff
@@ -2304,20 +2296,478 @@ crunches doIPCTransfer, possibleSwitchTo
...
+                                              reply_at' (the data) s \<and>
+                                              replySCs_of s (the data) = None)"
+                                         in hoare_strengthen_post[rotated])
+                   apply (clarsimp cong: conj_cong)
+                   apply (frule valid_objs'_valid_tcbs')
+                   apply (drule obj_at_ko_at')+
+                   apply (blast dest: no_replySC_valid_replies'_sc_asrt)
+                  apply wpsimp
+                 apply (wpfix add: receive_reply_sel)
+                 apply (rule_tac Q="\<lambda>_. valid_objs and pspace_aligned and pspace_distinct and
```
michaelmcinerney, line 2679: Can any of these large blocks be made simpler by using pattern matching against the statement of the lemma?

proof/refine/ARM/Ipc_R.thy
```diff
@@ -2304,20 +2296,478 @@ crunches doIPCTransfer, possibleSwitchTo
...
+                          \<longrightarrow> st_tcb_at ((=) (Structures_A.thread_state.BlockedOnReceive ep data pl)) a s \<and>
+                          ex_nonz_cap_to (the data) s \<and>
+                          reply_tcb_reply_at ((=) (Some a)) (the data) s \<and>
+                          reply_sc_reply_at (\<lambda>a. a = None) (the data) s \<and>
+                          the data \<notin> fst ` replies_with_sc s)"
+                          in hoare_strengthen_post[rotated])
+               apply (clarsimp split: option.splits)
+                  apply (intro conjI)
+                    apply (erule valid_objs_valid_tcbs)
+                   apply (clarsimp simp: pred_tcb_at_def obj_at_def is_tcb, fastforce)
```
michaelmcinerney, line 2694: Indentation seems off

proof/refine/ARM/Ipc_R.thy
```diff
@@ -2304,20 +2296,478 @@ crunches doIPCTransfer, possibleSwitchTo
...
+               apply fastforce
+                  apply (clarsimp simp: sk_obj_at_pred_def obj_at_def is_reply)
+                 apply (wpsimp wp: hoare_vcg_imp_lift hoare_vcg_all_lift simp: iff_conv_conj_imp)
+                apply (wpfix add: Structures_H.thread_state.sel)
```
michaelmcinerney, line 2698: Is the `Structures_H` necessary?

proof/refine/ARM/Ipc_R.thy
```diff
@@ -2304,20 +2296,478 @@ crunches doIPCTransfer, possibleSwitchTo
...
+                apply (rule_tac Q="\<lambda>_. tcb_at' t and tcb_at' a and valid_release_queue and
+                                       valid_release_queue' and valid_queues and valid_objs' and
+                                       valid_queues' and (\<lambda>s. data \<noteq> None \<longrightarrow> reply_at' (the data) s \<and>
+                                       replySCs_of s (the data) = None)"
+                                       in hoare_strengthen_post[rotated])
+                 apply (fastforce split: option.splits)
+                apply (wpsimp wp: hoare_vcg_imp_lift)
+               apply (assumption)
```
michaelmcinerney, line 2706: Unnecessary brackets

proof/refine/ARM/Ipc_R.thy
```diff
@@ -2304,20 +2296,478 @@ crunches doIPCTransfer, possibleSwitchTo
...
+              apply (prop_tac "(tcb_at' t and tcb_at' a and valid_pspace' and cur_tcb' and
+                                ep_at' ep and valid_release_queue and valid_release_queue' and
+                                valid_queues and valid_objs' and valid_mdb' and valid_queues') s'",
+                     assumption)
```
michaelmcinerney, line 2710: Indentation

proof/refine/ARM/Ipc_R.thy
```diff
@@ -2304,20 +2296,478 @@ crunches doIPCTransfer, possibleSwitchTo
...
+    apply clarsimp
+    apply (case_tac data; clarsimp simp: receive_reply_def)
+    apply (subgoal_tac "reply_at' aa s'", simp)
+    apply (frule (1) replySCs_of_cross, simp)
+    apply (erule cross_relF[OF _ reply_at'_cross_rel])
+    apply (clarsimp simp: obj_at_def reply_sc_reply_at_def is_reply)
+               apply (wpsimp wp: gts_wp)
```
michaelmcinerney, line 2717: Indentation

proof/refine/ARM/Ipc_R.thy
```diff
@@ -2304,20 +2296,478 @@ crunches doIPCTransfer, possibleSwitchTo
...
+              apply (wpsimp wp: gts_wp')
+             apply (clarsimp simp: ep_relation_def split: list.splits)
+            apply (wpsimp wp: hoare_vcg_all_lift hoare_vcg_imp_lift' hoare_vcg_disj_lift)
+           apply (wpsimp wp: hoare_vcg_all_lift hoare_vcg_imp_lift' hoare_vcg_disj_lift)
+          \<comment> \<open>end of main Hoare triples\<close>
+          apply (subgoal_tac "tcb_at a s")
+          apply (subgoal_tac "a \<noteq> ep")
```
michaelmcinerney, line 2724: Indentation

proof/refine/ARM/Ipc_R.thy
```diff
@@ -2304,20 +2296,478 @@ crunches doIPCTransfer, possibleSwitchTo
...
+                apply (clarsimp simp: obj_at_def live_def)
+               apply (erule weak_valid_sched_action_scheduler_action_not)
+               apply (clarsimp simp: obj_at_def pred_tcb_at_def)
+              apply (clarsimp, erule (1) FalseI[OF idle_no_ex_cap], clarsimp simp: valid_idle_def)
+        apply (case_tac "receive_reply x"; simp)
+        apply (subgoal_tac "data = Some aa", simp)
+         apply (subgoal_tac "reply_tcb_reply_at ((=) (Some a)) aa s", simp)
+         apply (subgoal_tac "reply_sc_reply_at (\<lambda>a. a = None) aa s", simp)
+          apply (intro conjI)
+                 apply (clarsimp simp: sk_obj_at_pred_def obj_at_def)
```
michaelmcinerney, line 2744: I think some of thse subgoals could be `fastforce`d, which could make it a little easier to read? No big deal, though

michaelmcinerney (approved):
Nice nice nice

proof/refine/ARM/Ipc_R.thy
MitchellBuckley: I agree.

proof/refine/ARM/Ipc_R.thy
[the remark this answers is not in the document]
corlewis: It's not a big deal, especially when reusing proofs like that, so it's up to you whether to change it here or not. My comment was really a reminder for the future.

proof/refine/ARM/Ipc_R.thy
MitchellBuckley: Nope, only there because I was reusing old proofs to make this one. Will change it.

proof/refine/ARM/Ipc_R.thy
corlewis: This is one which might be hard to get complete consensus on. Personally, I agree with both of you with having connectives like `and` and `|` at the start but at the same time still prefer things like `,` and `;` at the end.

MitchellBuckley: One could say "logical connectives should be at the start of new lines with the exception of puncuation (e.g. `,` and `;`"

corlewis: That would be my style but I know that some others like all connectives to go at the start, including `,` and `;`. As is usually the case, we're not going to make everyone happy so at some point we'll just need to pick an approach and stick with it until we all adjust.

mktnk3 (approved):
Great to see another big corres proof done!

proof/refine/ARM/Ipc_R.thy
```diff
@@ -2304,20 +2297,454 @@ crunches doIPCTransfer, possibleSwitchTo
...
+  done
+
+crunches doIPCTransfer
+  for reply_projs[wp]: "\<lambda>s. P (replyNexts_of s) (replyPrevs_of s) (replyTCBs_of s) (replySCs_of s)"
+  (wp: crunch_wps simp: crunch_simps)
+
+lemma BlockedOnReceive_reply_tcb_reply_at:
+  "\<lbrakk>\<exists>epptr pl. st_tcb_at ((=) (Structures_A.BlockedOnReceive epptr (Some rptr) pl)) tptr s;
+    sym_refs (state_refs_of s); valid_objs s\<rbrakk>
+   \<Longrightarrow> reply_tcb_reply_at (\<lambda>tp_opt. tp_opt = Some tptr) rptr s"
```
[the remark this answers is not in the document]
mktnk3: > This looks exactly the same as `DetSchedInvs_AI.BlockedOnReceive_reply_tcb_reply_at` (including the proof)...

So, if it is actually the same, we don’t need this lemma here, I suppose? And probably update the one in AInvs so that it doesn’t use `valid_objs`?

proof/refine/Move_R.thy
```diff
@@ -457,4 +457,20 @@ lemma bind_dummy_ret_val:
...
+lemma tcb_obj_at_equiv:
+  "tcb_at g s \<Longrightarrow> obj_at (P s) g s = (\<forall>ko. ko_at ko g s \<and> is_tcb ko \<longrightarrow> P s ko)"
+  by (clarsimp simp: obj_at_def is_tcb)
+
+(* FIXME RT: replace get_ep_valid_ep with this better version *)
+lemma get_ep_valid_ep'[wp]:
+  "\<lbrace> valid_objs and ep_at ep \<rbrace>
+   get_endpoint ep
+   \<lbrace> valid_ep \<rbrace>"
+  by (wpsimp simp: ep_at_def2 valid_ep_def2 simp_del: valid_simple_obj_def wp_del: get_ep_valid_ep)
```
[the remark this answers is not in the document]
MitchellBuckley: Do you mind if I leave this as a cleanup task? I've been trying to keep my PRs restricted to refine because working across two images tends to slow down the process (I have to rebuild things all the time). But I will do the replacement if you are very sure you want it now.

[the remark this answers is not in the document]
MitchellBuckley: Ah, but I see there is other overlap with AInvs on other comments. Hmm.

proof/refine/ARM/Ipc_R.thy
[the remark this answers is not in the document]
MitchellBuckley: I would not be surpised if this lemma exists, but I can't find it. Can you give me a line number in file in a branch?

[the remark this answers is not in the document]
mktnk3: Hi Mitch. Sorry for the confusion, but this lemma has been actually removed in #255 . I was on an old branch. So, this was replaced by `st_tcb_recv_reply_state_refs`. If you look at #255 , it should be pretty straightforward to use `st_tcb_recv_reply_state_refs` instead of this one.

proof/refine/Move_R.thy
```diff
@@ -457,15 +457,9 @@ lemma bind_dummy_ret_val:
...
-   lookup_cap_and_slot thread p
-   -,\<lbrace>\<lambda>ft s. valid_fault (ExceptionTypes_A.CapFault p' rp ft)\<rbrace>"
-  using lookup_cap_and_slot_valid_fault
-  apply (clarsimp simp: validE_E_def validE_def valid_def
-                 split: sum.splits)
-  apply (drule invs_valid_objs, fastforce)
-  done
+(* An "excluded middle" lemma for sc_at_ppred *)
+lemma sc_at_ppred_exm:
+  "sc_at g s \<Longrightarrow> sc_at_ppred p P g s = (\<not> sc_at_ppred p (\<lambda>x. \<not> P x) g s)"
```
mktnk3, line 462: I am not sure how this lemma is actually used in this PR, but it might be worth noting that the `sc_at` assumption is not needed for the implication from LH to RH. So, if the rewrite is actually used for simplification in this direction, we could do without this assumption and just have an imp lemma.

We do have lemmas similar to this one, such as:
```
lemma not_obj_at'_strengthen:
  "obj_at' (Not ∘ P) p s ⟹ ¬ obj_at' P p s"

lemma not_pred_tcb_at'_strengthen:
  "pred_tcb_at' f (Not ∘ P) p s ⟹ ¬ pred_tcb_at' f P p s"
```
we could write one for `sc_at_pred` along this line, too, maybe?

mktnk3: but I am not too sure about aligning all the negation.

MitchellBuckley: I agree that this approach will work if we are using this rewrite in the hypotheses of a proof, for in that case the full strength of the equality is not needed. However, if we are using this rewrite in the goal of a proof, then we would instead need "RH implies LH" which is not generally true. The use cases in my proofs are both substitutions in the goal and so the assumption is needed.

I could write two lemmas:
* LH implies RH, no assumption
* RH implies LH, with assumption
but I'm not sure that would be any neater.

MitchellBuckley: If it's ok, I'd like to leave this as it is.

mktnk3: yes, I said the assumption is redundant for going LH to RH. This lemma as an equality as it is now is not very good in that the use of it for simplifying LH -> RH will require us to prove an unnecessary condition, won’t it?

I would think having LH -> RH as an implication lemma is good. I am not sure about the other direction. There might be a better formulation. But, if it has to be an equality, I’d think it should be in the form of A = (B & C) (or something like that) rather than B => (A = C).

MitchellBuckley: I can do this
```
sc_at_ppred p P g s = (obj_at (λko. ∃sc n. ko = SchedContext sc n) g s ∧ ¬ sc_at_ppred p (λx. ¬ P x) g s)
```
but `sc_at` has something funky about valid sizes which is annoying.

mktnk3: Ah, I see. I think it can be `bound (scs_of s g)` in this case, instead of `sc_at`... we need to do some cleaning around these. Can we also use, say, `scp` instead of `g`?

mktnk3: but your `obj_at` version is equally good, I think.

MitchellBuckley: That should be equally easy for my proofs, is that the kind of thing you had in mind?

MitchellBuckley: Cool, thanks.

PR #262, opened 2021-03-30
Michaelm reply pop lemmas

michaelmcinerney:
This is a rebased and slightly updated version of Victor's previous PR. 

I probably should have just added some new commits, rather than squashing, to make reviewing easier. Sorry. The only significant change is in the commit that says "replyPop lemmas", and there, the only significant change is the addition of the proof of `replyPop_iflive`, though there were some small changes due to the latest spec change to `replyPop`.

### DIFF
--- a/lib/Monad_WP/NonDetMonadVCG.thy
+++ b/lib/Monad_WP/NonDetMonadVCG.thy
@@ -1214,6 +1214,15 @@ lemma hoare_vcg_conj_lift:
 lemmas hoare_vcg_conj_lift_pre_fix
   = hoare_vcg_conj_lift[where P=R and P'=R for R, simplified]
 
+lemma hoare_vcg_conj_lift':
+  assumes x: "\<lbrace>P\<rbrace> f \<lbrace>Q\<rbrace>"
+  assumes y: "\<lbrace>P'\<rbrace> f \<lbrace>Q'\<rbrace>"
+  shows      "\<lbrace>P and P'\<rbrace> f \<lbrace>\<lambda>rv s. Q rv s \<and> Q' rv s\<rbrace>"
+  apply (subst bipred_conj_def[symmetric], rule hoare_post_conj)
+   apply (rule hoare_pre_imp [OF _ x], simp)
+  apply (rule hoare_pre_imp [OF _ y], simp)
+  done
+
 lemma hoare_vcg_conj_liftE1:
   "\<lbrakk> \<lbrace>P\<rbrace> f \<lbrace>Q\<rbrace>,-; \<lbrace>P'\<rbrace> f \<lbrace>Q'\<rbrace>,\<lbrace>E\<rbrace> \<rbrakk> \<Longrightarrow>
   \<lbrace>P and P'\<rbrace> f \<lbrace>\<lambda>r s. Q r s \<and> Q' r s\<rbrace>,\<lbrace>E\<rbrace>"
--- a/proof/refine/ARM/Finalise_R.thy
+++ b/proof/refine/ARM/Finalise_R.thy
@@ -1436,7 +1436,7 @@ lemma emptySlot_invs'[wp]:
             \<and> (\<forall>sl'. info \<noteq> NullCap \<longrightarrow> sl' \<noteq> sl \<longrightarrow> cteCaps_of s sl' \<noteq> Some info)\<rbrace>
      emptySlot sl info
    \<lbrace>\<lambda>rv. invs'\<rbrace>"
-  apply (simp add: invs'_def valid_state'_def valid_pspace'_def)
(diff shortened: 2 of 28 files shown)


MitchellBuckley (approved):
Great!!

proof/refine/ARM/Finalise_R.thy
```diff
@@ -2326,6 +2326,342 @@ lemma replyClear_invs'[wp]:
   apply wpsimp
   sorry
 
+lemma isHeadSome:
+ "isHead (Some h) \<longleftrightarrow> (\<exists>p. h = Head p)"
+  by (clarsimp simp: isHead_def split: reply_next.split)
```
mktnk3, line 2331: This doesn’t seem to be used, and it is pretty much the same as existing `isHead_to_head`. So, probably we should remove this.

proof/refine/ARM/Finalise_R.thy
```diff
@@ -2326,6 +2326,342 @@ lemma replyClear_invs'[wp]:
...
+
+lemma in_list_refs_of_replies'_E:
+  assumes in_list: "(p, tp) \<in> list_refs_of_replies' s replyPtr"
+  assumes nextR: "\<And>reply. tp = ReplyNext \<Longrightarrow> replyNext reply = Some (Next p)
+                            \<Longrightarrow> replies_of' s replyPtr = Some reply \<Longrightarrow> R"
+  assumes prevR: "\<And>reply. tp = ReplyPrev \<Longrightarrow> replyPrev reply = Some p
+                            \<Longrightarrow> replies_of' s replyPtr = Some reply \<Longrightarrow> R"
+  shows "R"
```
mktnk3, line 2339: This doesn’t seem to be used, either. Do we want to keep it?

proof/refine/ARM/Finalise_R.thy
```diff
@@ -2326,6 +2326,342 @@ lemma replyClear_invs'[wp]:
...
+
+lemma updateReply_ko_at'_reply_sp:
+  "\<lbrace>ko_at' reply replyPtr\<rbrace>
+   updateReply replyPtr (replyTCB_update Map.empty)
+   \<lbrace>\<lambda>_. ko_at' (reply\<lparr>replyTCB := None\<rparr>) replyPtr\<rbrace>"
+  apply (simp add: updateReply_def)
+  apply (wpsimp wp: set_reply'.set_ko_at')
+  using ko_at'_inj by blast
+
+lemma replyUnlink_ko_at'_reply_sp:
```
mktnk3, line 2353: Are these sp rules (this one and the above) used somewhere? If not, probably we don’t need to keep them.

mktnk3 (approved):
Nice. Thanks for finishing this off. Some cleanup suggestions from me, but otherwise it looks good to go!

proof/refine/ARM/Finalise_R.thy
michaelmcinerney: Removed

proof/refine/ARM/Finalise_R.thy
michaelmcinerney: Removed

proof/refine/ARM/Finalise_R.thy
michaelmcinerney: I think I'll add a comment saying "FIXME RT: unused?"

mktnk3: Ok. I guess we can revisit when we finish Refine.

lib/Monad_WP/NonDetMonadVCG.thy
```diff
@@ -1214,6 +1214,15 @@ lemma hoare_vcg_conj_lift:
 lemmas hoare_vcg_conj_lift_pre_fix
   = hoare_vcg_conj_lift[where P=R and P'=R for R, simplified]
 
+lemma hoare_vcg_conj_lift':
+  assumes x: "\<lbrace>P\<rbrace> f \<lbrace>Q\<rbrace>"
+  assumes y: "\<lbrace>P'\<rbrace> f \<lbrace>Q'\<rbrace>"
+  shows      "\<lbrace>P and P'\<rbrace> f \<lbrace>\<lambda>rv s. Q rv s \<and> Q' rv s\<rbrace>"
```
corlewis, line 1220: I don't think this lemma is used at the moment, and I think that that adding an `and` like this is fighting against the simplifier. Is anyone aware of a particular use case for it?

michaelmcinerney: I will remove this as part of my forthcoming PR for doReplyTransfer_invs’, which contains some other work cleaning up some of the lemmas in this PR

mktnk3: Thanks for catching this. I was thinking about it and then forgot. I am more in favour of removing this, to be honest.

PR #263, opened 2021-04-01
doReplyTransfer_invs'

michaelmcinerney:


spec/haskell/src/SEL4/Object/Reply.lhs
```diff
@@ -93,6 +93,8 @@ This module specifies the behavior of reply objects.
 
 > replyRemove :: PPtr Reply -> PPtr TCB -> Kernel ()
 > replyRemove replyPtr tcbPtr = do
+>     stateAssert sym_refs_asrt
+>         "replyPush: `sym_refs (state_refs_of' s)` must hold"
```
michaelmcinerney, line 97: Replace `replyPush` with `replyRemove` in this error message

proof/refine/ARM/Syscall_R.thy
```diff
@@ -740,98 +740,69 @@ lemma isReply_awaiting_reply':
...
+   apply (intro conjI impI; fastforce?)
+    apply (fastforce dest: invs_valid_idle'
+                     simp: isReply_def valid_idle'_def obj_at'_def projectKOs idle_tcb'_def)
+   apply (fastforce intro: if_live_then_nonz_capE'
+                     simp: ko_wp_at'_def obj_at'_def projectKOs isReply_def)
+  apply (rule hoare_seq_ext[OF _ threadGet_sp], rename_tac fault)
+  apply (case_tac fault; clarsimp)
+   apply (wpsimp wp: doIPCTransfer_invs setThreadState_Running_invs')
+  apply (rule_tac Q="?pre and tcb_at' replyTCBPtr and ex_nonz_cap_to' replyTCBPtr
+                         and (\<lambda>s. replyTCBPtr \<noteq> ksIdleThread s)"
```
michaelmcinerney, line 792: Indentation

proof/refine/ARM/Finalise_R.thy
```diff
@@ -2482,6 +2518,36 @@ lemma replyPop_list_refs_of_replies'[wp]:
...
+  \<comment>\<open> opt_mapE will sometimes destroy the @{term "(|>)"} inside @{term replyNexts_of}
+      and @{term replyPrevs_of}, but we're using those as our local normal form. \<close>
+  supply opt_mapE[rule del]
+  apply (intro conjI impI allI)
+       \<comment>\<open> Our 6 cases correspond to various cases of @{term replyNext} and @{term replyPrev}.
+           We use @{thm ks_reply_at'_repliesD} to turn those cases into facts about
+           @{term replyNexts_of} and @{term replyPrevs_of}. \<close>
+      apply (all \<open>normalise_obj_at'\<close>)
+     apply (all \<open>drule(1) ks_reply_at'_repliesD[OF ko_at'_replies_of',
+                                                   folded protected_sym_refs_def]
```
michaelmcinerney, line 2541: Indentation

proof/refine/ARM/Syscall_R.thy
```diff
@@ -740,98 +740,69 @@ lemma isReply_awaiting_reply':
...
+                                                                        (tptr + n) s)"
+                   in hoare_post_impErr)
+        apply (rule sfi_invs_plus')
+       apply (wpsimp wp: getTCB_wp
+                   simp: isValidTimeoutHandler_def)+
+  apply (clarsimp simp: cte_wp_at'_obj_at' tcb_cte_cases_def  projectKOs obj_at'_def)
+  done
+
+crunches isValidTimeoutHandler
+  for inv: P
```
MitchellBuckley, line 765: nitpick: will it be ok, and better, to put this into `wp`?

proof/refine/ARM/Finalise_R.thy
```diff
@@ -2575,72 +2641,115 @@ lemma replyPop_iflive:
...
-                                       \<and> tcb_at' tcbPtr s \<and> tcbPtr \<noteq> ksIdleThread s"
-                       in hoare_post_imp)
-               apply (clarsimp split: if_split simp: valid_reply'_def)
-              apply (wpsimp wp: gts_wp')+
-  apply (clarsimp simp: obj_at_ko_at' valid_reply'_def split: if_splits)
-  apply (clarsimp simp: isHead_to_head split: if_splits)
-  apply (intro conjI; clarsimp; normalise_obj_at')
+  apply (clarsimp simp: replyPop_def)
+  apply (rule hoare_seq_ext[OF _ stateAssert_sp])
+  apply (intro hoare_seq_ext[OF _ get_reply_sp']
```
MitchellBuckley, line 2682: nice

MitchellBuckley (approved):
Love it! Neat trick with `intro hoare_seq_ext...`

proof/refine/ARM/Syscall_R.thy
michaelmcinerney: Maybe I'm being paranoid, and it's fine to put it into the `wp` set, but my thinking was that this rule throws away information about whether it *was* a valid timeout handler, and I didn't want to do that

proof/refine/ARM/Finalise_R.thy
```diff
@@ -2339,7 +2339,7 @@ lemma in_list_refs_of_replies'_E:
                  split: option.splits)
   by (auto elim: nextR prevR simp: opt_map_def)
 
-crunches replyPop
+crunches replyRemove
```
mktnk3, line 2342: Good idea. Makes sense.

mktnk3 (approved):
Nice! A fair amount of work for one non-corres sorry.

proof/refine/ARM/Syscall_R.thy
corlewis: You're right to be paranoid, but the rule that ends up being generated has `λ_. P` as the postcondition, so it can only be applied when the return value is not being considered. In my view, this means that it actually is safe to be in the `wp` set. 

There is then a theoretical alternative rule that has `⦃P⦄` as the postcondition that would need to include that information.

corlewis (approved):
Great job in removing all the side conditions in existing lemmas!

michaelmcinerney:
(Not sure how to reply to the comment you made about `isValidTimeoutHandler_invs` directly, but)

that's a very good point. I'll make it `[wp]`!

PR #264, opened 2021-04-04
yet another TOOLPREFIX for riscv toolchain

lsf37:
This one is relevant for MacOs

Signed-off-by: Gerwin Klein <kleing@unsw.edu.au>

### DIFF
--- a/spec/cspec/c/kernel.mk
+++ b/spec/cspec/c/kernel.mk
@@ -36,7 +36,7 @@ ifndef TOOLPREFIX
     ifeq ($(findstring ARM, ${L4V_ARCH}),ARM)
       TRY_TOOLPREFIX := arm-none-eabi-
     else ifeq (${L4V_ARCH},RISCV64)
-      TRY_TOOLPREFIX := riscv64-unknown-linux-gnu- riscv64-linux-gnu-
+      TRY_TOOLPREFIX := riscv64-unknown-linux-gnu- riscv64-linux-gnu- riscv64-unknown-elf-
     endif
   endif
   ifdef TRY_TOOLPREFIX


mbrcknl (approved):

axel-h:
Just curious, what does "relevant for MacOs" mean actually and why would an seL4 build care here?

mbrcknl:
> Just curious, what does "relevant for MacOs" mean actually and why would an seL4 build care here?

The `Makefile` here is used to produce pre-processed source for consumption by the C parser, and also to generate a `kernel.elf` for binary verification. It doesn't do much other than call into the seL4 build system using one of the "verified" config files. I added `TRY_TOOLPREFIX` as a convenience when some of us working on RISC-V binary verification were using the Debian-provided packages for the RISC-V cross-compilation toolchain, while the verification regression was still using a handcrafted toolchain with a different `TOOLPREFIX`. `TRY_TOOLPREFIX` meant that we didn't have to think about which `TOOLPREFIX` we needed. It looks like @lsf37 now wants the same convenience for running the RISC-V proofs on a Mac, where the RISC-V cross-compilation toolchain presumably has yet another `TOOLPREFIX`.

lsf37:
> It looks like @lsf37 now wants the same convenience for running the RISC-V proofs on a Mac, where the RISC-V cross-compilation toolchain presumably has yet another `TOOLPREFIX`.

Yes, that's all there is to it, really.

I'm on a Apple M1 chip for this one, which means yet another set of prefixes and different cross compilers. I'm not really trying to build all of seL4 with it, just to get to a correct `kernel_all.c_pp` file to produce a verification build. ARM and RISCV64 are working, X64 is still a question mark (the native clang and gcc on M1 now of course produce `arm64`, not `x86_64`, so I somewhat perversely need to cross-compile `x86_64` from an Arm gcc. Maybe tomorrow..).

PR #265, opened 2021-04-06
Initial style guide

MitchellBuckley:
I've drafted up a short style guide that is intended as a starting point for a more complete and rigorous guide which can be developed over time. I hope that the rules/examples in this version are correct.

I suggest:
1. Review and corrections of this limited guide up to a point where it is readable and accurate.
2. Merge.
3. Improvements, more details, and new sections can be proposed through pull-requests and go through the usual review process. 

In this way we can democratise the process somewhat, and also distribute the work-load across all interested parties.

### DIFF
--- a/docs/Style.thy
+++ b/docs/Style.thy
@@ -1,0 +1,235 @@
+(*
+     Copyright 2021, Data61, CSIRO (ABN 41 687 119 230)
+
+     SPDX-License-Identifier: CC-BY-SA-4.0
+*)
+
+chapter \<open>An Isabelle Syntax Style Guide\<close>
+
+theory Style
+  imports Main
+begin
+
+typedecl type_foo
+typedecl type_bar
+
+definition valid ::
+  "'a \<Rightarrow> 'b \<Rightarrow> 'c \<Rightarrow> bool" ("\<lbrace>_\<rbrace>/ _ /\<lbrace>_\<rbrace>")
+where
+  "\<lbrace>P\<rbrace> f \<lbrace>Q\<rbrace> \<equiv> undefined"
+
+definition
+  pred_conj :: "('a \<Rightarrow> bool) \<Rightarrow> ('a \<Rightarrow> bool) \<Rightarrow> ('a \<Rightarrow> bool)" (infixl "and" 35)
+where
+  "pred_conj P Q \<equiv> \<lambda>x. P x \<and> Q x"
+
+text \<open>
+This page describes the Isabelle syntax style conventions that we follow. For more on semantic
+style see Gerwin's style guide (parts 1 and 2).
+
+  * https://proofcraft.org/blog/isabelle-style.html
+  * https://proofcraft.org/blog/isabelle-style-part2.html
+
+Our Isabelle style is pretty much the same as in the Isabelle distribution. Look at files in 
+l4.verified/isabelle/src/HOL to get a feel for it.
+\<close>
+
+section \<open>General Principles\<close>
+
+text \<open>
+The main rules:
+  * Indent your proofs.
+  * Follow naming conventions.
+  * Be aware that others have to read what you write.
+
+To not drive proof maintainers insane:
+  * Do not use auto except at the end of a proof.
+  * Never check in proofs containing back.
+  * Instantiate bound variables in preference to leaving schematics in a subgoal 
+    (i.e. use erule_tac ... allE in preference to erule allE)
+  * Don't mix object and meta logic in a lemma statement.
+\<close>
+
+section \<open>Indentation\<close>
+
+text \<open>
+Isabelle code is much easier to maintain when indented consistently. In apply style proofs we
+indent by 2 spaces, and add an additional space for every additional subgoal. For example, the 
+rules iffI and conjI add a new subgoal, and fast removes a subgoal. The idea is that, when 
+something breaks, the indentation tells you whether a tactic used to solve a subgoal or produced 
+new ones.
+\<close>
+
+lemma disj_conj_distribR:
+  "(A \<and> B \<or> C) = ((A \<or> C) \<and> (B \<or> C))"
+  apply (rule iffI)
+   apply (rule conjI)
+    apply fast
+   apply fast
+  apply auto
+  done
+
+text \<open>
+The statement of Hoare triple lemmas are generally indented in the following way.
+\<close>
+
+lemma my_hoare_triple_lemma:
+  "\<lbrace> precondition_one and precondition_two and
+     precondition three \<rbrace>
+   my_function param_a param_b
+   \<lbrace> post_condition \<rbrace>"
+  oops
+
+text \<open> or \<close>
+
+lemma my_hoare_triple_lemma:
+  "\<lbrace> precondition_one and precondition_two
+     and precondition three \<rbrace>
+   my_function param_a param_b
+   \<lbrace> post_condition \<rbrace>"
+  oops
+
+text \<open>
+Definitions, record and datatypes are generally indented as follows and use \<equiv> rather than =. Short 
+datatypes can be on a single line.
+\<close>
+
+definition word_bits :: nat where
+  "word_bits \<equiv> 32"
+
+definition long_defn_type ::
+  "'this => 'is => 'a => 'long => 'type => 'that => 'does =>
+   'not => 'fit => 'in => 'one => 'line"
+  where
+  "long_defn_type \<equiv> undefined"
+
+fun foo_fun :: "nat \<Rightarrow> nat" where
+  "foo_fun 0 = 0" |
+  "foo_fun (Suc n) = n"
+
+record cdl_cnode =
+  cdl_cnode_caps :: type_foo
+  cdl_cnode_size_bits :: type_bar
+
+datatype cdl_object =
+    Endpoint
+  | Tcb type_foo
+  | CNode cdl_cnode
+  | Untyped
+ 
+datatype cdl_arch = IA32 | ARM11
+
+text \<open>
+There are tools to automatically indent proofs in both jEdit and Emacs. In terms of rules:
+  * definition and fun should be on the same line as the name.
+  * Everything after the first line of a definition or fun statement should be indented.
+  * Type definitions (record, type_synonym, datatype) have the defined type on the same line, the
+    rest only if it fits.
+  * Avoid long hanging indentation in data types, case expressions, etc. That is:
+
+Wrong:
+
+@{term\<open>
+case xs_blah of [] \<Rightarrow> foo
+              | y#ys \<Rightarrow> baz
+\<close>}
+
+Right:
+
+@{term\<open>
+case xs_blah of
+    [] \<Rightarrow> foo
+  | y#ys \<Rightarrow> baz
+\<close>}
+\<close>
+
+text \<open> 
+  If a definition body causes an overflow of the 100 char line limit, use one
+  of the following two patterns.
+\<close>
+
+definition hd_opt :: "'a list \<Rightarrow> 'a option" where
+  "hd_opt \<equiv> case_list
+            None
+            (\<lambda>h t. Some h)"
+
+definition hd_opt4 :: "'a list \<Rightarrow> 'a option" where
+  "hd_opt4 \<equiv>
+   case_list None (\<lambda>h t. Some h)"
+
+text \<open>
+  If an apply line causes an overflow of the 100 char line limit, use one
+  of the following two patterns: aligning on ":" or aligning on left.
+\<close>
+
+lemma test_lemma0:
+  "\<lbrakk>hd_opt l \<noteq> None; hd_opt (hd l) = Some x\<rbrakk> \<Longrightarrow>
+    hd_opt (concat l) = Some x"
+  apply (clarsimp simp: hd_opt_def
+                 split: list.splits)
+  done
+
+lemma test_lemma1:
+  "\<lbrakk>hd_opt l \<noteq> None; hd_opt (hd l) = Some x\<rbrakk> \<Longrightarrow>
+    hd_opt (concat l) = Some x"
+  apply (clarsimp simp: hd_opt_def
+                  split: list.splits)
+  done
+
+lemma test_lemma3:
+  "case_option None hd_opt (hd_opt l) = Some x \<Longrightarrow>
+   hd_opt (concat l) = Some x"
+  apply ((case_tac l; simp add: hd_opt_def),
+         rename_tac h t,
+         case_tac h; simp)
+  done
+
+section \<open>Other\<close>
+
+text \<open>
+General layout:
+* Wrap at 100 (if possible), avoid overly long lines, not everyone has a big screen
+* No hard tabs, use spaces instead, standard indentation is 2
+* Avoid starting lines with quotes ("), it breaks some automated tools.
+* Avoid unnecessary parentheses, unless it becomes really unclear what is going on. Roughly, 
+  parentheses are unnecessary if Isabelle doesn't show them in output.
+
+Other:
+* Avoid commands that produce "legacy" warnings. Add a JIRA issue with tag cleanup if you see them 
+  after an Isabelle update.
+\<close>
+
+section \<open>Comments\<close>
+
+text \<open>
+(* .. *) style comments are not printed in latex. Use them if you don't think anyone will want to 
+see this printed. Otherwise, prefer text {* .. *}
+
+In stark difference to the Isabelle distribution, we do comment stuff. If you have defined or done 
+anything that is tricky, required you to think deeply, or similar, save everyone else the duplicated 
+work and document why things are the way you've written them.
+
+We've been slack on commenting in the past. We should do more of it rather than less.
+There's no need to comment every single thing, though. Use your judgement. Prefer readable 
+code/proofs/definitions over commented ones.\<close>
+
+
+section \<open>More on proofs\<close>
+
+subsection \<open>prefer and defer\<close>
+
+text \<open> 
+   There are currently no hard rules on the use of `prefer` and `defer n`. Two general principles
+   apply:
+
+   1. If they are used too frequently then a proof may become unreadable. Do not use them
+      unnecessarily and consider including comments to indicate their purpose.
+   2. If they are used too "specifically" then a proof may break very frequently for not-
+      interesting reasons. This makes a proof less maintainable and so this should be avoided.
+  
+  A known use case for `prefer` is where a proof has a tree-like structure, and the prover wants to
+  approach it with a breadth-first approach. Since the default isabelle strategy is depth-first,
+  prefers (or defers) will be needed, e.g. corres proofs.
+\<close>
+
+end
--- a/docs/Style.thy
+++ b/docs/Style.thy
@@ -29,12 +29,35 @@ style see Gerwin's style guide (parts 1 and 2).
 
   * https://proofcraft.org/blog/isabelle-style.html
   * https://proofcraft.org/blog/isabelle-style-part2.html
-
-Our Isabelle style is pretty much the same as in the Isabelle distribution. Look at files in 
-l4.verified/isabelle/src/HOL to get a feel for it.
 \<close>
 
 section \<open>General Principles\<close>
+
+text \<open>
+  Goals:
+  * Readability and consistency:
+      reduce cognitive overhead.
+
+  * Make good use of horizontal space:
+      most screens are wide, not high
+
+  * Semantic indentation:
+      where possible group things according to meaning. But remain syntactically consistent:
+      a sub-term should not go more left than its parent term.
+
+  * Support tools like diff and grep:
+      diff tries to find the "function" each chunk belongs to by indentation. If we remain
+      compatible, diffs are much easier to read. Pull requests mean we read a lot of diffs.
+      grep and other regexp search is not needed as much as it used to, but it's still often
+      useful to be able to find constants and lemmas by regexp search outside of Isabelle
+
+  * Don't fight existing auto-indent where possible:
+      there can be exceptions, but each time we go against defaults, it will be annoying to follow.
+
+  * Compatibility with upstream style where it exists:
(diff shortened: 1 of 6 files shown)


docs/Style.thy
```diff
@@ -0,0 +1,182 @@
...
+
+subsection \<open>by\<close>
+
+(* The use of `by my_last_method` should be restricted to those cases where the final method
+   application is either:
+
+   1. Relatively simple and unlikely to break in the future, or
+   2. a very long-running method.
+
+   TO DO: Elaborate on the reasons for this standard.
```
lsf37, line 179: I thought we had actually decided that it doesn't matter in the end. The main complaint was that it's hard to figure out the error message if something breaks, but that has changed over Isabelle versions, and most methods do present their error message with `by`.

The reason I'd prefer not to make this is a rule is that upstream Isabelle style is to *prefer* `by`.

docs/Style.thy
```diff
@@ -0,0 +1,182 @@
...
+   A known use case for `prefer` is where a proof has a tree-like structure, and the prover wants to
+   approach it with a breadth-first approach. Since the default isabelle strategy is depth-first,
+   prefers (or defers) will be needed, e.g. corres proofs.
+*)
+
+subsection \<open>auto\<close>
+
+(* The `auto` method must only be used at the end of a proof, as the last method in a proof.
+
+   TO DO: Elaborate on the reasons for this standard.
```
lsf37, line 168: Could point at the section in my style guide for that.

docs/Style.thy
```diff
@@ -0,0 +1,182 @@
...
+ * @TAG(DATA61_BSD)
+ *)
+
+chapter \<open>An Isabelle Syntax Style Guide\<close>
+
+theory Style
+  imports Main
+begin
+
+(* This style guide is heavily influenced by Gerwin's Style Guide for Isabelle/HOL. See parts 1
```
lsf37, line 17: We should make these `text` comments instead of actual comments. That way, they are rendered readable on GitHub and they do show up in html/pdf if we produce that.

docs/Style.thy
```diff
@@ -0,0 +1,182 @@
...
+   and 2 here:
+
+   * https://proofcraft.org/blog/isabelle-style.html
+   * https://proofcraft.org/blog/isabelle-style-part2.html
+
+   This guide may repeat some parts of Gerwin's guide and should not conflict in any way.
```
lsf37, line 23: There is also a difference between the style guides: ny style guide avoids almost all syntactic considerations and is mostly about semantics things. This one here is more focussed on the syntactic aspect. Ideally we'd just link to mine for semantic considerations (like "auto in the middle of a proof") or when we have a different rule.

docs/Style.thy
```diff
@@ -0,0 +1,182 @@
...
+ *)
+
+section \<open>General Principles\<close>
+
+(* TO DO: What advice should go here? *)
```
lsf37, line 28: I'll write something for discussion.

docs/Style.thy
```diff
@@ -0,0 +1,182 @@
...
+
+
+section \<open>Definitions\<close>
+
+(* Definitions should be written over 2 lines in the following style.
+
+   Note that indentation is 2 spaces and the definitions do not run past 100 character line
+   limit. Note also that the \equiv symbol is used.
```
lsf37, line 36: \equiv is indeed our default. I wonder if we should make it mandatory or leave it open, because upstream style is to prefer the opposite (`=`). I do like \equiv better, mostly for reasons of syntax precedence (needs fewer parentheses). Upstream uses `<->` to get around the precedence for boolean definitions, but it doesn't work for other types and I find it strange for a definition.

The reason some people prefer `=` is that it's easier to work with in the logic, e.g. you can combine it easily with `sym` or similar without converting to object equality first, but I have never found that a hindrance in practice. Certainly not so far that I would write more parentheses everywhere.

Happy to keep it at \equiv, just wanted to point out that there are other opinions.

docs/Style.thy
```diff
@@ -0,0 +1,182 @@
...
+ *)
+
+definition hd_opt :: "'a list \<Rightarrow> 'a option" where
+  "hd_opt \<equiv> case_list None (\<lambda>h t. Some h)"
```
lsf37, line 40: It'd be nicer to use an example that has a parameter on the lhs, just to give a more familiar pattern.

docs/Style.thy
```diff
@@ -0,0 +1,182 @@
...
+
+   Note that the indentation of the type signature remains no less than the indentation level of
+   the initial `"`.
+ *)
+
+type_synonym my_very_long_nat_type = nat
+type_synonym 'a my_very_long_option_type = "'a option"
+type_synonym 'a my_very_long_list_type = "'a list"
+definition hd_opt2 :: "'a my_very_long_list_type \<Rightarrow> my_very_long_nat_type \<Rightarrow>
+                       'a my_very_long_list_type \<Rightarrow> 'a my_very_long_option_type" where
```
lsf37, line 53: I would not indent this way, actually. I'd write

```
definition hd_opt2 ::
  "'a my_very_long_list_type \<Rightarrow> my_very_long_nat_type \<Rightarrow>
   'a my_very_long_list_type \<Rightarrow> 'a my_very_long_option_type"
  where
```

`where` could also go on the end of the last line, I don't mind that. The main point is to break after the `::` to avoid a potentially very long right overhang (length depends on the name of the constant).

docs/Style.thy
```diff
@@ -0,0 +1,182 @@
...
+   gives a nice visual indication of the nesting depth of your proof.
+
+   The usual indentation for 0 subgoals is 2 spaces for a top-level proof.
+
+  For example:
+*)
+
+lemma test_lemma1:
+  "2 \<le> length l \<Longrightarrow> \<exists>h1 h2 t. l = h1 # h2 # t"
+  apply (case_tac l)
```
lsf37, line 114: Should probably have `cases` in the example since we should prefer that where possible.

docs/Style.thy
```diff
@@ -0,0 +1,182 @@
...
+   TO DO: Develop more rigorous advice on this issue.
+
+   Some examples.
+*)
+
+lemma test_lemma2:
+  "case_option None hd_opt (hd_opt l) = Some x \<Longrightarrow>
+   hd_opt (concat l) = Some x"
+  apply (clarsimp simp: hd_opt_def
+                 split: list.splits)
```
lsf37, line 138: It's in the TODO, but just wanted to comment here that we should allow both alignment styles for method modifiers: align on `:` and align left.

docs/Style.thy
```diff
@@ -0,0 +1,182 @@
+(*
+ * Copyright 2021, Data61, CSIRO
+ *
+ * This software may be distributed and modified according to the terms of
+ * the BSD 2-Clause license. Note that NO WARRANTY is provided.
+ * See "LICENSE_BSD2.txt" for details.
+ *
+ * @TAG(DATA61_BSD)
```
lsf37, line 8: Needs an SPDX license header now and should be CC-BY-SA-4.0 (see other README files)

docs/Style.thy
MitchellBuckley: Will remove for now.

lsf37:
I've added general goals that the rules in the style guide should aim towards. Hopefully this will give us a good basis for discussion if there is disagreement.

docs/Style.thy
```diff
@@ -0,0 +1,260 @@
...
+
+definition long_defn_type ::
+  "'this => 'is => 'a => 'long => 'type => 'that => 'does =>
+   'not => 'fit => 'in => 'one => 'line"
+  where
+  "long_defn_type \<equiv> undefined"
+
+fun foo_fun :: "nat \<Rightarrow> nat" where
+  "foo_fun 0 = 0" |
+  "foo_fun (Suc n) = n"
```
corlewis, line 132: I know that this came from the previous document, but my strong preference is for
```suggestion
fun foo_fun :: "nat \<Rightarrow> nat" where
    "foo_fun 0 = 0"
  | "foo_fun (Suc n) = n"
```

docs/Style.thy
```diff
@@ -0,0 +1,260 @@
...
+datatype cdl_object =
+    Endpoint
+  | Tcb type_foo
+  | CNode cdl_cnode
+  | Untyped
+
+datatype cdl_arch = IA32 | ARM11
+
+text \<open>
+There are tools to automatically indent proofs in both jEdit and Emacs. In terms of rules:
```
corlewis, line 147: I think that we can remove the reference to Emacs now.

docs/Style.thy
```diff
@@ -0,0 +1,260 @@
...
+
+text \<open>
+  If a definition body causes an overflow of the 100 char line limit, use one
+  of the following two patterns.
+\<close>
+
+definition hd_opt :: "'a list \<Rightarrow> 'a option" where
+  "hd_opt \<equiv> case_list
+            None
+            (\<lambda>h t. Some h)"
```
corlewis, line 178: Being pedantic, I think that this should be
```suggestion
  "hd_opt \<equiv> case_list
              None
              (\<lambda>h t. Some h)"
```

docs/Style.thy
```diff
@@ -0,0 +1,260 @@
...
+  apply (clarsimp simp: hd_opt_def
+                  split: list.splits)
+  done
+
+lemma test_lemma3:
+  "case_option None hd_opt (hd_opt l) = Some x \<Longrightarrow>
+   hd_opt (concat l) = Some x"
+  apply ((case_tac l; simp add: hd_opt_def),
+         rename_tac h t,
+         case_tac h; simp)
```
corlewis, line 208: Do we want to require extra bracketing when `;` and `,` interact? I don't know about others, but I haven't managed to internalise this yet and it's not obvious to me whether these lines are correctly indented.

corlewis (commented):
This is great, thanks for pushing this forwards Mitch!

docs/Style.thy
```diff
@@ -0,0 +1,260 @@
...
+
+text \<open>
+The statement of Hoare triple lemmas are generally indented in the following way.
+\<close>
+
+lemma my_hoare_triple_lemma:
+  "\<lbrace> precondition_one and precondition_two and
+     precondition three \<rbrace>
+   my_function param_a param_b
+   \<lbrace> post_condition \<rbrace>"
```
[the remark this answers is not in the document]
mbrcknl: No real argument, but a mild preference for no spaces, if only because it's the way I'm accustomed to the syntax translations printing things back to me.

docs/Style.thy
lsf37: I used to write them like this and still have. a slight preference for it, but the other style seems more popular upstream. I'm easy on this one. If people prefer it, happy with that.

docs/Style.thy
lsf37: I'm actually not that sure either. It'd be better to figure that out and document it better than requiring more parentheses, though.

docs/Style.thy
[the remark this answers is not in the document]
lsf37: Also mild preference for no spaces.

docs/Style.thy
```diff
@@ -0,0 +1,260 @@
...
+
+definition valid ::
+  "'a \<Rightarrow> 'b \<Rightarrow> 'c \<Rightarrow> bool" ("\<lbrace>_\<rbrace>/ _ /\<lbrace>_\<rbrace>")
+where
+  "\<lbrace>P\<rbrace> f \<lbrace>Q\<rbrace> \<equiv> undefined"
+
+definition
+  pred_conj :: "('a \<Rightarrow> bool) \<Rightarrow> ('a \<Rightarrow> bool) \<Rightarrow> ('a \<Rightarrow> bool)" (infixl "and" 35)
+where
+  "pred_conj P Q \<equiv> \<lambda>x. P x \<and> Q x"
```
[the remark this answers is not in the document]
lsf37: I vote for a separate helper theory. It'll be nicer when you can start reading actual content from the top.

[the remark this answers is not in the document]
mbrcknl: Is there a reason to not just import `Lib.NonDetMonad`? I guess it would mean adding a `ROOT` file with a `Docs` session, but I guess we would want that anyway, to avoid proliferation of unchecked Isabelle theories in the repo.

docs/Style.thy
MitchellBuckley: These parentheses are required since `,` apparently binds stronger than `;`. How could I improve this example, or should I remove the example?

docs/Style.thy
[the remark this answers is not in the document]
MitchellBuckley: I added a helper theory because it's easier and faster, I hope that's ok for now.

MitchellBuckley:
Attempted to apply most/all of the advice provided in feedback. I also rebased over master. Any further feedback?

docs/Style.thy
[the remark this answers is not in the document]
lsf37: A helper theory is fine from my side, but I don't see it popping up in the diff. Am I looking at the wrong version?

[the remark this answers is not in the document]
lsf37: Oh, I do see it, but it's not used in `Style.thy` yet. Can you please add that? Basically we want the theory to start with the text explanation (or at least as soon as practical).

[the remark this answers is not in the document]
MitchellBuckley: Oops, should be fixed now.

lsf37 (approved):
I'm happy with this as a first step. We can always tweak more later.

docs/Style_pre.thy
```diff
@@ -0,0 +1,28 @@
...
+text \<open>
+This a helper theory for Style.thy.
+\<close>
+
+typedecl type_foo
+typedecl type_bar
+
+definition valid ::
+  "'a \<Rightarrow> 'b \<Rightarrow> 'c \<Rightarrow> bool" ("\<lbrace>_\<rbrace>/ _ /\<lbrace>_\<rbrace>")
+where
```
mbrcknl, line 20:
```suggestion
  where
```

docs/Style_pre.thy
```diff
@@ -0,0 +1,28 @@
...
+  "\<lbrace>P\<rbrace> f \<lbrace>Q\<rbrace> \<equiv> undefined"
+
+definition
+  pred_conj :: "('a \<Rightarrow> bool) \<Rightarrow> ('a \<Rightarrow> bool) \<Rightarrow> ('a \<Rightarrow> bool)" (infixl "and" 35)
+where
```
mbrcknl, line 25:
```suggestion
  where
```

mbrcknl (approved):
Looks good!

MitchellBuckley:
https://bamboo.ts.data61.csiro.au/browse/L4V-GHTESTBOARD55-1

mbrcknl (commented):
One more comment: this should probably be included in a session somewhere. Is there a docs session?

MitchellBuckley:
> One more comment: this should probably be included in a session somewhere. Is there a docs session?

There is no `ROOT` in `l4v/docs/`, and no `docs` in `l4v/ROOTS`. We can make it though.

docs/Style.thy
```diff
@@ -0,0 +1,263 @@
...
+   case_list None (\<lambda>h t. Some h)"
+
+text \<open>
+  If an apply line causes an overflow of the 100 char line limit, use one
+  of the following two patterns: aligning on ":" or aligning on left.
+\<close>
+
+lemma test_lemma0:
+  "\<lbrakk>hd_opt l \<noteq> None; hd_opt (hd l) = Some x\<rbrakk> \<Longrightarrow>
+    hd_opt (concat l) = Some x"
```
Xaphiosis, line 189: Small detail: in this example is hd_opt indented one too many here? (same in test_lemma1); it seems it's because of the `\<lbrakk>` but I'm not familiar of doing things this way.

mbrcknl:
> > One more comment: this should probably be included in a session somewhere. Is there a docs session?
> 
> There is no `ROOT` in `l4v/docs/`, and no `docs` in `l4v/ROOTS`. We can make it though.

I just added a rather quickly assembled thing that appears to add this to the regression, according to some very superficial testing. Might be overkill; what do people think?

docs/Style.thy
```diff
@@ -0,0 +1,263 @@
...
+    precondition three\<rbrace>
+   my_function param_a param_b
+   \<lbrace>post_condition\<rbrace>"
+  oops
+
+text \<open> or \<close>
+
+lemma my_hoare_triple_lemma:
+  "\<lbrace>precondition_one and precondition_two
+    and precondition three\<rbrace>
```
Xaphiosis, line 106: I think there should be (maybe not in this revision) the concept of wrapping operators in the "operator left" style and "operator right" style. Some people prefer to wrap operations with the operator dangling on the right, e.g.

```
a + b +
c

a + b *
    c
```

while others prefer to wrap the operator and stick it on the left:

```
a + b
+ c

a + b
    * c
```

Where this is particularly useful, and I've been doing it this way for a long while, is in lemma statements (excuse ASCII art):

```
lemma "[| bunch of assumptions;
          more assumptions |]
       ==> conclusion"
```

which allows to clearly see in a large lemma where the conclusion is. Your style guide doesn't disallow that, but it might be worth pointing out that we did not settle this particular argument.
@lsf37 @mbrcknl any further memories/thoughts to add to this, while we're looking at the PR?

docs/Style.thy
```diff
@@ -0,0 +1,263 @@
...
+
+section \<open>Comments\<close>
+
+text \<open>
+(* .. *) style comments are not printed in latex. Use them if you don't think anyone will want to
+see this printed. Otherwise, prefer text \<open> .. \<close>
+
+In stark difference to the Isabelle distribution, we do comment stuff. If you have defined or done
+anything that is tricky, required you to think deeply, or similar, save everyone else the duplicated
+work and document why things are the way you've written them.
```
Xaphiosis, line 232: Good. I might add something like "That way even if your tactic/definition breaks, we will have an idea of what you were trying to achieve."

Xaphiosis:
I think this is a very good start. The operator wrapping thing should go in it at some point (maybe after more discussion), but we can amend this document after we settle disputes in future.

I  have another comment:
* add principle: `Do not change style of lemmas/functions/definitions "by the way", i.e. near to where you are working but that you are not otherwise modifying; add such changes to principled cleanup commits instead.` We'd rather live with some legacy style than die by style-fixing confetti.

docs/Style.thy
mbrcknl: Perhaps we don't want to try to resolve the argument now. :-)

It would be good to acknowledge that there are two styles, each with advantages and current adherents, and while we currently allow them both, we would prefer people not to constantly reformat each other's code.

That said, in the interests of getting something merged, I'd also be happy to leave this to the next iteration.

If/when we do say something, I would definitely like to make clear that we want the indenting to follow the operator precedence, as in your examples, so:
```
a
+ b
+ c
  * d
  * e
+ f
```
rather than:
```
a
+ b
+ c
* d
* e
```

lsf37: Yes, I'm happy with that (both, leaving it for later, and to explicitly allow both styles)

Xaphiosis: Yeah, that's fine, but I don't want to settle the argument by accident, since I do wrap `==>` and `-->` that way, and I remember from past discussions I'm not alone on that side. So recognition of the stalemate should go in at some point.

Agreed on your point about constantly changing styles, will add to my general principle observation, I think something like that observation should go in early.

Agreed on precedence; it can even be simpler: indent if it's a different operator, but do not mislead audience about precedence, e.g. anything like this is evil:

```
a
* b
  + c
  + d
```

lsf37:
> I think this is a very good start. The operator wrapping thing should go in it at some point (maybe after more discussion), but we can amend this document after we settle disputes in future.
> 
> I have another comment (and one more after Matt's excellent point):
> 
> * add principle: `Do not change style of lemmas/functions/definitions "by the way", i.e. near to where you are working but that you are not otherwise modifying; add such changes to principled cleanup commits instead.` We'd rather live with some legacy style than die by style-fixing confetti.
> * additionally: `When making very small changes (one-liners, renames, etc.), blend in to the style you see used around you, even if it is not your preferred style, and even if it is legacy. Having a proof accumulate a dozen styles over the decades rather than keep a suboptimal one is itself suboptimal.`

I'd also be happy with these as general guideline, applied with some discretion, i.e. not necessarily always a hard-and-fast rule.

lsf37:
> > > One more comment: this should probably be included in a session somewhere. Is there a docs session?
> > 
> > 
> > There is no `ROOT` in `l4v/docs/`, and no `docs` in `l4v/ROOTS`. We can make it though.
> 
> I just added a rather quickly assembled thing that appears to add this to the regression, according to some very superficial testing. Might be overkill; what do people think?

I think it's great!

Xaphiosis:
> I'd also be happy with these as general guideline, applied with some discretion, i.e. not necessarily always a hard-and-fast rule.

Yes, I agree. To confirm, my goal is to avoid in future, when young, new and eager people try do something and we get "change one line in a proof from 2012, and then do that line in the approved new style, and notice that by the way, this other term on the screen in another lemma has one too many spaces, ..."

docs/Style.thy
MitchellBuckley: > > > > One more comment: this should probably be included in a session somewhere. Is there a docs session?
> > > 
> > > 
> > > There is no `ROOT` in `l4v/docs/`, and no `docs` in `l4v/ROOTS`. We can make it though.
> > 
> > 
> > I just added a rather quickly assembled thing that appears to add this to the regression, according to some very superficial testing. Might be overkill; what do people think?
> 
> I think it's great!

Looks good to me!

docs/Style.thy
MitchellBuckley: Ok, thanks. I'm pretty keen to merge this as starting point and then accumulate this kind of more detailed and more accurate advice through subsequent pull-requests.

MitchellBuckley:
https://bamboo.ts.data61.csiro.au/browse/L4V-GHTESTBOARD58

PR #271, opened 2021-04-15
Remove remaining tab characters in .thy files

lsf37:
Closes [VER-748]

[VER-748]: https://sel4.atlassian.net/browse/VER-748


corlewis (approved):
Thanks! Is that officially the last tabs, and do you know whether our tests will catch any new ones?

lsf37:
> Thanks! Is that officially the last tabs, and do you know whether our tests will catch any new ones?

Yes, these are the last Raf's command on the Jira issue finds and the whitespace test should catch new ones.

PR #273, opened 2021-04-20
Overrun

michaelmcinerney:
This adds handling of overrun to `refill_budget_check/refillBudgetCheck`.

As far as I understand, the most stringent real time guarantees are applicable only in systems which respect a worst case execution time for threads running in user space, as well as the kernel itself. However, we would like to be able to properly handle the case where the WCET of a thread in user space is violated - that is, we'd still like to prove functional correctness for a kernel which handles this. `handle_overrun_loop/handleOverrunLoop` has been added to `refill_budget_check/refillBudgetCheck` to achieve this.

Our current approach does not wish to statically establish an upper bound on the release time of a scheduling context on which `refill_budget_check` is called. Instead, we check throughout `handle_overrun_loop` to see whether an operation would result in overflow. 

Because of this, we can no longer work with the formulation of `bounded_release_time` that we had (which said that the release time of a scheduling context was within a certain distance of the current time). `bounded_release_time` was only being used to show that `refill_update` preserves `valid_refills` (in particular, performing `refill_update` on some scheduling context would not result in overflow). The check in `handle_overrun_loop` allows us to show another property which suffices, namely that there is no scheduling context with release time within `2 * MAX_PERIOD` of `max_time`. Since `bounded_release_time` was used in fairly few places before, and the name is still reasonable for this new property, I've decided to stick with it (but happy to change it if it's confusing)

### DIFF
--- a/proof/invariant-abstract/DetSchedInvs_AI.thy
+++ b/proof/invariant-abstract/DetSchedInvs_AI.thy
@@ -2291,17 +2291,21 @@ abbreviation valid_ready_qs :: "'z state \<Rightarrow> bool" where
 
 lemmas valid_ready_qs_def = valid_ready_qs_2_def valid_ready_queued_thread_2_def
 
-definition bounded_release_time_2 :: "time \<Rightarrow> time \<Rightarrow> bool" where
-  "bounded_release_time_2 curtime reltime \<equiv>
-     unat reltime \<le> unat curtime + unat kernelWCET_ticks + unat MAX_PERIOD"
+\<comment> \<open>This predicate is currently used only in refill_update_valid_refills, in particular,
+    to show that performing refill_update will not result in overflow.
+    The constant 2 is required because refill_update will schedule a refill at most MAX_PERIOD
+    from the head r_time, of r_amount at most MAX_PERIOD\<close>
+definition bounded_release_time_2 :: "time \<Rightarrow> bool" where
+  "bounded_release_time_2 reltime \<equiv>
+     unat reltime + 2 * unat MAX_PERIOD \<le> unat max_time"
 
-abbreviation cfg_bounded_release_time :: "time \<Rightarrow> sc_refill_cfg \<Rightarrow> bool" where
-  "cfg_bounded_release_time t cfg \<equiv>
-     bounded_release_time_2 t (r_time (hd (scrc_refills cfg)))"
+abbreviation cfg_bounded_release_time :: "sc_refill_cfg \<Rightarrow> bool" where
+  "cfg_bounded_release_time cfg \<equiv>
+     bounded_release_time_2 (r_time (hd (scrc_refills cfg)))"
 
 abbreviation bounded_release_time :: "obj_ref \<Rightarrow> 'z state \<Rightarrow> bool" where
   "bounded_release_time scp s \<equiv>
-     pred_map (cfg_bounded_release_time (cur_time s)) (sc_refill_cfgs_of s) scp"
+     pred_map (cfg_bounded_release_time) (sc_refill_cfgs_of s) scp"
 
 lemmas cfg_bounded_release_time_def = bounded_release_time_2_def
 lemmas bounded_release_time_def = bounded_release_time_2_def
@@ -2312,7 +2316,7 @@ definition active_sc_valid_refills_2 :: "time \<Rightarrow> (obj_ref \<rightharp
   "active_sc_valid_refills_2 curtime sc_refill_cfgs \<equiv>
    \<forall>scp. pred_map active_scrc sc_refill_cfgs scp
          \<longrightarrow> pred_map cfg_valid_refills sc_refill_cfgs scp
-             \<and> pred_map (cfg_bounded_release_time curtime) sc_refill_cfgs scp"
+             \<and> pred_map cfg_bounded_release_time sc_refill_cfgs scp"
 
 abbreviation active_sc_valid_refills :: "'z state \<Rightarrow> bool" where
   "active_sc_valid_refills s \<equiv> active_sc_valid_refills_2 (cur_time s) (sc_refill_cfgs_of s)"
@@ -3831,8 +3835,8 @@ abbreviation cur_sc_active where
 
 lemmas cur_sc_active_lift = hoare_lift_Pf[where f=cur_sc and P=is_active_sc and m=f for f, rotated]
 
-abbreviation sc_bounded_release_time :: "time \<Rightarrow> sched_context \<Rightarrow> bool" where
-  "sc_bounded_release_time ct sc \<equiv> cfg_bounded_release_time ct (sc_refill_cfg_of sc)"
+abbreviation sc_bounded_release_time :: "sched_context \<Rightarrow> bool" where
+  "sc_bounded_release_time sc \<equiv> cfg_bounded_release_time (sc_refill_cfg_of sc)"
 
 locale valid_sched_pred_locale =
   fixes state_ext_t :: "'state_ext::state_ext itself"
--- a/proof/invariant-abstract/DetSchedSchedule_AI.thy
+++ b/proof/invariant-abstract/DetSchedSchedule_AI.thy
@@ -40,14 +40,14 @@ lemma current_time_bounded_strengthen:
   unfolding current_time_bounded_def
   by (erule order_trans[rotated], clarsimp)
 
-lemma update_time_stamp_current_time_bounded_3:
-  "\<lbrace>\<top>\<rbrace> update_time_stamp \<lbrace>\<lambda>_. current_time_bounded 3\<rbrace>"
+lemma update_time_stamp_current_time_bounded_5:
+  "\<lbrace>\<top>\<rbrace> update_time_stamp \<lbrace>\<lambda>_. current_time_bounded 5\<rbrace>"
   supply minus_add_distrib[simp del]
   apply (clarsimp simp: update_time_stamp_def getCurrentTime_def)
   apply wpsimp
   apply (clarsimp simp: current_time_bounded_def max_word_def)
   apply (rule_tac y="unat (-(getCurrentTime_buffer + 1))
-                     + unat kernelWCET_ticks + 3 * unat MAX_PERIOD"
+                     + unat kernelWCET_ticks + 5 * unat MAX_PERIOD"
                in order_trans)
    apply (prop_tac "min (unat (-(getCurrentTime_buffer + 1)))
                         (unat (last_machine_time_of s) + time_oracle (Suc (time_state_of s)))
@@ -57,11 +57,11 @@ lemma update_time_stamp_current_time_bounded_3:
   apply (clarsimp simp: minus_add_distrib)
   apply (meson le_trans min.cobounded1 unat_of_nat_closure)
   apply (subst unat_minus_plus_one)
-   apply (insert cur_time_bound_no_overflow cur_time_bound_no_overflow')
+   apply (insert getCurrentTime_buffer_no_overflow getCurrentTime_buffer_no_overflow')
    apply (clarsimp simp: kernelWCET_ticks_def MAX_PERIOD_def max_word_def)
-  apply (clarsimp simp:  max_word_def)
-  apply (insert cur_time_bound_no_overflow'_stronger cur_time_bound_no_overflow
-                cur_time_bound_no_overflow' MAX_PERIOD_mult)
+  apply (clarsimp simp: max_word_def)
+  apply (insert getCurrentTime_buffer_no_overflow'_stronger getCurrentTime_buffer_no_overflow
+                getCurrentTime_buffer_no_overflow' MAX_PERIOD_mult)
   by (fastforce simp: kernelWCET_ticks_def MAX_PERIOD_def max_word_def)
 
 lemma as_user_cur_sc_chargeable[wp]:
@@ -2034,7 +2034,7 @@ lemma set_refills_active_sc_valid_refills:
   "\<lbrace>active_sc_valid_refills and
     (\<lambda>s. is_active_sc sc_ptr s \<longrightarrow>
          pred_map (\<lambda>sc. cfg_valid_refills (sc\<lparr>scrc_refills := refills\<rparr>)) (sc_refill_cfgs_of s) sc_ptr
-         \<and> pred_map (\<lambda>sc. cfg_bounded_release_time (cur_time s) (sc\<lparr>scrc_refills := refills\<rparr>))
+         \<and> pred_map (\<lambda>sc. cfg_bounded_release_time (sc\<lparr>scrc_refills := refills\<rparr>))
                           (sc_refill_cfgs_of s) sc_ptr)\<rbrace>
    set_refills sc_ptr refills
    \<lbrace>\<lambda>rv. active_sc_valid_refills\<rbrace>"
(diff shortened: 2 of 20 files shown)


proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -8876,57 +8871,563 @@ lemma refill_budget_sched_schedule_used_r_time_helper:
...
+    apply (metis Nitpick.size_list_simp(2) tail_nonempty_length)
+   apply (metis Nitpick.size_list_simp(2) hd_conv_nth length_greater_0_conv nth_tl)
+  apply (rule word_le_nat_alt[THEN iffD1])
+  apply (rule schedule_used_head_r_time_shift_forward)
+   apply (metis Nitpick.size_list_simp(2))
+  apply (case_tac "sc_refills sc"; rename_tac list, case_tac list; simp)
+  using no_overflow_Cons apply blast
+  done
+
+(* FIXME RT Michael: move *)
```
michaelmcinerney, line 9076: I will move this to where `refills_unat_sum` is defined

proof/invariant-abstract/DetSchedInvs_AI.thy
```diff
@@ -2291,17 +2291,17 @@ abbreviation valid_ready_qs :: "'z state \<Rightarrow> bool" where
 
 lemmas valid_ready_qs_def = valid_ready_qs_2_def valid_ready_queued_thread_2_def
 
-definition bounded_release_time_2 :: "time \<Rightarrow> time \<Rightarrow> bool" where
-  "bounded_release_time_2 curtime reltime \<equiv>
-     unat reltime \<le> unat curtime + unat kernelWCET_ticks + unat MAX_PERIOD"
+definition bounded_release_time_2 :: "time \<Rightarrow> bool" where
+  "bounded_release_time_2 reltime \<equiv>
+     unat reltime + 2 * unat MAX_PERIOD \<le> unat max_time"
```
MitchellBuckley, line 2296: If it's not too much work, can we hide this `2` under a definition or (input) abbreviation?

proof/refine/ARM/Ipc_R.thy
```diff
@@ -2661,7 +2661,7 @@ lemma cteInsert_ct'[wp]:
 lemma maybeDonateSc_corres:
   "corres dc (tcb_at tcb_ptr and ntfn_at ntfn_ptr and invs and weak_valid_sched_action
               and valid_ready_qs and active_sc_valid_refills and valid_release_q
-              and current_time_bounded 1 and ex_nonz_cap_to tcb_ptr)
+              and current_time_bounded 2 and ex_nonz_cap_to tcb_ptr)
```
MitchellBuckley, line 2664: Given that it was I who introduced that parameter to `current_time_bounded`, I have only myself to blame. But is it possible to replace this `2` with a non magic-number or remove the paramater altogether? Does the new strategy for dealing with current_time allow for the latter?

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -8876,57 +8871,554 @@ lemma refill_budget_sched_schedule_used_r_time_helper:
...
                           sc_valid_refills_def word_le_nat_alt)
    apply (subst unat_sub)
-    apply (clarsimp simp: word_le_nat_alt)
-   apply (metis (no_types) add.commute add_increasing2 diff_le_self order_trans
-                sc_valid_refills_amounts_bounded sc_valid_refills_hd_in_list unat_arith_simps(1)
-                zero_le)
+    apply force
+   apply (clarsimp simp: word_less_nat_alt unat_sub)
+   apply (frule_tac refill="refill_hd sc" in refill_budget_check_refill_amount_helper
+          ; fastforce?)
```
MitchellBuckley, line 8877: nitpick: put semicolon on previous line?

MitchellBuckley (approved):
This is really great! I love the way loops and loop bodies are handled, and lots of the changes are fairly minimal. A few comments:
* nitpick on the location of semi-colons, but I know that style is not settled so I don't mind if you leave them your way.
* It would be really nice to remove the constant 5 from all of the current_time_bounded lines, if that's possible now. And then I imagine all the cases where it is 1 or 3 etc would still be fine. But I know this was already a lot of work, and I might be missing something that makes that impossible/infeasible. 
* It would be nice to abbreviate or make definition for 2 in release_time_bounded, but again I know that might be a lot of work on top of what you've done.
* The details refills work in DetSchedSchedule looks like it was difficult. The proofs there have always looked difficult. Now is not the time, but I hope one day we can streamline those a bit more. Congrats on getting them to work! Again!

proof/invariant-abstract/DetSchedInvs_AI.thy
michaelmcinerney: This `2` doesn't appear anywhere else, so I don't think we'd gain anything from having this as a definition or abbreviation. Happy to make a definition or abbreviation if others would like it, though

proof/refine/ARM/Ipc_R.thy
michaelmcinerney: We could certainly get rid of the number in `current_time_bounded`. We'd currently like `current_time_bounded 5` true after we call `getCurrentTime`. So we could define `current_time_bounded_const` as `5` (and even further abstract `getCurrentTime_buffer` to be `kernelWCET_ticks + current_time_bounded_const * MAX_PERIOD` if we'd like). The downside would be that many lemas would then be made weaker, because they require smaller constants like `1` or `2`. 

Any preference? This might be better put as a JIRA issue for now, and we can decide what'd we'd like later on when things are more settled

proof/invariant-abstract/DetSchedSchedule_AI.thy
michaelmcinerney: Over my dead body

spec/haskell/src/SEL4/Object/SchedContext.lhs
```diff
@@ -345,15 +345,46 @@ This module uses the C preprocessor to select a target architecture.
...
+> handleOverrunLoop :: Ticks -> Kernel Ticks
+> handleOverrunLoop usage =
+>     whileLoop (\usage -> fromJust . runReaderT (headTimeBuffer usage)) (\usage -> handleOverrunLoopBody usage) usage
+
 > refillBudgetCheck :: Ticks -> Kernel ()
 > refillBudgetCheck usage = do
 >     scPtr <- getCurSc
+
 >     sc <- getSchedContext scPtr
 >     assert (scRefillMax sc > 0) "CurSc should be active"
```
corlewis, line 380: Unrelated to this PR and probably not necessary, but it might be slightly more consistent to use `scActive` here (and maybe readerfy it).
```suggestion
>     active <- scActive scPtr
>     assert active "CurSc should be active"
```

spec/abstract/SchedContext_A.thy
```diff
@@ -264,28 +264,69 @@ where
...
+
+     when (usage' > 0 \<and> usage' < new_head_amount) $ do
+       refills \<leftarrow> get_refills sc_ptr;
+       sc \<leftarrow> get_sched_context sc_ptr;
+       period \<leftarrow> return (sc_period sc);
+       used \<leftarrow> return \<lparr>r_time = r_time (hd refills) + period, r_amount = usage'\<rparr>;
+       adjusted_hd \<leftarrow> return \<lparr>r_time = r_time (hd refills) + usage',
+                              r_amount = r_amount (hd refills) - usage'\<rparr>;
+       full \<leftarrow> refill_full sc_ptr;
+       set_refills sc_ptr $ schedule_used full (adjusted_hd # (tl refills)) used
```
corlewis, line 326: I guess it was ok since you've finished refine already, but was it an issue that the Haskell and C updates the refills in two steps while the abstract does it in one?

spec/abstract/SchedContext_A.thy
```diff
@@ -264,28 +264,69 @@ where
...
+
+     robin \<leftarrow> is_round_robin sc_ptr;
+     assert (\<not>robin);
 
+     usage' \<leftarrow> handle_overrun_loop usage;
+
+     refills \<leftarrow> get_refills sc_ptr;
+     new_head_amount \<leftarrow> return (r_amount (hd refills));
+
+     when (usage' > 0 \<and> usage' < new_head_amount) $ do
```
corlewis, line 318: Same comment as with the Haskell, it's not immediately obvious that it's correct to have the `usage' < new_head_amount` check here.

corlewis (commented):
I've still got `DetSchedSchedule_AI` to go through, but so far all the proofs look great! I have a couple of mostly unrelated comments for the specs and one potential concern about the new check in `refill_budget_check`.

proof/invariant-abstract/DetSchedInvs_AI.thy
MitchellBuckley: Cool, could you add a short comment, just a little snippet so we know why it's 2 and not 1 or 3? (Presumably it's because we the overrun code could push an sc `MAX_PERIOD` into the future, and the span of the refills could fill out another `MAX_PERIOD`?)

proof/refine/ARM/Ipc_R.thy
MitchellBuckley: JIRA issue sounds sensible.

spec/haskell/src/SEL4/Object/SchedContext.lhs
```diff
@@ -345,15 +345,46 @@ This module uses the C preprocessor to select a target architecture.
...
+
+>     return usage'
+
+> handleOverrunLoop :: Ticks -> Kernel Ticks
+> handleOverrunLoop usage =
+>     whileLoop (\usage -> fromJust . runReaderT (headTimeBuffer usage)) (\usage -> handleOverrunLoopBody usage) usage
+
 > refillBudgetCheck :: Ticks -> Kernel ()
 > refillBudgetCheck usage = do
 >     scPtr <- getCurSc
```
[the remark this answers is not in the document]
michaelmcinerney: If we know that it's not round robin in the abstract or exec level, it's very straightforward to carry that down. Do we know that? I'm not sure what the setup for CRefine looks like. It's very easy to add to the Haskell, so I can certainly do that

spec/haskell/src/SEL4/Object/SchedContext.lhs
```diff
@@ -345,15 +345,46 @@ This module uses the C preprocessor to select a target architecture.
...
 >     assert (scRefillMax sc > 0) "CurSc should be active"
 
->     usage' <- return (min usage (rAmount (refillHd sc)))
+>     usage' <- handleOverrunLoop usage
+
+>     sc <- getSchedContext scPtr
+>     newHeadAmount <- return (rAmount (refillHd sc))
 
->     when (usage' > 0) $ do
+>     when (usage' > 0 && usage' < newHeadAmount) $ do
```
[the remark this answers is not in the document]
michaelmcinerney: I requested Curtis change things around so that we always had `MIN_BUDGET` in the head. The changes are here

https://github.com/kent-mcleod/seL4/pull/10/files

He is also happy with `5` in the check, but I'm not sure that's been updated in the C anywhere. (The changes to `refill_update` aren't necessary any more, and I've let Curtis know about that, too)

spec/abstract/SchedContext_A.thy
michaelmcinerney: I meant to come back and clean up some of the abstract spec related to `schedule_used`. I'd prefer it to be a monadic function. `schedule_used` is now called in `handle_overrun_loop`, too, so it seems worthwhile, as long as it doesn't mess up the proofs too badly. I think I'll leave this for "another day"

spec/haskell/src/SEL4/Object/SchedContext.lhs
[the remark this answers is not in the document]
corlewis: Well, having looked a bit further, you assert that it's not round robin in the abstract spec so you must have already proven it as part of refine. If that's true then we would want to add the assert to Haskell as well, so we can reuse it during crefine. Otherwise we would have to reprove it.

spec/haskell/src/SEL4/Object/SchedContext.lhs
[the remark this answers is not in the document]
corlewis: Ah, cool, that makes sense. I was looking at https://github.com/kent-mcleod/seL4/pull/7/files instead.

spec/haskell/src/SEL4/Object/SchedContext.lhs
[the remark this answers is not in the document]
michaelmcinerney: The `corres` rules for `refillBudgeCheck` and `refillUnblockCheck` aren't done yet. That'll be fun. But OK, I'll add the assert

spec/haskell/src/SEL4/Object/SchedContext.lhs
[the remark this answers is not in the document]
michaelmcinerney: Yeah sorry I should have done a better job listing the various C commits

proof/invariant-abstract/DetSchedInvs_AI.thy
michaelmcinerney: Howsabout:

"This predicate is currently used only in `refill_update_valid_refills`, in particular, to show that performing `refill_update` will not result in overflow. The constant `2` is required because `refill_update` will schedule a refill at most `MAX_PERIOD` from the head `r_time`, of `r_amount` at most `MAX_PERIOD`"

MitchellBuckley: Love it.

spec/abstract/SchedContext_A.thy
mktnk3: > I meant to come back and clean up some of the abstract spec related to `schedule_used`. I'd prefer it to be a monadic function. `schedule_used` is now called in `handle_overrun_loop`, too, so it seems worthwhile, as long as it doesn't mess up the proofs too badly. I think I'll leave this for "another day"

Can you write down what you think we need to do in a JIRA ticket?

michaelmcinerney: I didn't mean anything substantial, just to perform the updates in `refill_budget_check` as in `refillBudgetCheck`, and to have `schedule_used` in the abstract a monadic function, as it is in the Haskell. Would you still like me to make a JIRA ticket?

mktnk3: Yes, even just a todo list would be helpful!

michaelmcinerney: https://sel4.atlassian.net/browse/VER-1396

mktnk3 (approved):
It is great that the overrun fix turned out to be not as bad as we feared, but still, this is again a huge chunk of work. Thanks for pushing this through, Michael. Again, it is good to see various pieces from the past work coming together nicely.

spec/haskell/src/SEL4/Object/SchedContext.lhs
michaelmcinerney: I decided against the readificationizing of `scActive`. There were a few too many lemmas that it would have touched, and it might be best to handle this in some future update that addresses reading/getting more generally

proof/refine/ARM/Ipc_R.thy
michaelmcinerney: https://sel4.atlassian.net/browse/VER-1397

PR #274, opened 2021-04-22
rt: prove valid_replies'

corlewis:
This removes one sorry in `Invariants_H.thy` by adding the `valid_replies'` invariants to `invs'`. This means that we now need to prove that it is preserved instead of just getting to assume that it holds.


corlewis:
This has not been rebased because it is now more difficult to cross the `valid_replies'_sc` assertion in `bindScReply`. The fix for this will possibly require changes to `sendIPC_corres` and `replyPush_corres`.

mktnk3 (approved):
This is a great work! Thanks Corey for finishing off this challenging proof. Only a few cleanup comments from me.

MitchellBuckley (approved):
This looks excellent!

PR #279, opened 2021-05-10
Haskell cleanup

mktnk3:
This is some cleanup, mainly on Haskell, that came up in Michael’s Overrun PR (and others).

I did this as part of the work to finish Refill corres PR, but the cleanup here also overlaps with the Schedule proofs that Mitch wants me to look at, in that both the PRs want `no_ofail` proof for ‘refillHeadOverlapping`. All things considered, I think it makes sense to have this change merged before working on that lemma and other related ones.

the changes included are:
- remove `readMapScPtr` (because this was an extra layer of definition that we could do without)
- cleanup for `updateSchedContext`
- remove dest! attribute for `readObject_misc_ko_at’`
  (we do need to add it quite often, but there are cases when we specifically don’t want it, and it is harder to figure out when the lemma is added by default that we need to remove it)

### DIFF
--- a/proof/refine/ARM/Detype_R.thy
+++ b/proof/refine/ARM/Detype_R.thy
@@ -3325,7 +3325,7 @@ lemma getTCB_det:
   by (clarsimp simp: ko_wp_at'_def getObject_def split_def gets_the_def
                      bind_def gets_def return_def get_def fail_def assert_opt_def
                      no_ofailD[OF no_ofail_tcb_at'_readObject] obj_at'_def projectKOs
-              split: if_splits option.split)
+              split: if_splits option.split dest!: readObject_misc_ko_at')
 
 lemma threadSet_det:
   "tcb_at' ptr s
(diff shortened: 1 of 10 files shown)


MitchellBuckley (approved):
I like these changes. And the proof changes look good!

While I can see that the original abstractions have a kind of neatness to them, for now it seems better to align things to the pre-existing style

corlewis (approved):
I'm happy for this to be merged, but I think we might want to put a bit more thought into `readObject_misc_ko_at'`. Is the issue that you're running in to that we often want `ko_at'` as an assumption but sometimes need `readObject`? If it is, this would suggest to me that our machinery is clashing with itself somewhere, and that the actual problem is that we need to be more consistent in which we use so that we can fully standardise on just one of them in the proofs.

Alternatively, I've just tested it and I'm able to prove that `readObject p s = Some (tcb :: tcb) = ko_at' tcb p s`. Assuming that that also holds for all the other types, then having something like that as a `simp` rule might change it so that the simplifier works for us and makes things consistent automatically.

PR #280, opened 2021-05-10
A small improvement to `rec_del_termination`

MitchellBuckley:
The improvements are:
* add comments into proof
* unwind some automation to clarify how each subgoal is solved.
* Remove some awkward "in monad" lemmas about `premption_point`

### DIFF
--- a/proof/invariant-abstract/CNodeInv_AI.thy
+++ b/proof/invariant-abstract/CNodeInv_AI.thy
@@ -129,11 +129,6 @@ locale CNodeInv_AI =
     "\<And>P d cap. vs_cap_ref (update_cap_data P d cap) = vs_cap_ref cap"
   assumes weak_derived_cap_is_device:
     "\<And>c c'. \<lbrakk>weak_derived c' c\<rbrakk> \<Longrightarrow>  cap_is_device c = cap_is_device c'"
-  assumes in_preempt[simp,intro]:
-    "\<And>rv s' (s::'state_ext state).
-      (Inr rv, s') \<in> fst (preemption_point s) \<Longrightarrow>
-      (\<exists>f es. s' = s \<lparr> machine_state := machine_state s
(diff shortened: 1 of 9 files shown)


MitchellBuckley:
In draft form until I'm sure that all tests pass.

proof/invariant-abstract/CNodeInv_AI.thy
```diff
@@ -902,70 +897,99 @@ lemma not_recursive_cspaces_irq_state_independent[intro!, simp]:
 
 context CNodeInv_AI begin
 
+(* FIXME: Move. Is this safe in wp_comb? *)
```
lsf37, line 900: `wp_comb` has potentially large performance impact, but it looks like straight `wp` might work here. Have you tried that?

lsf37 (approved):
That looks much nicer overall, thanks for the cleanup!

proof/invariant-abstract/CNodeInv_AI.thy
MitchellBuckley: I did try that but it didn't work. 

Since it's apparently wasn't needed until now, I can explicitly add it into the line where I use it and avoid any potential performance impact.

lsf37: That would work. If you want to, we could also measure if there is any real impact by running AInvs+Refine with and without -- if there is no clear difference there, then it should be fine. The `wb_comb` implementation changed a few times, so I might be over-cautious here.

MitchellBuckley: I'm sure it was useful at one point, but now it seems to be unused so assuming tests all pass we can forget about it.

mktnk3 (approved):
Nice cleanup. Thanks for doing this, Mitch!

MitchellBuckley:
All tests passing. @lsf37 happy for me to squash and merge?

lsf37:
Yes, go for it!

PR #281, opened 2021-05-10
Incremental progress towards `schedule_corres`

MitchellBuckley:
This is incremental progress towards `schedule_corres`. The main contributions are:
* `setNextInterrupt_corres`
* `getReprogramTimer_corres`
* `setDomainTime_corres`
* `setConsumedTime_corres`
* `setCurSc_corres`
* `machine_op_lift_corres_trivial`
* `refillSingle_corres`

This also introduces `standard_sc` which might be useful, I'd appreciate some feedback on that.

### DIFF
--- a/proof/refine/ARM/KHeap_R.thy
+++ b/proof/refine/ARM/KHeap_R.thy
@@ -4316,4 +4316,51 @@ global_interpretation set_simple_ko: typ_at_pres "set_simple_ko C ptr ep"
 global_interpretation update_sk_obj_ref: typ_at_pres "update_sk_obj_ref C update ref new"
   unfolding typ_at_pres_def by wpsimp
 
+lemma getReprogramTimer_corres:
+  "corres (=) \<top> \<top> (gets reprogram_timer) getReprogramTimer"
+  by (clarsimp simp: getReprogramTimer_def state_relation_def)
+
+lemma setDomainTime_corres:
+  "dt = dt' \<Longrightarrow>
+  corres dc \<top> \<top> (modify (domain_time_update (\<lambda>_. dt))) (setDomainTime dt')"
+  apply (clarsimp simp: setDomainTime_def, rule corres_modify)
+  by (clarsimp simp: state_relation_def swp_def)
+
+lemma setConsumedTime_corres:
+  "ct = ct' \<Longrightarrow>
+  corres dc \<top> \<top> (modify (consumed_time_update (\<lambda>_. ct))) (setConsumedTime ct')"
+  apply (clarsimp simp: setConsumedTime_def, rule corres_modify)
+  by (clarsimp simp: state_relation_def swp_def)
+
+lemma setCurSc_corres:
+  "sc = sc' \<Longrightarrow>
+   corres dc \<top> \<top> (modify (cur_sc_update (\<lambda>_. sc))) (setCurSc sc')"
+  apply (clarsimp simp: setCurSc_def, rule corres_modify)
+  by (clarsimp simp: state_relation_def swp_def)
+
+lemma refillSingle_equiv:
+  "valid_refills' sc \<Longrightarrow>
+   (length (refills_map (scRefillHead sc) (scRefillCount sc) (scRefillMax sc) (scRefills sc)) = Suc 0)
+    = (scRefillHead sc = refillTailIndex sc)"
+  apply (clarsimp simp: valid_sched_context'_def refillTailIndex_def refills_map_def)
+  apply (case_tac "scRefillCount sc = Suc 0"; simp)
+  apply (auto simp: Let_def)
+  done
+
+lemma refillSingle_corres:
+  "scp = scp' \<Longrightarrow>
+   corres (=) (sc_at scp) (obj_at' valid_refills' scp')
+     (refill_single scp)
+     (refillSingle scp')"
+  unfolding refill_single_def refillSingle_def
+  apply (simp add: refill_size_def get_refills_def)
+  apply (rule stronger_corres_guard_imp)
+    apply (rule_tac R'="\<lambda>sc s. valid_refills' sc" and R="\<lambda>_ _ . True" in corres_split)
+       apply (rule get_sc_corres)
+      apply (clarsimp simp: sc_relation_def)
+      apply (rule refillSingle_equiv; simp)
+     apply wpsimp+
+  apply (clarsimp simp: obj_at'_def)
+  done
+
 end
--- a/proof/refine/ARM/Schedule_R.thy
+++ b/proof/refine/ARM/Schedule_R.thy
@@ -3213,6 +3213,7 @@ lemma schedule_invs':
   apply (rule_tac hoare_seq_ext, rename_tac t)
    apply (rule_tac Q="invs'" in hoare_weaken_pre)
     apply (rule_tac hoare_seq_ext[OF _ getCurThread_sp])
+    apply (rule_tac hoare_seq_ext[OF _ isSchedulable_sp])
     apply (rule_tac hoare_seq_ext[OF _ getSchedulerAction_sp])
     apply (rule hoare_seq_ext)
      apply (wpsimp wp: switchSchedContext_invs')
@@ -3227,7 +3228,7 @@ lemma schedule_invs':
             apply (wpsimp simp: isHighestPrio_def')
            apply (wpsimp wp: curDomain_wp)
           apply (wpsimp simp: scheduleSwitchThreadFastfail_def)
-         apply (rename_tac tPtr isSchedulable x idleThread targetPrio)
+         apply (rename_tac tPtr x idleThread targetPrio)
          apply (rule_tac Q="\<lambda>_. invs' and st_tcb_at' runnable' tPtr and (\<lambda>s. action = ksSchedulerAction s)
                                 and tcb_in_cur_domain' tPtr" in hoare_strengthen_post[rotated])
           apply (prop_tac "st_tcb_at' runnable' tPtr s \<Longrightarrow> obj_at' (\<lambda>a. activatable' (tcbState a)) tPtr s")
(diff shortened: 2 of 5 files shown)


MitchellBuckley:
It only just occurred to me now that some of these may exist later in the proof stack. That seems unlikely, but I'll double-check tomorrow.

mktnk3:
> It only just occurred to me now that some of these may exist later in the proof stack. That seems unlikely, but I'll double-check tomorrow.

That’s a good point. I guess it’ll be SchedContextInv_R if any exists.

proof/refine/ARM/StateRelation.thy
```diff
@@ -281,9 +281,29 @@ lemma refills_heads_equal_valid_sched_context':
...
+(* Most sched contexts should satisfy these conditions. The lemma below (standard_sc_common_usage)
+   illustrates a common situation where these conditions hold.
+
+   It seems clear that an abbreviation of this kind is useful, but it's not clear exactly which
+   conditions it should contain. For instance, it does not contain "0 < scRefillMax sc" which one
+   may argue is equally standard.
+*)
+abbreviation (input) standard_sc where
+  "standard_sc sc \<equiv> scRefillMax sc \<le> length (scRefills sc) \<and> scRefillHead sc < scRefillMax sc \<and>
+                    scRefillCount sc \<le> scRefillMax sc \<and> 0 < scRefillCount sc"
```
mktnk3, line 297: I like this and I have myself encountered situations where I could use something like this. It makes sense to leave it as an abbreviation till we are sure of what conditions it should contain, but we could make it a definition in the future, maybe with a different name, though. “standard_sc” sounds rather too unspecific to me. These are conditions we need for the `wrap_slice` function to make sense, which we use to represent the circular refill list/buffer. Maybe something that features “refills”, like`refills_cond`?

I am starting to think that it would make sense to include these conditions as part of `sc_relation` because the relation between `sc_refills` and `scRefills` won’t hold without them.

proof/refine/ARM/KHeap_R.thy
```diff
@@ -4288,4 +4288,64 @@ global_interpretation set_simple_ko: typ_at_pres "set_simple_ko C ptr ep"
...
+   (length (refills_map (scRefillHead sca) (scRefillCount sca) (scRefillMax sca) (scRefills sca)) = Suc 0)
+    = (scRefillHead sca = refillTailIndex sca)"
+  apply (clarsimp simp: valid_sched_context'_def refillTailIndex_def refills_map_def)
+  apply (case_tac "scRefillCount sca = Suc 0"; simp)
+  apply (auto simp: Let_def)
+  done
+
+lemma refillSingle_corres:
+  "scp = scp' \<Longrightarrow>
+   corres (=) (sc_at scp) (sc_at' scp' and obj_at' standard_sc scp')
```
mktnk3, line 4337: This is just a comment but you can derive `sc_at’` from the `obj_at’` part.

spec/haskell/src/SEL4/Kernel/Thread.lhs
```diff
@@ -352,12 +353,11 @@ An IPC operation may request that the scheduler switch to a specific thread.
 We check here that the candidate has the highest priority in the system.
 
 >         SwitchToThread candidate -> do
->             isSchedulable <- isSchedulable curThread
 >             when isSchedulable (tcbSchedEnqueue curThread)
 >
 >             idleThread <- getIdleThread
 >             targetPrio <- threadGet tcbPriority candidate
->             curPrio <- if idleThread == curThread then return 0 else threadGet tcbPriority curThread
+>             curPrio <- if (curThread /= idleThread) then (threadGet tcbPriority curThread) else (return 0)
```
mktnk3, line 360: Just curious, did you have a specific reason for this change?

mktnk3 (approved):
Nice work! Only minor comments from me, maybe apart from the naming suggestion.

proof/refine/ARM/StateRelation.thy
mktnk3: Well, I guess it’s more of side conditions for valid sc_relation. We can introduce the notion of valid refills for the concrete side.

proof/refine/ARM/KHeap_R.thy
```diff
@@ -4288,4 +4288,64 @@ global_interpretation set_simple_ko: typ_at_pres "set_simple_ko C ptr ep"
...
+  by (clarsimp simp: state_relation_def swp_def)
+
+lemma setCurSc_corres:
+  "sc = sc' \<Longrightarrow>
+   corres dc \<top> \<top> (modify (cur_sc_update (\<lambda>_. sc))) (setCurSc sc')"
+  apply (clarsimp simp: setCurSc_def, rule corres_modify)
+  by (clarsimp simp: state_relation_def swp_def)
+
+lemma machine_op_lift_corres_trivial:
+  "f = f' \<Longrightarrow> corres_underlying Id False True dc \<top> \<top> (machine_op_lift f) (machine_op_lift f')"
```
corlewis, line 4315: This seems like a little bit of a strange lemma to me. It's useful, but couldn't it follow more directly from `corres_underlying_trivial`?

proof/refine/ARM/Schedule_R.thy
```diff
@@ -3864,6 +3865,114 @@ lemma awaken_corres:
...
+                  apply wpsimp
+                 apply wpsimp
+                apply (rule corres_return_eq_same)
+                apply (clarsimp simp: refills_heads_equal refill_map_def)
+               apply (rule corres_split [OF getReleaseQueue_corres])
+                 apply (rule corres_split [OF corres_if], simp)
+                     apply (rule corres_return_eq_same, simp)
+                    apply (rule corres_split_eqr)
+                       apply (rule corres_assert_opt_assume_l)
+                       apply (simp)
```
corlewis, line 3936: tiny nitpick
```suggestion
                       apply simp
```

proof/refine/ARM/Schedule_R.thy
```diff
@@ -3864,6 +3865,114 @@ lemma awaken_corres:
...
+                         apply (rule corres_return_eq_same)
+                         apply (clarsimp simp: refills_heads_equal refill_map_def)
+                        apply wpsimp
+                       apply wpsimp
+                      apply (simp, rule get_tcb_obj_ref_corres)
+                      apply (clarsimp simp: tcb_relation_def)
+                     apply (wpsimp wp: get_tcb_obj_ref_wp)
+                    apply (wpsimp wp: threadGet_wp)
+                   apply (rule corres_machine_op)
+                   apply (simp only:, rule setDeadline_corres, simp)
```
corlewis, line 3948: Out of curiosity, what is that `simp only:` needed for?

proof/refine/ARM/Schedule_R.thy
```diff
@@ -3864,6 +3865,114 @@ lemma awaken_corres:
...
+                  apply wpsimp+
+          apply (clarsimp simp: tcb_relation_def)
+         apply (wpsimp wp: get_tcb_obj_ref_wp)
+        apply (wpsimp wp: threadGet_wp)
+       apply wpsimp+
+   apply (subgoal_tac "release_queue s \<noteq> [] \<longrightarrow> active_sc_tcb_at (hd (release_queue s)) s")
+    apply (fastforce simp: cur_tcb_def pred_tcb_at_def vs_all_heap_simps obj_at_def is_sc_obj is_tcb
+                     elim: valid_sched_context_size_objsI)[1]
```
corlewis, line 3956: nitpick: unnecessary `[1]`

corlewis (approved):
Looks good!

spec/haskell/src/SEL4/Kernel/Thread.lhs
MitchellBuckley: It corresponds more closely to the other specifications.

proof/refine/ARM/KHeap_R.thy
MitchellBuckley: Oops! Good point!

proof/refine/ARM/StateRelation.thy
MitchellBuckley: How about `valid_refills'` or `standard_refills`?

proof/refine/ARM/Schedule_R.thy
MitchellBuckley: Good point, I've changed it to use `simp del: dc_simp`.

proof/refine/ARM/StateRelation.thy
mktnk3: I vote for `valid_refills’` :)

proof/refine/ARM/StateRelation.thy
```diff
@@ -281,9 +281,29 @@ lemma refills_heads_equal_valid_sched_context':
...
+(* Most sched contexts should satisfy these conditions. The lemma below (valid_refills'_common_usage)
+   illustrates a common situation where these conditions hold.
+
+   It seems clear that an abbreviation of this kind is useful, but it's not clear exactly which
+   conditions it should contain. For instance, it does not contain "0 < scRefillMax sc" which one
+   may argue is equally standard.
+*)
+abbreviation (input) valid_refills' where
+  "valid_refills' sc \<equiv> scRefillMax sc \<le> length (scRefills sc) \<and> scRefillHead sc < scRefillMax sc \<and>
+                    scRefillCount sc \<le> scRefillMax sc \<and> 0 < scRefillCount sc"
```
mktnk3, line 297: With its new name and all that, I would be happy to have this abbreviation without `(input)`. I have started to use this with projections and it seems to make sense. We can have this as a conjunct in `valid_sched_context’`. Not quite sure if we should handle this separately yet, although I realise we do have `valid_refills` as a notion separate from `valid_obj`.

mbrcknl (approved):

proof/refine/ARM/StateRelation.thy
mktnk3: Looking at the merged files, the previous version of this (`std_sc’)` is still sitting towards the end of this file. You originally called this one here as `standard_sc` and that made me think it was the same thing as your `std_sc’` from the previous PR.  But does this mean there was actually some duplication? Surely we don’t need both, do we?

mktnk3: I can clean it up in my PR.

MitchellBuckley: Any duplication is a mistake, I was only intending to use one notion.

PR #285, opened 2021-05-28
Import documentation: Haskell assertions

zaklogician:
We import some documentation regarding the role of assertions in Haskell, and how we use assertions in Haskell to transport information from abstract invariants to Haskell-to-C refinement proofs.

The file is a Markdown-ified version of previous documentation hosted at UNSW and Data61.

(This is a pure addition to docs, and should not change or impact verification in any way)

### DIFF
--- a/docs/haskell-assertions.md
+++ b/docs/haskell-assertions.md
@@ -1,0 +1,68 @@
+This is a snapshot of a monologue by @mbrcknl about assertions and monadic failure generally, and more specifically about how we can use assertions in Haskell to transport information from abstract invariants to proofs of Haskell-to-C refinement.
+
+> :mag: **tl;dr**
+> * assertions (assert) in Haskell give you:
+>   * free assumptions in wp
+>   * proof obligations in `corres` in `Refine`
+>   * free assumptions in `ccorres` in `CRefine`
+> * you can use them to transport information (properties, invariants) from `AInvs` down to Haskell and C:
+>   * if you assert `P'` in Haskell
+>   * and can derive `[| (s,s') : state_relation; P s |] ==> P' s'`
+>   * and can derive `P` in/from AInvs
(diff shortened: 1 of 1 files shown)


lsf37 (approved):
:+1:

There seems to be trailing whitespace that still needs to be fixed up

docs/haskell-assertions.md
```diff
@@ -0,0 +1,76 @@
...
+
+* If `nf` is `True`, then we get to assume that the failure flag is not set on the abstract side.
+* If `nf` is False, `corres_underlying` doesn't care about the failure flag on the abstract side.
+
+Concerning `nf'`:
+* If `nf'` is `True`, then we have to prove that the failure flag is not set on the concrete side.
+* If `nf'` is `False`, then `corres_underlying` doesn't care about the failure flag on the concrete side.
+
+
+For `corres`, we take the strongest combination `(False, True)`. So we don't get to assume that the failure flag is not set on the abstract side, but we have to prove that the failure flag is not set on the concrete side.
```
mbrcknl, line 34: Nit: consistent whitespace formatting.
```suggestion
Concerning `nf`:
* If `nf` is `True`, then we get to assume that the failure flag is not set on the abstract side.
* If `nf` is False, `corres_underlying` doesn't care about the failure flag on the abstract side.

Concerning `nf'`:
* If `nf'` is `True`, then we have to prove that the failure flag is not set on the concrete side.
* If `nf'` is `False`, then `corres_underlying` doesn't care about the failure flag on the concrete side.

For `corres`, we take the strongest combination `(False, True)`. So we don't get to assume that the failure flag is not set on the abstract side, but we have to prove that the failure flag is not set on the concrete side.
```

mbrcknl (approved):
Looks good! Just a couple of whitespace nits to fix.

corlewis (approved):
Looks good to me too! Thanks for doing this, and thanks @mbrcknl for originally writing it.

PR #288, opened 2021-06-17
Corres rename

MitchellBuckley:
This a revised version of Victor's original corres rename pull-request.

This does a mass renaming of corres lemmas across l4v. The renamings are done by session and only effect `Refine`, `Infoflow`, and `LibTest`. The renaming is done by looping through `renames.txt` and applying a suitable `sed` command. This is acheived by running the bash script in the appropriate directory.

There is a commit which adds the script and `renames.txt` and a commit that removes them. Unless we want to keep those scripts and list, the two commits should be squashed (to an empty commit) before merging.

### DIFF
--- a/lib/test/Corres_Test.thy
+++ b/lib/test/Corres_Test.thy
@@ -178,15 +178,15 @@ context begin interpretation Arch .
 
 
 lemmas load_hw_asid_corres_args[corres] =
-  load_hw_asid_corres[@lift_corres_args]
+  loadHWASID_corres[@lift_corres_args]
 
 lemmas invalidate_asid_corres_args[corres] =
-  invalidate_asid_corres[@lift_corres_args]
+  invalidateASID_corres[@lift_corres_args]
 
 lemmas invalidate_hw_asid_entry_corres_args[corres] =
-  invalidate_hw_asid_entry_corres[@lift_corres_args]
+  invalidateHWASIDEntry_corres[@lift_corres_args]
 
-lemma invalidate_asid_entry_corres:
+lemma invalidateASIDEntry_corres:
   "corres dc (valid_vspace_objs and valid_asid_map
                 and K (asid \<le> mask asid_bits \<and> asid \<noteq> 0)
                 and vspace_at_asid asid pd and valid_vs_lookup
@@ -218,10 +218,10 @@ crunch ksCurThread[wp]: invalidateASIDEntry, flushSpace "\<lambda>s. P (ksCurThr
 crunch obj_at'[wp]: invalidateASIDEntry, flushSpace "obj_at' P p"
 
 lemmas flush_space_corres_args[corres] =
-  flush_space_corres[@lift_corres_args]
+  flushSpace_corres[@lift_corres_args]
 
 lemmas invalidate_asid_entry_corres_args[corres] =
-  invalidate_asid_entry_corres[@lift_corres_args]
+  invalidateASIDEntry_corres[@lift_corres_args]
 
 
 lemma corres_inst_eq_ext:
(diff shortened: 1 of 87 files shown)


lsf37:
Nice. Starting testboard for it: @ssrg-bamboo `proof`

lsf37:
The bot doesn't seem to want to talk to me today, have kicked off a testboard manually.

lsf37 (approved):
Much improved consistency, thanks for finishing this off. Happy for it to be merged as is.

Do we have a plan for bringing this over or repeating it on the rt branch?

MitchellBuckley:
I just noticed that 6/261 listed renames were never applied (the first 6) because the renames.txt file was corrupted at some point. I could either:
1. Apply the last 6 renames and update this PR.
2. Leave them and fix do them in a following PR.
I've already attempted 1. but renaming one of those 6 interferes with CRefine and I haven't figured out how to fix that yet. So I'm tempted to go with 2.

No, I don't have a plan for how this interacts with `rt`.

lsf37:
I'm happy with option 2.

Interaction with `rt` is a bit of a worry for the next merge, because this touches a lot of lines. If we do the same on `rt`, we can probably then ignore this commit for the merge (merge everything before, then merge this commit resolving everything to `rt`, then continue merge).

Would it make sense to run the same `rename.txt` file on `rt`? Would that be a lot of work?

lsf37:
@Xaphiosis do you have an opinion on this?

MitchellBuckley:
Excellent. Should I squash the two commits (add and remove) before merging?

Running the same script on `rt` should be easy. I don't expect any complications there.

lsf37:
It'd be good to preserve the script, actually. We might want to use it for other things as well.

Instead of the add/remove commits, could you add the script somewhere in `misc` (probably its own subdirectory), and maybe leave the `rename.txt` in there as an example? You'd need to add a README to explain what it is and how it is used.

Xaphiosis:
My thoughts:
* Go with option 2 for now if that's easier.
* Would be nice to have one or more of the commits say what the reason for the rename is (consistency with something or other); "mass rename" alone doesn't indicate why.
* Agree keeping script around would be useful, agree with putting it in misc/, disagree with keeping rename.txt in its full glory. Would prefer a little documentation at the top of the script instructing how to use it, and the license checker will insist on putting a license in anyway. Otherwise a README.md?

MitchellBuckley:
> My thoughts:
> 
> * Go with option 2 for now if that's easier.
> * Would be nice to have one or more of the commits say what the reason for the rename is (consistency with something or other); "mass rename" alone doesn't indicate why.
> * Agree keeping script around would be useful, agree with putting it in misc/, disagree with keeping rename.txt in its full glory. Would prefer a little documentation at the top of the script instructing how to use it, and the license checker will insist on putting a license in anyway. Otherwise a README.md?

Done. Mind having a look and providing feedback on the README? Also, I don't know how to give the script the appropriate license.

lsf37:
Looks good from my side. The script should get this header:

```
# Copyright 2021, Data61, CSIRO (ABN 41 687 119 230)
#
# SPDX-License-Identifier: BSD-2-Clause
```

and the README.md the same as the top-level README.md (CC-BY-SA-4.0)

Xaphiosis:
README content looks good. Consider wrapping at some standard amount (80 chars?)

MitchellBuckley:
@lsf37 I don't understand the portable shell error. Any tips?

lsf37:
> @lsf37 I don't understand the portable shell error. Any tips?

The linter is complaining (correctly in this case), that the script is bash-specific and not portable to other shells. In particular, regular expressions are a bash feature that doesn't necessarily exist in other shells. I think that's fine to ignore here, and live with the fact that not all of our scripts are portable.

It's more important for seL4 scripts, because they might have to run in all kinds of environments, but the verification has a fairly narrow set of development platforms anyway.

I'll leave the check on, because if we can easily make a script portable, we should, but in this case, let's just ignore and merge.

MitchellBuckley:
> > @lsf37 I don't understand the portable shell error. Any tips?
> 
> The linter is complaining (correctly in this case), that the script is bash-specific and not portable to other shells. In particular, regular expressions are a bash feature that doesn't necessarily exist in other shells. I think that's fine to ignore here, and live with the fact that not all of our scripts are portable.
> 
> It's more important for seL4 scripts, because they might have to run in all kinds of environments, but the verification has a fairly narrow set of development platforms anyway.
> 
> I'll leave the check on, because if we can easily make a script portable, we should, but in this case, let's just ignore and merge.

Ah, that makes sense thank you. So this is a PR-only check (will not fail regression tests)?

Ok for this to merge now?

lsf37:
Yes, Ok to merge as is.

mktnk3:
It is great to see this merged! Thank you so much for finishing this off, Mitch. And thanks Gerwin & Raf for reviewing. I suppose we would like to apply Mitch’s script to MCS and for that we probably need to review the rename list first. Hope we get to do that soon.

PR #289, opened 2021-06-17
Enable CSpec in MCS regression test

lsf37:
Making sure that CKernel and CSpec continue to work in the MCS branch.

Also contains a small commit to sync `kernel.mk` with the `master` branch.


lsf37:
@ssrg-bamboo `proof`

Xaphiosis (approved):

lsf37:
Ok, two small failures still: CBaseRefine for RISCV64 (will exclude for now until Refine works), and `LibTest` fails for ARM, because the `wp` tutorial refers to functions that don't exist in MCS.

If that's less that 10 min to fix I'll do that, otherwise also exclude LibTest.

lsf37:
Ok, tests succeeded, will merge. It looks like the testboard does not handle skipped platforms gracefully. One more thing to look at in the future..

PR #293, opened 2021-06-24
refine: The state relation is not empty.

MitchellBuckley:
Defined a series of `zeroed` states on the abstract and haskell sides
that together satisfy the state relation. These states do not satisfy
invs or invs'.

This is connected to https://sel4.atlassian.net/browse/VER-1342

### DIFF
--- a/proof/refine/ARM/ADT_H.thy
+++ b/proof/refine/ARM/ADT_H.thy
@@ -10,6 +10,7 @@ theory ADT_H
 imports
   "AInvs.ADT_AI"
   Syscall_R
+  Init_R
 begin
 
 text \<open>
--- a/proof/refine/ARM/Init_R.thy
+++ b/proof/refine/ARM/Init_R.thy
@@ -1,0 +1,131 @@
+(*
+ * Copyright 2021, Data61, CSIRO (ABN 41 687 119 230)
+ *
+ * SPDX-License-Identifier: GPL-2.0-only
+ *)
+
+theory Init_R
+imports
+  KHeap_R
+
+begin
+
+context begin interpretation Arch . (*FIXME: arch_split*)
+
+(*
+  This provides a very simple witness that the state relation used in the first refinement proof is
+  non-trivial, by exhibiting a pair of related states. This helps guard against silly mistakes in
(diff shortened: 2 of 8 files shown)


MitchellBuckley:
This is ARM only, should I extend to all architectures?

lsf37:
> This is ARM only, should I extend to all architectures?

Yes, that would be good. This is mostly to guard against typos and stupid mistakes in the state relation, which we might introduce independently on each arch :-)

mbrcknl (commented):
Looks good, and will give some protection against stuff-ups in the state relation.

Would be good to have on all arches, especially if it's this simple.

lsf37 (approved):

mbrcknl (approved):
Would be good to have some comments at the top of the theory about what purpose this serves, and the caveats. Something like:

> This provides a very simple witness that the state relation used in the first refinement proof is non-trivial, by exhibiting a pair of related states. This helps guard against silly mistakes in the state relation, since we currently assume that the system starts in a state satisfying invariants and state relations.

> Note that the states we exhibit are *not* intended to be useful states. They are just the simplest possible states that prove non-triviality of the state relation. In particular, these states do *not* satisfy the respective invariant conditions. In future, this could be improved by exhibiting a tuple of more realistic states that are related across all levels of the refinement, and that also satisfy respective invariant. Ultimately, we would like to prove functional correctness of kernel initialisation. That would allow us to start from a minimal *but real* configuration that would allow us to make a much smaller set of assumptions about the initial configuration of the system.

With that, and versions for other arches, I think this is good to merge.

Xaphiosis:
Good idea!
Agree with what others have said (other arches, and especially Matt on documentation); maybe also add a little comment to the commit? It seems a bit cryptic without examination.

Also very minor style nitpick, I think if it fits on one line you don't need to wrap the type after the `::`, i.e.
`definition zeroed_intermediate_state :: global.kernel_state where`
Which reminds me, we should really go over your draft style guide when we have a chance, your efforts didn't really get the attention they deserved.

mbrcknl (approved):
Looks good!

PR #294, opened 2021-06-24
always use `addrFromKPPtr` for kernel addresses

mbrcknl:
This verifies a C kernel patch (seL4/seL4#409) which consolidates
translation between virtual and physical addresses, and makes it
consistent across architectures. In particular, we always use
`addrFromKPPtr`, even on architectures that don't use a distinct region
to map the kernel ELF. This will facilitate future improvements which
move the ELF mapping into a distinct virtual address region.

Signed-off-by: Matthew Brecknell <Matthew.Brecknell@data61.csiro.au>

### DIFF
--- a/proof/crefine/ARM/VSpace_C.thy
+++ b/proof/crefine/ARM/VSpace_C.thy
@@ -705,24 +705,29 @@ lemma pde_case_isPageTablePDE:
   by (cases pde, simp_all add: isPageTablePDE_def)
 
 lemma ptrFromPAddr_spec:
-  "\<forall>s. \<Gamma> \<turnstile>  {s}
-  Call ptrFromPAddr_'proc
-  \<lbrace>  \<acute>ret__ptr_to_void =  Ptr (ptrFromPAddr (paddr_' s) ) \<rbrace>"
+  "\<forall>s. \<Gamma> \<turnstile> {s}
+   Call ptrFromPAddr_'proc
+   \<lbrace>\<acute>ret__ptr_to_void =  Ptr (ptrFromPAddr (paddr_' s))\<rbrace>"
   apply vcg
-  apply (simp add: ARM.ptrFromPAddr_def pptrBaseOffset_def
-                   pptrBase_def physBase_def ARM.physBase_def)
+  apply (simp add: ptrFromPAddr_def pptrBaseOffset_def pptrBase_def physBase_def)
   done
 
 lemma addrFromPPtr_spec:
-  "\<forall>s. \<Gamma> \<turnstile>  {s}
-  Call addrFromPPtr_'proc
-  \<lbrace>  \<acute>ret__unsigned_long =  (addrFromPPtr (ptr_val (pptr_' s)) ) \<rbrace>"
+  "\<forall>s. \<Gamma> \<turnstile> {s}
+   Call addrFromPPtr_'proc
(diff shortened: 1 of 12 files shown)


mbrcknl:
This currently has a nasty hack to make `ARM` and `ARM_HYP` CRefine proofs work. While the tests run on this, I'll go back and update the specs to make something more reasonable.

xurtis:
> This currently has a nasty hack to make `ARM` and `ARM_HYP` CRefine proofs work. While the tests run on this, I'll go back and update the specs to make something more reasonable.

Do you plan to make this fix before merging?

mbrcknl:
> > This currently has a nasty hack to make `ARM` and `ARM_HYP` CRefine proofs work. While the tests run on this, I'll go back and update the specs to make something more reasonable.
> 
> Do you plan to make this fix before merging?

Oh, sorry, I should have been clearer that "ready to merge" meant that this is already fixed. :-)

mbrcknl:
The hack was taking advantage of the fact that `addrFromKPPtr` currently gives the same result as `addrFromPPtr` on the verified ARM and ARM_HYP configurations. The fix was to define them separately in the specs, so that if we ever give the ELF its own address space on those platforms, the proof update won't require the next person to figure out the hack. That update will just mirror the update to the C.

lsf37 (approved):
Looks good from my side, and thanks for the style cleanup as well!

@Xaphiosis should look at this to make sure that we don't clash with platform ports, but I think it's fine.

Xaphiosis:
I'm keep getting confused between `kernelELFPAddrBase` and `physBase`. After staring at this for a while, I think this works.

Have the updates to use addrFromKPPtr already make it to the Haskell/abstract spec for ARM and ARM_HYP? It would be nice to have conformance across arches on this, even if as you say "addrFromKPPtr currently gives the same result as addrFromPPtr on the verified ARM and ARM_HYP configurations". I don't remember if this was done, but it would be good not to introduce surprises.

mbrcknl:
> Have the updates to use addrFromKPPtr already make it to the Haskell/abstract spec for ARM and ARM_HYP? It would be nice to have conformance across arches on this, even if as you say "addrFromKPPtr currently gives the same result as addrFromPPtr on the verified ARM and ARM_HYP configurations". I don't remember if this was done, but it would be good not to introduce surprises.

Yeah, that's done. I only abused the fact that `addrFromKPPtr` and `addrFromPPtr` give the same result during initial proof exploration, but I don't do that any more. They're now defined separately on all arches, and used in the appropriate places in all the specs.

Xaphiosis (approved):

xurtis:
> The hack was taking advantage of the fact that `addrFromKPPtr` currently gives the same result as `addrFromPPtr` on the verified ARM and ARM_HYP configurations. The fix was to define them separately in the specs, so that if we ever give the ELF its own address space on those platforms, the proof update won't require the next person to figure out the hack. That update will just mirror the update to the C.

That is, in fact, the next step.

xurtis:
> Yeah, that's done. I only abused the fact that `addrFromKPPtr` and `addrFromPPtr` give the same result during initial proof exploration, but I don't do that any more. They're now defined separately on all arches, and used in the appropriate places in all the specs.

Does this mean a change that changes the implementation to use a dedicated ELF region on ARM wouldn't be too difficult to update in the proofs?

mbrcknl:
> Does this mean a change that changes the implementation to use a dedicated ELF region on ARM wouldn't be too difficult to update in the proofs?

That would be my hope. :-)

PR #290, opened 2021-06-18
Refill spec cleanup

mktnk3:
This PR contains several lines of cleanups and improvements that I hope will help with the upcoming refill related proofs. The brief descriptions of the changes are as follows:

- more systematic use of updateSchedContext and its derivatives, mainly in refill related functions
- abstract spec cleanup to go along with the above, with some rewrite rules to take advantage of existing proofs
- more streamlined approach for indexing haskell scRefills (`updateAt`)
- rewrite rules for simplifying projections in the abstract side statements
- projection style wp rules for `getSchedContext`, `updateSchedContext` and others
- rewrite rules for `updateSchedContext`, etc., for managing some complex corres proofs
- rewrite some refill related corres in a more modular style

There are some smaller lemmas that can be moved to KHeap_R or StateRelation.
I think the “projection” rewrites block can stay in Schedule_R where it currently sits, for the time being, unless we are going to them in the earlier theories.

PR #295, opened 2021-06-26
updateSchedContext_corres, opt_pred, some cleanup, etc.

mktnk3:
This PR contains the following changes, part of the refill spec cleanup PR (#290 ):

- a most generic corres rule and some wp rules for `updateSchedContext`
- some abbreviations and rewrite lemmas for simplifying AInvs projections
- add a bunch of wrap_slice and refill related lemmas
- add `length_replaceAt` to the simp set, and related cleanup

Some more detailed comments on those changes:

- the `updateSchedContext_corres` rules is generic enough to handle all the refill related corres rules in the main PR #290 . The way we implement refill circular list in Haskell introduces unusual interdependency and the rule requires assumptions involving both the abstract and concrete states that satisfy the state relation. This probably suggests one direction we could extend our corres framework, although it is likely to be rare that we need this level of generality and there are ways to get around it.

- there are a bunch of monadic_rewrite lemma on `updateSchedContext` that are used in #290 

- `opt_pred` will let us write conditions on projections in a concise way, and it can be unfolded as needed easily from the heap using `opt_map_left_Some`.
```
abbreviation
  opt_pred :: "('a => bool) => ('b => 'a option) => ('b => bool)" (infixl "|<" 50) where
  "P |< proj == (\<lambda>x. case_option False P (proj x))"
```
Basically, we can have the form `(predicate) |< (layers of projections)`, for instance:
```
bounded_release_time_2 |< (scs_of ||> sc_refills |> hd_opt ||> r_time)
```
There is freedom in the ways to write the predicate and the projections in that, for example, you can apply `r_time` in the predicate instead of projection in the case above. The choice of the direction of application is to mimic the usual predicates, but people might have other preferences.

This can replace `pred_map`, which contains one layer of definition that we could do without. The benefit of replacing will be more smooth and automatic but controlled unfolding and more compact notations. We can use the rewrite lemma to do on-demand replacement:
```
lemma pred_map_rewrite:
  "pred_map P proj = opt_pred P proj"
  by (fastforce simp: pred_map_def2)
```
We could rename `opt_pred` to `pred_map`, too, but I think I’d prefer featuring `opt` for this.

- some wp rules for `updateSchedContext` have preconditions of the form `sc_at’ scp & (conditions on projections)` instead of `!ko. ko_at’ ko scp --> conditions on ko`. This often works nicely because there is no need to deal with superfluous intermediate `ko`s being updated.

- there are some more abbreviations and rewrite lemmas for simplifying AInvs projections in addition to `opt_pred`, such as `scs_of2` which is a full abbreviation that can replace `scs_of`. These are part of my bigger project to simply AInvs DetSched* using simpler projections ([link to WIP](https://github.com/seL4/l4v/blob/projection_rewrite/proof/invariant-abstract/DetSchedInvs_AI.thy#L3857)). I have included some of those simplifications in #290 to make the abstract side conditions easier to handle and more in sync with the concrete side, and it seemed to work quite well.

- the reason for wanting to simply AInvs DetSched* projections is mainly because, although I believe what we currently have is part of the necessary journey to where we are now, it contains a fair bit of legacy definitions that we probably don’t need that can be quite unintuitive to unfold and potential cause of slow down. But the AInvs part is for later and open for discussion. What I would like to be merged soon is a smaller portion that we could use for finishing of the rest of Refine.

- I would really like a shorter name for `opt_map_left_Some`. It is often needed and very handy but has to be added manually, and it’s a bit tedious to type. Any suggestions?
```
lemma opt_map_left_Some:
  “f x = Some y ==> (f |> g) x = g y”
```

### DIFF
--- a/proof/refine/ARM/KHeap_R.thy
+++ b/proof/refine/ARM/KHeap_R.thy
@@ -3958,14 +3958,6 @@ lemma sc_replies_relation_rewrite:
 
 (* end : projection rewrites *)
 
-lemma sc_replies_relation_prevs_list':
-  "\<lbrakk> sc_replies_relation s s';
-     kheap s scp = Some (kernel_object.SchedContext sc n)\<rbrakk>
-    \<Longrightarrow> heap_ls (replyPrevs_of s') (scReplies_of s' scp) (sc_replies sc)"
-  apply (clarsimp simp: sc_replies_relation_def sc_replies_of_scs_def scs_of_kh_def map_project_def)
-  apply (clarsimp simp: opt_map_left_Some sc_of_def)
-  done
-
 (* updateSchedContext *)
 
 context begin interpretation Arch . (*FIXME: arch_split*)
@@ -4046,7 +4038,7 @@ lemma state_relation_sc_update:
      apply (erule obj_relation_cutsE)
             apply ((simp split: if_split_asm)+)[8]
     (* sc_replies_relation *)
-    apply (frule (1) sc_replies_relation_prevs_list'[simplified])
+    apply (frule (2) sc_replies_relation_prevs_list[simplified])
     apply (subst replyPrevs_of_non_reply_update[simplified]; (simp add: typ_at'_def ko_wp_at'_def)?)
     apply (simp add: sc_replies_relation_def)
     apply (rule conjI)
@@ -4148,8 +4140,6 @@ lemma updateSchedContext_corres_gen:
   done
 
 lemmas updateSchedContext_corres = updateSchedContext_corres_gen[where P=\<top> and P'=\<top>, simplified]
-
-declare opt_map_left_Some[simp]
 
 (* end : updateSchedContext *)
 
@@ -4579,8 +4569,8 @@ lemma refillSingle_corres:
   apply (rule stronger_corres_guard_imp)
     apply (rule_tac R'="\<lambda>sc s. valid_refills' sc" and R="\<lambda>_ _ . True" in corres_split)
        apply (rule get_sc_corres)
-      apply (clarsimp simp: sc_relation_def)
-      apply (rule refillSingle_equiv; simp)
+      apply simp
+      apply (metis (mono_tags, hide_lams) refillSingle_equiv sc_relation_def)
      apply wpsimp+
   apply (clarsimp simp: obj_at'_def)
   done
--- a/proof/refine/ARM/KHeap_R.thy
+++ b/proof/refine/ARM/KHeap_R.thy
@@ -3545,7 +3545,6 @@ lemma st_tcb_at_coerce_abstract:
   apply fastforce
   done
 
-end
 lemma st_tcb_at_coerce_concrete:
   assumes t: "st_tcb_at P t s"
   assumes sr: "(s, s') \<in> state_relation" "pspace_aligned s" "pspace_distinct s"
@@ -3664,8 +3663,6 @@ lemma valid_tcb_state_cross:
   by (fastforce dest: ep_at_cross reply_at_cross ntfn_at_cross
                 simp: valid_bound_obj'_def valid_tcb_state_def valid_tcb_state'_def
                split: Structures_A.thread_state.split_asm option.split_asm)
-
-context begin interpretation Arch . (*FIXME: arch_split*)
 
 lemma state_refs_of_cross_eq:
   "\<lbrakk>(s, s') \<in> state_relation; pspace_aligned s; pspace_distinct s\<rbrakk>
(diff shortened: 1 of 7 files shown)


proof/refine/ARM/Reply_R.thy
```diff
@@ -1160,27 +1160,33 @@ lemma valid_objs'_replyNexts_of_reply_at':
...
 (** sc_with_reply and sc_replies_relations : crossing information **)
 
+(* modified version of sc_replies_relation_prevs_list in StateRelation.thy;
+   updates only the abstract sc_relies; useful for the following few lemmas *)
 lemma sc_replies_relation_prevs_list':
   "\<lbrakk> sc_replies_relation s s';
-     kheap s scp = Some (kernel_object.SchedContext sc n)\<rbrakk>
-    \<Longrightarrow> heap_ls (replyPrevs_of s') (scReplies_of s' scp) (sc_replies sc)"
+     kheap s x = Some (kernel_object.SchedContext sc n)\<rbrakk>
+    \<Longrightarrow> heap_ls (replyPrevs_of s') (scReplies_of s' x) (sc_replies sc)"
```
mktnk3, line 1168: I will revert this back to `scp`a. This was once removed and then resurrected, but somehow didn’t quite make a full round-trip...

proof/refine/ARM/KHeap_R.thy
```diff
@@ -3917,6 +3914,180 @@ lemma refillSufficient_wp:
...
+    by (clarsimp simp: obj_at_simps)
+  have S: "\<And>(v::'a::pspace_storable). (1 :: word32) < 2 ^ (objBits v)"
+    by (clarsimp simp: obj_at_simps objBits_defs pteBits_def pdeBits_def scBits_pos_power2
+                split: kernel_object.splits arch_kernel_object.splits)
+  assume H: "(s, s') \<in> state_relation" "P s" "P' s'" "sc_at ptr s" "sc_at' ptr s'"
+  show ?thesis
+    using H S sz
+  apply -
+    apply (insert R1[rule_format, OF H(1) H(2) H(3) H(4) H(5)]
+                  R2[rule_format, OF H(1) H(2) H(3) H(4) H(5)])
```
mbrcknl, line 4014: tip: You can just say `OF H` if `H` has exactly 5 lemmas in the right order.

mbrcknl:
> We could rename `opt_pred` to `pred_map`, too, but I think I’d prefer featuring `opt` for this.

Just a historical  note: the `map` part came from Isabelle's naming the type `'a => 'b option` a `map`. I think you can find this by clicking through the right harpoon notation. I find the word `map` too heavily overloaded, though, so would be happy to switch to the name `opt_pred`.

mbrcknl:
> * I would really like a shorter name for `opt_map_left_Some`. It is often needed and very handy but has to be added manually, and it’s a bit tedious to type. Any suggestions?
> 
> ```
> lemma opt_map_left_Some:
>   “f x = Some y ==> (f |> g) x = g y”
> ```

`opt_reduce`? Not a whole lot shorter, but only requires you to use the shift key once.

proof/refine/ARM/KHeap_R.thy
mktnk3: ah, that’s good to know!

mktnk3:
>  I find the word `map` too heavily overloaded, though, so would be happy to switch to the name `opt_pred`.

I definitely agree about `map`.

`opt_reduce` sounds nice, but could it be shorter? e.g., `opt_red`, `opt_simp`, `opt_step`....hmm.

lsf37:
> > * I would really like a shorter name for `opt_map_left_Some`. It is often needed and very handy but has to be added manually, and it’s a bit tedious to type. Any suggestions?
> > 
> > ```
> > lemma opt_map_left_Some:
> >   “f x = Some y ==> (f |> g) x = g y”
> > ```
> 
> * `opt_reduce`, or...
> * `map_reduce` (given my previous comment about `pred_map`) 🥁

The constant is opt_map, so I would keep that as the prefix. Maybe opt_map_red?

lsf37:
I like opt_pred, but I’m not entirely sure if it should be on the left. It does make sense to have it left, because that’s probably what one wants to focus on. Just want to make sure we have explicitly talked about this. 

Would it make sense to make the priority of the rhs lower so that the projection chain can be written with parentheses?

mktnk3:
> Would it make sense to make the priority of the rhs lower so that the projection chain can be written with parentheses?

You mean changing the priority so that

`bounded_release_time_2 |< (scs_of ||> sc_refills |> hd_opt ||> r_time)`
                                                 ^                                                                  ^
these parentheses around the projection are always required, right?

If so, yes, I would think so. I briefly had that in mind but didn’t get to do it somehow. Tweaking the priory this way would make it easier to read.

lsf37:
> > Would it make sense to make the priority of the rhs lower so that the projection chain can be written with parentheses?
> 
> You mean changing the priority so that, for example,
> 
> `bounded_release_time_2 |< (scs_of ||> sc_refills |> hd_opt ||> r_time)`
> 
> the parentheses around the projection are always required, right?
> 
> If so, yes, I would think so. I briefly had that in mind but didn’t get to do it somehow. Tweaking the priory this way would make it easier to read.

Sorry, I mistyped, I meant *without*. I.e. I wanted to write

```
bounded_release_time_2 |< scs_of ||> sc_refills |> hd_opt ||> r_time
```
(I'm not sure what the current relative priorities, are, I just saw the parentheses in the example)

Looking at it now, I'm not that sure any more if I really want it without :-)

In general, it wouldn't make sense to write `(P |< f) ||> g`, so by default it should parse as `P <| (f |> g)` if we don't write any parentheses. For long chains like above, explicit parentheses do seem a bit clearer, so it probably *would* be nice to have them. So I'm changing my mind :-)

This brings me back to left/right. If we had a symbol to put the predicate at the right, e.g.

```
scs_of ||> sc_refills |> hd_opt ||> r_time ?> bounded_release_time_2
```

the reading flow is clearer and doesn't need parentheses. The attention is now on the rhs, though, which is deeper in the term. Not clear winner either side for me at the moment.

mktnk3:
Ah, ok. It is funny because, currently, we don’t need those parentheses on the RHS. I put them in the example (thinking, “oh, it’s easier to read this way...") but that's not currently how `opt_pred` is defined.

> This brings me back to left/right. If we had a symbol to put the predicate at the right, e.g.
...
> the reading flow is clearer and doesn't need parentheses. The attention is now on the rhs, though, which is deeper in the term. Not clear winner either side for me at the moment.

Having the predicate on the left-most is just like the normal predicates `P x s`, etc., and, often, one can tell from the predicate what parameters it takes, and it probably matters less whether they are applied normally or via `opt` version. This will let us have preconditions that start with meaningful bits of strings. But, I am not fully convinced yet myself.

Maybe people might have a better sense of which they prefer after reviewing? Regarding left/right, and with/without parentheses. If still undecided, we could pick and merge some combination and see how we feel, maybe?

mktnk3:
> The constant is opt_map, so I would keep that as the prefix. Maybe opt_map_red?

That is a fair point. I am happy to settle for `opt_map_red`.

mktnk3:
I wonder what is keeping people from approving this :)

proof/refine/ARM/KHeap_R.thy
```diff
@@ -3917,6 +3914,180 @@ lemma refillSufficient_wp:
   apply (wpsimp wp: getRefills_wp)
   by (clarsimp simp: sufficientRefills_def obj_at'_def)
 
+(* projection rewrites *)
+
+abbreviation
+  opt_pred :: "('a \<Rightarrow> bool) \<Rightarrow> ('b \<Rightarrow> 'a option) \<Rightarrow> ('b \<Rightarrow> bool)" (infixl "|<" 50) where
```
lsf37, line 3920: Could you please move opt-pred to the option mane file (roughly where opt-map is defined)

lsf37 (approved):

lsf37:
> I wonder what is keeping people from approving this :)

The discussion was too interesting to remember the button :-)

mbrcknl:
> I wonder what is keeping people from approving this :)

Sorry, it's been a weirdly distracting week. I'll try to finish my review today!

proof/refine/ARM/SchedContext_R.thy
```diff
@@ -799,4 +799,74 @@ lemmas getSchedContext_setSchedContext_decompose_decompose2
...
+  "\<lbrakk>\<forall>sc. objBits (f sc) = objBits sc; \<forall>sc. objBits (g sc) = objBits sc;
+    \<forall>sc. objBits (h sc) = objBits sc;  \<forall>sc. objBits (k sc) = objBits sc\<rbrakk> \<Longrightarrow>
+    monadic_rewrite False True
+     (sc_at' scPtr)
+     (updateSchedContext scPtr (k o h o g o f))
+     (do updateSchedContext scPtr f;
+         updateSchedContext scPtr g;
+         updateSchedContext scPtr h;
+         updateSchedContext scPtr k
+      od)"
```
mbrcknl, line 843: What's the context for using this rule?

I'm *guessing* that it would be where you have a single update of the form:
```
updateSchedContext scPtr (%sc. sc(| foo := k (foo sc), bar := h (bar sc), etc |))
```
in which case I guess you'd first need to do an ordinary (non-monadic) rewrite to the form:
```
updateSchedContext scPtr (foo_update (K k) o bar_update (K h) o ...)
```
before you can do this `monadic_rewrite`.

If so, then I think it would be nicer to have just a *single* `updateSchedContext_decompose` rule that rewrites a fold into a `mapM_x`, proved by induction:
```
lemma updateSchedContext_decompose:
  "\<forall>sc f. f : set fs --> objBits (f sc) = objBits sc \<Longrightarrow>
   monadic_rewrite False True
     (sc_at' scPtr)
     (updateSchedContext scPtr (fold (o) fs))
     (mapM_x (updateSchedContext scPtr) fs)"
```

Then for the first non-monadic rewrite above, you would instead rewrite to:
```
updateSchedContext scPtr (fold (o) [foo_update (K k), bar_update (K h), ...])
```

Another way to get exactly the `monadic_rewrite` you need from this list-based `updateSchedContext_decompose` would be to use:
```
updateSchedContext_deompose[where fs="[update_foo (K k), update_bar (K h), ...]", simplified]
```
With different length lists, the `fold` and `mapM_x` would expand out to give essentially your three rules, and the composition  of `update_foo` etc. would collapse into the `sc(| ... |)` form.

I haven't attempted to type-check any of this, so I probably got a bunch of details wrong. For example, I might not have the right `fold`, and if you pick the wrong fold, you might need  to `reverse fs` in the `mapM_x`.

mktnk3: Thanks, Matt. These are used in #290 , and the rewrite is done in the opposite direction. This is to bundle up multiple updates on the concrete side to match with a single step update on the abstract side. So, simplification from the record notation to composed updates that you mentioned is not needed there.

But I do agree that these rewrite rules could do with a bit more sophistication, probably along the line of
```
updateSchedContext_deompose[where fs="[update_foo (K k), update_bar (K h), ...]", simplified]
```
I think Michael had something similar, but that was mixed with function extensionality. But I believe there is something in common with what you say here that I can use. What I might do is to move lemmas to #290 and tweak them more there.

mktnk3:
Thanks for the approvals :)

I will update this PR with the following:
- rename `opt_map_left_Some` to `opt_map_red`
- move `opt_pred` to lib (OptionMonad)
- move updateSchedContext rewrite lemmas to #290

proof/refine/ARM/SchedContext_R.thy
mbrcknl: Right, that makes sense. You might still be able to get a modest improvement by proving a `monadic_rewrite` rule with `fold (o)` and `mapM_x`, and getting your three rules with something like:
```
lemmas updateSchedContext_decompose
  = updateSchedContext_decompose_fold[where fs="[f, g]" for f g, simplified]
lemmas updateSchedContext_decompose_twice
  = updateSchedContext_decompose_fold[where fs="[f, g, h]" for f g h, simplified]
lemmas updateSchedContext_decompose_thrice
  = updateSchedContext_decompose_fold[where fs="[f, g, h, k]" for f g h k, simplified]
```

proof/refine/ARM/KHeap_R.thy
```diff
@@ -3933,6 +4104,51 @@ lemma update_sched_context_rewrite:
...
+  and R2: "\<forall>s s'. (s, s') \<in> state_relation \<longrightarrow>
+          P s \<longrightarrow> P' s' \<longrightarrow> sc_at ptr s \<longrightarrow> sc_at' ptr s' \<longrightarrow>
+           heap_ls (replyPrevs_of s')  (scReply (the ((scs_of' s' ||> f') ptr)))
+             (sc_replies (the ((scs_of2 s ||> f) ptr)))"
+  and sz: "\<forall>sc'::sched_context. objBits sc' = objBits (f' sc')"
+  shows "corres dc
+         (sc_at ptr and P)
+         (sc_at' ptr and P')
+            (update_sched_context ptr f)
+            (updateSchedContext ptr f')"
```
mbrcknl, line 4122: Nice!

proof/refine/ARM/SchedContextInv_R.thy
```diff
@@ -640,42 +639,26 @@ lemma updateRefillHd_corres:
...
-         apply (fastforce simp: refills_map_def length_replaceAt)
+         apply (fastforce simp: refills_map_def)
 
         apply (case_tac "i=0")
          apply (clarsimp simp: refills_map_def)
          apply (subst hd_map, fastforce)
          apply (subst hd_wrap_slice; simp)
-         apply (subst nth_map)
-          apply (subst length_wrap_slice; simp?)
-          apply (metis length_replaceAt)
```
mbrcknl, line 655: Do you have a sense of which change allows you to remove all these bits? I'm not complaining, just curious. :-)

mbrcknl (approved):
I'm happy with this. On the question of definitions vs abbreviations, I guess I still have some vague doubts. But you're the one that's been doing all the work, so I'm happy that you're finding abbreviations easier here. I'm not going to stand in the way of progress. :-)

proof/refine/ARM/SchedContextInv_R.thy
mktnk3: It’s one of the consequences of putting `length_replaceAt` in the simplest.

I think it is safe to put `length_xxx` lemmas that prove "function `xxx` preserves the size" in the simp set.

mktnk3:
I will keep the operator priority for `|<` unchanged for now since I feel more undecided on this point. This means no need to put `( )` around the RHS. For the current PRs, there aren’t many big projections appearing on the RHS. Also, I have stacked up a bunch of PRs on top of this at the moment, and it is a bit of work to make the change in a (PR-by-PR) coherent way.

Just a bit of thoughts on this:

- if we have the predicate on the LHS of `|<`, it would be pretty useful to have parentheses around the `opt_map` sequence on the RHS to emphasise that the scope is the whole of the `opt_map` sequence.
- however, it is possible (quite easily) to tell the scope without them if you think of the type, and the `<` and `>` give hints about the direction of the operator. Also projections can often be parentheses heavy, so we might want to have less need for it.
- I would actually prefer it if `|>` applies in the opposite direction, but then we might want a different notation...

mktnk3:
I’ve tweaked the `updateSchedContext` rewrite lemmas. It’s somewhat fiddly and what I have here is probably not the most general thing one can write, perhaps, but at least unifies the two cases that we need.

I’ve renamed `opt_map_left_Some` to `opt_map_red`, but the commit will be added to a different PR (#290) to save some work for cleaning up.

I will squash the last commit and merge this PR later today.

PR #296, opened 2021-06-26
refine: fix regression caused by bad theory import

MitchellBuckley:
Importing Init_R into ADT_H was causing EmptyFail_H to fail. Since
no other theories actually depend on Init_R we can instead include
it in the Refine session directly.

Signed-off-by: Mitchell Buckley <mitchell.alan.buckley@gmail.com>


MitchellBuckley:
https://bamboo.ts.data61.csiro.au/browse/L4V-GHTESTBOARD53-1

lsf37 (approved):
👍

mbrcknl (approved):

PR #297, opened 2021-06-29
valid_refills'

mktnk3:
This PR contains the following changes:

- unify std_sc’ and valid_refills' and rename it as sc_valid_refills'
- define `valid_refills' ptr s'` as `(sc_valid_refills’ |< scs_of’ s’) ptr`  (using `opt_pred`)
- add some lemmas about `valid_refills’`
- update proofs for the change, some cleanup


mktnk3:
This is a small PR that follows the `opt_pred` PR (#295).  It would be great if this is reviewed promptly after #295, so that we can start reviewing #290.

In order for the `sc_refills` part of `sc_relation` to give a valid relation between the abstract and concrete refill lists, some fields of the concrete side scheduling context need to satisfy a set of conditions, which is
```
sc_valid_refills' sc == scRefillMax sc <= length (scRefills sc) & scRefillHead sc < scRefillMax sc
                                      & scRefillCount sc <= scRefillMax sc & 0 < scRefillCount sc
```
Currently these follow from `valid_objs’`, but it is often the case that you need these conditions when you don’t want to be carrying around `valid_objs’` through the proof. We can use `valid_refills’` instead as defined below.
```
definition valid_refills' where
   "valid_refills' sc_ptr s’ == (sc_valid_refills' |< scs_of' s') sc_ptr"
```
There are questions whether this should be a definition or an abbreviation, and whether to keep it as part of `valid_sched_context’` or handle it separately, which we can revisit later. For now, I have this as a definition and will leave `valid_sched_context’` as is.

mbrcknl (approved):
:+1:

proof/refine/ARM/StateRelation.thy
```diff
@@ -252,22 +252,44 @@ definition sc_relation ::
...
-   \<Longrightarrow> valid_refills' sc'"
-  by (clarsimp simp: sc_valid_refills_def valid_sched_context'_def active_sc_def sc_relation_def)
+definition valid_refills' where
+  "valid_refills' sc_ptr s' \<equiv> (sc_valid_refills' |< scs_of' s') sc_ptr"
+
+lemma valid_refills'_nonzero_scRefillCount:
+  "valid_refills' scp s' \<Longrightarrow> ((\<lambda>sc. 0 < scRefillCount sc) |< scs_of' s') scp"
+  by (clarsimp simp: valid_refills'_def split: option.splits)
+
+lemma valid_objs'_valid_refills'':
```
corlewis, line 285: The extra `'` in the name isn't clear to me. Maybe `valid_objs'_valid_refillsE'`? Possibly as an `[elim]` rule as well?

corlewis (approved):

proof/refine/ARM/StateRelation.thy
mktnk3: extra `’` is from the old version and should be removed. Thanks for catching this!

PR #298, opened 2021-06-30
add Sporadic flags

mktnk3:
this adds `scSporadic` as a new field of `sched_context` which will be set via `invokeSchedControl_ConfigureFlags` (which was previously called `invokeSchedControl_Configure`)


mktnk3:
This is the proof update for part of the changes in this commit, which has been squashed with other commits before being merged:
https://github.com/seL4/seL4/commit/afbea157103223f3fae3eab97362d9c2d4053908

mbrcknl (approved):
Looks good!

PR #299, opened 2021-06-30
Cleanup some FIXMEs in ainvs

MitchellBuckley:
This is an attempt to remove some `FIXME`s from ainvs.

* move lots of lemmas up to `lib`.
* replace `cte_wp_at_eqD` with `cte_wp_at_norm` (equal lemmas)
* `pd_shifting_gen` generalises `pd_shifting'` in 2 architectures
* remove some redundant crunch lemmas

I could have kept going with this to make a much larger pull-request, but thought I should check that I'm moving things to the right places.

Testboard here: https://bamboo.ts.data61.csiro.au/browse/L4V-GHTESTBOARD60-1

### DIFF
--- a/lib/LemmaBucket.thy
+++ b/lib/LemmaBucket.thy
@@ -467,6 +467,10 @@ lemma strenghten_False_imp:
   "\<not>P \<Longrightarrow> P \<longrightarrow> Q"
   by blast
 
+lemma None_in_range_Some [simp]:
+  "(None \<in> range Some) = False"
+  by auto
+
 lemma foldl_fun_or_alt:
   "foldl (\<lambda>x y. x \<or> f y) b ls = foldl (\<or>) b (map f ls)"
   apply (induct ls)
@@ -481,5 +485,32 @@ lemma sorted_imp_sorted_filter:
 lemma sorted_list_of_set_already_sorted:
   "\<lbrakk> distinct xs; sorted xs \<rbrakk> \<Longrightarrow> sorted_list_of_set (set xs) = xs"
   by (simp add: sorted_list_of_set_sort_remdups distinct_remdups_id sorted_sort_id)
+
+lemma inj_on_domD: "\<lbrakk>inj_on f (dom f); f x = Some z; f y = Some z\<rbrakk> \<Longrightarrow> x = y"
+  by (erule inj_onD) clarsimp+
+
+lemma not_emptyI:
+  "\<And>x A B. \<lbrakk>x\<in>A; x\<in>B\<rbrakk> \<Longrightarrow> A \<inter> B \<noteq> {}"
+  by auto
+
+lemma add_mask_lower_bits2:
+  "\<lbrakk>is_aligned (x :: 'a :: len word) n; p && ~~ mask n = 0\<rbrakk> \<Longrightarrow> x + p && ~~ mask n = x"
+  apply (subst word_plus_and_or_coroll)
+  apply (simp add: aligned_mask_disjoint and_mask_0_iff_le_mask)
+   apply (clarsimp simp: word_bool_alg.conj_disj_distrib2)
+  done
+
+(* FIXME: move to GenericLib *)
+lemma if3_fold2:
+  "(if P then x else if Q then x else y) = (if P \<or> Q then x else y)" by simp
+
+lemma inter_UNIV_minus[simp]:
+  "x \<inter> (UNIV - y) = x-y" by blast
+
+lemma imp_and_strg: "Q \<and> C \<longrightarrow> (A \<longrightarrow> Q \<and> C) \<and> C" by blast
+
+lemma cases_conj_strg: "A \<and> B \<longrightarrow> (P \<and> A) \<or> (\<not> P \<and> B)" by simp
+
+lemma and_not_not_or_imp: "(~ A & ~ B | C) = ((A | B) \<longrightarrow> C)" by blast
 
 end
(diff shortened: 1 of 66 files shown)


lib/LemmaBucket.thy
```diff
@@ -482,4 +486,44 @@ lemma sorted_list_of_set_already_sorted:
...
+  "\<lbrakk>b \<le> a; size pd - c \<le> a - b; is_aligned pd c \<rbrakk> \<Longrightarrow> pd + (vptr >> a << b) && ~~ mask c = pd"
+  apply (subgoal_tac "(vptr >> a << b) && ~~ mask c = 0")
+   apply (subst word_plus_and_or_coroll)
+    apply (erule aligned_mask_disjoint)
+    apply (simp add: and_mask_0_iff_le_mask[symmetric])
+   apply (simp add: word_bool_alg.conj_disj_distrib2)
+  apply (simp add: shiftr_shiftl1 neg_mask_twice word_bool_alg.conj.assoc)
+  apply (rule shiftr_not_mask_0)
+  apply (fastforce simp: max_def word_size)
+  done
```
Xaphiosis, line 514: This one is a bit specific for lib, but there isn't an ARM*-common area. It might be worth considering which of these could go into Word_Lib, but maybe as a future next step.

Xaphiosis (approved):
Nice work.
I'm a bit torn with the MCS situation on how much cleanup master can take before the merges become an issue, but I think this kind of cleanup is relatively safe.

lib/LemmaBucket.thy
MitchellBuckley: If it's too specific for lib, there's probably somewhere in ainvs that I could find a place for it. I'll try and do that.

And yes, not sure which of these can move up to Word_Lib or similar, but I'm glad you're happy to leave that for a future step.

MitchellBuckley:
> Nice work.
> I'm a bit torn with the MCS situation on how much cleanup master can take before the merges become an issue, but I think this kind of cleanup is relatively safe.

Yeah, good point. Do you have any concerns about impact on MCS @mktnk3 @lsf37 ?

mktnk3:
Hmm. I’d think this kind of clean up makes sense if it contributes to reduce the gap between master and MCS. Doing this only to master is likely to create more work in the future. This is not to stop doing this, but we really need to apply the same to MCS as part of this effort.

The corres rename task came up from MCS because we were naming new corres rules (MCS specific ones) consistently, while the existing ones were not, and that was causing some confusion to some. Victor started it on master for the purpose of staying in sync with master. So, the task is not really complete till it is applied to MCS.

mbrcknl (approved):
The clean-up looks good!

Not sure how to answer the question about merge issues, though. @lsf37 is probably the person who knows most about how well the branches are currently synced, and how to best keep them that way. But he seems quite overloaded at the moment.

mbrcknl:
And yes, I agree with @mktnk3 that this will work best if we make an effort now to sync these changes to both branches.

MitchellBuckley:
Ok, happy to apply parallel changes to MCS. I guess the obvious way would be to cherry-pick the commit and manually check that the changes are in sync. But I'll wait for Gerwin's input before I do that, and open to other ideas.

lsf37 (approved):
This is great! I’m not too worried about the merge impact with MCS on this one. Happy for it to go ahead.

lsf37:
> Ok, happy to apply parallel changes to MCS. I guess the obvious way would be to cherry-pick the commit and manually check that the changes are in sync. But I'll wait for Gerwin's input before I do that, and open to other ideas.

I’d actually wait with that, unless the patch just applies as is (worth trying, but chances are low). Cherry-picking is good when it applies cleanly, but if you need to resolve conflicts during that, you’ll get another set of conflicts from those in the merge, i.e. we’d be doing the work twice. 

I should probably start looking into the next merge of master into rt, just currently a bit busy with setting up GitHub CI, so it’ll be a few weeks until I get there.

mktnk3:
> I should probably start looking into the next merge of master into rt, just currently a bit busy with setting up GitHub CI, so it’ll be a few weeks until I get there.

It’s been about a year since the last merge, and master have changed to Isabelle2020 in the meantime. Do you think you can merge across Isabelle versions? If that’s possible and you think you can manage the changes here fine, I suppose I should leave it with you. Do you think the same for the corres rename as well?

lsf37:
The last master/rt merge was also across Isabelle versions. It's messy, but it's doable. The corres rename is probably also better to defer until after that, yes. At least if I get to it in the next 2 months :-)

mktnk3:
> The last master/rt merge was also across Isabelle versions. It's messy, but it's doable. The corres rename is probably also better to defer until after that, yes. At least if I get to it in the next 2 months :-)

Ah I see. I probably didn’t read the log close enough. I am not too sure about the two months timeframe :), but I suppose I can leave it with you then.

As for the corres renames, I am concerned about those that are already applied to master. Merging should take care of reflecting those in MCS (manually?), I would think? But wouldn’t it easier to run the same script on MCS before merging? But, either way, if merging is something that can happen soonish, we could leave it to revisit at that point maybe.

As for the new MCS ones, we have been writing them “correctly” for a while now, although there may be some older ones left to be updated.

lsf37:
> As for the corres renames, I am concerned about those that are already applied to master. Merging should take care of reflecting those in MCS (manually?), I would think? But wouldn’t it easier to run the same script on MCS before merging? But, either way, if merging is something that can happen soonish, we could leave it to revisit at that point maybe.

It's similar to this PR: if the same rename patch applies cleanly without conflict resolution, applying to to `rt` would be fine and would not create additional work, but if we recreate it manually and slightly differently (or we apply it and have to do conflict resolution), then we generate guaranteed conflicts in the merge. This would mean we're doing application/conflict resolution twice.
 
> As for the new MCS ones, we have been writing them “correctly” for a while now, although there may be some older ones left to be updated.

That's excellent, this should reduce the difference by a good bit.

PR #301, opened 2021-07-02
Support crunching lifted monadic functions

corlewis:
Previously, if `crunch` encountered monadic constants with different state spaces then it would throw a type unification failure due to trying to combine the different preconditions. This meant that any lifted functions had to be explicitly ignored.

This PR changes `crunch` so that it catches and ignores the type unification failure. This allows it to process lifted functions and even prove simple properties about them. However, if the lifted functions require any side conditions for the main property being crunched, then these will not be propagated and the proof of the top-level function will most likely fail.

I do not believe that it is possible to propagate these side conditions without manually passing in a state-lifting function, and I have not yet seen a use case that makes it worth exploring this as a feature.

The concrete example where this was observed involves extended_ops in the abstract state, when you have `('a state, unit) nondet_monad` functions calling lifted `(det_ext state, unit) nondet_monad` functions. This is why there is the slightly convoluted `do_extended_op` test case.

### DIFF
--- a/proof/invariant-abstract/CNodeInv_AI.thy
+++ b/proof/invariant-abstract/CNodeInv_AI.thy
@@ -187,7 +187,7 @@ locale CNodeInv_AI =
         \<Longrightarrow> (ptr, nat_to_cref (zombie_cte_bits zbits) m) \<in> cte_refs (Zombie ptr zbits n) irqn"
   assumes finalise_cap_emptyable[wp]:
     "\<And>sl c f.
-      \<lbrace>emptyable sl and (invs and valid_mdb)\<rbrace>
+      \<lbrace>emptyable sl and invs\<rbrace>
         finalise_cap c f
       \<lbrace>\<lambda>_. emptyable sl :: 'state_ext state \<Rightarrow> bool\<rbrace>"
   assumes deleting_irq_handler_emptyable[wp]:
--- a/lib/crunch-cmd.ML
+++ b/lib/crunch-cmd.ML
@@ -693,17 +693,39 @@ fun proof_failed_warnings const stack cfg wp ctxt t =
   else (warn_const_ignored const cfg ctxt; warn_helper_thms wp ctxt t;
         warn_stack const stack ctxt; all_tac t)
 
+fun collect_preconds (const, mapply, goal) current_goal pre extra ctxt =
+  let
+    val precond = get_inst_precond ctxt pre extra (mapply, goal);
+    val precond' = combine_preconds ctxt (get_precond current_goal) (the_list precond);
+  in Instance.put_precond precond' current_goal end
+  handle ERROR s =>
+    let
+      val default_precond = make_goal (read_const ctxt const) const pre extra ctxt
+        |> get_precond;
+      val goal_precond = Thm.concl_of goal |> HOLogic.dest_Trueprop |> get_precond;
+      val _ =
+        if not (Term.aconv_untyped (goal_precond, default_precond))
+        then (["Failed to collect preconditions from " ^
+               Proof_Context.markup_const ctxt const ^ "; continuing without them.",
+               "This might be due to monadic constants with different state spaces \
+                \being crunched."]
+              |> map Pretty.para |> Pretty.chunks |> Pretty.string_of |> warning;
+              debug_trace s)
+        else ()
+    in current_goal end;
+
 fun crunch cfg pre extra stack const' thms =
   let
     val ctxt = #ctxt cfg |> Context_Position.set_visible false;
     val const = real_const_from_name const' (#nmspce cfg) ctxt;
   in
     let
-      val _ = "crunching constant: " ^ const |> writeln;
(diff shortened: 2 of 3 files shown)


lsf37 (approved):
Nice idea, the more crunch can do the better.

In this case you'll still always get a warning. Would it make sense to change this to info level?

mbrcknl (approved):
I don't know enough about `crunch` internals to really grok how this works, but it kinda makes sense, and the test case is very nice.

Something I haven't been able to figure out is  whether  this is the *only* way that the guts  of `collect_preconds` can fail with `ERROR`. Is there any chance that the message could confuse people if there's another way it can fail?

If so, maybe consider weakening the message to something like "Failed to collect preconditions; continuing without them. Possible reason might be: monadic constants with different state spaces are begin crunched."

corlewis:
> In this case you'll still always get a warning. Would it make sense to change this to info level?

I wanted a warning because otherwise I think it might lead to confusing failures. If there are ever any cases with lifted functions and actual side conditions, then the proof will end up breaking at a later stage with very little indication that it was due to the side condition being dropped. My hope is that it being a warning at least gives a small chance of it being noticed.

It would be nice if this warning was only printed when a side condition is being dropped, but I think it is again almost impossible to do that without a state-lifting function

corlewis:
> Something I haven't been able to figure out is whether this is the _only_ way that the guts of `collect_preconds` can fail with `ERROR`. Is there any chance that the message could confuse people if there's another way it can fail?
> 
> If so, maybe consider weakening the message to something like "Failed to collect preconditions; continuing without them. Possible reason might be: monadic constants with different state spaces are begin crunched."

Good point! I think it is, but `get_inst_precond` and `combine_preconds` are both sufficiently complicated that it's hard to tell exactly what exceptions that they can raise. I'll have a bit of a think, but I'll probably change the warning to something like what you suggested.

lsf37:
> > In this case you'll still always get a warning. Would it make sense to change this to info level?
> 
> I wanted a warning because otherwise I think it might lead to confusing failures. 

What is the difference between info and warning there, though? Just more prominence?

A warning is something the user is supposed to fix, but it doesn't look like this is possible while using the feature.

> If there are ever any cases with lifted functions and actual side conditions, then the proof will end up breaking at a later stage with very little indication that it was due to the side condition being dropped. My hope is that it being a warning at least gives a small chance of it being noticed.

I do agree that it should print something for this case (definitely not just silently drop it), but I think warning is too high a level.

corlewis:
> A warning is something the user is supposed to fix, but it doesn't look like this is possible while using the feature.

I hadn't thought of it this way before, but this part convinced me. However, I also realised that there was a separate problem where a lifted function would cause valid side conditions from non-lifted functions to also get dropped. While fixing this I thought of a way to determine whether preconditions are actually being dropped, which means that I think it now makes sense to use `warning`.

The main change I've made is that `crunch` collects preconditions from one function at a time. This makes it very easy to only drop preconditions from specific functions, while others are collected as normal. Additionally, it also compares the preconditions being dropped with the default preconditions that `crunch` would have tried to prove. If they are equivalent then it's a 'safe' dropping and nothing is printed. However, if they're different then I expect that something really has gone wrong and so print the message as a warning.

It's almost definitely possible to construct an example where a function has preconditions that `crunch` can safely drop, but as far as I have thought of so far this would be sufficiently contrived that I'm ok with not considering it. In particular, I'm pretty convinced that this is not possible in any of our proofs. @lsf37 with this current implementation, are you ok with using `warning`?

proof/invariant-abstract/CNodeInv_AI.thy
```diff
@@ -187,7 +187,7 @@ locale CNodeInv_AI =
         \<Longrightarrow> (ptr, nat_to_cref (zombie_cte_bits zbits) m) \<in> cte_refs (Zombie ptr zbits n) irqn"
   assumes finalise_cap_emptyable[wp]:
     "\<And>sl c f.
-      \<lbrace>emptyable sl and (invs and valid_mdb)\<rbrace>
+      \<lbrace>emptyable sl and invs\<rbrace>
```
corlewis, line 190: As a side benefit of collecting preconditions one at a time, it looks like `crunch` now generates simpler preconditions in some cases.

While collecting preconditions, `crunch` has a very simple test that tries to work out whether the preconditions being added are actually necessary. From memory this is mainly to minimise duplication in the preconditions.

In this case, the test would have previously involved trying to prove `emptyable sl s ⟶ invs s ∧ valid_mdb s` and failing. Now it ends up successfully proving `emptyable sl s ∧ invs s ⟶ valid_mdb s`, and so doesn't collect the `valid_mdb` condition.

lsf37:
> It's almost definitely possible to construct an example where a function has preconditions that `crunch` can safely drop, but as far as I have thought of so far this would be sufficiently contrived that I'm ok with not considering it. In particular, I'm pretty convinced that this is not possible in any of our proofs. @lsf37 with this current implementation, are you ok with using `warning`?

Yes, I think this makes sense now. Happy to go ahead with it!

mbrcknl (approved):
This looks good!

PR #308, opened 2021-07-13
github: complete "switch proof runs to AWS"

lsf37:
This completes the previous commit to run all proof tests on reasonably high-powered AWS VMs instead of GitHub runners. All tests run in one go for efficiency.

This needs seL4/ci-actions#113 merged first.

For now we test on each push to master and rt, and for all pull requests. Since runs for fork-PRs need to be approved for the first run, this should be reasonably safe against abuse.

One full run from scratch costs about USD 2.70 of the free, but limited seL4 foundation AWS credits. Most runs will not be from scratch, but if you don't need the results of a run, it's a good idea to cancel. Assuming that, we can do a few thousand test per year, which should be plenty, but we'll monitor cost closely for the first few months.

I'm planning to make the logs a bit prettier, and potentially to upload the full Isabelle logs as build artefacts, but that is a separate step.

### DIFF
--- a/.github/workflows/proof.yml
+++ b/.github/workflows/proof.yml
@@ -9,20 +9,25 @@ on:
     branches:
       - master
       - rt
-  pull_request:
+  # this action needs access to secrets.
+  # The actual test runs in a no-privilege VM, so it's Ok to run on untrusted PRs.
+  pull_request_target:
 
 jobs:
-  aspec:
-    name: ASpec
+  all:
+    name: All
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


lsf37:
The current AWS limits are set to 16 concurrent VMs, i.e. 4 concurrent runs of 4 architectures (master) or 8 concurrent runs for PRs on RT (2 architectures) or any combination of these that gets us <= 16 VMs.

The action will fail somewhat gracefully when that limit is reached, and it is safe to just go on "Rerun jobs" to try again.

lsf37:
Turns out logs and prettier GH output were easy and smallish, so I added them here.

mbrcknl (approved):

PR #309, opened 2021-07-14
run rt proofs on AWS

lsf37:
This mirrors the corresponding commit on the master branch. We'll unfortunately only see if everything works properly after this is merged, but it seems to be going fine on `master` and PRs there.

The benefit is (should be) that the tests for MCS and PRs to the rt branch will be complete and run faster than on GitHub or Bamboo.

### DIFF
--- a/.github/workflows/proof.yml
+++ b/.github/workflows/proof.yml
@@ -1,4 +1,4 @@
-# Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+# Copyright 2021, Proofcraft Pty Ltd
 #
 # SPDX-License-Identifier: BSD-2-Clause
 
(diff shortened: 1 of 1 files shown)


mbrcknl (approved):

mktnk3 (approved):
By “complete”, do you mean running CRefine as well? Do we want it at this point? I suppose we can ignore CRefine and merge if necessary.

lsf37:
> By “complete”, do you mean running CRefine as well? Do we want it at this point? I suppose we can ignore CRefine and merge if necessary.

With "complete" I meant everything that is enabled in `run_tests` in the branch. Before, which tests were run was duplicated in the github action setup, now it picks up `run_tests` and runs all available sessions by default. I think this currently excludes CRefine, but includes CSpec for `rt`, which sounds about right (same as Bamboo does).

mktnk3:
It might include CBaseRefine. I am not sure, actually.
As for the tests on the PR, all the sessions now appear bundled up as `Proofs (ARM)` for each arch. I guess we can open the log to see the progress.

lsf37:
> It might include CBaseRefine. I am not sure, actually.

I think it might, yes, we'll see what it does. I should switch on `SKIP_DUPLICATED_PROOFS`, actually. That should speed up `CBaseRefine` significantly.

> As for the tests on the PR, all the sessions now appear bundled up as Proofs / MCS (ARM) for each arch. I guess we can open the log to see the progress.

Yes, that's the plan. Starting a new VM for each sub-test was too many VMs, and it turns out that GitHub has limit of 20 concurrent jobs for all tests in the repo, so having 4 (or here 2) instead of 4 * main sessions helps run more proofs.

If it becomes too annoying to have to look into the logs, I might look at whether the VM can report test status to the github checks API directly as well, but let's see how it is in practice first.

mktnk3:
Ok, I see. Yes, let’s see. As you probably are aware, #298 is currently running.

lsf37:
:+1: (subsequent runs should be faster, because it saves the images between runs)

mktnk3:
Hmm. CBaseRefine is run only for ARM, but it failed (timed out in Tcb_R). Refine has passed fine, though. Setting `SKIP_DUPLICATED_PROOFS` would be good, probably.

Also, both ARM and RISCV failed with archiving the logs, it seems. Something to do with the path, it looks like.

Looking at the log as the tests are running might be not as nice as it could be. It seems you can see the log lines as they are generated but can’t see the older lines till the tests are all done. Or maybe it is different if there is a failure, I am not sure. Anyway, I suppose this is how tests on GitHub have always been. I am sure we can live with what we have now in terms of seeing the log on the PR.

lsf37:
> Hmm. CBaseRefine is run only for ARM, but it failed (timed out in Tcb_R). Refine has passed fine, though. Setting SKIP_DUPLICATED_PROOFS would be good, probably.

Yes, will add that this afternoon. It does mean we have problem with CBaseRefine, though, because it does need to pass -- the only effect of setting the option should be time, not failure/non-failure. Probably just need to increase the timeout.

> Also, both ARM and RISCV failed with archiving the logs, it seems. Something to do with the path, it looks like.

It looks like the logs path is different between Isabelle2019 and Isabelle2020, which is a bit annoying, but will fix.

> Looking at the log as the tests are running might be not as nice as it could be. It seems you can see the log lines as they are generated but can’t see the older lines till the tests are all done. Or maybe it is different if there is a failure, I am not sure. Anyway, I suppose this is how tests on GitHub have always been. I am sure we can live with what we have now in terms of seeing the log on the PR.

It does unfortunately seem to be the way GitHub processes logs (while the test is running, you only get to see new ones). I'll look into that separate reporting option when I'm done with the more urgent tests, so we can at least see which sessions have already passed/failed.

I do have a PR in preparation that will make the logs easier to read at the end of a run.

mktnk3:
> > Hmm. CBaseRefine is run only for ARM, but it failed (timed out in Tcb_R). Refine has passed fine, though. Setting SKIP_DUPLICATED_PROOFS would be good, probably.
> 
> Yes, will add that this afternoon. It does mean we have problem with CBaseRefine, though, because it does need to pass -- the only effect of setting the option should be time, not failure/non-failure. Probably just need to increase the timeout.

The issue should be just the time. The PR doesn’t touch Tcb_R and shouldn’t have any impact on the proof there. CBaseRefine has been passing fine on Bamboo so far.

I thought we can merge the PR without having green for the proof tests. At least, I think we have been merging PRs while the tests are still running sometimes (after getting green results previously somewhere).

Having said that, I am happy to wait to merge #298 for further test runs.

Improvements around reporting/logging when you have time for it would be great. No need to rush.

lsf37:
Happy for the PR to merge, the tests as such all pass. It's just the framework that needs a bit more work.

> CBaseRefine has been passing fine on Bamboo so far.

For RT, Bamboo is only testing with SKIP_DUPLICATED_PROOFS on, so we'd not have noticed on Bamboo. I think only the master branch gets tested once a week without it.

Let's increase the timeout and see if it passes then.

mktnk3:
Great, the PR passes all the tests green now :) Will rebase and merge soon.

PR #310, opened 2021-07-15
More readable proof logs

lsf37:
This switches on colour for GitHub and makes sure the proof output is actually placed into the corresponding log groups.

### DIFF
--- a/misc/regression/run_tests.py
+++ b/misc/regression/run_tests.py
@@ -47,9 +47,13 @@ ANSI_WHITE = "\033[37m"
 ANSI_BOLD = "\033[1m"
 
 
-def output_color(color, s):
+def running_on_github():
+    return os.environ.get("GITHUB_REPOSITORY") != None
+
+
+def output_color(color, s, github=running_on_github()):
     """Wrap the given string in the given color."""
-    if sys.stdout.isatty():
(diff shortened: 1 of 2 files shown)


misc/regression/run_tests.py
```diff
@@ -47,9 +47,13 @@
 ANSI_BOLD = "\033[1m"
 
 
-def output_color(color, s):
+def running_on_github():
+    return os.environ.get("GITHUB_REPOSITORY") != None
+
+
+def output_color(color, s, github=running_on_github()):
```
mbrcknl, line 54: I think this will call `running_on_github()` once when the function is parsed, not when it's called. This is fine, because the result ought to be always the same. But in that case, you might as well just make it a global variable.

mbrcknl (approved):
Looks good!

misc/regression/run_tests.py
mbrcknl: If you do want it to be more dynamic, I think you need to use the `None` idiom:

```
def output_color(color, s, github=None):
    if github is None:
        github = running_on_github()
```

lsf37: yes, I think this is run at definition time only. I had it as a global variable in the beginning, but found that also somewhat ugly.

PR #312, opened 2021-07-16
abstract x64+riscv: clean up some Word_Lib imports

Xaphiosis:
These are already imported upstream.


lsf37 (approved):
Let's see if all images build, but if yes, definitely a good idea to remove.

PR #315, opened 2021-07-29
AInvs projections

mktnk3:
The main point of this PR is to change the existing AInvs DetSched projections (`tcb_of`, `sc_of` and `ep_of`) to from “definition” to “primrec”.

It also adds `ntfn_of` and `reply_of`, but they are not used yet.

Defining them as primrec seems to work well in terms of simplification and often results in simpler codes. The proof update for this change resulted in slight clean-up overall.

---
So far, tcb_of and sc_of in AInvs are case statements on kernel objects
defined as definitions, and they needed to be explicitly unfolded to
simplify to an option value.

Making them abbreviations is one way to automate the simplification, but
it has a drawback in that the case statement often has unfortunate
interaction with Map.empty, which is an abbreviation of its own, that
tends to produce unwantedly expanded case expressions in subgoals.

It is probably possible to automate the simplication of definitions
by writing a simplification rule for each kernel object case and add
all of such rules to the simp set, but defining these projections as
primrec seems to work just as well, and also often results in shorter
proofs as we can skip case distinctions which would be needed otherwise.

### DIFF
--- a/proof/invariant-abstract/DetSchedInvs_AI.thy
+++ b/proof/invariant-abstract/DetSchedInvs_AI.thy
@@ -584,23 +584,15 @@ end
 definition tcb_of :: "kernel_object \<rightharpoonup> tcb" where
   "tcb_of ko \<equiv> case ko of TCB tcb \<Rightarrow> Some tcb | _ \<Rightarrow> None"
 
-lemma tcb_of_simps:
-  "tcb_of (TCB tcb) = Some tcb"
-  "tcb_of (SchedContext sc n) = None"
-  "tcb_of (CNode n cnode) = None"
-  "tcb_of (Endpoint ep) = None"
-  "tcb_of (Notification ntfn) = None"
-  "tcb_of (Reply reply) = None"
-  "tcb_of (ArchObj aobj) = None"
-  by (auto simp: tcb_of_def)
+lemmas tcb_of_simps [simp] = tcb_of_def [split_simps kernel_object.split]
 
-lemma tcb_of_Some:
+lemma tcb_of_Some[simp]:
   "tcb_of ko = Some tcb \<longleftrightarrow> ko = TCB tcb"
-  by (simp add: tcb_of_def split: kernel_object.splits)
+  by (cases ko; simp)
 
 lemma tcb_of_None:
   "tcb_of ko = None \<longleftrightarrow> (\<forall>tcb. ko \<noteq> TCB tcb)"
-  by (simp add: tcb_of_def split: kernel_object.splits)
+  by (cases ko; simp)
 
 definition tcbs_of_kh :: "('obj_ref \<rightharpoonup> kernel_object) \<Rightarrow> 'obj_ref \<rightharpoonup> tcb" where
   "tcbs_of_kh kh \<equiv> kh |> tcb_of"
@@ -721,25 +713,17 @@ lemma etcb_at'_pred_map:
 \<comment> \<open>Project scheduling contexts from the kernel heap\<close>
 
 definition sc_of :: "kernel_object \<rightharpoonup> sched_context" where
-  "sc_of ko \<equiv> case ko of SchedContext sc n \<Rightarrow> Some sc | _ \<Rightarrow> None"
+  "sc_of ko \<equiv> case ko of SchedContext sc _ \<Rightarrow> Some sc | _ \<Rightarrow> None"
 
-lemma sc_of_simps:
-  "sc_of (SchedContext sc n) = Some sc"
-  "sc_of (CNode n cnode) = None"
-  "sc_of (TCB tcb) = None"
-  "sc_of (Endpoint ep) = None"
-  "sc_of (Notification ntfn) = None"
-  "sc_of (Reply reply) = None"
-  "sc_of (ArchObj aobj) = None"
-  by (auto simp: sc_of_def)
+lemmas sc_of_simps [simp] = sc_of_def [split_simps kernel_object.split]
 
-lemma sc_of_Some:
+lemma sc_of_Some[simp]:
   "sc_of ko = Some sc \<longleftrightarrow> (\<exists>n. ko = SchedContext sc n)"
-  by (simp add: sc_of_def split: kernel_object.splits)
+ by (cases ko; simp)
 
 lemma sc_of_None:
   "sc_of ko = None \<longleftrightarrow> (\<forall>sc n. ko \<noteq> SchedContext sc n)"
-  by (simp add: sc_of_def split: kernel_object.splits)
+ by (cases ko; simp)
 
 definition scs_of_kh :: "('obj_ref \<rightharpoonup> kernel_object) \<Rightarrow> 'obj_ref \<rightharpoonup> sched_context" where
   "scs_of_kh kh \<equiv> kh |> sc_of"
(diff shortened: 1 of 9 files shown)


mktnk3:
I find this `Map.empty`, which is given in “src/HOL/Map.thy” as
```
abbreviation
  empty :: "'a \<rightharpoonup> 'b" where
  "empty \<equiv> \<lambda>x. None"
```
quite invasive. Is this how it is meant to be? Anything that is defined with “... = None” matches this abbreviation, it seems. Maybe making it input only would be better? Or is this a known thing?

lsf37:
> Is this how it is meant to be? Anything that is defined with “... = None” matches this abbreviation, it seems. Maybe making it input only would be better? Or is this a known thing?

I would have thought it applies only if the lambda is present, but I might be wrong. With lambda present, it is as intended. Without it's a problem. There's not really that much we can do about it, though. We could override, but overrides of things in `Main` will be overwritten in any theory merge that doesn't itself include the override, so these tend to come back.

lsf37:
> The main point of this PR is to change the existing AInvs DetSched projections (tcb_of, sc_of and ep_of) to from “definition” to “primrec”.

I don't think this part of it is a good idea. We actually do have a standard strategy for this: use a `definition` instead of `fun`/`primrec`, so that we get the case split rules (and don't have to always explicitly do `cases`), and use something like

```isabelle
lemmas tcb_of_simps [simp] = tcb_of_def [split_simps kernel_object.splits]
```

to get the automatic simplification. We don't always want automatic simplification, but for `tcbs_of` etc it makes sense.

proof/invariant-abstract/DetSchedInvs_AI.thy
```diff
@@ -581,26 +581,30 @@ end
 
 \<comment> \<open>Project threads from the kernel heap\<close>
 
-definition tcb_of :: "kernel_object \<rightharpoonup> tcb" where
```
lsf37, line 584: See my other comment for this -- we should leave it as `definition` with `case`, and use `[split_simps ..]` to get the simp rules.

proof/invariant-abstract/DetSchedInvs_AI.thy
```diff
@@ -581,26 +581,30 @@ end
...
-  "tcb_of (ArchObj aobj) = None"
-  by (auto simp: tcb_of_def)
+| "tcb_of (Endpoint _) = None"
+| "tcb_of (Notification _) = None"
+| "tcb_of (CNode _ _) = None"
+| "tcb_of (Reply _) = None"
+| "tcb_of (SchedContext _ _) = None"
+| "tcb_of (ArchObj _) = None"
+
+lemma tcb_of_SomeD[dest!]:
```
lsf37, line 593: Not sure any more, I might have written the `tcb_of_Some` version below, but I think this could be an equality instead of a `==>`. With an equality, this could then be `[simp]`, and we can probably remove `tcb_of_Some`.

proof/invariant-abstract/DetSchedInvs_AI.thy
```diff
@@ -581,26 +581,30 @@ end
...
+  "tcb_of ko = Some obj \<Longrightarrow> ko = TCB obj"
+  by (cases ko; simp)
+
+lemma tcb_of_NoneD[simp]:
```
lsf37, line 597: I don't think this is a good simp rule, the conclusion is not in simp normal form and the left-hand side might often be not true, so if the preprocessed conclusion fits often, this will be applied often and fail often, i.e. slow down proofs where it doesn't work.

But: if we can make it an equality instead of `==>` similar to my comment on `tcb_of_SomeD`, then `tcb_of ko = None` becomes the lhs of the rewrite rule, which would be perfectly nice for a `[simp]` rule.

proof/invariant-abstract/DetSchedInvs_AI.thy
```diff
@@ -720,26 +724,30 @@ lemma etcb_at'_pred_map:
 
 \<comment> \<open>Project scheduling contexts from the kernel heap\<close>
 
-definition sc_of :: "kernel_object \<rightharpoonup> sched_context" where
```
lsf37, line 723: See `definition` comment

proof/invariant-abstract/DetSchedInvs_AI.thy
```diff
@@ -835,54 +843,47 @@ global_interpretation sc_replies:
 
 \<comment> \<open>Project endpoints from the kernel heap\<close>
 
-definition ep_of :: "kernel_object \<rightharpoonup> endpoint" where
```
lsf37, line 838: See `definition` comment for these as well.

proof/invariant-abstract/DetSchedInvs_AI.thy
```diff
@@ -916,6 +917,54 @@ global_interpretation ep_recv_qs:
   opt_map_opt_map_cons_def_locale _ ep_of eps_of_kh Endpoint recv_q_of RecvEP ep_recv_qs_of_eps
   using ep_recv_qs_of_eps_def recv_q_of_None recv_q_of_Some by unfold_locales
 
+\<comment> \<open>Notifications\<close>
+
+\<comment> \<open>Project notifications from the kernel heap\<close>
+
+primrec ntfn_of :: "Structures_A.kernel_object \<rightharpoonup> notification" where
```
lsf37, line 924: It's a good idea to add these, but also as definitions, and with equality setup for the automation lemmas (if that works, I haven't tried out if equality is actually possible)

proof/invariant-abstract/DetSchedInvs_AI.thy
```diff
@@ -916,6 +917,54 @@ global_interpretation ep_recv_qs:
...
+  by (cases ko; simp)
+
+abbreviation ntfns_of :: "'z state \<Rightarrow> obj_ref \<rightharpoonup> Structures_A.notification" where
+  "ntfns_of \<equiv> (\<lambda>s. kheap s |> ntfn_of)"
+
+\<comment> \<open>Replies\<close>
+
+\<comment> \<open>Project replies from the kernel heap\<close>
+
+primrec reply_of :: "Structures_A.kernel_object \<rightharpoonup> reply" where
```
lsf37, line 948: Same here (`definition` comment). Just marking all of them so you don't have to search.

mktnk3:
> > The main point of this PR is to change the existing AInvs DetSched projections (tcb_of, sc_of and ep_of) to from “definition” to “primrec”.
> 
> I don't think this part of it is a good idea. We actually do have a standard strategy for this: use a `definition` instead of `fun`/`primrec`, so that we get the case split rules (and don't have to always explicitly do `cases`), and use something like
> 
> ```isabelle-theory
> lemmas tcb_of_simps [simp] = tcb_of_def [split_simps kernel_object.splits]
> ```
> 
> to get the automatic simplification. We don't always want automatic simplification, but for `tcbs_of` etc it makes sense.

Ah, Ok. Great, that’s the kind of feedback I was hoping for. I can certainly set things up in that way instead. Is this “standard strategy” shared somewhere so that people can refer to?

proof/invariant-abstract/DetSchedInvs_AI.thy
```diff
@@ -948,9 +997,6 @@ lemmas vs_pred_map_simps =
   pred_map_def[where m="kheap s" for s :: "'z state"]
 
 lemmas vs_heap_simps =
-  tcb_of_simps
```
lsf37, line 951: It's fine to leave them removed here if they are declared `[simp]` at the point we generate them with `[split_simps ..]`

proof/invariant-abstract/DetSchedInvs_AI.thy
```diff
@@ -581,26 +581,30 @@ end
 
 \<comment> \<open>Project threads from the kernel heap\<close>
 
-definition tcb_of :: "kernel_object \<rightharpoonup> tcb" where
-  "tcb_of ko \<equiv> case ko of TCB tcb \<Rightarrow> Some tcb | _ \<Rightarrow> None"
-
-lemma tcb_of_simps:
+primrec tcb_of :: "Structures_A.kernel_object \<rightharpoonup> tcb" where
```
lsf37, line 584: In general inside `_AI` theories, we should leave off qualifiers for ASpec, i.e. no `Structures_A.`. ASpec should be the default name space in AInvs, and if it is not for some reason, things should break as immediately as possible.

lsf37 (changes requested):
The effect of this is quite nice, i.e. the proof gains shouldn't change. I'd just change the implementation to use `[split_simps ..]` and investigate if we can make the `_SomeD` rules equalities.

lsf37:
> Ah, Ok. Great, that’s the kind of feedback I was hoping for. I can certainly set things up in that way instead. 

👍 

> Is this “standard strategy” shared somewhere so that people can refer to?

I thought it was, but it's not in the repo. It might be in one of the old wiki pages, but we should add it under `docs` here. I can do that in a separate PR if you like.

lsf37:
It might actually be better to move all these `tcbs_of` etc definitions into ASpec even, because they can be used there in the future as well. Happy for that to come later, though, if you want to finish this first.

mktnk3:
> I would have thought it applies only if the lambda is present, but I might be wrong. With lambda present, it is as intended. Without it's a problem. There's not really that much we can do about it, though. We could override, but overrides of things in `Main` will be overwritten in any theory merge that doesn't itself include the override, so these tend to come back.

It is probably due to the combination with eta equivalence and function composition (of projections). I guess we just have to live with it, I see what you say about overriding what is in Main.

mktnk3:
> It might actually be better to move all these `tcbs_of` etc definitions into ASpec even, because they can be used there in the future as well. Happy for that to come later, though, if you want to finish this first.

I had a similar thought when I changed the PR name from DetSched to AInvs. I will see how things go, but I will move these to ASpec now/soon.

proof/invariant-abstract/DetSchedInvs_AI.thy
mktnk3: Yes, I should have removed these qualifiers. These definitions are initially used while working in Refine, and that’s where they are coming from.

lsf37:
> It is probably due to the combination with eta equivalence and function composition (of projections). 

Yes, that makes sense, that could produce the pattern. We could see if we could add another abbreviation that merges this with some other part of the projection chain, i.e. `empty |> ..`, but it might not be necessary. Come to think of it, we should probably have a `simp` rule for projection chains that have `empty` in them, which should shorten them to just `empty` or something similar.

> I guess we just have to live with it, I see what you say about overriding what is in Main.

If we go back to definitions, the impact might be less than before (although simp will still produce them).

proof/invariant-abstract/DetSchedInvs_AI.thy
mktnk3: Yes, there are equality versions below and I didn't need to copy these ones across. I can add the equalities to the simp set.

proof/invariant-abstract/DetSchedInvs_AI.thy
mktnk3: This might not quite work as an automatic simp rule, I am afraid. For the `Some` case, the equality in the conclusion allows us to retrieve the equality in the LHS, so we don’t lose anything. But for this `None` case, we won’t be able to use the LHS equality once this simp rule triggers. This leaves us with stray `tcb_of ko` in other expressions which we’d want to rewrite to `None`.

lsf37: Hm, yes, you're right. There's no good mechanism for saying this should only be applied when nothing else works. Still not sure it should be globally `simp`. Do you have an example goal where this rule applies? (Just trying to figure out how the pattern works in practise)

mktnk3: I now pushed a version with all the `xxx_of_None` in the simp set. There are two `auto` proofs in DetSched files that get broken by having `tcb_of_None` in the simp set.The subgoal for one of them is like this (it's quite messy):
```
 ⟦P (the (map_option (r_time ∘ hd ∘ scrc_refills)
                 (case case tcb_of x2 of None ⇒ None
                       | Some x ⇒ (id ∘ tcb_sched_context) x of
                  None ⇒ None
                  | Some x ⇒
                      map_project sc_refill_cfg_of
                       (λsa. case kheap s sa of None ⇒ None
                             | Some x ⇒ sc_of x)
                       x)));
        ∀x. sc_refills (f x) = sc_refills x;
        kheap s sc_ptr = Some (SchedContext x xa);
        a_type (SchedContext x xa) = a_type (SchedContext (f x) xa);
        kheap s t = Some x2; ∀tcb. x2 ≠ TCB tcb⟧
       ⟹ P (the None)
       ```
It's only `tcb_of_None` causing issues, but it is probably because others aren't used as much. I didn't fix Refine, so we might see more breakage there.

**Comment by mktnk3** [comment]
> It might actually be better to move all these `tcbs_of` etc definitions into ASpec even, because they can be used there in the future as well. Happy for that to come later, though, if you want to finish this first.

For this, I am thinking of adding a few more changes to these basic projections, so I will wait to go through them before moving the whole lot to ASpec.

File: proof/invariant-abstract/DetSchedInvs_AI.thy (hunk 3 above)
Comment (mktnk3) [reply proof/invariant-abstract/DetSchedInvs_AI.thy]: I have a version that I haven’t pushed that has the equalities for `Some` cases in the simp set while those for `None` not in the simp set, and this seems to work well (doesn’t break any proof). I would be happy if we can settle for this version.

Taking the `Some` equalities out of the simp set does break the proof (`sc_of_Some` is needed in many places in Refine).

Comment (mktnk3) [reply proof/invariant-abstract/DetSchedInvs_AI.thy]: I pushed the above version as a separate commit. If there is no further comments, I would like an approval so that I can merge this soon. I will squash and update the commit messages before merging.

Comment (lsf37) [reply proof/invariant-abstract/DetSchedInvs_AI.thy]: I'm happy with this part. Still need to look through the rest. Should be done soon, though.

File: proof/invariant-abstract/DetSchedInvs_AI.thy [hunk 10]
```diff
@@ -584,23 +584,15 @@ end
...
-  "tcb_of (Notification ntfn) = None"
-  "tcb_of (Reply reply) = None"
-  "tcb_of (ArchObj aobj) = None"
-  by (auto simp: tcb_of_def)
-
-lemma tcb_of_Some:
+lemmas tcb_of_simps [simp] = tcb_of_def [split_simps kernel_object.split]
+
+lemma tcb_of_Some[simp]:
   "tcb_of ko = Some tcb \<longleftrightarrow> ko = TCB tcb"
```
lsf37, line 590: I don't think this one is right yet. What I meant was that we should have something like
```
(tcb_of ko = Some tcb) = (ko = TCB tcb)
```
in the simpset (which may already exist later, not sure). But we should not have `tcb_of_SomeD` in the simp set, because that would apply to any term of type `kernel_object` (it'll fail for most of them, but it will slow down all other proofs). If we have the rule above in the simpset, then `tcb_of_SomeD` is just that rule with `[THEN iffD1]`, which we should either prove that way if we actually need it as a dest rule, or leave out completely.

mktnk3: This is iff and this is the version that you added?

mktnk3: This is the pre-existing one and I didn't change the statement. I just added it to the simp set  (I had to change the proof, though).

lsf37: I can't find that rule, did I already add it? In any case, it should be added if it doesn't exist yet. If it's already there somewhere, it'd be good to move it close to the definition.

`[iff]` is probably safe for that one, yes, which means we can leave out the current form `tcb_of_Some` because that should be generated as part of `iff`.

mktnk3: What I am saying is that the statement has "longleftrightarrow". It is an implication in both ways.

lsf37: Oh dear, my brain so did not see that, sorry, I kept seeing "longrightarrow". Thanks for pointing that out. And then everything is fine as is. We might make it `iff`, but it's not a priority.

lsf37 (approved):

mktnk3:
I think there is something about the precedence - people might choose to use <--> over = because = asks for more parentheses, maybe :wink:  :eyeglasses:

lsf37:
> I think there is something about the precedence - people might choose to use <--> over = because = asks for more parentheses, maybe 😉 👓

Yes, indeed. I really should try to get Scott's symbol script for github diffs working in Safari, then it would have been obvious. Sorry for wasting so much time..

PR #316, opened 2021-07-30
rt ainvs: valid_ready_qs substatement and cleanup

mktnk3:
`valid_ready_qs` is a statement about a thread in a ready_queue that
has three components: one is about the ready queue matching the domain
and the priority of the thread, and the others are the thread being in a
runnable state and has a released SC.

- this commit adds an abbreviation, `ready_queues_etcb_eq`, which states the validity of the first component only, i.e., "if in a ready_queue, it must be in the correct queue (corresponding to its domain and priority)"

- `etcb_eq` is used for better compatibility with current `valid_ready_qs`, for the time being

- `valid_ready_qs` preconditions in some wp rules are replaced with this, as they were too generous; this leads to some simplifications in proofs

- most notable one is the `not_queued` wp rules for `tcb_sched_dequeue`; it removes a thread from its "correct" queue, but we need to assume that that is the only queue possible. Previously, `valid_ready_qs` was used for this purpose.

- a few other cleanups added along the way

This is a small side work of "add `refill_unblock_check`" spec change.

### DIFF
--- a/proof/invariant-abstract/DetSchedSchedule_AI.thy
+++ b/proof/invariant-abstract/DetSchedSchedule_AI.thy
@@ -7120,16 +7120,11 @@ lemma no_ofail_refill_head_overlapping:
   apply (clarsimp simp: obind_def obj_at_def read_sched_context_def)
   done
 
-lemma bound_refill_head_overlapping:
-  "kheap s sc_ptr = Some (SchedContext sc n) \<Longrightarrow> bound (refill_head_overlapping sc_ptr s)"
-  apply (clarsimp simp: refill_head_overlapping_def obind_def read_sched_context_def)
-  done
-
 lemma refill_head_overlapping_true_imp_length_at_least_two:
   "\<lbrakk>the (refill_head_overlapping sc_ptr s); pred_map \<top> (sc_refill_cfgs_of s) sc_ptr\<rbrakk>
    \<Longrightarrow> pred_map (\<lambda>cfg. Suc 0 < length (scrc_refills cfg)) (sc_refill_cfgs_of s) sc_ptr"
   apply (clarsimp simp: vs_all_heap_simps)
-  apply (frule bound_refill_head_overlapping)
+  apply (insert no_ofailD[OF no_ofail_refill_head_overlapping])
   apply (fastforce simp: refill_head_overlapping_def vs_all_heap_simps read_sched_context_def)
   done
 
@@ -7139,7 +7134,7 @@ lemma refill_head_overlapping_refills_overlapping:
                        \<le> r_time (hd (scrc_refills cfg)) + r_amount (hd (scrc_refills cfg)))
                  (sc_refill_cfgs_of s) sc_ptr"
   apply (clarsimp simp: vs_all_heap_simps)
-  apply (frule bound_refill_head_overlapping)
+  apply (insert no_ofailD[OF no_ofail_refill_head_overlapping])
   apply (clarsimp simp: refill_head_overlapping_def obind_def read_sched_context_def)
   done
 
@@ -7150,7 +7145,7 @@ lemma refill_head_overlapping_refills_not_overlapping:
                          \<le> r_time (hd (scrc_refills cfg)) + r_amount (hd (scrc_refills cfg))))
                   (sc_refill_cfgs_of s) sc_ptr"
   apply (clarsimp simp: vs_all_heap_simps)
-  apply (frule bound_refill_head_overlapping)
+  apply (insert no_ofailD[OF no_ofail_refill_head_overlapping])
   apply (clarsimp simp: refill_head_overlapping_def obind_def read_sched_context_def)
   done
 
--- a/proof/invariant-abstract/DetSchedSchedule_AI.thy
+++ b/proof/invariant-abstract/DetSchedSchedule_AI.thy
@@ -1460,11 +1460,20 @@ lemma set_tcb_sched_context_valid_sched_pred':
 lemmas set_tcb_sched_context_valid_sched_pred[valid_sched_wp] =
   set_tcb_sched_context_valid_sched_pred'[THEN hoare_drop_assertion]
 
-lemma set_tcb_sched_context_valid_ready_qs_not_queued:
-  "\<lbrace>valid_ready_qs and not_queued ref\<rbrace>
+lemma set_tcb_sched_context_valid_ready_qs:
+  "\<lbrace>valid_ready_qs and (\<lambda>s. in_ready_q ref s \<longrightarrow> (\<exists>p. sc = Some p \<and> released_sc_at p s))\<rbrace>
    set_tcb_obj_ref tcb_sched_context_update ref sc
    \<lbrace>\<lambda>_. valid_ready_qs\<rbrace>"
-  by (wpsimp wp: valid_sched_wp simp: valid_ready_qs_def vs_all_heap_simps not_queued_def)
+  unfolding set_tcb_obj_ref_def
+  apply simp
+  apply (rule hoare_seq_ext[OF _ gets_the_sp])
+  apply (wpsimp simp: vs_all_heap_simps obj_at_def set_object_def
+                  wp: get_object_wp split: option.splits)
+  apply (cases sc; simp)
+   apply (clarsimp simp: valid_ready_qs_def vs_all_heap_simps not_queued_def dest!: get_tcb_SomeD)+
+  apply (drule_tac x=d and y=p in spec2, clarsimp, drule_tac x=ref in bspec, simp, clarsimp)
+  apply (clarsimp simp: in_queues_2_def)
+  done
 
 lemma set_tcb_sched_context_None_released_ipc_queues:
   "\<lbrace>released_ipc_queues\<rbrace>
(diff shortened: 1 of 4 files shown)


proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -1460,11 +1460,20 @@ lemma set_tcb_sched_context_valid_sched_pred':
...
    set_tcb_obj_ref tcb_sched_context_update ref sc
    \<lbrace>\<lambda>_. valid_ready_qs\<rbrace>"
-  by (wpsimp wp: valid_sched_wp simp: valid_ready_qs_def vs_all_heap_simps not_queued_def)
+  unfolding set_tcb_obj_ref_def
+  apply simp
+  apply (rule hoare_seq_ext[OF _ gets_the_sp])
+  apply (wpsimp simp: vs_all_heap_simps obj_at_def set_object_def
+                  wp: get_object_wp split: option.splits)
+  apply (cases sc; simp)
+   apply (clarsimp simp: valid_ready_qs_def vs_all_heap_simps not_queued_def dest!: get_tcb_SomeD)+
```
michaelmcinerney, line 1473: nitpick: is the `+` necessary here?

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -1502,30 +1504,22 @@ lemma valid_release_q_no_sc_not_in_release_q:
...
-  apply (wpsimp wp: valid_sched_wp)
-  apply (clarsimp simp: valid_release_q_def vs_all_heap_simps in_release_q_def sc_ready_time_eq_iff
-                 elim!: sorted_release_qE)
-  by (auto simp: tcb_ready_times_defs map_project_simps opt_map_simps map_join_simps)
+  unfolding set_tcb_obj_ref_def
+  apply simp
+  apply (rule hoare_seq_ext[OF _ gets_the_sp])
+  apply (wpsimp simp: vs_all_heap_simps obj_at_def set_object_def
+                  wp: get_object_wp split: option.splits)
+  by (cases scopt;
```
michaelmcinerney, line 1518: These big blocks always scare me. When they break you don't immediately know where it broke. Is it possble to replace the bit from `rule conjI` with a `fastforce` of some kind?

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -13958,27 +13883,60 @@ lemma maybe_donate_sc_active_sc_valid_refills[wp]:
...
    \<lbrace>\<lambda>_. ready_or_release\<rbrace>"
   unfolding sched_context_resume_def
-  by (wpsimp wp: hoare_drop_imps hoare_vcg_all_lift hoare_vcg_if_lift simp: is_schedulable_def)
+  by (wpsimp wp: thread_get_wp is_schedulable_wp)
+     (clarsimp simp: obj_at_def is_schedulable_bool_def2 pred_tcb_at_def is_tcb)
+
+lemma tcb_sched_enqueue_ready_queues_etcb_eq[wp]:
+  "tcb_sched_action tcb_sched_enqueue tp \<lbrace>ready_qs_etcb_eq\<rbrace>"
+  unfolding tcb_sched_action_def tcb_sched_enqueue_def get_tcb_queue_def bind_assoc
+  apply (rule hoare_seq_ext[OF _ thread_get_sp])
```
michaelmcinerney, line 13903: Are these `hoare_seq_ext`s necessary? (And in the next couple proofs)

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -14074,37 +14032,37 @@ lemma send_signal_WaitingNtfn_helper:
...
-                     dest!: symreftype_inverse')
-              apply (clarsimp simp: get_refs_def2)+
-    by (intro conjI impI;
-        fastforce simp: is_tcb obj_at_def state_refs_of_def get_refs_def2 ntfn_q_refs_of_def
-                  dest!: symreftype_inverse'
-                  split: option.splits list.splits if_splits)
+    apply (prop_tac "ipc_queued_thread (hd wnlist) s", simp add: tcb_at_kh_simps pred_map_def)
+    apply (rule released_ipc_queues_blocked_on_recv_ntfn_D, simp)
+    by (fastforce elim: pred_map_weakenE simp: tcb_at_kh_simps is_blocked_thread_state_defs)
+      (* sym_refs *) (* FIXME RT: this is nasty, can it be improved? *)
```
michaelmcinerney, line 14046: I suppose this is more for the `refill_unblock_check` everywhere PR, but I'm fairly sure I was able to improve, or maybe even get rid of, all these `sym_refs` bits in the draft I had.

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -14391,7 +14349,8 @@ lemma send_signal_BOR_helper:
                       maybe_donate_sc_released_ipc_queues'
                 simp: valid_sched_def)
    apply (wpsimp cong: conj_cong simp: obj_at_kh_kheap_simps)
-  apply (clarsimp simp: valid_sched_def obj_at_kh_kheap_simps vs_all_heap_simps current_time_bounded_def)
+  apply (clarsimp simp: valid_sched_def obj_at_kh_kheap_simps vs_all_heap_simps current_time_bounded_def
+                  dest!: valid_ready_qs_etcb_eq)
```
michaelmcinerney, line 14353: This indentation is fine, of course, but I thought you preferred lining up the colons

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -17002,22 +16962,13 @@ lemma charge_budget_active_reply_scs:
...
 
 lemma charge_budget_ready_or_release:
-  "\<lbrace>ready_or_release and valid_ready_qs and ct_not_in_release_q and ct_not_queued
+  "\<lbrace>ready_or_release and ready_qs_etcb_eq and ct_not_in_release_q and ct_not_queued
     and cur_sc_chargeable and active_sc_valid_refills
     and cur_sc_active\<rbrace>
    charge_budget consumed canTimeout
    \<lbrace>\<lambda>_. ready_or_release :: 'state_ext state \<Rightarrow> _ \<rbrace>"
   apply (clarsimp simp: charge_budget_def when_def)
-  apply (wpsimp wp: hoare_drop_imp)
```
michaelmcinerney, line 17011: Nice

michaelmcinerney (approved):
So did you end up using this new abbreviation in the definition of `valid_ready_qs`? If not, that might be nice, if possible

proof/invariant-abstract/DetSchedSchedule_AI.thy
mktnk3: that is not my comment, that’s an existing one. I changed the indentation only. You can see the diff without whitespace changes by changing the GitHub setting, I think.

proof/invariant-abstract/DetSchedSchedule_AI.thy
mktnk3: yes, otherwise we’ll need to assume `tcb_at` I think.

> So did you end up using this new abbreviation in the definition of `valid_ready_qs`? If not, that might be nice, if possible

It might be possible eventually when we tweak `valid_ready_qs`, but currently it is a multilayered definition/abbreviation and `\<forall>tp d p. tp \<in> set (ready_queues s d p)` part is in a different layer from the rest of the statement. So it is not quite possible to use this as is. The abbreviation `etcb_eq` does appear in `valid_ready_qs` at some level.

mktnk3:
> So did you end up using this new abbreviation in the definition of `valid_ready_qs`? If not, that might be nice, if possible

It might be possible eventually when we tweak `valid_ready_qs`, but currently it is a multilayered definition/abbreviation and `\<forall>tp d p. tp \<in> set (ready_queues s d p)` part is in a different layer from the rest of the statement. So it is not quite possible to use this as is. The abbreviation `etcb_eq` does appear in `valid_ready_qs` at some level.

proof/invariant-abstract/DetSchedSchedule_AI.thy
mktnk3: I believe so. It discharges the first subgoal and then applies to the second. Can repeat instead.

proof/invariant-abstract/DetSchedSchedule_AI.thy
mktnk3: This is not too bad if you know what `valid_release_q` consists of. But it is a bit fiddly to leave out the `sorted_release_q` part at the end. I can certainly pull it apart by repeating things...

mktnk3:
I was too happy to get to the end of DetSchedSchedule_AI that I forgot that Refine existed 😅 hang on...

proof/invariant-abstract/DetSchedSchedule_AI.thy
mktnk3: well, it wasn't actually too bad to make it a bit simpler, although I do think we need two steps, not just one fastforce.

mktnk3:
I've added a couple more cleanup changes as separate commits.

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -2532,21 +2516,15 @@ lemma tcb_sched_dequeue_valid_sched_action_2_ct_upd:
   done
 
 lemma tcb_dequeue_not_queued:
-  "\<lbrace>valid_ready_qs\<rbrace> tcb_sched_action tcb_sched_dequeue tptr \<lbrace>\<lambda>_. not_queued tptr\<rbrace>"
+  "\<lbrace>\<lambda>s. (t = tptr \<and> ready_qs_etcb_eq s) \<or> not_queued t s\<rbrace>
+    tcb_sched_action tcb_sched_dequeue tptr
```
corlewis, line 2520: nitpick: indentation

proof/invariant-abstract/DetSchedSchedule_AI.thy
corlewis: Hmm, it's hard to tell without looking at the proofs, but that surprises me. In theory there shouldn't be any difference in power between sp and wp style proofs, and if this is true then I think that that would mean that the wp rules could be improved.

proof/invariant-abstract/DetSchedSchedule_AI.thy
corlewis: Yeah, awesome!

corlewis (approved):
Nice work, it definitely looks to simplify things!

proof/invariant-abstract/DetSchedSchedule_AI.thy
mktnk3: You are quite right there, actually. It was because of `thread_get_wp` not quite doing its job. And we of course have two versions - apparently `thread_get_wp'` is the right form and what we need here. We should probably get rid of the first and just keep the `wp'` version, and also for the master.

corlewis: We probably should make an issue so that we don't forget about that.

PR #318, opened 2021-08-05
New corres rule for whileLoop, and headInsufficientLoop_corres

michaelmcinerney:
This PR restores the original `corres` rule for `whileLoop`; it seems that the rule which included crossing is unnecessary. A new `corres` rules for `whileLoop` with termination and `no_fail` conditions stated on the abstract side is added. This is then used to prove `headInsufficientLoop_corres`.

### DIFF
--- a/lib/ExtraCorres.thy
+++ b/lib/ExtraCorres.thy
@@ -278,17 +278,16 @@ lemma wp_from_corres_unit:
   f' \<lbrace>\<lambda>_ s'. \<exists>s. (s,s') \<in> state_relation \<and> Q s \<and> Q' s'\<rbrace>"
   by (auto intro!: wp_from_corres_u_unit)
 
-lemma corres_whileLoop_results:
+lemma corres_whileLoop_results_helper:
   assumes cond: "\<And>r r' s s'. \<lbrakk>rrel r r'; (s, s') \<in> srel; P s; P' s'\<rbrakk> \<Longrightarrow> C r s = C' r' s'"
   assumes body_corres:
            "\<And>r r'. rrel r r'
-                    \<Longrightarrow> corres_underlying srel False nf' rrel (P and A and C r) (P' and A' and C' r') (B r) (B' r')"
-  assumes body_inv: "\<And>r. \<lbrace>P and A and C r\<rbrace> B r \<lbrace>\<lambda>_. P\<rbrace>" "\<And>r'. \<lbrace>P' and A' and C' r'\<rbrace> B' r' \<lbrace>\<lambda>_. P'\<rbrace>"
-  assumes cross: "\<And>s s'. \<lbrakk>(s, s') \<in> srel; P s; P' s'\<rbrakk> \<Longrightarrow> A s \<and> A' s'"
+                    \<Longrightarrow> corres_underlying srel False nf' rrel (P and C r) (P' and C' r') (B r) (B' r')"
+  assumes body_inv: "\<And>r. \<lbrace>P and C r\<rbrace> B r \<lbrace>\<lambda>_. P\<rbrace>" "\<And>r'. \<lbrace>P' and C' r'\<rbrace> B' r' \<lbrace>\<lambda>_. P'\<rbrace>"
   shows "(rv', t') \<in> fst (whileLoop C' B' r' s')
          \<Longrightarrow> \<forall>s r. (s,s') \<in> srel \<and> rrel r r' \<and> P s \<and> P' s'
                    \<longrightarrow> (\<exists>rv t. (rv, t) \<in> fst (whileLoop C B r s) \<and> (t,t') \<in> srel \<and> rrel rv rv')"
-  apply (rule_tac r=r' and B=B' and C=C' in in_whileLoop_induct)
+  apply (rule_tac r=r' and B=B' in in_whileLoop_induct)
     apply simp
    apply clarsimp
    apply (rename_tac r_conc s_conc s_abs r_abs)
@@ -298,8 +297,6 @@ lemma corres_whileLoop_results:
   apply (thin_tac "_ \<in> fst _")
   apply (rename_tac r s' r' t' r'' u')
   apply clarsimp
-  apply (prop_tac "A s \<and> A' s'")
-   apply (fastforce dest: cross)
   apply (rename_tac s r_abs)
   apply (insert body_corres)
   apply (drule_tac x=r_abs in meta_spec)
@@ -322,15 +319,11 @@ lemma corres_whileLoop_results:
   apply (simp add: cond whileLoop_def whileLoop_results.intros)
   done
 
-lemmas corres_whileLoop_results_inv = corres_whileLoop_results[where A=\<top> and A'=\<top>, simplified]
-
-lemma whileLoop_no_fail:
-  assumes body_guard: "\<And>r'. \<lbrace>P' and C' r'\<rbrace> B' r' \<lbrace>\<lambda>_. P'\<rbrace>"
+lemma corres_whileLoop_no_fail_helper:
+  assumes body_inv: "\<And>r'. \<lbrace>P' and C' r'\<rbrace> B' r' \<lbrace>\<lambda>_. P'\<rbrace>"
   assumes nf': "\<And>r'. no_fail (P' and C' r') (B' r')"
-  assumes termin: "\<And>r' s'. P' s' \<Longrightarrow> C' r' s' \<Longrightarrow> whileLoop_terminates C' B' r' s'"
-  shows "no_fail P' (whileLoop C' B' r')"
-  apply (simp add: no_fail_def)
-  apply (intro allI impI)
+  assumes termin: "\<And>r' s'. \<lbrakk>P' s'; C' r' s'\<rbrakk> \<Longrightarrow> whileLoop_terminates C' B' r' s'"
+  shows "P' s \<Longrightarrow> \<not> snd (whileLoop C' B' r' s)"
   apply (rule_tac I="\<lambda>_ s. P' s"
               and R="{((r', s'), r, s). C' r s \<and> (r', s') \<in> fst (B' r s)
                                         \<and> whileLoop_terminates C' B' r s}"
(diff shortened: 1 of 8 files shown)


michaelmcinerney:
Whoops I cut off the "Signed-off-by" in one of the commits when squashing. I'll add it back before merging

lib/ExtraCorres.thy
```diff
@@ -278,17 +278,16 @@ lemma wp_from_corres_unit:
   f' \<lbrace>\<lambda>_ s'. \<exists>s. (s,s') \<in> state_relation \<and> Q s \<and> Q' s'\<rbrace>"
   by (auto intro!: wp_from_corres_u_unit)
 
-lemma corres_whileLoop_results:
+lemma corres_whileLoop_results_helper:
```
corlewis, line 281: Out of curiosity, what has changed that this is now a helper?

proof/refine/ARM/SchedContextInv_R.thy
```diff
@@ -1012,6 +1012,268 @@ lemma refillUpdate_corres:
...
+  apply (subst refillHdInsufficient_simp; simp)
+  apply (frule refill_hd_relation2[where s'=s'])
+    apply (fastforce simp: vs_all_heap_simps opt_map_def)
+   apply (clarsimp simp: sc_ko_at_valid_objs_valid_sc' opt_map_def projectKOs obj_at'_def)
+  apply (clarsimp simp: obj_at_def sc_at_ppred_def obj_at'_def projectKOs minBudget_def
+                        MIN_BUDGET_def kernelWCETTicks_def opt_map_def vs_all_heap_simps)
+  done
+
+lemma set_object_no_fail[wp]:
+  "no_fail (obj_at (\<lambda>k. a_type obj = a_type k) sc_ptr) (set_object sc_ptr obj)"
```
corlewis, line 1170: Not necessary, but this and the `update_sched_context` could probably be moved to an earlier file like `KHeap_R`.

proof/refine/ARM/SchedContextInv_R.thy
```diff
@@ -1012,6 +1012,268 @@ lemma refillUpdate_corres:
...
+  apply (clarsimp simp: set_object_def)
+  apply (wpsimp wp: no_fail_obj_at get_object_wp)
+  apply (clarsimp simp: obj_at_def)
+  done
+
+lemma update_sched_context_no_fail:
+  "no_fail (sc_at sc_ptr) (update_sched_context sc_ptr f)"
```
corlewis, line 1177: I think that this one could be `[wp]` as well?

corlewis (approved):
Very smart, looks good to me!

lib/ExtraCorres.thy
michaelmcinerney: This was in restoring the original version of the `corres_whileLoop` proofs

michaelmcinerney: (The diff is a bit hard to read. If you open the source for this file, it's easy to see that it's just the old version plus the new stuff for abstract termination and `no_fail`.)

proof/refine/ARM/SchedContextInv_R.thy
michaelmcinerney: I was a bit suprised that we didn't already have a rule like this, and thought maybe there was a good reason we didn't. But if not, maybe I can move it to KHeap_A, even. Any further thoughts?

proof/refine/ARM/SchedContextInv_R.thy
michaelmcinerney: So `sc_at` wants `valid_sched_context_size` as well, but this isn't necessary here. I went with just `sc_at` for this rule because it avoided some messing around with quantifiers. I can look into it more if you think it's a good idea

proof/refine/ARM/SchedContextInv_R.thy
```diff
@@ -722,7 +722,7 @@ lemma refillResetRR_corres:
 lemma refillPopHead_corres:
   "corres (\<lambda>refill refill'. refill = refill_map refill')
               (pspace_aligned and pspace_distinct and sc_at sc_ptr and is_active_sc sc_ptr
-               and sc_refills_sc_at (\<lambda>refills. 1 < length refills) sc_ptr)
+               and (\<lambda>s. pred_map (\<lambda>cfg. Suc 0 < length (scrc_refills cfg)) (sc_refill_cfgs_of s) sc_ptr))
```
mktnk3, line 725: I don't think this change is necessary. It would be better to adjust your use site to accommodate the current version, I'd think?

To give a further context, `refillPopHead_corres` is used in my `refillUnblockCheck_corres` proof, too. I should have thought about this, but there might be other overlapping things.
https://github.com/mktnk3/l4v/commit/d92215ab52ff919fa77149bd2490f9a25b86e9df

michaelmcinerney: I much prefer the `pred_map` formulations. But if you would like me to change this, I shall. We need a policy regarding `obj_at` versus `pred_map` alternatives

mktnk3: well, isn't this actually the same thing as `is_active_sc`?

michaelmcinerney: I don't think we have this anywhere in the proofs. Carrying around that invariant would be its own pain

michaelmcinerney: Oh wait no. This wants length greater than 1, not 0. We can certainly have active scs which have length 1

mktnk3: ah, yes, it's 1, not 0. So, it's the `is_active_sc` part that we could do without.

mktnk3: I think, as a general rule, we'd better refrain from tweaking notations of existing lemmas if there is no semantic needs/improvements (i.e., you do need to fix that lemma). That is kind of a base line for multiple people working closely together on something.

As for the choice of `obj_at` or `pred_map`, we know that we can't easily agree on this point and there might not even be point in agreeing on one of those two choices right now. We are working towards converging (most likely to a third option) but it will be a while before that can happen.

proof/refine/ARM/SchedContextInv_R.thy
```diff
@@ -1012,6 +1012,268 @@ lemma refillUpdate_corres:
    apply clarsimp+
   done
 
+lemma head_insufficient_length_cross:
+  "\<lbrakk>(s,s') \<in> state_relation; valid_objs s; pspace_aligned s; pspace_distinct s;
+    pred_map (\<lambda>cfg. unat MIN_BUDGET \<le> refills_unat_sum (scrc_refills cfg))
+             (sc_refill_cfgs_of s) scPtr;
+    the (head_insufficient scPtr s); valid_objs' s'\<rbrakk>
+   \<Longrightarrow> obj_at' (\<lambda>sc'. Suc 0 < scRefillCount sc') scPtr s'"
+  apply (frule (1) head_insufficient_length_at_least_two[rotated])
```
mktnk3, line 1021: Hmm. This is a rather odd crossing lemma. Crossing the information in the assumption would give a lot more than the conclusion here. I think this should be two separate lemmas. One for proving that `head_insufficient` implies "refill length greater than 1" on the abstract side, which is actually proved below. Then, the second one is a lemma for crossing the conditions on the refill length, which would require a bit more than simply unfolding the `sc_relation`. I am not sure if we already have such a lemma.

proof/refine/ARM/SchedContextInv_R.thy
```diff
@@ -1012,6 +1012,268 @@ lemma refillUpdate_corres:
...
+   apply (fastforce dest!: sc_ko_at_valid_objs_valid_sc')
+  apply (auto simp: sc_relation_def refills_map_def valid_sched_context'_def)
+  done
+
+lemma non_overlapping_merge_refills_rewrite:
+  "update_refill_hd scPtr (r_time_update (\<lambda>t. t - r_amount old_head)
+                           \<circ> r_amount_update (\<lambda>m. m + r_amount old_head))
+   = do update_refill_hd scPtr (r_amount_update (\<lambda>m. m + r_amount old_head));
+        update_refill_hd scPtr (r_time_update (\<lambda>t. t - r_amount old_head))
+     od"
```
mktnk3, line 1048: Can't we do this simply by unfolding `update_refill_hd` and using an existing lemma about `update_sched_context`?

proof/refine/ARM/SchedContextInv_R.thy
mktnk3: Actually, I did write this length crossing lemma myself when I proved `refillUnblockCheck_corres`..
```
lemma length_sc_refills_cross:
  "\<lbrakk>(s, s') \<in> state_relation; sc_at scp s; sc_at' scp s'; valid_refills'' scp s';
   ((\<lambda>sc. P (length (sc_refills sc))) |< scs_of2 s) scp\<rbrakk>
   \<Longrightarrow> ((\<lambda>sc'. P (scRefillCount sc')) |< scs_of' s') scp"
  apply (drule (2) state_relation_sc_relation)
  apply (clarsimp simp: obj_at_simps is_sc_obj valid_refills''_def sc_relation_def opt_map_left_Some)
  done
  ```
  this is in https://github.com/mktnk3/l4v/commit/d92215ab52ff919fa77149bd2490f9a25b86e9df

File: proof/refine/ARM/SchedContextInv_R.thy [hunk 7]
```diff
@@ -1012,6 +1012,270 @@ lemma refillUpdate_corres:
...
+
+lemma headInsufficientLoop_corres:
+  "sc_ptr = scPtr
+   \<Longrightarrow> corres dc (pspace_aligned and pspace_distinct and sc_at sc_ptr and is_active_sc sc_ptr
+                  and valid_objs
+                  and (\<lambda>s. pred_map (\<lambda>cfg. unat MIN_BUDGET \<le> refills_unat_sum (scrc_refills cfg))
+                                    (sc_refill_cfgs_of s) sc_ptr)
+                  and (\<lambda>s. pred_map (\<lambda>cfg. refills_unat_sum (scrc_refills cfg) \<le> unat max_time)
+                                    (sc_refill_cfgs_of s) sc_ptr))
+                 (valid_objs' and valid_refills' scPtr)
```
mktnk3, line 1260: You can derive `valid_refills'` from `valid_objs'`. See `valid_objs'_valid_refills'` in StateRelation.thy.
You will need to cross `is_active_sc`, with a few hops, and there are lemmas for that, too.

proof/refine/ARM/SchedContextInv_R.thy
michaelmcinerney: Oh cool. I'll use that

proof/refine/ARM/SchedContextInv_R.thy
michaelmcinerney: I tried that, but I couldn't get it to work. I might take another look at it

proof/refine/ARM/SchedContextInv_R.thy
michaelmcinerney: This can be made into an if and only if, which has been helpful. I'll include this vesion of this rule, if you're OK with that

proof/refine/ARM/KHeap_R.thy
```diff
@@ -4677,4 +4674,16 @@ lemma active_sc_at'_cross:
                         active_sc_def)
   done
 
+lemma is_active_sc'2_cross:
+  "\<lbrakk>(s,s') \<in> state_relation; pspace_aligned s; pspace_distinct s; is_active_sc sc_ptr s;
+    sc_at sc_ptr s\<rbrakk>
+   \<Longrightarrow> is_active_sc' sc_ptr s'"
```
mktnk3, line 4680: I guess it makes sense to have this lemma. Just pointing out that we have `active_sc_at'_imp_is_active_sc'` and `active_sc_at'_cross` right above here which can bridge this equality. But I am happy for this to be merged as is. What we need to do one day soon is to consolidate all the "the SC is active" variations.

proof/refine/ARM/SchedContextInv_R.thy
```diff
@@ -1012,6 +1012,289 @@ lemma refillUpdate_corres:
...
+
+lemma length_sc_refills_cross:
+  "\<lbrakk>(s, s') \<in> state_relation; sc_at scp s; sc_at' scp s'; valid_refills' scp s'\<rbrakk>
+   \<Longrightarrow> ((\<lambda>sc. P (length (sc_refills sc))) |< scs_of2 s) scp
+        = ((\<lambda>sc'. P (scRefillCount sc')) |< scs_of' s') scp"
+  apply (drule (2) state_relation_sc_relation)
+  apply (clarsimp simp: obj_at_simps is_sc_obj valid_refills'_def sc_relation_def opt_map_red)
+  done
+
+lemma head_insufficient_length_cross:
```
mktnk3, line 1033: I think you can now remove this lemma. It is not used and we probably don't want it in.

proof/refine/ARM/SchedContextInv_R.thy
```diff
@@ -1012,6 +1012,289 @@ lemma refillUpdate_corres:
...
+      apply simp
+     apply (clarsimp simp: active_sc_at'_def)
+    apply (fastforce elim: valid_objs'_valid_refills'
+                     simp: active_sc_at'_def is_active_sc'_def obj_at'_def projectKOs
+                           opt_map_red)
+   apply (clarsimp simp: opt_map_red vs_all_heap_simps)
+  apply (clarsimp simp: obj_at'_def projectKOs active_sc_at'_def opt_map_red)
+  done
+
+lemma non_overlapping_merge_refills_rewrite:
```
mktnk3, line 1052: Glad this worked! We probably should give this a better name though.

proof/refine/ARM/Schedule_R.thy
```diff
@@ -2376,8 +2376,16 @@ crunches refillHeadOverlappingLoop, headInsufficientLoop
...
   and valid_release_queue'[wp]: valid_release_queue'
   and typ_at'[wp]: "\<lambda>s. P (typ_at' T p s)"
+  and sc_at'_n[wp]: "\<lambda>s. P (sc_at'_n T p s)"
   (wp: crunch_wps)
 
+end
+
+global_interpretation refillPopHead: typ_at_all_props' "refillPopHead scPtr"
+  by typ_at_props'
+
```
mktnk3, line 2386: Nice!

proof/refine/ARM/Schedule_R.thy
```diff
@@ -3859,7 +3834,7 @@ lemma awaken_corres:
   apply (rule corres_cross[where Q'="\<lambda>s'. distinct (ksReleaseQueue s')"
                            , OF ksReleaseQueue_distinct_cross_rel], blast)
   apply (clarsimp simp: awaken_def Schedule_A.awaken_def runReaderT_def)
-  apply (rule corres_whileLoop_inv; simp)
+  apply (rule corres_whileLoop; simp)
```
mktnk3, line 3837: :+1:

mktnk3 (approved):
Thanks for sorting out the `whileLoop_wp` and the crossing issue. This looks great. A few further cleanup comments but apart from those this looks good to be merged.

proof/invariant-abstract/KHeap_AI.thy
```diff
@@ -2620,4 +2620,40 @@ lemma st_tcb_recv_reply_state_refs:
...
+
+lemma set_object_no_fail[wp]:
+  "no_fail (obj_at (\<lambda>k. a_type obj = a_type k) ptr) (set_object ptr obj)"
+  apply (clarsimp simp: set_object_def)
+  apply (wpsimp wp: get_object_wp)
+  apply (clarsimp simp: obj_at_def)
+  done
+
+lemma update_sched_context_no_fail[wp]:
+  "no_fail (\<lambda>s. \<exists>sc n. ko_at (SchedContext sc n) sc_ptr s) (update_sched_context sc_ptr f)"
```
mktnk3, line 2653: I thought we could stick with `sc_at`, and now I am not sure this is the right direction to go. Isn't `kheap s sc_ptr = Some (SchedContext sc n)` what we want here? We can skip unfolding `obj_at` if we are happy to take existentials?

mktnk3 (approved):
Thanks for moving those `no_fail` lemmas to AInvs.

proof/invariant-abstract/KHeap_AI.thy
michaelmcinerney: I think `sc_at` would be easiest, since we already have so many rules for it. But it wouldn't be weakest precondition then. There are already some issues (e.g. VER-1350) mentioning `sc_at` versus some alternatives. Maybe for now, we can leave it as is, and I can mention in that issue that we might want to also look at this rule when we get around to deciding on an approach?

mktnk3: It shouldn’t have been `sc_at` to start with, as we do have another way to express the weakest precondition. Let’s go with `kheap s sc_ptr = Some (SchedContext sc n)` with quantifiers, since you’ve moved it to where `scs_of` is not yet available.

lsf37 (approved):
Looks good from my side as well.

PR #319, opened 2021-08-06
docs: add setup instructions for PIDE plug-ins

zaklogician:
Add a section on installation and configuration procedures of common Isabelle PIDE (jEdit) plug-ins.

Signed-off-by: Zoltan Kocsis <zoltan.kocsis@data61.csiro.au>


lsf37 (approved):
There is trailing whitespace in one of the lines, which would be good to fix. Otherwise, good to go. Thanks for putting this up!

PR #320, opened 2021-08-06
refill_budget_check guard

michaelmcinerney:
This brings the guard in `refill_budget_check` in line with the C: see

https://github.com/seL4/seL4/blob/master/src/kernel/sporadic.c#L284

### DIFF
--- a/proof/invariant-abstract/DetSchedSchedule_AI.thy
+++ b/proof/invariant-abstract/DetSchedSchedule_AI.thy
@@ -7099,7 +7099,7 @@ lemma no_ofail_refill_head_overlapping:
   done
 
 lemma refill_head_overlapping_true_imp_length_at_least_two:
-  "\<lbrakk>the (refill_head_overlapping sc_ptr s); pred_map \<top> (sc_refill_cfgs_of s) sc_ptr\<rbrakk>
+  "\<lbrakk>the (refill_head_overlapping sc_ptr s); pred_map \<top> (scs_of s) sc_ptr\<rbrakk>
    \<Longrightarrow> pred_map (\<lambda>cfg. Suc 0 < length (scrc_refills cfg)) (sc_refill_cfgs_of s) sc_ptr"
   apply (clarsimp simp: vs_all_heap_simps)
   apply (insert no_ofailD[OF no_ofail_refill_head_overlapping])
@@ -7107,7 +7107,7 @@ lemma refill_head_overlapping_true_imp_length_at_least_two:
   done
 
 lemma refill_head_overlapping_refills_overlapping:
-  "\<lbrakk>the (refill_head_overlapping sc_ptr s); pred_map \<top> (sc_refill_cfgs_of s) sc_ptr\<rbrakk>
+  "\<lbrakk>the (refill_head_overlapping sc_ptr s); pred_map \<top> (scs_of s) sc_ptr\<rbrakk>
    \<Longrightarrow> pred_map (\<lambda>cfg. r_time (hd (tl (scrc_refills cfg)))
                        \<le> r_time (hd (scrc_refills cfg)) + r_amount (hd (scrc_refills cfg)))
                  (sc_refill_cfgs_of s) sc_ptr"
@@ -7490,7 +7490,7 @@ lemma bound_head_insufficient:
   done
 
 lemma head_insufficient_true_imp_insufficient:
-  "\<lbrakk>the (head_insufficient sc_ptr s); pred_map \<top> (sc_refill_cfgs_of s) sc_ptr\<rbrakk>
+  "\<lbrakk>the (head_insufficient sc_ptr s); pred_map \<top> (scs_of s) sc_ptr\<rbrakk>
    \<Longrightarrow> pred_map (\<lambda>cfg. r_amount (scrc_refill_hd cfg) < MIN_BUDGET) (sc_refill_cfgs_of s) sc_ptr"
   apply (clarsimp simp: vs_all_heap_simps)
   apply (frule bound_head_insufficient)
@@ -7498,7 +7498,7 @@ lemma head_insufficient_true_imp_insufficient:
   done
 
 lemma head_insufficient_false_imp_sufficient:
-  "\<lbrakk>\<not> (the (head_insufficient sc_ptr s)); pred_map \<top> (sc_refill_cfgs_of s) sc_ptr\<rbrakk>
+  "\<lbrakk>\<not> (the (head_insufficient sc_ptr s)); pred_map \<top> (scs_of s) sc_ptr\<rbrakk>
    \<Longrightarrow> pred_map (\<lambda>cfg. MIN_BUDGET \<le> r_amount (scrc_refill_hd cfg)) (sc_refill_cfgs_of s) sc_ptr"
   apply (clarsimp simp: vs_all_heap_simps)
   apply (frule bound_head_insufficient)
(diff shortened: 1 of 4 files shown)


proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -8847,24 +8847,34 @@ lemma bound_head_time_buffer:
...
 
 lemma head_time_buffer_true_imp_buffer:
-  "\<lbrakk>the (head_time_buffer usage s); pred_map \<top> (sc_refill_cfgs_of s) (cur_sc s)\<rbrakk>
-   \<Longrightarrow> pred_map (\<lambda>cfg. max_time - r_time (hd ((scrc_refills cfg))) \<ge> 5 * MAX_PERIOD)
-                (sc_refill_cfgs_of s) (cur_sc s)"
-  apply (frule bound_head_time_buffer[where usage=usage])
-  apply (clarsimp simp: head_time_buffer_def obind_def read_sched_context_def ogets_def
-                        vs_all_heap_simps
-                 split: kernel_object.splits)
+  "pred_map \<top> (sc_refill_cfgs_of s) (cur_sc s)
```
mktnk3, line 8850: Since a fragment of an SC cannot exist in the kheap without an whole SC existing there, I think this should be phrased in terms of `scs_of`, probably simply something like `bound (scs_of s) (cur_sc s)`. It could be `sc_at` even.

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -9408,6 +9423,14 @@ lemma handle_overrun_loop_length:
            | wps)+
   done
 
+lemma handle_overrun_loop_body_sc_at_cur_sc[wp]:
```
mktnk3, line 9426: I think this name suggests something different from what is stated there, but it probably should be actually about `sc_at` or something equivalent. See the comment above.

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -8847,24 +8847,34 @@ lemma bound_head_time_buffer:
...
 lemma head_time_buffer_true_imp_unat_buffer:
-  "\<lbrakk>the (head_time_buffer usage s); pred_map \<top> (sc_refill_cfgs_of s) (cur_sc s)\<rbrakk>
-   \<Longrightarrow> pred_map (\<lambda>cfg. unat (r_time (hd (scrc_refills cfg))) + 5 * unat MAX_PERIOD \<le> unat max_time)
-                (sc_refill_cfgs_of s) (cur_sc s)"
-  apply (frule head_time_buffer_true_imp_buffer[where usage=usage])
-   apply (clarsimp simp: vs_all_heap_simps obj_at_kh_kheap_simps)
-  apply (clarsimp simp: vs_all_heap_simps obj_at_kh_kheap_simps)
-  apply (clarsimp simp: word_le_nat_alt unat_sub)
-  apply (metis MAX_PERIOD_mult add_diff_cancel_left' le_diff_iff nat_le_iff_add unat_bounded_above)
+  "pred_map \<top> (sc_refill_cfgs_of s) (cur_sc s)
```
mktnk3, line 8867: The same comment as above.

spec/abstract/SchedContext_A.thy
```diff
@@ -254,14 +254,15 @@ where
      \<equiv> whileLoop (\<lambda>_ s. the ((head_insufficient sc_ptr) s))
                   (\<lambda>_. non_overlapping_merge_refills sc_ptr) ()"
 
+definition "MAX_RELEASE_TIME = max_time - 5 * MAX_PERIOD"
```
mktnk3, line 257: If we are adding this notion this far down in the abstract spec, we should use lower case letters as a derived notion. Then we'd want to state the equality between that and `maxReleaseTime` in StateRelation.thy.

As an alternative, we could move it to MachineExports.thy or the like and have the equality added in Hardware_H.

I can see that, on the haskell side, we are missing the `MaxPeriod` alias and always going back to `MaxPeriodUs`. It would make sense to add and use that, too.

mktnk3 (approved):
Nice. The specs look nicer with the notion of max_release_time. A few comments mostly from the overall coherence perspective.

spec/abstract/SchedContext_A.thy
michaelmcinerney: OK. Does that mean that `MAX_PERIOD` should be lowercase too? It's derived from `MAX_PERIOD_US` and `us_to_ticks`.

mktnk3: What I meant is that this is not a good place to define a C level constant, which all upper case naming suggests. We could either move this definition to an appropriate place, or use lower case to consider it as an abstract spec constant. But either way we probably should maintain that distinction. `MAX_PERIOD` is defined in MachineOps. So, that makes sense.
Also, even if we are to choose using lower case, we'd better move the definition to Structure_A or somewhere anyway.

corlewis: I agree with Miki, if this can be moved to MachineOps and used by both specs then that would be nice. We're generally keen to reuse more things when possible instead of defining them twice.

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -8847,24 +8847,34 @@ lemma bound_head_time_buffer:
...
+                   (sc_refill_cfgs_of s) (cur_sc s)
+          \<and> pred_map (\<lambda>cfg. r_amount (scrc_refill_hd cfg) \<le> usage )
+                     (sc_refill_cfgs_of s) (cur_sc s))"
+  apply (fastforce simp: head_time_buffer_def obind_def read_sched_context_def ogets_def
+                         vs_all_heap_simps MAX_RELEASE_TIME_def
+                  split: kernel_object.splits)
+  done
+
+lemma unat_MAX_RELEASE_TIME:
+  "unat MAX_RELEASE_TIME = unat max_time - 5 * unat MAX_PERIOD"
```
corlewis, line 8862: nitpick: it wouldn't hurt to have this earlier, maybe in MachineOps as well if `MAX_RELEASE_TIME` is moved

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -9275,8 +9287,10 @@ lemma handle_overrun_loop_head_bound:
    apply (rule_tac y="unat (r_time (refill_hd sc)) + unat (sc_period sc) + 4 * unat MAX_PERIOD"
                 in order_trans)
     apply linarith
-   apply (clarsimp simp: vs_all_heap_simps)
-  apply (prop_tac "unat (sc_period sc) + 4 * unat MAX_PERIOD \<le> 5 * unat MAX_PERIOD", linarith)
```
corlewis, line 9279: Very nice that we can get rid of these hardcoded values!

corlewis (approved):
Looks good!

spec/abstract/SchedContext_A.thy
michaelmcinerney: `MachineOps` is an arch-dependent file. If I were to move `MAX_RELEASE_TIME` there, we'd have to do it for all architectures. And if we wanted to modify the definition (say, changing the `5` to a `6`), we'd need to update it in all files. Currently, `MAX_RELEASE_TIME` is using the appropriate arch dependent constant, namely `MAX_PERIOD`, in both architectures. Is this not ideal?

Is there a convention again having constant being defined in certain files?

And while we're discussing this constant, I actually think it's a bad name. It is not an upper bound on the release time of a scheduling context: `refill_budget_check` and `refill_update` are both able to push refills past this point. It would be better to have it called `OVERRUN_LOOP_CAP` or something like that, but I know that the systems people won't want to change the name.

mktnk3: Whether this constant should be arch specific or not is a separate question... this is not a good place to define it either way. We can assume that it is generic for now and define it maybe in MachineExports.thy as I suggested in the above.

After all, Refine isn't arch-split yet anyway.

I don't think we want to be dealing with different names from C for constants coming from C. You are welcome to challenge the naming with the systems team.

PR #321, opened 2021-08-10
`scheduleUsed_corres`

michaelmcinerney:
This proves `scheduleUsed_corres`.

I've rewritten `schedule_used` in the abstract to be monadic. I had hoped that the changes to AInvs would be minimal, but it turned out to be quite a lot of reorganising. Most of the more difficult word proofs (`schedule_used_no_overflow`, `schedule_used_odered_disjoint`, `schedule_used_window`) are pretty much unchanged. Some of the helper lemmas (`refill_budget_check_ordered_disjoint_helper'`, etc.) are new and replace older, clunkier versions. Nothing very exciting is going on in AInvs.

As expected after rewriting `schedule_used`, the `corres` rule was very straightforward. I had to slightly modify `refillAddTail_corres` to allow it to be applied in my lemma. This did not break anything in this branch, but perhaps it is being used elsewhere

### DIFF
--- a/lib/Corres_UL.thy
+++ b/lib/Corres_UL.thy
@@ -537,6 +537,13 @@ lemmas corres_when2 =
                 and Q="\<top>" and Q'="\<top>" and r=dc, unfolded if_apply_def2, simplified,
                 folded when_def]
 
+lemma corres_if_split:
+  "\<lbrakk> G = G';
+     G \<Longrightarrow> corres_underlying sr nf nf' r P P' a c;
+     \<not>G \<Longrightarrow> corres_underlying sr nf nf' r P P' b d \<rbrakk>
+   \<Longrightarrow> corres_underlying sr nf nf' r P P' (if G then a else b) (if G' then c else d)"
+  by (clarsimp simp: corres_underlying_def)
+
 text \<open>Some equivalences about liftM and other useful simps\<close>
 
 lemma snd_liftM [simp]:
--- a/proof/invariant-abstract/DetSchedSchedule_AI.thy
+++ b/proof/invariant-abstract/DetSchedSchedule_AI.thy
@@ -7510,16 +7510,19 @@ method non_overlapping_merge_refills_simple
      , (clarsimp simp: vs_all_heap_simps obj_at_def)?)
 
 lemma non_overlapping_merge_refills_nonempty_refills:
-  "non_overlapping_merge_refills sc_ptr
-   \<lbrace>\<lambda>s. pred_map (\<lambda>cfg. scrc_refills cfg \<noteq> []) (sc_refill_cfgs_of s) sc_ptr\<rbrace>"
+  "\<lbrace>\<lambda>s. pred_map (\<lambda>cfg. scrc_refills cfg \<noteq> []) (sc_refill_cfgs_of s) sc_ptr \<and> sc_ptr = cur_sc s\<rbrace>
+   non_overlapping_merge_refills sc_ptr
+   \<lbrace>\<lambda>_ s. pred_map (\<lambda>cfg. scrc_refills cfg \<noteq> []) (sc_refill_cfgs_of s) sc_ptr\<rbrace>"
   apply non_overlapping_merge_refills_simple
   done
 
 lemma head_insufficient_loop_nonempty_refills:
-  "head_insufficient_loop sc_ptr
-   \<lbrace>\<lambda>s. pred_map (\<lambda>cfg. scrc_refills cfg \<noteq> []) (sc_refill_cfgs_of s) sc_ptr\<rbrace>"
+  "\<lbrace>\<lambda>s. pred_map (\<lambda>cfg. scrc_refills cfg \<noteq> []) (sc_refill_cfgs_of s) sc_ptr \<and> sc_ptr = cur_sc s\<rbrace>
+   head_insufficient_loop sc_ptr
+   \<lbrace>\<lambda>_ s. pred_map (\<lambda>cfg. scrc_refills cfg \<noteq> []) (sc_refill_cfgs_of s) sc_ptr\<rbrace>"
+  (is "valid ?pre _ _")
   apply (clarsimp simp: head_insufficient_loop_def)
-  apply (wpsimp wp: whileLoop_wp'; fastforce?)
+  apply (wpsimp wp: valid_whileLoop[where I="\<lambda>_. ?pre"]; fastforce?)
   apply (wpsimp wp: non_overlapping_merge_refills_nonempty_refills)
   done
 
@@ -7582,7 +7585,7 @@ lemma head_insufficient_loop_refills_sum:
         \<and> pred_map (\<lambda>cfg. unat MIN_BUDGET \<le> refills_unat_sum (scrc_refills cfg))
                     (sc_refill_cfgs_of s) sc_ptr
         \<and> pred_map (\<lambda>cfg. refills_unat_sum (scrc_refills cfg) \<le> unat max_time)
-                    (sc_refill_cfgs_of s) sc_ptr\<rbrace>
+                    (sc_refill_cfgs_of s) sc_ptr \<and> sc_ptr = cur_sc s\<rbrace>
    head_insufficient_loop sc_ptr
    \<lbrace>\<lambda>_ s. pred_map (\<lambda>cfg. refills_sum (scrc_refills cfg) = scrc_budget cfg)
                    (sc_refill_cfgs_of s) sc_ptr\<rbrace>"
@@ -7590,7 +7593,7 @@ lemma head_insufficient_loop_refills_sum:
   apply (clarsimp simp: head_insufficient_loop_def)
   apply (wpsimp wp: valid_whileLoop[where I="\<lambda>_. ?pre"]
          ; fastforce?)
-  apply (intro hoare_vcg_conj_lift_pre_fix)
+  apply (intro hoare_vcg_conj_lift_pre_fix; (solves wpsimp)?)
     defer
     apply (wpsimp wp: non_overlapping_merge_refills_refills_unat_sum_lower_bound)
     apply (fastforce dest: head_insufficient_length_at_least_two)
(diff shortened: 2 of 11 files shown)


proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -8244,20 +8258,21 @@ lemma refill_update_valid_refills:
...
 
-lemma schedule_used_sum[simp]:
-  "refills_sum (schedule_used b refills new) = refills_sum (refills @ [new])"
-  apply (cases refills)
-   apply (clarsimp simp: schedule_used_def)
+lemma schedule_used_sum:
+  "\<lbrace>\<lambda>s. pred_map (\<lambda>cfg. refills_sum ((scrc_refills cfg) @ [new]) = t) (sc_refill_cfgs_of s) (cur_sc s)
+        \<and> sc_ptr = cur_sc s\<rbrace>
+   schedule_used sc_ptr new
+   \<lbrace>\<lambda>_ s. pred_map (\<lambda>cfg. refills_sum (scrc_refills cfg) = t) (sc_refill_cfgs_of s) (cur_sc s)\<rbrace>"
```
corlewis, line 8265: It's very hard to say without actually playing around with it, but would it be easier if these lemmas about `schedule_used` were generalised away from `cur_sc`?

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -8329,153 +8344,82 @@ lemma valid_refills_r_amount_bounded_max_sc_period:
...
-   apply (prop_tac "unat (r_amount (last lista)) \<le> sum_list (map unat (map r_amount lista))")
-    apply (simp add: member_le_sum_list)
-   apply (clarsimp simp: max_word_def)
-  apply simp
-  apply (prop_tac "butlast lista @ [last lista] = lista", fastforce)
-  apply (prop_tac "sum_list (map unat (map r_amount [last lista])) = unat (r_amount (last lista))")
-    apply simp
-  apply (metis sum_list_append map_append)
+lemma unat_max_time:
+  "2 ^ LENGTH(64) - 1 = unat max_time"
```
corlewis, line 8363: This and the next lemma should probably be about `max_word` instead of `max_time`, and also marked for being moved (possible even into `Word_Lib`, although it's hard to tell whether they're distinct enough from the already existing lemmas)

corlewis (approved):
It's a bit sad to get rid of the nice abstract version of `schedule_used`, but this definitely seems like the best way forward as opposed to doing something more complicated in Refine. Great work!

If possible I always like to properly generalise lemmas for all objects instead of just the `cur_sc` (/`cur_tcb`), but that might not be necessary for this PR.

proof/refine/ARM/SchedContextInv_R.thy
```diff
@@ -1012,6 +1012,120 @@ lemma refillUpdate_corres:
...
+  apply (clarsimp simp: get_refills_def)
+  apply (wpsimp wp: get_sched_context_exs_valid)
+   apply (fastforce intro: sc_atD1)
+  apply simp
+  done
+
+(* FIXME RT: move? *)
+lemma corres_if4:
+  "\<lbrakk>G = G'; G \<Longrightarrow> corres_underlying sr nf nf' r P P' a c; \<not>G \<Longrightarrow> corres_underlying sr nf nf' r P P' b d\<rbrakk>
+   \<Longrightarrow> corres_underlying sr nf nf' r P P' (if G then a else b) (if G' then c else d)"
```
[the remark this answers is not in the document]
michaelmcinerney: I suppose this isn't needed. I could just use `corres_if3` (maybe with some sort of `where blah, simplified blah` statement). But this rule is exactly what my (probably eccentric) thought process wanted. Should I get rid of it? Or move it to `Corres_UL.thy`?

proof/invariant-abstract/DetSchedSchedule_AI.thy
michaelmcinerney: It certainly wouldn't be any easier if these weren't `cur_sc`. I found it very annoying to get this all right, and I dreaded trying to use more general versions along with `wps` (which I've never had a good handle on). If you'd like me to take another look, I can, or maybe I can make a JIRA issue for it

proof/invariant-abstract/DetSchedSchedule_AI.thy
michaelmcinerney: Sure. I've got some nice rules generalised to `'a :: len word`. Maybe `Word_Lemmas_Internal.thy` in `Word_Lib`?

lsf37: Yes, the `*_Internal` theories are the right place to put these.

lsf37: Apart from this one, of course :-) (sorry, saw that as soon as I wrote it). `max_time` is not going to be defined there yet, so somewhere close to where `max_time` is defined would be a good spot.

lsf37: (when generalised to `max_word`)

proof/refine/ARM/SchedContextInv_R.thy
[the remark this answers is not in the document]
corlewis: Eh, if it's more convenient then I'm fine with keeping this one, with a preference for moving it to `Corres_UL`. It just feels a bit strange with how close it is to `corres_if3`.

michaelmcinerney:
I've updated and rebased this PR.

Git pretty badly mangled the rebase. It ended up deleting quite a few lemmas from DetSchedScheulde, but I've added those in a separate commit. And considering there were a number of things I wanted to modify in Refine, and the rebase there was not done well automatically, I decided to remove the old commit for Refine and make a new one. Apologies if this makes reviewing more difficult. I've cherry-picked a commit for several word lemmas from the `us_to_ticks` PR, and put `corres_if4` into its own commit, too

lib/Corres_UL.thy
```diff
@@ -537,6 +537,13 @@ lemmas corres_when2 =
                 and Q="\<top>" and Q'="\<top>" and r=dc, unfolded if_apply_def2, simplified,
                 folded when_def]
 
+lemma corres_if4:
+  "\<lbrakk> G = G';
+     G \<Longrightarrow> corres_underlying sr nf nf' r P P' a c;
+     \<not>G \<Longrightarrow> corres_underlying sr nf nf' r P P' b d \<rbrakk>
+   \<Longrightarrow> corres_underlying sr nf nf' r P P' (if G then a else b) (if G' then c else d)"
+  by (clarsimp simp: corres_underlying_def)
+
```
mktnk3, line 546: Hmm. This looks quite trivial to me. `simp` or `clarsimp` will split the conclusion into a conjunction of the two implications in the assumption or something similar. So, `clarsimp` and `conjI/impI` would be pretty much the same thing as applying this rule, woundn’t it?

proof/refine/ARM/SchedContextInv_R.thy
```diff
@@ -1268,6 +1268,98 @@ lemma headInsufficientLoop_corres:
...
+  apply (wpsimp wp: refillFull_wp)
+  apply (clarsimp simp: obj_at'_def)
+  done
+
+lemma length_refills_map_is_scRefillCount:
+  "sc_valid_refills' sc
+   \<Longrightarrow> (length (refills_map (scRefillHead sc) (scRefillCount sc) (scRefillMax sc) (scRefills sc)))
+        = scRefillCount sc"
+  apply (clarsimp simp: valid_sched_context'_def refillTailIndex_def refills_map_def)
+  done
```
mktnk3, line 1288: We already have
```
lemma length_refills_map[simp]:
  "⟦ mx ≤ length list; count ≤ mx ⟧ ⟹ length (refills_map start count mx list) = count"
  by (clarsimp simp: refills_map_def)
```
in the simp set. Given that `sc_valid_refills’` is an abbreviation, I don’t see why we need to have a separate lemma for this particular instance.

lib/Corres_UL.thy
michaelmcinerney: It is quite trivial. `clarsimp` seems to be greedy and will pull apart all `if`s. `schedule_used` contains three `if`s, so here it will create duplication which I can't see how to handle easily, and disturbs the structure of the proof I was wanting

proof/refine/ARM/SchedContextInv_R.thy
michaelmcinerney: Removed

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -8305,153 +8320,82 @@ lemma valid_refills_r_amount_bounded_max_sc_period:
...
+  apply (clarsimp simp: max_word_def)
   done
```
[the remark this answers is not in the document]
michaelmcinerney: Yes. Removed

lib/Corres_UL.thy
mktnk3: Ok, you want to proceed one layer at a time. Did you try using `corres_guard_imp` and `corres_if3`? Or `corres_guard_cong`?

michaelmcinerney: I could use `corres_guard_imp` and `corres_if3`, but I was aiming for a more "fowards" style, where the guards are either provided by strongest postcondition rules, "skip" rules, comparison with the initial guards, or (as a last resort) manual instantiation. `corres_if4` is exactly 

```
corres_if3[where P=T and Q=T for T, where P'=T' and Q'=T' for T', simplified]
```

so I could just use that instead, but I thought that a simple rule like this might be more convenient for people who like this (even more) forwards reasoning style in `corres` proofs

michaelmcinerney: I would certainly prefer to keep it. `corres_if_split` isn't taken. Would you be happy for me to rename it to `corres_if_split`?

mktnk3 (approved):
Great to see the monadic version of `schedule_used` working nicely. A few more comments for moving smaller lemmas out of DetSchedSchedule_AI.

lib/Corres_UL.thy
mktnk3: that would be good, if there’s no objection from other people.

PR #322, opened 2021-08-11
add opt_predD and other opt_map/opt_pred lemmas

mktnk3:
A small addition to OptionMonad, a few lemmas for controlling `opt_map` and `opt_pred`. `opt_predD` has been useful for me.

### DIFF
--- a/lib/Monad_WP/OptionMonad.thy
+++ b/lib/Monad_WP/OptionMonad.thy
@@ -116,12 +116,16 @@ lemma opt_predD:
   "(P |< proj) x \<Longrightarrow> \<exists>y. proj x = Some y \<and> P y"
   by (clarsimp split: option.splits)
 
-lemma opt_pred_unfold:
-  "(P |< f |> g)=  ((P |< g) |< f)"
+lemma opt_predE:
+  "\<lbrakk>(P |< proj) x; \<And>y. \<lbrakk>proj x = Some y; P y\<rbrakk> \<Longrightarrow> R\<rbrakk> \<Longrightarrow> R"
+  by (clarsimp split: option.splits)
+
+lemma opt_pred_unfold_map:
+  "(P |< f |> g) = ((P |< g) |< f)"
   by (fastforce simp: opt_map_def split: option.splits)
 
 lemma opt_pred_unfold_proj:
-  "(P |< f ||> g)=  ((\<lambda>x. P (g x)) |< f)"
+  "(P |< f ||> g)=  (P o g |< f)"
   by (clarsimp simp: opt_map_def split: option.splits)
 
 (* obind, etc. *)
--- a/lib/Monad_WP/OptionMonad.thy
+++ b/lib/Monad_WP/OptionMonad.thy
@@ -112,6 +112,18 @@ lemma opt_pred_disj:
   "((P1 |< hp) p \<or> (P2 |< hp) p) = (((P1 or P2) |< hp) p)"
   by (fastforce simp: pred_disj_def split: option.splits)
 
+lemma opt_predD:
+  "(P |< proj) x \<Longrightarrow> \<exists>y. proj x = Some y \<and> P y"
+  by (clarsimp split: option.splits)
+
+lemma opt_pred_unfold:
+  "(P |< f |> g)=  ((P |< g) |< f)"
+  by (fastforce simp: opt_map_def split: option.splits)
+
+lemma opt_pred_unfold_proj:
+  "(P |< f ||> g)=  ((\<lambda>x. P (g x)) |< f)"
+  by (clarsimp simp: opt_map_def split: option.splits)
+
 (* obind, etc. *)
 
 definition
--- a/lib/Monad_WP/OptionMonad.thy
+++ b/lib/Monad_WP/OptionMonad.thy
@@ -62,6 +62,30 @@ lemma opt_map_upd_Some:
 
 lemmas opt_map_upd[simp] = opt_map_upd_None opt_map_upd_Some
 
+lemma opt_map_Some_comp:
+  "f ||> h o g = f ||> g ||> h"
+  by (fastforce simp: opt_map_def split: option.split)
+
+lemma opt_map_unfold_l:
+  "(f |> g) x = (id |> g) (f x)"
+  by (clarsimp simp: opt_map_def)
+
+(* LHS is the same as (f ||> id) *)
+lemma opt_map_unit_r:
+  "(f |> Some) x = f x"
+  by (clarsimp simp: opt_map_def split: option.split)
+
+lemma opt_map_unit_l:
+  "(Some |> f) x = f x"
+  by (clarsimp simp: opt_map_def split: option.split)
+
+lemma opt_map_zero_l:
+  "(Map.empty |> g) x = None"
+  by (clarsimp simp: opt_map_def)
+
+lemma opt_map_zero_r:
+  "(f |> Map.empty) x = None"
+  by (clarsimp simp: opt_map_def split: option.split)
 
 lemma case_opt_map_distrib:
   "((\<lambda>s. case_option None g (f s)) |> h)
--- a/lib/Monad_WP/OptionMonad.thy
+++ b/lib/Monad_WP/OptionMonad.thy
@@ -101,7 +101,7 @@ lemma "\<lbrakk> (f |> g) x = None; g v = None \<rbrakk> \<Longrightarrow> f(x \
 (* opt_pred *)
 
 abbreviation
-  opt_pred :: "('a \<Rightarrow> bool) \<Rightarrow> ('b \<Rightarrow> 'a option) \<Rightarrow> ('b \<Rightarrow> bool)" (infixl "|<" 50) where
+  opt_pred :: "('a \<Rightarrow> bool) \<Rightarrow> ('b \<Rightarrow> 'a option) \<Rightarrow> ('b \<Rightarrow> bool)" (infixl "|<" 55) where
   "P |< proj \<equiv> (\<lambda>x. case_option False P (proj x))"
 
 lemma opt_pred_conj:
@@ -121,11 +121,11 @@ lemma opt_predE:
   by (clarsimp split: option.splits)
 
 lemma opt_pred_unfold_map:
-  "(P |< f |> g) = ((P |< g) |< f)"
+  "(P |< (f |> g)) = ((P |< g) |< f)"
   by (fastforce simp: opt_map_def split: option.splits)
 
 lemma opt_pred_unfold_proj:
-  "(P |< f ||> g)=  (P o g |< f)"
+  "(P |< (f ||> g))=  (P o g |< f)"
   by (clarsimp simp: opt_map_def split: option.splits)
 
 (* obind, etc. *)
--- a/lib/Monad_WP/OptionMonad.thy
+++ b/lib/Monad_WP/OptionMonad.thy
@@ -62,28 +62,28 @@ lemma opt_map_upd_Some:
 
 lemmas opt_map_upd[simp] = opt_map_upd_None opt_map_upd_Some
 
-lemma opt_map_Some_comp:
+lemma opt_map_Some_comp[simp]:
   "f ||> h o g = f ||> g ||> h"
   by (fastforce simp: opt_map_def split: option.split)
 
-lemma opt_map_unfold_l:
-  "(f |> g) x = (id |> g) (f x)"
+lemma opt_map_fold_l:
+  "(id |> g) (f x) = (f |> g) x"
   by (clarsimp simp: opt_map_def)
 
 (* LHS is the same as (f ||> id) *)
-lemma opt_map_unit_r:
+lemma opt_map_unit_r[simp]:
   "(f |> Some) x = f x"
   by (clarsimp simp: opt_map_def split: option.split)
 
-lemma opt_map_unit_l:
+lemma opt_map_unit_l[simp]:
   "(Some |> f) x = f x"
   by (clarsimp simp: opt_map_def split: option.split)
 
-lemma opt_map_zero_l:
+lemma opt_map_zero_l[simp]:
   "(Map.empty |> g) x = None"
   by (clarsimp simp: opt_map_def)
 
-lemma opt_map_zero_r:
+lemma opt_map_zero_r[simp]:
   "(f |> Map.empty) x = None"
   by (clarsimp simp: opt_map_def split: option.split)
 


lib/Monad_WP/OptionMonad.thy
```diff
@@ -62,6 +62,30 @@ lemma opt_map_upd_Some:
 
 lemmas opt_map_upd[simp] = opt_map_upd_None opt_map_upd_Some
 
+lemma opt_map_Some_comp:
+  "f ||> h o g = f ||> g ||> h"
```
corlewis, line 66: Hard to tell without opening Isabelle, but this and the four following after the next one all look like they would be nice to have as `[simp]`.

michaelmcinerney (approved):
Looks good. 

Just out of curiosity, is it possible to add fancier symbols for `opt_map` and `opt_pred`? Does Isabelle or jEdit only work with a certain character set? I don't have any particular symbol in mind, but if these symbols will be everywhere in our proofs soon, it'd be nice to have it look as nice and coherent as possible

corlewis (approved):

lib/Monad_WP/OptionMonad.thy
```diff
@@ -88,6 +112,18 @@ lemma opt_pred_disj:
   "((P1 |< hp) p \<or> (P2 |< hp) p) = (((P1 or P2) |< hp) p)"
   by (fastforce simp: pred_disj_def split: option.splits)
 
+lemma opt_predD:
+  "(P |< proj) x \<Longrightarrow> \<exists>y. proj x = Some y \<and> P y"
+  by (clarsimp split: option.splits)
+
+lemma opt_pred_unfold:
```
lsf37, line 119: This should be called `opt_pred_assoc`, I think. If there is no other reason against, it should be oriented the other way around, and potentially be `[simp]` so that parentheses are automatically removed (same as for `+` etc).

lsf37: Hm, this is a bit more subtle than a standard `assoc`, maybe `opt_pred_map_assoc` to express that two operators are involved here. Maybe even `opt_pred_map_distrib`.

lib/Monad_WP/OptionMonad.thy
```diff
@@ -88,6 +112,18 @@ lemma opt_pred_disj:
...
+  "(P |< f |> g)=  ((P |< g) |< f)"
+  by (fastforce simp: opt_map_def split: option.splits)
+
+lemma opt_pred_unfold_proj:
```
lsf37, line 123: This one actually unfolds, so that is fine. Not sure about the use site, but I'm tempted to add a lemma with `o` for this. Is this true `P |< f ||> g = P o g |< f`?

lsf37 (approved):
Nice additions. My comments for making things `[simp]` might break proofs (hopefully in a good way), but I think it would be worth investigating.

lsf37:
> Just out of curiosity, is it possible to add fancier symbols for `opt_map` and `opt_pred`? Does Isabelle or jEdit only work with a certain character set? I don't have any particular symbol in mind, but if these symbols will be everywhere in our proofs soon, it'd be nice to have it look as nice and coherent as possible

You can see the symbols the font supports in the symbols panel in jEdit somewhere. This syntax is already heavily used, though, so while changing it is possible, it would be a bunch of work, and we'd want a compatibility layer, because this is distributed with AutoCorres, so has users outside l4v.

lib/Monad_WP/OptionMonad.thy
mktnk3: It is actually the one below that I had the need to write, and this is a full `opt_map` version of it.

The idea for having this and the one below is to tweak what sits lhs and rhs of the top level `opt_pred` (`|<`) so that the rhs is a particular projection you need. Then we can combine it with another `opt_pred` statement via `opt_pred_conj`, etc., and it would be nice if we can do some reasoning inside the lhs of `opt_pred`.
So, I'm roughly aiming to have something like this:
`(P |< f |> g) and (P' |< f) = ((P |< g and P') |< f)`
`(P |< f ||> proj) and (P' |< f) = ((P o proj and P') |< f)`

mktnk3:
Thanks for all the feedback. I am happy to add these lemmas to the simp set as suggested and update the proof, to start using them.

lib/Monad_WP/OptionMonad.thy
mktnk3: > I'm tempted to add a lemma with `o` for this. Is this true `P |< f ||> g = P o g |< f`?

Ah, that should work, yes.

mktnk3:
> > Just out of curiosity, is it possible to add fancier symbols for `opt_map` and `opt_pred`? Does Isabelle or jEdit only work with a certain character set? I don't have any particular symbol in mind, but if these symbols will be everywhere in our proofs soon, it'd be nice to have it look as nice and coherent as possible
> 
> You can see the symbols the font supports in the symbols panel in jEdit somewhere. This syntax is already heavily used, though, so while changing it is possible, it would be a bunch of work, and we'd want a compatibility layer, because this is distributed with AutoCorres, so has users outside l4v.

Also, you probably don't want anything lengthy to type for these two because you'd be using them quite heavily. `opt_map` already so, and, as for `opt_pred`, the intention is to replace `pred_map` and `pred_map` has more than 2000 occurrences.

But then, I can see even from the discussion so far on this PR that interleaving `|<` and `|>` can be confusing for quick scanning. Maybe adjusting the operator precedence so that we use parentheses on the both sides might be helpful (if we are happy with more parentheses).

lib/Monad_WP/OptionMonad.thy
lsf37: Those look both good. I think you could prove them still with the rule oriented in the other direction (this one, the one below is fine).  Maybe not, depending on how it is applied. Happy to leave that to your experience with the rule.

lib/Monad_WP/OptionMonad.thy
```diff
@@ -62,6 +62,30 @@ lemma opt_map_upd_Some:
 
 lemmas opt_map_upd[simp] = opt_map_upd_None opt_map_upd_Some
 
+lemma opt_map_Some_comp:
+  "f ||> h o g = f ||> g ||> h"
+  by (fastforce simp: opt_map_def split: option.split)
+
+lemma opt_map_unfold_l:
+  "(f |> g) x = (id |> g) (f x)"
```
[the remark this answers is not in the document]
mktnk3: Yes, I will flip this one.

lsf37:
> But then, I can see even from the discussion so far on this PR that interleaving `|<` and `|>` can be confusing for quick scanning. Maybe adjusting the operator precedence so that we use parentheses on the both sides might be helpful (if we are happy with more parentheses).

I don't remember if we discussed this already, but would `?<` be better syntax? It's a bit further away from `|>` visually and the `?` would be for predicate. I.e. `P ?< f |> g` or `P <? f |> g`. We could use \triangle for opt_map, but I'm not sure it helps that much. Happy to try a few more variants.

You are probably right that we should have parentheses on the right even with `?`, i.e. something like `P ?< (f |> g)`.

This whole `?` is just an idea, I'm not sure it's really better -- what do people think?

lib/Monad_WP/OptionMonad.thy
```diff
@@ -88,6 +112,18 @@ lemma opt_pred_disj:
   "((P1 |< hp) p \<or> (P2 |< hp) p) = (((P1 or P2) |< hp) p)"
   by (fastforce simp: pred_disj_def split: option.splits)
 
+lemma opt_predD:
+  "(P |< proj) x \<Longrightarrow> \<exists>y. proj x = Some y \<and> P y"
```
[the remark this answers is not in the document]
mktnk3: Here is the `E` version:
```
lemma opt_predE:
  "⟦(P |< proj) x; ⋀y. ⟦proj x = Some y; P y⟧ ⟹ R⟧ ⟹ R"
  by (clarsimp split: option.splits)
```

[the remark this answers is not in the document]
lsf37: Having both is good.

michaelmcinerney:
I don't have nearly as much experience with these things, so don't take these suggestions too seriously, but maybe `\<box>` or `\<Down>`? (It seems like the box is used in `sep_algebra` so I don't know if that conflicts.) `opt_map` is like chaining heaps together, so the arrow there seems good. But `opt_pred` is like testing a property *of* a heap, so a left or right arrow doesn't seem as appealing to me there. Maybe something that's more reminiscent of function application would be good? Who knows.

I agree that the parentheses should be on the right (as in Gerwin's comment).

And as an aside, it's been ages since I've looked in the Symbols list. It's atrocious. How on earth are we ever going to compete with Agda?!

lsf37:
> I don't have nearly as much experience with these things, so don't take these suggestions too seriously, but maybe `\<box>` or `\<Down>`? 

`\box` and `\Down` both already have common meanings in CS (in separation logic, temporal+modal logic and operational semantics), they would not be good to use, esp since we do use operational semantics in C and we might use temporal logic for concurrency at some point.

> (It seems like the box is used in `sep_algebra` so I don't know if that conflicts.) 

That is the separation logic meaning (empty heap, also in scope in CRefine)

> `opt_map` is like chaining heaps together, so the arrow there seems good. But `opt_pred` is like testing a property _of_ a heap, so a left or right arrow doesn't seem as appealing to me there. Maybe something that's more reminiscent of function application would be good? Who knows.

Yes, `opt_pred` is the one that is hard to figure out. I haven't really gotten up with anything good either yet.


> And as an aside, it's been ages since I've looked in the Symbols list. It's atrocious. How on earth are we ever going to compete with Agda?!

Those symbols appeared several years (decade?) before Agda was a thing, so it's more like Agda competed and overtook. But we're conflating jEdit and Isabelle. Isabelle doesn't care about symbols, it just sees strings of the form `\<symbol>`. It's a question of font and editor what that gets mapped to. And jEdit is indeed not exactly at the forefront of that wave.

mktnk3:
It looks like adding the lemmas to the simp set doesn't break anything that is currently in the repo. So that's good.

One argument against using `?` in the notation is the clash with the schematics. It might be hard to read in the find_theorem output.

`opt_pred` is like predicate/function application but on an option value. So, my initial thinking was to have something as small as possible just to mark the distinction between the predicate on the left and the projection on the left, hopefully. Putting the rhs in the parentheses makes sense as function application, so I will adjust the priority and let's see how it feels.

mktnk3:
> Thinking of it as function application, we could use something like `$`, which we already have in Haskell for function application. `<$>` is a bit long, how about `<$`? We could go even further and do something like `P $[ f |> g ]`, i.e. make the right brackets part of the syntax. This one looks a bit too much like a list, but maybe there is a variation that makes sense.

I quite like `<$`. Making it mixfix as you suggest might also be a way to go. Even with normal parentheses, because adjusting the priority doesn't impose us to put parentheses around the rhs all the time. We don't need them when no `|>` or `||>` is visible in rhs. That means when rhs is a single level (abbreviated) heap we don't need parentheses, and this is actually the majority of the case we have...I did change the operator priority but it required hardly any change with the proof.

So, with what is said in the above, we would have to write `P <$ (scs_of s)`, while what we have now is `P <$ scs_of s`.

mktnk3:
I pushed the version with adjusted `opt_pred` priority (that didn't require proof update).

I think I will try out the `P <$ (scs_of s)` mixfix option next, probably at some point tomorrow morning. I'll wait for further suggestions/comments, if any, for now.

lib/Monad_WP/OptionMonad.thy
```diff
@@ -62,6 +62,30 @@ lemma opt_map_upd_Some:
...
+  "f ||> h o g = f ||> g ||> h"
+  by (fastforce simp: opt_map_def split: option.split)
+
+lemma opt_map_fold_l:
+  "(id |> g) (f x) = (f |> g) x"
+  by (clarsimp simp: opt_map_def)
+
+(* LHS is the same as (f ||> id) *)
+lemma opt_map_unit_r[simp]:
+  "(f |> Some) x = f x"
```
lsf37, line 75: Sorry for spotting this only now, but it would make sense to eta-contract this one to `f |> Some = f`. This would mean it can apply in the middle of a chain as well.

lib/Monad_WP/OptionMonad.thy
```diff
@@ -62,6 +62,30 @@ lemma opt_map_upd_Some:
...
+  by (clarsimp simp: opt_map_def split: option.split)
+
+lemma opt_map_unit_l[simp]:
+  "(Some |> f) x = f x"
```
lsf37, line 79: Can also eta contract

lib/Monad_WP/OptionMonad.thy
```diff
@@ -62,6 +62,30 @@ lemma opt_map_upd_Some:
...
+  by (clarsimp simp: opt_map_def split: option.split)
+
+lemma opt_map_zero_l[simp]:
+  "(Map.empty |> g) x = None"
+  by (clarsimp simp: opt_map_def)
+
+lemma opt_map_zero_r[simp]:
+  "(f |> Map.empty) x = None"
```
lsf37, line 87: And in the spirit of eta contraction for more applicability, this one would become `f |> Map.empty = Map.empty`, which shows that `empty` absorbs, which is also a nice algebraic property to have (and `Some` as the identity).

lsf37 (approved):

lib/Monad_WP/OptionMonad.thy
lsf37: If we don't have it yet, this suggest we might also want `Map.empty |> f = Map.empty` in the simp set.

lsf37:
It's Ok to merge this now and do the tweaks separate. Up to you.

lib/Monad_WP/OptionMonad.thy
mktnk3: Ah, yes, that makes sense. I was hoping these rules to be something of algebra-like, too.

mktnk3:
> It's Ok to merge this now and do the tweaks separate. Up to you.

Ok, I think I will merge this now as is and give a bit more time about `opt_pred` notation discussion/tweak, and I will include the eta contraction change there as well. I actually have a next lot of AInvs changes, too, so it can go there, too.

mktnk3:
Oops, I meant to squash but forgot.

lsf37:
> Oops, I meant to squash but forgot

I can sqash on `rt` for you if you like (needs admin rights)

mktnk3:
If it is easy, please do. Squashing 3rd with 1st, and 4th & 5th with the 2nd, would be good (So one commit each for opt_map and opt_pred). Thanks.

mktnk3:
> If it is easy, please do. Squashing 3rd with 1st, and 4th & 5th with the 2nd, would be good (So one commit each for opt_map and opt_pred). Thanks.

Or all together in one...

lsf37:
> If it is easy, please do. Squashing 3rd with 1st, and 4th & 5th with the 2nd, would be good (So one commit each for opt_map and opt_pred). Thanks.

Done!

lib/Monad_WP/OptionMonad.thy
mktnk3: I added this one to the simp set as well, but it seems that this needs to be in the opposite direction. I will fix it in my next batch. It affects my WIP only, I am pretty sure :)

lsf37: The test worked fine after the merge, so all good.

PR #327, opened 2021-08-15
testboard improvements

lsf37:
A few small improvements:

- use https URLs
- less noisy output
- add dry-run option
- print GitHub commit URL

The https URLs are needed because the test runner does not have ssh access to GitHub.

With this the testboard should work as before, and you have testboard access when you have push access to the gh-testboard repo.

(added lots of people as reviewers mostly to make sure you are aware this exists now again)

### DIFF
--- a/misc/testboard/testboardpush
+++ b/misc/testboard/testboardpush
@@ -53,12 +53,15 @@ def subprocess_output(*args, **kwargs):
         return out
 
 # github remotes are of the form git@github.com:user/repo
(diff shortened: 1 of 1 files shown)


lsf37:
It occurs to me that, since the trigger is now on GitHub, we could move away again from having each testboard commit being its own branch. Then again, we don't really need to, and with branches we could add something automatic that cleans up old branches after a few months, which we can't do with accumulating commits.

Anyway, for another time.

mbrcknl (approved):

PR #330, opened 2021-08-16
more opt map lemmas: updates

mktnk3:
Another small addition to our `opt_map` lemma collection.
For state updates and others. The eta_contraction tweak from the last PR is also included.

### DIFF
--- a/lib/Monad_WP/OptionMonad.thy
+++ b/lib/Monad_WP/OptionMonad.thy
@@ -71,21 +71,21 @@ lemma opt_map_fold_l:
   by (clarsimp simp: opt_map_def)
 
 (* LHS is the same as (f ||> id) *)
-lemma opt_map_unit_r[simp]:
-  "(f |> Some) x = f x"
-  by (clarsimp simp: opt_map_def split: option.split)
+lemma opt_map_Some_id_r[simp]:
+  "f |> Some = f"
+  by (fastforce simp: opt_map_def split: option.split)
 
-lemma opt_map_unit_l[simp]:
-  "(Some |> f) x = f x"
+lemma opt_map_Some_id_l[simp]:
+  "Some |> f = f"
   by (clarsimp simp: opt_map_def split: option.split)
 
 lemma opt_map_zero_l[simp]:
-  "(Map.empty |> g) x = None"
+  "Map.empty |> g = Map.empty"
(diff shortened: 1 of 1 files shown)


lib/Monad_WP/OptionMonad.thy
```diff
@@ -60,32 +68,66 @@ lemma opt_map_upd_Some:
...
-lemmas opt_map_upd[simp] = opt_map_upd_None opt_map_upd_Some
+lemma opt_map_Some_upd_Some:
+  "f(x \<mapsto> v) ||> g = (f ||> g)(x \<mapsto> g v)"
+  by (simp add: opt_map_upd_Some)
+
+lemmas opt_map_upd[simp]
+  = opt_map_upd_None opt_map_upd_Some opt_map_Some_upd_Some
+
+lemma opt_map_upd_triv[simp]:
+  "t k = Some x \<Longrightarrow> (t |> f)(k := f x) = t |> f"
```
lsf37, line 79: This is potentially dangerous as a `[simp]` rule for performance, because it will always try to prove the precondition when the pattern occurs, but it is not that likely for that precondition to actually be true.

Might be better to collect this and the next two lemmas in a `lemmas` set, so that they are easy to add to a `simp` locally when they are needed.

lsf37 (approved):
:+1:

lib/Monad_WP/OptionMonad.thy
mktnk3: ah, ok, true. I'll remove from the simp set. This is quite useful for state updates etc but probably local enough.

lib/Monad_WP/OptionMonad.thy
```diff
@@ -60,32 +68,66 @@ lemma opt_map_upd_Some:
...
 
 lemma opt_map_zero_r[simp]:
-  "(f |> Map.empty) x = None"
-  by (clarsimp simp: opt_map_def split: option.split)
+  "f |> Map.empty = Map.empty"
+  by (fastforce simp: opt_map_def split: option.split)
+
+lemma opt_map_Some_eta_fold:
+  "f |> (\<lambda>x. Some (g x)) = f ||> g"
+  by (simp add: o_def)
```
mktnk3, line 130: This is handy but loops with `o_def`.

corlewis (approved):

PR #336, opened 2021-08-19
`cur_sc_tcb`

michaelmcinerney:
This modifies the definition of `cur_sc_tcb` so that we do not need to assume that the scheduler action is resume current thread in as many places. I've attempted to remove this condition from all preconditions in lemmas before AInvs.thy.

I've finished AInvs for the domain time fix. This change to `cur_sc_tcb` works well. I'll remove `schact_is_rct` from a bunch of places in DetSchedSchedule as part of the domain time work

### DIFF
--- a/proof/invariant-abstract/Invariants_AI.thy
+++ b/proof/invariant-abstract/Invariants_AI.thy
@@ -1276,10 +1276,8 @@ definition
   "cur_tcb s \<equiv> tcb_at (cur_thread s) s"
 
 definition
-  "cur_sc_tcb s \<equiv>
-     sc_tcb_sc_at
-       (\<lambda>t. scheduler_action s = resume_cur_thread \<longrightarrow> t = Some (cur_thread s))
-       (cur_sc s) s"
+  "cur_sc_tcb s
+    \<equiv> (scheduler_action s = resume_cur_thread) \<longrightarrow> sc_tcb_sc_at (\<lambda>t. t = Some (cur_thread s)) (cur_sc s) s"
 
 definition valid_machine_time_2 :: "time \<Rightarrow> time \<Rightarrow> bool" where
    "valid_machine_time_2 ct lmt \<equiv> lmt \<le> - getCurrentTime_buffer - 1 \<and> ct \<le> lmt"
--- a/proof/invariant-abstract/SchedContext_AI.thy
+++ b/proof/invariant-abstract/SchedContext_AI.thy
@@ -1365,7 +1365,7 @@ lemma tcb_sched_action_cur_sc_tcb [wp]:
   "\<lbrace>cur_sc_tcb\<rbrace> tcb_sched_action action thread \<lbrace>\<lambda>_. cur_sc_tcb\<rbrace>"
   by (wpsimp simp: tcb_sched_action_def set_tcb_queue_def)
 
-lemma reschedule_required_cur_sc_tcb [wp]:
+lemma reschedule_required_cur_sc_tcb_inv:
   "\<lbrace>cur_sc_tcb\<rbrace> reschedule_required \<lbrace>\<lambda>_. cur_sc_tcb\<rbrace>"
   by (wpsimp simp: reschedule_required_def set_scheduler_action_def tcb_sched_action_def
                    set_tcb_queue_def get_tcb_queue_def thread_get_def is_schedulable_def
@@ -1382,13 +1382,21 @@ lemma sched_context_bind_tcb_invs[wp]:
                     update_sched_context_valid_objs_same valid_ioports_lift
                     update_sched_context_iflive_update update_sched_context_refs_of_update
                     update_sched_context_cur_sc_tcb_None update_sched_context_valid_idle
-                    hoare_vcg_all_lift hoare_vcg_conj_lift | wp set_tcb_obj_ref_wp)+
+                    hoare_vcg_all_lift hoare_vcg_conj_lift
+                    reschedule_required_cur_sc_tcb_inv
+         | wp set_tcb_obj_ref_wp)+
   apply (fastforce simp: obj_at_def tcb_cap_cases_def tcb_st_refs_of_def is_sc_obj_def
                          pred_tcb_at_def sc_tcb_sc_at_def valid_sched_context_def
                          is_tcb valid_idle_def state_refs_of_def get_refs_def2
                    elim: ex_cap_to_after_update' delta_sym_refs valid_objs_valid_sched_context_size
                   dest!: symreftype_inverse' split: if_splits)
   done
+
+lemma reschedule_required_cur_sc_tcb[wp]:
+  "\<lbrace>\<top>\<rbrace> reschedule_required \<lbrace>\<lambda>_. cur_sc_tcb\<rbrace>"
+  by (wpsimp simp: reschedule_required_def set_scheduler_action_def tcb_sched_action_def
+                   set_tcb_queue_def get_tcb_queue_def thread_get_def is_schedulable_def
+                   cur_sc_tcb_def sc_tcb_sc_at_def obj_at_def)
 
 lemma maybe_sched_context_bind_tcb_invs[wp]:
   "\<lbrace>invs and (\<lambda>s. tcb_at tcb s \<and> (bound_sc_tcb_at (\<lambda>x. x \<noteq> Some sc) tcb s \<longrightarrow>
--- a/proof/invariant-abstract/Syscall_AI.thy
+++ b/proof/invariant-abstract/Syscall_AI.thy
@@ -475,8 +475,8 @@ lemma do_reply_invs[wp]:
   done
 
 lemma pinv_invs[wp]:
-  "\<lbrace>\<lambda>s. invs s \<and> ct_active s \<and> valid_invocation i s \<and>
-        scheduler_action s = resume_cur_thread\<rbrace>
+  "\<lbrace>\<lambda>s. invs s \<and> ct_active s \<and> valid_invocation i s
+        \<and> bound_sc_tcb_at (\<lambda>a. \<exists>y. a = Some y) (cur_thread s) s\<rbrace>
      perform_invocation blocking call can_donate i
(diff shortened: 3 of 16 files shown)


proof/invariant-abstract/Syscall_AI.thy
```diff
@@ -475,16 +475,16 @@ lemma do_reply_invs[wp]:
   done
 
 lemma pinv_invs[wp]:
-  "\<lbrace>\<lambda>s. invs s \<and> ct_active s \<and> valid_invocation i s \<and>
-        scheduler_action s = resume_cur_thread\<rbrace>
+  "\<lbrace>\<lambda>s. invs s \<and> ct_active s \<and> valid_invocation i s
+        \<and> bound_sc_tcb_at (\<lambda>a. \<exists>y. a = Some y) (cur_thread s) s\<rbrace>
```
michaelmcinerney, line 479: This new condition arises from a precondition in the `invs` rule for `send_ipc`, which wants to know that if the `can_donate` flag is true, then the `thread` argument has an sc. It seems like the `thread` argument is (currently) always the current thread (and I think we can discharge this precondition since the current thread is always bound to the current sc when `send_ipc` is called, though I'm not totally sure on this). It would make more sense to me for us to perform some sort of check before calling (a function which calls) `send_ipc`, but that's not how things currently are...

proof/invariant-abstract/Invariants_AI.thy
```diff
@@ -1276,10 +1276,8 @@ definition
   "cur_tcb s \<equiv> tcb_at (cur_thread s) s"
 
 definition
-  "cur_sc_tcb s \<equiv>
-     sc_tcb_sc_at
-       (\<lambda>t. scheduler_action s = resume_cur_thread \<longrightarrow> t = Some (cur_thread s))
-       (cur_sc s) s"
+  "cur_sc_tcb s
+    \<equiv> (scheduler_action s = resume_cur_thread) \<longrightarrow> sc_tcb_sc_at (\<lambda>t. t = Some (cur_thread s)) (cur_sc s) s"
```
lsf37, line 1280:
```suggestion
  "cur_sc_tcb s
    \<equiv> scheduler_action s = resume_cur_thread \<longrightarrow> sc_tcb_sc_at (\<lambda>t. t = Some (cur_thread s)) (cur_sc s) s"
```

Could potentially also write
```
    \<equiv> scheduler_action s = resume_cur_thread \<longrightarrow> sc_tcb_sc_at ((=) Some (cur_thread s)) (cur_sc s) s"
```
but that will turn the equality around, i.e make it into `Some .. = t`. Potentially less nice in the proofs.

proof/invariant-abstract/SchedContext_AI.thy
```diff
@@ -1365,7 +1365,7 @@ lemma tcb_sched_action_cur_sc_tcb [wp]:
   "\<lbrace>cur_sc_tcb\<rbrace> tcb_sched_action action thread \<lbrace>\<lambda>_. cur_sc_tcb\<rbrace>"
   by (wpsimp simp: tcb_sched_action_def set_tcb_queue_def)
 
-lemma reschedule_required_cur_sc_tcb [wp]:
+lemma reschedule_required_cur_sc_tcb_inv:
```
lsf37, line 1368: Seems like a reasonably safe `[wp]` rule. Any specific reason for removing it? (not that it's a problem, just curious)

lsf37: Oh, I see the rule below it, yes, that is definitely better. But then we should actually delete the rule `reschedule_required_cur_sc_tcb_inv` entirely, because `reschedule_required_cur_sc_tcb` is always better to use.

proof/invariant-abstract/Syscall_AI.thy
lsf37: I assume `pinv_invs` is called further up in the chain. If you managed to repair those proof, then the condition is fine.

I would however suggest to write `(\<lambda>a. a ~= None)` in the statement instead. `wpsimp` and friends will turn this into the "exists" form automatically.

proof/invariant-abstract/Syscall_AI.thy
```diff
@@ -539,8 +539,8 @@ context Syscall_AI begin
 
 lemma pinv_tcb[wp]:
   "\<And>tptr blocking call can_donate i.
-    \<lbrace>\<lambda>s. invs s \<and> st_tcb_at active tptr s \<and> ct_active s \<and> valid_invocation i s \<and>
-         scheduler_action s = resume_cur_thread\<rbrace>
+    \<lbrace>\<lambda>s. invs s \<and> st_tcb_at active tptr s \<and> ct_active s \<and> valid_invocation i s
+         \<and> bound_sc_tcb_at (\<lambda>a. \<exists>y. a = Some y) (cur_thread s) s\<rbrace>
```
lsf37, line 543: `~= None` here as well.  Do we not have an abbreviation for just "is bound"? Might have been a different proof that I'm remembering.

lsf37 (approved):
So many removed preconditions, very nice.

I'd prefer the `~= None` form everywhere over `exists .. = Some ..`. Otherwise all good.

The Refine test failed because the AWS instance disappeared. I've restarted the proof run. (This is possible, because we're using cheap spot instances that are not guaranteed to remain online for longer stretches, i.e. they get zapped when demand increases. Usually that doesn't happen, but sometimes it does).

proof/invariant-abstract/SchedContext_AI.thy
michaelmcinerney: I couldn't get the nicer rule working in the proof where I did use the `inv` version, but I'll give it another shot

proof/invariant-abstract/Syscall_AI.thy
michaelmcinerney: We do have `bound`, which is an abbreviation for `_ ~= None`. But when I rewrite the rules and statements in preconditions in this way, I think the simplifier will turn this into the exists form, and then I can't use my rules (I have to use `not_None_eq` to get it back into a form that can be used). Any thoughts?

mktnk3: You can use `your_rule[simplified]` and then it should match, I think.

proof/invariant-abstract/SchedContext_AI.thy
michaelmcinerney: Yeah it's really not nice. It will end up making me deal with `sym_refs` and `valid_sched_context`, and while I might be able to fix it, it seems like it'll be a fairly complicated proof. Is it worth pursuing this further, or leaving it as it was?

proof/invariant-abstract/Syscall_AI.thy
michaelmcinerney: Thanks Miki, that did wok. It's a bit unfortunate that you have to take a rule already in the `wp` set and then mention it explicity with the simplified attribute. If this is the best way to go, that's fine, but I'd be curious if there is another way

michaelmcinerney: I have next to no experience with simp sets or building something like `wpsimp`, but if `wpsimp` is basically alternating `wp` and `simp`, doesn't it make sense for `wpsimp` and friends to be trying to work both plain and simplified versions of rules? Maybe this would kill performance

proof/invariant-abstract/Invariants_AI.thy
michaelmcinerney: I originally had it like this, but then something wasn't working and I became paranoid about precedence. But I've made the change you've suggested

proof/invariant-abstract/Syscall_AI.thy
lsf37: A few options:
- change `bound` to the current simp-normal-form, which is `Ex .. = Some`. This doesn't look nice in the definition of `bound`, but would probably be the most stable one.
- change the attribute at the lemma declaration from `[wp]` to `[simplified, wp]`. This would work, but is not generally nice style, because now the lemma that is written down is not the actual lemma statement.
- remove the rule `not_None_eq` from the simp set where we need these lemmas to apply. This is not very stable in general, and sometimes you might want the rule present because you want other `~= None` rewritten, but I reckon it would work here.

proof/invariant-abstract/SchedContext_AI.thy
lsf37: That seems really curious, because it should make things strictly easier. Maybe it makes another lemma not apply that expects to prove a `cur_sc_tcb` that is not there any more or something. Ok to leave it in for now, we can always remove it later.

If we do leave it in, we should prove the `_inv` lemma as a weakening of `reschedule_required_cur_sc_tcb`, though (`wpsimp` should do it).

proof/invariant-abstract/Invariants_AI.thy
lsf37: If you are unsure about precedence, you can use the `term` command with more parentheses to let Isabelle echo back to you what the "standard" parentheses are. I.e. if Isabelle leaves them out in the output and you have had them in the input, they are safe to leave out.

michaelmcinerney: Oooh that's a good idea, thanks

proof/invariant-abstract/Syscall_AI.thy
michaelmcinerney: I'd prefer the first option. Is this something we should be doing on master? If so, I might just make a JIRA issue for this. If just for RT, and others like the idea too, I might make this change first and then rebase this PR on top

lsf37: It's Ok to do on `rt` first, but it would be nice to make the `bound` part a separate commit if possible, so that we can apply it to master later as well.

michaelmcinerney: OK, I'll do it on `rt` first and make it separate commit(s)

michaelmcinerney:
Closed in favour of https://github.com/seL4/l4v/pull/343

PR #337, opened 2021-08-20
`bound`

michaelmcinerney:
This (seemingly) didn't require any proof fix


lsf37:
That looks like a refreshingly small change. Do the proofs pass without further change or are you waiting for the test to come back to determine that?

michaelmcinerney:
> That looks like a refreshingly small change. Do the proofs pass without further change or are you waiting for the test to come back to determine that?

I didn't test RISCV, but ARM AInvs and Refine passed on my machine, so I'm pretty sure there won't be any breakage

lsf37 (approved):
Cool.

mktnk3 (approved):
Nice

PR #338, opened 2021-08-21
use datatype package for abstract endpoint

mktnk3:
This adds a common selector `ep_queue` for `SendEP` and `RecvEP` in the definition of `endpoint` in Structures_A, consequently triggering the datatype package to generate a bunch of additional facts.

`get_ep_queue` and `valid_ep` are redefined using `ep_queue`. `queue_of` is removed because it is now a duplication of `ep_queue`.

Proof updates are mostly due to the above definition changes, and also from using `endpoint.case_eq_if` in place of locally defined lemmas to simplify the case distinction. Various versions are essentially the same and easily replaced by the `case_eq_if` lemma.

One interesting observation is that the Haskell `endpoint` is defined in a suboptimal order - first `RecvEP`, then `IdleEP`, followed by `SendEP`. It uses a shared queue selector, which is good, but we'd want to swap `RecvEP` and `IdleEP` so that the `case_eq_if` lemma will have `if IdleEP` in the top level (because this is what we mostly use).

### DIFF
--- a/proof/invariant-abstract/Invariants_AI.thy
+++ b/proof/invariant-abstract/Invariants_AI.thy
@@ -3501,8 +3501,8 @@ lemma has_reply_cap_cte_lift:
   by (simp add: cte_wp_at_caps_of_state, rule ctes)
 
 lemma pred_tcb_at_disj:
-  "(pred_tcb_at proj P t s \<or> pred_tcb_at proj Q t s) = pred_tcb_at proj (\<lambda>a. P a \<or> Q a) t s"
-  by (fastforce simp add: pred_tcb_at_def obj_at_def)
+  "pred_tcb_at proj (\<lambda>x. P x \<or> P' x) p s = (pred_tcb_at proj (\<lambda>x. P x) p s \<or> pred_tcb_at proj (\<lambda>x. P' x) p s)"
+  by (fastforce simp: pred_tcb_at_def obj_at_def)
 
 lemma dom_empty_cnode: "dom (empty_cnode us) = {x. length x = us}"
   unfolding empty_cnode_def
--- a/proof/invariant-abstract/Invariants_AI.thy
+++ b/proof/invariant-abstract/Invariants_AI.thy
@@ -693,10 +693,7 @@ definition
 where
   "valid_ep ep s \<equiv> case ep of
     IdleEP \<Rightarrow> True
-  | SendEP ts \<Rightarrow>
-      (ts \<noteq> [] \<and> (\<forall>t \<in> set ts. tcb_at t s) \<and> distinct ts)
-  | RecvEP ts \<Rightarrow>
-      (ts \<noteq> [] \<and> (\<forall>t \<in> set ts. tcb_at t s) \<and> distinct ts)"
+  | _ \<Rightarrow> (\<lambda>ts. ts \<noteq> [] \<and> (\<forall>t \<in> set ts. tcb_at t s) \<and> distinct ts) (ep_queue ep)"
 
 definition
   valid_sched_context :: "sched_context \<Rightarrow> 'z::state_ext state \<Rightarrow> bool"
--- a/proof/refine/ARM/IpcCancel_R.thy
+++ b/proof/refine/ARM/IpcCancel_R.thy
@@ -2258,7 +2258,7 @@ lemma cancelAllIPC_loop_body_valid_queues:
   apply (wpsimp wp: sts_valid_queues sts_st_tcb_at'_cases hoare_drop_imps)
   done
 
-lemma cancel_all_ipc_corres_helper:
+lemma cancelAllIPC_corres_helper:
   "distinct list \<Longrightarrow>
    corres dc
           ((\<lambda>s. \<forall>t \<in> set list. blocked_on_send_recv_tcb_at t s \<and> t \<noteq> idle_thread s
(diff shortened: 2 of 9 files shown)


proof/invariant-abstract/Invariants_AI.thy
```diff
@@ -693,10 +693,7 @@ definition
 where
   "valid_ep ep s \<equiv> case ep of
     IdleEP \<Rightarrow> True
-  | SendEP ts \<Rightarrow>
-      (ts \<noteq> [] \<and> (\<forall>t \<in> set ts. tcb_at t s) \<and> distinct ts)
-  | RecvEP ts \<Rightarrow>
-      (ts \<noteq> [] \<and> (\<forall>t \<in> set ts. tcb_at t s) \<and> distinct ts)"
+  | _ \<Rightarrow> (\<lambda>ts. ts \<noteq> [] \<and> (\<forall>t \<in> set ts. tcb_at t s) \<and> distinct ts) (ep_queue ep)"
```
lsf37, line 696: I'm almost tempted to use `let` here, but the lambda is actually quite nice.

lsf37 (approved):
I like it.

I'd be also happy with changing the order in Haskell, but the proof update might be more involved, because that will also change the order of goals in `cases` and case splits.

mktnk3:
> I like it.
> 
> I'd be also happy with changing the order in Haskell, but the proof update might be more involved, because that will also change the order of goals in `cases` and case splits.

Yes, it will be of a bit of work, so maybe later. JIRA issue here: https://sel4.atlassian.net/jira/software/c/projects/VER/issues/VER-1407

proof/invariant-abstract/Invariants_AI.thy
michaelmcinerney: This is cool. I'm not sure what `ts` is supposed to abbreviate or remind me of. Would `q` or something make more sense? I'm not sure

proof/refine/ARM/IpcCancel_R.thy
```diff
@@ -2336,74 +2336,12 @@ lemma in_send_ep_queue_TCBBlockedSend:
...
-  apply (clarsimp simp: valid_ep'_def)
-  apply (prop_tac "(t, EPRecv) \<in> state_refs_of' s epPtr")
-   apply (clarsimp simp: state_refs_of'_def obj_at'_def projectKOs)
-  apply (clarsimp simp: sym_refs_def obj_at'_def)
-  apply (fastforce simp: state_refs_of'_def tcb_bound_refs'_def st_tcb_at'_def get_refs_def2
-                         isBlockedOnReceive_def obj_at'_def projectKOs
-                  split: if_splits thread_state.splits)
-  done
-
-lemma IdleEP_versus_SendEP_and_RecvEP:
```
michaelmcinerney, line 2381: Glad to be rid of these ad hoc lemmas

michaelmcinerney (approved):
Awesome

proof/invariant-abstract/Invariants_AI.thy
mktnk3: I suppose that’s meant to mean “(potentially) many `t`’s” just like using `xs` for a list...

michaelmcinerney: Ahh. I guess that makes some sense

lsf37: The general convention for lists is an `s` suffix for `sequence`, pronounced as plural (many `t`s as @mktnk3 says). So `ts` is a list/sequence of `t`.

Should create a TODO to add that to the conventions file, I thought it was in there already or part of the [general Isabelle naming conventions](https://isabelle.systems/conventions/naming.html), but couldn't find it. It's a tradition from SML and Haskell.

proof/refine/ARM/IpcCancel_R.thy
```diff
@@ -2336,74 +2336,12 @@ lemma in_send_ep_queue_TCBBlockedSend:
...
-   apply (fastforce simp: valid_objs_def valid_obj_def dest!: invs_valid_objs)
-  apply (clarsimp simp: state_refs_of_def valid_ep_def split: option.splits)
-  apply (intro conjI impI allI; (fastforce simp: obj_at_def)?)
-  apply (prop_tac "(t, EPRecv) \<in> state_refs_of s epptr", clarsimp simp: state_refs_of_def)
-  apply (clarsimp simp: sym_refs_def dest!: invs_sym_refs)
-  apply (fastforce simp: state_refs_of_def)
-  done
+(* FIXME : Move *)
+lemma pred_tcb_at_disj:
+  "(pred_tcb_at proj (\<lambda>x. P x \<or> P' x) p s) = (pred_tcb_at proj (\<lambda>x. P x) p s \<or> pred_tcb_at proj (\<lambda>x. P' x) p s)"
```
corlewis, line 2341: Could we quickly do this move now? Or at least to Move_R?

proof/refine/ARM/IpcCancel_R.thy
corlewis: Definitely!

corlewis (approved):

proof/refine/ARM/IpcCancel_R.thy
mktnk3: Surely... and what I discovered is that we actually have this in Invariants_AI:
```
lemma pred_tcb_at_disj:
  "(pred_tcb_at proj P t s ∨ pred_tcb_at proj Q t s) = pred_tcb_at proj (λa. P a ∨ Q a) t s"
  by (fastforce simp add: pred_tcb_at_def obj_at_def)
```
My natural thinking would go in the way of splitting `pred_tcb_at` with the disjunction, but people have different ideas apparently...

mktnk3: And in its only use in AInvs is with `symmetric`... so I might as well flip it :D

PR #340, opened 2021-08-21
thread_state with datatype package (test)

mktnk3:
This PR is mainly for a demonstration purpose, for now, on JIRA issue: https://sel4.atlassian.net/browse/VER-1381

I went on and tried using the datatype package with `thread_state`. I added the `reply_object` selector as Corey did, and also added on discriminator just to see how it goes.

The AInvs proof breaks in a few places, in a way that has nothing to do with the added selector or discriminator. The breakage wasn't too difficult to fix it but it is quite obscure overall.

I did some digging and tracing, and it seems that one of the generated facts with `[dest]` attribute interacts badly with the simplifier. It is the `thread_state.distinct_disc` rule which looks like:
```
thread_state.distinct_disc:
      ?thread_state = Running ⟹ ?thread_state ≠ Inactive
      ?thread_state = Inactive ⟹ ?thread_state ≠ Running
      ?thread_state = Running ⟹ ¬ is_blocked_on_receive ?thread_state
      is_blocked_on_receive ?thread_state ⟹ ?thread_state ≠ Running
      ?thread_state = Running ⟹ ¬ is_BlockedOnSend ?thread_state
       ....
      ?thread_state = IdleThreadState ⟹ ¬ is_BlockedOnReply ?thread_state
      is_BlockedOnNotification ?thread_state ⟹ ?thread_state ≠ IdleThreadState
      ?thread_state = IdleThreadState ⟹ ¬ is_BlockedOnNotification ?thread_state
``` 
The breakage is often where `idle st` is present, which simplifies to `st = IdleThreadState`, and, from what I saw, this rule and congruence go around in loops.

With the proof fix being not too bad, we could just embrace this situation and go ahead with the change, or we could consider some alternatives.

I haven't checked Refine yet, so there is a chance it breaks.

### DIFF
--- a/spec/abstract/Structures_A.thy
+++ b/spec/abstract/Structures_A.thy
@@ -397,6 +397,8 @@ datatype thread_state
   | "reply_object (BlockedOnNotification _) = None"
   | "reply_object IdleThreadState = None"
 
+declare thread_state.distinct_disc[rule del]
+
 (* FIXME RT: generating the following discriminators and selectors automatically breaks
              unrelated proofs in strange ways *)
 (*
--- a/spec/abstract/Structures_A.thy
+++ b/spec/abstract/Structures_A.thy
@@ -383,24 +383,33 @@ datatype thread_state
   = Running
   | Inactive
   | Restart
-  | BlockedOnReceive obj_ref "obj_ref option" receiver_payload
+  | is_blocked_on_receive: BlockedOnReceive obj_ref (reply_object: "obj_ref option") receiver_payload
   | BlockedOnSend obj_ref sender_payload
   | BlockedOnReply obj_ref
   | BlockedOnNotification obj_ref
   | IdleThreadState
+  where
+  "reply_object (BlockedOnReply reply) = Some reply"
+  | "reply_object Running = None"
+  | "reply_object Inactive = None"
+  | "reply_object Restart = None"
+  | "reply_object (BlockedOnSend _ _) = None"
+  | "reply_object (BlockedOnNotification _) = None"
+  | "reply_object IdleThreadState = None"
 
 (* FIXME RT: generating the following discriminators and selectors automatically breaks
              unrelated proofs in strange ways *)
-
+(*
 fun reply_object where
   "reply_object (BlockedOnReceive _ reply_opt _) = reply_opt"
 | "reply_object (BlockedOnReply reply) = Some reply"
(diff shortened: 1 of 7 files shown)


spec/abstract/Structures_A.thy
```diff
@@ -383,24 +383,33 @@ datatype thread_state
...
   | BlockedOnNotification obj_ref
   | IdleThreadState
+  where
+  "reply_object (BlockedOnReply reply) = Some reply"
+  | "reply_object Running = None"
+  | "reply_object Inactive = None"
+  | "reply_object Restart = None"
+  | "reply_object (BlockedOnSend _ _) = None"
+  | "reply_object (BlockedOnNotification _) = None"
+  | "reply_object IdleThreadState = None"
```
lsf37, line 398: Would it be possible to use a default rule here? (I'm not sure any more which preprocessor is used for these)
```isabelle
"reply_object (BlockedOnReply reply) = Some reply"
   | "reply_object _ = None"
```

lsf37:
Nice detective work! That dest rule does not actually look very useful tbh, so we could also remove it from `dest` directly after the definition.

spec/abstract/Structures_A.thy
mktnk3: I’d think yes, I’ll have a go.

mktnk3:
> That dest rule does not actually look very useful tbh, so we could also remove it from `dest` directly after the definition.

I agree, I think we should remove it (it would be good if the package doesn’t give it `dest` attribute...). It seems that the proof slowed down a fair bit, possibly because of this rule but maybe also something else. I’ll remove this rule and see how it compares.

spec/abstract/Structures_A.thy
mktnk3: We can’t use the default rule here, actually.

mktnk3:
It seems I forgot to save one change, so the timeout is probably because of that.

mktnk3:
I removed this rule from `dest` and dropped the proof fix commit. And AInvs checks fine with that.

spec/abstract/Structures_A.thy
lsf37: Ok, good to know.

lsf37:
Closing the PR since we're not planning to merge it as is. Issue #471 links back to it in case we want to revive it.

PR #344, opened 2021-08-24
Ipc_R cleanup

mktnk3:
so that we stay with a nice decreasing trend :)

It turns out that the two of the three remaining sorries in Ipc_R are not needed. This PR deletes them.

The last sorry in Ipc_R is the one for `refill_unblock_corres`, which itself is done but waiting for the spec update. So we are almost finished with Ipc_R.


michaelmcinerney (approved):

lsf37 (approved):
:+1:

Best way to solve sorries :-)

corlewis (approved):

PR #346, opened 2021-08-25
sched_context_resume cleanup

mktnk3:
This PR contains a further cleanup on DetSchedSchedule_AI, in particular, about `sched_context_resume`.

We know that, with `valid_sched` in the precondition, after calling `sched_context_resume` on a thread, the thread being schedulable implies it is ready. We had several versions of lemmas essentially stating the same thing. Those are consolidated into `sched_context_resume_schedulable_imp_ready`.

Also, the existing lemmas mostly argue for "ready and sufficient" but it is actually only the readiness that matters (that could have changed) under `valid_sched`. We carry around separately (as part of `valid_sched`) in `valid_refills` the fact that "active implies sufficient".

Somewhat related, I think it probably makes sense to modify the notion of `released` in our proof to drop the sufficiency condition, so that it means "active and ready" only. We have `valid_sched` which gives us the guarantee that "active implies sufficient" separately, and in C, `sc_released` asserts sufficiency, so it is not actually a guard for it. Unless we are asking too much in `valid_sched`, which is unlikely (I hope).

A few other minor cleanups are also contained in this PR. It might help to review commit by commit.

### DIFF
--- a/proof/invariant-abstract/DetSchedSchedule_AI.thy
+++ b/proof/invariant-abstract/DetSchedSchedule_AI.thy
@@ -4957,6 +4957,17 @@ lemma tcb_release_enqueue_valid_release_q[wp]:
   apply (erule sorted_tcb_release_enqueue_upd)
   by (auto simp: tcb_sc_refill_cfgs_2_def opt_map_simps map_join_simps vs_all_heap_simps)
 
+lemma tcb_release_enqueue_weak_valid_sched_action[wp]:
+  "\<lbrace>weak_valid_sched_action and scheduler_act_not tptr\<rbrace>
+   tcb_release_enqueue tptr
+   \<lbrace>\<lambda>_. weak_valid_sched_action\<rbrace>"
+  unfolding tcb_release_enqueue_def
+  apply (rule hoare_seq_ext[OF _ get_sc_time_sp'])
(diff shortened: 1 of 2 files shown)


proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -6790,12 +6806,12 @@ lemma tcs_valid_sched:
    apply (rule hoare_vcg_E_elim, wpsimp)
    apply (rule valid_validE_R)
    apply ((wpsimp wp: hoare_vcg_all_lift install_tcb_cap_ct_active
-                      hoare_vcg_imp_lift install_tcb_cap_timeout_faulted_tcb_at
+                       hoare_vcg_imp_lift install_tcb_cap_timeout_faulted_tcb_at
```
michaelmcinerney, line 6809: Indentation

michaelmcinerney (approved):

mktnk3:
Rebased and added some more cleanups on `sched_context_resume` and `valid_sched_action`, and also about `possible_switch_to` and `valid_ready_qs`. Happy to be deleting more lines than added ones :)

corlewis (approved):
Awesome cleanup, great to see so much consolidation!

PR #347, opened 2021-08-30
`us_to_ticks`

michaelmcinerney:
This makes `us_to_ticks` a definition, and axiomatizes some of the necessary properties, in a way that is hopefully consistent. If the axioms are all fine, I'd be happy to add a comprehensive comment which motivates each of the axioms. 

The sorry that I recently added to `Refine.thy` related to `valid_domain_list'` being true of the initial state is (easily) solved.

I was hoping that I'd be able to use a more uniform approach that gives no overflow and non zero properties for some collection of constants, but I wasn't able to find a way that was evidently consistent. Any ideas would be appreciated.

I was unsure of what to call `factor1` and `factor2`. Maybe `timer_frequency` would be suitable for `factor1`, but I'm not entirely sure. I wanted to avoid something as boring as `us_to_ticks_mult` and `us_to_ticks_div` if possible, but maybe the latter is fine for `factor2`.

### DIFF
--- a/spec/machine/ARM/MachineOps.thy
+++ b/spec/machine/ARM/MachineOps.thy
@@ -120,6 +120,7 @@ consts' max_us_to_ticks :: "64 word"
 
 end
 
+
 qualify ARM (in Arch)
 
 type_synonym ticks = "64 word"
@@ -128,6 +129,75 @@ type_synonym time  = "64 word"
 abbreviation max_time :: time where
   "max_time \<equiv> max_word"
 
+\<comment> \<open>The following notepad shows that the axioms introduced below, which provide various results
+    about several constants and their conversion via us_to_ticks, are consistent.\<close>
+
+notepad begin
+
+define ticks_per_timer_unit :: "64 word" where "ticks_per_timer_unit = of_nat 50"
+define timer_unit :: "64 word" where "timer_unit = of_nat 1"
+define kernelWCET_us :: "64 word" where "kernelWCET_us = of_nat 100"
+define MAX_PERIOD_US :: "64 word" where "MAX_PERIOD_US = 60 * 60 * 1000 * 1000"
+define \<mu>s_in_ms :: "64 word" where "\<mu>s_in_ms = 1000"
+
+have ticks_per_timer_unit_non_zero:
+  "ticks_per_timer_unit \<noteq> 0"
+  by (simp add: ticks_per_timer_unit_def)
+
+have MIN_BUDGET_bound: "2 * unat kernelWCET_us * unat ticks_per_timer_unit < unat max_time"
+  apply (subst unat_max_word[symmetric])
+  apply clarsimp
+  apply (prop_tac "unat kernelWCET_us \<le> 100")
+   apply (fastforce simp: kernelWCET_us_def)
+  apply (prop_tac "unat ticks_per_timer_unit \<le> 50")
+   apply (fastforce simp: ticks_per_timer_unit_def)
+  apply (rule_tac y="10000" in le_less_trans)
+   apply (rule_tac j1=200 and l1=50 in order.trans[OF mult_le_mono]; fastforce)
+  apply simp
+  done
+
+have getCurrentTime_buffer_bound:
+  "unat kernelWCET_us * unat ticks_per_timer_unit
+    + 5 * unat MAX_PERIOD_US * unat ticks_per_timer_unit
+   < unat max_time"
+  apply (subst unat_max_word[symmetric])
+  apply clarsimp
+  apply (rule_tac j1=5000 and l1="5 * 60 * 60 * 1000 * 1000 * 50"
+               in order.strict_trans1[OF add_le_mono])
+    apply (fastforce simp: kernelWCET_us_def ticks_per_timer_unit_def)
+   apply (fastforce simp: kernelWCET_us_def ticks_per_timer_unit_def MAX_PERIOD_US_def)
+  apply linarith
+  done
+
+have kernelWCET_pos': "0 < (kernelWCET_us * ticks_per_timer_unit) div timer_unit"
+  apply (clarsimp simp: word_less_nat_alt)
+  apply (subst unat_mult_lem' | subst unat_div
+         | fastforce simp: kernelWCET_us_def ticks_per_timer_unit_def timer_unit_def max_word_def)+
+  done
+
+have MIN_BUDGET_pos': "0 < 2 * ((kernelWCET_us * ticks_per_timer_unit) div timer_unit)"
+  apply (clarsimp simp: word_less_nat_alt)
+  apply (subst unat_mult_lem' | subst unat_div
+         | fastforce simp: kernelWCET_us_def ticks_per_timer_unit_def timer_unit_def max_word_def)+
+  done
+
+have domain_time_pos: "0 < ((15 * \<mu>s_in_ms) * ticks_per_timer_unit) div timer_unit"
+  apply (clarsimp simp: word_less_nat_alt)
+  apply (subst unat_mult_lem' | subst unat_div
+         | fastforce simp: \<mu>s_in_ms_def ticks_per_timer_unit_def timer_unit_def max_word_def)+
+  done
+
+have getCurrentTime_buffer_pos:
+  "0 < (kernelWCET_us * ticks_per_timer_unit) div timer_unit
+       + 5 * (MAX_PERIOD_US * ticks_per_timer_unit div timer_unit)"
+  apply (clarsimp simp: word_less_nat_alt)
+  apply (subst unat_add_lem'' | subst unat_mult_lem' | subst unat_div
+         | fastforce simp: kernelWCET_us_def MAX_PERIOD_US_def ticks_per_timer_unit_def
+                           timer_unit_def max_word_def)+
+  done
+
+end
+
 axiomatization
   kernelWCET_us :: ticks
 where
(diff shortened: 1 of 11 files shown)


lib/Word_Lib/Word_Lemmas_Internal.thy
```diff
@@ -416,4 +416,24 @@ lemma unat_minus_plus_one:
    using max_word_wrap apply blast
   by (metis less_is_non_zero_p1 word_less_nat_alt word_overflow_unat)
 
+lemma unat_max_word:
+  "2 ^ LENGTH('a) - 1 = unat (max_word :: 'a :: len word)"
```
lsf37, line 420: Wow, I would have thought we had a lemma like this already, but I can't find it either. Nice addition.

spec/machine/ARM/MachineOps.thy
```diff
@@ -143,25 +143,29 @@ definition
...
   "MAX_PERIOD_US \<equiv> 60 * 60 * 1000 * 1000"
 
-axiomatization
-  us_to_ticks :: "ticks \<Rightarrow> ticks"
-where
-  us_to_ticks_mono[intro!]: "mono us_to_ticks"
+definition "\<mu>s_in_ms = 1000"
+
+consts' factor1 :: ticks
+consts' factor2 :: ticks
```
lsf37, line 149: We should give these better names. I was just using them to explain the form the term should take. Maybe something like `ticks_per_timer` for `factor1` and `timer_unit` for `factor2`.

spec/machine/ARM/MachineOps.thy
```diff
@@ -143,25 +143,29 @@ definition
...
+
+definition
+  "us_to_ticks us = (us * factor1) div factor2"
+
+axiomatization where
```
lsf37, line 154: Cool, the new axioms all look safe to me. Good job!

To really make sure, we should separately (in a test theory or a notepad or something where the axioms are not in scope) show that they are satisfiable. I.e. pick concrete numbers `kernelWCET_us`, `factor1`, `factor2`, etc and show that we can prove these rules. Ok to do that outside this PR, but it would be good to commit to the repo.

At some point later I'd like to auto-generate that for the kernel config, then we can replace the axioms, but that is in the future.

spec/machine/MachineExports.thy
```diff
@@ -67,32 +67,42 @@ requalify_consts
   timerPrecision
   max_time
   getCurrentTime_buffer
+  factor1
+  factor2
```
lsf37, line 71: Do we actually need to export them? I might have missed them, but I don't think I saw their names mentioned in the rest of the proof (which is a good thing).

lsf37: (We should be able to export eg `factor1_non_zero` without exporting `factor1`)

spec/machine/MachineExports.thy
```diff
@@ -67,32 +67,42 @@ requalify_consts
...
+  factor1_non_zero
+  MIN_BUDGET_bound
+  getCurrentTime_buffer_bound
+  kernelWCET_pos'
+  MIN_BUDGET_pos'
+  domain_time_pos
+  getCurrentTime_buffer_pos
+  unat_max_word
+  unat_add_lem''
+  unat_mult_lem'
```
lsf37, line 94: These are in Word_Lib now and don't need to be exported here.

spec/machine/RISCV64/MachineOps.thy
```diff
@@ -109,25 +109,29 @@ and
...
    "MAX_PERIOD_US \<equiv> 60 * 60 * 1000 * 1000"
 
-axiomatization
-  us_to_ticks :: "ticks \<Rightarrow> ticks"
-where
-  us_to_ticks_mono[intro!]: "mono us_to_ticks"
+definition "\<mu>s_in_ms = 1000"
+
+consts' factor1 :: ticks
+consts' factor2 :: ticks
```
lsf37, line 115: naming (same as for ARM)

spec/machine/RISCV64/MachineOps.thy
```diff
@@ -136,7 +140,7 @@ axiomatization
 
 end_qualify
 
-context Arch begin global_naming RISCV64
+context Arch begin global_naming ARM
```
lsf37, line 143: This looks like a copy/paste error

lsf37 (commented):
A few small things to fix up, but nothing substantial.

It's nice to have these finally sorted out!

spec/machine/ARM/MachineOps.thy
lsf37: better `ticks_per_timer_unit` than `ticks_per_timer`

michaelmcinerney: Thanks, those sound good

spec/machine/ARM/MachineOps.thy
michaelmcinerney: Sure, I can have a go at that

spec/machine/MachineExports.thy
michaelmcinerney: I know that there is at least one lemma that is using at least one of these at the moment. But maybe a helper lemma I wrote later in the game could be used instead. I'll take another look

lsf37: Cool. If it is mentioned in a lemma statement, then I just missed it and no change needed.

spec/machine/ARM/MachineOps.thy
michaelmcinerney: I've pushed a new commit that has a first attempt at this. I wasn't sure how to add a `defines` line for a function: when I tried, the "argument" to the function was appearing as an unknown schematic variable. So I've got some lemma statements that are like an unfolding of the `us_to_ticks` mentioned in the axioms.

(I haven't made any other changes for the PR yet, but I'll get to that soon.)

I'm not sure where it would be best to have this. It's not dependent on the architecture, so it shouldn't be in `MachineOps.thy`. Maybe a separate file immediately after?

michaelmcinerney: Well, in saying that it isn't dependent on the architecture, `MAX_PERIOD_US` comes from the C config file, so I guess that's some sort of dependence, and we'd want to be using that number in this demonstration...

spec/machine/RISCV64/MachineOps.thy
```diff
@@ -491,4 +491,68 @@ definition setVSpaceRoot :: "paddr \<Rightarrow> machine_word \<Rightarrow> unit
   "setVSpaceRoot pt asid \<equiv> machine_op_lift $ setVSpaceRoot_impl pt asid"
 
 end
+
+lemma mult_le_mono':
+  "\<lbrakk>a \<le> a'; b \<le> b'; a' * b' \<le> c\<rbrakk> \<Longrightarrow> a * b \<le> (c :: nat)"
+  using le_trans mult_le_mono by blast
+
+notepad begin
```
lsf37, line 499: This is nice. For extra peace of mind, it'd be better to put this before the `axiomatization` and `consts'` statements that introduce `factor1` etc to make it really obvious that these aren't accidentally used in here (I don't think they can, but the more obvious the better).

Should probably also add a comment that this notepad shows that the axioms introduced later are consistent.

spec/machine/ARM/MachineOps.thy
lsf37: The two factors will eventually also come out of the kernel config and will depend on the platform, so I think that is Ok to leave in here.

spec/machine/RISCV64/MachineOps.thy
michaelmcinerney: Sure, I'll add a comment and put this before the `axiomatization`. 

Are you happy for it to be put in the arch dependent files? I can say in the comment that the proofs are currently independent of the architecture

lsf37: I think it's fine to duplicate this one. Usually we'd want to avoid duplication, but there is a lot of it in MachineOps already and now trying hard for this specific bit doesn't really add much more.

We probably do want a deeper reform of these definitions at some point, but not now.

spec/machine/MachineExports.thy
michaelmcinerney: I was able to remove the references to `factor1` and `factor2` (these have been renamed, too) so I've removed these from the `requalify_consts`

michaelmcinerney:
This has been updated with the notepad section for both ARM and RISCV, and addressing the other comments that Gerwin has made

lsf37 (approved):

spec/machine/MachineExports.thy
```diff
@@ -67,32 +67,38 @@ requalify_consts
...
   MAX_PERIOD_mult
-  MAX_PERIOD_mult'
+  ticks_per_timer_unit_non_zero
+  MIN_BUDGET_bound
+  getCurrentTime_buffer_bound
+  kernelWCET_pos'
+  MIN_BUDGET_pos'
+  domain_time_pos
+  getCurrentTime_buffer_pos
+  us_to_ticks_def
```
mktnk3, line 90: Do we need to export the definition of `us_to_ticks`? It is not meant to be available in the generic context, right?

spec/machine/MachineExports.thy
```diff
@@ -67,32 +67,38 @@ requalify_consts
...
+  getCurrentTime_buffer_no_overflow
+  us_to_ticks_mono
+  MIN_BUDGET_helper
+  \<mu>s_in_ms_def
```
mktnk3, line 94: This one, too?

mktnk3 (approved):
Thanks for sorting this out. This looks great. Just a few comments from me for further consistency, etc.

spec/machine/ARM/MachineOps.thy
```diff
@@ -120,6 +120,10 @@ consts' max_us_to_ticks :: "64 word"
 
 end
 
+lemma less_mono:
+  "\<lbrakk>a \<le> a'; b \<le> b'; a' + b' < c\<rbrakk> \<Longrightarrow> a + b < (c :: nat)"
+  by fastforce
```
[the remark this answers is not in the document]
michaelmcinerney: Removed

spec/machine/ARM/MachineOps.thy
```diff
@@ -128,40 +132,118 @@ type_synonym time  = "64 word"
 abbreviation max_time :: time where
   "max_time \<equiv> max_word"
 
+lemma mult_le_mono':
+  "\<lbrakk>a \<le> a'; b \<le> b'; a' * b' \<le> c\<rbrakk> \<Longrightarrow> a * b \<le> (c :: nat)"
+  using le_trans mult_le_mono by blast
```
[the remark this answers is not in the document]
michaelmcinerney: Removed

spec/machine/ARM/MachineOps.thy
```diff
@@ -184,53 +266,81 @@ lemma replicate_no_overflow:
...
-  done
+lemma us_to_ticks_helper:
+  "unat a * unat ticks_per_timer_unit \<le> unat max_time
+   \<Longrightarrow> unat (us_to_ticks a) \<le> unat a * unat ticks_per_timer_unit"
+  apply (clarsimp simp: us_to_ticks_def)
+  by (subst unat_div | subst unat_mult_lem' | simp)+
+
+lemma le_mono:
+  "\<lbrakk>a \<le> a'; b \<le> b'; a' + b' \<le> c\<rbrakk> \<Longrightarrow> a + b \<le> (c :: nat)"
+  by fastforce
```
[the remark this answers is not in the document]
michaelmcinerney: Removed

spec/machine/MachineExports.thy
michaelmcinerney: Nope, removed

spec/machine/MachineExports.thy
michaelmcinerney: This is needed to show that \<mu>s_in_ms is equivalent to `usInMs`. This is in the same boat as `MAX_RELEASE_TIME` in needing to be consolidated somewhere appropriate. Shall I leave it in here for now and make a JIRA issue for this constant business?

spec/machine/ARM/MachineOps.thy
```diff
@@ -184,53 +266,81 @@ lemma replicate_no_overflow:
...
+
+lemma getCurrentTime_buffer_no_overflow:
+  "unat kernelWCET_ticks + 5 * unat MAX_PERIOD < unat max_time"
```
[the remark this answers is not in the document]
michaelmcinerney: This is a bit of a complicated area. The constant `5` is currently used in both `getCurrentTime` and `MAX_RELEASE_TIME`, but it is coincidental that they both do use `5`. It is likely that we'll want to change these in the future, but also likely they'll need to be different values at some point. 

We could introduce an alias for this constant factor and use it in the axiomatisation and the definition of `getCurrentTime_buffer`, but its usefullness might be limited; we currently have a lot of lemmas that use `current_time_bounded n` for some `n` less than 5. We may want to (try to) make everything the same constant in lemma statements (or just remove the constant as a parameter to `current_time_bounded`), but I'm not sure whether this will be possible, easy, or all that helpful

spec/machine/MachineExports.thy
mktnk3: I see, that’s fair enough. Yes, making a JIRA issue would  be great.

spec/machine/ARM/MachineOps.thy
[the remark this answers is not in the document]
mktnk3: Thanks for the comment, and, as we discussed in the meeting, let’s leave these as is for now.

spec/machine/MachineExports.thy
michaelmcinerney: https://sel4.atlassian.net/browse/VER-1411

PR #348, opened 2021-09-01
Sel4pr523

mktnk3:


lsf37 (approved):
This will need a rebase after the force-push to `rt`, but otherwise all good. (There was also a gitlint error to fix up)

PR #349, opened 2021-09-02
arm_hyp cspec: accommodate new bitfield file

lsf37:
This is an update for seL4/seL4#378, which amongst other things factors out the definition of virq into its own file.

This change is only visible on ARM_HYP, and the new file does not exist for the other architectures, so we need to make KernelInc_C arch-dependent, which needs to happen in its own directory. Since the directory of the new file will also not exist on other architectures, but the ROOT file must mention it, we forcefully create that directory in the build Makefile.

Test with seL4/seL4#378


lsf37:
Closing this, because we managed to not split the corresponding bitfield definitions.

PR #350, opened 2021-09-04
`refillBudgetCheck_corres`

michaelmcinerney:
This proves `refillBudgetCheck_corres`.

To help prove the termination of `handle_overrun_loop`, a new conjunct has been added to `sp_valid_refills` which says that each refill has nonzero `r_amount`. The first commit contains all the changes for this addition. I also generalised all of the lemmas used in `refill_budget_check_valid_refills` which were previously specialised to `cur_sc`. 

There are a couple `no_fail` rules and one small `corres` rule which I found convenient for these proofs. I am happy to move them to Lib, or to rewrite these proofs without them, if that is preferred


proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -8994,27 +9210,34 @@ lemma handle_overrun_loop_head_bound:
...
                in hoare_pre_add[THEN iffD2, simplified pred_conj_def])
    apply (fastforce dest!: head_time_buffer_implies_no_overflow[rotated 1]
                      simp: vs_all_heap_simps)
-
   apply (clarsimp simp: handle_overrun_loop_body_def)
   apply (rule hoare_seq_ext[OF _ gets_sp])
   apply (rule hoare_seq_ext[OF _ refill_single_sp])
   apply (rule hoare_seq_ext[OF _ get_sched_context_sp])
-  apply (rule_tac B="\<lambda>_ s. ?post s
+  apply (rule_tac B="\<lambda>_ s. pred_map (\<lambda>cfg. unat (r_time (hd (scrc_refills cfg))) + 4 * unat MAX_PERIOD \<le> unat max_time)
```
michaelmcinerney, line 9235: Clean this up

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -7040,10 +7040,13 @@ method merge_refills_simple
...
-                 (sc_refill_cfgs_of s) sc_ptr
-        \<and> pred_map (\<lambda>cfg. 1 < length (scrc_refills cfg)) (sc_refill_cfgs_of s) sc_ptr\<rbrace>
-   merge_refills sc_ptr
+  "\<lbrace>\<lambda>s. if sc_ptr' = sc_ptr
+        then pred_map (\<lambda>cfg. refills_sum (scrc_refills cfg) = scrc_budget cfg)
+                      (sc_refill_cfgs_of s) sc_ptr
+             \<and> pred_map (\<lambda>cfg. 1 < length (scrc_refills cfg)) (sc_refill_cfgs_of s) sc_ptr
+        else pred_map (\<lambda>cfg. refills_sum (scrc_refills cfg) = scrc_budget cfg)
+                      (sc_refill_cfgs_of s) sc_ptr\<rbrace>
+   merge_refills sc_ptr'
```
mktnk3, line 7049: We’d want to refactor the common clause out of if-then-else here. And then having an implication `sc_ptr’ = sc_ptr --> ...` as the second precondition would be good.

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -7052,22 +7055,35 @@ lemma merge_refills_refills_sum:
...
-                 (sc_refill_cfgs_of s) sc_ptr
-        \<and> pred_map (\<lambda>cfg. 1 < length (scrc_refills cfg)) (sc_refill_cfgs_of s) sc_ptr\<rbrace>
-   merge_refills sc_ptr
+  "\<lbrace>\<lambda>s. if sc_ptr' = sc_ptr
+        then pred_map (\<lambda>cfg. refills_unat_sum (scrc_refills cfg) \<le> unat max_time)
+                      (sc_refill_cfgs_of s) sc_ptr
+             \<and> pred_map (\<lambda>cfg. 1 < length (scrc_refills cfg)) (sc_refill_cfgs_of s) sc_ptr
+        else pred_map (\<lambda>cfg. refills_unat_sum (scrc_refills cfg) \<le> unat max_time)
+                      (sc_refill_cfgs_of s) sc_ptr\<rbrace>
+   merge_refills sc_ptr'
```
mktnk3, line 7086: The same here. No need to repeat the common part.

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -7083,66 +7099,78 @@ lemma merge_refills_refills_unat_sum:
...
-                 (sc_refill_cfgs_of s) sc_ptr
-        \<and> pred_map (\<lambda>cfg. Suc 0 < length (scrc_refills cfg)) (sc_refill_cfgs_of s) sc_ptr\<rbrace>
-   merge_refills sc_ptr
+  "\<lbrace>\<lambda>s. if sc_ptr' = sc_ptr
+        then pred_map (\<lambda>cfg. r_time (scrc_refill_hd cfg) \<le> P (cur_time s))
+                      (sc_refill_cfgs_of s) sc_ptr
+             \<and> pred_map (\<lambda>cfg. Suc 0 < length (scrc_refills cfg)) (sc_refill_cfgs_of s) sc_ptr
+        else pred_map (\<lambda>cfg. r_time (scrc_refill_hd cfg) \<le> P (cur_time s))
+                      (sc_refill_cfgs_of s) sc_ptr\<rbrace>
+   merge_refills sc_ptr'
```
mktnk3, line 7108: Here, too.

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -7083,66 +7099,78 @@ lemma merge_refills_refills_unat_sum:
...
-                 (sc_refill_cfgs_of s) sc_ptr
-        \<and> pred_map (\<lambda>cfg. Suc 0 < length (scrc_refills cfg)) (sc_refill_cfgs_of s) sc_ptr\<rbrace>
-   merge_refills sc_ptr
+  "\<lbrace>\<lambda>s. if sc_ptr' = sc_ptr
+        then  pred_map (\<lambda>cfg. unat (r_time (scrc_refill_hd cfg)) \<le> P (cur_time s))
+                       (sc_refill_cfgs_of s) sc_ptr
+              \<and> pred_map (\<lambda>cfg. Suc 0 < length (scrc_refills cfg)) (sc_refill_cfgs_of s) sc_ptr
+        else pred_map (\<lambda>cfg. unat (r_time (scrc_refill_hd cfg)) \<le> P (cur_time s))
+                      (sc_refill_cfgs_of s) sc_ptr\<rbrace>
+   merge_refills sc_ptr'
```
mktnk3, line 7134: The same here.

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -7159,13 +7187,17 @@ lemma merge_refills_hd_r_amount:
...
-  "\<lbrace>\<lambda>s. pred_map (\<lambda>cfg. a \<le> r_amount (hd (scrc_refills cfg))) (sc_refill_cfgs_of s) sc_ptr
-        \<and> pred_map (\<lambda>cfg. refills_unat_sum (scrc_refills cfg) \<le> unat max_time)
-                    (sc_refill_cfgs_of s) sc_ptr\<rbrace>
-   refill_head_overlapping_loop sc_ptr
+  "\<lbrace>\<lambda>s. if sc_ptr' = sc_ptr
+        then pred_map (\<lambda>cfg. a \<le> r_amount (hd (scrc_refills cfg))) (sc_refill_cfgs_of s) sc_ptr
+             \<and> pred_map (\<lambda>cfg. refills_unat_sum (scrc_refills cfg) \<le> unat max_time)
+                         (sc_refill_cfgs_of s) sc_ptr
+        else pred_map (\<lambda>cfg. a \<le> r_amount (hd (scrc_refills cfg))) (sc_refill_cfgs_of s) sc_ptr\<rbrace>
+   refill_head_overlapping_loop sc_ptr'
```
mktnk3, line 7195: The same here.

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -7177,11 +7209,15 @@ lemma refill_head_overlapping_loop_hd_r_amount:
...
-   merge_refills sc_ptr
+  "\<lbrace>\<lambda>s. if sc_ptr' = sc_ptr
+        then pred_map (\<lambda>cfg. unat (r_time (scrc_refill_hd cfg)) + refills_unat_sum (scrc_refills cfg)
+                             \<le> unat max_time)
+                      (sc_refill_cfgs_of s) sc_ptr
+             \<and> pred_map (\<lambda>cfg. Suc 0 < length (scrc_refills cfg)) (sc_refill_cfgs_of s) sc_ptr
+        else pred_map (\<lambda>cfg. unat (r_time (scrc_refill_hd cfg)) + refills_unat_sum (scrc_refills cfg)
+                          \<le> unat max_time)
+                   (sc_refill_cfgs_of s) sc_ptr\<rbrace>
+   merge_refills sc_ptr'
```
mktnk3, line 7220: The same here. I will stop pointing out the same issues for the rest of the PR, but please refactor all the `if P then A & B else A` instances in this PR to `A & (P --> B)`. These `A`s are often quite big and repeating them affects readability, and it is always better to be clear about what is common and what is not. Also, when we need to make modifications in the future, we don’t want to be spending time on duplicating changes or going through all of them to see which was is the same or not.

michaelmcinerney: I've pushed a commit which rearranges the preconditions, but I'm still curious about this idea with `if_helper`. I think there are many situations where we could use this approach

proof/refine/ARM/SchedContextInv_R.thy
```diff
@@ -1352,6 +1354,544 @@ lemma scheduleUsed_corres:
...
+  apply (clarsimp simp: is_active_sc2_def obj_at_def opt_map_red)
+  done
+
+lemma refillBudgetCheck_corres:
+  "usage = usage'
+   \<Longrightarrow> corres dc ((\<lambda>s. sc_at (cur_sc s) s \<and> is_active_sc (cur_sc s) s
+                       \<and> valid_objs s
+                       \<and> pspace_aligned s \<and> pspace_distinct s)
+                  and (\<lambda>s. \<not> round_robin (cur_sc s) s \<and> valid_refills (cur_sc s) s))
+                 valid_objs'
```
mktnk3, line 1750: I am pretty sure this can be weakened to `valid_refills’`. Similarly for `handleOverrunLoop_corres` and so on. I did the same with `scheduleUsed_corres`, etc., in my `refillUnblockCheck_corres` PR.

spec/abstract/SchedContext_A.thy
```diff
@@ -304,13 +304,12 @@ where
...
-     new_head_amount \<leftarrow> return (r_amount (hd refills));
 
      when (usage' > 0 \<and> r_time (hd refills) < MAX_RELEASE_TIME) $ do
        sc \<leftarrow> get_sched_context sc_ptr;
-       period \<leftarrow> return (sc_period sc);
-       used \<leftarrow> return \<lparr>r_time = r_time (hd (sc_refills sc)) + period, r_amount = usage'\<rparr>;
-       update_refill_hd sc_ptr (r_time_update (\<lambda>t. t + usage') o (r_amount_update (\<lambda>m. m - usage')));
+       used \<leftarrow> return \<lparr>r_time = r_time (hd (sc_refills sc)) + sc_period sc, r_amount = usage'\<rparr>;
+       set_refill_hd sc_ptr \<lparr>r_time = r_time (refill_hd sc) + usage',
+                             r_amount = r_amount (refill_hd sc) - usage'\<rparr>;
```
mktnk3, line 312: I am not too sure of this change, to be honest. What is the reason behind it? If it is just aligning with Haskell, wouldn’t it be better to change Haskell to be similar to the abstract for this case, if you consider how the C code is?

proof/refine/ARM/SchedContextInv_R.thy
michaelmcinerney: Did you want to merge the `refillUnblockCheck_corres` PR first and then I can make the changes to these new `corres` rules?

spec/abstract/SchedContext_A.thy
michaelmcinerney: It was to ease the refinement. If you're happy for me to change this part of `refillBudgetCheck`, I'll get rid of the commits related to the change to the definition of `refill_budget_check` and will use two calls to `update_refill_hd/updateRefillHd` in `refill_budget_check/refillBudgetCheck`

michaelmcinerney: I suppose I could just leave the asbtract spec as it is and then use a rewrite rule for the `corres` rule. I might just go ahead with that

mktnk3: It would be helpful for later to change Haskell to use `updateRefillHd` instead of `setRefillHd` while we are here, wouldn’t it?

michaelmcinerney: Yeah, I'd like to leave the definition of `refill_budget_check` as it is in current `rt`, replace the call to `setRefillHd` in `refillBudgetCheck` with two calls to `updateRefillHd`, and then use a rewrite rule to help with the proof of `refillBudgetCheck_corres`

mktnk3: Great, that’ll be perfect :)

michaelmcinerney: The changes are here

https://github.com/michaelmcinerney/l4v/commits/refillBudgetCheck_corres3

I squashed the commits relevant to the `valid_refills` conjunct and the generalised helper lemmas into the first commit in that branch. I can force push that branch to this branch if that's better.

I think the best way forward would be to wait for the `refillUnblockCheck_corres` PR to be merged, and then I can rebase the linked branch over that, move the work in `SchedContextInv_R.thy` to `Schedule_R.thy`, and then update the guards in the new `corres` rules I have

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -8804,63 +8921,98 @@ lemma handle_overrun_loop_body_ordered_disjoint:
   apply (subst unat_add_lem'; fastforce simp: max_word_def window_def last_tl)
   done
 
+(* FIXME RT: keep? move? *)
+lemma hoare_imp_helper:
+  "\<lbrakk> \<lbrace>P and Q\<rbrace> f \<lbrace>\<lambda>rv s. R rv s\<rbrace>; f \<lbrace>\<lambda>s. \<not> Q s\<rbrace> \<rbrakk> \<Longrightarrow> \<lbrace>\<lambda>s. P s\<rbrace> f \<lbrace>\<lambda>rv s. Q s \<longrightarrow> R rv s\<rbrace>"
```
mktnk3, line 8926: We could keep this maybe but please give a bit more descriptive name.

proof/refine/ARM/SchedContextInv_R.thy
```diff
@@ -1352,6 +1354,544 @@ lemma scheduleUsed_corres:
...
+  apply (clarsimp simp: get_refills_def)
+  apply (wpsimp wp: get_sched_context_exs_valid)
+   apply (erule sc_atD1)
+  apply simp
+  done
+
+(* FIXME RT: keep? move to Lib? *)
+lemma corres_when3:
+  "\<lbrakk>G = G'; G \<Longrightarrow> corres_underlying sr nf nf' dc P P' a c\<rbrakk>
+   \<Longrightarrow> corres_underlying sr nf nf' dc P P' (when G a) (when G' c)"
```
mktnk3, line 1732: I don’t think we’d want to keep adding these weaker lemmas, to be honest. Surely you can replace this with `corres_if_split` and a few more steps?

proof/refine/ARM/SchedContextInv_R.thy
```diff
@@ -1352,6 +1354,544 @@ lemma scheduleUsed_corres:
...
+                  wp: update_sched_context_wp)
+  done
+
+(* FIXME RT: keep and move the following no_fail lemmas? *)
+lemma no_fail_skip:
+  "\<lbrakk>no_fail P f; f \<lbrace>P\<rbrace>; \<And>rv. no_fail P (g rv)\<rbrakk> \<Longrightarrow> no_fail P (f >>= (\<lambda>rv. g rv))"
+  by (wpsimp, blast+)
+
+lemmas no_fail_bind'
+  = no_fail_bind[rotated 2, where P=P and Q=P for P, simplified pred_conj_def, simplified]
```
mktnk3, line 1497: I suppose we can keep these but please don’t rotate `no_fail_bind’`. We don’t want such difference between `no_fail_bind` and `no_fail_bind’`, right? Please keep the precondition order the same.

mktnk3 (commented):
I understand that this will become a refreshed PR in the near future, but adding a few more comments so that cleanup can be done in the meantime.

proof/refine/ARM/SchedContextInv_R.thy
michaelmcinerney: Removed

proof/refine/ARM/SchedContextInv_R.thy
michaelmcinerney: Removed the `rotated 2`

proof/refine/ARM/SchedContextInv_R.thy
michaelmcinerney: Done

michaelmcinerney:
This has been rebased and updated, moving the lemmas which were in `SchedContextInv_R.thy` to `Schedule_R.thy` (as well as some others which were necessary, which has increased the size of the diff), and addressing all the comments received so far.

proof/refine/ARM/Schedule_R.thy
```diff
@@ -4210,6 +4209,1003 @@ lemma refillUnblockCheck_corres':
...
+   apply (corressimp corres: updateRefillTl_corres
+                       simp: refill_map_def)
+  apply (rule corres_split'[rotated 2, OF refill_full_sp refillFull_sp])
+   apply (corressimp corres: refillFull_corres)
+  apply (rule corres_if_split; (solves simp)?)
+   apply (corressimp corres: refillAddTail_corres)
+   apply (clarsimp simp: refill_map_def obj_at_simps opt_map_red)
+  apply (corressimp corres: updateRefillTl_corres simp: refill_map_def)
+  done
+
```
michaelmcinerney, line 4703: Everything in this block before here was just moved

proof/refine/ARM/Schedule_R.thy
```diff
@@ -4210,6 +4209,1003 @@ lemma refillUnblockCheck_corres':
...
+  apply (rule_tac Q="\<lambda>s'. sc_at' (ksCurSc s') s'" in corres_cross_add_guard)
+   apply (fastforce intro: sc_at_cross simp: state_relation_def)
+  apply (rule_tac Q="\<lambda>s'. is_active_sc' (ksCurSc s') s'" in corres_cross_add_guard)
+   apply (fastforce intro: is_active_sc'_cross simp: state_relation_def)
+  apply (clarsimp simp: handle_overrun_loop_body_def handleOverrunLoopBody_def)
+  apply (rule corres_split'[rotated 2, OF gets_sp getCurSc_sp])
+   apply (corressimp corres: getCurSc_corres)
+  apply (rule corres_split'[rotated 2, OF refill_single_sp refillSingle_sp])
+   apply (corressimp corres: refillSingle_corres)
+   apply (fastforce dest!: valid_objs'_valid_refills'
```
mktnk3, line 4779: I am guessing that these `valid_objs’_valid_refills’` (here and twice below) are probably not needed? If they are needed, I wonder where `valid_objs’` gets resurrected.

proof/refine/ARM/Schedule_R.thy
```diff
@@ -4210,6 +4209,1003 @@ lemma refillUnblockCheck_corres':
...
+        apply (rule scheduleUsed_corres)
+         apply simp
+        apply (clarsimp simp: refill_map_def sc_relation_def)
+       apply wpsimp
+      apply (wpsimp wp: updateRefillHd_valid_objs')
+     apply wpsimp
+    apply (wpsimp wp: updateRefillHd_valid_objs')
+   apply (clarsimp simp: is_active_sc_rewrite)
+  apply (fastforce intro: valid_objs'_valid_refills'
+                    simp: active_sc_at'_def is_active_sc'_def obj_at_simps opt_map_red)
```
mktnk3, line 5206: Does `valid_objs’` actually show up in the proofs here? If so, where does it come from I wonder?

mktnk3 (approved):
Great to see another big corres rule sorted.

It looks like the majority of the changes for AInvs, which is in DetSchedSchedule_AI, is about `cur_sc`, and that is not mentioned at all in the commit message. I think it would be helpful for the future if we update the commit message accordingly (or split the commit, which would be more helpful...).

proof/refine/ARM/Schedule_R.thy
```diff
@@ -4210,6 +4209,1003 @@ lemma refillUnblockCheck_corres':
...
+lemma nonOverLappingMergeRefills_valid_refills'[wp]:
+  "nonOverlappingMergeRefills scPtr \<lbrace>valid_refills' scPtr\<rbrace>"
+  apply (wpsimp simp: nonOverlappingMergeRefills_def)
+  apply (clarsimp simp: obj_at'_def)
+  done
+
+definition head_insufficient_loop_measure where
+  "head_insufficient_loop_measure sc_ptr
+     \<equiv> measure (\<lambda>(_, s). case kheap s sc_ptr of Some (Structures_A.SchedContext sc _)
+                                             \<Rightarrow> (length (sc_refills sc)))"
```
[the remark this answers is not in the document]
michaelmcinerney: Yours is on the Refine level, mine is on the abstract level. If `corres_whileLoop_abs` were used for the proof of `refillUnblockCheck_corres`, then this measure could be used there, too, but there's probably very little benefit in doing that

proof/refine/ARM/Schedule_R.thy
michaelmcinerney: Removed, including an instance in `scheduleUsed_corres`

proof/refine/ARM/Schedule_R.thy
michaelmcinerney: Removed

proof/refine/ARM/Schedule_R.thy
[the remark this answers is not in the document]
mktnk3: The point is that we have managed in duplicating essentially the same proofs for both abstract and concrete. They contain the same information for the purpose. We don’t want to have such duplications in maintaining because any change is needed in the future we would have to update both if we keep the both.

proof/refine/ARM/Schedule_R.thy
```diff
@@ -4210,6 +4209,1003 @@ lemma refillUnblockCheck_corres':
...
+                                                (sc_refill_cfgs_of s) (cur_sc s))
+                                \<and> pred_map (\<lambda>cfg. scrc_refills cfg \<noteq> [])
+                                            (sc_refill_cfgs_of s) sc_ptr"
+               and Q'="\<lambda>_ s'. valid_refills' scPtr s' \<and> active_sc_at' scPtr s' \<and> scPtr = ksCurSc s'"
+               in corres_split')
+     apply (corressimp corres: handleOverrunLoop_corres)
+     apply (fastforce intro: valid_refills_refills_unat_sum_equals_budget
+                       simp: vs_all_heap_simps cfg_valid_refills_def round_robin_def
+                             sp_valid_refills_def is_active_sc_rewrite[symmetric])
+    defer
```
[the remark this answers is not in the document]
michaelmcinerney: Removed

proof/refine/ARM/Schedule_R.thy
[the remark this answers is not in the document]
mktnk3: When I wrote the `refillUnblockCheck_corres` proof, `corres_whileLoop_abs` didn’t exist, just to mention.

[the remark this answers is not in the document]
michaelmcinerney: Would you like me to rewrite the proof of `handleOverrunLoop_corres` with `corres_whileLoop`?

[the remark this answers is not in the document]
michaelmcinerney: Oh, `headInsufficientLoop_corres`, it would be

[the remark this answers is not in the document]
mktnk3: I am not sure. What I think I know is that this measure probably deserves a more generic name, maybe we can name it to mean it’s about the length of refills.

And, since we have a crossing lemma for refill length, it probably makes sense to use the argument on the abstract side. So, we can do the consolidation later, not as a part of this PR. The next question is whether we should remove the `whileLoop` corres rule for termination on the concrete side so that we kind of force ourselves to do the termination on the abstract side. Maybe we don’t need to decide now.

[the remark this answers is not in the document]
michaelmcinerney: I am happy to rename the measure. Would you like me to do this as part of this PR? It was originally written for `headInsufficientLoop`, and it appears here only because I moved it.

I think `corres_whileLoop` is more general than `corres_whileleLoop_abs`: it seems possble for a concrete function to terminate, but to be in correspondence with an abstract function which does not terminate (roughly speaking). I cannot imagine that we'd come across such a situation in the l4v proofs, but it's worth mentioning.

proof/refine/ARM/Schedule_R.thy
```diff
@@ -4210,6 +4209,1003 @@ lemma refillUnblockCheck_corres':
...
+       apply fastforce
+      apply (corressimp corres: handleOverrunLoopBody_corres)
+     apply (wpsimp wp: handle_overrun_loop_body_no_fail)
+     apply (clarsimp simp: vs_all_heap_simps)
+    apply (intro hoare_vcg_conj_lift_pre_fix; (solves wpsimp)?)
+        apply (rule hoare_weaken_pre)
+         apply (rule_tac f=cur_sc in hoare_lift_Pf2)
+          apply wpsimp
+         apply wpsimp
+        apply wpsimp
```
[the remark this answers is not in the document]
michaelmcinerney: I really can't get `wps` to work. @corlewis do you have any tips?

I did experiment before with a little Eisbach method which does work

```
method wp_proj for proj :: "'a state => _" uses wp simp
   = (rule hoare_weaken_pre, rule hoare_lift_Pf2[where f=proj], wpsimp wp: wp simp: simp, wpsimp+)
```

It unfortunately requires you to write `apply (wp_proj "cur_sc :: det_state => _")` if you'd like to use it, and I can't see how to get rid of the type annotation. Plus, I'm sure it's less general than `wps`, so it's not ideal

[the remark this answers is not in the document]
michaelmcinerney: The `'a state` in the method declaration can be replaced with `_` and this will also work for the `kernel_state`. But still, I'm not that happy with this approach

proof/refine/ARM/Schedule_R.thy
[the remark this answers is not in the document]
mktnk3: Actually, it is enough if we know what terminates on the concrete side terminates on the abstract side as well. So, that suggests we should base ourselves on the concrete side, I suppose? I think we can leave this PR as is on this issue and deal with it separately.

proof/refine/ARM/Schedule_R.thy
mktnk3: 👍 I must have missed that one. While you are there, can you check one more thing? In `scheduleUsed_corres`, I think crossing of `valid_refills’` is now necessary and should be removed.

michaelmcinerney: I've removed that crossing, too

proof/refine/ARM/Schedule_R.thy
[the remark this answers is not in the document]
michaelmcinerney: Well, `corres_whileLoop_abs` is certainly proving more than what strictly needs to be proved, but this is done for convenience: Mitch had an attempt at `headInsufficientLoop_corres` with the concrete termination, and from memory, it was not all that pleasant. He had to introduce analogues of `refills_unat_sum` etc., which prompted me to think about abstract termination and no_fail. There's not much difference between abstract and concrete in the other loops we have, but my preference would be to have more of these general rules available, and to then select the best one for the proof at hand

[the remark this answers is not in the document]
mktnk3: So, it is the concrete version of `head_insufficient_length_at_least_two` that is hard to show, I suppose? It is quite obvious that `nonOverlappingMergeRefills` is refill length decreasing. And I see why he wanted to have crossing there. And *if* that had worked, one of my refill-length-decreasing proofs could have been used. We could be generally proving the “refill-length-decreasing” property for functions. Maybe. It is a bit surprising that there was no other way to cross that information.

This is not specifically about the case here, but the potential problem is that those more specific rules are actually less general, and more often than not things go in the opposite way - people pick their rule to guide the proof, and if the proof you happened to pick is not general enough, it might lead to a hard point or a dead-end. Also, the search process becomes harder and more costly with many similar but specialised rules. I guess this is an argument for having a smaller set of more general rules.

proof/refine/ARM/Schedule_R.thy
```diff
@@ -4210,6 +4209,1003 @@ lemma refillUnblockCheck_corres':
...
+  apply (rule corres_split'[rotated 2, OF get_sched_context_sp get_sc_sp'])
+   apply (corressimp corres: get_sc_corres)
+  apply (corressimp corres: corres_return_eq_same)
+  apply (fastforce simp: sc_relation_def obj_at_simps valid_refills'_def opt_map_red)
+  done
+
+end
+
+global_interpretation updateRefillTl: typ_at_all_props' "updateRefillTl scPtr f"
+  by typ_at_props'
```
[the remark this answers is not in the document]
michaelmcinerney: All the `typ_at_all_props'` interpretations in this file could be fairly easily moved into one block if we're happy to move some `crunches` around. 

But maybe we could even move a bunch more to one place. I tried, and I can get the `crunches` and the interpretation working for `refillBudgetCheck` at the end of `KHeap_R.thy`, so I'm pretty sure this is viable. @corlewis do you have any thoughts? Maybe we could create a file of its own for this. Of course this'd be moving a lot of things around, so should be done separately. Should I make a JIRA issue for this?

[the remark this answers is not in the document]
mktnk3: Well, moving to the top of this file before the first `context begin interpretation Arch .`, if possible, would be enough for minimising the number of arch interpretation. I don’t see much benefit in moving further beyond that point...

proof/refine/ARM/Schedule_R.thy
```diff
@@ -4210,6 +4205,920 @@ lemma refillUnblockCheck_corres':
...
+  unfolding readRefillReady_def readSchedContext_def ovalid_def
+  by (fastforce simp: obind_def opt_map_red obj_at'_def projectKOs
+                dest: use_ovalid[OF ovalid_readCurTime]
+               dest!: readObject_misc_ko_at'
+               split: option.split_asm)+
+
+lemma refillReady_wp':
+  "\<lbrace>\<lambda>s. sc_at' scp s \<longrightarrow>
+        P (((\<lambda>sc'. rTime (refillHd sc') \<le> ksCurTime s + kernelWCETTicks) |< scs_of' s) scp) s\<rbrace>
+    refillReady scp \<lbrace>P\<rbrace>"
```
corlewis, line 4224: tiny nitpick: indentation

proof/refine/ARM/Schedule_R.thy
```diff
@@ -4210,6 +4205,920 @@ lemma refillUnblockCheck_corres':
...
+        = ((\<lambda>sc'. P (scRefillCount sc')) |< scs_of' s') scp"
+  apply (drule (2) state_relation_sc_relation)
+  apply (clarsimp simp: obj_at_simps is_sc_obj valid_refills'_def sc_relation_def opt_map_red)
+  done
+
+lemma update_refill_hd_rewrite:
+  "update_refill_hd scPtr (f \<circ> g)
+   = do update_refill_hd scPtr g;
+        update_refill_hd scPtr f
+     od"
```
corlewis, line 4389: Is this a duplicate? A lemma by the same name seems to be used earlier in the file, so I think it's either an accidental duplication or a name clash.

corlewis (approved):
Another massive effort! As I'm sure you expect, I particularly like the generalisations away from `cur_sc`.

proof/refine/ARM/Schedule_R.thy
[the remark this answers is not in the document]
corlewis: For completeness, I think I agree with both of you. It would be nice to have a self-contained place to put all of the `typ_at_all_props'` interpretations, especially with how crunchable they are, but I don't think there's enough benefit to actually invest the time in moving everything.

proof/refine/ARM/Schedule_R.thy
michaelmcinerney: Yep, name clash. I've renamed it to `update_refill_hd_comp`

michaelmcinerney:
The latest commit contains a new Eisbach method to help with the `wps` stuff. 

I wanted it to be able to completely solve the goal if possible, so I added the second-to-last line in case it was able to get down to the final conjunct. I couldn't find a way to get rid of this: because we want this method to only work if it actually solves the goal, we do want `(solves blah)?` there, but then if we have this, we can't have a `(foo | bar)+` anywhere if either foo or bar always succeeds, since it will loop. So I don't know whether we can remove this line.

And the big line is also wrapped in its own question mark, in case the postcondition is not a conjunction and the `solves` bit can still solve it.

A bit complicated, so maybe not all that useful. Let me know if you'd like it moved to the `Eisbach_Methods.thy` file in `lib`

michaelmcinerney:
Yet another commit with yet another version.

This new one is a bit nicer. It's shorter, so hopefully easier to understand. It won't pause on the first conjunct it can't solve. Instead, it'll pull it all apart (if the postcondition is in fact a conjunction) and will attempt to solve each individually. If it can't solve one, it'll leave you with it. And if it's not a conjunction in the postcondition, it'll try to solve it with the last line

mktnk3:
> Yet another commit with yet another version.
> 
> This new one is a bit nicer. It's shorter, so hopefully easier to understand. It won't pause on the first conjunct it can't solve. Instead, it'll pull it all apart (if the postcondition is in fact a conjunction) and will attempt to solve each individually. If it can't solve one, it'll leave you with it. And if it's not a conjunction in the postcondition, it'll try to solve it with the last line

Ah this is nice. But I am not sure if this should belong to Eisbach_Methods.thy. Things there seem to me a bit different in nature... Maybe it can stay here? If the proofs change in the future and it needs a tweak, having it locally would be an advantage. Either way, it might be nice to have some descriptions of that it does and/or when/how to use it.

PR #353, opened 2021-09-07
`PageTableDuplicates` last sorry

michaelmcinerney:
This clears the last sorry in `PageTableDuplicates` by proving a lemma I've called `scBits_at_least_2`, which is a slight strengthening of `scBits_pos'`. It was a little annoying replacing the latter with the former, so I decided to keep the latter and prove it using the former. I'm not sure about the name for the new lemma but let me know if you have suggestions


lsf37 (approved):
Nice to see that sorry count going down!

mktnk3 (approved):
Nice!

corlewis (approved):

PR #355, opened 2021-09-15
“add refill_unblock_check” spec update, part I

mktnk3:
This is the first PR of “adding refill_unblock_check” for thread activation/switching.
The C side change is now part of the "mcs: Add sporadic flag to SchedControl_Configure” commit.

This PR contains the following:
- definition of `if_cond_refill_unblock_check`/`ifCondRefillUnblockCheck`, which package up a call to `refill_unblock_check` with various guards encoded as two bool option parameters
- this encoding currently has unsimplified duplication due to temporary changes added to ease verification (see the commit message for the details)
- remove `refill_unblock_check` from `sched_context_yield_to`, proof fix for that
- add `refill_unblock_check` to `switch_sched_context`, proof fix for that
- add `ifCondRefillUnblockCheck_corres` rule, to be used in the sequel PRs
- also contains missed Haskell fix for seL4PR#467, no proof update needed
- some cleanup, removing duplication and unused lemmas
- add a new file Sporadic_A.thy in ASpec, `refill_unblock_check` definition and what it depends on are moved there
- in DetSchedSchedule_AI, `refill_unblock_check` lemmas are moved to the beginning of the file

There are many commits but it is probably easier to review commit by commit. Happy to be still removing more lines than adding :)

### DIFF
--- a/proof/invariant-abstract/Deterministic_AI.thy
+++ b/proof/invariant-abstract/Deterministic_AI.thy
@@ -4200,6 +4200,8 @@ lemma delete_objects_valid_list[wp]: "\<lbrace>valid_list\<rbrace> delete_object
 
 lemmas mapM_x_def_bak = mapM_x_def[symmetric]
 
+crunch valid_list[wp]: refill_unblock_check valid_list
+  (wp: hoare_drop_imps crunch_wps simp: crunch_simps is_round_robin_def)
 crunch valid_list[wp]: maybe_donate_sc valid_list (wp: maybeM_inv)
 
 locale Deterministic_AI_2 = Deterministic_AI_1 +
--- a/proof/invariant-abstract/SchedContext_AI.thy
+++ b/proof/invariant-abstract/SchedContext_AI.thy
@@ -532,12 +532,24 @@ lemma refill_budget_check_if_live_then_nonz_cap[wp]:
               simp: refill_budget_check_def update_refill_hd_def)
   done
 
-crunches refill_budget_check
+lemma refill_unblock_check_refs_of[wp]:
+  "refill_unblock_check sc_ptr \<lbrace>\<lambda>s. P (state_refs_of s)\<rbrace>"
+  apply (wpsimp simp: refill_unblock_check_defs
+                  wp: hoare_drop_imp whileLoop_wp')
+  apply (clarsimp simp: state_refs_of_def)
+  done
+
+crunches refill_budget_check, if_cond_refill_unblock_check
   for aligned[wp]: pspace_aligned
   and distinct[wp]: pspace_distinct
+  and iflive[wp]: if_live_then_nonz_cap
+  and typ_at[wp]: "\<lambda>s. P (typ_at T p s)"
   and sc_at[wp]: "sc_at sc_ptr"
+  and tcb_at[wp]: "tcb_at tptr"
+  and it[wp]: "\<lambda>s. P (idle_thread s)"
   and cte_wp_at[wp]: "cte_wp_at P c"
   and interrupt_irq_node[wp]: "\<lambda>s. P (interrupt_irq_node s)"
+  and interrupt_states[wp]: "\<lambda>s. P (interrupt_states s)"
   and caps_of_state[wp]: "\<lambda>s. P (caps_of_state s)"
   and no_cdt[wp]: "\<lambda>s. P (cdt s)"
   and no_revokable[wp]: "\<lambda>s. P (is_original_cap s)"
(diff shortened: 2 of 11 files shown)


spec/abstract/Sporadic_A.thy
```diff
@@ -0,0 +1,121 @@
...
+definition if_cond_refill_unblock_check where
+  "if_cond_refill_unblock_check sc_opt active asst \<equiv>
+    maybeM (\<lambda>scp. do
+      sc \<leftarrow> get_sched_context scp;
+      cur_sc_ptr \<leftarrow> gets cur_sc;
+      guard \<leftarrow> return (case active of
+                         None \<Rightarrow> (\<not> sc_sporadic sc)
+                       | Some True \<Rightarrow> sc_sporadic sc \<and> sc_active sc
+                       | Some False \<Rightarrow> sc_sporadic sc \<and> sc_active sc);
+      when (guard \<and> (if asst = Some False then scp \<noteq> cur_sc_ptr else True)) $
```
michaelmcinerney, line 93: Would it be better to write this `if _ then _ else True` as `_ --> _`, and the one on the next line?

michaelmcinerney (approved):
Cool

spec/abstract/Sporadic_A.thy
mktnk3: what kind of benefits do you have in mind?

mktnk3:
It's probably good if #357 goes in before this one.

proof/invariant-abstract/SchedContext_AI.thy
```diff
@@ -831,7 +773,7 @@ crunches head_insufficient_loop, handle_overrun_loop
   for valid_replies_pred[wp]: "valid_replies_pred P"
   (wp: crunch_wps ignore: update_sched_context)
 
-lemma refill_budget_check_valid_replies[wp]:
+lemma refill_budget_check_valid_replies_red[wp]:
```
corlewis, line 776: Where does the `red` in the name come from?

proof/refine/ARM/Schedule_R.thy
```diff
@@ -4104,6 +4104,34 @@ lemma refillUnblockCheck_corres':
       apply (clarsimp simp: obj_at_simps valid_refills'_def ps_clear_domE scBits_simps opt_map_red)
   done
 
+lemma ifCondRefillUnblockCheck_corres:
+  "corres dc
+     (\<lambda>s. case_option True (\<lambda>scp. sc_at scp s \<and> active_sc_valid_refills s
+                                \<and> pspace_aligned s \<and> pspace_distinct s
+      \<and> (((\<lambda>sc. case_option (sc_active sc) \<top> act) |< scs_of2 s) scp)) scp_opt)
```
corlewis, line 4111: It's going to be tough to fit it all in, but this condition needs better indentation.

corlewis (approved):

spec/abstract/Sporadic_A.thy
```diff
@@ -79,4 +80,42 @@ where
...
+        when (if asst = Some True then scp \<noteq> cur_sc_ptr else True) $ refill_unblock_check scp
```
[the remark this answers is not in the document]
lsf37: Does this mirror a specific code structure in C/Haskell? Usually I'd prefer to write `when (A /\ B)` to `when A $ when B`.

spec/abstract/Sporadic_A.thy
```diff
@@ -0,0 +1,82 @@
+(*
+ * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
```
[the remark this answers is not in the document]
mktnk3: This is probably a case of adding a UNSW copyright. It was created (by Michael as a co-author) earlier and has mostly old contents moved, but it does have some recent addition.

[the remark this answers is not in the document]
lsf37: Yes, adding a 2021 UNSW line makes sense for that.

spec/abstract/Sporadic_A.thy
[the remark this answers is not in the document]
mktnk3: Yes, this is meant to be mirroring the C code. When `asst` is `Some False`, it always shows up as a conjunct in the first guard. But when `asst` is `Some True`, the “if without else (= when)” always sits inside the first guard, and originally this was an assert rather than a test - we changed it to a test to defer proving the asserts, making this case do the same test as the `Some False` case. But the structure of the C code remains the same, so I am keeping it here and in Haskell just to save the effort of changing now and also in refinement.

There is also some repetition with the `act` parameter, too. This is again meant to be two different guard, `sporadic` and `sporadic & active` tests, but since we are deferring of the proof of “sporadic implies active” for those call sites, the `sporadic` test is currently replaced with `sporadic & active`, which makes the two the same, but the original structure is kept (meaning the duplication is still there) because in C these correspond to slightly different codes and also because it should be changed back in the future.

`refill_unblock_check` is typically called under certain conditions, and there are a few patterns for those guards. This function aims to package those conditions up in an effective way so that we can make necessary arguments without unfolding it. I have written (and might add a few more) some wp rules that should cater for what we need for the cases where the difference in the parameters matters.

So, these were my thinking for keeping the original C structure as much as possible in these definitions. But that is based on the assumption that this will become useful when we revert these changes for “ease of verification” in the “near” future. If that’s unlikely, maybe we can just embrace these changes and optimise things in that way, including the C code. But we might want to do that separately, because we’d want to be merging this PR soon so that we can continue with the `schedule_corres` work.

proof/invariant-abstract/SchedContext_AI.thy
mktnk3: ah, it should be `pred` (ie. what the invariant is about).

spec/abstract/Sporadic_A.thy
mktnk3: This is a wrapper function for `refill_unblock_check`, just adding these guards to the main function. So, when these guard conditions are irrelevant to what you are proving, you can mostly crunch things away. And, when the guards do matter for the proof, you are most likely to be wanting to split things apart. So, having if-then-else in this case might be in a sense effort saving, or at least the benefit mentioned above is not as much as what you might think. But I can certainly change it to an implication, too, maybe for the sake of consistency.

mktnk3: And also, in actual call sites, the wrapper function is always called with concrete parameters, so these `if`s simplify away and there is no problem. It is only when we want to prove something generic about it that matters, and that is what my comment above is about.

mktnk3: Actually, for this particular case, aggressive splitting won't be an issue because the simplifier will just change “if `P` then `X` else `True`” to `P --> X`. We can certainly use this result of simplification in the spec from the start, but I seem to have preference in spelling things out mainly due to the convoluted nature of the spec and also the possibility of the future code change, though it is not a very strong preference/argument.

lsf37: `the simplifier will just change “if P then X else True” to P --> X` is not correct -- the simplifier will usually pull out the implication to the top of the goal, which is not what you usually want, esp if your goal is a Hoare triple or similar.

lsf37: (even without that, I find `A --> B` a lot clearer and easier to read than `if A then B else True`)

mktnk3: Hmm. Well, I had a look at the wp rules I proved, and apparently, what actually happens is that it doesn’t rewrite anything but it does not split this `if` either. Maybe because it is inside the `when` or with the conjunction, or maybe because of what is preceding I am not sure, but, for the case here, these `if`s are not being pulled out as implications. Anyway, happy to rewrite them as implications.

But I am a bit puzzled, too. This is in ASpec and I had the impression that we aren’t really trying to avoid if-then-else, generally. Maybe we should be doing that anyway.

proof/invariant-abstract/Deterministic_AI.thy
```diff
@@ -4200,6 +4200,8 @@ lemma delete_objects_valid_list[wp]: "\<lbrace>valid_list\<rbrace> delete_object
 
 lemmas mapM_x_def_bak = mapM_x_def[symmetric]
 
+crunch valid_list[wp]: refill_unblock_check valid_list
+  (wp: hoare_drop_imps crunch_wps simp: crunch_simps is_round_robin_def)
 crunch valid_list[wp]: maybe_donate_sc valid_list (wp: maybeM_inv)
```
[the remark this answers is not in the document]
mktnk3: This crunch is removed in my next PR, so I will leave this as is for now. Also, this file is full of `crunch`es, so maybe we can do a proper cleaning up separately.

PR #357, opened 2021-09-15
Remove "is_active_sc" variants from preconditions / remove haskell assert

mktnk3:
Working on Refine fix for "add refill_unblock_check everywhere" change, I started to find the "is_active_sc2" precondition of the refill_unblock_check corres rule rather difficult to deal with. So, I had a look, and it turned out that we don't need it. The haskell assert is also removed.

This also contains a small cleanup in Schedule_R that removes duplications about `tcbReleaseDequeue`.

### DIFF
--- a/proof/refine/ARM/Schedule_R.thy
+++ b/proof/refine/ARM/Schedule_R.thy
@@ -2932,22 +2932,7 @@ lemma setReleaseQueue_pred_tcb_at'[wp]:
   by (wpsimp simp: setReleaseQueue_def)
 
 crunches tcbReleaseDequeue
-  for valid_pspace'[wp]: valid_pspace'
-  and state_refs_of'[wp]: "\<lambda>s. P (state_refs_of' s)"
-  and list_refs_of_replies'[wp]: "\<lambda>s. P (list_refs_of_replies' s)"
-  and valid_global_refs'[wp]: valid_global_refs'
-  and valid_arch_state'[wp]: valid_arch_state'
-  and irq_node'[wp]: "\<lambda>s. P (irq_node' s)"
-  and typ_at'[wp]: "\<lambda>s. P (typ_at' T p s)"
-  and sc_at'_n[wp]: "\<lambda>s. Q (sc_at'_n n p s)"
-  and valid_irq_states'[wp]: valid_irq_states'
-  and ksInterruptState[wp]: "\<lambda>s. P (ksInterruptState s)"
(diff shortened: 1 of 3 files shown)


mktnk3:
We could do this (or maybe should) but I realise that the change here isn't enough to resolve my issue and I need a different approach. Anyway, I will leave this for review for now.

michaelmcinerney (approved):

proof/refine/ARM/Schedule_R.thy
```diff
@@ -2385,20 +2385,23 @@ global_interpretation refillPopHead: typ_at_all_props' "refillPopHead scPtr"
...
+  apply (clarsimp simp: valid_sched_context'_def obj_at_simps valid_obj'_def
+                        valid_sched_context_size'_def opt_map_red
                  dest!: sc_ko_at_valid_objs_valid_sc')
   done
 
-crunches mergeRefills, nonOverlappingMergeRefills
+crunches mergeRefills, nonOverlappingMergeRefills, scheduleUsed
   for active_sc_at'[wp]: "active_sc_at' scPtr"
+  and sc_at'[wp]: "\<lambda>s. P (sc_at' scPtr s)"
+  (simp: objBits_simps)
```
corlewis, line 2404: Tiny nitpick: this one and a later `sc_at'` crunch should probably be in terms of `typ_at'` and `typ_at_all_props'`

corlewis (approved):

proof/refine/ARM/Schedule_R.thy
mktnk3: Doesn’t `crunch` generate more lemmas about `sc_at’` in this case? The proofs below use not just the lemmas for the top level functions (i.e., mergeRefills, nonOverlappingMergeRefills, scheduleUsed) but also those for “sub definitions”, so it is a bit of pain to list all locale interpretations needed. They do generate all other good stuff, though. Also, it would be nice if we can do 2-dimensional locale interpretation like `crunches`.

mktnk3: I added a bunch of interpretations and tweaked the crunches. Crunches are better done on `refillUnblockCheck` and `refillBudgetCheck` and put together before the interpretations.

corlewis: Looks good to me! https://sel4.atlassian.net/browse/VER-1371 would definitely be nice to have in this case with all of those interpretations.

PR #358, opened 2021-09-20
`valid_sched_init`

michaelmcinerney:


corlewis (approved):

mktnk3 (approved):
Nice. It would be nice if we have a few more sorries like this. Let’s see :)

PR #360, opened 2021-09-22
testboard: alwaus use https URLs

lsf37:
The AWS test runner does not have a GitHub ssh key, so can only access repositories via https.


mbrcknl (approved):
Looks good!

PR #361, opened 2021-09-22
Remove `ct_not_inQ` from `invs'`

michaelmcinerney:


proof/refine/ARM/Finalise_R.thy
```diff
@@ -1389,7 +1390,21 @@ lemma deletedIRQHandler_ct_not_inQ[wp]:
   apply (simp add: comp_def)
   done
 
-crunch ct_not_inQ[wp]: emptySlot "ct_not_inQ"
```
michaelmcinerney, line 1392: I have no idea why this `crunch` stopped working... Any ideas on why would be interesting. I'm not sure this is used anywhere, but it might still be nice to have, considering that `ct_not_inQ` does still appear in a few places.

proof/refine/ARM/Untyped_R.thy
```diff
@@ -5225,8 +5225,27 @@ crunch valid_machine_state'[wp]: insertNewCap valid_machine_state'
 crunch pspace_domain_valid[wp]: insertNewCap pspace_domain_valid
   (wp: crunch_wps)
 
-crunch ct_not_inQ[wp]: insertNewCap "ct_not_inQ"
```
michaelmcinerney, line 5228: Ditto for this `crunch`

proof/refine/ARM/Tcb_R.thy
```diff
@@ -2463,7 +2444,6 @@ lemma schedContextBindTCB_invs':
              threadSet_valid_replies'
           | clarsimp simp: tcb_cte_cases_def cteCaps_of_def)+
   apply (clarsimp simp: invs'_def valid_state'_def valid_pspace'_def valid_dom_schedule'_def)
-  (* slow: around 70s *)
```
corlewis, line 2466: Very nice that this has been sped up!

proof/refine/ARM/Finalise_R.thy
corlewis: What does it get stuck on? My only guess is that some other lemma that this crunch depended on now either no longer exists or has stopped being generated by a different crunch.

corlewis (approved):
This is some great clean-up!

lsf37 (approved):

proof/refine/ARM/KHeap_R.thy
```diff
@@ -3743,6 +3743,28 @@ lemma state_refs_of_cross:
...
+  apply (prop_tac "obj_at' tcbQueued (ksCurThread s') s'")
+   apply (clarsimp simp: obj_at_simps cur_tcb'_def)
+  apply (frule curthread_relation)
+  apply (frule state_relation_ready_queues_relation)
+  apply (clarsimp simp: valid_queues'_def inQ_def ready_queues_relation_def obj_at_simps)
+  by (metis Structures_H.kernel_object.case(6))
+
+method add_ct_not_inQ =
+  rule_tac Q="\<lambda>s'. ct_not_inQ s'" in corres_cross_add_guard
+  , fastforce intro: ct_not_inQ_cross simp: valid_sched_def
```
mktnk3, line 3766: Seeing that the one for the next PR of this line has another method defined that is almost the same, maybe we can have a “meta” method, something like
```
method add_sch_act_wf P R=
   rule_tac Q=R in corres_cross_add_guard, fastforce intro: R simp: valid_sched_def
   ```
then we can define each instance using this. I am not sure the difference among the instances are, but I’d imagine it is possible to write something that cover most of the cases, if not all.

Also, there probably will be situations where we want to be adding more than one conjunct at a time. Then it will be possible to use this abstracted (parameterised) method to handle multiple additions, instead of applying multiple method sequentially.

michaelmcinerney: I think these methods are simple enough that we're not gaining much from doing this.

There is a bit of variation in these methods, though I'm not entirely sure it's necessary that they do vary. One uses `dest` rather than `intro`, another uses `intro!` instead. It seems plausible that we would like to add further things to `simp:` later. Having one method that generates all these means that it's possible that if that method were changed, we could get breakages in many places. The asserts are used all over the place now, so it could be a lot of work to fix that.

If you would like me to try this, then it might be best to merge all these PRs first and then try, so that I can see whether one general method will work

mktnk3: Since these crossing PRs follow basically the same pattern, I am just wondering if we could be more systematic with this process. Then things will be easier when we want to repeat this process for a new invariant etc. in the future (or even now).

For example, currently these `add_xxx` methods are defined in various locations (in KHeap_R mostly I think). But thinking ahead, it probably makes sense to collect them in one place. And maybe abstract things away a bit more if possible as I mentioned in the first comment, by having a template method.

I am not sure if breakage etc. mentioned is a real issue here. We are talking about having a template for applying a one particular rule (`corres_cross_add_guard`). The potential variations come from how we write the actual cross lemmas (the parameters). In that sense, I believe it is worthwhile to consider if we can write these cross lemmas in a more uniform way. But as for the template itself, if we have to accomodate something new, we can most likely work out a conservative change or if it has to be the new thing can be dealt with separately, most likely.

And, when we call these methods, the way currently things are done means we might have 3-4 (or potentially more in the future) lines of `add_xxx`, and all of them use the same lemma `corres_cross_add_guard`. So, having a template method that is It would make sense to have a way to say “cross A, B, and C” or “cross all that can be crossed”.

mktnk3:
And, I would also like to see all the commas in the “right” place, too :)
It would be nice to see this clarified in the style guide.

PR #363, opened 2021-09-24
Remove `sch_act_wf (ksSchedulerAction s) s` and `ct_idle_or_in_cur_domain'` from `invs'`

michaelmcinerney:
I think it was just luck that many of the preconditions that I have weakened could be weakened - that is, it wasn't because of the bits of `invs'` I removed. While I did try to go through all `invs'` lemmas, I'm sure there are some other lemmas elsewhere that could be weakened because of these bits being removed, so it would be good if everyone could keep an eye out for that.

It seems like `ct_idle_or_in_cur_domain'` wasn't actually used in the proofs. I haven't added an assert for it, or made a little `add` method for it, but the cross lemma is there in case it becomes necessary later.


spec/haskell/src/SEL4/Model/StateData.lhs
```diff
@@ -430,3 +430,9 @@ does hold. We may prove that this assertion is true using ct_not_inQ_cross.
 
 > ct_not_inQ_asrt :: KernelState -> Bool
 > ct_not_inQ_asrt _ = True
+
+An assert that will say that `sch_act_wf (ksSchedulerAction s) s` holds. We may prove this assertion
+is true using the ruse sch_act_wf_cross.
```
michaelmcinerney, line 435: Sigh, there is a spelling error here

proof/refine/ARM/Schedule_R.thy
```diff
@@ -837,35 +831,20 @@ lemma sch_act_wf:
...
-                              state_refs_of'_def ps_clear_def valid_irq_node'_def valid_queues'_def
-                              ct_not_inQ_ct  ct_idle_or_in_cur_domain'_def
-                              bitmapQ_defs valid_queues_no_bitmap_def valid_dom_schedule'_def
-                        cong: option.case_cong)
-    done
-qed
-
-(* Don't use this rule when considering the idle thread. The invariant ct_idle_or_in_cur_domain'
-   says that either "tcb_in_cur_domain' t" or "t = ksIdleThread s".
-   Use setCurThread_invs_idle_thread instead. *)
```
corlewis, line 863: It's so good to be able to get rid of all of this

corlewis (approved):
Very nice!

lsf37 (approved):
Really nice to see so many lines going.

There is one of those method definitions in here that starts a line with a comma, but you know my opinion on those ;-)

PR #366, opened 2021-09-27
Remove `valid_idle'` from `invs'`

michaelmcinerney:
This unfortunately adds lines, but hopefully it will lead to simpler proofs for future `invs'` lemmas


lsf37 (approved):
I think this is good and overall still a win for maintenance.

PR #367, opened 2021-09-28
testboardpush: fix URL processing

lsf37:
Had missed the `:` suffix in the git URL, so the ssh URL test would never succeed.

This is a fix for my prevision commit ce302e0c6586dedfc09d740e3b2c58ee37c77099 which didn't quite achieve what it promised.


lsf37:
Proof tests are failing because we have too many concurrent proof runs at the moment. This commit shouldn't affect any proofs anyway, though.

PR #369, opened 2021-09-28
Remove`cur_tcb'` from `invs'`

michaelmcinerney:
This removes `cur_tcb'` from `invs'`, and also the requirement that the `ksReadyQueue`s are distinct from `valid_queues`.

I chose to remove the `valid_state'` definition now that `cur_tcb'` is no longer part of `invs'`. Much of the diff is just due to this


proof/refine/ARM/KHeap_R.thy
```diff
@@ -3592,6 +3592,10 @@ lemma cur_tcb_cross:
   apply (erule (3) tcb_at_cross)
   done
 
+method add_cur_tcb' =
+  rule_tac Q="\<lambda>s'. cur_tcb' s'" in corres_cross_add_guard
+  , fastforce intro!: cur_tcb_cross
```
lsf37, line 3597: I know that people have started with `;` at the beginning of lines
, which I also object to
, but`,` is really not an operator that should start a line
. The whole point of these operators being chosen as punctuation is that they should go with the preceding statement
, not the next statement
. If you imagine the same in English,
that would look entirely weird
.

lsf37 (approved):
This is great! I wish we had done this ages ago on master already.

I would appreciate the `,` at the end of the line for the method definition :-)

proof/refine/ARM/KHeap_R.thy
mktnk3: I totally agree with Gerwin here. Punctuations are...punctuations.

mktnk3:
> This is great! I wish we had done this ages ago on master already.

I suppose we will be doing this on RISCV in the near future, and I think it probably makes sense to do it as master than as MCS. So, we can think about what the best approach would be.

PR #373, opened 2021-10-05
Make `schact_is_rct` an abbreviation

michaelmcinerney:
Because `valid_sched` is also an abbreviation, and it occurs quite often with `schact_is_rct`, `simp` and friends often rewrite `valid_sched` to `valid_sched_2` with its many many fields, and with `scheduler_action s` replaced by `resume_cur_thread`. Any proof breakages due to this can be fairly straightforwardly solved by unfolding things like `valid_sched_def`, but that seems to slow things down a lot in some places. And there's one nasty instance in `Untyped_R.thy` where I ended up writing out the entire `valid_sched_2` expression in an Isar proof. If there are better ways of handling these things, please let me know.

### DIFF
--- a/proof/invariant-abstract/DetSchedInvs_AI.thy
+++ b/proof/invariant-abstract/DetSchedInvs_AI.thy
@@ -3091,20 +3091,11 @@ abbreviation simple_sched_action :: "'z state \<Rightarrow> bool" where
 
 lemmas simple_sched_action_def = simple_sched_action_2_def
 
-definition schact_is_rct_2 :: "scheduler_action \<Rightarrow> bool" where
-  "schact_is_rct_2 schact \<equiv> schact = resume_cur_thread"
-
 abbreviation schact_is_rct :: "'z state \<Rightarrow> bool" where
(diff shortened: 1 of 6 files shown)


mktnk3:
Hmm. Sounds like this wasn’t such a great idea. Maybe there are ways to tweak things. Let me have a look later when I can.

lsf37:
What's the status of this PR? Can we close it?

michaelmcinerney:
> What's the status of this PR? Can we close it?

Miki asked me to look into this a while ago. I don't think we should proceed with it, but she said she wanted to take a look at it herself

lsf37:
Ah, yes, I should read more closely :-) @mktnk3 do you still want to pursue this one now that Refine is finished?

mktnk3:
> Ah, yes, I should read more closely :-) @mktnk3 do you still want to pursue this one now that Refine is finished?

No. Controlling the interaction with the valid_shed etc. would be a non-trivial work and it’s probably not worth the effort.

lsf37:
Cool, thanks for confirming, then I'll close this one here.

PR #375, opened 2021-10-08
`ceil_log`

mktnk3:
This PR contains changes for replacing Complex_Main `log` used in `scBitsFromRefillLength’` with `ceil_log`.

`ceil_log` is a function that returns `n` for a given `x` such that `2^(n-1) < x <= 2^n`.

`ceil_log` is defined at the end of `lib/Lib.thy` using a copied version of `Discrete.log` (renamed to `discrete_log`)
to avoid importing Complex_Main.

### DIFF
--- a/lib/Lib.thy
+++ b/lib/Lib.thy
@@ -2769,4 +2769,175 @@ lemma findIndex_member:
   apply (clarsimp simp: findIndex_def findIndex'_app)
   done
 
+
+(* ceil\_log : log 2 ceiling function for nat, which is used for SC size encoding *)
+
+(* ceil\_log is meant to be defined in terms of Discrete.log. In the below, however, we are using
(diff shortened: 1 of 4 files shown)


lsf37:
Awesome, thanks, this is good. Instead of merging it here, I'll pull this into the merge branch directly, I think, to avoid any conflicts. I might tweak slightly (the `\_`, and potentially I'll put the Lib part into a separate theory file after all, since we had to copy a bit of Discrete.Log).

We should at some point later investigate why `Complex_Main` breaks `rec_del`, but for now this will do fine.

mktnk3:
Ok, I will leave it with you then. Thanks. I can do a JIRA issue for `rec_del` to remind us.

michaelmcinerney (approved):

PR #379, opened 2021-10-15
merge master into rt

lsf37:
This merges the `master` branch into `rt`, and spans two Isabelle releases from there (2020 and 2021).

This branch is meant to replace `rt`, not to be merged into `rt`, because it contains commits from `rt` which were cherry-picked and adjusted since the time the merge started. I'm only putting this up as a PR so people can have a look if they like.

It is likely too large to review in any usable form.

Testbard is available here: <https://github.com/seL4/gh-testboard/commit/82307e2>


lsf37:
The conflicts are expected and from the shared/adjusted commits from `rt` since `rt-merge` started.

mktnk3:
> Testboard has succeeded for ARM and RISCV64. If you're happy with it, we can merge.

I am happy with that. I had a look around but not sure how much further to look, so we might as well get moving, if others are ok with it. We will be figuring out the details as we update our current work, too.

michaelmcinerney:
I'm happy to go ahead with the merge

PR #380, opened 2021-10-15
RISC-V InfoFlow

ryybrr:
Add InfoFlow proofs for RISCV. The access control proofs have also been updated to use a stronger integrity condition.

### DIFF
--- a/lib/EquivValid.thy
+++ b/lib/EquivValid.thy
@@ -195,6 +195,11 @@ lemma bind_ev:
   shows "equiv_valid I A A (\<lambda>s. P' s \<and> P'' s) (f >>= g)"
   using assms by (blast intro: bind_ev_general)
 
+lemma equiv_valid_weaken_pre:
+  "\<lbrakk> equiv_valid I A' B P f; \<forall>st t. I st t \<and> A st t \<longrightarrow> A' st t \<rbrakk>
+     \<Longrightarrow> equiv_valid I A B P f"
+  by (fastforce simp: equiv_valid_def spec_equiv_valid_def equiv_valid_2_def)
+
 lemma equiv_valid_guard_imp:
   assumes reads_res: "equiv_valid I A B Q f"
   assumes guard_imp: "\<And> s. P s \<Longrightarrow> Q s"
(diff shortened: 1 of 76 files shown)


spec/abstract/RISCV64/Arch_Structs_A.thy
```diff
@@ -222,6 +222,11 @@ definition max_pt_level :: vm_level
   where
   "max_pt_level = asid_pool_level - 1"
 
+lemma vm_level_exhaust:
+  fixes n :: vm_level
+  shows "\<lbrakk> n = 0 \<Longrightarrow> Q; n = 1 \<Longrightarrow> Q; n = 2 \<Longrightarrow> Q; n = 3 \<Longrightarrow> Q \<rbrakk> \<Longrightarrow> Q"
+  by (induct n; clarsimp)
```
lsf37, line 228: This lemma is problematic, because the value of `vm_levels` might change in the future (both up and down) based on configuration choices. The reason we don't have one like it is because we explicitly want proofs to reason about levels generically (apart from the existence of the first two levels and the existence of a bottom level).

It makes sense to use in concrete example states, but otherwise we should not use it and not make it available here.

lsf37 (commented):
Only the `vm_level_exhaust` lemma to fix up (best to move into ExampleState), otherwise I think this is ready to merge.

spec/abstract/RISCV64/Arch_Structs_A.thy
ryybrr: Any idea what lemmas I should be using instead? I run into annoying subgoals that amount to:
`⟦level < 2; level ≠ 0; level ≠ 1 ⟧ ⟹ False"`

lsf37: That lemma above would be fine, we are guaranteed to have at least 2 levels. What we don't want is a lemma that is equivalent to `vm_levels` having a specific value.

lsf37: (Actually that lemma would even be true for `vm_levels < 2`, so it's definitely fine).

lsf37 (approved):

lsf37:
Cool, good to merge once tests have passed!

PR #383, opened 2021-10-25
`refillNew_invs'` and `refillUpdate_invs'`

michaelmcinerney:
Please see the short description in the commit message

### DIFF
--- a/proof/refine/ARM/Invariants_H.thy
+++ b/proof/refine/ARM/Invariants_H.thy
@@ -670,7 +670,8 @@ definition valid_sched_context' :: "sched_context \<Rightarrow> kernel_state \<R
      \<and> scRefillMax sc \<le> length (scRefills sc)
      \<and> (0 < scRefillMax sc \<longrightarrow> scRefillHead sc < scRefillMax sc
                                \<and> scRefillCount sc \<le> scRefillMax sc
-                               \<and> 0 < scRefillCount sc)"
+                               \<and> 0 < scRefillCount sc)
+     \<and> length (scRefills sc) = refillAbsoluteMax' (scBitsFromRefillLength' (length (scRefills sc)))"
 
 definition valid_reply' :: "reply \<Rightarrow> kernel_state \<Rightarrow> bool" where
   "valid_reply' reply s \<equiv>
--- a/proof/refine/ARM/Retype_R.thy
+++ b/proof/refine/ARM/Retype_R.thy
@@ -153,7 +153,7 @@ lemma valid_obj_makeObject_sched_context [simp]:
   unfolding valid_obj'_def valid_sched_context'_def
   using refillAbsoluteMax'_lb sc_size_bounds_def minRefillLength_ARM
   by (clarsimp simp: valid_sc_size'_makeObject_sc')
-     (clarsimp simp: makeObject_sc; fastforce)
+     (clarsimp simp: makeObject_sc scBits_inverse_us)
 
 lemma valid_obj_makeObject_user_data [simp]:
   "valid_obj' (KOUserData) s"
--- a/proof/refine/ARM/SchedContextInv_R.thy
+++ b/proof/refine/ARM/SchedContextInv_R.thy
@@ -1085,6 +1085,141 @@ lemma refillUpdate_corres:
    apply clarsimp+
   done
 
+crunches maybeAddEmptyTail, setRefillHd
+  for invs'[wp]: invs'
+  (simp: crunch_simps wp: crunch_wps)
+
+lemma refillNew_invs':
+  "\<lbrace>\<lambda>s. invs' s \<and> (\<exists>n. sc_at'_n n scPtr s \<and> valid_refills_number' maxRefills n)
+        \<and> ex_nonz_cap_to' scPtr s \<and> MIN_REFILLS \<le> maxRefills\<rbrace>
+   refillNew scPtr maxRefills budget period
+   \<lbrace>\<lambda>_. invs'\<rbrace>"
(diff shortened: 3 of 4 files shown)


corlewis (approved):
I can't particularly comment on the addition to `valid_sched_context'`, but the proofs look good!

proof/refine/ARM/Invariants_H.thy
```diff
@@ -670,7 +670,8 @@ definition valid_sched_context' :: "sched_context \<Rightarrow> kernel_state \<R
      \<and> scRefillMax sc \<le> length (scRefills sc)
      \<and> (0 < scRefillMax sc \<longrightarrow> scRefillHead sc < scRefillMax sc
                                \<and> scRefillCount sc \<le> scRefillMax sc
-                               \<and> 0 < scRefillCount sc)"
+                               \<and> 0 < scRefillCount sc)
+     \<and> length (scRefills sc) = refillAbsoluteMax' (scBitsFromRefillLength' (length (scRefills sc)))"
```
mktnk3, line 674: It is true that this should hold, but that is because `length (scRefills sc)` is the maximum number of refills for a given SC size, i.e., `length (scRefills sc) = refillAbsoluteMax’ n` for `n` which is carried in `sc_at’_n`.

I think we do need to have the relation between the SC size `n` and the refill length in Refine, and it looks like adding this information to `sc_at’_n` makes sense.

michaelmcinerney: OK. Would you like to make this change to `sc_at'_n`, or shall I try? It might be best to wait for the `pspace_bounded'` work to be merged

michaelmcinerney: I did have a quick look at changing `sc_at'_n` to include a statement to this effect. There's already a breakage in `typ_at_lift_valid_untyped'`. While I'm sure this could be repaired, it seems like it's already going to be more work than the approach that's taken here.

(The extra conjunct in `valid_sched_context'` can be made slightly nicer by writing `length (scRefills sc) = refillAbsoluteMax' (objBits sc)`, in case that helps.)

Again, maybe it's not going to be bad, but there are an awful lot of `crunches` that crunch any property of `sc_at'_n n`, so there might be a difficulty there.

In current `rt,` `sc_at'_n` and `valid_sched_context_size'` are purely about the size of the scheduling context, without reference to the fields of the sc, and `valid_sched_context'` doesn't make any mention of the size of the sc at all. Seeing that we do want a statement about the length of the refills and the overall size, we'll have to break this nice "purity" somehow. 

It is easy to imagine a scheduling context whose refill buffer has length that's not as great as possible, and in that case, we'd say that it's not a valid sc. So I think it does make sense to have it in `valid_sched_context'`. But ultimately, the size of the sc comes from the refill length anyway, so maybe the distinction isn't as pure as I made it out to be, and there's no particularly strong case for either way.

Seeing that I am much less informed about the sc size stuff than you, @mktnk3, perhaps it would be best for you to add something like this to `sc_at'_n`, if you would like. I can still try, but I might not do a very good job of it

mktnk3: Ah, thanks for having a go. Yes, I can look at this later today.

corlewis: I don't have that much input here, due to it both being a while since I worked directly on these proofs and also never doing much with refills, but my first reaction was that we wouldn't want to add more to `sc_at’_n`. Haven't we already had a lot of complications due to needing to extend `sc_at'` with information about the size?

@mktnk3 Where do you see potential benefits from having this condition in `sc_at’_n` instead of `valid_sched_context_size'`. Does the suggestion to phrase it as `length (scRefills sc) = refillAbsoluteMax' (objBits sc)` help at all?

mktnk3: `length (scRefills sc) = refillAbsoluteMax' (objBits sc)` is the same things as what is currently used because `objBits sc` is equal to `scBitsFromRefillLength' (length (scRefills sc))` by definition.

The information that is currently missing in Refine is that this `length (scRefills sc)` encodes `n`. So, `n` is what exists first, and `length (scRefills sc)` is derived from it. I am saying that we should have this relation explicit somewhere.

Oddly enough, we do have this encoding stated in `sc_relation`, where `length (scRefills sc)` is linked to the `n` in the abstract. So, what we have been using this in the corres settings. But what is in this PR is only about Refine, and we have a problem because we don’t  have anything that links `n` in Refine with the refill length. This could come an issue down the track because `scRefills` is a Refine-only concept introduced to encode `n`.

So, I do feel the need to have a link between `n` and `length (scRefills sc)` just for Refine. And it seems to me that `sc_at’_n` is the right (and probably only) place we could do it. Neither of `valid_sched_context’` or `valid_sched_context_size’` mentions `n`.

mktnk3: > It is easy to imagine a scheduling context whose refill buffer has length that's not as great as possible, and in that case, we'd say that it's not a valid sc. So I think it does make sense to have it in `valid_sched_context'`. But ultimately, the size of the sc comes from the refill length anyway, so maybe the distinction isn't as pure as I made it out to be, and there's no particularly strong case for either way.

So, what I am saying is that this is actually the opposite. We start with the size of the SC `n`. And the `length (scRefills sc)` is defined to encode `n`, as the maximum number of refills for the given `n`. And that is the only way we can have an SC object.

mktnk3: But maybe this is somewhat a separate issue that can be dealt with it later, and it is best to proceed with what we have in this PR for now.

mktnk3 (approved):

PR #384, opened 2021-10-27
Weaken `akernel_invs` precondition

michaelmcinerney:
This weakens the precondition of `akernel_invs` as described in VER-1301. This also renames `is_schedulable_bool` to `schedulable`, and introduces the abbreviation `ct_schedulable`.

I will continue on with the remainder of VER-1301, namely proving a Hoare triple for `call_kernel` with postcondition `ct_running s --> ct_schedulable s`

### DIFF
--- a/proof/invariant-abstract/AInvs.thy
+++ b/proof/invariant-abstract/AInvs.thy
@@ -22,16 +22,22 @@ lemma pred_tcb_at_upd_apply:
   pred_tcb_at proj P t (s\<lparr>kheap := (kheap s)(t := p' t)\<rparr>)"
   by (simp add: pred_tcb_at_def obj_at_def)
 
-lemma thread_set_tcb_arch_is_schedulable_bool[wp]:
-  "\<lbrace>\<lambda>s. is_schedulable_bool (cur_thread s) s\<rbrace>
-     thread_set (\<lambda>tcb. tcb\<lparr>tcb_arch := arch_tcb_context_set us (tcb_arch tcb)\<rparr>) t
-   \<lbrace>\<lambda>rv s. is_schedulable_bool (cur_thread s) s\<rbrace>"
+lemma thread_set_tcb_arch_ct_schedulable[wp]:
+  "thread_set (\<lambda>tcb. tcb\<lparr>tcb_arch := arch_tcb_context_set us (tcb_arch tcb)\<rparr>) t \<lbrace>ct_schedulable\<rbrace>"
   apply (simp add: thread_set_def)
   apply (rule hoare_seq_ext[OF _ assert_get_tcb_ko'])
   apply (wpsimp wp: set_object_wp)
-  apply (fastforce simp: is_schedulable_bool_def is_sc_active_def get_tcb_def ko_atD
+  apply (fastforce simp: schedulable_def is_sc_active_def get_tcb_def ko_atD
                          in_release_queue_def
                   split: option.splits )
+  done
+
+lemma thread_set_tcb_arch_ct_not_running[wp]:
+  "thread_set (\<lambda>tcb. tcb\<lparr>tcb_arch := arch_tcb_context_set us (tcb_arch tcb)\<rparr>) t \<lbrace>\<lambda>s. \<not> ct_running s\<rbrace>"
+  apply (simp add: thread_set_def)
+  apply (rule hoare_seq_ext[OF _ assert_get_tcb_ko'])
+  apply (wpsimp wp: set_object_wp)
+  apply (clarsimp simp: ct_in_state_def pred_tcb_at_def obj_at_def split: if_splits)
   done
 
 text \<open>The top-level invariance\<close>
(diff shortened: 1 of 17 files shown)


proof/invariant-abstract/KHeap_AI.thy
```diff
@@ -2268,13 +2268,8 @@ lemma is_schedulable_sp:
...
 lemma is_schedulable_sp':
-  "\<lbrace>P\<rbrace> is_schedulable tp \<lbrace>\<lambda>rv. (\<lambda>s. rv = is_schedulable_bool tp s) and P\<rbrace>"
-  apply (clarsimp simp: is_schedulable_def)
-  apply (wpsimp simp: hoare_vcg_if_lift2 obj_at_def is_tcb wp: get_sched_context_wp)
-  apply(rule conjI)
-   apply (clarsimp simp: Option.is_none_def is_schedulable_bool_def get_tcb_def)
-  by (clarsimp simp: is_schedulable_bool_def get_tcb_def is_sc_active_def
-              split: option.splits)
+  "\<lbrace>P\<rbrace> is_schedulable tp \<lbrace>\<lambda>rv. (\<lambda>s. rv = schedulable tp s) and P\<rbrace>"
+  by (wpsimp wp: is_schedulable_wp)
```
corlewis, line 2272: Haha, that's some nice cleanup.

proof/invariant-abstract/AInvs.thy
```diff
@@ -22,26 +22,32 @@ lemma pred_tcb_at_upd_apply:
...
 
 lemma akernel_invs:
   "\<lbrace>\<lambda>s. invs s \<and> (e \<noteq> Interrupt \<longrightarrow> ct_running s) \<and>
         scheduler_action s = resume_cur_thread \<and>
-        is_schedulable_bool (cur_thread s) s\<rbrace>
-     (call_kernel e)
-   \<lbrace>\<lambda>rv s. (invs s \<and> (ct_running s \<or> ct_idle s))\<rbrace>"
+        (ct_running s \<longrightarrow> ct_schedulable s)\<rbrace>
+   call_kernel e
+   \<lbrace>\<lambda>_ s. (invs s \<and> (ct_running s \<or> ct_idle s))\<rbrace>"
```
corlewis, line 50: Feel free to ignore this nitpick, but the postcondition doesn't need the outer brackets

corlewis (approved):

spec/abstract/Schedule_A.thy
```diff
@@ -212,8 +212,8 @@ where
 abbreviation (input)
   schedule_switch_thread_branch :: "obj_ref \<Rightarrow> obj_ref \<Rightarrow> bool \<Rightarrow> (unit, 'z::state_ext) s_monad"
 where
-  "schedule_switch_thread_branch candidate ct ct_schedulable \<equiv> do
-     when ct_schedulable (tcb_sched_action tcb_sched_enqueue ct); \<comment> \<open>schedulable\<close>
+  "schedule_switch_thread_branch candidate ct ct_schdble \<equiv> do
```
mktnk3, line 215: It is just a matter of taste, but I kind of prefer something like `ct_sched`...

mktnk3 (approved):
Nice. I can see that most of the changes are coming from the rename.

To be honest, I am not too comfortable with `schedulable` rename. One thing to consider probably is that the name of this test function is `isSchedulable` in both Haskell and C. And `schedulable` is often used as variable names for the return value.

For the abstract, there is also `is_schedulable_opt`, which was there first, I think, and that is why we have `is_scheduable_bool`. We can maybe settle for renaming `is_schedulable_bool` to `is_schedulable` and remove `is_schedulable_opt`.

michaelmcinerney:
> Nice. I can see that most of the changes are coming from the rename.
> 
> To be honest, I am not too comfortable with `schedulable` rename. One thing to consider probably is that the name of this test function is `isSchedulable` in both Haskell and C. And `schedulable` is often used as variable names for the return value (in Haskell mostly).
> 
> For the abstract, there is also `is_schedulable_opt`, which was there first, I think, and that is why we have `is_scheduable_bool`. We can maybe settle for renaming `is_schedulable_bool` to `is_schedulable` and remove `is_schedulable_opt`.

`is_schedulable` is the name of the monadic function in the abstract spec, so we will have to have something else here. I don't think there were any clashes in Refine due to having `schedulable` and then using `schedulable` for the name of the return value (there were a couple with `ct_schedulable`, which is why I have written `ct_schdble` in a couple places). Seeing that there are so many instances of `schedulable` in AInvs, I think `schedulable` is a good way to go

mktnk3:
ah yes, of course. So, there is that correspondence, and this is for the non-monadic one. But doesn’t that mean that ASpec still has a lot of uses of `schedulable` as return value variable?

michaelmcinerney:
Yeah the ASpec uses `schedulable` as the name of the return value a fair bit, but nothing breaks because of it. Isabelle will just name this `schedulablea` if it appears as a green variable in a proof. Should I use `sched` as the name of the return value instead?

mktnk3:
I for myself wasn’t really feeling the need for introducing this renaming to start with, but I suppose you felt the need so you did this, and it’s up to you how to wrap it up...

mktnk3:
Then again, we have `isSchedulable_bool` in Refine, too...

michaelmcinerney:
Gerwin suggested the renaming in VER-1301, and the names themselves. I do think that `is_schedulable_bool` is a bit cumbersome, especially for something that occurs so frequently.

I forgot about `isSchedulable_bool`. If everyone is happy with `schedulable` and `ct_schedulable`, would `schedulable'` for Refine be reasonable?

Seeing that the `call_kernel` lemmas work is already done, I would prefer to merge this PR as it is now, and then come back and do any renaming and modifications after the `call_kernel` lemmas work is merged. (I am a bit worried that rebasing so many commits would be painful.) Is that OK?

mktnk3:
I’ll leave it up to you to judge

lsf37:
I'd be fine with `schedulable` and `schedulable'`.

PR #389, opened 2021-11-07
pspace_bounded'

mktnk3:
In master, we have
```
lemma objBitsKO_bounded[simp]:
  "objBitsKO ko < word_bits”
```
which is implicitly used in many many places where we need to argue alignments, etc.

In MCS, we cannot have this fact without an assumption on the SC size. The decode function guarantees that the user-defined size for an SC object is bounded, but we need to pass around that information or something equivalent in one way or another. So far, we had a sorry in StateRelation.thy that basically gave this assumption for free. This PR removes this sorry and sorts the situation out properly.

This PR adds an invariant `pspace_bounded’`, which states that all the kernel object `ko` in the `ksPSpace` satisfy `objBitsKO ko < word_bits`. This essentially replaces the above fact in master.

`pspace_bounded'` goes along with `pspace_aligned’` and `pspace_distinct’`, but no abstract counterpart is needed. In the corres settings, we can always have it because `pspace_relation` implies `pspace_bounded’`.

Correspondingly, `ko_wp_at’` now contains a condition on the bound along with those on alignment and distinctiveness. `loadObject_default` and `updateObject_default` also assert the bound.

One more thing: we need this bound to be able to state non-overflow for the kernel object size. In master, this is again something you can get simply by using the fixed size for the kernel, but in MCS we need an assumption:
```
lemma objBitsKO_no_overflow[simp, intro!]:
  "objBitsKO ko < word_bits
    \<Longrightarrow> (1::machine_word) < (2::machine_word)^(objBitsKO ko)"
 ```
With `pspace_bounded’`, we can derive this conclusion, and I have for now put this in the simp set, which seems to be working fine and have led to some simplifications in a few places.

### DIFF
--- a/proof/refine/ARM/Detype_R.thy
+++ b/proof/refine/ARM/Detype_R.thy
@@ -497,6 +497,7 @@ lemma  valid_objs: "valid_objs' s'"
   and      pspace: "valid_pspace' s'"
   and          pa: "pspace_aligned' s'"
   and          pd: "pspace_distinct' s'"
+  and          bd: "pspace_bounded' s'"
   and          vq: "valid_queues s'"
   and         vq': "valid_queues' s'"
   and        vrlq: "valid_release_queue s'"
@@ -841,7 +842,7 @@ lemma freeMemory_deletionIsSafe[wp]:
   done
 
 lemma detype_ReplyPrevs_of:
-  "\<lbrakk>pspace_aligned' s'; pspace_distinct' s'; \<forall>p. p \<in> S \<longrightarrow> \<not> ko_wp_at' live' p s'\<rbrakk>
+  "\<lbrakk>pspace_aligned' s'; pspace_distinct' s'; pspace_bounded' s'; \<forall>p. p \<in> S \<longrightarrow> \<not> ko_wp_at' live' p s'\<rbrakk>
    \<Longrightarrow> ((\<lambda>x. if x \<in> S then None else ksPSpace s' x) |> reply_of' |> replyPrev)
        = replyPrevs_of s'"
   apply (prop_tac "\<And>p reply_ptr. (replyPrevs_of s' p = Some reply_ptr) \<Longrightarrow> p \<notin> S")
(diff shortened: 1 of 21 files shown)

### CODE REVIEW

File: proof/refine/ARM/Invariants_H.thy [hunk 1]
```diff
@@ -50,9 +50,27 @@ definition
...
+  | (KONotification _) \<Rightarrow>    True
+  | (KOCTE _) \<Rightarrow>    True
+  | (KOTCB _) \<Rightarrow>    True
+  | (KOUserData) \<Rightarrow>    True
+  | (KOUserDataDevice) \<Rightarrow>    True
+  | (KOKernelData) \<Rightarrow>    True
+  | (KOArch _) \<Rightarrow>    True
+  | (KOSchedContext sc) \<Rightarrow> scBitsFromRefillLength sc < word_bits
+  | (KOReply _) \<Rightarrow>    True
+  )"
```
lsf37, line 68: You can use a default pattern in case, e.g.:

```suggestion
lemma valid_sz_simps:
"objBitsKO ko < word_bits =
 (case ko of
  | KOSchedContext sc \<Rightarrow> scBitsFromRefillLength sc < word_bits
  | _ \<Rightarrow>    True
  )"
```

lsf37 (approved):
Nice, I'm glad this sorry is gone. It always looked like it would be another invariant or at least a nontrivial amount of work.

proof/refine/ARM/Detype_R.thy
```diff
@@ -3212,9 +3263,16 @@ lemma copyGlobalMappings_setCTE_commute:
   apply (clarsimp simp: pteBits_def pdeBits_def)
   done
 
+lemma dmo_bounded'[wp]:
+  "\<lbrace>pspace_bounded'\<rbrace> doMachineOp f \<lbrace>\<lambda>_. pspace_bounded'\<rbrace>"
```
michaelmcinerney, line 3267: Use invariant notation

proof/refine/ARM/KHeap_R.thy
```diff
@@ -875,7 +867,8 @@ lemma get_ntfn'_valid_ntfn[wp]:
 
 lemma get_sc_valid_sc'[wp]:
   "\<lbrace> valid_objs' \<rbrace> getSchedContext sc \<lbrace> valid_sched_context' \<rbrace>"
-  unfolding getSchedContext_def by getObject_valid_obj
+  unfolding getSchedContext_def
+ by getObject_valid_obj
```
michaelmcinerney, line 871: Another space needed

michaelmcinerney (approved):
Nice. Just a couple boring nitpicks from me

corlewis (approved):
Great job! As Gerwin said, this definitely looked like it would take a fair bit of work to resolve.

proof/refine/ARM/PageTableDuplicates.thy
```diff
@@ -1220,17 +1225,15 @@ lemma createNewObjects_valid_duplicates'[wp]:
       apply (simp add: word_bits_def)
      apply wp
       apply (wp snoc.hyps)
+      apply (subst conj_assoc[symmetric])
+      apply (subst conj_assoc[symmetric])
+      apply (subst conj_assoc[symmetric])
```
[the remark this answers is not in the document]
mktnk3: Sure,  I can add a comment. I can’t say I have explored all possibilities to avoid this, but I believe what I have is a slight improvement to the previous version. The wp rule used here had three conjuncts and now it has four. And the original proof was stepping it through one by one three times. I didn’t want to make it four times repeats. So, instead, I am bundling up the four conjuncts together first to apply the rule in just one step.

proof/refine/ARM/Retype_R.thy
```diff
@@ -553,8 +553,8 @@ lemma cte_wp_at_retype':
                    cong: if_cong)
      apply (insert al ko)
      apply (simp, safe, simp_all)
-              apply (fastforce simp: makeObjectKO_def makeObject_cte makeObject_tcb tcb_cte_cases_def
-                              split: if_split_asm)+
+              apply ((fastforce simp: makeObjectKO_def makeObject_cte makeObject_tcb tcb_cte_cases_def
+                              split: if_split_asm)+)[10]
```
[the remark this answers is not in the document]
mktnk3: I can replace this line with
```
apply simp
 apply (safe; simp)
```
so that the second `simp` is applied only to what `safe` generates.

PR #390, opened 2021-11-08
`hw_invs'`

michaelmcinerney:


proof/refine/ARM/Syscall_R.thy
```diff
@@ -1402,41 +1402,31 @@ lemma lookupCap_refs[wp]:
...
+   apply (fastforce simp: ct_in_state'_def)
+  apply (rename_tac epCap)
+  apply (case_tac epCap; clarsimp?, (solves wpsimp)?)
+   apply (wpsimp wp: getNotification_wp)
+   apply (clarsimp simp: obj_at_simps isNotificationCap_def)
+  apply (intro conjI impI; (solves wpsimp)?)
+  by (wpsimp simp: lookupReply_def getCapReg_def
+               wp: hoare_vcg_conj_liftE
+      | wp (once) hoare_drop_imps)+
+     (clarsimp simp: obj_at_simps ct_in_state'_def pred_tcb_at'_def)
```
corlewis, line 1429: I have a very mild preference for splitting this up into the two different steps, but it's mild enough that you should leave it as is if you prefer them together.

corlewis (approved):
Very nice, especially with simplifying all of those preconditions!

proof/refine/ARM/Syscall_R.thy
michaelmcinerney: Oh I meant to ask about this one: the first step here takes about 20 to 25 seconds, which is quite unusual, I thought. The `apply_trace` spits out a bunch of things, so maybe it's all fine, but I can't remember a `wpsimp` step taking this long.

I'm not sure it's worth investigating, and I'm not sure I'd be the best person to investigate it if it is. But if you'd like to look at it together, I'd be interested

michaelmcinerney: After chatting with Corey a bit more, we've decided to leave this as is, but with a comment saying that this takes a little while

mktnk3 (approved):
Nice. Good to see another sorry go, and it looks like with some cleanup, too.

mktnk3:
The test on ARM proof failed but it looks like due to AWS going offline or something. I can see that ARM Refine was successful. I guess we can rerun the test anyway.

PR #391, opened 2021-11-09
`invokeSchedControlConfigureFlags_corres`

michaelmcinerney:
The first commit in the branch was cherry-picked from the `call_kernel` lemmas PR

### DIFF
--- a/lib/Corres_UL.thy
+++ b/lib/Corres_UL.thy
@@ -542,6 +542,11 @@ lemma corres_if_split:
    \<Longrightarrow> corres_underlying sr nf nf' r P P' (if G then a else b) (if G' then c else d)"
   by (clarsimp simp: corres_underlying_def)
 
+lemma abs_ex_lift_corres:
+  "(\<And>n. corres_underlying srel nf nf' rrel (\<lambda>s. P n s \<and> Q s) P' G G' )
+   \<Longrightarrow> corres_underlying srel nf nf' rrel (\<lambda>s. (\<exists>n. P n s) \<and> Q s) P' G G'"
+  by (fastforce simp: corres_underlying_def)
+
 text \<open>Some equivalences about liftM and other useful simps\<close>
 
 lemma snd_liftM [simp]:
--- a/proof/invariant-abstract/DetSchedSchedule_AI.thy
+++ b/proof/invariant-abstract/DetSchedSchedule_AI.thy
@@ -12835,19 +12835,27 @@ lemma sched_context_donate_wp:
   unfolding sched_context_donate_def
   by (wpsimp wp: ssc_bound_tcb_at' sc_tcb_update_sc_tcb_sc_at)
 
-lemma tcb_release_remove_sc_not_in_release_q[wp]:
-  "\<lbrace>sc_not_in_release_q sc_ptr\<rbrace>
-    tcb_release_remove tptr
-   \<lbrace>\<lambda>rv. sc_not_in_release_q sc_ptr\<rbrace>"
+lemma tcb_release_remove_sc_not_in_release_q_inv:
+  "tcb_release_remove tptr \<lbrace>sc_not_in_release_q sc_ptr\<rbrace>"
   apply (wpsimp simp: tcb_release_remove_def)
-  by (clarsimp simp: pred_tcb_at_def obj_at_def tcb_sched_dequeue_def not_in_release_q_def in_queue_2_def)
+  by (clarsimp simp: pred_tcb_at_def obj_at_def tcb_sched_dequeue_def not_in_release_q_def
+                     in_queue_2_def)
+
+lemma tcb_release_remove_sc_not_in_release_q:
+  "\<lbrace>\<lambda>s. heap_ref_eq sc_ptr tptr (tcb_scps_of s) \<and> heap_refs_inj (tcb_scps_of s)\<rbrace>
+   tcb_release_remove tptr
+   \<lbrace>\<lambda>_. sc_not_in_release_q sc_ptr\<rbrace>"
+  apply (clarsimp simp: tcb_release_remove_def)
+  apply wpsimp
+  apply (clarsimp simp: tcb_sched_dequeue_def in_queue_2_def heap_refs_inj_eq)
+  done
 
 lemma sched_context_donate_sc_not_in_release_q:
   "\<lbrace>sc_not_in_release_q sc_ptr and not_in_release_q tptr\<rbrace>
-    sched_context_donate scp tptr
-   \<lbrace>\<lambda>rv. sc_not_in_release_q sc_ptr\<rbrace>"
+   sched_context_donate scp tptr
+   \<lbrace>\<lambda>_. sc_not_in_release_q sc_ptr\<rbrace>"
   apply (clarsimp simp: sched_context_donate_def)
-  by (wpsimp wp: get_sc_obj_ref_wp hoare_vcg_imp_lift' tcb_release_remove_not_in_release_q)
+  by (wpsimp wp: get_sc_obj_ref_wp hoare_vcg_imp_lift' tcb_release_remove_sc_not_in_release_q_inv)
 
 lemma maybe_donate_sc_valid_release_q:
   "\<lbrace> valid_release_q and not_in_release_q tcb_ptr\<rbrace>
--- a/proof/invariant-abstract/DetSchedSchedule_AI.thy
+++ b/proof/invariant-abstract/DetSchedSchedule_AI.thy
@@ -11883,7 +11883,6 @@ lemma refill_budget_check_active_sc_valid_refills:
    the cur_sc is ready at this point *)
 lemma commit_time_active_sc_valid_refills:
   "\<lbrace>active_sc_valid_refills
-    and cur_sc_in_release_q_imp_zero_consumed
     and cur_sc_more_than_ready
     and current_time_bounded 5
     and consumed_time_bounded\<rbrace>
--- a/proof/invariant-abstract/DetSchedSchedule_AI.thy
+++ b/proof/invariant-abstract/DetSchedSchedule_AI.thy
@@ -269,10 +269,29 @@ lemma tcb_sched_append_ct_not_in_q[wp]:
      \<lbrace>\<lambda>_. ct_not_in_q\<rbrace>"
   by (wpsimp wp: valid_sched_wp simp: valid_sched_wpsimps)
 
-lemma tcb_sched_dequeue_ct_not_in_q[wp]:
-  "\<lbrace>ct_not_in_q\<rbrace> tcb_sched_action tcb_sched_dequeue thread
-   \<lbrace>\<lambda>_. ct_not_in_q\<rbrace>"
+lemma tcb_sched_dequeue_ct_not_in_q:
+  "tcb_sched_action tcb_sched_dequeue thread \<lbrace>ct_not_in_q\<rbrace>"
   by (wpsimp wp: valid_sched_wp simp: valid_sched_wpsimps)
+
+lemma tcb_sched_dequeue_sc_not_in_ready_q:
+  "\<lbrace>\<lambda>s. heap_ref_eq sc_ptr tptr (tcb_scps_of s) \<and> heap_refs_inj (tcb_scps_of s) \<and> valid_ready_qs s
+        \<and> tcb_at tptr s\<rbrace>
+   tcb_sched_action tcb_sched_dequeue tptr
+   \<lbrace>\<lambda>rv. sc_not_in_ready_q sc_ptr\<rbrace>"
+  apply (clarsimp simp: tcb_sched_action_def set_tcb_queue_def)
+  apply (wpsimp wp: thread_get_wp)
+  apply (clarsimp simp: pred_tcb_at_def obj_at_def tcb_sched_dequeue_def heap_refs_inj_eq
+                        in_queues_2_def  valid_ready_qs_def is_tcb_def)
+  apply (rename_tac ko)
+  apply (case_tac ko; clarsimp)
+  apply (drule_tac x=d in spec)
+  apply (drule_tac x=p in spec)
+  apply clarsimp
+  apply (drule_tac x=t in bspec, blast)
+  apply (prop_tac "t=tptr")
+   apply (clarsimp simp:heap_refs_inj_eq)
+  apply (clarsimp simp: vs_all_heap_simps)
+  done
 
 lemma valid_blocked_discharge_except:
   assumes "valid_blocked_thread id id S queues rlq sa ct tcb_sts tcb_scps sc_refill_cfgs t"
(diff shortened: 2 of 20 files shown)


proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -17782,12 +17800,13 @@ lemma invoke_sched_control_configure_flags_valid_sched:
...
                      hoare_vcg_conj_lift tcb_release_remove_cur_sc_in_release_q_imp_zero_consumed)
-   apply (clarsimp simp: valid_sched_valid_sched_except_blocked current_time_bounded_def)
+   apply (clarsimp simp: valid_sched_valid_sched_except_blocked current_time_bounded_def
+                         sc_at_pred_n_def obj_at_def)
 
   apply (rule_tac B="\<lambda>rv s. iscc_valid_sched_predicate sc_ptr tcb_ptr mrefills budget s
                             \<and> budget \<le> period \<and> period \<le> MAX_PERIOD \<and> current_time_bounded 3 s
                             \<and> ex_nonz_cap_to sc_ptr s \<and> consumed_time_bounded s
-                            \<and> sc_refill_max_sc_at (\<lambda>rm. rm = sc_refill_max sc) sc_ptr s
+                           \<and> sc_refill_max_sc_at (\<lambda>rm. rm = sc_refill_max sc) sc_ptr s
```
michaelmcinerney, line 17809: I'll revert this

proof/refine/ARM/SchedContextInv_R.thy
```diff
@@ -1220,17 +1220,409 @@ lemma refillUpdate_invs':
...
+           apply fastforce
+          apply (wpsimp wp: commit_time_active_sc_valid_refills
+                            hoare_vcg_imp_lift')
+          apply (fastforce intro: cur_sc_active_offset_ready_and_sufficient_implies_cur_sc_more_than_ready)
+         apply (wpsimp wp: commit_time_valid_ready_qs hoare_vcg_imp_lift'
+                           tcb_sched_dequeue_valid_ready_qs hoare_vcg_disj_lift)
+         apply (fastforce intro: cur_sc_active_offset_ready_and_sufficient_implies_cur_sc_more_than_ready)
+        apply (wpsimp wp: commit_time_valid_release_q hoare_vcg_imp_lift'
+                          tcb_release_remove_cur_sc_in_release_q_imp_zero_consumed'
+                          tcb_release_remove_cur_sc_in_release_q_imp_zero_consumed
```
michaelmcinerney, line 1420: This one can be removed

proof/refine/ARM/Syscall_R.thy
```diff
@@ -1119,7 +1119,7 @@ lemma handleInvocation_corres:
             apply (rule corres_split [OF setThreadState_corres], simp)
               apply (rule corres_splitEE [OF _ performInvocation_corres])
                   apply simp
-                  apply (rule corres_split [OF getThreadState_corres])
+                  apply (rule corres_split [OF getThreadState_corres]; (solves simp)?)
```
corlewis, line 1127: Would a `lemmas getThreadState_corres' = getThreadState_corres[OF refl]` be worth making to reduce the number of places you need this `solves simp` step?

corlewis (approved):
Great job getting through another sorry!

lib/Corres_UL.thy
```diff
@@ -542,6 +542,11 @@ lemma corres_if_split:
    \<Longrightarrow> corres_underlying sr nf nf' r P P' (if G then a else b) (if G' then c else d)"
   by (clarsimp simp: corres_underlying_def)
 
+lemma abs_ex_lift_corres:
+  "(\<And>n. corres_underlying srel nf nf' rrel (\<lambda>s. P n s \<and> Q s) P' G G' )
+   \<Longrightarrow> corres_underlying srel nf nf' rrel (\<lambda>s. (\<exists>n. P n s) \<and> Q s) P' G G'"
```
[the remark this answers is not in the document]
michaelmcinerney: This is used in `invokeSchedControlConfigureFlags_corres`. It's all a bit of a blur, but I think this was used to deal with `\exists n. sc_obj_at n sc s`, which can be gotten from `sc_at`, in order for `refillUpdate_corres` and `refillNew_corres` to be applicable: their assumptions are written in a form that refers to the `n` as it appears in the abstract size of the sc, and the `n` there is a blue variable. I did try to rewrite these `corres` rules to use the form that's immediate from `valid_sc_ctrl_inv' sc_inv'`, which would have meant that this rule I wrote, or something like it, wasn't needed, but that was very very difficult, and I gave up.

Is there a better approach?

proof/refine/ARM/Syscall_R.thy
michaelmcinerney: I'd be happy to do that. But if I did it to this rule, should I do it for the others too? What does everyone think?

proof/refine/ARM/SchedContext_R.thy
```diff
@@ -185,6 +185,46 @@ lemma scConsumed_update_corres:
   apply (clarsimp simp: sc_relation_def objBits_def objBitsKO_def obj_at_def)
   done
 
+lemma scBadge_update_corres:
+  "\<lbrakk>\<forall>x y. x = y \<longrightarrow> f x = f' y\<rbrakk> \<Longrightarrow>
+    corres dc (sc_at scPtr) (ko_at' sc' scPtr)
+          (update_sched_context scPtr (sc_badge_update f))
+          (setSchedContext scPtr (scBadge_update f' sc'))"
```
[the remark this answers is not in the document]
michaelmcinerney: They are equivalent. I know that the `scConsumed_update` version was used in that other proof you pointed out that is extremely manual. I can try rewriting the assumption, though, and see how it goes

lib/Corres_UL.thy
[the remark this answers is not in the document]
michaelmcinerney: @lsf37 Is there a good way to mark `corres` rules that aren't safe? We've got `wp_unsafe`, but nothing similar for `corres`, I don't think. Or are you aware of a better way of handling this situation?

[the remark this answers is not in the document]
lsf37: They aren't used automatically, so there is no marker, but we could something in the name (or a comment)

proof/refine/ARM/TcbAcc_R.thy
```diff
@@ -1886,8 +1886,9 @@ lemma asUser_setRegister_corres:
 end
 
 lemma getThreadState_corres:
-  "corres thread_state_relation (tcb_at t) (tcb_at' t)
-          (get_thread_state t) (getThreadState t)"
+  "t = t' \<Longrightarrow>
+   corres thread_state_relation (tcb_at t) (tcb_at' t)
+          (get_thread_state t) (getThreadState t')"
```
mktnk3, line 1891: I am finding it hard to appreciate the merit of this change. It seems odd to talk about different pointers when we assume `obj_at` and `obj_at’` for the same `t`. And it does make to assume those for such a basic function. So, probably it is a bit too much to apply this change for this case.  The same for the `isRunnable_corres` below.

proof/refine/ARM/Tcb_R.thy
```diff
@@ -582,8 +582,9 @@ lemma copyreg_invs':
   by (rule hoare_strengthen_post, rule copyreg_invs'', simp)
 
 lemma isRunnable_corres:
-  "corres (\<lambda>ts runn. runnable ts = runn) (tcb_at t) (tcb_at' t)
-     (get_thread_state t) (isRunnable t)"
+  "t = t' \<Longrightarrow>
+   corres (\<lambda>ts runn. runnable ts = runn) (tcb_at t) (tcb_at' t)
+     (get_thread_state t) (isRunnable t')"
```
mktnk3, line 587: See the comment for `getThreadState_corres`.

proof/refine/ARM/TcbAcc_R.thy
mktnk3: or maybe just change the `t`’s in the assumptions as well and it might work, actually. But it is generating more work to apply in many places it seems.

michaelmcinerney: These rules will always be applied when the guards are schematic, so it doesn't matter what the guards say (I could also use a cross lemma so that the concrete guard is just `Top` if that is preferred, though there might not be much benefit). 

I was really just wanting to make it easier to apply these sorts of rules (without first having to do a `simp` or a `corres_gen_asm` or something). Do we now think that this is not a good idea to have `corres` rules in this form? Or would Corey's suggestion with the `lemmas` statement be a good compromise?

mktnk3: But is it right to say that, with this change, we will “always” have to simplify after applying the rule? Doesn't that mean this gives us more work overall, compared with having to maybe preprocess sometimes? I agree Corey’s suggestion is one solution, and if we are doing that, it probably makes sense to just keep the original statement with the original name so that we don’t need to update the existing proofs, and use the name with prime (‘) for the new version to be used only when it is needed.

michaelmcinerney: Yes we will always have to simplify after applying a rule in this form. 

I am not sure whether it would be more work, but it certainly seems more convenient to me to always have the rule apply, and then resolve this obligation with a `simp` or `clarsimp simp: tcb_relation_def` or something after. We discussed this in the RT channel and there were not any objections then. From memory, I have often had to do `corres_req` or `corres_gen_asm` before applying the rule, which is much more manual and prone to breaking. @corlewis @lsf37 do you have any further thoughts on this?

To avoid having to state and prove similar lemmas twice, I think the `lemmas` statement is the best approach. So I could have a primed name for the more complicated version, and then give the `lemmas` statement the unprimed name

corlewis: My understanding is that we often need this sort of assumption to handle cases where the various bits of machinery haven't already worked out that they are the same t's. I think it's a bit simpler here than with similar cases in hoare triple, in that due to the way that we use return value relations we can have this as an assumption and not a precondition. However, my feeling is that at the moment it is as Michael describes, with us needing to have some sort of step either before or after applying the corres rule.

Assuming that we're not going to invest the time to try and find some sort of solution like making `corressimp` more powerful, my preference is something along the lines of what is here. Having the simpler `lemmas` version is a bit of a trade-off, it makes things simpler for us but I can see it being a bit confusing for new proof engineers in the future. Even considering that, I prefer having the simpler version for when it can be used.

What I'm not certain about is the question about which `t` to refer to in the preconditions. Now that I'm looking at it, it does feel a bit odd to have the wrong `t` but also it logically doesn't make a difference and generally we do prefer to keep things in terms of their abstract versions.

mktnk3: I don’t think this style of giving the equality will “always” work, though. If you need the state relation, or even some of the preconditions on either of the abs/conc sides (such as `valid_objs`), to show the equality, you will get stuck, unless you have shown the equality using the usual `corres_req`, etc., prior to applying the rule. It might be rare to have such a situation but I don’t think we can exclude the possibility, particularly with MCS. I suppose we can backtrack if we get stuck like that and show what needs to be shown, which I guess could still work overall.

proof/refine/ARM/SchedContextInv_R.thy
```diff
@@ -1220,17 +1220,409 @@ lemma refillUpdate_invs':
...
+  apply (clarsimp simp: updateSchedContext_def)
+  apply (wpsimp wp: set_sc'.set_wp)
+  apply (fastforce simp: active_sc_at'_def obj_at_simps scBits_simps ps_clear_def)
+  done
+
+lemma updateSchedContext_scSporadic_update_active_sc_at'[wp]:
+  "updateSchedContext scPtr' (scSporadic_update f) \<lbrace>active_sc_at' scPtr\<rbrace>"
+  apply (clarsimp simp: updateSchedContext_def)
+  apply (wpsimp wp: set_sc'.set_wp)
+  apply (fastforce simp: active_sc_at'_def obj_at_simps scBits_simps ps_clear_def)
```
mktnk3, line 1256: We have `updateSchedContext_active_sc_at’` which should cover these cases.

proof/refine/ARM/SchedContext_R.thy
```diff
@@ -185,6 +185,46 @@ lemma scConsumed_update_corres:
...
+                              [where f="sc_badge_update f" and f'="scBadge_update f'" ])
+  apply (clarsimp simp: sc_relation_def objBits_def objBitsKO_def obj_at_def)
+  done
+
+lemma scSporadic_update_corres:
+  "\<lbrakk>\<forall>x y. x = y \<longrightarrow> f x = f' y\<rbrakk> \<Longrightarrow>
+    corres dc (sc_at scPtr) (ko_at' sc' scPtr)
+          (update_sched_context scPtr (sc_sporadic_update f))
+          (setSchedContext scPtr (scSporadic_update f' sc'))"
+  apply (clarsimp simp: update_sched_context_def)
```
mktnk3, line 213: For this one and the one above (`scBadge_update_corres`), and also `scConsumed_update_corres `, we probably don’t need separate rules for each of these. We should be able to have a generic version to cover all these cases, for corresponding between `setSchedContext` and `update_sched_context`, when updating non-reply, non-refills fields.

mktnk3 (approved):
Nice. Great to see the last sorry in SchedContextInv_R go.

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -12667,19 +12699,27 @@ lemma sched_context_donate_wp:
...
+  "tcb_release_remove tptr \<lbrace>sc_not_in_release_q sc_ptr\<rbrace>"
   apply (wpsimp simp: tcb_release_remove_def)
-  by (clarsimp simp: pred_tcb_at_def obj_at_def tcb_sched_dequeue_def not_in_release_q_def in_queue_2_def)
+  by (clarsimp simp: pred_tcb_at_def obj_at_def tcb_sched_dequeue_def not_in_release_q_def
+                     in_queue_2_def)
+
+lemma tcb_release_remove_sc_not_in_release_q:
+  "\<lbrace>\<lambda>s. heap_ref_eq sc_ptr tptr (tcb_scps_of s) \<and> heap_refs_inj (tcb_scps_of s)\<rbrace>
+   tcb_release_remove tptr
+   \<lbrace>\<lambda>_. sc_not_in_release_q sc_ptr\<rbrace>"
```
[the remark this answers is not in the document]
michaelmcinerney: I've combined these two cases into a lemma called `tcb_release_remove_sc_not_in_release_q`

proof/refine/ARM/SchedContextInv_R.thy
michaelmcinerney: Removed

PR #394, opened 2021-11-10
`performInv_invs`

michaelmcinerney:
Signed-off-by: Michael McInerney <m.mcinerney@unsw.edu.au>

### DIFF
--- a/proof/refine/ARM/Syscall_R.thy
+++ b/proof/refine/ARM/Syscall_R.thy
@@ -858,6 +858,150 @@ lemma threadSet_not_curthread_ct_domain:
   apply clarsimp
   done
 
+lemma schedContextBindNtfn_invs':
+  "\<lbrace>invs' and ex_nonz_cap_to' scPtr and ex_nonz_cap_to' ntfnPtr\<rbrace>
+   schedContextBindNtfn scPtr ntfnPtr
+   \<lbrace>\<lambda>_. invs'\<rbrace>"
+  apply (clarsimp simp: schedContextBindNtfn_def)
+  apply (wpsimp wp: setSchedContext_invs' setNotification_invs' hoare_vcg_imp_lift'
+                    hoare_vcg_all_lift getNotification_wp)
+  apply (rule conjI)
+   apply (fastforce dest: ntfn_ko_at_valid_objs_valid_ntfn'
+                    simp: valid_ntfn'_def
+                   split: ntfn.splits)
+  apply (fastforce dest: sc_ko_at_valid_objs_valid_sc'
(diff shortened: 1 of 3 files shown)


proof/refine/ARM/Syscall_R.thy
```diff
@@ -858,6 +858,171 @@ lemma threadSet_not_curthread_ct_domain:
...
+      apply fastforce
+     apply (frule sc_ko_at_valid_objs_valid_sc')
+      apply fastforce
+     apply (clarsimp simp: valid_sched_context'_def scBits_simps obj_at_simps)
+    apply (wpsimp | wps)+
+  apply (clarsimp simp: ct_in_state'_def st_tcb_at'_def obj_at_simps runnable_eq_active')
+  done
+
+lemma tcbInReleaseQueue_update_st_tcb_at'[wp]:
+  "threadSet (tcbInReleaseQueue_update b) t \<lbrace>\<lambda>s. Q (st_tcb_at' P t' s)\<rbrace>"
```
corlewis, line 966: Probably unimportant if this hasn't been needed before, but maybe move this to TcbAcc_R?

corlewis: Some of the following more general lemmas could be moved earlier as well.

corlewis (approved):
Very nice!

proof/refine/ARM/Syscall_R.thy
```diff
@@ -858,6 +858,171 @@ lemma threadSet_not_curthread_ct_domain:
...
+   apply (fastforce simp: obj_at_simps valid_tcb'_def tcb_cte_cases_def comp_def
+                          valid_sched_context'_def valid_sched_context_size'_def
+                          valid_release_queue'_def inQ_def)
+   done
+
+lemma setSchedContext_ct_in_state':
+  "\<lbrace>ct_in_state' P and sc_at' ptr\<rbrace> setSchedContext ptr sc \<lbrace>\<lambda>_. ct_in_state' P\<rbrace>"
+  apply (wpsimp wp: set_sc'.set_wp)
+  apply (fastforce simp: ct_in_state'_def st_tcb_at'_def obj_at_simps ps_clear_def)
+  done
```
mktnk3, line 899: I don’t think we need `sc_at’` in the precondition. We already have a lemma about `setSchedContext` and `pred_tcb_at’`. That with `hoare_lift_Pf` or something like that should work.

mktnk3 (approved):
Nice.

proof/refine/ARM/Syscall_R.thy
michaelmcinerney: Done, and put in `SchedContext_R.thy`

PR #396, opened 2021-11-18
"add refill_unblock_check” spec update, part VI

mktnk3:
This is the last PR for the “add refill_unblock_check” spec update.

This adds `refill_unblock_check` via `if_cond_refill_unblock_check` in `send_signal`,  `receive_ipc`, and `do_reply_transfer`.


mktnk3:
The last commit plugs in the refillUnblockCheck_corres proof in Schedule_R to replace the sorry in Ipc_R (removes one sorry). This is much easier now (than when I first attempted a while ago) because maybeDonateSc doesn't call refillUnblockCheck any more.

michaelmcinerney (approved):

corlewis (approved):
Awesome, it's great to see this all come together so nicely!

PR #397, opened 2021-11-30
Current thread activatable

michaelmcinerney:
This shows that after `handle_event` and `preemption_path`, if the scheduler action is resume current thread, then the current thread is activatable. This will be important in showing that `cur_sc_offset_ready (consumed_timed s) s` and `cur_sc_offset_sufficient (consumed_time s) s` are true after `call_kernel`.

Many of the proofs are much more manual than I would have liked. I think this is because the statement is an implication, and because so many functions can potentially change the thread state of the current thread. `crunches` was unfortunately not very helpful at all. If anyone has any ideas on how to make these proofs less manual, please let me know.

### DIFF
--- a/proof/invariant-abstract/AInvs.thy
+++ b/proof/invariant-abstract/AInvs.thy
@@ -1373,4 +1373,708 @@ lemma call_kernel_consumed_time_bounded:
   apply (wpsimp wp: preemption_path_consumed_time_bounded)
   done
 
+lemma set_thread_state_schact_is_rct_imp_ct_activatable[wp]:
+  "set_thread_state ref ts \<lbrace>\<lambda>s. schact_is_rct s \<longrightarrow> ct_in_state activatable s\<rbrace>"
+  apply (clarsimp simp: set_thread_state_def set_thread_state_act_def)
+  apply (wpsimp simp: set_scheduler_action_def
+                  wp: is_schedulable_wp set_object_wp)
+  apply (clarsimp simp: schact_is_rct_def ct_in_state_def pred_tcb_at_def obj_at_def
+                        schedulable_def2 get_tcb_def)
+  done
+
+lemma reply_unlink_tcb_schact_is_rct_imp_ct_activatable[wp]:
+  "reply_unlink_tcb t r \<lbrace>\<lambda>s :: det_state. schact_is_rct s \<longrightarrow> ct_in_state activatable s\<rbrace>"
+  apply (clarsimp simp: reply_unlink_tcb_def)
+  apply (wpsimp wp: update_sk_obj_ref_wp gts_inv get_simple_ko_inv
+         | wp (once) hoare_drop_imps)+
+  apply (clarsimp simp: ct_in_state_def pred_tcb_at_def obj_at_def)
+  done
+
+lemma reschedule_required_schact_is_rct_imp_ct_activatable[wp]:
+  "\<lbrace>\<top>\<rbrace> reschedule_required \<lbrace>\<lambda>_ s. schact_is_rct s \<longrightarrow> ct_in_state activatable s\<rbrace>"
+  apply (rule_tac Q="\<lambda>_ s. scheduler_action s = choose_new_thread" in hoare_post_imp)
+   apply (clarsimp simp: schact_is_rct_def)
+  apply wpsimp
+  done
+
+lemma sched_context_donate_schact_is_rct_imp_ct_activatable[wp]:
+  "sched_context_donate sc_ptr tcb_ptr
+   \<lbrace>\<lambda>s :: det_state. schact_is_rct s \<longrightarrow> ct_in_state activatable s\<rbrace>"
+  (is "_ \<lbrace>?P\<rbrace>")
+  apply (clarsimp simp: sched_context_donate_def test_reschedule_def)
+  apply (rule hoare_seq_ext_skip, wpsimp)
+  apply (rule_tac B="\<lambda>_. ?P" in hoare_seq_ext)
+   apply (wpsimp wp: hoare_vcg_imp_lift' set_tcb_obj_ref_wp update_sched_context_wp)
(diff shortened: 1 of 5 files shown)


proof/invariant-abstract/ARM/ArchDetSchedSchedule_AI.thy
```diff
@@ -341,9 +345,7 @@ qed
 
 context Arch begin global_naming ARM
 
-lemma handle_hyp_fault_trivial[wp]:
-  "handle_hypervisor_fault t fault \<lbrace>Q\<rbrace>"
-  by (cases fault; wpsimp)
+
```
michaelmcinerney, line 348: Remove new line

proof/invariant-abstract/AInvs.thy
```diff
@@ -1373,4 +1373,711 @@ lemma call_kernel_consumed_time_bounded:
...
+  apply (clarsimp simp: reply_unlink_tcb_def)
+  apply (wpsimp wp: update_sk_obj_ref_wp gts_inv get_simple_ko_inv
+         | wp (once) hoare_drop_imps)+
+  apply (clarsimp simp: ct_in_state_def pred_tcb_at_def obj_at_def)
+  done
+
+lemma reschedule_required_schact_is_rct_imp_ct_activatable[wp]:
+  "\<lbrace>\<top>\<rbrace> reschedule_required \<lbrace>\<lambda>_ s. schact_is_rct s \<longrightarrow> ct_in_state activatable s\<rbrace>"
+  apply (rule_tac Q="\<lambda>_ s. scheduler_action s = choose_new_thread" in hoare_post_imp)
+  apply (clarsimp simp: schact_is_rct_def)
```
michaelmcinerney, line 1396: Extra space needed

mktnk3 (approved):
Nice! Great to see this working.

I am not sure what to do with the crunch issue. Maybe encapsulating this implication invariant as a definition might work, but it will be a bit of work to change, and there are other reasons that we don’t want to that, I suppose.

corlewis:
> I am not sure what to do with the crunch issue. Maybe encapsulating this implication invariant as a definition might work, but it will be a bit of work to change, and there are other reasons that we don’t want to that, I suppose.

I don't have it open to check, but from memory `crunch_wps` has `hoare_drop_imps` in it, which means that it wouldn't work well for an invariant that includes an implication. If that's the issue then you would either need to not use `crunch_wps` in any `crunches`, which would be annoying, or go with Miki's suggestion of making a definition.

If you do want to try out a definition then I would suggest either writing it as a projection or writing a lifting rule, to make it trivial to prove in the majority of simple cases.

proof/invariant-abstract/AInvs.thy
```diff
@@ -1373,4 +1373,711 @@ lemma call_kernel_consumed_time_bounded:
...
+  apply (wpsimp wp: hoare_vcg_imp_lift')
+  done
+
+crunches deleting_irq_handler
+  for schact_is_rct_imp_ct_activatable[wp]: "\<lambda>s :: det_state. schact_is_rct s \<longrightarrow> ct_in_state activatable s"
+
+lemma finalise_cap_schact_is_rct_imp_ct_activatable:
+  "finalise_cap cap final
+   \<lbrace>\<lambda>s :: det_state . (schact_is_rct s \<longrightarrow> ct_in_state activatable s)
+                      \<and> (\<exists>slot. cte_wp_at ((=) cap) slot s) \<and> invs s\<rbrace>"
```
corlewis, line 1586: This postcondition seems a bit unusual. Out of curiosity, do you remember why you needed to prove them together like this?

proof/invariant-abstract/AInvs.thy
```diff
@@ -1373,4 +1373,711 @@ lemma call_kernel_consumed_time_bounded:
...
+                    in hoare_strengthen_post)
+        apply (wpsimp wp: finalise_cap_schact_is_rct_imp_ct_activatable)
+        apply blast
+       apply (wpsimp wp: hoare_vcg_imp_lift')
+     apply (wpsimp wp: preemption_point_inv simp: ct_in_state_def)
+    apply blast+
+  done
+
+lemma cap_revoke_schact_is_rct_imp_ct_activatable:
+  "cap_revoke cap \<lbrace>\<lambda>s :: det_state. (schact_is_rct s \<longrightarrow> ct_in_state activatable s) \<and> invs s\<rbrace>"
```
corlewis, line 1644: Same question as above with this postcondition.

corlewis (approved):
Great work!

It's definitely not worth moving now, but it might be nice to have some of the more generic lemmas earlier, like the `ct_in_state` ones.

michaelmcinerney:
> > I am not sure what to do with the crunch issue. Maybe encapsulating this implication invariant as a definition might work, but it will be a bit of work to change, and there are other reasons that we don’t want to that, I suppose.
> 
> I don't have it open to check, but from memory `crunch_wps` has `hoare_drop_imps` in it, which means that it wouldn't work well for an invariant that includes an implication. If that's the issue then you would either need to not use `crunch_wps` in any `crunches`, which would be annoying, or go with Miki's suggestion of making a definition.
> 
> If you do want to try out a definition then I would suggest either writing it as a projection or writing a lifting rule, to make it trivial to prove in the majority of simple cases.

I did try making it a definition, but crunches struggles with even the most basic lemmas. `hoare_drop_imps` is in `crunch_wps`, but even not using that and using `hoare_vcg_imp_lift'` instead doesn't seem to work any better. I might just leave it as is. While the proofs are manual, they're very straightforward, so hopefully not too much of a pain

proof/invariant-abstract/AInvs.thy
michaelmcinerney: I was able to rephrase the preconditions to be a little bit more usual. The changes are in the last commit

PR #398, opened 2021-12-02
riscv machine: use address size consistently

lsf37:
In the rest of the proofs we use machine_word to refer to addresses.
This commit brings the machine definitions in line with that.


lsf37:
(We actually mostly use obj_ref once that is defined, but my point is we shouldn't hard code word32 or word64 when it's not necessary)

Xaphiosis:
This makes sense for addresses, agreed. There are sometimes ABIs that specify a width that's not intrinsically `machine_word`, but rather coincidentally, but I don't think RISC-V actually has anything like that visible in our specs. I remember something about ASIDs specified with 64-bit words but not being sure if `machine_word` works there (like `obj_ref`, the asid type is not defined yet).

Xaphiosis (approved):

PR #399, opened 2021-12-05
add some corres split rules

noneGMJ:
During my project validation, I constructed some decomposition rules for bind statements, and simple decomposition statements for if and condition. I hope this will be of some small help to you.


lsf37:
Sorry we never got to this. The rules are not wrong, but by now they should all be covered by the existing corres_underlying mechanisms and would need some cleanup to conform with the mandated proof style (see docs/ directory).

I'll close the PR, but thanks for making the effort.

PR #400, opened 2021-12-06
crefine: update for PR seL4/seL4#321

lsf37:
The aim of the PR was readability, but it actually also brings the C more in line with the spec.

Test with seL4/seL4#321


Xaphiosis:
Looks good. So I would, like basically, drop the "actually" from the commit message if I'm just being totally picky :P

Xaphiosis (approved):

lsf37:
I would like argue that this is like actually a reasonable use of actually :-) At least in the sense that this outcome wasn't expected (by me). But I can change it you think it's better.

Xaphiosis:
If you want it there, all good. Had to go look up the corresponding seL4 PR to see what you meant.

lsf37:
(had forgotten about the other architectures, they are now there)

Xaphiosis:
> (had forgotten about the other architectures, they are now there)

Sorry, I should have caught that.

lsf37:
Me too :-)

PR #404, opened 2021-12-09
Final invariants

michaelmcinerney:


proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -26446,10 +26458,12 @@ lemma handle_event_preemption_path_valid_sched:
...
 
 lemma call_kernel_valid_sched:
-  "\<lbrace>\<lambda>s. valid_sched s \<and> invs s \<and> valid_machine_time s \<and> schact_is_rct s \<and> ct_not_in_release_q s
-        \<and> ct_released s \<and> (ct_running s \<or> ct_idle s) \<and> consumed_time_bounded s
-        \<and> cur_sc_active s \<and> cur_sc_offset_ready (consumed_time s) s \<and> current_time_bounded 5 s
-        \<and> cur_sc_offset_sufficient (consumed_time s) s \<and> (e \<noteq> Interrupt \<longrightarrow> ct_running s)\<rbrace>
+  "\<lbrace>\<lambda>s. valid_sched s \<and> invs s \<and> schact_is_rct s
+        \<and> cur_sc_active s \<and> ct_not_in_release_q s
+        \<and> (ct_running s \<or> ct_idle s) \<and> (e \<noteq> Interrupt \<longrightarrow> ct_running s)
+        \<and> valid_machine_time s \<and> current_time_bounded 5 s \<and> consumed_time_bounded s
```
lsf37, line 26464: `current_time_bounded  5` looks like an oddly specific number, I'm guessing some WCET multiple from the end of time, but e should at least have a named constant for it.

lsf37 (approved):
Apart from the magic number, this looks good. Should I infer from the title that this is now the full round trip for the invariants?

proof/invariant-abstract/DetSchedSchedule_AI.thy
michaelmcinerney: Yeah `current_time_bounded n` roughly says we are the WCET plus `n` times the `MAX_PERIOD` from the end of time.

We have wondered before whether all instances of `current_time_bounded n` can be replaced with `current_time_bounded 5`. Some of the proofs for handling overrun are very tricky, so it's hard to say without trying it. Would that be the best outcome? Would you be happy for me to merge these PRs for these invariants, and to come back to this? At the very least, yes, we can make a named constant for this number

michaelmcinerney:
> Apart from the magic number, this looks good. Should I infer from the title that this is now the full round trip for the invariants?

Yes I'm pretty sure this covers the full round trip for the invariants. I'll try to clear the sorries in Refine.thy that relate to this, so we should know for sure soon

proof/invariant-abstract/DetSchedSchedule_AI.thy
lsf37: Happy for you to clean it up later, yes. As long as it actually happens :-)

proof/invariant-abstract/AInvs.thy
```diff
@@ -2060,4 +2061,496 @@ lemma handle_event_preemption_path_schact_is_rct_imp_ct_activatable:
...
+
+lemma call_kernel_cur_sc_offset_ready_and_sufficient:
+  "\<lbrace>\<lambda>s. cur_sc_offset_ready (consumed_time s) s \<and> cur_sc_offset_sufficient (consumed_time s) s
+        \<and> cur_sc_active s \<and> ct_not_in_release_q s
+        \<and> (ct_running s \<or> ct_idle s) \<and> (e \<noteq> Interrupt \<longrightarrow> ct_running s)
+        \<and> invs s \<and> schact_is_rct s \<and> valid_sched s
+        \<and> current_time_bounded 5 s \<and> valid_machine_time s \<and> consumed_time_bounded s\<rbrace>
+   call_kernel e
+   \<lbrace>\<lambda>_ s :: det_state. cur_sc_offset_ready (consumed_time s) s
+                       \<and> cur_sc_offset_sufficient (consumed_time s) s\<rbrace>"
```
corlewis, line 2517: It's definitely not worth changing any of this, but out of curiosity do you remember whether it was necessary to prove these together?

corlewis (approved):
Great job sorting all of these preconditions out!

proof/invariant-abstract/AInvs.thy
michaelmcinerney: Definitely not necessary, but it did cut down on a lot of repetition. Sometimes readiness and sufficiency are dealt with separately: for example, in `preemption_point`, an sc that is ready will remain ready as time advances, but might become insufficient as the consumed time increases. But a lot of the time they're dealt with together, and `cur_sc_more_than_ready`, which is a major ingredient of these proofs, has them together

mktnk3 (approved):
Nice :)

PR #406, opened 2021-12-16
`kernel_entry_invs`

michaelmcinerney:
This proves `kernel_entry_invs`. 

This also makes some modificatin to `full_invs` and `full_invs'`. I'm not entirely sure that these are OK, or whether they're optimal. The final theorem, `refinement`, works without modification to either `ADT_H` or `ADT_A`, so I think something like what is here is OK.

I have made an effort to weaken the postconditions of a couple `invs'` lemmas which are still sorried. I don't think it's of any benefit to weaken the concrete guards of any of the remaining `corres` rules - and I think doing so would also break some of the proofs here which are a little annoying.

### DIFF
--- a/proof/refine/ARM/Refine.thy
+++ b/proof/refine/ARM/Refine.thy
@@ -228,73 +228,166 @@ lemma absKState_correct:
 
 text \<open>The top-level invariance\<close>
 
-lemma kernel_entry_invs:
-  "\<lbrace>einvs and (\<lambda>s. e \<noteq> Interrupt \<longrightarrow> ct_running s)
-    and (\<lambda>s. 0 < domain_time s) and valid_domain_list and (ct_running or ct_idle)
-    and (\<lambda>s. scheduler_action s = resume_cur_thread)\<rbrace>
+lemma kernel_entry_invs_det_ext:
+  "\<lbrace>\<lambda>s. invs s \<and> schact_is_rct s \<and> cur_sc_active s \<and> ct_not_in_release_q s
+          \<and> (e \<noteq> Interrupt \<longrightarrow> ct_running s)\<rbrace>
    kernel_entry e us
-   \<lbrace>\<lambda>_. einvs and (\<lambda>s. ct_running s \<or> ct_idle s)
-        and (\<lambda>s. 0 < domain_time s) and valid_domain_list
-        and (\<lambda>s. scheduler_action s = resume_cur_thread)\<rbrace>"
-  apply (rule_tac Q="\<lambda>rv. invs and (\<lambda>s. ct_running s \<or> ct_idle s) and valid_sched and
-                           (\<lambda>s. 0 < domain_time s) and valid_domain_list and
-                           valid_list and (\<lambda>s. scheduler_action s = resume_cur_thread)"
+   \<lbrace>\<lambda>_ s :: det_state. invs s \<and> (ct_running s \<or> ct_idle s)\<rbrace>"
+  apply (simp add: kernel_entry_def)
+  apply (wp akernel_invs_det_ext thread_set_invs_trivial thread_set_ct_in_state select_wp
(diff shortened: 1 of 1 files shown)


proof/refine/ARM/Refine.thy
```diff
@@ -228,73 +228,166 @@ lemma absKState_correct:
...
+  "\<lbrace>\<lambda>s. valid_sched s \<and> invs s \<and> schact_is_rct s
+        \<and> cur_sc_active s \<and> ct_not_in_release_q s
+        \<and> (ct_running s \<or> ct_idle s) \<and> (e \<noteq> Interrupt \<longrightarrow> ct_running s)
+        \<and> valid_machine_time s \<and> current_time_bounded 5 s \<and> consumed_time_bounded s
+        \<and> cur_sc_offset_ready (consumed_time s) s
+        \<and> cur_sc_offset_sufficient (consumed_time s) s\<rbrace>
+   kernel_entry e us
+   \<lbrace>\<lambda>_. valid_sched :: det_state \<Rightarrow> _\<rbrace>"
+  apply (simp add: kernel_entry_def)
+  apply (wp call_kernel_valid_sched  thread_set_invs_trivial thread_set_ct_in_state
```
michaelmcinerney, line 252: Extra spaces in this

proof/refine/ARM/Refine.thy
```diff
@@ -228,73 +228,166 @@ lemma absKState_correct:
...
+                                          \<and> (m = KernelMode \<longrightarrow> e \<noteq> None)
+                                          \<and> (m = UserMode \<longrightarrow> ct_running s)
+                                          \<and> (m = IdleMode \<longrightarrow> ct_idle s)
+                                          \<and> (e \<noteq> None \<and> e \<noteq> Some Interrupt \<longrightarrow> ct_running s)}"
+
+crunches do_user_op, check_active_irq
+  for valid_list[wp]: valid_list
+  and valid_sched[wp]: valid_sched
+  and sched_act[wp]: "\<lambda>s. P (scheduler_action s)"
+  and domain_time[wp]: "\<lambda>s. 0 < domain_time s"
```
michaelmcinerney, line 342: I think this can be strengthened to `\lambdas. P (domain_time s)`

proof/refine/ARM/Refine.thy
```diff
@@ -811,30 +954,48 @@ lemma ckernel_invariant:
     apply (clarsimp simp: ex_abs_def lift_state_relation_def)
     apply (frule akernel_init_invs[THEN bspec])
     apply (rule_tac x = s in exI)
-    apply (clarsimp simp: Init_A_def)
+    apply (clarsimp simp: Init_A_def )
+
+    apply (clarsimp simp: Init_A_def init_A_st_def )
```
michaelmcinerney, line 959: Extra spaces before these end parentheses

lsf37 (approved):
This is looking very promising!

We should try to find a better name than `abstract_bundle`, because this name is visible in all further images down the chain (and `bundle` is an existing Isabelle concept which will confuse those who use it). Would `mcs_invs` be terrible?

michaelmcinerney:
> This is looking very promising!
> 
> We should try to find a better name than `abstract_bundle`, because this name is visible in all further images down the chain (and `bundle` is an existing Isabelle concept which will confuse those who use it). Would `mcs_invs` be terrible?

`abstract_bundle` is an input abbreviation, so it won't be visible, but yeah, it's not the greatest name either, and I can definitely change it. `mcs_invs` is good. I will go with that unless we can think up something better

lsf37:
> `abstract_bundle` is an input abbreviation, so it won't be visible

Should have been more precise -- I meant in the name space (i.e. the name is now used). Input abbreviations are still visible there, they are just not printed in the output.

mktnk3 (approved):
Based on what is in this PR, I could close the remaining sorries in Refine.thy, and it all seemed to work nicely.

PR #407, opened 2021-12-16
Remove parameter from `current_time_bounded`

michaelmcinerney:
Most of the diff is entirely uninteresting - you can probably just ignore `DetSchedSchedule_AI.thy`. Apart from removing the various numbers, several lemmas which are now unnecessary have also been removed

### DIFF
--- a/proof/invariant-abstract/AInvs.thy
+++ b/proof/invariant-abstract/AInvs.thy
@@ -635,7 +635,7 @@ lemma schedule_cur_sc_active:
 lemma call_kernel_cur_sc_active:
   "\<lbrace>\<lambda>s. cur_sc_active s \<and> invs s \<and> schact_is_rct s \<and> valid_sched s \<and> ct_not_in_release_q s
         \<and> (ct_running s \<or> ct_idle s) \<and> (e \<noteq> Interrupt \<longrightarrow> ct_running s)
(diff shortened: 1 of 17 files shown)


lib/Simulation.thy
```diff
@@ -96,7 +96,7 @@ where
 
 text \<open>
   Refinement is a global property over all executions and/or all
-  hoare triples. As this is hard to show, we define the weaker concept
+  hoare triples. As this is hard to show, we define the stronger concept
```
lsf37, line 99: I think it turns out that they are actually equivalent, so probably better to just remove the word entirely (just "we define the concept of")

lib/Simulation.thy
```diff
@@ -51,7 +51,7 @@ definition
 
 text \<open>
   The sequence of operations in the data type is then executed
-  in an initial state by initialising the private state, executing
+  in an initial state by initialising the observable state, executing
```
lsf37, line 54: It has to do both. Initially we only had trivial observable state, so the private state was the important bit, but I would vote for just saying "initialising the state".

lsf37 (approved):
Cool, thanks for doing that, it's much nicer without the concrete numbers.

Looks like the PR needs a rebase before it can be merged.

corlewis (approved):
This looks great!

PR #409, opened 2021-12-20
Prove `he_invs’`

mktnk3:
This proves `he_invs’` (`invs’` for `handleEvent`).

This also introduces three asserts for the properties we want to have in `handleEvent`, after the calls to first phase `handleInvocation` prior to `handleRecv`. In the proof for `handleEvent_corres` (which will be in a separate PR after this), we can validate these asserted properties by crossing over from the abstract side. For the proof in this PR, only one of the three asserts is needed (`ct_active’`).

### DIFF
--- a/proof/refine/ARM/Syscall_R.thy
+++ b/proof/refine/ARM/Syscall_R.thy
@@ -2573,7 +2573,7 @@ crunches check_budget
   for cur_tcb[wp]: cur_tcb
   and pspace_aligned[wp]: pspace_aligned
   and pspace_distinct[wp]: pspace_distinct
-  (wp: crunch_wps)
+  (wp: crunch_wps simp: crunch_simps)
 
 crunches checkBudgetRestart
   for valid_duplicates''[wp]: "\<lambda>s. vs_valid_duplicates' (ksPSpace s)"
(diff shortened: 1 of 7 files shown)


spec/haskell/src/SEL4/Model/StateData.lhs
```diff
@@ -448,3 +448,14 @@ is true using the rule cur_tcb'_cross.
 
 > cur_tcb'_asrt :: KernelState -> Bool
 > cur_tcb'_asrt _ = True
+
+Asserts for first phase, non-blocking, non-calling handle_invocation.
```
michaelmcinerney, line 452: I suppose these asserts could be used elsewhere, so maybe we want to rephrase this comment

proof/refine/ARM/Syscall_R.thy
```diff
@@ -2466,9 +2466,156 @@ crunches handleFault, receiveSignal, receiveIPC, asUser
...
+  and ksCurThread[wp]: "\<lambda>s. P (ksCurThread s)"
+  (wp: crunch_wps simp: crunch_simps)
+
+lemma getCurrentTime_invs'[wp]:
+  "doMachineOp getCurrentTime \<lbrace>invs'\<rbrace>"
+  apply (simp add: getCurrentTime_def modify_def)
+  apply (wpsimp wp: dmo_invs' simp: modify_def)
+  by (simp add: in_get put_def gets_def in_bind in_return)
+
+lemma invs'_ksCurTime_update[iff]:
```
michaelmcinerney, line 2561: I'm not bothered if you don't move this, but I know there are some "update" rules for `invs'` in a much earlier file, and this might be better there

michaelmcinerney (approved):
Nice

PR #411, opened 2021-12-21
abs-config: lift ASpec+ExecSpec config values from C config

lsf37:
This is the result of an investigation into lifting the configuration values the user can set in the kernel build system directly into the abstract and executable specifications without manual intervention. The investigation was successful, and this PR implements the feature.

Specifically, we add:
 - a CMake build target in `cspec` for just generating config (without needing compiler or preprocessor tool chain)
 - a python script for parsing the generated config header and generating Isabelle `Kernel_Config` theories for each `L4V_ARCH` configuration. Changing these kernel configs will now change values such as `numDomains`, `retypeFanOutLimit`, `timeSlice` etc

The PR also make long names mandatory for definition unfolding of these constants, so it becomes clear which proofs depend on their value and might fail because of a change.
 
This does not necessarily make the proofs generic in these constants, but you can now set different values in the kernel build system and rerun the proofs to see if they still work. There is a good chance that this works for some of them (e.g. `CONFIG_TIME_SLICE`, `CONFIG_RESET_CHUNK_BITS`, `CONFIG_MAX_NUM_WORK_UNITS_PER_PREEMPTION`). More work like PR #410 is needed for other values, but with together PR #410 `CONFIG_NUM_DOMAINS` is now the only value that needs to be changed for adjusting the number of domains.

This PR is on top of #410 and should be merged afterwards.

Test with seL4/seL4#713


lsf37:
A sample generated config file for ARM looks like this:

```isabelle
(*
 * Copyright 2021, Proofcraft Pty Ltd
 *
 * SPDX-License-Identifier: GPL-2.0-only
 *)

chapter "Kernel Configuration"

theory Kernel_Config
imports Platform
begin

(*
  GENERATED -- DO NOT EDIT! Changes will be overwritten.

  This file was generated from ~/ver/l4v/spec/cspec/c/config-build/ARM/gen_config/kernel/gen_config.h
  by the script ~/ver/l4v/spec/cspec/c/gen-config-thy.py.
*)

definition CONFIG_WORD_SIZE :: nat where
  "CONFIG_WORD_SIZE \<equiv> 32"

definition CONFIG_USER_TOP :: machine_word where
  "CONFIG_USER_TOP \<equiv> 0xa0000000"

definition CONFIG_L1_CACHE_LINE_SIZE_BITS :: nat where
  "CONFIG_L1_CACHE_LINE_SIZE_BITS \<equiv> 5"

definition CONFIG_PADDR_USER_DEVICE_TOP :: machine_word where
  "CONFIG_PADDR_USER_DEVICE_TOP \<equiv> 4294967295"

definition CONFIG_ROOT_CNODE_SIZE_BITS :: nat where
  "CONFIG_ROOT_CNODE_SIZE_BITS \<equiv> 12"

definition CONFIG_TIMER_TICK_MS :: machine_word where
  "CONFIG_TIMER_TICK_MS \<equiv> 2"

definition timeSlice :: nat where
  "timeSlice \<equiv> 5"  (* CONFIG_TIME_SLICE *)

definition retypeFanOutLimit :: machine_word where
  "retypeFanOutLimit \<equiv> 256"  (* CONFIG_RETYPE_FAN_OUT_LIMIT *)

definition workUnitsLimit :: nat where
  "workUnitsLimit \<equiv> 100"  (* CONFIG_MAX_NUM_WORK_UNITS_PER_PREEMPTION *)

definition resetChunkBits :: nat where
  "resetChunkBits \<equiv> 8"  (* CONFIG_RESET_CHUNK_BITS *)

definition CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS :: nat where
  "CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS \<equiv> 230"

definition numDomains :: nat where
  "numDomains \<equiv> 16"  (* CONFIG_NUM_DOMAINS *)

definition numPriorities :: nat where
  "numPriorities \<equiv> 256"  (* CONFIG_NUM_PRIORITIES *)

definition CONFIG_MAX_NUM_NODES :: nat where
  "CONFIG_MAX_NUM_NODES \<equiv> 1"

definition CONFIG_KERNEL_STACK_BITS :: nat where
  "CONFIG_KERNEL_STACK_BITS \<equiv> 12"

definition CONFIG_MAX_NUM_TRACE_POINTS :: nat where
  "CONFIG_MAX_NUM_TRACE_POINTS \<equiv> 0"

definition CONFIG_USER_STACK_TRACE_LENGTH :: nat where
  "CONFIG_USER_STACK_TRACE_LENGTH \<equiv> 0"


(* These definitions should only be unfolded consciously and carefully: *)
hide_fact (open) CONFIG_WORD_SIZE_def
hide_fact (open) CONFIG_USER_TOP_def
hide_fact (open) CONFIG_L1_CACHE_LINE_SIZE_BITS_def
hide_fact (open) CONFIG_PADDR_USER_DEVICE_TOP_def
hide_fact (open) CONFIG_ROOT_CNODE_SIZE_BITS_def
hide_fact (open) CONFIG_TIMER_TICK_MS_def
hide_fact (open) timeSlice_def
hide_fact (open) retypeFanOutLimit_def
hide_fact (open) workUnitsLimit_def
hide_fact (open) resetChunkBits_def
hide_fact (open) CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS_def
hide_fact (open) numDomains_def
hide_fact (open) numPriorities_def
hide_fact (open) CONFIG_MAX_NUM_NODES_def
hide_fact (open) CONFIG_KERNEL_STACK_BITS_def
hide_fact (open) CONFIG_MAX_NUM_TRACE_POINTS_def
hide_fact (open) CONFIG_USER_STACK_TRACE_LENGTH_def

end
```

lsf37:
The theory linter fails because the file `Kernel_Config` is mentioned in the diff, but is removed. Will put up an issue to fix the linter.

lsf37:
@mbrcknl adding you to the reviewers in case you wanted to have a look (IIRC, you were involved with some of the C build setup). No pressure, though.

lsf37:
(going to let the test runs finish before I rebase on the new `numdomains` state)

spec/Makefile
```diff
@@ -64,8 +64,13 @@ c-kernel: c-parser .FORCE
...
 .PHONY: c-kernel
 
+# Generate C config headers
+c-config: .FORCE
+	cd cspec/c && L4V_REPO_PATH=$(L4V_REPO_PATH) L4V_ARCH=$(L4V_ARCH) $(MAKE) config
+.PHONY: c-config
+
 # Run the haskell translator
-design-spec: .FORCE
+design-spec: c-config .FORCE
```
Xaphiosis, line 73: sanity check: ASpec depends on design-spec, right?

spec/cspec/c/gen-config-thy.py
```diff
@@ -0,0 +1,274 @@
...
+#
+# This script generates a theory Kernel_Config.thy from the file gen_config.h,
+# which in turn is produced by the kernel build system, containing kernel
+# configuration options.
+#
+# The script makes the following assumptions about gen_config.h:
+#
+#   - only lines that start with "#define" are config options that are used
+#   - the value of the config option is separated from its name by two spaces
+#   - there is an optional comment after the value, separated by two spaces
```
Xaphiosis, line 17: perhaps an example? proverbial picture is worth a thousand words?

spec/cspec/c/kernel.mk
```diff
@@ -115,3 +115,16 @@ ${KERNEL_BUILD_ROOT}/.cmake_done: ${KERNEL_DEPS} ${CONFIG_DOMAIN_SCHEDULE}
 
 ${UMM_TYPES}: ${KERNEL_BUILD_ROOT}/kernel_all.c_pp
 	${CSPEC_DIR}/mk_umm_types.py --root $(L4V_REPO_PATH) ${KERNEL_BUILD_ROOT}/kernel_all.c_pp $@
+
+# This target generates config files and headers only. It does not invoke
+# the C tool chain or preprocessor. We force CMake to skip tests for these,
+# so that ASpec and ExecSpec can be built with fewer dependencies.
```
Xaphiosis, line 121: out of curiosity: how did you manage this, via no-compiler.cmake? how do you stop it trying to build sources?

spec/Makefile
lsf37: correct

spec/machine/Kernel_Config_Lemmas.thy
```diff
@@ -0,0 +1,28 @@
...
+
+theory Kernel_Config_Lemmas
+imports "$L4V_ARCH/Kernel_Config"
+begin
+
+text \<open>
+  seL4's build system allows configuration of some architecture-independent constants, such as the
+  number of domains.
+
+  The long-term goal is to make the proofs resilient in the face of changes of these configuration
```
Xaphiosis, line 15: well, for some of them, don't think we want to commit to many or most of them

spec/cspec/c/kernel.mk
lsf37: `cmake` itself doesn't build, it's more like `autoconf` and friends, it just generates info for another build system. That other build system could be `make` for instance, but for seL4 it's `ninja`. For this one we just never call `ninja`.

The one remaining bit is that `cmake` tests for compiler availability when it does all it's build system generation stuff, and the `no-compiler.cmake` settings force it to skip most of that. It will still check that *some* compiler is installed, but it will not demand any more that it works or that it works for the selected platform.

spec/machine/Kernel_Config_Lemmas.thy
Xaphiosis: the really dumb thing is I originally wrote that :/ Fine to leave

Xaphiosis (approved):

spec/machine/Kernel_Config_Lemmas.thy
lsf37: I'd say longer term we do want that for most of them -- at least to the degree that it becomes very easy to check if config changes are expected to pass the proofs or not. But that is for actual config parameters, not for selecting feature combinations or platform/architecture settings.

spec/cspec/c/gen-config-thy.py
lsf37: Yes, good idea. Will do.

- [ ] add example

Xaphiosis:
I really like this change, it creates a lot of future improvement potential, and also syncs up the variables. It was super annoying to finish abstract/design proofs, and then build CKernel+CSpec, and in CRefine realise oh, the constant is different in the C, meaning a full rebuild. 

Meta thought/concern: original plan for Kernel_Config and its hidden definitions was to indicate that unfolding of Kernel_Config.*_def is strongly discouraged. With these changes, we get a lot of such unfolds, decreasing the deterrent effect. I don't think this is something to address right now, but I'm interested in how you see this play out.

lsf37:
> Meta thought/concern: original plan for Kernel_Config and its hidden definitions was to indicate that unfolding of Kernel_Config.*_def is strongly discouraged. With these changes, we get a lot of such unfolds, decreasing the deterrent effect. I don't think this is something to address right now, but I'm interested in how you see this play out.

Ideally we'd do a pass similar to the one for `numDomains` to make it actually generic in the constants that get unfolded a lot. They probably don't need to be, and like for `numDomains` it's likely to improve the proofs. Main question is when we will have time for it.

spec/cspec/c/kernel.mk
Xaphiosis: Great. The part I missed is that it's cmake that generates the config, and not a part of a target in a Makefile-analogue (in this case ninja). Now I got it.

spec/machine/Kernel_Config_Lemmas.thy
Xaphiosis: I agree. All good.

Xaphiosis:
> Ideally we'd do a pass similar to the one for `numDomains` to make it actually generic in the constants that get unfolded a lot. They probably don't need to be, and like for `numDomains` it's likely to improve the proofs. Main question is when we will have time for it.

Indeed. `numDomains` genericity proved more time-consuming than I anticipated, due to the large amount of unfolds and magic numbers and the many test runs needed. However, my points are incoherent:
- how do we stop people unfolding numDomains_def again given they see the other constants unfolded all over?
- your path forwards sounds good, and we're planning the weekly test to check multiple domains with the help of the above, but if we generalise other variables that's going to be a *lot* of tests as well; not something solvable at the moment

lsf37:
> * your path forwards sounds good, and we're planning the weekly test to check multiple domains with the help of the above, but if we generalise other variables that's going to be a _lot_ of tests as well; not something solvable at the momen

Yes, ideally we'd somehow extract all necessary conditions into a small theory that can be checked mostly independently. Unclear how well that can work without having access to at least the content in `CKernel`, but even checking a small theory on top of `CKernel` would already be much faster than checking all proofs. For that we'd need to get to a point where those constants are guaranteed to not be unfolded anywhere. The work is never done :-)

corlewis (approved):
This looks really good to me as well (with the caveat that I didn't look too closely at gen-config-thy.py).

I agree with what @Xaphiosis said, it's really exciting to see progress in this direction, with all of the potential improvements and clean-up that it opens up.

spec/cspec/c/gen-config-thy.py
```diff
@@ -0,0 +1,274 @@
...
+bool = 'boolean'
+string = 'string'
+nat = 'nat'
+word = 'machine_word'
+
+# All known config keys that could be set in a gen_config.h file.
+# The dict maps config key to (type, definition name)
+#
+# This table duplicates some information from the kernel build system, but
+# only which keys exist, not their values.
```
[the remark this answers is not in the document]
lsf37: It will by default ignore new keys. The idea is that the proofs will fail if the value was used in a verification-visible way in the C, but should keep succeeding when not. That way this won't block things if a new debug setting is added for instance, but will fail the regression test if the verification-relevant values are refactored somehow.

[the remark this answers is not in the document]
corlewis: Perfect, that's even better! Thanks for the clarification.

lsf37:
(rebased on master)

Will add the example now and then squash/merge.

PR #415, opened 2021-12-23
prove handleEvent_corres + fix sorries in Refine.thy

mktnk3:
This removes the last sorry in Syscal_R.thy.

The main proof and its helper lemmas make use of the asserts added in https://github.com/seL4/l4v/pull/409.

Update: this PR now contains #416 , which removes all the remaining sorries in Refine. Since this PR touches quite a lot of places in Syscall_R, it should make sense to review the two together rather than one after the other.

### DIFF
--- a/proof/refine/ARM/Refine.thy
+++ b/proof/refine/ARM/Refine.thy
@@ -499,18 +499,29 @@ lemma akernel_invariant:
   done
 
 lemma ckernel_invs:
-  "\<lbrace>invs' and (\<lambda>s. vs_valid_duplicates' (ksPSpace s)) and
-               (\<lambda>s. e \<noteq> Interrupt \<longrightarrow> ct_running' s) and
-               (\<lambda>s. ksSchedulerAction s = ResumeCurrentThread)\<rbrace>
+  "\<lbrace>invs' and (\<lambda>s. vs_valid_duplicates' (ksPSpace s))
+    and (\<lambda>s. e \<noteq> Interrupt \<longrightarrow> ct_running' s) and (ct_running' or ct_idle')
+    and (\<lambda>s. is_active_sc' (ksCurSc s) s) and sym_heap_tcbSCs
+    and (\<lambda>s. obj_at' (\<lambda>sc. scTCB sc = Some (ksCurThread s)) (ksCurSc s) s)
+    and (\<lambda>s. pred_map (\<lambda>tcb. \<not> tcbInReleaseQueue tcb) (tcbs_of' s) (ksCurThread s))
+    and (\<lambda>s. ksSchedulerAction s = ResumeCurrentThread)\<rbrace>
    callKernel e
    \<lbrace>\<lambda>_. invs'\<rbrace>"
-  apply (simp add: callKernel_def)
+  apply (simp add: callKernel_def mcsPreemptionPoint_def)
   apply (rule hoare_pre)
-   apply (wp activate_invs'  schedule_sch
-             schedule_sch_act_simple he_invs' schedule_invs'
-             hoare_drop_imp[where R="\<lambda>_. kernelExitAssertions"]
-          | simp add: no_irq_getActiveIRQ)+
-  sorry
+   apply (wpsimp wp: hoare_drop_imp[where R="\<lambda>_. kernelExitAssertions"] activate_invs')
+     apply (rule hoare_drop_imp)
+     apply (wpsimp wp: schedule_invs')
+    apply (wpsimp wp: stateAssert_wp)
+   apply (wpsimp wp: isSchedulable_wp hoare_drop_imp)
+    apply (intro iffI; clarsimp simp: isScActive_def isSchedulable_bool_def)
+   apply (rule hoare_post_impErr[where E="\<lambda>_. invs'" and Q=Q and R=Q for Q])
+     apply wpsimp
+    apply (clarsimp simp: active_from_running')+
+  apply (clarsimp simp: sym_heap_def pred_map_def)
+  apply (rule_tac x="ksCurSc s" in exI)
+  apply (clarsimp simp: obj_at_simps is_active_sc'_def isScActive_def opt_map_red pred_map_def)
+  done
 
 (* abstract and haskell have identical domain list fields *)
 abbreviation valid_domain_list' :: "'a kernel_state_scheme \<Rightarrow> bool" where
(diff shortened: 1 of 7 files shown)


proof/refine/ARM/Refine.thy
```diff
@@ -499,18 +499,29 @@ lemma akernel_invariant:
...
 
 lemma ckernel_invs:
-  "\<lbrace>invs' and (\<lambda>s. vs_valid_duplicates' (ksPSpace s)) and
-               (\<lambda>s. e \<noteq> Interrupt \<longrightarrow> ct_running' s) and
-               (\<lambda>s. ksSchedulerAction s = ResumeCurrentThread)\<rbrace>
+  "\<lbrace>invs' and (\<lambda>s. vs_valid_duplicates' (ksPSpace s))
+    and (\<lambda>s. e \<noteq> Interrupt \<longrightarrow> ct_running' s) and (ct_running' or ct_idle')
+    and (\<lambda>s. is_active_sc' (ksCurSc s) s) and sym_heap_tcbSCs
+    and (\<lambda>s. obj_at' (\<lambda>sc. scTCB sc = Some (ksCurThread s)) (ksCurSc s) s)
+    and (\<lambda>s'. pred_map (\<lambda>tcb. \<not> tcbInReleaseQueue tcb) (tcbs_of' s') (ksCurThread s'))
```
lsf37, line 506: Any specific reason for the `s'` in this one instead of `s`?

proof/refine/ARM/Refine.thy
```diff
@@ -525,23 +536,47 @@ lemma callKernel_domain_time_left:
...
-           (ct_running' or ct_idle') and
-           (\<lambda>s. vs_valid_duplicates' (ksPSpace s)) and
-           (\<lambda>s. ksSchedulerAction s = ResumeCurrentThread) and
-           (\<lambda>s. 0 < ksDomainTime s) and valid_domain_list' \<rbrace>
+  "\<lbrace> invs' and (\<lambda>s. e \<noteq> Interrupt \<longrightarrow> ct_running' s)
+    and (ct_running' or ct_idle')
+    and (\<lambda>s. vs_valid_duplicates' (ksPSpace s))
+    and (\<lambda>s. is_active_sc' (ksCurSc s) s) and sym_heap_tcbSCs
+    and (\<lambda>s. obj_at' (\<lambda>sc. scTCB sc = Some (ksCurThread s)) (ksCurSc s) s)
+    and (\<lambda>s'. pred_map (\<lambda>tcb. \<not> tcbInReleaseQueue tcb) (tcbs_of' s') (ksCurThread s'))
```
lsf37, line 562: `s'` likely copy/paste from above

lsf37 (approved):
Very nice!

We're adding a bunch of new preconditions to `kernel_corres` and `entry_corres`. I can't really tell from the diff if these are part of the ADT invariants, i.e. if we know that these conditions are established at kernel exit.

lsf37:
> We're adding a bunch of new preconditions to `kernel_corres` and `entry_corres`. I can't really tell from the diff if these are part of the ADT invariants, i.e. if we know that these conditions are established at kernel exit.

Loading it up in Isabelle, I can see this is all good, most of the bits are crossed over from the abstract side and we have the full ADT refinement at the bottom of the theory. Very nice job!

proof/refine/ARM/Syscall_R.thy
```diff
@@ -2677,6 +2593,400 @@ proof -
...
+     (einvs and ct_running and (\<lambda>s. scheduler_action s = resume_cur_thread) and
+      valid_machine_time and
+      current_time_bounded 5 and
+      consumed_time_bounded and
+      cur_sc_active and
+      ct_released and
+      ct_not_in_release_q and
+      ct_not_queued and
+      (\<lambda>s. cur_sc_offset_ready (consumed_time s) s) and
+      (\<lambda>s. cur_sc_offset_sufficient (consumed_time s) s))
```
michaelmcinerney, line 2639: `mcs_invs` can be moved earlier and used here

proof/refine/ARM/Syscall_R.thy
```diff
@@ -2677,6 +2593,400 @@ proof -
...
+     apply wpsimp
+    apply (clarsimp simp: invs_def valid_state_def valid_pspace_def active_from_running
+                          schedulable_def2 released_sc_tcb_at_def)
+    apply (clarsimp simp: ct_in_state_def st_tcb_weakenE)
+   apply (clarsimp simp: active_from_running')
+  apply simp
+  done
+
+abbreviation (input)
+  "a_pre \<equiv> (einvs and ct_running and
```
michaelmcinerney, line 2718: This is almost the same as `mcs_invs`. Can they be consolidated or reused somehow?

proof/refine/ARM/Syscall_R.thy
```diff
@@ -2677,6 +2593,400 @@ proof -
...
+
+lemma handleEvent_corres:
+  "corres (dc \<oplus> dc)
+     (einvs and (\<lambda>s. event \<noteq> Interrupt \<longrightarrow> ct_running s)
+      and (\<lambda>s. scheduler_action s = resume_cur_thread)
+      and valid_machine_time and current_time_bounded 5
+      and consumed_time_bounded and cur_sc_active and (ct_active or ct_idle)
+      and ct_not_in_release_q and ct_not_queued
+      and (\<lambda>s. cur_sc_offset_ready (consumed_time s) s)
+      and (\<lambda>s. cur_sc_offset_sufficient (consumed_time s) s))
```
michaelmcinerney, line 2839: `mcs_invs`

proof/refine/ARM/Syscall_R.thy
```diff
@@ -2677,6 +2593,400 @@ proof -
...
+     apply (frule (1) invs_strengthen_cur_sc_tcb_are_bound[OF _ invs_cur_sc_tcb])
+      apply (clarsimp simp: invs_def valid_state_def valid_pspace_def)
+     apply (erule (5) schact_is_rct_ct_released)
+     apply (erule (2) cur_sc_not_idle_sc_ptr')
+    apply (case_tac event)
+         apply (simp_all add: handleEvent_def)
+         apply (rename_tac syscall)
+         apply (rule updateTimeStamp_checkBudgetRestart_helperE)
+         apply (case_tac syscall; simp)
+                   apply (auto intro: corres_guard_imp[OF handleSend_corres]
```
michaelmcinerney, line 2874: This `auto` needs to be removed. Maybe an Eisbach method, similar to the Eisbach methods we have in `DetSchedSchedule` for the `handle_event` Hoare triples, would be a good idea (though they are annoying to fix when they break...)

michaelmcinerney (approved):

proof/refine/ARM/Refine.thy
```diff
@@ -617,11 +652,178 @@ lemma doUserOp_valid_duplicates':
...
   done
 
+(* FIXME RT: generalise similar lemmas in InterruptAcc_R with this one *)
+lemma ct_in_state_irq_state_independent[intro!, simp]:
+  "ct_in_state P (s \<lparr>machine_state := machine_state s \<lparr>irq_state := f (irq_state (machine_state s)) \<rparr> \<rparr>)
+   = ct_in_state P s"
+  by (simp add: ct_in_state_def)
+
+lemma valid_tcbs_machine_state_update[iff]:
+  "valid_tcbs (machine_state_update f s) = valid_tcbs s"
```
corlewis, line 662: Low priority, but this lemma could be moved as well

proof/refine/ARM/Refine.thy
```diff
@@ -630,59 +832,154 @@ lemma kernel_corres':
...
        apply (rule corres_split_handle [OF _ handleEvent_corres])
+         (* handle *)
+         apply (rule kernel_preemption_corres)
+        apply (rule_tac E="\<lambda>_ s. einvs s \<and> scheduler_act_sane s \<and> cur_sc_chargeable s \<and>
+                                 (schact_is_rct s \<longrightarrow> cur_sc_active s) \<and> ct_not_queued s \<and>
+                                 (schact_is_rct s \<longrightarrow> ct_in_state activatable s) \<and>
+                                 (cur_sc_active s \<longrightarrow> cur_sc_offset_ready (consumed_time s) s) \<and>
+                                 ct_not_blocked s \<and> current_time_bounded 5 s \<and>
+                                 ct_not_in_release_q s \<and> consumed_time_bounded s"
+              in hoare_post_impErr[where Q=Q and R=Q for Q])
```
corlewis, line 880:
```suggestion
               in hoare_post_impErr[where Q=Q and R=Q for Q])
```

corlewis (approved):
This is awesome, great job pushing it through to the end!

proof/refine/ARM/Syscall_R.thy
mktnk3: This auto doesn’t need to be removed. This proof has pretty much (but not exactly) the same structure as the master one up to this point. This auto discharges all the subgoals that it is applied to. This auto may be removed if you like altogether with the master one, I suppose, but that would be total restructuring the proofs which is beyond the purpose of the current PR (which is to fix the sorries as the first priority).

proof/refine/ARM/Syscall_R.thy
mktnk3: These abbreviations (`pre_a` and `pre_c`) are only meant to be used very locally in the two helper lemmas that follow, and the purpose is to ensure that they should have same preconditions. I wouldn’t have needed to introduce them if I could unify these two helper lemmas that are almost identical but in different monads. It would be nice if we can abstract away the structure somehow to avoid duplication, but our current corres framework doesn’t seem to allow that.

mktnk3:
I am not moving `mcs_invs` because there are enough deviations in this PR it seems that doesn’t seem to make doing so worthwhile. I also think that it makes sense to have it close to `full_invs` at this point, maybe till we know the final version of it. If this is final enough, the right place for it is probably not Syscall_R but somewhere in AInvs DetSched*.  Anyway, that is again beyond the scope of this PR.

proof/refine/ARM/Syscall_R.thy
lsf37: In terms of our style rules, this `auto` is permissible, yes. The important part is that it is only applied to goals that it solves, which seems to be the case here.

PR #416, opened 2021-12-26
fix sorries in Refine.thy

mktnk3:
This PR removes all the remaining sorries in Refine.


mktnk3:
Merged into #415 so that we can review all the changes together.

PR #417, opened 2022-01-06
RT merge from master

lsf37:
This merge includes the following work from `master`:

- infoflow arch split + RISCV
- generic numDomains
- kernel config lifting from C

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@
 *#*#
 
 /spec/cspec/c/build
+/spec/cspec/c/config-build
 
 /spec/haskell/stack.yaml.lock
 /spec/haskell/doc/**/*.aux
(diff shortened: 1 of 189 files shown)


lsf37:
As always, these merges are hard to review. The bit after the merge, i.e. 6c48989..4608cc2 might be interesting to look at -- those are the proof updates for numDomains and kernel config lifting. Relatively minor, but includes changes to Haskell which also bring Haskell more in line with C (on MCS).

michaelmcinerney (approved):
I carefully looked at the commits in the range you mentioned, and it all looks good. If you'd like me to look at some other commits, just let me know which would be good, and I can do that.

I'm glad that we could keep the proofs about the domain time (with the implication) with minimal change.

lsf37:
> I carefully looked at the commits in the range you mentioned, and it all looks good. If you'd like me to look at some other commits, just let me know which would be good, and I can do that.

All good, thanks. The rest should all be commits we have already reviewed on master.

> I'm glad that we could keep the proofs about the domain time (with the implication) with minimal change.

:+1:

lsf37:
I'll merge this one somewhat quickly since we can't rebase merge commits.

PR #419, opened 2022-01-12
RISCV MCS Refine

michaelmcinerney:
This contains some work for RISCV MCS Refine. I have completed all files in the Refine session up to and including `KHeap_R.thy`, and it would be good to carefully check these (in particular, `Invariants_H.thy` and `StateRelation.thy` as well as `KHeap_R.thy`). The changes in the files after are just from the patch.

### DIFF
--- a/proof/refine/RISCV64/Bits_R.thy
+++ b/proof/refine/RISCV64/Bits_R.thy
@@ -12,15 +12,18 @@ crunch_ignore (add:
   bind return "when" get gets fail assert put modify unless select alternative assert_opt gets_the
   returnOk throwError lift bindE liftE whenE unlessE throw_opt assertE liftM liftME sequence_x
   zipWithM_x mapM_x sequence mapM sequenceE_x sequenceE mapME mapME_x catch select_f
-  handleE' handleE handle_elseE forM forM_x zipWithM filterM forME_x
+  handleE' handleE handle_elseE forM forM_x forME_x zipWithM filterM forME_x
   withoutFailure throw catchFailure rethrowFailure capFaultOnFailure lookupErrorOnFailure
   nullCapOnFailure nothingOnFailure without_preemption withoutPreemption preemptionPoint
   cap_fault_on_failure lookup_error_on_failure const_on_failure ignore_failure ignoreFailure
   empty_on_failure emptyOnFailure unifyFailure unify_failure throw_on_false
-  storeWordVM loadWord setRegister getRegister getRestartPC debugPrint
-  setNextPC maskInterrupt clearMemory throw_on_false unifyFailure ignoreFailure empty_on_failure
-  emptyOnFailure clearMemoryVM null_cap_on_failure setNextPC getRestartPC assertDerived
-  throw_on_false setObject getObject updateObject loadObject)
+  storeWordVM loadWord setRegister getRegister getRestartPC
+  debugPrint setNextPC maskInterrupt clearMemory throw_on_false
+  unifyFailure ignoreFailure empty_on_failure emptyOnFailure clearMemoryVM null_cap_on_failure
+  setNextPC getRestartPC assertDerived throw_on_false getObject setObject updateObject loadObject
+  ifM andM orM whenM whileM haskell_assert)
+
+
 
 context Arch
 begin
@@ -52,7 +55,9 @@ lemma isCap_simps:
   "isNotificationCap v = (\<exists>v0 v1 v2 v3. v = NotificationCap v0 v1 v2 v3)"
   "isEndpointCap v = (\<exists>v0 v1 v2 v3 v4 v5. v = EndpointCap v0 v1 v2 v3 v4 v5)"
   "isUntypedCap v = (\<exists>d v0 v1 f. v = UntypedCap d v0 v1 f)"
-  "isReplyCap v = (\<exists>v0 v1 v2. v = ReplyCap v0 v1 v2)"
+  "isReplyCap v = (\<exists>v0 v1. v = ReplyCap v0 v1)"
+  "isSchedContextCap v = (\<exists>v0 v1. v = SchedContextCap v0 v1)"
+  "isSchedControlCap v = (v = SchedControlCap)"
   "isIRQControlCap v = (v = IRQControlCap)"
   "isIRQHandlerCap v = (\<exists>v0. v = IRQHandlerCap v0)"
   "isNullCap v = (v = NullCap)"
@@ -85,6 +90,27 @@ lemma projectKO_ntfn:
   "(projectKO_opt ko = Some t) = (ko = KONotification t)"
   by (cases ko) (auto simp: projectKO_opts_defs)
 
+lemma projectKO_reply:
+  "(projectKO_opt ko = Some t) = (ko = KOReply t)"
+  by (cases ko) (auto simp: projectKO_opts_defs)
+
+lemma reply_of'_KOReply[simp]:
+  "reply_of' (KOReply reply) = Some reply"
+  apply (clarsimp simp: projectKO_reply)
+  done
+
+lemma projectKO_sc:
+  "(projectKO_opt ko = Some t) = (ko = KOSchedContext t)"
+  by (cases ko) (auto simp: projectKO_opts_defs)
+
+lemma sc_of'_Sched[simp]:
+  "sc_of' (KOSchedContext sc) = Some sc"
+  by (simp add: projectKO_sc)
+
+lemma tcb_of'_TCB[simp]:
+  "tcb_of' (KOTCB tcb) = Some tcb"
+  by (simp add: projectKO_tcb)
+
 lemma projectKO_ASID:
   "(projectKO_opt ko = Some t) = (ko = KOArch (KOASIDPool t))"
   by (cases ko)
@@ -106,9 +132,9 @@ lemma projectKO_user_data_device:
      (auto simp: projectKO_opts_defs split: arch_kernel_object.splits)
 
 lemmas projectKOs[simp] =
-  projectKO_ntfn projectKO_ep projectKO_cte projectKO_tcb
+  projectKO_ntfn projectKO_ep projectKO_cte projectKO_tcb projectKO_reply projectKO_sc
   projectKO_ASID projectKO_PTE projectKO_user_data projectKO_user_data_device
-  projectKO_eq projectKO_eq2
+  projectKO_eq
 
 lemma capAligned_epI:
   "ep_at' p s \<Longrightarrow> capAligned (EndpointCap p a b c d e)"
@@ -117,7 +143,7 @@ lemma capAligned_epI:
   apply (drule ko_wp_at_norm)
   apply clarsimp
   apply (drule ko_wp_at_aligned)
-  apply (simp add: objBits_simps capUntypedPtr_def isCap_simps objBits_defs)
+  apply (simp add: objBits_simps' capUntypedPtr_def isCap_simps)
   done
 
 lemma capAligned_ntfnI:
@@ -136,20 +162,58 @@ lemma capAligned_tcbI:
                   dest!: ko_wp_at_aligned simp: objBits_simps')
   done
 
-lemma capAligned_reply_tcbI:
-  "tcb_at' p s \<Longrightarrow> capAligned (ReplyCap p m r)"
+lemma capAligned_replyI:
+  "reply_at' p s \<Longrightarrow> capAligned (ReplyCap p r)"
   apply (clarsimp simp: obj_at'_real_def capAligned_def
                         objBits_simps word_bits_def capUntypedPtr_def isCap_simps)
   apply (fastforce dest: ko_wp_at_norm
                   dest!: ko_wp_at_aligned simp: objBits_simps')
   done
 
+lemma capAligned_sched_contextI:
+  "\<lbrakk>sc_at'_n r p s; sc_size_bounds r\<rbrakk>
+      \<Longrightarrow> capAligned (SchedContextCap p r)"
+  by (clarsimp simp: obj_at'_real_def capAligned_def sc_size_bounds_def ko_wp_at'_def isCap_simps
+                     objBits_simps word_bits_def capUntypedPtr_def maxUntypedSizeBits_def)
+
+lemma sc_at'_n_sc_at':
+  "sc_at'_n n p s \<Longrightarrow> sc_at' p s"
+  apply (clarsimp simp: ko_wp_at'_def obj_at'_def projectKOs)
+  by (case_tac ko; clarsimp)
+
 lemma ko_at_valid_objs':
   assumes ko: "ko_at' k p s"
   assumes vo: "valid_objs' s"
   assumes k: "\<And>ko. projectKO_opt ko = Some k \<Longrightarrow> injectKO k = ko"
   shows "valid_obj' (injectKO k) s" using ko vo
-  by (clarsimp simp: valid_objs'_def obj_at'_def project_inject ranI)
+  by (clarsimp simp: valid_objs'_def obj_at'_def projectKOs
+                     project_inject ranI)
+
+lemmas ko_at_valid_objs'_pre =
+  ko_at_valid_objs'[simplified project_inject, atomized, simplified, rule_format]
+
+lemmas ep_ko_at_valid_objs_valid_ep' =
+  ko_at_valid_objs'_pre[where 'a=endpoint, simplified injectKO_defs valid_obj'_def, simplified]
+
+lemmas ntfn_ko_at_valid_objs_valid_ntfn' =
+  ko_at_valid_objs'_pre[where 'a=notification, simplified injectKO_defs valid_obj'_def,
+                        simplified]
+
+lemmas tcb_ko_at_valid_objs_valid_tcb' =
+  ko_at_valid_objs'_pre[where 'a=tcb, simplified injectKO_defs valid_obj'_def, simplified]
+
+lemmas cte_ko_at_valid_objs_valid_cte' =
+  ko_at_valid_objs'_pre[where 'a=cte, simplified injectKO_defs valid_obj'_def, simplified]
+
+lemmas sc_ko_at_valid_objs_valid_sc' =
+  ko_at_valid_objs'_pre[where 'a=sched_context, simplified injectKO_defs valid_obj'_def,
+                        simplified]
+
+lemmas reply_ko_at_valid_objs_valid_reply' =
+  ko_at_valid_objs'_pre[where 'a=reply, simplified injectKO_defs valid_obj'_def, simplified]
+
+lemmas pte_ko_at_valid_objs_valid_pte' =
+  ko_at_valid_objs'_pre[where 'a=pte, simplified injectKO_pte valid_obj'_def]
 
 lemma obj_at_valid_objs':
   "\<lbrakk> obj_at' P p s; valid_objs' s \<rbrakk> \<Longrightarrow>
@@ -184,6 +248,9 @@ lemma getIdleThread_corres:
 
 lemma git_wp [wp]: "\<lbrace>\<lambda>s. P (ksIdleThread s) s\<rbrace> getIdleThread \<lbrace>P\<rbrace>"
   by (unfold getIdleThread_def, wp)
+
+lemma getIdleSc_wp [wp]: "\<lbrace>\<lambda>s. P (ksIdleSC s) s\<rbrace> getIdleSC \<lbrace>P\<rbrace>"
+  by (unfold getIdleSC_def, wp)
 
 lemma gsa_wp [wp]: "\<lbrace>\<lambda>s. P (ksSchedulerAction s) s\<rbrace> getSchedulerAction \<lbrace>P\<rbrace>"
   by (unfold getSchedulerAction_def, wp)
--- a/proof/refine/RISCV64/EmptyFail.thy
+++ b/proof/refine/RISCV64/EmptyFail.thy
@@ -11,39 +11,28 @@ begin
 (* Collect empty_fail lemmas here. naming convention is emtpy_fail_NAME.
    Unless there is a good reason, they should all be [intro!, wp, simp] *)
 
-lemma empty_fail_projectKO [simp, intro!]:
-  "empty_fail (projectKO v)"
-  unfolding empty_fail_def projectKO_def
-  by (simp add: return_def fail_def split: option.splits)
+lemma empty_fail_fun_app[intro!]:
+  "empty_fail (f x) \<Longrightarrow> empty_fail (f $ x)"
+  by simp
 
 lemma empty_fail_alignCheck [intro!, wp, simp]:
   "empty_fail (alignCheck a b)"
-  unfolding alignCheck_def
-  by (simp add: alignError_def)
+  unfolding alignCheck_def by wpsimp
 
 lemma empty_fail_magnitudeCheck [intro!, wp, simp]:
   "empty_fail (magnitudeCheck a b c)"
-  unfolding magnitudeCheck_def
-  by (simp split: option.splits)
+  unfolding magnitudeCheck_def by wpsimp
 
 lemma empty_fail_loadObject_default [intro!, wp, simp]:
-  shows "empty_fail (loadObject_default x b c d)"
-  by (auto simp: loadObject_default_def
-           split: option.splits)
+  shows "empty_fail (gets_the $ loadObject_default x b c d)" by wpsimp
 
 lemma empty_fail_threadGet [intro!, wp, simp]:
   "empty_fail (threadGet f p)"
-  by (simp add: threadGet_def getObject_def split_def)
+  by (wpsimp simp: threadGet_def)
 
 lemma empty_fail_getCTE [intro!, wp, simp]:
   "empty_fail (getCTE slot)"
-  apply (simp add: getCTE_def getObject_def split_def)
-  apply (intro empty_fail_bind, simp_all)
-  apply (simp add: loadObject_cte typeError_def alignCheck_def alignError_def
-                   magnitudeCheck_def
-            split: Structures_H.kernel_object.split)
-  apply (auto split: option.split)
-  done
+  by (wpsimp simp: getCTE_def getObject_def)
 
 lemma empty_fail_updateObject_cte [intro!, wp, simp]:
   "empty_fail (updateObject (v :: cte) ko a b c)"
@@ -69,12 +58,13 @@ lemma empty_fail_getSlotCap [intro!, wp, simp]:
 context begin interpretation Arch . (*FIXME: arch_split*)
 
 lemma empty_fail_getObject:
-  assumes x: "(\<And>b c d. empty_fail (loadObject x b c d::'a :: pspace_storable kernel))"
-  shows "empty_fail (getObject x :: 'a :: pspace_storable kernel)"
-  apply (simp add: getObject_def split_def)
-  apply (safe intro!: empty_fail_bind empty_fail_gets empty_fail_assert_opt)
-  apply (rule x)
+  "empty_fail (getObject x :: 'a :: pspace_storable kernel)"
+  apply (wpsimp simp: getObject_def split_def)
   done
+
+lemma empty_fail_getObject_tcb [intro!, wp, simp]:
+  shows "empty_fail (getObject x :: tcb kernel)"
+  by (wpsimp wp: empty_fail_getObject)
 
 lemma empty_fail_updateTrackedFreeIndex [intro!, wp, simp]:
   shows "empty_fail (updateTrackedFreeIndex p idx)"
@@ -97,7 +87,7 @@ lemma empty_fail_getIRQSlot [intro!, wp, simp]:
 
 lemma empty_fail_getObject_ntfn [intro!, wp, simp]:
   "empty_fail (getObject p :: Structures_H.notification kernel)"
-  by (simp add: empty_fail_getObject)
+  by (wpsimp wp: empty_fail_getObject)
 
 lemma empty_fail_getNotification [intro!, wp, simp]:
   "empty_fail (getNotification ep)"
@@ -110,11 +100,11 @@ lemma empty_fail_lookupIPCBuffer [intro!, wp, simp]:
 
 lemma empty_fail_updateObject_default [intro!, wp, simp]:
   "empty_fail (updateObject_default v ko a b c)"
-  by (simp add: updateObject_default_def typeError_def unless_def split: kernel_object.splits )
+  by (wpsimp simp: updateObject_default_def)
 
 lemma empty_fail_threadSet [intro!, wp, simp]:
   "empty_fail (threadSet f p)"
-  by (simp add: threadSet_def getObject_def setObject_def split_def)
+  by (wpsimp simp: threadSet_def setObject_def)
 
 lemma empty_fail_getThreadState[iff]:
   "empty_fail (getThreadState t)"
--- a/proof/refine/RISCV64/InvariantUpdates_H.thy
+++ b/proof/refine/RISCV64/InvariantUpdates_H.thy
@@ -38,30 +38,13 @@ lemma invs'_machine:
 proof -
   show ?thesis
     apply (cases "ksSchedulerAction s")
-    apply (simp_all add: invs'_def valid_state'_def cur_tcb'_def ct_in_state'_def ct_idle_or_in_cur_domain'_def tcb_in_cur_domain'_def
-                    valid_queues_def valid_queues_no_bitmap_def bitmapQ_defs
-                    vms ct_not_inQ_def
-                    state_refs_of'_def ps_clear_def
-                    valid_irq_node'_def mask
-              cong: option.case_cong)
-    done
-qed
-
-lemma invs_no_cicd'_machine:
-  assumes mask: "irq_masks (f (ksMachineState s)) =
-                 irq_masks (ksMachineState s)"
-  assumes vms: "valid_machine_state' (ksMachineState_update f s) =
-                valid_machine_state' s"
-  shows "invs_no_cicd' (ksMachineState_update f s) = invs_no_cicd' s"
-proof -
-  show ?thesis
-    apply (cases "ksSchedulerAction s")
-    apply (simp_all add: all_invs_but_ct_idle_or_in_cur_domain'_def valid_state'_def cur_tcb'_def ct_in_state'_def ct_idle_or_in_cur_domain'_def tcb_in_cur_domain'_def
-                    valid_queues_def valid_queues_no_bitmap_def bitmapQ_defs
-                    vms ct_not_inQ_def
-                    state_refs_of'_def ps_clear_def
-                    valid_irq_node'_def mask
-              cong: option.case_cong)
+    apply (simp_all add: invs'_def cur_tcb'_def ct_in_state'_def
+                         ct_idle_or_in_cur_domain'_def tcb_in_cur_domain'_def
+                         valid_queues_def valid_queues_no_bitmap_def bitmapQ_defs
+                         vms ct_not_inQ_def valid_dom_schedule'_def
+                         state_refs_of'_def ps_clear_def
+                         valid_irq_node'_def mask
+                   cong: option.case_cong)
     done
 qed
 
@@ -93,10 +76,6 @@ lemma valid_tcb'_tcbQueued[simp]:
 lemma valid_tcb'_tcbFault_update[simp]:
   "valid_tcb' tcb s \<Longrightarrow> valid_tcb' (tcbFault_update f tcb) s"
   by (clarsimp simp: valid_tcb'_def  tcb_cte_cases_def cteSizeBits_def)
-
-lemma valid_tcb'_tcbTimeSlice_update[simp]:
-  "valid_tcb' (tcbTimeSlice_update f tcb) s = valid_tcb' tcb s"
-  by (simp add:valid_tcb'_def tcb_cte_cases_def cteSizeBits_def)
 
 lemma valid_queues_ksSchedulerAction_update[simp]:
   "valid_queues (ksSchedulerAction_update f s) = valid_queues s"
@@ -160,6 +139,10 @@ lemma valid_queues'_ksWorkUnitsCompleted[simp]:
   "valid_queues' (ksWorkUnitsCompleted_update f s) = valid_queues' s"
   by (simp add: valid_queues'_def)
 
+lemma valid_queues'_ksReprogramTimer[simp]:
+  "valid_queues' (ksReprogramTimer_update f s) = valid_queues' s"
+  by (simp add: valid_queues'_def)
+
 lemma valid_queues_ksCurDomain[simp]:
   "valid_queues (ksCurDomain_update f s) = valid_queues s"
   by (simp add: valid_queues_def valid_queues_no_bitmap_def bitmapQ_defs)
@@ -178,6 +161,10 @@ lemma valid_queues_ksDomainTime[simp]:
 
 lemma valid_queues_ksWorkUnitsCompleted[simp]:
   "valid_queues (ksWorkUnitsCompleted_update f s) = valid_queues s"
+  by (simp add: valid_queues_def valid_queues_no_bitmap_def bitmapQ_defs)
+
+lemma valid_queues_ksReprogramTimer[simp]:
+  "valid_queues (ksReprogramTimer_update f s) = valid_queues s"
   by (simp add: valid_queues_def valid_queues_no_bitmap_def bitmapQ_defs)
 
 lemma valid_irq_node'_ksCurDomain[simp]:
@@ -199,6 +186,58 @@ lemma valid_irq_node'_ksDomainTime[simp]:
 lemma valid_irq_node'_ksWorkUnitsCompleted[simp]:
   "valid_irq_node' w (ksWorkUnitsCompleted_update f s) = valid_irq_node' w s"
   by (simp add: valid_irq_node'_def)
+
+lemma valid_irq_node'_ksReprogramTimer[simp]:
+  "valid_irq_node' w (ksReprogramTimer_update f s) = valid_irq_node' w s"
+  by (simp add: valid_irq_node'_def)
+
+lemma valid_release_queue_ksWorkUnitsCompleted[simp]:
+  "valid_release_queue (ksWorkUnitsCompleted_update f s) = valid_release_queue s"
+  by (simp add: valid_release_queue_def)
+
+lemma valid_release_queue_ksReprogramTimer[simp]:
+  "valid_release_queue (ksReprogramTimer_update f s) = valid_release_queue s"
+  by (simp add: valid_release_queue_def)
+
+lemma valid_release_queue_ksDomainTime[simp]:
+  "valid_release_queue (ksDomainTime_update f s) = valid_release_queue s"
+  by (simp add: valid_release_queue_def)
+
+lemma valid_release_queue_ksCurDomain[simp]:
+  "valid_release_queue (ksCurDomain_update f s) = valid_release_queue s"
+  by (simp add: valid_release_queue_def)
+
+lemma valid_release_queue_ksDomScheduleIdx[simp]:
+  "valid_release_queue (ksDomScheduleIdx_update f s) = valid_release_queue s"
+  by (simp add: valid_release_queue_def)
+
+lemma valid_release_queue_ksSchedulerAction[simp]:
+  "valid_release_queue (ksSchedulerAction_update f s) = valid_release_queue s"
+  by (simp add: valid_release_queue_def)
+
+lemma valid_release_queue'_ksSchedulerAction[simp]:
+  "valid_release_queue' (ksSchedulerAction_update f s) = valid_release_queue' s"
+  by (simp add: valid_release_queue'_def)
+
+lemma valid_release_queue'_ksWorkUnitsCompleted[simp]:
+  "valid_release_queue' (ksWorkUnitsCompleted_update f s) = valid_release_queue' s"
+  by (simp add: valid_release_queue'_def)
+
+lemma valid_release_queue'_ksReprogramTimer[simp]:
+  "valid_release_queue' (ksReprogramTimer_update f s) = valid_release_queue' s"
+  by (simp add: valid_release_queue'_def)
+
+lemma valid_release_queue'_ksDomainTime[simp]:
+  "valid_release_queue' (ksDomainTime_update f s) = valid_release_queue' s"
+  by (simp add: valid_release_queue'_def)
+
+lemma valid_release_queue'_ksCurDomain[simp]:
+  "valid_release_queue' (ksCurDomain_update f s) = valid_release_queue' s"
+  by (simp add: valid_release_queue'_def)
+
+lemma valid_release_queue'_ksDomScheduleIdx[simp]:
+  "valid_release_queue' (ksDomScheduleIdx_update f s) = valid_release_queue' s"
+  by (simp add: valid_release_queue'_def)
 
 lemma ex_cte_cap_wp_to_work_units[simp]:
   "ex_cte_cap_wp_to' P slot (ksWorkUnitsCompleted_update f s)
@@ -277,17 +316,22 @@ lemma ct_in_state_ksSched[simp]:
   apply auto
   done
 
-lemma invs'_wu [simp]:
+lemma invs'_wu[simp]:
   "invs' (ksWorkUnitsCompleted_update f s) = invs' s"
-  apply (simp add: invs'_def cur_tcb'_def valid_state'_def Invariants_H.valid_queues_def
-                   valid_queues'_def valid_irq_node'_def valid_machine_state'_def
-                   ct_not_inQ_def ct_idle_or_in_cur_domain'_def tcb_in_cur_domain'_def
-                   bitmapQ_defs valid_queues_no_bitmap_def)
+  apply (simp add: invs'_def cur_tcb'_def valid_queues_def
+                   valid_queues'_def valid_release_queue_def valid_release_queue'_def
+                   valid_irq_node'_def valid_machine_state'_def ct_not_inQ_def
+                   ct_idle_or_in_cur_domain'_def tcb_in_cur_domain'_def bitmapQ_defs
+                   valid_queues_no_bitmap_def valid_dom_schedule'_def)
   done
 
 lemma valid_arch_state'_interrupt[simp]:
   "valid_arch_state' (ksInterruptState_update f s) = valid_arch_state' s"
   by (simp add: valid_arch_state'_def cong: option.case_cong)
+
+lemma valid_inQ_queues_ksSchedulerAction_update[simp]:
+  "valid_inQ_queues (ksSchedulerAction_update f s) = valid_inQ_queues s"
+  by (simp add: valid_inQ_queues_def)
 
 lemma valid_bitmapQ_ksSchedulerAction_upd[simp]:
   "valid_bitmapQ (ksSchedulerAction_update f s) = valid_bitmapQ s"
@@ -334,9 +378,15 @@ lemma sch_act_simple_ksReadyQueuesL2Bitmap[simp]:
 
 lemma ksDomainTime_invs[simp]:
   "invs' (ksDomainTime_update f s) = invs' s"
-  by (simp add:invs'_def valid_state'_def
-    cur_tcb'_def ct_not_inQ_def ct_idle_or_in_cur_domain'_def
-    tcb_in_cur_domain'_def valid_machine_state'_def)
+  by (simp add: invs'_def cur_tcb'_def
+                valid_release_queue_def valid_release_queue'_def
+                valid_machine_state'_def ct_not_inQ_def
+                ct_idle_or_in_cur_domain'_def tcb_in_cur_domain'_def valid_dom_schedule'_def)
+
+lemma ksSchedulerAction_invs'[simp]:
+  "invs' (ksSchedulerAction_update f s) = invs' s"
+  by (auto simp: invs'_def valid_release_queue_def valid_release_queue'_def
+                 valid_machine_state'_def  valid_irq_node'_def valid_dom_schedule'_def)
 
 lemma valid_machine_state'_ksDomainTime[simp]:
   "valid_machine_state' (ksDomainTime_update f s) = valid_machine_state' s"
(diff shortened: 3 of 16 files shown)


proof/refine/ARM/Invariants_H.thy
```diff
@@ -2490,6 +2490,20 @@ lemma cte_wp_at_cases':
                          unless_def when_def bind_def
                   split: kernel_object.splits if_split_asm option.splits
                     del: disjCI)
+apply (rule disjI2)
+
+\<comment> \<open>apply (fastforce simp: field_simps elim: rsubst[where P="\<lambda>x. ksPSpace s x = v" for s v])\<close>
+apply (rule_tac x="0x10" in exI)
+apply (rule_tac x=x6 in exI)
```
lsf37, line 2497: It looks like there is an additional case in MCS in this proof. Can you comment on what that is? Ideally we want to solve that case without an explicit instantiation.

proof/refine/ARM/Invariants_H.thy
```diff
@@ -2742,7 +2756,13 @@ lemma typ_at_lift_page_table_at'_strong:
...
   apply -
   apply (rule P_bool_lift[where P=P])
-  apply (wpsimp wp: hoare_vcg_const_Ball_lift hoare_vcg_bex_lift hoare_vcg_imp_lift
+  apply_trace (wpsimp wp: hoare_vcg_const_Ball_lift hoare_vcg_bex_lift hoare_vcg_imp_lift
+         | fastforce)
+  apply_trace (wpsimp wp: hoare_vcg_const_Ball_lift hoare_vcg_bex_lift hoare_vcg_imp_lift
+         | fastforce)
+  apply_trace (wpsimp wp: hoare_vcg_const_Ball_lift hoare_vcg_bex_lift hoare_vcg_imp_lift
+         | fastforce)
+  apply_trace (wpsimp wp: hoare_vcg_const_Ball_lift hoare_vcg_bex_lift hoare_vcg_imp_lift
```
lsf37, line 2765: (in `typ_at_lift_page_table_at'_strong`) I would not have expected this proof to change at all. Can you say what breaks in here for MCS?

lsf37: Oh, wait, this is in `ARM/Invariants_H.thy`! Sorry, should have looked closer, I think this is just diagnostics.

proof/refine/RISCV64/Invariants_H.thy
```diff
@@ -51,66 +71,82 @@ definition pspace_no_overlap' :: "obj_ref \<Rightarrow> nat \<Rightarrow> kernel
      \<lambda>s. \<forall>x ko. ksPSpace s x = Some ko \<longrightarrow>
                 (mask_range x (objBitsKO ko)) \<inter> {ptr .. (ptr && ~~ mask bits) + mask bits} = {}"
 
-definition ko_wp_at' :: "(kernel_object \<Rightarrow> bool) \<Rightarrow> obj_ref \<Rightarrow> kernel_state \<Rightarrow> bool" where
+definition
```
lsf37: This might have been part of the automatic patch, but we shouldn't remove the type signature here. I've explicitly added a lot of these in the Invariants file to improve style and maintainability. Would be sad to lose all that in the merge.

proof/refine/RISCV64/InvariantUpdates_H.thy
```diff
@@ -200,6 +187,58 @@ lemma valid_irq_node'_ksWorkUnitsCompleted[simp]:
...
+  "valid_release_queue' (ksDomainTime_update f s) = valid_release_queue' s"
+  by (simp add: valid_release_queue'_def)
+
+lemma valid_release_queue'_ksCurDomain[simp]:
+  "valid_release_queue' (ksCurDomain_update f s) = valid_release_queue' s"
+  by (simp add: valid_release_queue'_def)
+
+lemma valid_release_queue'_ksDomScheduleIdx[simp]:
+  "valid_release_queue' (ksDomScheduleIdx_update f s) = valid_release_queue' s"
+  by (simp add: valid_release_queue'_def)
```
Xaphiosis, line 240: @corlewis : could crunch or something like it automate some of this? I also saw some `empty_fail` that  looked derivable, unless I'm mistaken (happens a lot) crunch has a mode that can deal with those?

proof/refine/RISCV64/Invariants_H.thy
```diff
@@ -51,66 +71,82 @@ definition pspace_no_overlap' :: "obj_ref \<Rightarrow> nat \<Rightarrow> kernel
...
+                     tcbBlockSizeBits_def epSizeBits_def ntfnSizeBits_def cteSizeBits_def word_size
+                     pteBits_def wordSizeCase_def wordBits_def replySizeBits_def
+              split: arch_kernel_object.splits)
 
-definition obj_at' :: "('a::pspace_storable \<Rightarrow> bool) \<Rightarrow> machine_word \<Rightarrow> kernel_state \<Rightarrow> bool" where
-  obj_at'_real_def:
-  "obj_at' P p s \<equiv> ko_wp_at' (\<lambda>ko. \<exists>obj. projectKO_opt ko = Some obj \<and> P obj) p s"
+definition
+  obj_at' :: "('a::pspace_storable \<Rightarrow> bool) \<Rightarrow> obj_ref \<Rightarrow> kernel_state \<Rightarrow> bool"
+where obj_at'_real_def:
```
lsf37: Probably also part of the automated patch bit, but please restore the original correct style (name on same line as `definition` and `where` on same line as type when it fits (and all on same line if possible))

lsf37: In fact, is there any actual change in this one? It's hard to spot.

proof/refine/RISCV64/KHeap_R.thy
```diff
@@ -207,64 +373,277 @@ lemma obj_at_setObject2:
...
+  using assms
+  apply (clarsimp simp: objBits_def objBitsKO_def inj_def
+                        same_size_obj_at'_set_obj'_iff[where 'a=tcb, simplified])
+  done
+
+\<comment>\<open>
+  Keeps a generic @{term obj_at'} (rather than a specific @{term "obj_at' (\<lambda>_. True)"}) to match
+  in more simp contexts.
+\<close>
+lemma tcb_obj_at'_set_obj'_iff:
```
Xaphiosis: who makes these strange multi-line comment style?

michaelmcinerney:
Now with some style updates.

The second commit contains all the files in Refine I have gone through so far. (The last commit contains the changes from the patch, as well as the original files and the `.rej` files, and can be ignored for now)

proof/refine/RISCV64/EmptyFail.thy
```diff
@@ -12,38 +12,26 @@ begin
    Unless there is a good reason, they should all be [intro!, wp, simp] *)
 
 lemma empty_fail_projectKO [simp, intro!]:
-  "empty_fail (projectKO v)"
-  unfolding empty_fail_def projectKO_def
-  by (simp add: return_def fail_def split: option.splits)
+  "empty_fail (gets_the $ projectKO v)" by wpsimp
```
lsf37, line 15: This looks unlikely to apply with `simp`, because the `$` would usually already be gone when this rule is used. I think a general lemma `empty_fail (gets_the f)` (for any f) would be better here.

lsf37: Hm, but that would not actually solve a `empty_fail (gets_the $ _)` goal. So maybe two rules, the one I suggested above with `[simp, intro!]`, and `empty_fail (gets_the $ _)` with just `[intro!]`.

proof/refine/RISCV64/Bits_R.thy
```diff
@@ -105,10 +131,10 @@ lemma projectKO_user_data_device:
   by (cases ko)
      (auto simp: projectKO_opts_defs split: arch_kernel_object.splits)
 
-lemmas projectKOs[simp] =
-  projectKO_ntfn projectKO_ep projectKO_cte projectKO_tcb
+lemmas projectKOs =
```
lsf37, line 134: I made `projectKOs` `[simp]` in RISCV64 on purpose, because you pretty much always want them when a `projectKO` occurs. So I'd vote for putting the `[simp]` back in (and removing the explicit `simp add: projectKOs` where they occur in new proofs -- the old ones in RISCV should have any)

proof/refine/RISCV64/Bits_R.thy
```diff
@@ -117,39 +143,77 @@ lemma capAligned_epI:
...
+  apply (simp add: objBits_simps' projectKOs capUntypedPtr_def isCap_simps)
   done
 
 lemma capAligned_ntfnI:
   "ntfn_at' p s \<Longrightarrow> capAligned (NotificationCap p a b c)"
   apply (clarsimp simp: obj_at'_real_def capAligned_def
                         objBits_simps word_bits_def capUntypedPtr_def isCap_simps)
   apply (fastforce dest: ko_wp_at_norm
-                  dest!: ko_wp_at_aligned simp: objBits_simps')
+                  dest!: ko_wp_at_aligned simp: objBits_simps' projectKOs)
```
lsf37, line 154: could be reverted

proof/refine/RISCV64/Bits_R.thy
```diff
@@ -117,39 +143,77 @@ lemma capAligned_epI:
   apply (drule ko_wp_at_norm)
   apply clarsimp
   apply (drule ko_wp_at_aligned)
-  apply (simp add: objBits_simps capUntypedPtr_def isCap_simps objBits_defs)
+  apply (simp add: objBits_simps' projectKOs capUntypedPtr_def isCap_simps)
```
lsf37, line 146: `projectKOs` could be removed

proof/refine/RISCV64/Bits_R.thy
```diff
@@ -117,39 +143,77 @@ lemma capAligned_epI:
...
+                  dest!: ko_wp_at_aligned simp: objBits_simps' projectKOs)
   done
 
 lemma capAligned_tcbI:
   "tcb_at' p s \<Longrightarrow> capAligned (ThreadCap p)"
   apply (clarsimp simp: obj_at'_real_def capAligned_def
                         objBits_simps word_bits_def capUntypedPtr_def isCap_simps)
   apply (fastforce dest: ko_wp_at_norm
-                  dest!: ko_wp_at_aligned simp: objBits_simps')
+                  dest!: ko_wp_at_aligned simp: objBits_simps' projectKOs)
```
lsf37, line 162: could be reverted

proof/refine/RISCV64/Bits_R.thy
```diff
@@ -117,39 +143,77 @@ lemma capAligned_epI:
...
 
-lemma capAligned_reply_tcbI:
-  "tcb_at' p s \<Longrightarrow> capAligned (ReplyCap p m r)"
+lemma capAligned_replyI:
+  "reply_at' p s \<Longrightarrow> capAligned (ReplyCap p r)"
   apply (clarsimp simp: obj_at'_real_def capAligned_def
                         objBits_simps word_bits_def capUntypedPtr_def isCap_simps)
   apply (fastforce dest: ko_wp_at_norm
-                  dest!: ko_wp_at_aligned simp: objBits_simps')
+                  dest!: ko_wp_at_aligned simp: objBits_simps' projectKOs)
```
lsf37, line 170: could be reverted

proof/refine/RISCV64/Bits_R.thy
```diff
@@ -117,39 +143,77 @@ lemma capAligned_epI:
...
+  apply (clarsimp simp: ko_wp_at'_def obj_at'_def projectKOs)
+  by (case_tac ko; clarsimp)
+
 lemma ko_at_valid_objs':
   assumes ko: "ko_at' k p s"
   assumes vo: "valid_objs' s"
   assumes k: "\<And>ko. projectKO_opt ko = Some k \<Longrightarrow> injectKO k = ko"
   shows "valid_obj' (injectKO k) s" using ko vo
-  by (clarsimp simp: valid_objs'_def obj_at'_def project_inject ranI)
+  by (clarsimp simp: valid_objs'_def obj_at'_def projectKOs
```
lsf37, line 189: `projectKOs` could be removed

proof/refine/RISCV64/Bits_R.thy
```diff
@@ -169,6 +233,7 @@ lemma tcb_in_valid_state':
   apply (clarsimp simp: pred_tcb_at'_def)
   apply (drule obj_at_valid_objs')
    apply fastforce
+  apply (clarsimp simp: projectKOs)
```
lsf37, line 236: could be reverted

proof/refine/RISCV64/Bits_R.thy
```diff
@@ -329,8 +397,9 @@ lemma ko_at_imp_cte_wp_at':
   fixes x :: cte
   shows "\<lbrakk> ko_at' x ptr s \<rbrakk> \<Longrightarrow> cte_wp_at' (\<lambda>cte. cte = x) ptr s"
   apply (erule obj_atE')
-  apply (clarsimp simp: objBits_simps')
-  apply (erule cte_wp_at_cteI'; simp add: cte_level_bits_def)
+  apply (clarsimp simp: projectKOs objBits_simps')
+  apply (erule cte_wp_at_cteI')
+    apply (simp add: cte_level_bits_def)+
```
lsf37, line 402: could be reverted

proof/refine/RISCV64/Bits_R.thy
```diff
@@ -429,9 +498,9 @@ lemma withoutPreemption_R:
 
 lemma ko_at_cte_ipcbuffer:
   "ko_at' tcb p s \<Longrightarrow> cte_wp_at' (\<lambda>x. x = tcbIPCBufferFrame tcb) (p + tcbIPCBufferSlot * 0x20) s"
-  apply (clarsimp simp: obj_at'_def objBits_simps)
+  apply (clarsimp simp: obj_at'_def projectKOs objBits_simps)
```
lsf37, line 501: could be reverted

proof/refine/RISCV64/Bits_R.thy
```diff
@@ -429,9 +498,9 @@ lemma withoutPreemption_R:
...
   apply (erule (2) cte_wp_at_tcbI')
-   apply (fastforce simp add: tcb_cte_cases_def tcbIPCBufferSlot_def cteSizeBits_def)
+   apply (fastforce simp add: tcb_cte_cases_def cteSizeBits_def tcbIPCBufferSlot_def)
```
lsf37, line 503: could be reverted (to keep the final diff small)

proof/refine/RISCV64/EmptyFail.thy
```diff
@@ -12,38 +12,26 @@ begin
...
 lemma empty_fail_getCTE [intro!, wp, simp]:
   "empty_fail (getCTE slot)"
-  apply (simp add: getCTE_def getObject_def split_def)
-  apply (intro empty_fail_bind, simp_all)
-  apply (simp add: loadObject_cte typeError_def alignCheck_def alignError_def
-                   magnitudeCheck_def
-            split: Structures_H.kernel_object.split)
-  apply (auto split: option.split)
-  done
+  by (wpsimp simp: getCTE_def getObject_def)
```
lsf37, line 34: nice!

proof/refine/RISCV64/EmptyFail.thy
lsf37: We should do this properly: if `$` does occur in terms with `empty_fail`, we should have a rule for `$`, then we never need any combination rules. So we should have `empty_fail (f x) ==> empty_fail (f $ x)` as `[intro!]`, and then no other rules with `$` in that position, i.e. only `empty_fail (gets_the f)` as `[simp, intro!]`.

proof/refine/RISCV64/InvariantUpdates_H.thy
lsf37: This does smell automatable. If not with crunch, then probably with an update locale like we have for some of the other updators. Or potentially the `add_upd_simps` command which helpfully has no documentation at all, but sounds like it should be useful here.

proof/refine/RISCV64/Machine_R.thy
```diff
@@ -22,6 +22,31 @@ lemma irq_state_independent_HI[intro!, simp]:
    \<Longrightarrow> irq_state_independent_H P"
   by (simp add: irq_state_independent_H_def)
 
+definition "getCurrentTime_independent_H (P :: kernel_state \<Rightarrow> bool)
+  \<equiv> \<forall>f s. P s \<longrightarrow>
```
lsf37, line 26: This is not style conforming. The `definition` line should have a type signature with `where` and the actual definition should start on a separate line.

proof/refine/RISCV64/Machine_R.thy
```diff
@@ -22,6 +22,31 @@ lemma irq_state_independent_HI[intro!, simp]:
...
+   "\<lbrakk>\<And>s f.
+     P (s\<lparr>ksMachineState
+          := (ksMachineState s)\<lparr>last_machine_time :=
+                                f (last_machine_time (ksMachineState s)) (time_state (ksMachineState s))\<rparr>\<rparr>)
+     = P s\<rbrakk>
+    \<Longrightarrow> getCurrentTime_independent_H P"
+   by (simp add: getCurrentTime_independent_H_def)
+
+definition "time_state_independent_H (P :: kernel_state \<Rightarrow> bool)
+  \<equiv> \<forall>f s. P s \<longrightarrow>
```
lsf37, line 41: same style violation as above

proof/refine/RISCV64/Machine_R.thy
```diff
@@ -22,6 +22,31 @@ lemma irq_state_independent_HI[intro!, simp]:
...
 
+definition "getCurrentTime_independent_H (P :: kernel_state \<Rightarrow> bool)
+  \<equiv> \<forall>f s. P s \<longrightarrow>
+           P (s\<lparr>ksMachineState :=
+                       ksMachineState s\<lparr>last_machine_time :=
+                          f (last_machine_time (ksMachineState s)) (time_state (ksMachineState s))\<rparr>\<rparr>)"
+
+lemma getCurrentTime_independent_HI[intro!, simp]:
+   "\<lbrakk>\<And>s f.
+     P (s\<lparr>ksMachineState
```
lsf37, line 33: Put `P ..` on the same line as `\<And>`

proof/refine/RISCV64/Invariants_H.thy
```diff
@@ -7,9 +7,15 @@
...
   LevityCatch
-  "AInvs.ArchDetSchedSchedule_AI"
+  "AInvs.AInvs"
+  "Lib.AddUpdSimps"
+  ArchMove_R
 begin
 
+(* FIXME: this should go somewhere in spec *)
+translations
+  (type) "'a kernel" <=(type) "kernel_state \<Rightarrow> ('a \<times> kernel_state) set \<times> bool"
```
lsf37: Indeed it should :-)

proof/refine/RISCV64/Invariants_H.thy
```diff
@@ -130,29 +167,151 @@ lemma st_tcb_at'_def:
...
+  "getReplyNextPtr (Some (Next rn)) = Some rn" by (simp add: getReplyNextPtr_def)
+
+lemma getHeadScPtr_Some_Head[simp]:
+  "getHeadScPtr (Some (Head sc)) = Some sc" by (simp add: getHeadScPtr_def)
+
+lemma theReplyNextPtr_Some_Next[simp]:
+  "theReplyNextPtr (Some (Next rn)) = rn" by (simp add: theReplyNextPtr_def)
+
+lemma theHeadScPtr_Some_Head[simp]:
+  "theHeadScPtr (Some (Head sc)) = sc" by (simp add: theHeadScPtr_def)
```
lsf37: these should be replaced with lemmas of the form

`lemmas getReplyNextPr_simps[simp] = getReplyNextPtr_def[split_simps option.split reply_next.split]`

(This will cover all cases and make it resistant to change in constructor names and pattern)

lsf37: I'm pretty sure I said the same on the original PR where this was introduced, but apparently that has not made it in.

proof/refine/RISCV64/Invariants_H.thy
```diff
@@ -130,29 +167,151 @@ lemma st_tcb_at'_def:
...
-   | (BlockedOnSend x _ _ _ _)  => {(x, TCBBlockedSend)}
-   | (BlockedOnNotification x)  => {(x, TCBSignal)}
-   | _                          => {}"
+type_synonym ref_set = "(obj_ref \<times> reftype) set"
+
+definition tcb_st_refs_of' :: "thread_state \<Rightarrow> ref_set" where
+  "tcb_st_refs_of' z \<equiv> case z of
+    Running                    => {}
+  | Inactive                   => {}
+  | Restart                    => {}
```
lsf37: should use a default pattern `_ => {}` for these cases, so the the definition is resistant to change in that datatype

proof/refine/RISCV64/Invariants_H.thy
```diff
@@ -130,29 +167,151 @@ lemma st_tcb_at'_def:
...
+  | (BlockedOnReply r)         => if bound r then {(the r, TCBReply)} else {}
+  | IdleThreadState            => {}
+  | (BlockedOnReceive x _ r)   => if bound r then {(x, TCBBlockedRecv), (the r, TCBReply)}
+                                  else {(x, TCBBlockedRecv)}
+  | (BlockedOnSend x _ _ _ _)  => {(x, TCBBlockedSend)}
+  | (BlockedOnNotification x)  => {(x, TCBSignal)}"
+
+definition
+  tcb_bound_refs' ::
```
lsf37: name should go on same lines as `definition`

proof/refine/RISCV64/Invariants_H.thy
```diff
@@ -169,50 +328,117 @@ definition ntfn_q_refs_of' :: "Structures_H.ntfn \<Rightarrow> (obj_ref \<times>
...
+
+definition list_refs_of_reply' :: "reply \<Rightarrow> ref_set" where
+  "list_refs_of_reply' r = get_refs ReplyNext (replyNext_of r) \<union> get_refs ReplyPrev (replyPrev r)"
+
+abbreviation list_refs_of_replies_opt' :: "kernel_state \<Rightarrow> obj_ref \<Rightarrow> ref_set option" where
+  "list_refs_of_replies_opt' s \<equiv> replies_of' s ||> list_refs_of_reply'"
+
+(* FIXME RT: move to Option_Monad? *)
+definition map_set :: "('a \<Rightarrow> 'b set option) \<Rightarrow> 'a \<Rightarrow> 'b set" where
+  "map_set f \<equiv> case_option {} id \<circ> f"
```
lsf37: It's probably a good time to do this move now before we copy/paste it to more architectures in the future.

proof/refine/RISCV64/Invariants_H.thy
```diff
@@ -169,50 +328,117 @@ definition ntfn_q_refs_of' :: "Structures_H.ntfn \<Rightarrow> (obj_ref \<times>
...
-definition state_refs_of' :: "kernel_state \<Rightarrow> obj_ref \<Rightarrow> (obj_ref \<times> reftype) set" where
-  "state_refs_of' s \<equiv> \<lambda>x.
-     case ksPSpace s x of
-       None \<Rightarrow> {}
-     | Some ko \<Rightarrow> if is_aligned x (objBitsKO ko) \<and> ps_clear x (objBitsKO ko) s
-                  then refs_of' ko else {}"
+definition refs_of' :: "kernel_object \<Rightarrow> ref_set" where
+  "refs_of' x \<equiv> case x of
+    (KOTCB tcb)           => refs_of_tcb' tcb
+  | (KOCTE cte)           => {}
```
lsf37: This line can be removed (covered by default case)

proof/refine/RISCV64/Invariants_H.thy
```diff
@@ -169,50 +328,117 @@ definition ntfn_q_refs_of' :: "Structures_H.ntfn \<Rightarrow> (obj_ref \<times>
...
 | "live' (KOEndpoint ep)       = (ep \<noteq> IdleEP)"
-| "live' (KONotification ntfn) = (bound (ntfnBoundTCB ntfn) \<or> (\<exists>ts. ntfnObj ntfn = WaitingNtfn ts))"
-| "live' _                     = False"
+| "live' (KONotification ntfn) = live_ntfn' ntfn"
+| "live' (KOSchedContext sc)   = live_sc' sc"
+| "live' (KOReply r)           = live_reply' r"
+| "live' (KOUserData)          = False"
+| "live' (KOUserDataDevice)    = False"
+| "live' (KOKernelData)        = False"
+| "live' (KOArch ako)          = False"
```
lsf37: use `_ => False` case like in the RISCV version

proof/refine/RISCV64/Invariants_H.thy
```diff
@@ -169,50 +328,117 @@ definition ntfn_q_refs_of' :: "Structures_H.ntfn \<Rightarrow> (obj_ref \<times>
...
 
 fun zobj_refs' :: "capability \<Rightarrow> obj_ref set" where
   "zobj_refs' (EndpointCap r _ _ _ _ _)  = {r}"
 | "zobj_refs' (NotificationCap r _ _ _)  = {r}"
 | "zobj_refs' (ThreadCap r)              = {r}"
-| "zobj_refs' _                          = {}"
+| "zobj_refs' (SchedContextCap r _)      = {r}"
+| "zobj_refs' (ReplyCap r _)             = {r}"
+| "zobj_refs'                 _          = {}"
```
lsf37: Move `_` left, i.e. `zobj_refs' _`

proof/refine/RISCV64/Invariants_H.thy
```diff
@@ -169,50 +328,117 @@ definition ntfn_q_refs_of' :: "Structures_H.ntfn \<Rightarrow> (obj_ref \<times>
...
-| "cte_refs' (Zombie ref _ n) x        = (\<lambda>x. ref + (x << cteSizeBits)) ` {0 ..< of_nat n}"
-| "cte_refs' (IRQHandlerCap irq) x     = {x + (ucast irq << cteSizeBits)}"
-| "cte_refs' _ _                       = {}"
-
+primrec cte_refs' :: "capability \<Rightarrow> obj_ref \<Rightarrow> obj_ref set" where
+  "cte_refs' (UntypedCap _ _ _ _) _      = {}"
+| "cte_refs' NullCap _                   = {}"
+| "cte_refs' DomainCap _                 = {}"
+| "cte_refs' (EndpointCap _ _ _ _ _ _) _ = {}"
+| "cte_refs' (NotificationCap _ _ _ _) _ = {}"
```
lsf37: use `fun` and default case (I don't think `primrec` supports the default case syntax, otherwise primrec would be fine)

proof/refine/RISCV64/Invariants_H.thy
```diff
@@ -169,50 +328,117 @@ definition ntfn_q_refs_of' :: "Structures_H.ntfn \<Rightarrow> (obj_ref \<times>
...
+| "cte_refs' (CNodeCap ref bits _ _) x   = (\<lambda>x. ref + (x * 2^cteSizeBits)) ` {0 .. 2 ^ bits - 1}"
+| "cte_refs' (ThreadCap ref) x           = (\<lambda>x. ref + x) ` (dom tcb_cte_cases)"
+| "cte_refs' (Zombie r _ n) x            = (\<lambda>x. r + (x * 2 ^ cteSizeBits)) ` {0 ..< of_nat n}"
+| "cte_refs' (ArchObjectCap _) _         = {}"
+| "cte_refs' (IRQControlCap) _           = {}"
+| "cte_refs' (IRQHandlerCap irq) x       = {x + (ucast irq) * 16}"
+| "cte_refs' (ReplyCap _ _) _            = {}"
+| "cte_refs' (SchedContextCap _ _) _     = {}"
+| "cte_refs' SchedControlCap _           = {}"
```
lsf37: In fact, we should revert the entire change for this one, the RISCV version looks correct, it uses symbolic names and `mask` etc instead of magic numbers, which we will need if we want to base aarch64 etc on it

proof/refine/RISCV64/Invariants_H.thy
```diff
@@ -402,28 +652,51 @@ definition valid_ep' :: "Structures_H.endpoint \<Rightarrow> kernel_state \<Righ
...
-     \<and> (case ntfnBoundTCB ntfn of Some tcb \<Rightarrow> ts = [tcb] | _ \<Rightarrow> True))
-  | ActiveNtfn b \<Rightarrow> True)
-  \<and> valid_bound_tcb' (ntfnBoundTCB ntfn) s"
+    Structures_H.IdleNtfn \<Rightarrow> True
+  | Structures_H.WaitingNtfn ts \<Rightarrow>
+      ts \<noteq> [] \<and> (\<forall>t \<in> set ts. tcb_at' t s) \<and> distinct ts
+      \<and> (case ntfnBoundTCB ntfn of Some tcb \<Rightarrow> ts = [tcb] | _ \<Rightarrow> True)
+  | Structures_H.ActiveNtfn b \<Rightarrow> True)
+      \<and> valid_bound_tcb' (ntfnBoundTCB ntfn) s
+      \<and> valid_bound_sc' (ntfnSc ntfn) s"
```
lsf37: Please check if we can remove the `Structures_H` qualifier here. The Haskell version should be the default name space in Refine. With that, the diff for this one should be much smaller.

proof/refine/RISCV64/Invariants_H.thy
```diff
@@ -710,14 +993,34 @@ where
...
   "no_0_obj' \<equiv> \<lambda>s. ksPSpace s 0 = None"
 
+abbreviation is_reply_linked :: "obj_ref \<Rightarrow> kernel_state \<Rightarrow> bool" where
+  "is_reply_linked rptr s \<equiv> replyNexts_of s rptr \<noteq> None \<or> replyPrevs_of s rptr \<noteq> None"
+
+definition valid_replies'_except :: "obj_ref set \<Rightarrow> kernel_state \<Rightarrow> bool" where
+  "valid_replies'_except RS s \<equiv>
+     (\<forall>rptr. rptr \<notin> RS \<and> is_reply_linked rptr s
+             \<longrightarrow> (\<exists>tptr. replyTCBs_of s rptr = Some tptr
+                         \<and> st_tcb_at' ((=) (BlockedOnReply (Some rptr))) tptr s))"
```
lsf37: remove outer parentheses (not necessary for `\<equiv>`)

proof/refine/RISCV64/Invariants_H.thy
```diff
@@ -728,8 +1031,8 @@ where
 | "runnable' (Structures_H.Inactive)                = False"
 | "runnable' (Structures_H.Restart)                 = True"
 | "runnable' (Structures_H.IdleThreadState)         = False"
-| "runnable' (Structures_H.BlockedOnReceive a b)    = False"
-| "runnable' (Structures_H.BlockedOnReply)          = False"
+| "runnable' (Structures_H.BlockedOnReceive _ _ _)  = False"
+| "runnable' (Structures_H.BlockedOnReply _)        = False"
```
lsf37: :+1:

Looks like I didn't clean this one up in my pass back then (probably hadn't touched this part). If you feel like it, you could remove the `Structures_H` and make this a `fun` with default case `False`. This could just be written as `runnable' st == st = Running \/ st = Restart`, but I have the feeling we wanted the `simp` rules we get from primrec/fun

proof/refine/RISCV64/Invariants_H.thy
```diff
@@ -751,9 +1054,14 @@ definition
   valid_queues_no_bitmap :: "kernel_state \<Rightarrow> bool"
 where
  "valid_queues_no_bitmap \<equiv> \<lambda>s.
-   (\<forall>d p. (\<forall>t \<in> set (ksReadyQueues s (d, p)). obj_at' (inQ d p and runnable' \<circ> tcbState) t s)
-    \<and>  distinct (ksReadyQueues s (d, p))
-    \<and> (d > maxDomain \<or> p > maxPriority \<longrightarrow> ksReadyQueues s (d,p) = []))"
+   (\<forall>d p. (\<forall>t \<in> set (ksReadyQueues s (d, p)). obj_at' (inQ d p) t s)
+          \<and> (d > maxDomain \<or> p > maxPriority \<longrightarrow> ksReadyQueues s (d,p) = []))"
+
```
lsf37: should remove out parentheses (they were there in the original, but if we touch it, we should clean it up)

proof/refine/RISCV64/Invariants_H.thy
```diff
@@ -751,9 +1054,14 @@ definition
...
+defs ready_qs_runnable_def:
+  "ready_qs_runnable \<equiv> \<lambda>s. \<forall>d p. \<forall>t \<in> set (ksReadyQueues s (d, p)). st_tcb_at' runnable' t s"
+
+definition notQueued :: "obj_ref \<Rightarrow> kernel_state \<Rightarrow> bool" where
+  "notQueued t s \<equiv> (\<forall>a b. t \<notin> set (ksReadyQueues s (a, b)))"
```
lsf37: remove outer parentheses

lsf37: In fact, this should not have `a b ` and then pass `(a,b)` to ksReadyQueues. If we really want two components here, it should be `d` and `p`, and I would usually recommend just a single variable for the entire pair. Let's not do the latter in this case, because the proofs are going to assume two now and if we fix this here, we're going to have to fix it in the proofs, which we should only do once we have them working.

proof/refine/RISCV64/Invariants_H.thy
```diff
@@ -799,11 +1107,29 @@ where
 lemmas bitmapQ_defs = valid_bitmapQ_def valid_bitmapQ_except_def bitmapQ_def
                        bitmapQ_no_L2_orphans_def bitmapQ_no_L1_orphans_def
 
+(* valid_queues is too strong sometimes *)
+definition valid_inQ_queues :: "kernel_state \<Rightarrow> bool" where
+  "valid_inQ_queues \<equiv> \<lambda>s. \<forall>d p. (\<forall>t\<in>set (ksReadyQueues s (d, p)). obj_at' (inQ d p) t s)"
```
lsf37: I don't think the parentheses around the inner forall are needed.

proof/refine/RISCV64/Invariants_H.thy
```diff
@@ -799,11 +1107,29 @@ where
...
+
+(* when in the middle of updates, a particular queue might not be entirely valid *)
+definition valid_inQ_queues_except :: "obj_ref \<Rightarrow> kernel_state \<Rightarrow> bool" where
+ "valid_inQ_queues_except t'
+   \<equiv> \<lambda>s. (\<forall>d p. (\<forall>t \<in> set (ksReadyQueues s (d, p)). t \<noteq> t' \<longrightarrow> obj_at' (inQ d p) t s))"
```
lsf37: remove parentheses for outer and inner forall

proof/refine/RISCV64/Invariants_H.thy
```diff
@@ -842,19 +1178,42 @@ definition sch_act_sane :: "kernel_state \<Rightarrow> bool" where
 abbreviation
   "sch_act_not t \<equiv> \<lambda>s. ksSchedulerAction s \<noteq> SwitchToThread t"
 
-definition idle_tcb'_2 :: "Structures_H.thread_state \<times> machine_word option \<Rightarrow> bool" where
-  "idle_tcb'_2 \<equiv> \<lambda>(st, ntfn_opt). (idle' st \<and> ntfn_opt = None)"
+definition
+  idle_tcb'_2 :: "Structures_H.thread_state \<times> obj_ref option \<times> obj_ref option \<times> obj_ref option
```
lsf37: name should go on same line as `definition`, type can be on next line if it is too long.

(the idea is that you can grep for "definition idle_tbc" and find the def)

proof/refine/RISCV64/Invariants_H.thy
```diff
@@ -2102,52 +2509,192 @@ proof -
...
-  apply (wp hoare_vcg_disj_lift hoare_vcg_ex_lift x)
+  apply (rule P_bool_lift[where P=P])
+   apply (wpsimp wp: hoare_vcg_disj_lift hoare_vcg_ex_lift hoare_vcg_all_lift x)+
+  done
+
+lemma typ_at_lift_page_table_at'_strong:
+  assumes x: "\<And>p. f \<lbrace>\<lambda>s. P (typ_at' (ArchT PTET) p s)\<rbrace>"
+  shows      "f \<lbrace>\<lambda>s. P (page_table_at' p s)\<rbrace>"
+  unfolding page_table_at'_def All_less_Ball
+  using x
```
lsf37: This pattern of `assumes x`, .., `using x` seems to be recurring. If we are dealing with a single `assumes`, we should leave it unnamed and instead say `using assms`

lsf37: It looks to me like in this case, we don't actually need `using assms`, but could add `assms` as a parameter to the `wp` set in `wpsimp`. (Might not work, but would be worth a shot)

proof/refine/RISCV64/Invariants_H.thy
```diff
@@ -2676,19 +3312,6 @@ lemma ex_cte_cap_to'_pres:
...
-  To relate the two, we must look at the values of kernel configuration constants.
-  To allow the proofs to work for all permitted values of these constants, their definitions
-  should only be unfolded in this section, and the derived properties kept to a minimum.\<close>
-
-lemma le_maxDomain_eq_less_numDomains:
-  shows "x \<le> unat maxDomain \<longleftrightarrow> x < Kernel_Config.numDomains"
-        "y \<le> maxDomain \<longleftrightarrow> unat y < Kernel_Config.numDomains"
-  by (auto simp: Kernel_Config.numDomains_def maxDomain_def word_le_nat_alt)
-
-
```
lsf37: This one should stay

proof/refine/RISCV64/Invariants_H.thy
```diff
@@ -3246,16 +3894,180 @@ private lemma ko_at_defn_ko_wp_at':
...
+  unfolding valid_replies'_sc_asrt_def
+  by (simp)
+
+(** sc_with_reply' **)
+
+definition sc_with_reply' where
+  "sc_with_reply' rp s' =
+     the_pred_option
+         (\<lambda>sc_ptr. \<exists>xs. heap_ls (replyPrevs_of s') (scReplies_of s' sc_ptr) xs
+                     \<and> rp \<in> set xs)"
```
lsf37: style: should have a type signature, should not break after `the_pred_option`

proof/refine/RISCV64/StateRelation.thy
```diff
@@ -188,6 +318,22 @@ primrec aobj_relation_cuts :: "RISCV64_A.arch_kernel_obj \<Rightarrow> machine_w
 | "aobj_relation_cuts (PageTable pt) x =
      (\<lambda>y. (x + (ucast y << pteBits), pte_relation y)) ` UNIV"
 
+abbreviation
+  sc_relation_cut :: "Structures_A.kernel_object \<Rightarrow> Structures_H.kernel_object \<Rightarrow> bool"
+  where
+  "sc_relation_cut obj obj' \<equiv>
```
lsf37, line 324: style: name should be on same line as `abbreviation`, `where` on same line as type

proof/refine/RISCV64/StateRelation.thy
```diff
@@ -188,6 +318,22 @@ primrec aobj_relation_cuts :: "RISCV64_A.arch_kernel_obj \<Rightarrow> machine_w
...
+  (case (obj, obj') of
+        (Structures_A.SchedContext sc n, KOSchedContext sc') \<Rightarrow> sc_relation sc n sc'
+      | _ \<Rightarrow> False)"
```
lsf37, line 327: remove outer parentheses

proof/refine/RISCV64/StateRelation.thy
```diff
@@ -262,28 +427,45 @@ definition pspace_relation ::
...
   "sched_act_relation choose_new_thread a' = (a' = ChooseNewThread)" |
   "sched_act_relation (switch_thread x) a' = (a' = SwitchToThread x)"
 
-definition ready_queues_relation ::
-  "(Deterministic_A.domain \<Rightarrow> Structures_A.priority \<Rightarrow> Deterministic_A.ready_queue) \<Rightarrow>
-   (domain \<times> priority \<Rightarrow> KernelStateData_H.ready_queue) \<Rightarrow> bool" where
+definition
+  ready_queues_relation :: "(Structures_A.domain \<Rightarrow> Structures_A.priority \<Rightarrow> Structures_A.ready_queue)
+                         \<Rightarrow> (domain \<times> priority \<Rightarrow> KernelStateData_H.ready_queue) \<Rightarrow> bool"
+  where
```
lsf37, line 461: style

proof/refine/RISCV64/StateRelation.thy
```diff
@@ -262,28 +427,45 @@ definition pspace_relation ::
...
   "ready_queues_relation qs qs' \<equiv> \<forall>d p. (qs d p = qs' (d, p))"
 
+definition
+  release_queue_relation :: "Structures_A.release_queue \<Rightarrow> KernelStateData_H.release_queue \<Rightarrow> bool"
+  where
```
lsf37, line 466: style

proof/refine/RISCV64/StateRelation.thy
```diff
@@ -423,69 +611,102 @@ definition state_relation :: "(det_state \<times> kernel_state) set" where
...
+  machine_state s = ksMachineState s';
+  work_units_completed s = ksWorkUnitsCompleted s';
+  domain_index s = ksDomScheduleIdx s';
+  domain_list s = ksDomSchedule s';
+  cur_domain s = ksCurDomain s';
+  domain_time s = ksDomainTime s';
+  consumed_time s = ksConsumedTime s';
+  cur_time s = ksCurTime s';
+  cur_sc s = ksCurSc s';
+  reprogram_timer s = ksReprogramTimer s' \<rbrakk> \<Longrightarrow> R"
```
lsf37, line 709: The previous indent was correct, they should all align with `pspace_relation`

proof/refine/RISCV64/KHeap_R.thy
```diff
@@ -6,80 +6,174 @@
 
 theory KHeap_R
 imports
+  "AInvs.AInvs"
```
lsf37: Remove `AInvs` re-import here, it is already imported and will destroy any simpset deletion setup from before.

proof/refine/RISCV64/KHeap_R.thy
```diff
@@ -6,80 +6,174 @@
...
+  unfolding valid_replies'_def
+  apply (rule iffI; clarsimp simp: obj_at'_def projectKOs valid_sz_simps)
+   apply (drule_tac x=rptr in spec, clarsimp simp: opt_map_def)
+  apply (clarsimp simp: pspace_alignedD' pspace_distinctD' opt_map_def projectKOs
+                  split: option.splits)
+  done
+
+primrec
+  same_caps' :: "Structures_H.kernel_object \<Rightarrow> Structures_H.kernel_object \<Rightarrow> bool"
+where
```
lsf37: style

proof/refine/RISCV64/KHeap_R.thy
```diff
@@ -6,80 +6,174 @@
...
+  "same_caps' (KOReply r) val = (\<exists>r'. val = KOReply r')"
+  by (cases val; fastforce)+
+
 lemma lookupAround2_known1:
   "m x = Some y \<Longrightarrow> fst (lookupAround2 x m) = Some (x, y)"
   by (fastforce simp: lookupAround2_char1)
 
+abbreviation (input)
+  set_ko' :: "machine_word \<Rightarrow> kernel_object \<Rightarrow> kernel_state \<Rightarrow> kernel_state"
+where
```
lsf37: style

proof/refine/RISCV64/KHeap_R.thy
```diff
@@ -6,80 +6,174 @@
...
+  "set_ko' ptr ko s \<equiv> s\<lparr>ksPSpace := (ksPSpace s)(ptr := Some ko)\<rparr>"
+
+abbreviation (input)
+  set_obj' :: "machine_word \<Rightarrow> ('a :: pspace_storable) \<Rightarrow> kernel_state \<Rightarrow> kernel_state"
+where
```
lsf37: style

proof/refine/RISCV64/KHeap_R.thy
```diff
@@ -155,42 +320,43 @@ lemma updateObject_cte_is_tcb_or_cte:
...
   "\<lbrace> obj_at' (\<lambda>x::'a::pspace_storable. True) t \<rbrace>
    setObject p (v::'a::pspace_storable)
   \<lbrace> \<lambda>rv. obj_at' (\<lambda>x::'a::pspace_storable. True) t \<rbrace>"
   apply (simp add: setObject_def split_def)
   apply (rule hoare_seq_ext [OF _ hoare_gets_post])
-  apply (clarsimp simp: valid_def in_monad obj_at'_def lookupAround2_char1 project_inject
-                 dest!: R)
-  apply (subgoal_tac "objBitsKO (injectKO v) = objBitsKO (injectKO obj)")
+  apply (clarsimp simp: valid_def in_monad obj_at'_def
+                        projectKOs lookupAround2_char1
```
lsf37: remove projectKOs

proof/refine/RISCV64/KHeap_R.thy
```diff
@@ -207,64 +373,261 @@ lemma obj_at_setObject2:
   apply (clarsimp simp: valid_def in_monad)
   apply (frule updateObject_type)
   apply (drule R)
-  apply (clarsimp simp: obj_at'_def)
+  apply (clarsimp simp: obj_at'_def projectKOs)
```
lsf37: revert projectKOs

proof/refine/RISCV64/KHeap_R.thy
```diff
@@ -207,64 +373,261 @@ lemma obj_at_setObject2:
...
 
-lemma updateObject_tcb_eta:
-  "updateObject (v :: tcb) = updateObject_default v"
-  by ((rule ext)+, simp)
+lemma obj_at'_set_obj'_distinct:
+  assumes "p \<noteq> p'"
+          "obj_at' Q p' s"
+  shows "obj_at' P p (set_ko' p' ko s) = obj_at' P p s"
+  using assms apply -
+  apply (rule iffI; fastforce simp: obj_at'_def projectKO_eq project_inject ps_clear_upd)
```
lsf37: `apply` should be on its own line

lsf37: To avoid the whole `using assms apply -` thing, we could rewrite the lemma statement into non-Isar from (no assumes/shows). The proof doesn't make use of any of the Isar features and the statement might actually be more readable.

proof/refine/RISCV64/KHeap_R.thy
```diff
@@ -207,64 +373,261 @@ lemma obj_at_setObject2:
...
+         (case_tac ko; clarsimp),
+         (rename_tac arch, case_tac arch; clarsimp)?)+
   done
 
+\<comment>\<open> Used to show a stronger variant of @{thm obj_at_setObject2} for many concrete types.
+
+    Needs to be a definition so we can easily refer to it within ML as a constant. \<close>
+definition distinct_updateObject_types ::
+  "('a :: pspace_storable) itself \<Rightarrow> ('b :: pspace_storable) itself \<Rightarrow> bool"
+where
```
lsf37: where should go on previous line (or be indented by 2 if it doesn't fit), so that `diff` and friends recognise the definition as one block.

proof/refine/RISCV64/KHeap_R.thy
```diff
@@ -445,147 +825,81 @@ lemma setObject_nosch:
...
-                 dest!: updateObject_objBitsKO)
-   apply (fastforce dest: bspec[OF _ domI])
-  apply (fastforce dest: bspec[OF _ domI])
-  done
+lemma get_sc_valid_sc'[wp]:
+  "\<lbrace> valid_objs' \<rbrace> getSchedContext sc \<lbrace> valid_sched_context' \<rbrace>"
+  unfolding getSchedContext_def by getObject_valid_obj
 
-lemma setObject_canonical[wp]:
-  shows     "\<lbrace>pspace_canonical'\<rbrace> setObject p val \<lbrace>\<lambda>rv. pspace_canonical'\<rbrace>"
```
lsf37: I was initially worried about these disappearing, but they are now much nicer further down. :+1:

proof/refine/RISCV64/KHeap_R.thy
```diff
@@ -645,6 +958,29 @@ proof -
                   elim: cte_wp_atE' canonical_address_add)
 qed
 
+(* FIXME rt merge: move to Word_lib *)
+lemma max_word_minus_1[simp]: "0xFFFFFFFF + 2^x = (2^x - 1::32 word)"
+  by simp
```
lsf37: This lemma is unlikely to be useful in RISCV (should need 64 word). It would be nice to find a generic phrasing of this, but it would be Ok for now to just add a 64 bit version.

proof/refine/RISCV64/KHeap_R.thy
```diff
@@ -794,34 +1130,29 @@ declare foldl_True[simp]
...
-  apply (simp add: get_simple_ko_def getEndpoint_def get_object_def
+  apply (simp add: get_simple_ko_def getEndpoint_def get_object_def gets_the_def
                    getObject_def bind_assoc ep_at_def2)
-  apply (clarsimp simp: in_monad split_def bind_def gets_def get_def return_def)
+  apply (clarsimp simp: in_monad split_def bind_def gets_def get_def return_def
+                 dest!: readObject_misc_ko_at')
   apply (clarsimp simp: assert_def fail_def obj_at_def return_def is_ep partial_inv_def)
-  apply (clarsimp simp: loadObject_default_def in_monad in_magnitude_check objBits_simps')
-  apply (clarsimp simp add: state_relation_def pspace_relation_def)
+  apply (rename_tac ep' ep)
```
lsf37: The names look unused, so the rename_tac can probably be removed

proof/refine/RISCV64/KHeap_R.thy
```diff
@@ -794,34 +1130,29 @@ declare foldl_True[simp]
...
+  apply (clarsimp simp: state_relation_def pspace_relation_def obj_at'_def projectKOs)
```
lsf37: revert projectKOs

proof/refine/RISCV64/KHeap_R.thy
```diff
@@ -920,13 +1287,12 @@ lemma setObject_other_corres:
...
   assumes t: "is_other_obj_relation_type (a_type ob)"
   assumes b: "\<And>ko. P ko \<Longrightarrow> objBits ko = objBits ob'"
-  assumes e: "\<And>ko. P ko \<Longrightarrow> exst_same' (injectKO ko) (injectKO ob')"
   assumes P: "\<And>(v::'a::pspace_storable). (1 :: machine_word) < 2 ^ (objBits v)"
   shows      "other_obj_relation ob (injectKO (ob' :: 'a :: pspace_storable)) \<Longrightarrow>
   corres dc (obj_at (\<lambda>ko. a_type ko = a_type ob) ptr and obj_at (same_caps ob) ptr)
             (obj_at' (P :: 'a \<Rightarrow> bool) ptr)
             (set_object ptr ob) (setObject ptr ob')"
-  supply image_cong_simp [cong del] projectKOs[simp del]
+  supply image_cong_simp [cong del]
```
lsf37: like will need to be reverted

proof/refine/RISCV64/KHeap_R.thy
```diff
@@ -935,8 +1301,8 @@ lemma setObject_other_corres:
   apply (unfold set_object_def setObject_def)
   apply (clarsimp simp: in_monad split_def bind_def gets_def get_def Bex_def
                         put_def return_def modify_def get_object_def x
-                        projectKOs obj_at_def
-                        updateObject_default_def in_magnitude_check [OF _ P])
+                        projectKOs obj_at_def in_magnitude_check[OF _ P]
+                        updateObject_default_def)
```
lsf37: can probably be reverted (projectKOs)

proof/refine/RISCV64/KHeap_R.thy
```diff
@@ -960,35 +1326,31 @@ lemma setObject_other_corres:
...
+                         is_other_obj_relation_type_def a_type_def other_obj_relation_def
+                  split: Structures_A.kernel_object.split_asm if_split_asm
+                         arch_kernel_obj.split_asm kernel_object.split_asm arch_kernel_object.split_asm)
+  apply (drule replyPrevs_of_non_reply_update[simplified])
+   apply (clarsimp simp: other_obj_relation_def; cases ob; cases "injectKO ob'";
+          simp split: arch_kernel_obj.split_asm)
+  apply (clarsimp simp add: opt_map_def)
+  done
+
+lemmas obj_at_simps = obj_at_def obj_at'_def projectKOs map_to_ctes_upd_other
```
lsf37: remove projectKOs

proof/refine/RISCV64/KHeap_R.thy
```diff
@@ -1008,18 +1370,315 @@ lemma setNotification_corres:
...
+                     other_obj_relation_def pte_relation_def
+              split: Structures_A.kernel_object.split_asm if_split_asm
+                     arch_kernel_obj.split_asm)
+
+lemma set_reply_corres: (* for reply update that doesn't touch the reply stack *)
+  "reply_relation ae ae' \<Longrightarrow>
+  corres dc \<top>
+            (obj_at' (\<lambda>ko. replyPrev_same ae' ko) ptr)
+            (set_reply ptr ae) (setReply ptr ae')"
+  proof -
```
lsf37: `proof` indents same as `lemma`

proof/refine/RISCV64/KHeap_R.thy
```diff
@@ -1008,18 +1370,315 @@ lemma setNotification_corres:
...
+    apply (drule_tac x=p in spec)
+    by (subst replyPrevs_of_replyPrev_same_update[simplified, where ob'=ae', simplified];
+        simp add: typ_at'_def ko_wp_at'_def obj_at'_def project_inject opt_map_def)
+  qed
+
+lemma setReply_not_queued_corres: (* for reply updates on replies not in fst ` replies_with_sc *)
+  "reply_relation r1 r2 \<Longrightarrow>
+  corres dc (\<lambda>s. ptr \<notin> fst ` replies_with_sc s) (reply_at' ptr)
+            (set_reply ptr r1) (setReply ptr r2)"
+  proof -
```
lsf37: indent 0

proof/refine/RISCV64/KHeap_R.thy
```diff
@@ -1502,526 +2098,850 @@ lemma setObject_state_refs_of_eq:
...
+lemma updateObject_cte[simp]:
+  "fst (updateObject (v::'a) (KOCTE cte) p x n s) = {}"
+  by (clarsimp simp: default_update updateObject_default_def in_monad projectKOs not_cte bind_def)
+
+lemma pspace_aligned'[wp]: "f p v \<lbrace>pspace_aligned'\<rbrace>"
+  and pspace_distinct'[wp]: "f p v \<lbrace>pspace_distinct'\<rbrace>"
+  and pspace_bounded'[wp]: "f p v \<lbrace>pspace_bounded'\<rbrace>"
+  and no_0_obj'[wp]: "f p v \<lbrace>no_0_obj'\<rbrace>"
+  and pspace_canonical'[wp]: "f p v \<lbrace>pspace_canonical'\<rbrace>"
+  and pspace_in_kernel_mappings'[wp]: "f p v \<lbrace>pspace_in_kernel_mappings'\<rbrace>"
```
lsf37: :+1:

proof/refine/RISCV64/Invariants_H.thy
michaelmcinerney: This translation was already in `KernelStateData_H.thy`, so I've removed this

proof/refine/RISCV64/KHeap_R.thy
```diff
@@ -2105,32 +3034,1807 @@ lemma doMachineOp_invs_bits[wp]:
...
+
+lemma pred_map_rewrite:
+  "pred_map P proj = opt_pred P proj"
+  by (fastforce simp: pred_map_def2)
+
+abbreviation sc_of2 :: "Structures_A.kernel_object \<rightharpoonup> Structures_A.sched_context" where
+  "sc_of2 ko \<equiv> case ko of kernel_object.SchedContext sc n \<Rightarrow> Some sc | _ \<Rightarrow> None"
+
+abbreviation scs_of2 :: "'z state \<Rightarrow> obj_ref \<rightharpoonup> Structures_A.sched_context" where
+  "scs_of2 \<equiv> (\<lambda>s. kheap s |> sc_of2)"
```
lsf37: This looks like translation between the two projection styles. Not sure there is much we can/should do about this, but it's probably worth taking a moment to make a conscious decision. I naturally tend towards the abbreviation style, because it is more lightweight and I used it in RISCV, but converting all MCS AInvs use to that would be quite a bit of work.

proof/refine/RISCV64/KHeap_R.thy
```diff
@@ -2105,32 +3034,1807 @@ lemma doMachineOp_invs_bits[wp]:
...
+abbreviation sc_replies_of2 :: "'z state \<Rightarrow> obj_ref \<Rightarrow>obj_ref list option" where
+  "sc_replies_of2 s \<equiv> scs_of2 s ||> sc_replies"
+
+lemma sc_replies_of_rewrite:
+  "sc_replies_of s = sc_replies_of2 s"
+  by (fastforce simp: sc_heap_of_state_def sc_replies_of_scs_def opt_map_def map_project_def
+              split: option.splits Structures_A.kernel_object.splits)
+
+definition
+  sc_replies_relation2_2 ::
```
lsf37: style

proof/refine/RISCV64/KHeap_R.thy
```diff
@@ -2105,32 +3034,1807 @@ lemma doMachineOp_invs_bits[wp]:
...
+
+lemmas updateSchedContext_corres = updateSchedContext_corres_gen[where P=\<top> and P'=\<top>, simplified]
+
+(* end : updateSchedContext *)
+
+end
+
+(* FIXME RT: rename *)
+(* this lets cross the sc size information from concrete to abstract *)
+lemma ko_at'_cross:
```
lsf37: We should perform the rename to something like `ko_at_sc_cross` before we perpetuate the name. Might be relatively simple to do on ARM as well.

proof/refine/RISCV64/KHeap_R.thy
```diff
@@ -2105,32 +3034,1807 @@ lemma doMachineOp_invs_bits[wp]:
...
+  "cross_rel simple_sched_action sch_act_simple"
+  apply (clarsimp simp: cross_rel_def)
+  by (fastforce simp: simple_sched_action_def sch_act_simple_def
+                dest: state_relation_sched_act_relation
+               split: Structures_A.scheduler_action.splits)
+
+(* FIXME RT: move to Corres_UL.thy *)
+lemma cross_relF:
+  "(s, s') \<in> state_relation \<Longrightarrow> cross_rel A B \<Longrightarrow> A s \<Longrightarrow> B s'"
+  by (clarsimp simp: cross_rel_def)
```
lsf37: perform move and remove on ARM as well.

lsf37:
Ok, the content looks very good! Lots of minor style comments mostly from things that are already cleaned up in RISCV, but where not yet on ARM.

I've looked at the first two commits, and after fixing up the style things, they would be good as is from my side.

Before we do the rest, I think it would be worth doing a pass through ARM Refine looking for FIXMEs that were introduced there and either decide to remove them or address them (or defer them if they are Ok to be duplicated). Otherwise we're duplicating the cleanup work we need to do later.

I've commented on the FIXMEs in the two commits that I think we should address. If applicable we should do them on ARM as well so that we stay mostly in sync.

proof/refine/RISCV64/Invariants_H.thy
michaelmcinerney: Yep, we can remove them

proof/refine/RISCV64/Invariants_H.thy
michaelmcinerney: Nope, removed

proof/refine/RISCV64/Invariants_H.thy
michaelmcinerney: This block appears in `master` at the very top of the file in ARM and ARM_HYP, but in this position in RISCV64. I moved it to the top in this commit. Where would be best?

proof/refine/RISCV64/EmptyFail.thy
michaelmcinerney: We do have `gets_the_empty_fail` which is generated by a `crunch` in `EmptyFail_AI`. It is `[wp]` only. Nothing else in that file seems to have other attributes. Should I remove `empty_fail_projectKO`, and then just add the `empty_fail (f x) ==> empty_fail (f $ x)` rule as `[intro!]`, and that should do?

proof/refine/RISCV64/Invariants_H.thy
lsf37: Ah, I missed that it was already at the top. Then it's fine to leave as is, I don't have a particular preference for position.

proof/refine/RISCV64/EmptyFail.thy
lsf37: Yes, that sounds good. I can't guarantee that we haven't missed anything, but I think it should work.

proof/refine/RISCV64/KHeap_R.thy
```diff
@@ -307,61 +670,18 @@ lemma setObject_cte_wp_at':
...
-  apply (auto dest: updateObject_default_result)
-  done
-
 lemma obj_at_setObject3:
   fixes Q::"'a::pspace_storable \<Rightarrow> bool"
   fixes P::"'a::pspace_storable \<Rightarrow> bool"
   assumes R: "\<And>ko s y n. (updateObject v ko p y n s)
                    = (updateObject_default v ko p y n s)"
-  assumes P: "\<And>(v::'a::pspace_storable). (1 :: machine_word) < 2 ^ (objBits v)"
+  assumes P: "\<And>(v::'a::pspace_storable). (1 :: word32) < 2 ^ (objBits v)"
```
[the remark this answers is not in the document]
michaelmcinerney: Reverted. I think this was applied automatically

proof/refine/RISCV64/KHeap_R.thy
michaelmcinerney: Should anything be done about this now, or should it be part of deciding on a heap projection approach?

lsf37: Let's make this part of the heap projection approach decision.

proof/refine/RISCV64/Invariants_H.thy
```diff
@@ -130,29 +163,132 @@ lemma st_tcb_at'_def:
...
-   | _                          => {}"
+type_synonym ref_set = "(obj_ref \<times> reftype) set"
+
+definition tcb_st_refs_of' :: "thread_state \<Rightarrow> ref_set" where
+  "tcb_st_refs_of' z \<equiv> case z of
+    (BlockedOnReply r)         \<Rightarrow> if bound r then {(the r, TCBReply)} else {}
+  | (BlockedOnReceive x _ r)   \<Rightarrow> if bound r then {(x, TCBBlockedRecv), (the r, TCBReply)}
+                                  else {(x, TCBBlockedRecv)}
+  | (BlockedOnSend x _ _ _ _)  \<Rightarrow> {(x, TCBBlockedSend)}
+  | (BlockedOnNotification x)  \<Rightarrow> {(x, TCBSignal)}
```
lsf37: Missed this in the previous run: `case` branches don't need parentheses on the left. Better to remove them.

proof/refine/RISCV64/InvariantUpdates_H.thy
michaelmcinerney: https://sel4.atlassian.net/browse/VER-1415

lsf37 (approved):
One final small style thing that I overlooked last time, otherwise I'm happy for this to be merged!

michaelmcinerney:
I made the same style change (no parentheses on the left of a case statement) to a couple other definitions in that area, and I also didn't include `LevityCatch.thy` in the right commit, but I don't think there's much going on there

PR #422, opened 2022-01-13
Page table abstraction for experimental-timeprot

robs-cse:
Have our abstraction of the microarchitecture support making the necessary distinction between virtual and physical addresses. Includes a basic test instantiation.


scottbuckley (approved):
Yep this all looks good. I wonder if we might want to remove the pagetable_invs stuff some time in future if it causes conflicts, but it makes sense to have it at least for now

PR #423, opened 2022-01-17
Remove unused lemmas

michaelmcinerney:


lsf37 (approved):
Nice!

Xaphiosis:
Thank you. Does this (partially) apply to the RISCV changes you made as well?

Xaphiosis (approved):

michaelmcinerney:
> Thank you. Does this (partially) apply to the RISCV changes you made as well?

None of the files here were part of the RISCV changes. 

Is it possible to redo the patch once I've done the FIXMEs on ARM, and then apply it only to those files I haven't gone through yet?

Xaphiosis:
> > Thank you. Does this (partially) apply to the RISCV changes you made as well?
> 
> None of the files here were part of the RISCV changes.
> 
> Is it possible to redo the patch once I've done the FIXMEs on ARM, and then apply it only to those files I haven't gone through yet?

In the same way one would use `git diff | patch ...` you can use `git show <my commit> | patch ...` to apply changes from a commit to a different path, and then if files don't exist or diff hunks don't apply, you'll get messages about it or .rej files whose contents you can (after a brief look) ignore.

michaelmcinerney:
There was one lemmas that I missed. I might just squash it with the rest of them to make the history slightly nicer

lsf37:
Yup, no problem.

proof/refine/ARM/Finalise_R.thy
```diff
@@ -3163,19 +3150,6 @@ lemma finaliseCapTrue_standin_bound_tcb_at':
   apply (case_tac cap; simp add: finaliseCapTrue_standin_def isCap_simps)
   by wpsimp
 
-lemma capDeleteOne_bound_tcb_at':
-  "\<lbrace>bound_tcb_at' P tptr and cte_wp_at' (isReplyCap \<circ> cteCap) callerCap\<rbrace>
-      cteDeleteOne callerCap \<lbrace>\<lambda>rv. bound_tcb_at' P tptr\<rbrace>"
```
corlewis, line 3168: Is there a more general version of this lemma now? We clearly don't use it since it can be removed, but it feels like one which might be useful to have available.

corlewis (approved):

proof/refine/ARM/Finalise_R.thy
michaelmcinerney: There isn't a more general version of this lemma. I can add it back in, though for some reason the proof requires a slight update (I think part of the reason I removed it was because this proof broke and I thought that if it wasn't used then it might not be worth fixing)

corlewis: I'm ok with not bothering. I can imagine it being useful but it's probably not worth fixing now with it not being used.

corlewis: Whoops, sorry, I see you've already done it.

PR #424, opened 2022-01-17
Move to lib

michaelmcinerney:


lsf37 (approved):
:+1:

proof/refine/ARM/Schedule_R.thy
```diff
@@ -2303,19 +2303,6 @@ lemma setReprogramTimer_active_sc_at'[wp]:
   unfolding active_sc_at'_def
   by wpsimp
 
-(* FIXME RT: move these whileM lemmas, prove `whileM_*_inv` using `whileM_wp_gen`. *)
```
corlewis, line 2306: Very minor, but do we want to keep the second half of this FIXME? I have to admit, that it's not completely clear to me what the intention of it was, so maybe we don't care about it anymore

corlewis (approved):
Thanks for cleaning these up!

proof/refine/ARM/Schedule_R.thy
michaelmcinerney: This turned out to be easy. The tests failed for some probably annoying reason, so we may as well update this and rerun the tests

lib/LemmaBucket.thy
```diff
@@ -561,4 +561,28 @@ lemma cases_conj_strg: "A \<and> B \<longrightarrow> (P \<and> A) \<or> (\<not>
...
+  apply (case_tac list; simp)
+  apply (fastforce simp: filter_id_conv)
+  done
+
+lemma case_list_when:
+  "(case l of
+      [] \<Rightarrow> return ()
+      | r # x \<Rightarrow> f r x)
+   = (when (l \<noteq> []) $ f (hd l) (tl l))"
+  by (clarsimp simp: list_case_If2)
```
Xaphiosis, line 586: since we're moving these, could we also clean up style? I don't think this one follows any

Xaphiosis:
Had one comment, not super relevant, otherwise looks good. Thanks for doing cleanup.

Xaphiosis (approved):

lib/LemmaBucket.thy
michaelmcinerney: Is it that the `[]` and the `r` should be indented the same amount? That's the only thing I can see here

lsf37: The `|` should align with `case`. This this case it looks like the whole case expression would fit on one line, which I would usually prefer. Also the naming convention for lists is with `s`, so it should be `r # xs`

michaelmcinerney: I put it all on the one line, and update the name of the variable

PR #426, opened 2022-01-20
Move to AInvs

michaelmcinerney:


lsf37 (approved):
There seems to be a proof failure (missing `thread_set_valid_tcbs`), but otherwise this looks good.

lsf37:
There is one issue after all: if we're moving a lemma, it's usually better to have the remove and add in one commit, not separate commits (i.e. to have an actual move commit). The idea is that the intermediate state has at least a chance of working.

In this case I'd suggest to squash them all. I'll make it a fairly big commit, but it's reasonably clear what is going on inside it.

michaelmcinerney:
> There is one issue after all: if we're moving a lemma, it's usually better to have the remove and add in one commit, not separate commits (i.e. to have an actual move commit). The idea is that the intermediate state has at least a chance of working.
> 
> In this case I'd suggest to squash them all. I'll make it a fairly big commit, but it's reasonably clear what is going on inside it.

Sure, if you're happy with one big commit for all the sessions, that's easier for me (separating these things was a big pain, actually)

lsf37:
> Sure, if you're happy with one big commit for all the sessions, that's easier for me (separating these things was a big pain, actually)

Yes, happy with that. It's fine for "move" commits to go over multiple sessions (since they are moving things). Ideally it would be nice to have units that fit together thematically, but I don't think it's worth the work in this case.

PR #428, opened 2022-01-27
More FIXME RTs

michaelmcinerney:
This removes many FIXME RTs from Refine. Most of the remaining FIXME RTs are either related to JIRA issues, or we've decided to leave them for now.

### DIFF
--- a/proof/invariant-abstract/ARM/ArchRetype_AI.thy
+++ b/proof/invariant-abstract/ARM/ArchRetype_AI.thy
@@ -646,8 +646,7 @@ lemma valid_untyped_helper [Retype_AI_assms]:
   have obj_at_pres: "\<And>P x. obj_at P x s \<Longrightarrow> obj_at P x ?ns"
   by (clarsimp simp: obj_at_def dest: domI)
    (erule pspace_no_overlapC [OF pn _ _ cover vp])
-  note blah[simp del] = atLeastAtMost_iff atLeastatMost_subset_iff atLeastLessThan_iff
-        Int_atLeastAtMost atLeastatMost_empty_iff
+  note blah[simp del] = atLeastAtMost_simps
   have cover':"range_cover ptr sz (obj_bits (default_object ty dev us (cur_domain s))) n"
     using cover tyunt tysc
     by (clarsimp simp: obj_bits_dev_irr)
@@ -733,8 +732,7 @@ lemma valid_cap:
     "(s::'state_ext state) \<turnstile> cap \<and> untyped_range cap \<inter> {ptr .. (ptr && ~~ mask sz) + 2 ^ sz - 1} = {}"
   shows "s' \<turnstile> cap"
   proof -
-  note blah[simp del] = atLeastAtMost_iff atLeastatMost_subset_iff atLeastLessThan_iff
-        Int_atLeastAtMost atLeastatMost_empty_iff
+  note blah[simp del] = atLeastAtMost_simps
   have cover':"range_cover ptr sz (obj_bits (default_object ty dev us (cur_domain s))) n"
     using cover tyunt tysc
     by (clarsimp simp: obj_bits_dev_irr)
@@ -1105,9 +1103,7 @@ lemma pspace_respects_device_region:
       and cap_refs_resp_dev: "cap_refs_respects_device_region s"
   shows "pspace_respects_device_region s'"
   proof -
-    note blah[simp del] = untyped_range.simps usable_untyped_range.simps atLeastAtMost_iff
-          atLeastatMost_subset_iff atLeastLessThan_iff
-          Int_atLeastAtMost atLeastatMost_empty_iff split_paired_Ex
+    note blah[simp del] = untyped_range.simps usable_untyped_range.simps atLeastAtMost_simps
   show ?thesis
   apply (cut_tac vp)
   apply (rule pspace_respects_device_regionI)
--- a/proof/invariant-abstract/RISCV64/ArchDetype_AI.thy
+++ b/proof/invariant-abstract/RISCV64/ArchDetype_AI.thy
@@ -495,9 +495,7 @@ proof -
 qed
 
 lemma in_user_frame_eq:
-  notes [simp del] = atLeastAtMost_iff atLeastatMost_subset_iff atLeastLessThan_iff
-                     Int_atLeastAtMost atLeastatMost_empty_iff split_paired_Ex
-                     order_class.Icc_eq_Icc
+  notes [simp del] = atLeastAtMost_simps order_class.Icc_eq_Icc
     and [simp] = p2pm1_to_mask
   shows "p \<notin> untyped_range cap \<Longrightarrow>
          in_user_frame p (s \<lparr>kheap := \<lambda>x. if x \<in> untyped_range cap then None else kheap s x\<rparr>)
@@ -520,10 +518,7 @@ lemma in_user_frame_eq:
     done
 
 lemma in_device_frame_eq:
-  notes blah[simp del] =  atLeastAtMost_iff
-          atLeastatMost_subset_iff atLeastLessThan_iff
-          Int_atLeastAtMost atLeastatMost_empty_iff split_paired_Ex
-          order_class.Icc_eq_Icc
+  notes blah[simp del] =  atLeastAtMost_simps order_class.Icc_eq_Icc
      and  p2pm1[simp] = p2pm1_to_mask
   shows "p \<notin> untyped_range cap
        \<Longrightarrow> in_device_frame p (s \<lparr>kheap := \<lambda>x. if x \<in> untyped_range cap then None else kheap s x\<rparr>)
--- a/proof/refine/ARM/Schedule_R.thy
+++ b/proof/refine/ARM/Schedule_R.thy
@@ -8,6 +8,10 @@ theory Schedule_R
 imports SchedContext_R InterruptAcc_R
 begin
 
+crunches scReleased, getReprogramTimer, getCurTime, getRefills, getReleaseQueue, refillSufficient,
+         refillReady, isRoundRobin
+  for inv[wp]: P
+
 context begin interpretation Arch . (*FIXME: arch_split*)
 
 declare static_imp_wp[wp_split del]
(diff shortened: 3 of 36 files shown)


proof/invariant-abstract/ARM/ArchRetype_AI.thy
```diff
@@ -646,8 +646,7 @@ lemma valid_untyped_helper [Retype_AI_assms]:
   have obj_at_pres: "\<And>P x. obj_at P x s \<Longrightarrow> obj_at P x ?ns"
   by (clarsimp simp: obj_at_def dest: domI)
    (erule pspace_no_overlapC [OF pn _ _ cover vp])
-  note blah[simp del] = atLeastAtMost_iff atLeastatMost_subset_iff atLeastLessThan_iff
-        Int_atLeastAtMost atLeastatMost_empty_iff
+  note blah[simp del] = atLeastAtMost_simps
```
lsf37, line 649: You don't have to change it in this iteration, but if we don't make reference to the name `blah` further down, it is actually Ok to not have a name here, because the scope is global.

```isabelle
note [simp del] = atLeastAtMost_simps
```

would be legal syntax and Ok here (we just don't want un-named global simp lemmas)

lsf37 (approved):
Very nice! And really got some milage out of `atLeastAtMost_simps` :-)

We could remove the `blah` names, which never made any sense and got copied around a lot, but that should probably be separate and done on master as well.

proof/invariant-abstract/RISCV64/ArchDetype_AI.thy
```diff
@@ -495,9 +495,7 @@ proof -
 qed
 
 lemma in_user_frame_eq:
-  notes [simp del] = atLeastAtMost_iff atLeastatMost_subset_iff atLeastLessThan_iff
-                     Int_atLeastAtMost atLeastatMost_empty_iff split_paired_Ex
-                     order_class.Icc_eq_Icc
+  notes [simp del] = atLeastAtMost_simps order_class.Icc_eq_Icc
```
Xaphiosis, line 498: I like this kind of thing instead of the `blah[simp del]` seen in other places. If we're doing cleanup, would it make sense to get rid of the `blah` at some point?

proof/invariant-abstract/ARM/ArchRetype_AI.thy
Xaphiosis: Oh yeah, I said a similar thing in another comment, sorry for the dupe. Agreed.

proof/refine/ARM/Untyped_R.thy
```diff
@@ -3397,24 +3396,20 @@ lemma updateFreeIndex_updateCap_caps_no_overlap'':
   apply (clarsimp simp:caps_no_overlap''_def)
   apply (wp updateCap_ctes_of_wp)
   apply (clarsimp simp: modify_map_def ran_def cte_wp_at_ctes_of
-              simp del: atLeastAtMost_iff atLeastatMost_subset_iff atLeastLessThan_iff
-                        Int_atLeastAtMost atLeastatMost_empty_iff split_paired_Ex)
+              simp del: atLeastAtMost_simps)
   apply (case_tac "a = src")
-   apply (clarsimp simp del: atLeastAtMost_iff atLeastatMost_subset_iff atLeastLessThan_iff
-     Int_atLeastAtMost atLeastatMost_empty_iff split_paired_Ex)
+   apply (clarsimp simp del: atLeastAtMost_simps)
```
Xaphiosis, line 3401: This raises an issue for me, the amount of times that `atLeastAtMost_simps` is removed from the simpset: please tell me we're not the ones who put it in the simpset? If we did, it might be wise to take it out.

proof/refine/RISCV64/KHeap_R.thy
```diff
@@ -2943,6 +2996,10 @@ lemma ep_redux_simps':
   by (fastforce split: list.splits
                 simp: valid_ep_def valid_ntfn_def)+
 
+lemma endpoint_live':
+  "\<lbrakk>ko_at' ep ptr s; ep \<noteq> IdleEP\<rbrakk> \<Longrightarrow> ko_wp_at' live' ptr s"
+  apply (clarsimp simp: ko_wp_at'_def obj_at'_def)
+  done
```
Xaphiosis, line 3002: by?

proof/refine/RISCV64/KHeap_R.thy
```diff
@@ -4307,6 +4364,178 @@ lemma update_sched_context_rewrite:
                          is_sc_obj_def)
   done
 
+lemmas sc_inv_state_eq' = getObject_sc_inv[THEN use_valid[rotated], rotated
+                                           , where s=s and P="(=) s" for s, OF _ refl]
+
+lemma sc_inv_state_eq:
+  "(a :: sched_context, s') \<in> fst (getSchedContext p s) \<Longrightarrow> s' = s"
+  apply (fastforce dest: sc_inv_state_eq' simp: getSchedContext_def)
+  done
```
Xaphiosis, line 4373: by?

proof/refine/RISCV64/KHeap_R.thy
```diff
@@ -4307,6 +4364,178 @@ lemma update_sched_context_rewrite:
...
+  apply (drule_tac ob1="g (f sc)" in no_failD[OF no_fail_setObject_other, rotated])
+   apply simp+
+  done
+
+lemmas getSchedContext_setSchedContext_decompose_decompose_ext
+  = getSchedContext_setSchedContext_decompose[where f="f x" and g="g y" for f g x y]
+lemmas getSchedContext_setSchedContext_decompose_decompose2
+  = getSchedContext_setSchedContext_decompose[where g="\<lambda>sc. g (h sc)" for g h]
+lemmas getSchedContext_setSchedContext_decompose_decompose_ext2
+  = getSchedContext_setSchedContext_decompose[where f="f x" and g="g y" for f g x y]
```
Xaphiosis, line 4495: @lsf37 : how much do we indent the `=`? I've seen four spaces in this PR, I've seen people putting the `=` under the lemmas name, etc.

proof/refine/RISCV64/KHeap_R.thy
```diff
@@ -4307,6 +4364,178 @@ lemma update_sched_context_rewrite:
...
+    apply (clarsimp simp: split: option.splits)
+    apply (metis (no_types) array_rules(2) lookupAround2_char2 mcs(1) order.strict_trans2
+                            word_le_less_eq word_le_not_less)
+   apply (simp add: in_monad getSchedContext_def getObject_def)
+   apply (frule no_ofailD[OF no_ofail_sc_at'_readObject])
+   apply (clarsimp del: readObject_misc_ko_at' simp del: readObject_misc_obj_at')
+   apply (clarsimp simp: setSchedContext_def setObject_def projectKOs in_monad ps_clear_upd obj_at'_def
+                         split_def updateObject_default_def magnitudeCheck_assert RISCV64_H.fromPPtr_def
+                  dest!: readObject_misc_ko_at')
+
```
Xaphiosis, line 4437: I don't know if the extra empty lines help in this lemma, might be worth dropping them.

proof/refine/ARM/TcbAcc_R.thy
```diff
@@ -1442,6 +1425,13 @@ lemma threadSet_valid_tcbs':
   apply (clarsimp simp: obj_at'_def projectKOs valid_tcbs'_def)
   done
 
+lemma asUser_valid_tcbs'[wp]:
+  "asUser t f \<lbrace>valid_tcbs'\<rbrace>"
+  apply (simp add: asUser_def split_def)
+  apply (wpsimp wp: threadSet_valid_tcbs' hoare_drop_imps
+              simp: valid_tcb'_def tcb_cte_cases_def)
```
Xaphiosis, line 1432: optional: can one-line this by adding the initial defs to wpsimp's simp

proof/refine/ARM/Schedule_R.thy
```diff
@@ -8,6 +8,9 @@ theory Schedule_R
 imports SchedContext_R InterruptAcc_R
 begin
 
+crunches scReleased
+  for inv: P
```
Xaphiosis, line 12: can this be squashed with other `inv` crunches?

Xaphiosis:
I have a few comments, which if addressed would likely need to apply to all arches (I picked whichever one I saw first). In general, this is very welcome work, and I'm grateful you're doing it.

proof/refine/RISCV64/KHeap_R.thy
lsf37: I would of course prefer it to be on the end of the previous line. Otherwise, standard indent would be 2, I think.

Xaphiosis: The nitpicky question is "to whom does the `=`" belong to. Clearly it's under `lemmas`, but if it is under the name as well, then:

```
lemmas xxx
       = ...
```

In your version with the `=` at EOL, the same question arises, where does the `...` go? Does it belong to the name or the `lemmas`? At the moment I'm siding with it belonging with the name, not `lemmas`, but not clear cut on it.

Xaphiosis: (of course this doesn't need changing in this commit, but more for style guide later)

lsf37: If it was a normal `lemma` statement (not `lemmas`), I think we would also continue with a 2-indent, and not indent with the name, so I'm for staying consistent with that.

Xaphiosis: Works for me!

proof/refine/ARM/Untyped_R.thy
michaelmcinerney: No, I think all of the rules in `atLeastAtMost_simps` are added to the simpset within the HOL theory files

proof/refine/ARM/TcbAcc_R.thy
corlewis: I feel like this is almost another style question. This approach of having an explicit step unfolding the definitions separately to the 'actual proof' is very common.

Personally, I find it useful when repairing a broken proof to be able to so easily see the definition of the function, but this isn't enough of a benefit for me to care much. It would be nice however for us to make a conscious decision about what the preferred style is so that we can be consistent going forwards.

Xaphiosis: Agreed, but if we are unfolding, I prefer `unfolding X Y by ...` to the double-simp.

proof/refine/ARM/Schedule_R.thy
michaelmcinerney: Actually, I just saw that there is another `crunches` for `inv` lower in the file. I'll move those to the top of this file

proof/refine/RISCV64/KHeap_R.thy
michaelmcinerney: I generally prefer using `by` only when the last step or two takes more than a few seconds. I sometimes use it for proofs which are very short and which we wouldn't expect to change often. Is there enough of a preference for `by` for me to change these? Or to go through and change a whole bunch of them?

proof/refine/RISCV64/KHeap_R.thy
michaelmcinerney: I'm a big fan of extra lines in nasty proofs like this one. The indentation does already say "yes this is the end of that subgoal", of course, but I do like the extra separation. Any further thoughts on what to do with extra lines?

proof/refine/RISCV64/KHeap_R.thy
Xaphiosis: From my perspective: It doesn't look like an interesting proof, or one that's likely to break. The extra line wastes visual space, adding nothing. If it breaks, you'll need to change something on the one line anyway, at which point having to change by to apply won't matter much. @lsf37 might have a different view, but I picked up the habit from him

proof/refine/ARM/Untyped_R.thy
Xaphiosis: Shame. If that's the case, then this is a fine way of handling the problem.

proof/refine/RISCV64/KHeap_R.thy
Xaphiosis: Fine to leave them if they add something. Looking over this specific proof, I didn't see the benefit, but you'd know better than me.

proof/refine/RISCV64/KHeap_R.thy
lsf37: I tend to use `by` for most one-liners, yes. I don't have an extremely strong opinion on it, but I'd say a preference for.

PR #433, opened 2022-02-07
RISCV Refine till end of `TcbAcc_R.thy`

michaelmcinerney:


proof/refine/RISCV64/CSpace1_R.thy
```diff
@@ -189,23 +179,20 @@ lemma can_be_is:
...
-                        loadObject_cte split_def in_monad
-                 dest!: in_singleton
-             split del: if_split)
-  apply (clarsimp simp: in_monad typeError_def objBits_simps
-                        magnitudeCheck_def
-                 split: kernel_object.split_asm if_split_asm option.split_asm
-             split del: if_split)
-       apply simp+
-  done
+  by (wpsimp simp: getCTE_def)
```
lsf37, line 195: very nice!

proof/refine/RISCV64/CSpace1_R.thy
```diff
@@ -427,17 +411,17 @@ proof -
...
-       apply (simp_all add: simps)[4]
+  apply (cases c; simp add: simps)
+   defer
    apply (clarsimp simp: simps the_arch_cap_def)
    apply (rename_tac arch_cap)
-   apply (case_tac arch_cap; simp add: simps arch_update_cap_data_def
-                             RISCV64_H.updateCapData_def)
+   apply (case_tac arch_cap)
+        apply (simp_all add: arch_update_cap_data_def
+                             RISCV64_H.updateCapData_def)[5]
```
lsf37, line 420: `simp_all` is a bit old-style, these days I would prefer something like
```isabelle
   apply (case_tac arch_cap; simp add: arch_update_cap_data_def RISCV64_H.updateCapData_def)
```
This doesn't need the explicit goal limiter and won't leak if the number of goals change.

proof/refine/RISCV64/ArchAcc_R.thy
```diff
@@ -12,19 +12,126 @@ theory ArchAcc_R
...
+
+lemmas setObject_ASID_cteCaps_of[wp] = ctes_of_cteCaps_of_lift [OF setObject_asidpool.ctes_of]
+lemmas storePTE_cteCaps_of[wp] = ctes_of_cteCaps_of_lift [OF storePTE.ctes_of]
+
+lemma no_0_obj'_abstract:
+  "(s, s') \<in> state_relation \<Longrightarrow> no_0_obj' s' \<Longrightarrow> kheap s 0 = None"
+  by (auto intro: pspace_relation_None simp add: no_0_obj'_def)
+
+declare if_cong[cong]
+
```
lsf37, line 128: This is a dangerous declaration, a) for non-termination and b) for general simp performance. We should not mess with if congruence rules globally and I thought we had managed to eliminate that. Did we re-introduce this in `rt`?

In any case, now that you have the theory working, I'd remove this and add only a local `notes if_cong[cong]` to the lemmas that need it (or even just the simp invocations that need it).

lsf37 (approved):
Nice. I have only the `if_cong` declaration I'm concerned about.

proof/refine/RISCV64/ArchAcc_R.thy
michaelmcinerney: There is a `declare if_cong[cong]` in RISCV on master. And there's a seemingly different `declare if_cong`in ARM on rt. But sure, I will try to remove it

lsf37: Then I was less successful eradicating it than I thought (maybe I just did AInvs). If it can be removed, that would be really good. If it makes everything unmanageable, it's Ok to put a FIXME on it and leave for now.

michaelmcinerney: Yeah a few proofs did break in the CSpace files, too, so I will add the FIXME RT. Is there anything you meant by "tagged"?

lsf37: Cool, `FIXME RT` is good (`RT` is the tag)

lsf37:
:+1:

PR #434, opened 2022-02-08
arm+arm-hyp machine: align platform constants to C

corlewis:
This matches the C code, as seen at https://github.com/seL4/seL4/blob/master/include/arch/arm/arch/32/mode/hardware.h#L72.

I explored aligning more of the platform constants, both for ARM and the other architectures, but it very rapidly became more work than what seemed worthwhile.

### DIFF
--- a/spec/machine/ARM/Platform.thy
+++ b/spec/machine/ARM/Platform.thy
@@ -48,11 +48,11 @@ definition physBase :: word32 where
 definition pptrBaseOffset :: word32 where
   "pptrBaseOffset \<equiv> pptrBase - physBase"
 
-definition kernelELFBase :: word32 where
-  "kernelELFBase \<equiv> pptrBase + (physBase && mask 22)"
-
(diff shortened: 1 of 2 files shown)


Xaphiosis:
Could we get a mention of why you're doing it in the commit message? Also, have you looked at the way it's handled on other arches (especially RISCV64 where it does a written-out calculation that is checked against C)?
I'm happy with the change, but I want to have a bit more context for the historians.

corlewis:
I've updated the commit message. Does the explanation make sense now?

I did look at how RISCV64 handles it and explored going in that direction but unfortunately I don't have capacity at the moment to properly change ARM and ARM-HYP to that approach.

Xaphiosis (approved):

Xaphiosis:
I'd probably have mentioned the constant you removed, but it's fine. For ARM_HYP there isn't much finesse with these constants, as they get unfolded most of the time anyway... which is likely why you didn't see proof changes upon representation changes.

Issue #413, opened 2021-12-21
add CI for different `numDomains` settings


lsf37:
When PR #410 is merged, the proofs will be generic in the number of protection domains.

Testing that fact with every CI job is going to be too resource-intensive, but we should test at least values of `1`, `2`, and something like `16` in the weekly job so we get an alert when we accidentally break this genericity.

Xaphiosis:
We already test 16 as it is the default. Did you mean another number?

lsf37:
I mostly meant the weekly should also keep testing the normal default. Not sure if we want to leave it at `16`, but `16` is as good as any I guess.

Xaphiosis:
Then you propose *adding* `1` and `2` to the weekly test? That makes sense.
Other thoughts:
- we need a branch with a commit that changes all the verified seL4 configs that keeps getting rebased before we test (we already discussed that)
- when someone adds a new config, the above commit will not change the new config, and I have not a clue how one would track or notice that; an alternative to the above method is to take master and search+replace the number of domains in all configs

lsf37:
> Then you propose adding 1 and 2 to the weekly test? That makes sense.

yes, that would have been much clearer :-)

> - we need a branch with a commit that changes all the verified seL4 configs that keeps getting rebased before we test (we already discussed that)

Ah, I had the branch discussion put under "for different platform", not for different config. For changing the config in only the number of domains, I'd propose for the CI job to edit the `*_verified.config` files (via regexp) before it starts the test. That should be pretty targeted and test the entire chain.

Xaphiosis:
Search+replace via regexp sounds like a good plan and will catch other arches should they appear.
Yes, you're right, I'm conflating discussions. You proposed the floating commit for a platform port. My bad.

lsf37:
This has been implemented in https://github.com/seL4/l4v/pull/701 and related PRs.

PR #435, opened 2022-02-17
update to Isabelle 2021-1

lsf37:
Mainly Word_Lib changes this time, and consequences of tighter context checking in Isabelle-2021-1.

The PR proof tests will fail, because they will test with Isabelle2021, but there is a [testboard commit](https://github.com/seL4/gh-testboard/commit/727123b) for the changes here that uses `ts-2021-1`.

One thing that should still be tested is if everything also works with `numDomains=1` (see also #413). Will run that locally.
Edit: `numDomains = 1` has passed.

Also notable is that I removed some old theories like `AutoLevity*` and `TSubst` which were mostly bitrotted, but also the `wpx` method. The `wpx` method broke in the update and isn't really used much, so to reduce maintenance burden for the future, I removed it and updated the proofs instead (most just worked with `wp`).


lsf37:
The theory linter failure is for a demo file, which should be fine.

michaelmcinerney (approved):
I carefully looked at one of the commits for AInvs and another for Refine, and they looked good. I did have a random look at another half a dozen commits, and of what I understand, they looked good too

lib/Monad_WP/wp/WPI.thy
```diff
@@ -254,7 +254,7 @@ ML \<open>
     val frees = Term.add_frees t [];
     val frees' = Term.add_frees t' [];
   in
-   exists (member (=) frees') frees
+   exists (member (op =) frees') frees
```
Xaphiosis, line 257: `op` is back?

lsf37: It never went away in ML, only in Isabelle. It's a miracle that this ever worked, but polyml allows it, it just complains that it is not legal ML but moves on. (A bunch of these got introduced by a too-global search and replace when `op` was removed in Isabelle)

lib/Word_Lib/Bit_Shifts_Infix_Syntax.thy
```diff
@@ -0,0 +1,209 @@
+(*
+ * Copyright Data61, CSIRO (ABN 41 687 119 230)
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ *)
+
+(* Author: Jeremy Dawson, NICTA *)
```
Xaphiosis, line 7: github thinks this file is new; sanity check of what happened?

lib/Word_Lib/Bitwise.thy
```diff
@@ -165,22 +166,22 @@ lemma rbl_sshiftr:
...
    apply (case_tac "LENGTH('a)", simp_all)
   apply (rule word_eqI)
-  apply (simp add: nth_sshiftr word_size test_bit_of_bl
+  apply (simp add: bit_simps word_size test_bit_of_bl
       msb_nth)
   done
 
 lemma nth_word_of_int:
-  "(word_of_int x :: 'a::len word) !! n = (n < LENGTH('a) \<and> bin_nth x n)"
+  "bit (word_of_int x :: 'a::len word) n = (n < LENGTH('a) \<and> bit x n)"
```
Xaphiosis, line 174: we're losing `!!` syntax entirely?

lib/Word_Lib/Bit_Shifts_Infix_Syntax.thy
lsf37: It is new, Florian refactored a whole bunch of these. (This is re-import from AFP)

lib/Word_Lib/Bitwise.thy
lsf37: It's recreated later again as optional syntax. I don't really like it either, but at least optional does make sense. In our proofs nothing changes in the syntax.

lib/Word_Lib/Signed_Words.thy
```diff
@@ -93,6 +93,21 @@ lemma zero_sle_ucast:
...
+         \<longleftrightarrow> (bit (((2::'a::len  word) ^ n) :: 'a word) m)"
+  by (simp add: bit_simps)
+
+lemma scast_nop1 [simp]:
+  "((scast ((of_int x)::('a::len) word))::'a signed word) = of_int x"
+  by transfer (simp add: take_bit_signed_take_bit)
+
+lemma scast_nop2 [simp]:
+  "((scast ((of_int x)::('a::len) signed word))::'a word) = of_int x"
+  by transfer (simp add: take_bit_signed_take_bit)
```
Xaphiosis, line 107: when these are proved by `transfer`, what is is transferring to here?

proof/access-control/ARM/ArchRetype_AC.thy
```diff
@@ -165,20 +165,19 @@ lemma copy_global_mappings_pas_refined:
        copy_global_mappings pd
        \<lbrace>\<lambda>_. pas_refined aag\<rbrace>"
   apply (simp add: copy_global_mappings_def)
-  apply (rule hoare_weaken_pre)
   apply wp
-  (* Use \<circ> to avoid wp filtering out the global_pd condition here
+    (* Use \<circ> to avoid wp filtering out the global_pd condition here
      TODO: see if we can clean this up *)
```
Xaphiosis, line 170: TODO?

lib/Word_Lib/More_Word_Operations.thy
```diff
@@ -977,14 +972,27 @@ lemma from_to_bool_last_bit:
...
 lemma sint_ctz:
-  "LENGTH('a) > 2
-   \<Longrightarrow> 0 \<le> sint (of_nat (word_ctz (x :: 'a :: len word)) :: 'a signed word)
-        \<and> sint (of_nat (word_ctz x) :: 'a signed word) \<le> int (LENGTH('a))"
-  apply (subgoal_tac "LENGTH('a) < 2 ^ (LENGTH('a) - 1)")
-   apply (rule conjI)
-    apply (metis len_signed order_le_less_trans sint_of_nat_ge_zero word_ctz_le)
-   apply (metis int_eq_sint len_signed sint_of_nat_le word_ctz_le)
-  using small_powers_of_2 [of \<open>LENGTH('a)\<close>] by simp
+  \<open>0 \<le> sint (of_nat (word_ctz (x :: 'a :: len word)) :: 'a signed word)
```
Xaphiosis, line 975: is this proof yours or upstream?

lib/Word_Lib/Word_Lemmas.thy
```diff
@@ -895,26 +804,57 @@ lemma ucast_leq_mask:
   done
 
 lemma shiftl_inj:
```
Xaphiosis, line 806: are these proofs faster or something? this is way longer than it used to be...

lib/Word_Lib/Word_Lemmas.thy
```diff
@@ -1480,7 +1441,14 @@ lemma mask_shift_eq_mask_mask:
...
   "\<lbrakk> a \<ge> b; unat n = unat (p AND mask b) \<rbrakk>
    \<Longrightarrow> (p AND NOT(mask a)) + (p AND mask a >> b) * (1 << b) + n = (p :: 'a :: len word)"
-  by (metis and_not_mask mask_rshift_mult_eq_rshift_lshift mask_split_sum_twice word_unat.Rep_eqD)
+  apply (simp add: shiftl_def shiftr_def flip: push_bit_eq_mult take_bit_eq_mask word_unat_eq_iff)
+  apply (subst disjunctive_add)
+   apply (auto simp add: bit_simps)
+  apply (subst disjunctive_add)
+   apply (auto simp add: bit_simps)
+    apply (rule bit_word_eqI)
+  apply (auto simp add: bit_simps)
```
Xaphiosis, line 1450: this can't be your work, not sure why someone did this change

proof/crefine/ARM/Finalise_C.thy
```diff
@@ -1262,8 +1260,7 @@ lemma deleteASID_ccorres:
                            asid_low_bits_def order_le_less_trans [OF word_and_le1])
     apply wp
    apply vcg
-  apply (clarsimp simp: Collect_const_mem if_1_0_0
```
Xaphiosis, line 1265: is Collect_const_mem now in the simpset?

lib/Word_Lib/Signed_Words.thy
lsf37: I have no idea, these are all Florian's doing and just re-imported. I assume that the base type for this is `int` (because of the `of_int`, so `transfer` would use the corresponding theorems for `of_int` to produce an `int` goal from this.

proof/crefine/ARM/SR_lemmas_C.thy
```diff
@@ -507,14 +507,21 @@ lemma fst_setCTE:
...
+  obtain s' where
+    "((), s')\<in>fst (setCTE dest cte s)"
+    "s' = s\<lparr>ksPSpace := ksPSpace s'\<rparr>"
+    "dom (ksPSpace s) = dom (ksPSpace s')"
+    "(\<forall>p \<in> dom (ksPSpace s').
+       case the (ksPSpace s p) of
+         KOTCB t \<Rightarrow> \<exists>t'. ksPSpace s' p = Some (KOTCB t') \<and> tcb_no_ctes_proj t = tcb_no_ctes_proj t'
+       | KOCTE _ \<Rightarrow> \<exists>cte. ksPSpace s' p = Some (KOCTE cte)
+       | _ \<Rightarrow> ksPSpace s' p = ksPSpace s p)"
+    by clarsimp
```
Xaphiosis, line 520: what happened here? is the yuck comment still applicable?

proof/crefine/ARM/SyscallArgs_C.thy
```diff
@@ -735,13 +735,19 @@ lemma ccap_relation_page_is_device:
     apply (auto split: if_splits simp: to_bool_def generic_frame_cap_get_capFIsDevice_CL_def Let_def)
    done
 
+(* FIMXE isa: move to Word_Lib/Word_32.thy*)
```
Xaphiosis, line 738: misspelled FIXME, maybe you missed it?

proof/crefine/ARM_HYP/Arch_C.thy
```diff
@@ -1664,30 +1664,36 @@ lemma performPageInvocationMapPTE_ccorres:
              apply (rule conjI, clarsimp)
               apply (clarsimp split: if_split_asm)
                apply (case_tac a; clarsimp simp: isLargePagePTE_def)
-               apply (clarsimp simp: cpte_relation_def pte_lift_small split del: split_of_bool)
+               apply (clarsimp simp: cpte_relation_def pte_lift_small
+                               split del: split_of_bool)
```
Xaphiosis, line 1668: why all these? there's no difference to original except reindenting

Xaphiosis: if you're going this far, a `supply` might make more sense?

tools/c-parser/umm_heap/CTypesBase.thy
```diff
@@ -319,9 +319,9 @@ lemma intvl_plus_sub_Suc:
 
 lemma intvl_neq_start:
   "\<lbrakk> (q::'a::len word) \<in> {p..+n}; p \<noteq> q \<rbrakk> \<Longrightarrow> q \<in> {p + 1..+n - Suc 0}"
-  by (clarsimp simp: intvl_def)
-     (metis One_nat_def diff_Suc_1 less_Suc_eq_0_disj less_imp_Suc_add of_nat_Suc take_bit_eq_0_iff
-            take_bit_of_0)
+  apply (clarsimp simp: intvl_def)
+  by (metis One_nat_def Suc_eq_plus1 gr0_conv_Suc less_diff_conv not_gr_zero of_nat_Suc
+            semiring_1_class.of_nat_0)
```
Xaphiosis: this is unusual for you, why the split from a single `by`?

tools/c-parser/Simpl/ex/Quicksort.thy
```diff
@@ -217,65 +167,55 @@ apply (hoare_rule anno =
...
+    apply simp
+    apply (rule_tac x="ps" in exI)
+    apply simp
+   apply (metis insertCI set_mset_add_mset_insert set_mset_mset)
+  apply clarsimp
+  apply (rule_tac ?x=grs in exI)
+  apply (rule conjI)
+  apply (erule heap_eq_ListI1)
+   apply clarify
+   apply (erule_tac x=x in allE) back
```
Xaphiosis: are these style failures due to `back` ancient Harvey code?

tools/asmrefine/SimplExport.thy
```diff
@@ -1,4 +1,4 @@
-(*
+    (*
```
Xaphiosis: - [ ] whoops reindent

lib/Word_Lib/Word_Lemmas.thy
lsf37: Oh wow, that violates AFP rules as well. Need to fix there.

spec/cspec/KernelState_C.thy
```diff
@@ -40,4 +40,9 @@ lemmas conseqPost = HoarePartialDef.conseqPost
 (* Likewise, we'd prefer to get HOL.conj_cong over StateFun.conj_cong *)
 lemmas conj_cong = HOL.conj_cong
 
+(* FIXME isa:  Removes two rules that are too eager for the bitfield proofs.
```
Xaphiosis: FIXME isa

proof/refine/X64/Schedule_R.thy
```diff
@@ -155,10 +155,7 @@ lemma removeFromBitmap_valid_queues_no_bitmap_except[wp]:
 lemma removeFromBitmap_bitmapQ:
   "\<lbrace> \<lambda>s. True \<rbrace> removeFromBitmap d p \<lbrace>\<lambda>_ s. \<not> bitmapQ d p s \<rbrace>"
   unfolding bitmapQ_defs bitmap_fun_defs
-  apply (wp| clarsimp simp: bitmap_fun_defs)+
-  apply (subst (asm) complement_nth_w2p, simp_all)
-  apply (fastforce intro!: order_less_le_trans[OF word_unat_mask_lt] simp: word_size wordRadix_def')
-  done
+  by (wp| clarsimp simp: bitmap_fun_defs)+
```
Xaphiosis: space after wp, since it's redone?

lib/Word_Lib/More_Word_Operations.thy
lsf37: Upstream (word changes are 99% upstream, I've only added things in _Internal)

proof/refine/ARM_HYP/Schedule_R.thy
```diff
@@ -486,44 +459,27 @@ crunch global_refs'[wp]: tcbSchedAppend valid_global_refs'
...
 crunch irq_states[wp]: tcbSchedDequeue valid_irq_states'
 
-crunch ct'[wp]: tcbSchedEnqueue "\<lambda>s. P (ksCurThread s)"
-  (simp: unless_def)
 crunch ct'[wp]: tcbSchedAppend "\<lambda>s. P (ksCurThread s)"
   (simp: unless_def)
-crunch ct'[wp]: tcbSchedDequeue "\<lambda>s. P (ksCurThread s)"
 
-crunch pde_mappings'[wp]: tcbSchedEnqueue "valid_pde_mappings'"
-  (simp: unless_def)
```
Xaphiosis: lots of crunches went away here; removing un-needed ones, or something update related?

proof/infoflow/RISCV64/Example_Valid_State.thy
```diff
@@ -1345,103 +1345,8 @@ lemma pspace_distinct_s0:
...
-     apply simp
-     apply (rule_tac sz=59 in machine_word_plus_mono_right_split)
-      apply (simp add: unat_word_ariths mask_def shiftl_t2n)
-      apply (cut_tac x=irq and 'a=64 in ucast_less)
-       apply simp
-      apply (simp add: word_less_nat_alt)
-     apply (simp add: word_bits_def)
-    apply simp
-   apply simp
+   apply word_bitwise
```
Xaphiosis: oh wow, that's special. nicely done

proof/crefine/X64/StoreWord_C.thy
```diff
@@ -50,16 +50,15 @@ lemma byte_to_word_heap_upd_outside_range:
...
-  done
+  by (clarsimp intro!: intvl_inter_le [where k=0 and ka=7, simplified, OF refl]
+                       intvl_inter_le [where k=0 and ka=6, simplified, OF refl]
+                       intvl_inter_le [where k=0 and ka=5, simplified, OF refl]
+                       intvl_inter_le [where k=0 and ka=4, simplified, OF refl]
+                       intvl_inter_le [where k=0 and ka=3, simplified, OF refl]
+                       intvl_inter_le [where k=0 and ka=2, simplified, OF refl]
+                       intvl_inter_le [where k=0 and ka=1, simplified, OF refl]
+                       intvl_inter_le [where k=0 and ka=0, simplified, OF refl]
+                split: if_split_asm) (* long, many if_asm splits *)
```
Xaphiosis, line 61: does this change make much difference?

proof/crefine/RISCV64/SyscallArgs_C.thy
```diff
@@ -757,13 +757,19 @@ lemma ccap_relation_page_is_device:
     apply (auto split: if_splits simp: to_bool_def Let_def cap_frame_cap_lift_def)
    done
 
+(* FIMXE isa: move to Word_Lib/Word_32.thy*)
```
Xaphiosis, line 760: FIXME isa

Xaphiosis:
Finished going through the files. Comments:
- very nice work, this is a huge amount of work and I really appreciate you go through it
- if you agree with a comment, please check the other arches as well, because I did not have the mental capacity to notice the same issue on multiple arches
- I was surprised and have a little mixed feelings about the amount of crunches that got removed in AInvs (e.g. Ipc_AI) and Refine (several files)... from a commit perspective it looks like these were removed due to the Isabelle update, from a cleanup perspective this is great, and from a debugging perspective (i.e. when something breaks and you get to pin it down a bit faster with more granularity)...  I'm not sure. What do you think, did I miss something?
- What *did* wpx do? I never used it and completely forgot if anyone ever explained it to me; would be nice to know what we're losing
- What are your thoughts on some proofs switching from simpler or simp-based to metis with a bunch of rules? As soon as I see them, I think they got spat out by sledgehammer and worry about how slow they might be (I remember going through a separation logic file once where someone used sledgehammer for everything because it was new and shiny, and my simpler non-metis proofs were significantly faster). Again, curious what you think and how to manage future speed degradation.

proof/crefine/ARM/SyscallArgs_C.thy
lsf37: Indeed I did, well spotted.

proof/crefine/ARM/Finalise_C.thy
lsf37: No, still unchanged, it just happened to not actually be used here (was probably a copy/paste). The change I was after was probably the `if_1_0_0`, because that generates a warning.

lib/Word_Lib/Word_Lemmas.thy
lsf37: This probably just needs a very different proof now because all the fundamentals changed, but I haven't done any of these, I don't really know.

proof/crefine/ARM/SR_lemmas_C.thy
lsf37: The `guess` command was removed. I think the `Yuck` is now more justified than before :-). (But really could also be removed, it doesn't add anything useful)

proof/crefine/ARM_HYP/Arch_C.thy
lsf37: This is a result of having added `split del: split_of_bool_asm` before and then removing it again (moving it to a bundle). Better would be to revert that as well, but I didn't find any good way of getting just the bits that changed because of add/remove cycle.

tools/c-parser/Simpl/ex/Quicksort.thy
lsf37: This is `Simpl/ex/Quicksort`, so a demo (which seems to have changed in the AFP import). No Harvey connection this time :-)

spec/cspec/KernelState_C.thy
lsf37: This is resolved in a later commit

proof/refine/ARM_HYP/Schedule_R.thy
lsf37: removing unneeded ones. It's unclear if warnings are from the update or not, these specific ones probably are not, but I usually remove anything that looks feasible in an update instead of trying to figure out where it comes from.

lsf37:
> * if you agree with a comment, please check the other arches as well, because I did not have the mental capacity to notice the same issue on multiple arches

Will do and well spotted on that FIMXE tag.

> * I was surprised and have a little mixed feelings about the amount of crunches that got removed in AInvs (e.g. Ipc_AI) and Refine (several files)... from a commit perspective it looks like these were removed due to the Isabelle update, from a cleanup perspective this is great, and from a debugging perspective (i.e. when something breaks and you get to pin it down a bit faster with more granularity)...  I'm not sure. What do you think, did I miss something?

Not sure I follow on the debugging. The current state works and they are removed because they complained loudly that they didn't do or prove anything.

I try to remove warnings in updates, because updates create new warnings -- if we don't do it there we will never do it and just accumulate them. Trying to figure out if a warning is new or old and not remove it when it's old is too expensive and I don't think there is real benefit to doing that in a large update like this. The commit would get a bit nicer, but if you're hitting an Isabelle update there is no real point any more in bisecting or honing in to single commits anyway.


> * What _did_ wpx do? I never used it and completely forgot if anyone ever explained it to me; would be nice to know what we're losing

It had no documentation, of course :-) One of the reasons it was almost unused, I guess. From the code, it was lifting over operators more automatically, and there was one proof where I needed some tweaking to get wpsimp to work. It was an experiment, but not successful as a replacement or real improvement.

> * What are your thoughts on some proofs switching from simpler or simp-based to metis with a bunch of rules? As soon as I see them, I think they got spat out by sledgehammer and worry about how slow they might be (I remember going through a separation logic file once where someone used sledgehammer for everything because it was new and shiny, and my simpler non-metis proofs were significantly faster). Again, curious what you think and how to manage future speed degradation.

They are all sledgehammer, of course, but for proofs that did not work, so nothing got slower, it got from not working to working. Usually these are faster than simp, I don't think there is any degradation here. (That is not always true, of course, but current sledgehammer will suggest simp proofs if there are obvious simp proofs).

spec/cspec/KernelState_C.thy
lsf37: It seems that is not true. Will update (basically just dropping the FIXME part of it, because there is nothing really to fix -- I thought I could do something upstream in Word_Lib, but that doesn't work).

Xaphiosis:
> I try to remove warnings in updates, because updates create new warnings -- if we don't do it there we will never do it and just accumulate them. Trying to figure out if a warning is new or old and not remove it when it's old is too expensive and I don't think there is real benefit to doing that in a large update like this. The commit would get a bit nicer, but if you're hitting an Isabelle update there is no real point any more in bisecting or honing in to single commits anyway.

That's a very solid justification, new warnings appearing as part of Isabelle update, or warning reduction. Sold!
 
> They are all sledgehammer, of course, but for proofs that did not work, so nothing got slower, it got from not working to working. Usually these are faster than simp, I don't think there is any degradation here. (That is not always true, of course, but current sledgehammer will suggest simp proofs if there are obvious simp proofs).

I don't think there's any point changing any of these for this PR. It is interesting that our experiences with sledgehammer->metis are so different. By decomposing the metis proof into some other invocation that effectively uses the simplifier in a more targeted fashion (often using inspiration of what sledgehammer dredged up for metis), I almost always get a faster proof (sometimes even a shorter one), but not one that sledgehammer would suggest. That said, there's no generic way of doing that step, and sometimes metis really is the right move (seems to do well at flipping equalities). I'm interested in the cumulative performance effect of using sledgehammered metis in this manner, but maybe there's no point thinking about it.

lsf37:
> That's a very solid justification, new warnings appearing as part of Isabelle update, or warning reduction. Sold!

👍 
 
> > They are all sledgehammer, of course, but for proofs that did not work, so nothing got slower, it got from not working to working. Usually these are faster than simp, I don't think there is any degradation here. (That is not always true, of course, but current sledgehammer will suggest simp proofs if there are obvious simp proofs).
> 
> I don't think there's any point changing any of these for this PR. It is interesting that our experiences with sledgehammer->metis are so different. 

Sledgehammer has improved drastically in the last 10 years and that separation logic experience is quite a while ago. I don't think you would get metis proofs for much of that any more, you'd mainly get simp and blast suggestions (might be nice to try out, actually, maybe I'm completely wrong about this). metis is still frequent and slow metis proofs still do happen, esp when there are lots of rules, but my main experience these days is that most metis proofs are fast.

> By decomposing the metis proof into some other invocation that effectively uses the simplifier in a more targeted fashion (often using inspiration of what sledgehammer dredged up for metis), I almost always get a faster proof (sometimes even a shorter one), but not one that sledgehammer would suggest. 

For slow proofs that can absolutely work, yes. 

> That said, there's no generic way of doing that step, and sometimes metis really is the right move (seems to do well at flipping equalities). I'm interested in the cumulative performance effect of using sledgehammered metis in this manner, but maybe there's no point thinking about it.

I maintain that these are improvement: there may be one or two slow ones in the update, but on average they are fast and I would think that overall speed has improved rather than decreased. That is hard to measure of course, because the rest of the runtime speed also has changed and neither proof works with the other Isabelle version. But working with it interactively I didn't see any problems introduced from them.

proof/crefine/X64/StoreWord_C.thy
lsf37: Took me a while to dig this one out. Yes, absolutely worth it, this otherwise blocks for almost a full minute. I first thought something new was slowing this down, but it seems to always have been slow.

lsf37:
Ok, have fixed up the comments and rebased over current master. This includes the recent VCPU invariant work and other commits, so there may be some failures. Will looks at these once the [testboard](https://github.com/seL4/gh-testboard/runs/5713160128) has some info.

proof/crefine/ARM/SR_lemmas_C.thy
lsf37: Have removed the comment.

Xaphiosis (approved):

PR #437, opened 2022-02-22
asmrefine pfx (prefix for finding constants via name) + Kernel_C tweaks

Xaphiosis:
@mbrcknl : hoping this doesn't break other BV efforts

Needs full regression test to make sure I didn't derp anything else.

### DIFF
--- a/tools/asmrefine/SimplExport.thy
+++ b/tools/asmrefine/SimplExport.thy
@@ -101,39 +101,41 @@ fun get_field ctxt s = let
     handle ERROR _ => NONE
        | Bind => NONE
 
-val read_const = Proof_Context.read_const {proper = true, strict = true}
+fun read_const ctxt (pfx : string) (name : string) = let
+    val nm = if pfx = "" then name else (pfx ^ "." ^ name)
+  in Proof_Context.read_const {proper = true, strict = true} ctxt nm end
 
-fun process_struct ctxt csenv (nm, flds) = let
+fun process_struct ctxt csenv pfx (nm, flds) = let
     val offs = map (ProgramAnalysis.offset csenv (map snd flds))
         (0 upto (length flds - 1))
-    val cons = read_const ctxt (nm ^ "." ^ nm)
+    val cons = read_const ctxt pfx (nm ^ "." ^ nm)
     val typ = dest_Const cons |> snd |> strip_type |> snd
     val sz = ProgramAnalysis.sizeof csenv (Absyn.StructTy nm)
     val algn = ProgramAnalysis.alignof csenv (Absyn.StructTy nm)
     val accs = map (fst #> prefix (nm ^ ".")
-        #> read_const ctxt) flds
+        #> read_const ctxt pfx) flds
   in (nm, (cons, typ, sz, algn, map fst flds ~~ (accs ~~ offs))) end
 
-fun structs ctxt csenv = ProgramAnalysis.get_senv csenv
-  |> map (process_struct ctxt csenv)
+fun structs ctxt csenv pfx = ProgramAnalysis.get_senv csenv
+  |> map (process_struct ctxt csenv pfx)
   |> Symtab.make
 
-fun structs_by_typ ctxt csenv = Symtab.dest (structs ctxt csenv)
+fun structs_by_typ ctxt csenv pfx = Symtab.dest (structs ctxt csenv pfx)
   |> map (fn (nm, (cons, typ, sz, algn, flds)) => (fst (dest_Type typ), (nm, cons, sz, algn, flds)))
   |> Symtab.make |> Symtab.lookup
 
-fun cons_fields ctxt csenv = structs ctxt csenv |> Symtab.dest
+fun cons_fields ctxt csenv pfx = structs ctxt csenv pfx |> Symtab.dest
   |> map (fn (_, (cons, typ, _, _, flds))
     => (fst (dest_Const cons), (fst (dest_Type typ),
            map (snd #> fst #> dest_Const #> fst) flds)))
   |> Symtab.make |> Symtab.lookup
 
-fun enums ctxt csenv = let
+fun enums ctxt csenv pfx = let
     val Absyn.CE ecenv = ProgramAnalysis.cse2ecenv csenv
   in
     #enumenv ecenv |> Symtab.dest
     |> map (fn (s, (n, _)) =>
-        (read_const ctxt s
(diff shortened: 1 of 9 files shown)


tools/asmrefine/SimplExport.thy
```diff
@@ -1105,15 +1110,18 @@ fun emit_func_body ctxt outfile eparams name = let
     else (emit outfile ("1 Basic Ret 0");
           emit_body ctxt outfile eparams body 2 "1" ("ErrExc", "ErrExc")
             |> snd |> prefix "EntryPoint " |> emit outfile
+(*
           handle TERM (s, ts) => raise TERM ("emit_func_body: " ^ name ^ ": " ^ s, body :: @{term True} :: ts)
             | TYPE (s, Ts, ts) => raise TYPE ("emit_func_body: " ^ name ^ ": " ^ s, Ts, body :: @{term True} :: ts)
             | Empty => raise TERM ("emit_func_body: Empty", [body]))
+*)
+)
```
mbrcknl, line 1118: Did you mean to leave this commented out?

tools/asmrefine/SimplExport.thy
```diff
@@ -1063,10 +1067,11 @@ fun emit_body ctxt outfile params (Const (@{const_name Seq}, _) $ a $ b) n c e =
   | emit_body _ _ _ t _ _ _ = raise TERM ("emit_body", [t])
 
 fun emit_func_body ctxt outfile eparams name = let
+    val _ = warning ("EMIT_FUNC_BODY " ^ name)
```
mbrcknl, line 1070: Did you mean to leave this tracing here?

mbrcknl (approved):
I've not kept up with the Kernel_Config development, so I don't fully grok everything here. But I think I understand enough to see roughly why the `pfx` is needed in asmrefine.

In any case, I think this should be fine for the rest of BV. If it makes it through SimplExportAndRefine, then it shouldn't affect anything downstream.

tools/asmrefine/SimplExport.thy
Xaphiosis: Embarrassing medical incident in the git surgery (and the tracing too), thank you for spotting! Will fix.

Xaphiosis: Though I'll wait until regression proof tests complete, as these don't change the success criteria.

Xaphiosis:
Oh great, AsmRefine failed on the ARMs. Needs work.

mbrcknl:
> Oh great, AsmRefine failed on the ARMs. Needs work.

Probably not anything special about the ARMs. Apparently, we never turned on AsmRefine for RISC-V. The contents of the AsmRefine session get imported by SimplExportAndRefine, so the only thing that's not checked on RISC-V is AsmRefineTest. Of course, that's what failed on the ARMs.

I don't understand the error, though...

```
*** exception Io {cause = SysErr ("No such file or directory", SOME ENOENT),
***               function = "TextIO.openIn", name =
***               "/home/test-runner/ver/l4v/tools/asmrefine/testfiles/global_array_swap_Cfuns.txt"} raised (line 284 of "./basis/TextIO.sml")
*** At command "ML" (line 114 of "~/ver/l4v/tools/asmrefine/testfiles/global_array_swap_gref.thy")
```

I can't see how you get to global_array_swap_gref.thy:114 without creating the file at line 28 (`emit_C_everything_relative`), assuming that the `@{theory}` at line 114 is the same as the theory of the `@{context}` at line 28.

Xaphiosis:
> I can't see how you get to global_array_swap_gref.thy:114 without creating the file at line 28 (`emit_C_everything_relative`), assuming that the `@{theory}` at line 114 is the same as the theory of the `@{context}` at line 28.

This is a derp by me. I added an extra argument `pfx` but then I don't pass anything for it for `emit_C_everything_relative` in the `asmrefine/testfiles`, so ML happily hands me back a partial function and doesn't generate any files, hence the IO Error.

Xaphiosis:
It's still somewhat strange that we test AsmRefine on ARM and ARM_HYP, but SimplExportAndRefine on ARM and RISCV64... should we have turned that on for RISCV64?

lsf37:
> @mbrcknl :
> It's still somewhat strange that we test AsmRefine on ARM and ARM_HYP, but SimplExportAndRefine on ARM and RISCV64... should we have turned that on for RISCV64?

I think it would make sense to turn it on for RISCV64 (at least if you can get it to work without putting in a lot of time)

Xaphiosis:
> > @mbrcknl :
> > It's still somewhat strange that we test AsmRefine on ARM and ARM_HYP, but SimplExportAndRefine on ARM and RISCV64... should we have turned that on for RISCV64?
> 
> I think it would make sense to turn it on for RISCV64 (at least if you can get it to work without putting in a lot of time)

Ok I'll give it a go.

mbrcknl:
> > I can't see how you get to global_array_swap_gref.thy:114 without creating the file at line 28 (`emit_C_everything_relative`), assuming that the `@{theory}` at line 114 is the same as the theory of the `@{context}` at line 28.
> 
> This is a derp by me. I added an extra argument `pfx` but then I don't pass anything for it for `emit_C_everything_relative` in the `asmrefine/testfiles`, so ML happily hands me back a partial function and doesn't generate any files, hence the IO Error.

Ah yes, I didn't make that connection. :-)

> @mbrcknl :
> It's still somewhat strange that we test AsmRefine on ARM and ARM_HYP, but SimplExportAndRefine on ARM and RISCV64... should we have turned that on for RISCV64?

This is probably just an oversight (by me) when we set up SimplExportAndRefine on RISC-V. We should have also turned on AsmRefine.

Turning AsmRefine on for ARM_HYP was probably just because it's cheap and we could, but we didn't turn on SimplExportAndRefine because it's expensive.

Xaphiosis:
Yeah, I can see why it wasn't turned on, it needs some kind of per-arch testfiles folder thing because of `global_asm_stmt.c` which uses `int` to do an inline asm register read, which yields a nasty looking error:
```
exception TYPE raised (line 474 of "~/repos/v2/l4v/tools/c-parser/isar_install.ML"):
  type_of: type mismatch in application
  32 word
  64 word
  (λx _. x) B.1
```
Which is fair enough, using an ARM test file with inline assembly isn't going to work on RISCV64. To get this to work, we'd need separate arch folders and an inline asm example for each. I might leave that to another pull request.

PR #438, opened 2022-02-24
Add `vcpu_switch` to `associate_vcpu_tcb`, and some WIP on `cur_vcpu_valid`

michaelmcinerney:


proof/invariant-abstract/ARM_HYP/ArchVSpace_AI.thy
```diff
@@ -2656,6 +2656,37 @@ lemma invs_current_vcpu_update': "\<lbrakk> cur_vcpu_at v s \<and> invs s \<rbra
             valid_global_objs_def valid_global_vspace_mappings_def
            split: option.split)
 
+lemma vgic_update_sym_refs_hyp[wp]:
+  "\<lbrace>\<lambda>s. sym_refs (state_hyp_refs_of s)\<rbrace> vgic_update vcpuptr f \<lbrace>\<lambda>rv s. sym_refs (state_hyp_refs_of s)\<rbrace>"
```
lsf37, line 2660: These would be slightly nicer with invariant notation `f {| .. |}`, but I see they are just moved from ARchFinalise, so not really a must to change.

proof/invariant-abstract/ARM_HYP/ArchArch_AI.thy
```diff
@@ -910,7 +910,7 @@ lemma associate_vcpu_tcb_sym_refs_hyp[wp]:
        apply (clarsimp simp add: sym_refs_def)
        apply (case_tac "x = t"; case_tac "x = vr"; clarsimp simp add: state_hyp_refs_of_def obj_at_def
                                                                  dest!: get_tcb_SomeD)
-       apply fastforce
+        apply fastforce
       apply (rule hoare_pre)
```
lsf37, line 914: Is this one intentional? Looks inconsistent with the `hoare_pre` line.

proof/invariant-abstract/ARM_HYP/ArchArch_AI.thy
```diff
@@ -983,24 +987,196 @@ lemma valid_global_vspace_mappings_vcpu_update_str:
...
+  for valid_global_refs[wp]: valid_global_refs
+  (wp: crunch_wps)
+
+crunches vcpu_restore_reg
+  for valid_irq_states[wp]: valid_irq_states
+  (wp: crunch_wps dmo_valid_irq_states simp: writeVCPUHardwareReg_def)
+
+method repeat_unless methods stop cont
+  = (stop | (cont, repeat_unless stop cont))
+
```
lsf37, line 1048: Cool, nice one. This one looks more generally useful and could move up to somewhere in Lib where we define more Eisbach methods (maybe with a small example).

lsf37: Maybe swap the arguments? With the name `repeat_unless` I would expect the `cont` first and the `stop` second.

lsf37 (approved):
Nice! Looking very good.

proof/invariant-abstract/ARM_HYP/ArchArch_AI.thy
michaelmcinerney: I'll doube check this one

proof/invariant-abstract/ARM_HYP/ArchArch_AI.thy
michaelmcinerney: This (as well as `hoare_seq_ext_skip`) is just moved from Lib. They mustn't have found their way into `master` yet. There aren't many uses of `repeat_unless` in `rt`. I agree that the arguments would be better swapped, but should we do it both here and on `rt`, or leave it for now?

lsf37: Hm, that changes things. Ideally we'd change it in `rt` first and then backport the Lib change, but that sounds like a bunch of work that is a distraction.

Let's maybe at least find the `rt` commit that puts it into Lib and cherry-pick that onto this branch, so that it ends up in master when we merge here. This should avoid a merge conflict for updates later. For the argument order, let's leave it as is for now, maybe with a FIXME. It is documented, just not the order of least surprise.

Xaphiosis: Probably worth to point out that with this argument order and definition this looks more like `repeat_until`. It repeats `cont` until `stop` applies. So I would second putting in a FIXME and reconsidering merge with RT later.

proof/invariant-abstract/ARM_HYP/ArchArch_AI.thy
```diff
@@ -983,24 +987,196 @@ lemma valid_global_vspace_mappings_vcpu_update_str:
...
+  and pspace_respects_device_region[wp]: pspace_respects_device_region
+  and cur_tcb[wp]: cur_tcb
+  and valid_vspace_objs[wp]: valid_vspace_objs
+  and valid_arch_caps[wp]: valid_arch_caps
+  and valid_kernel_mappings[wp]: valid_kernel_mappings
+  and equal_kernel_mappings[wp]: equal_kernel_mappings
+  and valid_asid_map[wp]: valid_asid_map
+  and valid_global_vspace_mappings[wp]: valid_global_vspace_mappings
+  and pspace_in_kernel_window[wp]: pspace_in_kernel_window
+  (wp: crunch_wps dmo_valid_irq_states device_region_dmos simp: crunch_simps)
```
Xaphiosis, line 1019: This looks like a very large crunch addition for one function. Are there not other `crunches` that could have `associate_vcpu_tcb` added to them to collapse some of these?

lsf37: I think it's nicer to concentrate them, because you get only one place to update if you change the function. One of the nice things `crunches` enabled.

proof/invariant-abstract/ARM_HYP/ArchArch_AI.thy
```diff
@@ -983,24 +987,196 @@ lemma valid_global_vspace_mappings_vcpu_update_str:
...
+  (wp: crunch_wps dmo_valid_irq_states simp: writeVCPUHardwareReg_def)
+
+method repeat_unless methods stop cont
+  = (stop | (cont, repeat_unless stop cont))
+
+lemma is_irq_active_sp:
+  "\<lbrace>P\<rbrace> get_irq_state irq \<lbrace>\<lambda>rv s. P s \<and> (rv = interrupt_states s irq)\<rbrace>"
+  by (wpsimp simp: get_irq_state_def)
+
+\<comment> \<open>For forward reasoning in oare proofs, these lemmas allow skipping over the
```
Xaphiosis, line 1053: *Hoare

proof/invariant-abstract/ARM_HYP/ArchArch_AI.thy
lsf37: (of course if there is a second such function with a comparable set we should merge them)

Xaphiosis: Objection your honour: `crunches` was created to do a *2D* scan of a set of properties over several functions. I guess we can place this one on the fire of aesthetic appeal... from the other perspective, if you have a bunch of functions crunching the same property, that gives you a different form of concentration. I'd ideally prefer more clustering together, especially for uninteresting things (e.g. hardly anything changes valid_asid_map/valid_global_vspace_mappings)

proof/invariant-abstract/ARM_HYP/ArchArch_AI.thy
```diff
@@ -983,24 +987,196 @@ lemma valid_global_vspace_mappings_vcpu_update_str:
...
+  "vcpu_restore vcpu \<lbrace>valid_irq_states\<rbrace>"
+  unfolding vcpu_restore_def isb_def
+  by (wpsimp wp: dmo_valid_irq_states set_gic_vcpu_ctrl_hcr_irq_masks set_gic_vcpu_ctrl_vmcr_irq_masks
+                 set_gic_vcpu_ctrl_lr_irq_masks mapM_wp_inv set_gic_vcpu_ctrl_apr_irq_masks)
+
+crunches vcpu_switch
+  for valid_irq_states[wp]: valid_irq_states
+  (wp: crunch_wps dmo_valid_irq_states
+   simp: get_gic_vcpu_ctrl_lr_def get_gic_vcpu_ctrl_apr_def
+         get_gic_vcpu_ctrl_vmcr_def get_gic_vcpu_ctrl_hcr_def isb_def)
```
Xaphiosis, line 1085: Trying to understand crunch limitations here... wouldn't this be able to handle vcpu_restore_valid_irq_states as well, given the extra rules above? I think vcpu_switch calls vcpu_restore? I keep getting the feeling that some of these crunches could be compressed fairly easily. If this is copied from elsewhere, might not be worth the bother.

proof/invariant-abstract/ARM_HYP/ArchArch_AI.thy
```diff
@@ -983,24 +987,196 @@ lemma valid_global_vspace_mappings_vcpu_update_str:
...
+  "do_machine_op (set_cntv_off_64 w') \<lbrace>valid_machine_state\<rbrace>"
+  "do_machine_op (maskInterrupt m irq) \<lbrace>valid_machine_state\<rbrace>"
+  "do_machine_op (setHCR word) \<lbrace>valid_machine_state\<rbrace>"
+  "do_machine_op isb \<lbrace>valid_machine_state\<rbrace>"
+  "do_machine_op dsb \<lbrace>valid_machine_state\<rbrace>"
+  "do_machine_op (set_gic_vcpu_ctrl_hcr f) \<lbrace>valid_machine_state\<rbrace>"
+  "do_machine_op (set_gic_vcpu_ctrl_lr n w'') \<lbrace>valid_machine_state\<rbrace>"
+  "do_machine_op (set_gic_vcpu_ctrl_apr w''') \<lbrace>valid_machine_state\<rbrace>"
+  "do_machine_op (set_gic_vcpu_ctrl_vmcr w''')  \<lbrace>valid_machine_state\<rbrace>"
+  "do_machine_op (get_gic_vcpu_ctrl_lr w'''')  \<lbrace>valid_machine_state\<rbrace>"
```
Xaphiosis, line 1138: `"  "` -> `" "`

proof/invariant-abstract/ARM_HYP/ArchArch_AI.thy
```diff
@@ -983,24 +987,196 @@ lemma valid_global_vspace_mappings_vcpu_update_str:
...
-   by (wp get_vcpu_wp arch_thread_get_wp weak_if_wp as_user_only_idle hoare_vcg_all_lift
-        | wp (once) hoare_drop_imps
-        | wpc
-        | clarsimp
-        | strengthen valid_arch_state_vcpu_update_str valid_global_refs_vcpu_update_str
-                     valid_global_vspace_mappings_vcpu_update_str valid_global_objs_vcpu_update_str
-        | simp add: associate_vcpu_tcb_def valid_obj_def[abs_def] valid_vcpu_def
-                    dissociate_vcpu_tcb_def vcpu_invalidate_active_def vcpu_disable_def
-        | simp add: obj_at_def)+
+  by (wpsimp simp: invs_def valid_state_def valid_pspace_def pred_conj_def)
```
Xaphiosis, line 1179: nice

proof/invariant-abstract/ARM_HYP/ArchVCPU_AI.thy
```diff
@@ -0,0 +1,354 @@
+(*
+ * Copyright 2022, ???
```
Xaphiosis, line 2: FIXME copyright

lsf37: Well spotted. In this case UNSW, unless Mitch had already created the file, the D61 and UNSW

proof/invariant-abstract/ARM_HYP/ArchVCPU_AI.thy
```diff
@@ -0,0 +1,354 @@
...
+  apply (clarsimp simp: cur_vcpu_valid_def active_cur_vcpu_of_def pred_tcb_at_def obj_at_def)
+  done
+
+lemma vgic_update_cur_vcpu_valid[wp]:
+  "vgic_update vr f \<lbrace>cur_vcpu_valid\<rbrace>"
+  apply (clarsimp simp: vgic_update_def vcpu_update_def)
+  apply (rule hoare_seq_ext_skip, wpsimp)+
+  apply (wpsimp wp: set_vcpu_wp get_vcpu_wp)
+  apply (clarsimp simp: cur_vcpu_valid_def active_cur_vcpu_of_def pred_tcb_at_def obj_at_def)
+  done
```
Xaphiosis, line 97: these 4 lemmas look a bit copy-pasta heavy... is there no way to compress or automate this a little better?

proof/invariant-abstract/ARM_HYP/ArchVCPU_AI.thy
```diff
@@ -0,0 +1,354 @@
...
+  (wp: cur_vcpu_valid_lift crunch_wps mapM_x_wp_inv simp: active_cur_vcpu_of_def crunch_simps)
+
+lemma switch_vcpu_cur_vcpu_valid[wp]:
+  "\<lbrace>\<lambda>s. arch_tcb_at (\<lambda>a. itcb_vcpu a = v) t s\<rbrace>
+   vcpu_switch v
+   \<lbrace>\<lambda>_ s. cur_vcpu_valid (s\<lparr>cur_thread := t\<rparr>)\<rbrace>"
+  apply (clarsimp simp: vcpu_switch_def)
+  apply (cases v; clarsimp)
+   apply (rule hoare_seq_ext[OF _ gets_sp], rename_tac cur_vcpu)
+   apply (case_tac cur_vcpu; clarsimp)
```
Xaphiosis, line 143: clarsimp unnecessary here, you apply wpsimp to one branch and clarsimp to the other

proof/invariant-abstract/ARM_HYP/ArchVCPU_AI.thy
```diff
@@ -0,0 +1,354 @@
...
+   apply (clarsimp simp: when_def)
+   apply (intro conjI impI)
+    apply (rule hoare_seq_ext_skip, wpsimp)
+    apply (wpsimp simp: cur_vcpu_valid_def active_cur_vcpu_of_def)
+   apply (wpsimp simp: cur_vcpu_valid_def active_cur_vcpu_of_def)
+  apply (rule hoare_seq_ext[OF _ gets_sp], rename_tac cur_vcpu)
+  apply (case_tac cur_vcpu; clarsimp)
+   apply (rule hoare_seq_ext_skip, wpsimp)
+   apply (wpsimp simp: cur_vcpu_valid_def active_cur_vcpu_of_def)
+  apply (intro conjI impI)
```
Xaphiosis, line 154: from this line onwards, it looks like nothing interesting happens. is it possible to finish this up via normal weakest-precondition reasoning, or compress it a bit? Up until the case splitting, I can see the use of forward reasoning, but then this proof starts looking like it's hammering in a screw.

proof/invariant-abstract/ARM_HYP/ArchVCPU_AI.thy
```diff
@@ -0,0 +1,354 @@
...
+  apply (clarsimp simp: vcpu_switch_def)
+  apply (cases v; clarsimp)
+   apply (rule hoare_seq_ext[OF _ gets_sp], rename_tac cur_vcpu)
+   apply (case_tac cur_vcpu; clarsimp)
+    apply (wpsimp simp: cur_vcpu_valid_def active_cur_vcpu_of_def)
+   apply (clarsimp simp: when_def)
+   apply (intro conjI impI)
+    apply (rule hoare_seq_ext_skip, wpsimp)
+    apply (wpsimp simp: cur_vcpu_valid_def active_cur_vcpu_of_def)
+   apply (wpsimp simp: cur_vcpu_valid_def active_cur_vcpu_of_def)
```
Xaphiosis, line 149: same comment as below, something about this feels... very repetitive; it might be necessary, but looking at it again couldn't hurt. Hard to say without seeing the proof states, happy to be wrong here.

proof/invariant-abstract/ARM_HYP/ArchVCPU_AI.thy
```diff
@@ -0,0 +1,354 @@
...
+   \<lbrace>\<lambda>_. cur_vcpu_valid\<rbrace>"
+  apply (simp add: Schedule_A.schedule_def schedule_switch_thread_fastfail_def ethread_get_when_def
+                   ethread_get_def)
+  by (wpsimp wp: hoare_drop_imps gts_wp )
+
+lemma schedule_cur_vcpu_valid[wp]:
+  "\<lbrace>cur_vcpu_valid and valid_idle\<rbrace>
+   (schedule :: (unit,unit) s_monad)
+   \<lbrace>\<lambda>_. cur_vcpu_valid\<rbrace>"
+  apply (simp add: Schedule_A.schedule_def allActiveTCBs_def)
```
Xaphiosis, line 349: consider `unfolding` in these situations

spec/abstract/ARM_HYP/ArchVSpace_A.thy
```diff
@@ -612,6 +596,21 @@ where
...
+    case t_vcpu of Some p \<Rightarrow> dissociate_vcpu_tcb p t
+                 | _ \<Rightarrow> return ();
+    v \<leftarrow> get_vcpu vr;
+    case vcpu_tcb v of Some p \<Rightarrow> dissociate_vcpu_tcb vr p
+                     | _ \<Rightarrow> return ();
+    arch_thread_set (\<lambda>x. x \<lparr> tcb_vcpu := Some vr \<rparr>) t;
+    set_vcpu vr (v\<lparr> vcpu_tcb := Some t \<rparr>);
+    ct \<leftarrow> gets cur_thread;
+    when (t = ct) $ vcpu_switch (Some vr)
+  od"
```
Xaphiosis, line 612: thank you for fixing style

Xaphiosis:
The commit needs changing I think. If you have a complete feature, you can merge the complete feature into one commit with a full description. Here, you are making a change to the abstract spec, which is relevant to a bug we found and should get some mention in the description of the change. Then you have WIP on AInvs ... I would much rather see, in the end, the spec updates (aspec+haskell) squashed into one commit, and the proof updates into another. For work-in-progress, having finer granularity of commits is more flexible anyway.
Along the same lines, I was wondering "why are we doing this on rt? that doesn't make sense", until I realised it isn't on rt, only the commit says it's on rt. Possibly due to trauma from working on rt too long...

PR #439, opened 2022-03-01
Add `vcpu_switch` to `associate_vcpu_tcb`

michaelmcinerney:
Please see the commit messages, as well as https://sel4.atlassian.net/browse/SELFOUR-2947 and https://sel4.atlassian.net/browse/VER-770


proof/crefine/ARM_HYP/Finalise_C.thy
```diff
@@ -2179,29 +2179,38 @@ lemma associateVCPUTCB_ccorres:
...
             apply ceqv
-           apply (rule ccorres_return_Skip)
-          apply (rule wp_post_taut)
+           apply (rule ccorres_pre_getCurThread, rename_tac curThread)
+           apply (subst ccorres_seq_skip'[symmetric])
+           apply (rule ccorres_split_nothrow[where r'=dc and xf'=xfdc])
+               apply (rule_tac R="\<lambda>s. curThread = ksCurThread s" in ccorres_when)
+                apply (intro allI impI)
+                apply (unfold mem_simps)[1]
+                apply (clarsimp simp: rf_sr_ksCurThread)
```
Xaphiosis, line 2192: This is a bit odd, looks like a fight with clarsimp, and I'm not sure what `mem_simps` is exactly, but  it's an awkward-looking combination.

proof/crefine/ARM_HYP/Finalise_C.thy
```diff
@@ -2179,29 +2179,38 @@ lemma associateVCPUTCB_ccorres:
...
+               apply (ctac add: vcpu_switch_ccorres_Some)
+              apply ceqv
+             apply (rule ccorres_return_Skip)
+            apply (rule wp_post_taut)
+           apply (vcg exspec=vcpu_switch_modifies)
+          apply (rule_tac Q="\<lambda>_. pspace_aligned' and pspace_distinct' and valid_objs' and no_0_obj'
+                                 and valid_arch_state' and vcpu_at' vcpuptr"
+                 in hoare_post_imp, fastforce)
```
Xaphiosis, line 2200: indent `in` by two as it falls under `rule_tac`

Xaphiosis: also, there seems a fair bit of hoare_post_imp added in this proof; can you say why wp couldn't deal with it (with some assistance)?

proof/invariant-abstract/ARM_HYP/ArchArch_AI.thy
```diff
@@ -983,24 +987,179 @@ lemma valid_global_vspace_mappings_vcpu_update_str:
...
+  "do_machine_op (set_gic_vcpu_ctrl_vmcr w''') \<lbrace>valid_machine_state\<rbrace>"
+  "do_machine_op (get_gic_vcpu_ctrl_lr w'''') \<lbrace>valid_machine_state\<rbrace>"
+  "do_machine_op get_gic_vcpu_ctrl_apr \<lbrace>valid_machine_state\<rbrace>"
+  "do_machine_op get_gic_vcpu_ctrl_vmcr \<lbrace>valid_machine_state\<rbrace>"
+  "do_machine_op get_gic_vcpu_ctrl_hcr \<lbrace>valid_machine_state\<rbrace>"
+  unfolding valid_machine_state_def set_cntv_cval_64_def read_cntpct_def set_cntv_off_64_def
+            maskInterrupt_def setHCR_def set_gic_vcpu_ctrl_hcr_def set_gic_vcpu_ctrl_lr_def
+            set_gic_vcpu_ctrl_apr_def set_gic_vcpu_ctrl_vmcr_def get_gic_vcpu_ctrl_lr_def
+            get_gic_vcpu_ctrl_apr_def get_gic_vcpu_ctrl_vmcr_def get_gic_vcpu_ctrl_hcr_def
+  by (wpsimp wp: hoare_vcg_all_lift hoare_vcg_disj_lift dmo_machine_state_lift)+
```
Xaphiosis, line 1129: @corlewis : do you have any thoughts on crunch and doing this kind of thing? it seems rather repetitive and the `w''''` is indicative of that

proof/refine/ARM_HYP/Arch_R.thy
```diff
@@ -1306,6 +1306,24 @@ lemma archThreadSet_VCPU_Some_corres[corres]:
...
 
+crunches dissociateVCPUTCB
+  for no_0_obj'[wp]: no_0_obj'
+  and ksCurThread[wp]: "\<lambda>s. P (ksCurThread s)"
+  (simp: crunch_simps wp: crunch_wps)
+
+lemma vcpuSwitch_corres'':
+  assumes "vcpu' = vcpu"
+  shows
+  "corres dc (\<lambda>s. (vcpu \<noteq> None \<longrightarrow> vcpu_at  (the vcpu) s) \<and> valid_arch_state s)
```
Xaphiosis, line 1317: strange indent (at same level as `shows`) and not really needed assumes/shows... is this only moved?

proof/refine/ARM_HYP/Arch_R.thy
```diff
@@ -2284,18 +2308,14 @@ lemma assoc_invs':
...
-    f
-  od" by (simp add: bind_assoc)
+lemma associateVCPUTCB_helper_rewrite:
+  "do archThreadSet (atcbVCPUPtr_update (\<lambda>_. Just v)) tcb;
+      setObject v $ vcpuTCBPtr_update (\<lambda>_. Just tcb) vcpu;
+      f
+   od = do associateVCPUTCB_helper vcpu v tcb;
+           f
+        od"
+  by (simp add: associateVCPUTCB_helper_def bind_assoc)
```
Xaphiosis, line 2318: Something here I don't get: you're defining associateVCPUTCB_helper to be whatever you want it to be, and then you have this extra rule to fold it up that's just a bit different to your definition. Would it make more sense to define associateVCPUTCB_helper to match exactly what you need it to be and use `fold`?

Xaphiosis: Also, the `Just` is something that comes from Haskell, and should go away with a simp along with the `$`, so it's unusual seeing them here. I'd experiment below in `associateVCPUTCB_invs` with simplifying first before trying a `subst associateVCPUTCB_helper_rewrite[symmetric]`

proof/refine/ARM_HYP/Arch_R.thy
```diff
@@ -2246,16 +2266,20 @@ lemma archThreadSet_ex_nonz_cap_to'[wp]:
...
   "\<lbrace>invs' and
     ko_at' (vcpu\<lparr>vcpuTCBPtr:= None\<rparr>) v and
     obj_at' (\<lambda>tcb. atcbVCPUPtr (tcbArch tcb) = None) t and
     ex_nonz_cap_to' v and ex_nonz_cap_to' t\<rbrace>
-   do
-    archThreadSet (atcbVCPUPtr_update (\<lambda>_. Just v)) t;
-    setObject v $ vcpuTCBPtr_update (\<lambda>_. Just t) vcpu
-   od \<lbrace>\<lambda>_. invs'\<rbrace>"
-  unfolding invs'_def valid_state'_def valid_pspace'_def valid_mdb'_def
+   associateVCPUTCB_helper vcpu v t
```
Xaphiosis, line 2280: I'm really not clear how `associateVCPUTCB_helper` is helping compared to what we had before. Is there a reason you're doing it this way? Even if you want to force a forward proof, you could split off two statements with an assoc rule... here it looks like you're collapsing two statements into a def, making a rewrite rule for those two lines to fold them into the def, then proving a wp rule over a def where it used to be two lines, and then doing a symmetric subst for the def later on... hard to follow/understand why.

Xaphiosis: (where you = either actual you or whoever actually did this)

corlewis:
Is the first `repeat_unless` commit taken from the rt branch? I was going to suggest squashing the two commits together, but if it is I guess that makes it a bit more complicated.

Similarly, it would be nice to actually implement the FIXME now, but I can see how that would also be complicated with the rt branch. Would the nicest way to handle it require simultaneously fixing both master and rt?

proof/invariant-abstract/ARM_HYP/ArchArch_AI.thy
corlewis: Unfortunately, `crunch` currently wouldn't be any help here unless it turned out that you could prove `do_machine_op f ⦃valid_machine_state⦄`. There isn't a way to specify parameters for the functions being crunched.

michaelmcinerney: Let me know if I should experiment with another way of doing this

proof/refine/ARM_HYP/Arch_R.thy
michaelmcinerney: This is very similar to `vcpuSwitch_corres'`, and I copied that lemma and modified it here to much more easily apply with the `corressimp` in the `associateVCPUTCB_corres` proof. I will at least remove the `assumes/shows`

proof/refine/ARM_HYP/Arch_R.thy
michaelmcinerney: This was very annoying. I was able to easily do forward reasoning the whole way through, but then I would have been specifying preconditions that would have come out of the wp machinery... So I tried to follow the previous proof.

The problem is that this bundling (that now occurs in this helper function) is now not right towards the end of the function, now that `vcpuSwitch` is there. And the rewrite rule that was there before wouldn't apply nicely after doing some `wpsimp`ing to take care of the `vcpuSwitch`. 

I think I can get rid of the helper function and the rewrite rule with a forwards proof. Would that be the best way of going?

michaelmcinerney:
> Is the first `repeat_unless` commit taken from the rt branch? I was going to suggest squashing the two commits together, but if it is I guess that makes it a bit more complicated.

And yeah the `repeat_unless` commit is from the `rt` branch, too. I'm guessing it'd be easier to keep them separate (and make the changes Gerwin suggested on Mattermost to the commit message), but I can squash if that is best

corlewis:
> And yeah the `repeat_unless` commit is from the `rt` branch, too. I'm guessing it'd be easier to keep them separate (and make the changes Gerwin suggested on Mattermost to the commit message), but I can squash if that is best

Oh right, I was also going to suggest that in cases where the commit messages aren't changed, it would be helpful to mention any cherry-picked commits in the PR description.

proof/refine/ARM_HYP/Arch_R.thy
Xaphiosis: You can do something like what you're doing now, but with less repetition and more constraint to within the single proof you're using it in (unless you're using it in more proofs that I missed?). At the very least, the definition and rewrite (fold) are redundant, because you can already pick a definition that folds the way you want.
This might also be an opportunity to practice Isar-style proofs using proof-internal definitions.

proof/crefine/ARM_HYP/Finalise_C.thy
Xaphiosis: Oh man, that (indenting to first param) is actually not bad either. I support doing that unless you're really running low on horizontal space in the proof. @lsf37 @corlewis  : agree?

proof/crefine/ARM_HYP/Finalise_C.thy
Xaphiosis: Do those three applies then collapse into one that solves the subgoal?

michaelmcinerney: Yes!

proof/refine/ARM_HYP/Arch_R.thy
michaelmcinerney: I tried for hours and hours to get it working without the helper definition, but I think the associativity of bind was killing me: there didn't seem to be a nice way to have a function with many many binds, and then separate precisely the last two of them. Any ideas?

proof/crefine/ARM_HYP/Finalise_C.thy
corlewis: I think you've gotten to the core of the different viewpoints, I've definitely been considering the `in` as belonging to the `rule_tac`.

lsf37 (approved):
:+1: Happy with the new version.

proof/refine/ARM_HYP/Arch_R.thy
```diff
@@ -1306,6 +1306,20 @@ lemma archThreadSet_VCPU_Some_corres[corres]:
...
   done
 
+crunches dissociateVCPUTCB
+  for no_0_obj'[wp]: no_0_obj'
+  and ksCurThread[wp]: "\<lambda>s. P (ksCurThread s)"
+  (simp: crunch_simps wp: crunch_wps)
+
+lemma vcpuSwitch_corres'':
+  "vcpu' = vcpu
+   \<Longrightarrow> corres dc (\<lambda>s. (vcpu \<noteq> None \<longrightarrow> vcpu_at  (the vcpu) s) \<and> valid_arch_state s)
```
corlewis, line 1316: Feel free to leave this as is, but is our official style to have the arrow in the previous line?
```suggestion
  "vcpu' = vcpu \<Longrightarrow>
   corres dc (\<lambda>s. (vcpu \<noteq> None \<longrightarrow> vcpu_at  (the vcpu) s) \<and> valid_arch_state s)
```

corlewis (approved):
Looks good to me

proof/invariant-abstract/ARM_HYP/ArchArch_AI.thy
```diff
@@ -983,24 +987,179 @@ lemma valid_global_vspace_mappings_vcpu_update_str:
...
-   by (wp get_vcpu_wp arch_thread_get_wp weak_if_wp as_user_only_idle hoare_vcg_all_lift
-        | wp (once) hoare_drop_imps
-        | wpc
-        | clarsimp
-        | strengthen valid_arch_state_vcpu_update_str valid_global_refs_vcpu_update_str
-                     valid_global_vspace_mappings_vcpu_update_str valid_global_objs_vcpu_update_str
-        | simp add: associate_vcpu_tcb_def valid_obj_def[abs_def] valid_vcpu_def
-                    dissociate_vcpu_tcb_def vcpu_invalidate_active_def vcpu_disable_def
-        | simp add: obj_at_def)+
+  by (wpsimp simp: invs_def valid_state_def valid_pspace_def)
```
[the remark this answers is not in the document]
michaelmcinerney: This was only possible because I pulled apart `invs'` completely in the previous proofs

proof/refine/ARM_HYP/Arch_R.thy
michaelmcinerney: Ah yeah I think what you suggest is the official style. I can change if people would like

PR #440, opened 2022-03-03
`valid_cur_vcpu`

michaelmcinerney:
This adds the new invariant `valid_cur_vcpu` that says that the active current vcpu is the vcpu of the current thread.

### DIFF
--- a/proof/invariant-abstract/ARM_HYP/ArchArch_AI.thy
+++ b/proof/invariant-abstract/ARM_HYP/ArchArch_AI.thy
@@ -640,12 +640,11 @@ lemma max_index_upd_no_cap_to:
   apply (clarsimp simp:table_cap_ref_def)
   done
 
-
-lemma perform_asid_control_invocation_st_tcb_at:
-  "\<lbrace>st_tcb_at (P and (Not \<circ> inactive) and (Not \<circ> idle)) t
-    and ct_active and invs and valid_aci aci\<rbrace>
-    perform_asid_control_invocation aci
-  \<lbrace>\<lambda>y. st_tcb_at P t\<rbrace>"
+lemma perform_asid_control_invocation_pred_tcb_at:
+  "\<lbrace>\<lambda>s. pred_tcb_at proj Q t s \<and> st_tcb_at (P and (Not \<circ> inactive) and (Not \<circ> idle)) t s
+        \<and> ct_active s \<and> invs s \<and> valid_aci aci s\<rbrace>
+   perform_asid_control_invocation aci
+   \<lbrace>\<lambda>_. pred_tcb_at proj Q t\<rbrace>"
   supply
     is_aligned_neg_mask_eq[simp del]
     is_aligned_neg_mask_weaken[simp del]
@@ -675,8 +674,6 @@ lemma perform_asid_control_invocation_st_tcb_at:
       apply (simp add:page_bits_def)+
     apply (simp add:invs_valid_objs invs_psp_aligned)+
   apply (rule conjI)
-   apply (erule pred_tcb_weakenE, simp)
-  apply (rule conjI)
    apply (frule st_tcb_ex_cap)
      apply clarsimp
     apply (clarsimp split: Structures_A.thread_state.splits)
@@ -704,6 +701,14 @@ lemma perform_asid_control_invocation_st_tcb_at:
   apply (auto simp:page_bits_def detype_clear_um_independent)
   done
 
+lemma perform_asid_control_invocation_st_tcb_at:
+  "\<lbrace>st_tcb_at (P and (Not \<circ> inactive) and (Not \<circ> idle)) t
+    and ct_active and invs and valid_aci aci\<rbrace>
+   perform_asid_control_invocation aci
+   \<lbrace>\<lambda>_. st_tcb_at P t\<rbrace>"
+  apply (wpsimp wp: perform_asid_control_invocation_pred_tcb_at)
+  apply (fastforce simp: pred_tcb_at_def obj_at_def)
+  done
 
 lemma set_cap_idx_up_aligned_area:
   "\<lbrace>K (\<exists>idx. pcap = UntypedCap dev ptr pageBits idx) and cte_wp_at ((=) pcap) slot
--- a/proof/invariant-abstract/ARM_HYP/ArchVCPU_AI.thy
+++ b/proof/invariant-abstract/ARM_HYP/ArchVCPU_AI.thy
@@ -1,0 +1,662 @@
+(*
+ * Copyright 2022, UNSW (ABN 57 195 873 197)
+ *
+ * SPDX-License-Identifier: GPL-2.0-only
+ *)
+
+theory ArchVCPU_AI
+imports AInvs
+begin
+
+context Arch begin global_naming ARM_HYP (*FIXME: arch_split*)
+
+definition active_cur_vcpu_of :: "'z state \<Rightarrow> obj_ref option" where
+  "active_cur_vcpu_of s \<equiv> case arm_current_vcpu (arch_state s) of Some (vr, True) \<Rightarrow> Some vr
+                                                                 | _  \<Rightarrow> None"
+
+definition valid_cur_vcpu :: "'z::state_ext state \<Rightarrow> bool" where
+  "valid_cur_vcpu s \<equiv> arch_tcb_at (\<lambda>itcb. itcb_vcpu itcb = active_cur_vcpu_of s) (cur_thread s) s"
+
+lemma valid_cur_vcpu_lift_ct_Q:
+  assumes arch_tcb_of_cur_thread: "\<And>P. \<lbrace>\<lambda>s. arch_tcb_at P (cur_thread s) s \<and> Q s\<rbrace>
+                                        f \<lbrace>\<lambda>_ s. arch_tcb_at P (cur_thread s) s\<rbrace>"
+  and active_cur_vcpu_of: "\<And>P. \<lbrace>\<lambda>s. P (active_cur_vcpu_of s) \<and> Q s\<rbrace>
+                                f \<lbrace>\<lambda>_ s. P (active_cur_vcpu_of s)\<rbrace>"
+  shows "\<lbrace>\<lambda>s. valid_cur_vcpu s \<and> Q s\<rbrace> f \<lbrace>\<lambda>_. valid_cur_vcpu\<rbrace>"
+  unfolding valid_cur_vcpu_def valid_def
+  using use_valid[OF _ active_cur_vcpu_of] use_valid[OF _ arch_tcb_of_cur_thread]
+  by (fastforce simp: active_cur_vcpu_of_def)
+
+lemmas valid_cur_vcpu_lift_ct = valid_cur_vcpu_lift_ct_Q[where Q=\<top>, simplified]
+
+lemma valid_cur_vcpu_lift:
+  "\<lbrakk>\<And>P. f \<lbrace>\<lambda>s. P (active_cur_vcpu_of s)\<rbrace>; \<And>P t. f \<lbrace>\<lambda>s. arch_tcb_at P t s\<rbrace>;
+    \<And>P. f \<lbrace>\<lambda>s. P (cur_thread s)\<rbrace>\<rbrakk> \<Longrightarrow>
+   f \<lbrace>valid_cur_vcpu\<rbrace>"
+  apply (rule valid_cur_vcpu_lift_ct)
+   apply (rule_tac f=cur_thread in hoare_lift_Pf3)
+   apply fastforce+
+  done
+
+lemma valid_cur_vcpu_lift_weak:
+  "\<lbrakk>\<And>P. f \<lbrace>\<lambda>s. P (arch_state s)\<rbrace>; \<And>P t. f \<lbrace>\<lambda>s. arch_tcb_at P t s\<rbrace>;
+    \<And>P. f \<lbrace>\<lambda>s. P (cur_thread s)\<rbrace>\<rbrakk> \<Longrightarrow>
+   f \<lbrace>valid_cur_vcpu\<rbrace>"
+  apply (rule valid_cur_vcpu_lift_ct)
+   apply (rule_tac f=cur_thread in hoare_lift_Pf3)
+   apply fastforce+
+  apply (clarsimp simp: valid_cur_vcpu_def valid_def)
+  by (fastforce simp: active_cur_vcpu_of_def)
+
+lemma valid_cur_vcpu_lift_cur_thread_update:
+  assumes arch_tcb_at: "\<And>P. f \<lbrace>arch_tcb_at P t\<rbrace>"
+  and active_cur_vcpu_of: "\<And>P. f \<lbrace>\<lambda>s. P (active_cur_vcpu_of s)\<rbrace>"
+  shows "f \<lbrace>\<lambda>s. valid_cur_vcpu (s\<lparr>cur_thread := t\<rparr>)\<rbrace>"
+  unfolding valid_cur_vcpu_def valid_def
+  using use_valid[OF _ active_cur_vcpu_of] use_valid[OF _ arch_tcb_at]
+  by (fastforce simp: active_cur_vcpu_of_def)
+
+crunches as_user
+  for active_cur_vcpu_of[wp]: "\<lambda>s. P (active_cur_vcpu_of s)"
+  (wp: crunch_wps simp: active_cur_vcpu_of_def)
+
+lemma as_user_valid_cur_vcpu[wp]:
+  "as_user tptr f \<lbrace>valid_cur_vcpu\<rbrace>"
+  by (rule valid_cur_vcpu_lift; wpsimp wp: as_user_pred_tcb_at)
+
+lemma machine_state_update_active_cur_vcpu_of[simp]:
+  "P (active_cur_vcpu_of (s\<lparr>machine_state := ms\<rparr>)) = P (active_cur_vcpu_of s)"
+  by (fastforce simp: active_cur_vcpu_of_def)
+
+crunches do_machine_op
+  for valid_cur_vcpu[wp]: valid_cur_vcpu
+  and valid_cur_vcpu_cur_thread_update[wp]: "\<lambda>s. valid_cur_vcpu (s\<lparr>cur_thread := t\<rparr>)"
+  (wp: valid_cur_vcpu_lift_cur_thread_update valid_cur_vcpu_lift crunch_wps)
+
+lemma valid_cur_vcpu_vcpu_update[simp]:
+  "vcpu_at v s \<Longrightarrow> valid_cur_vcpu (s\<lparr>kheap := kheap s(v \<mapsto> ArchObj (VCPU vcpu))\<rparr>) = valid_cur_vcpu s"
+  by (clarsimp simp: valid_cur_vcpu_def active_cur_vcpu_of_def pred_tcb_at_def obj_at_def)
+
+crunches vcpu_save_reg, vcpu_write_reg, save_virt_timer, vgic_update, vcpu_disable
+  for valid_cur_vcpu[wp]: valid_cur_vcpu
+  (wp: set_vcpu_wp)
+
+lemma set_vcpu_arch_tcb_at_cur_thread[wp]:
+  "set_vcpu ptr vcpu \<lbrace>\<lambda>s. arch_tcb_at P (cur_thread s) s\<rbrace>"
+  apply (wpsimp wp: set_vcpu_wp get_vcpu_wp)
+  apply (clarsimp simp: pred_tcb_at_def obj_at_def)
+  done
+
+crunches vcpu_disable, vcpu_restore, vcpu_save, set_vm_root
+  for arch_tcb_at_cur_thread[wp]: "\<lambda>s. arch_tcb_at P (cur_thread s) s"
+  (wp: crunch_wps)
+
+crunches vcpu_update, do_machine_op, invalidate_hw_asid_entry, invalidate_asid
+  for active_cur_vcpu_of[wp]: "\<lambda>s. P (active_cur_vcpu_of s)"
+  (simp: active_cur_vcpu_of_def)
+
+lemma vcpu_save_reg_active_cur_vcpu_of[wp]:
+  "vcpu_save_reg vr reg \<lbrace>\<lambda>s. P (active_cur_vcpu_of s)\<rbrace>"
+  by (wpsimp simp: vcpu_save_reg_def)
+
+crunches vcpu_restore, do_machine_op, vcpu_save_reg, vgic_update, save_virt_timer,
+         vcpu_save_reg_range, vgic_update_lr, vcpu_enable, vcpu_save, set_vcpu
+  for valid_cur_vcpu[wp]: valid_cur_vcpu
+  (wp: valid_cur_vcpu_lift crunch_wps simp: active_cur_vcpu_of_def)
+
+lemma switch_vcpu_valid_cur_vcpu_cur_thread_update[wp]:
+  "\<lbrace>arch_tcb_at (\<lambda>itcb. itcb_vcpu itcb = v) t\<rbrace>
+   vcpu_switch v
+   \<lbrace>\<lambda>_ s. valid_cur_vcpu (s\<lparr>cur_thread := t\<rparr>)\<rbrace>"
+  unfolding vcpu_switch_def
+  apply (wpsimp simp: valid_cur_vcpu_def active_cur_vcpu_of_def)
+  by fastforce
+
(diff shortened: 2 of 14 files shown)


michaelmcinerney:
I stupidly imported a file generically that doesn't exist on arches other than ARM_HYP. But seeing that this PR is targetting the spec/proof update, I might wait until that PR is merged before significantly updating this. Otherwise, the AInvs content for ARM_HYP should still be good to review. Happy to mark this PR as a draft for now if that's preferable

lsf37:
Since the tests won't be working until this is sorted (so we won't merge accidentally), I think it's fine to leave it on non-draft for now. For reviewing it is good enough.

proof/invariant-abstract/ARM_HYP/ArchVCPU_AI.thy
```diff
@@ -0,0 +1,729 @@
+(*
+ * Copyright 2022, ???
```
michaelmcinerney, line 2: Still need to add the proper copyright. The only thing I took from Sidney and Mitch's work was the definition of `active_cur_vcpu_of`. Should this be only UNSW then?

lsf37: Yes, that is good.

proof/invariant-abstract/ARM_HYP/ArchVCPU_AI.thy
```diff
@@ -0,0 +1,727 @@
...
+   (call_kernel e) :: (unit, det_ext) s_monad
+   \<lbrace>\<lambda>_ . valid_cur_vcpu\<rbrace>"
+  unfolding call_kernel_def
+  apply (simp flip: bind_assoc)
+  by (wpsimp wp: handle_event_valid_cur_vcpu hoare_vcg_if_lift2 hoare_drop_imps
+      | strengthen invs_valid_idle)+
+
+end
+
+end
```
lsf37, line 727: please add line feed at last line

proof/invariant-abstract/ARM_HYP/ArchVCPU_AI.thy
```diff
@@ -0,0 +1,727 @@
...
+definition valid_cur_vcpu :: "'z::state_ext state \<Rightarrow> bool" where
+  "valid_cur_vcpu s \<equiv> arch_tcb_at (\<lambda>itcb. itcb_vcpu itcb = active_cur_vcpu_of s) (cur_thread s) s"
+
+lemma valid_cur_vcpu_lift_ct_Q:
+  assumes arch_tcb_of_cur_thread: "\<And>P. \<lbrace>\<lambda>s. arch_tcb_at P (cur_thread s) s \<and> Q s\<rbrace>
+                                        f \<lbrace>\<lambda>_ s. arch_tcb_at P (cur_thread s) s\<rbrace>"
+  and active_cur_vcpu_of: "\<And>P. \<lbrace>\<lambda>s. P (active_cur_vcpu_of s) \<and> Q s\<rbrace>
+                                f \<lbrace>\<lambda>_ s. P (active_cur_vcpu_of s)\<rbrace>"
+  shows "\<lbrace>\<lambda>s. valid_cur_vcpu s \<and> Q s\<rbrace> f \<lbrace>\<lambda>_. valid_cur_vcpu\<rbrace>"
+  apply (clarsimp simp: valid_cur_vcpu_def valid_def)
```
lsf37, line 26: could this `clarsimp` be an `unfolding ..`?

proof/invariant-abstract/ARM_HYP/ArchVCPU_AI.thy
```diff
@@ -0,0 +1,727 @@
...
+crunches vcpu_restore, do_machine_op, vcpu_save_reg, vgic_update, save_virt_timer,
+         vcpu_save_reg_range, vgic_update_lr, vcpu_enable, vcpu_save
+  for valid_cur_vcpu[wp]: valid_cur_vcpu
+  (wp: valid_cur_vcpu_lift crunch_wps simp: active_cur_vcpu_of_def)
+
+lemma switch_vcpu_valid_cur_vcpu_cur_thread_update[wp]:
+  "\<lbrace>arch_tcb_at (\<lambda>itcb. itcb_vcpu itcb = v) t\<rbrace>
+   vcpu_switch v
+   \<lbrace>\<lambda>_ s. valid_cur_vcpu (s\<lparr>cur_thread := t\<rparr>)\<rbrace>"
+  apply (clarsimp simp: vcpu_switch_def)
```
lsf37, line 113: this one also looks like it could be an `unfolding` (it's not super important, but it's nice when the unfolding step is stable and you know in maintenance that nothing interesting is supposed to be happening here).

proof/invariant-abstract/ARM_HYP/ArchVCPU_AI.thy
```diff
@@ -0,0 +1,727 @@
...
+
+lemma find_free_hw_asid_active_cur_vcpu_of[wp]:
+  "find_free_hw_asid \<lbrace>\<lambda>s. P (active_cur_vcpu_of s)\<rbrace>"
+  apply (clarsimp simp: find_free_hw_asid_def find_pd_for_asid_assert_def)
+  apply (intro hoare_seq_ext[OF _ gets_inv])
+  apply (clarsimp split: option.splits)
+  apply (rule hoare_seq_ext_skip, wpsimp)+
+   apply (clarsimp simp: active_cur_vcpu_of_def pred_tcb_at_def obj_at_def valid_cur_vcpu_def)
+  apply (wpsimp simp: find_pd_for_asid_assert_def)
+  done
```
lsf37, line 129:
```suggestion
lemma active_cur_vcpu_of_next_asid[simp]:
  "active_cur_vcpu_of (s⦇arch_state := arch_state s⦇arm_next_asid := v⦈⦈) = active_cur_vcpu_of s"
  by (clarsimp simp: active_cur_vcpu_of_def pred_tcb_at_def obj_at_def valid_cur_vcpu_def)

lemma find_free_hw_asid_active_cur_vcpu_of[wp]:
  "find_free_hw_asid ⦃λs. P (active_cur_vcpu_of s)⦄"
  by (wpsimp simp: find_free_hw_asid_def find_pd_for_asid_assert_def)
```

proof/invariant-abstract/ARM_HYP/ArchVCPU_AI.thy
```diff
@@ -0,0 +1,727 @@
...
+  apply (rule hoare_seq_ext[OF _ gets_the_sp])
+  apply (clarsimp simp: get_vcpu_def bind_assoc)
+  apply (rule hoare_seq_ext[OF _ get_object_sp])
+  apply (clarsimp split: kernel_object.splits arch_kernel_obj.splits)
+  apply (rule hoare_seq_ext[OF _ assert_sp])
+  apply (rule hoare_seq_ext[OF _ gets_sp])
+  apply (simp flip: bind_assoc)
+  apply (rule hoare_seq_ext)
+   apply wpsimp
+  apply (simp add: bind_assoc)
```
lsf37, line 187: This looks like too much explicit forward single-stepping to me for a proof like this. Those single steps are very brittle in maintenance, because they spell out each single step of the function in the proof, so even small reorderings/additions will break. They shouldn't really be necessary here. Of course, that's hard to say from just the diff. Some of that could be because of the inline unfolding of `vcpu_invalidate_active`, which probably needs its own lemma(s).

The lemma below `associate_vcpu_tcb_valid_cur_vcpu` has the same flavour.`wp`/`wpsimp` should be effective on these, but they probably need some hand holding in the form of lemmas.

proof/invariant-abstract/ARM_HYP/ArchVCPU_AI.thy
```diff
@@ -0,0 +1,727 @@
...
+  (wp: crunch_wps simp: crunch_simps)
+
+lemma send_fault_ipc_arch_tcb_at[wp]:
+  "send_fault_ipc tptr fault \<lbrace>arch_tcb_at P t\<rbrace>"
+  unfolding send_fault_ipc_def thread_set_def Let_def
+  apply (rule validE_valid)
+  apply (rule hoare_seq_ext_skipE, solves wpsimp)+
+  apply (wpsimp wp: set_object_wp)
+  apply (clarsimp simp: pred_tcb_at_def obj_at_def get_tcb_def)
+  done
```
lsf37, line 343: `wp` gets stuck at `lookup_cap` because we haven't provided any full `wp` rules for it, but we do have a rule for `arch_tcb_at` and all the implications are just accumulated information that is not actually necessary for that to be preserved. `hoare_drop_imps` will get rid of these implications automatically and `hoare_vcg_all_lift_R` will deal with the quantifiers that are introduced in the case distinction:
```suggestion
  by (wpsimp wp: set_object_wp hoare_drop_imps hoare_vcg_all_lift_R
             simp: pred_tcb_at_def obj_at_def get_tcb_def)
```

proof/invariant-abstract/ARM_HYP/ArchVCPU_AI.thy
```diff
@@ -0,0 +1,727 @@
...
+
+lemma do_reply_transfer_arch_tcb_at[wp]:
+  "do_reply_transfer sender receiver slot grant \<lbrace>arch_tcb_at P t\<rbrace>"
+  unfolding do_reply_transfer_def thread_set_def
+  apply (rule hoare_seq_ext_skip, solves wpsimp, clarsimp?)+
+  apply (rename_tac fault_opt)
+  apply (case_tac fault_opt; clarsimp)
+   apply wpsimp
+  apply (rule hoare_seq_ext_skip, solves wpsimp)+
+  apply (wpsimp wp: set_object_wp hoare_drop_imps | intro conjI impI)+
```
lsf37, line 362: The `intro conjI impI` is not a good pattern in combination with `wpsimp`, it's usually a sign that something has gone wrong. In this case it's splitting an `if` into a conjunction instead of progressing with `wp`.

Try 
`  apply (wpsimp wp: set_object_wp hoare_drop_imps split_del: split_if)`
instead (note no `+`, which `wpsimp` should never need for a single goal).

This specific thing (too eager `if_split`) is a problem with the `clarsimp` inside `wpsimp` being too eager. We could try to remove `if_split` in there by default, but we'd get lots of annoying warnings in contexts where `if_split` is already removed. Maybe we should put this one on the list of things to spend some time thinking about.

proof/invariant-abstract/ARM_HYP/ArchVCPU_AI.thy
```diff
@@ -0,0 +1,727 @@
...
+
+crunches reschedule_required, set_scheduler_action, tcb_sched_action
+  for arch_tcb_at[wp]: "arch_tcb_at P t"
+  (simp: reschedule_required_def set_scheduler_action_def tcb_sched_action_def set_tcb_queue_def
+         get_tcb_queue_def)
+
+lemma do_reply_transfer_arch_tcb_at[wp]:
+  "do_reply_transfer sender receiver slot grant \<lbrace>arch_tcb_at P t\<rbrace>"
+  unfolding do_reply_transfer_def thread_set_def
+  apply (rule hoare_seq_ext_skip, solves wpsimp, clarsimp?)+
```
lsf37, line 357: We should stop using this pattern in a single ad-hoc proof style. I was fine with it as emergency measure for some of the RT proofs, but it's not a good thing to use everywhere. Unless the part we're splitting off is itself some big case distinction or something like that, this should never be necessary. `wpsimp` alone will solve all of these parts automatically if the harder later part is already solved.

If we really want to do this, then we should make it a proof method that is carefully constructed and deals natively with e.g. the `K_bind` that things get stuck on here.

proof/invariant-abstract/ARM_HYP/ArchVCPU_AI.thy
```diff
@@ -0,0 +1,727 @@
...
+  for arch_tcb_at[wp]: "arch_tcb_at P t"
+  (wp: valid_cur_vcpu_lift mapM_x_wp_inv crunch_wps)
+
+crunches reschedule_required, set_scheduler_action, tcb_sched_action
+  for arch_tcb_at[wp]: "arch_tcb_at P t"
+  (simp: reschedule_required_def set_scheduler_action_def tcb_sched_action_def set_tcb_queue_def
+         get_tcb_queue_def)
+
+lemma do_reply_transfer_arch_tcb_at[wp]:
+  "do_reply_transfer sender receiver slot grant \<lbrace>arch_tcb_at P t\<rbrace>"
```
lsf37, line 355: So what this proof is getting stuck on is the unfolding (wp-rule use) of `set_object`. That is one level too low. In fact, unfolding `thread_set` is already one level too low, because `thread_set` with fault updates should always preserve `arch_tcb_at`. Once you have that, the lemma becomes almost entirely automatic:

```isabelle
lemma thread_set_fault_arch_tcb_at[wp]:
  "thread_set (tcb_fault_update f) r ⦃arch_tcb_at P t⦄"
  unfolding thread_set_def
  by (wpsimp wp: set_object_wp simp: pred_tcb_at_def obj_at_def get_tcb_def)

lemma do_reply_transfer_arch_tcb_at[wp]:
  "do_reply_transfer sender receiver slot grant ⦃arch_tcb_at P t⦄"
  unfolding do_reply_transfer_def
  by (wpsimp wp: gts_wp split_del: if_split)
```

proof/invariant-abstract/ARM_HYP/ArchVCPU_AI.thy
Xaphiosis: Dropping the `if_split` is one way, and `hoare_drop_imps` is useful when you don't need to the condition of the `if`. If you *do* need the condition of the `if`, consider adding the simp rule `if_apply_def2` which converts `if` to implications, then you can decide whether you want to drop them or not. I don't know if this applies here, but it's a very useful tool to know about.

lsf37: The `if` is in the code this time (somewhere deep in a term), so this specific rule wouldn't help here, but in general it's a good idea to check, yes.

proof/invariant-abstract/ARM_HYP/ArchVCPU_AI.thy
```diff
@@ -0,0 +1,727 @@
...
+    apply simp
+   apply simp
+  apply (rule hoare_seq_ext_skipE)
+   apply (case_tac buffer; clarsimp)
+    apply wpsimp
+   apply (rule hoare_seq_ext_skipE)
+    apply (wpsimp wp: cap_delete_valid_cur_vcpu)
+   apply (wpsimp wp: check_cap_inv hoare_drop_imps thread_set_hyp_refs_trivial thread_set_valid_cur_vcpu)
+  apply wpsimp
+  done
```
lsf37, line 638: This is one where forward stepping with invariant make sense, because with pure backwards wp reasoning would blow up over the case distinctions. No need to do all those case distinctions manually, though, `wpsimp` can do them on its own. In fact, you can nicely collapse the entire proof into:

```suggestion
  \<comment> \<open>ThreadControl\<close>
  apply (rule validE_valid)
  apply (rule hoare_seq_ext_skipE, solves ‹wpsimp wp: check_cap_inv›)+
  by (wpsimp wp: check_cap_inv hoare_drop_imps thread_set_hyp_refs_trivial thread_set_valid_cur_vcpu)
```

lsf37: Even for this, though, we should ideally use a proof method that does the skipping step, reliably handles special cases, and works for all variants of `valid`, `validE`, etc.

Not sure it makes sense to put that into this PR, but it's not good to let the assembly-level proof steps proliferate too much, because they'll be hard to remove later if we wait too long.

lsf37 (approved):
Overall I'm happy with this, my current pet peeve is the forward stepping with `hoare_seq_ext_skip` :-) This is usually too manual and hard to maintain.

There is only one case that I can see where it really is beneficial, otherwise they are usually a sign of something having gone wrong in the proofs with `wp`, either a missing rule or unfolding too deeply which leads to the same effect (not hitting the spot where the automation works).

That one case where the forward stepping is better is when `wp` would otherwise lead to exponentially many cases. You get that when you have a chain of case distinctions -- the forward stepping joins the two case branches every time, so doesn't blow up. You can't do that easily with `wp`.

For those cases, and maybe for proof exploration, we should wrap up the forward stepping into a proof method. I'm fine with doing that separately, as long as we don't leave too long.

michaelmcinerney:
Thanks for the review, Gerwin. I've hopefully addressed all your comments.

I've included an Eisbach method that hopefully achieves what we would be after. I didn't want the method to apply `rule valid_validE` unless `rule validE_valid` could be applied, so that the method wouldn't mess with the proof state any more than necessary. If we aren't so worried about that, the method could be made slightly less repetitive:

```
method hoare_skip uses wp simp =
  (rule validE_valid)?,
  (rule hoare_seq_ext_skip | rule hoare_seq_ext_skipE),
  solves <wpsimp wp: wp simp: simp>,
  (rule valid_validE)?,
  clarsimp?
```

lsf37:
> Thanks for the review, Gerwin. I've hopefully addressed all your comments.

:+1: Looks good!

> I've included an Eisbach method that hopefully achieves what we would be after. I didn't want the method to apply `rule valid_validE` unless `rule validE_valid` could be applied, so that the method wouldn't mess with the proof state any more than necessary. If we aren't so worried about that, the method could be made slightly less repetitive:
> 
> ```
> method hoare_skip uses wp simp =
>   (rule validE_valid)?,
>   (rule hoare_seq_ext_skip | rule hoare_seq_ext_skipE),
>   solves <wpsimp wp: wp simp: simp>,
>   (rule valid_validE)?,
>   clarsimp?
> ```

The proof already looks a lot nicer, but the method still needs a bit more refinement:

 - we shouldn't end in `clarsimp?`, because that can strongly affect the remaining goal (incl the monadic code) instead of just splitting things off. If it is for resolving `K_bind`, we should use/make a specific rule for it (`hoare_seq_ext_nobind` looks like the one to try before `hoare_seq_ext` to avoid that particular issue)

 - instead of the validE -> skip -> validE conversion, we should make rules that do this in one step, i.e. produce rules out of `hoare_seq_ext_nobind` that are already properly chained together so only combinations that make sense are possible. 

 - we should complete the rule set so it works for `validE_R` and `validR_R`

 - naming: the hardest part :-). "skip" is what we're used to, but it's not a good name. We're not really skipping anything.  Maybe `inv_step`? Doesn't quite express that it works from the front when everything else works from the back. Is `forward_inv_step` too long?

- taken all of the above together, we should probably make a named rule set for these. The order of rules in that set is important (e.g. so the ones with K_bind are tried first). With a rule set the method can then become just

       rule forward_inv_step_rules, solves <wpsimp wp: wp simp: simp>

   That would have the added advantage that if the `wpsimp` is not quite strong enough on its own, you can do a separate `apply (rule forward_inv_step_rules)` in the proof, and it should already take care of all the special cases, just leaving you with the two main wp goals to be solved (not to be abused, though :-))

michaelmcinerney:
Those are all very good points. I really like the idea of a named rule set. I'll look into that

michaelmcinerney:
I think I have a good rule set for the `forward_inv_step_rules` set. Where should the Eisbach method go? Immediately after the rules? I think that makes most sense. Or in the `Eisbach_Methods.thy` file?

lsf37:
> I think I have a good rule set for the `forward_inv_step_rules` set. Where should the Eisbach method go? Immediately after the rules? I think that makes most sense. Or in the `Eisbach_Methods.thy` file?

The Eisbach_Methods file is for more general ones, this one would fit somewhere in the NondetMonadVCG space I think, close to where the rules are available.

lsf37:
> > I think I have a good rule set for the `forward_inv_step_rules` set. Where should the Eisbach method go? Immediately after the rules? I think that makes most sense. Or in the `Eisbach_Methods.thy` file?
> 
> The Eisbach_Methods file is for more general ones, this one would fit somewhere in the NondetMonadVCG space I think, close to where the rules are available.

Thinking more about it, it might be worth putting those rules and the method into their own theory file, actually (and including that somewhere transparently so that we don't need to change other imports).

michaelmcinerney:
> Thinking more about it, it might be worth putting those rules and the method into their own theory file, actually (and including that somewhere transparently so that we don't need to change other imports).

If this new file imports `NonDetMonadVCG`, then I'm not sure how I would go about not changing other imports. `NonDetMonadVCG` is imported a dozen or so times, so I'm guessing we wouldn't want to replace all of those imports with importing this new file. Or maybe if I import the new file in `Include_AI`? Then it would at least be available everywhere in AInvs, Refine, CRefine

lsf37:
> If this new file imports `NonDetMonadVCG`, then I'm not sure how I would go about not changing other imports. `NonDetMonadVCG` is imported a dozen or so times, so I'm guessing we wouldn't want to replace all of those imports with importing this new file. 

Yes, that'd be too much. But we could include the new file in `NonDetMonadVCG`, provided the lemmas that are needed in the new file can be moved into the new file without dragging too many other things with them.

michaelmcinerney:
> Yes, that'd be too much. But we could include the new file in `NonDetMonadVCG`, provided the lemmas that are needed in the new file can be moved into the new file without dragging too many other things with them.

The new rules for the new file require `hoare_seq_ext` and the `validE` and "nobind" versions, all of which are in `NonDetMonadVCG`. We wouldn't want to move those to the new file, would we?

lsf37:
Cool, new version looks good to me!

proof/invariant-abstract/ARM_HYP/ArchVCPU_AI.thy
corlewis: This feels like a specific enough guideline that it would be worth recording. Would [`docs/style`](https://github.com/seL4/l4v/blob/master/docs/Style.thy#L251) be the right place to put it?

proof/invariant-abstract/ARM_HYP/ArchArch_AI.thy
```diff
@@ -640,12 +640,11 @@ lemma max_index_upd_no_cap_to:
...
-lemma perform_asid_control_invocation_st_tcb_at:
-  "\<lbrace>st_tcb_at (P and (Not \<circ> inactive) and (Not \<circ> idle)) t
-    and ct_active and invs and valid_aci aci\<rbrace>
-    perform_asid_control_invocation aci
-  \<lbrace>\<lambda>y. st_tcb_at P t\<rbrace>"
+lemma perform_asid_control_invocation_pred_tcb_at:
+  "\<lbrace>\<lambda>s. pred_tcb_at proj Q t s \<and> st_tcb_at (P and (Not \<circ> inactive) and (Not \<circ> idle)) t s
+        \<and> ct_active s \<and> invs s \<and> valid_aci aci s\<rbrace>
+   perform_asid_control_invocation aci
+   \<lbrace>\<lambda>_. pred_tcb_at proj Q t\<rbrace>"
```
corlewis, line 647: Apologies for commenting on this way too late, but I just noticed that the `P` in this precondition no longer seems necessary. Leaving it in like this shouldn't break anything but it might lead to confusing schematic variables when this rule is used

proof/invariant-abstract/Untyped_AI.thy
```diff
@@ -3868,26 +3868,33 @@ lemmas invoke_untyp_invs[wp] =
...
-lemma invoke_untyped_st_tcb_at[wp]:
-  "\<lbrace>invs and st_tcb_at (P and (Not \<circ> inactive) and (Not \<circ> idle)) t
-         and ct_active and valid_untyped_inv ui\<rbrace>
-     invoke_untyped ui
-   \<lbrace>\<lambda>rv. \<lambda>s :: 'state_ext state. st_tcb_at P t s\<rbrace>"
+lemma invoke_untyped_pred_tcb_at:
+  "\<lbrace>\<lambda>s. pred_tcb_at proj Q t s \<and> invs s \<and> st_tcb_at (P and (Not \<circ> inactive) and (Not \<circ> idle)) t s
+        \<and> ct_active s \<and> valid_untyped_inv ui s\<rbrace>
+   invoke_untyped ui
+   \<lbrace>\<lambda>_ s :: 'state_ext state. pred_tcb_at proj Q t s\<rbrace>"
```
corlewis, line 3875: Same with the `P` in this lemma.

proof/invariant-abstract/ARM_HYP/ArchVCPU_AI.thy
lsf37: Yes, that would be a good place

proof/invariant-abstract/ARM_HYP/ArchArch_AI.thy
lsf37: Ah yes, you're right, I missed that. We should fix this in a new PR. I don't think it'll break much (or anything), but it's better to not have to worry about unused schematics popping up.

PR #441, opened 2022-03-13
rt run_tests: make sure L4V_FEATURES is set

lsf37:
If L4V_FEATURES is not already set to something specific, set it to
MCS so we don't accidentally run the wrong configuration.

### DIFF
--- a/run_tests
+++ b/run_tests
@@ -14,6 +14,10 @@ import subprocess
 L4V_ARCH_DEFAULT="ARM"
 L4V_ARCH_LIST=["ARM","ARM_HYP","X64","RISCV64"]
 
+# Set MCS for this branch by default
+L4V_FEATURES=os.environ.get("L4V_FEATURES", "MCS")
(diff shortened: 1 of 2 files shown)


mbrcknl (approved):
Would it make sense to also set it in `spec/cspec/c/Makefile`, for strange people like myself who like to build using `make`?

lsf37:
> Would it make sense to also set it in `spec/cspec/c/Makefile`, for strange people like myself who like to build using `make`?

Hm, we should set it only once, but maybe `spec/cspec/c/Makefile` is the better place because it will catch both uses. Happy to change it. @Xaphiosis do you have an opinion on this? (It turns out CI has been testing the wrong config all this time)

Xaphiosis:
I think the Makefile might be a better place for this; there aren't any scenarios I can think of on `rt` where we might want to run non-MCS either. It'll make for fewer `make` surprises (I use `make` a fair bit as well).

PR #442, opened 2022-03-13
ci proof-deploy: save C graph-lang

mbrcknl:
In the l4v proof-deploy workflow, export the C graph-lang produced by SimplExportAndRefine. In the near future, this will feed into binary verification. Also add a new job matrix to produce C graph-lang for MCS configurations supported by binary verification.

### DIFF
--- a/.github/workflows/proof-deploy.yml
+++ b/.github/workflows/proof-deploy.yml
@@ -48,6 +48,12 @@ jobs:
         AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
         AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
         AWS_SSH: ${{ secrets.AWS_SSH }}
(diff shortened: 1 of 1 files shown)


lsf37 (approved):
Looks good from my side.

lsf37:
The only slight inconsistency is that the rt manifest doesn't necessarily point to the same seL4 version as the master/devel manifest. This something we should fix separately, though. It's time for the rt manifest to track the C kernel more closely.

mbrcknl:
> The only slight inconsistency is that the rt manifest doesn't necessarily point to the same seL4 version as the master/devel manifest. This something we should fix separately, though. It's time for the rt manifest to track the C kernel more closely.

That's a good point. I guess a better way to handle this would be to have a second trigger for BV coming from the MCS proof workflow. But as you suggest, I might go ahead with this approach for now. This way, I can focus on a single trigger, but still get BV running on MCS configs. Once that's in place, I can come back and move things over to the MCS workflow.

PR #443, opened 2022-03-21
ci: add binary verification workflow

mbrcknl:
This adds a workflow that prepares the inputs needed by graph-refine. It is triggered at the end of a proof-deploy workflow. It collects C graph-lang artifacts from the proof-deploy workflow, and runs the decompiler to generate ASM graph-lang. Finally, it submits a job to a graph-refine back end over SSH.

Depends on https://github.com/seL4/ci-actions/pull/204.

### DIFF
--- a/.github/workflows/binary.yml
+++ b/.github/workflows/binary.yml
@@ -1,0 +1,320 @@
+# Copyright 2022 Kry10 Limited
+#
+# SPDX-License-Identifier: BSD-2-Clause
+
+name: Prepare binary verification
+
+on:
+  repository_dispatch:
+    types:
+      - binary-verification
+  workflow_dispatch:
+    inputs:
+      repo:
+        description: 'Repository'
+        required: true
+        default: 'seL4/l4v'
+      run_id:
+        description: 'Workflow run ID'
+        required: true
+
+jobs:
+  artifacts:
+    name: Initialise artifacts
+    runs-on: ubuntu-latest
+    outputs:
+      enabled_configs: ${{ steps.prepare.outputs.enabled_configs }}
+      sel4_commit: ${{ steps.prepare.outputs.sel4_commit }}
+      graph_refine_commit: ${{ steps.prepare.outputs.graph_refine_commit }}
+      isabelle_commit: ${{ steps.prepare.outputs.isabelle_commit }}
+    steps:
+      - name: Retrieve artifacts
+        id: retrieve
+        uses: actions/github-script@v6
+        with:
+          github-token: ${{ secrets.PRIV_REPO_TOKEN }}
+          script: |
+            const inputs = (function() {
+              if ("${{ github.event_name }}" === "repository_dispatch") {
+                return { repo: "${{ github.event.client_payload.repo }}",
+                         run_id: "${{ github.event.client_payload.run_id }}" };
+              } else {
+                return { repo: "${{ github.event.inputs.repo }}",
+                         run_id: "${{ github.event.inputs.run_id }}" };
+              }
+            })();
+            console.log(`Triggered by ${inputs.repo} run_id ${inputs.run_id}`)
+            console.log(`::set-output name=trigger_repo::${inputs.repo}`)
+            console.log(`::set-output name=trigger_run_id::${inputs.run_id}`)
+            const repo_parts = inputs.repo.split("/");
+            const artifacts = await github.rest.actions.listWorkflowRunArtifacts({
+               owner: repo_parts[0],
+               repo: repo_parts[1],
+               run_id: inputs.run_id,
+            });
+            const bv_artifacts = artifacts.data.artifacts.filter((artifact) => {
+              const name = artifact.name;
+              return name == "manifest" || name == "c-graph-lang";
+            });
+            const fs = require('fs/promises');
+            const files = bv_artifacts.map(async function(artifact) {
+              let download = await github.rest.actions.downloadArtifact({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                artifact_id: artifact.id,
+                archive_format: 'zip',
+              });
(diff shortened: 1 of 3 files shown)


mbrcknl:
Example test run: https://github.com/seL4/l4v-bv-ci/runs/5620776603?check_suite_focus=true

.github/workflows/binary.yml
```diff
@@ -0,0 +1,290 @@
...
+        # binary verificaation can't handle.
+        working-directory: seL4
+        run: |
+          # Add compile options to gcc.cmake
+          ( echo
+            echo "# Binary verification cannot handle cloned functions."
+            echo "if(KernelVerificationBuild)"
+            echo "    add_compile_options(-fno-partial-inlining -fno-ipa-cp -fno-ipa-sra)"
+            echo "endif()"
+          ) >> gcc.cmake
```
lsf37, line 230: Do we already have an issue for this in Jira? If not, can you please add a GitHub issue on seL4. We should measure the performance impact and either switch these off globally or, if there is significant impact provide a config switch for it, so people can easily select the config that BV can support.

.github/workflows/binary.yml
```diff
@@ -0,0 +1,290 @@
...
+        if: steps.enabled.outputs.config
+        uses: cachix/install-nix-action@v16
+        with:
+          nix_path: nixpkgs=channel:nixos-unstable
+      - name: Install Cachix
+        if: steps.enabled.outputs.config
+        uses: cachix/cachix-action@v10
+        with:
+          name: sel4-bv
+          authToken: ${{ secrets.BV_CACHIX_AUTH_TOKEN }}
```
lsf37, line 217: What does this part do?

.github/workflows/binary.yml
```diff
@@ -0,0 +1,290 @@
...
+          ( echo
+            echo "# Binary verification cannot handle cloned functions."
+            echo "if(KernelVerificationBuild)"
+            echo "    add_compile_options(-fno-partial-inlining -fno-ipa-cp -fno-ipa-sra)"
+            echo "endif()"
+          ) >> gcc.cmake
+      - name: Build target
+        if: steps.enabled.outputs.config
+        working-directory: graph-refine
+        shell: nix-shell --run "bash -eo pipefail {0}"
```
lsf37, line 234: I guess this is where `nix` comes in, but it's still unclear to me what exactly is going on there.

lsf37 (approved):
Overall I'm happy with this. Can you expand on the `nix` connection a bit?

.github/workflows/binary.yml
```diff
@@ -0,0 +1,290 @@
...
+on:
+  repository_dispatch:
+    types:
+      - binary-verification
+  workflow_dispatch:
+    inputs:
+      repo:
+        description: 'Repository'
+        required: true
+        default: 'seL4/l4v-bv-ci'
```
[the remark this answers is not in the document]
mbrcknl: Yes, thanks for noticing!

.github/workflows/binary.yml
mbrcknl: This was always going to be the most controversial bit. I should have added some notes in the commit message, and also in some comments in the workflow. Here's some explanation for a start...

`install-nix-action` installs the [Nix package manager](https://nixos.org/manual/nix/stable/), which provides various `nix-*` commands, and sets it up to use the [nixpkgs collection of packages](https://nixos.org/manual/nixpkgs/stable/). The only command I'm using in this workflow is `nix-shell`, which creates a sub-shell with the environment specified by [`shell.nix`](https://github.com/seL4/graph-refine/blob/ci-riscv64/shell.nix) in the working directory. The environment created by that Nix expression provides the Python modules, cross compilers and other tools the workflow script needs. `nix-shell` will download the packages it needs to set up that environment, building if necessary.

The motivation is to have a declarative specification of the required environment, and a tool to quickly get a script running in that environment. A Dockerfile would serve the same purpose, but Nix is more declarative and makes it easier to ensure build reproducibility. That makes caching work better and speeds up development iteration. Nix also has a very nice way to turn any set of packages [into a Docker image](https://github.com/seL4/graph-refine/blob/ci-riscv64/decompiler/default.nix#L81-L85). I used that to make Docker images for the decompiler and graph-refine. I would have liked to make a Docker image for everything needed here, but ran out of time.

`cachix-action` is optional. It adds a cache for any packages that `nix-shell` needs to build, so it doesn't need to build the same package more than once. The cache is provided by cachix.org, currently running on a 5GB free account, which is plenty big enough for these workflows.

.github/workflows/binary.yml
mbrcknl: This is running the `run` script under a sub-shell which provides the tools needed by the `run` script. `nix-shell` sets up the environment according to the [`shell.nix`](https://github.com/seL4/graph-refine/blob/ci-riscv64/shell.nix) in the working directory.

.github/workflows/binary.yml
mbrcknl: Issue logged here: https://github.com/seL4/seL4/issues/805

.github/workflows/binary.yml
mbrcknl: I added some comments in [a fixup commit](https://github.com/seL4/l4v/pull/443/commits/a432632b2bc558e62e16e545b9082b7acbdaca14).

lsf37: Ok, thanks for those. You're right, this is a bit controversial, because it introduces yet another way of specifying dependencies. I think it's fine for development at the moment and getting things working, but longer term we should unify these.

Good to merge for now, though.

mbrcknl: I'll leave it open for a day or two in case it needs further tweaks. I'll ping you if there's anything significant.

PR #448, opened 2022-03-29
github: changed AutoCorresSEL4 default

lsf37:
`ci-actions/aws-proofs` no longer excludes the AutoCorresSEL4 session by default, so we no longer need to provide a fake argument to the session parameter to not exclude it.

This is significant, because we now want the default to be non-verbose since we're running multiple sessions in parallel.

### DIFF
--- a/.github/workflows/external.yml
+++ b/.github/workflows/external.yml
@@ -24,7 +24,6 @@ jobs:
       with:
         L4V_ARCH: ${{ matrix.arch }}
         manifest: default.xml
-        session: '-v'  # non-empty argument to `./run-tests` to not exclude AutoCorresSEL4
       env:
         AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
         AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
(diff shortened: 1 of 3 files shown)


lsf37:
(killed the proof sessions, because `pull_request_target` will still use the old workflows from `master`, not the ones from the PR)

lsf37:
This should be the last PR in the default-switch series.

mbrcknl (approved):
Interesting. It didn't occur to me that `-j 2` would mean we won't see detailed logs until the workflow is finished. But I guess that's a small price to pay for the proofs to be able to finish at all. :)

lsf37:
> Interesting. It didn't occur to me that `-j 2` would mean we won't see detailed logs until the workflow is finished. But I guess that's a small price to pay for the proofs to be able to finish at all. :)

I was initially very keen on these detailed log, but it turns they are not as useful as I thought, because you only get to see what is being produced while you are on the page. I.e. to really see what is going on, you basically have to have a browser window open on that run the entire time. For a short job that's fine, but for the proofs that is fairly annoying.

mbrcknl:
> I was initially very keen on these detailed log, but it turns they are not as useful as I thought, because you only get to see what is being produced while you are on the page. I.e. to really see what is going on, you basically have to have a browser window open on that run the entire time. For a short job that's fine, but for the proofs that is fairly annoying.

Yeah, that's a good point. I found it useful for decompilation workflows. Since the decompiler constantly produces output, you can pop into a job to see how far along it is. But with the way Isabelle looks ahead in batch mode, that won't work as well for these proofs.

PR #455, opened 2022-04-11
MCS RISCV Refine

michaelmcinerney:


proof/refine/RISCV64/TcbAcc_R.thy
```diff
@@ -1659,7 +1666,7 @@ lemma asUser_sch_act_wf[wp]:
 lemma asUser_idle'[wp]:
   "\<lbrace>valid_idle'\<rbrace> asUser t m \<lbrace>\<lambda>rv. valid_idle'\<rbrace>"
   apply (simp add: asUser_def split_def)
-  apply (wpsimp wp: threadSet_idle' select_f_inv)
+  apply_trace (wpsimp wp: threadSet_idle' select_f_inv)
```
michaelmcinerney, line 1669: Remove the `_trace`

proof/refine/RISCV64/RAB_FN.thy
```diff
@@ -78,7 +78,7 @@ declare resolveAddressBitsFn.simps[simp del]
 
 lemma isCNodeCap_capUntypedPtr_capCNodePtr:
   "isCNodeCap c \<Longrightarrow> capUntypedPtr c = capCNodePtr c"
-  by (clarsimp simp: isCap_simps)
+  by (clarsimp simp: isCap_simps State_H.RISCV64_H.PPtr_def)
```
lsf37, line 81:
```suggestion
  by (clarsimp simp: isCap_simps RISCV64_H.PPtr_def)
```

Do you remember if the `RISCV64_H` is really necessary here?

proof/refine/RISCV64/Untyped_R.thy
```diff
@@ -4324,9 +4324,8 @@ lemma resetUntypedCap_corres:
               apply (subst is_aligned_weaken[OF is_aligned_mult_triv2])
                apply (simp add: Kernel_Config.resetChunkBits_def minUntypedSizeBits_def)
               apply (subst unat_mult_simple)
-               apply (clarsimp)
                apply (rule order_less_trans[rotated],
-                      rule_tac n=sz in power_strict_increasing, simp+)
+                      rule_tac n=sz in power_strict_increasing, (simp add: word_bits_def)+)
```
lsf37, line 4328: Does `;` work here instead of `, (..)+`? (The latter is an older idiom from before `;` -- generally if `;` works it is nicer, because it won't spill out into other subgoals)

lsf37:
Still in the middle of reviewing, but before I forget: since this PR basically closes the work on Refine for now, I think it is time to remove the quick-and-dirty settings for AInvs and Refine and switch the test to "real" proof mode. This would be accomplished by removing the following 3 lines in `run_tests`

```python
# For the time being, we allow sorries in invariant proofs:
os.environ["AINVS_QUICK_AND_DIRTY"]="1"
os.environ["REFINE_QUICK_AND_DIRTY"]="1"
print("Testing with QUICK_AND_DIRTY for AInvs and Refine")
```

This also removes the annoying re-build issue between images made from `run_tests` and images made from a plain `make`, because the settings are then the same.

proof/refine/RISCV64/ADT_H.thy
```diff
@@ -7,9 +7,16 @@
 chapter \<open>Abstract datatype for the executable specification\<close>
 
 theory ADT_H
-  imports Syscall_R
+imports
+  "AInvs.ADT_AI"
```
lsf37, line 11: Is the new import necessary for something specific or did this just come in from applying the patch?

proof/refine/RISCV64/ADT_H.thy
```diff
@@ -405,15 +444,28 @@ lemma distinct_word_add_inj_ptes:
...
+context
+begin
+
+private method ako =
+  find_goal \<open>match premises in "kheap s p = Some (ArchObj ako)" for s p ako \<Rightarrow> succeed\<close>,
+  (rename_tac ako, case_tac ako;
+   clarsimp simp: other_obj_relation_def pte_relation_def split: if_split_asm)
+
+private method ko =
+  case_tac ko; clarsimp simp: other_obj_relation_def cte_relation_def split: if_split_asm
```
lsf37, line 456: Nice, this might be worth back-porting to master for AArch64 when we get to Refine there.

proof/refine/RISCV64/Finalise_R.thy
```diff
@@ -2206,11 +2196,6 @@ definition
   finaliseCapTrue_standin_simple_def:
   "finaliseCapTrue_standin cap fin \<equiv> finaliseCap cap fin True"
 
-context
-begin
-
-declare if_cong [cong]
```
lsf37, line 2212: Thanks for removing that. This does not actually achieve anyway what it looks like it wants to do: even in a local context the `declare` remains global. 

Just for to reinforce the right pattern, to remove something locally, it'd be:

```
context
  notes if_cong[cong]
begin
...
end
```

proof/refine/RISCV64/RAB_FN.thy
michaelmcinerney: Turns out the `RISCV64_H` is necessary, but I've removed the `State_H`

proof/refine/RISCV64/ADT_H.thy
michaelmcinerney: This can be removed!

proof/refine/RISCV64/IpcCancel_R.thy
```diff
@@ -78,261 +121,305 @@ locale delete_one_conc_pre =
...
 end
 
-crunch tcb_at'[wp]: cancelSignal "tcb_at' t"
-  (wp: crunch_wps simp: crunch_simps)
+sublocale delete_one_conc_pre < delete_one: typ_at_all_props' "cteDeleteOne slot"
+  by typ_at_props'
 
-context delete_one_conc_pre
-begin
+declare if_weak_cong [cong]
```
lsf37, line 149: This is one of these problematic ones, probably auto-inserted from the merge. `if_weak_cong` is the default, so in a sense this should be harmless, but it means that the context expects `if_cong` to be set as the cong rule for if at this point. If this line is producing a warning that it's duplicate, then it should be removed. If the line does not produce a warning, we should eventually eliminate the global declare further up in the tree somewhere that makes this one necessary. But not in this PR. For now, if we can move it up to the top of this theory and add a FIXME, that would be good already. If proofs break because of the move, you might have to add a local declaration for `if_cong[cong]`

proof/refine/RISCV64/IpcCancel_R.thy
```diff
@@ -2036,108 +3098,189 @@ proof -
...
-         \<or> (ep_ptr, TCBBlockedSend) \<in> state_refs_of' s t"
-    apply -
-    apply (frule(1) sym_refs_ko_atD')
-    apply (drule ko_at_state_refs_ofD')
-    apply (case_tac r)
-      apply (clarsimp simp: st_tcb_at_refs_of_rev'
-             | drule(1) bspec | drule st_tcb_at_state_refs_ofD')+
-    done
+  proof
+    assume t_epQ: "t \<in> set (epQueue r)"
```
lsf37, line 3250: The indent here looks off, `proof` should be on the same indent as the corresponding statement (lemma here), i.e. 0. Auto-indent should fix this automatically.

proof/refine/RISCV64/IpcCancel_R.thy
```diff
@@ -2154,35 +3297,36 @@ lemma not_in_ntfnQueue:
...
-
-  with ko_at have "st_tcb_at' (Not \<circ> simple') t s"
-    apply -
-    apply (drule state_refs_of'_elemD)
-    apply (simp add: st_tcb_at_refs_of_rev')
-    apply (erule pred_tcb'_weakenE)
-    apply (clarsimp)
-    done
+  proof
+    assume t_epQ: "t \<in> set (ntfnQueue (ntfnObj r))"
```
lsf37, line 3301: same as above (`proof` should be at indent 0)

proof/refine/RISCV64/Ipc_R.thy
```diff
@@ -3107,316 +4041,1056 @@ lemma replyFromKernel_corres:
...
+   apply (fastforce simp: valid_obj'_def valid_sched_context_size'_def objBits_def objBitsKO_def)
+  apply (clarsimp simp: valid_idle'_def obj_at'_def sym_refs_asrt_def)
+  apply (drule_tac ko="tcb" and p=tptr in sym_refs_ko_atD'[rotated])
+   apply (fastforce simp: obj_at'_def)
+  apply (clarsimp simp: ko_wp_at'_def refs_of_rev')
+  done
+
+lemma
+  shows
+  maybeReturnSc_invs':
```
lsf37, line 4397: these should be on the same line

proof/refine/RISCV64/IpcCancel_R.thy
michaelmcinerney: Moved to the top of the file, with a `FIXME RT: remove`

proof/refine/RISCV64/Schedule_R.thy
```diff
@@ -2197,89 +2905,2607 @@ lemma scheduleChooseNewThread_ct_activatable'[wp]:
...
+lemma nonOverLappingMergeRefills_valid_refills'[wp]:
+  "nonOverlappingMergeRefills scPtr \<lbrace>valid_refills' scPtr\<rbrace>"
+  apply (wpsimp simp: nonOverlappingMergeRefills_def)
+  apply (clarsimp simp: obj_at'_def)
+  done
+
+definition head_insufficient_loop_measure where
+  "head_insufficient_loop_measure sc_ptr
+     \<equiv> measure (\<lambda>(_, s). case kheap s sc_ptr of Some (Structures_A.SchedContext sc _)
+                                             \<Rightarrow> (length (sc_refills sc)))"
```
lsf37, line 4197:
```suggestion
                                             \<Rightarrow> length (sc_refills sc))"
```

proof/refine/RISCV64/Schedule_R.thy
```diff
@@ -2197,89 +2905,2607 @@ lemma scheduleChooseNewThread_ct_activatable'[wp]:
...
+    Note that the method works best when the precondition is not schematic.
+    First, if the postcondition is not a conjunction, it will try to solve the goal with the
+    method given to `solves`. The goal will be left untouched if it cannot solve the goal.
+    If the postcondition is a conjunction, the method will pull it apart into all the conjuncts
+    using hoare_vcg_conj_lift_pre_fix. It will then attempt to solve each of these individually.\<close>
+method wps_conj_solves uses wp simp wps
+   = (clarsimp simp: pred_conj_def)?
+     , (intro hoare_vcg_conj_lift_pre_fix
+        ; (solves \<open>rule hoare_weaken_pre, (wpsimp wp: wp simp: simp | wps wps)+\<close>)?)
+       | (solves \<open>rule hoare_weaken_pre, (wpsimp wp: wp simp: simp | wps wps)+\<close>)?
```
lsf37, line 4536: If this is more generally useful, we should move it up into NondetMonadVCG or similar so that it is in scope for other proofs. (Plus adding one or two small usage examples)

proof/refine/RISCV64/Syscall_R.thy
```diff
@@ -1076,10 +1119,15 @@ lemma lcs_eq [wp]:
   "\<lbrace>\<top>\<rbrace> lookupCapAndSlot t cptr \<lbrace>\<lambda>rv. cte_wp_at' ((=) (fst rv) o cteCap) (snd rv)\<rbrace>,-"
   by (wpsimp simp: lookupCapAndSlot_def)
 
-lemma lec_eq[wp]:
+\<comment> \<open>lemma lec_eq[wp]:
   "\<lbrace>\<top>\<rbrace>
      lookupExtraCaps t buffer info
    \<lbrace>\<lambda>rv s. (\<forall>x\<in>set rv. cte_wp_at' ((=) (fst x) o cteCap) (snd x) s)\<rbrace>,-"
+  by (wpsimp wp: mapME_set simp: lookupExtraCaps_def)\<close>
```
lsf37, line 1126: should be removed

proof/refine/RISCV64/Tcb_R.thy
```diff
@@ -605,65 +626,393 @@ lemma threadSet_valid_objs_tcbPriority_update:
...
+  apply (wpsimp wp: hoare_vcg_const_imp_lift rescheduleRequired_invs' hoare_vcg_all_lift)
+       apply (wpsimp wp: threadGet_wp threadSetPriority_invs' tcbSchedDequeue_not_queued)+
+  apply (drule invs_queues')
+  apply (fastforce simp: ready_qs_runnable_def valid_queues'_def inQ_def
+                         obj_at'_def ct_in_state'_def runnable_eq_active')
+  done
+
+lemma runnable'_case_thread_state_If:
+  "(case rv of Structures_H.thread_state.Running \<Rightarrow> threadSetPriority_onRunning t x
+             | Structures_H.thread_state.Restart \<Rightarrow> threadSetPriority_onRunning t x
```
lsf37, line 746:
```suggestion
  "(case rv of Structures_H.Running \<Rightarrow> threadSetPriority_onRunning t x
             | Structures_H.Restart \<Rightarrow> threadSetPriority_onRunning t x
```

proof/refine/RISCV64/Tcb_R.thy
```diff
@@ -835,6 +1189,38 @@ lemma checkCapAt_cteInsert_corres:
...
+     apply (clarsimp simp: cte_wp_at_caps_of_state)
+     apply (rule conjI)
+      apply (frule same_object_as_cap_master)
+      apply (clarsimp simp: cap_master_cap_simps is_cnode_or_valid_arch_def
+                            is_cap_simps is_valid_vtable_root_def
+                     dest!: cap_master_cap_eqDs)
+     apply (erule disjE)
+      apply (erule(1) checked_insert_is_derived)
+      apply (fastforce simp: is_derived_def is_cap_simps same_object_as_def cte_wp_at_caps_of_state)+
+\<comment> \<open>
```
lsf37, line 1223: remove commented-out proof bits

lsf37 (approved):
Cool, this looks good. Ready to merge from my side once the smaller comments are fixed up.

One more small thing to fix up: I'd say something like "finish Refine session" or something similar in the commit message (instead of saying that it's getting them working) because the former better indicates that this is a real milestone.

proof/refine/RISCV64/Schedule_R.thy
michaelmcinerney: I did find this method useful in a fair few places, but with the `hoare_vcg_conj_lift_pre_fix`, it's more well suited to a forwards proof style. I would be happy to get rid of this method, or change that rule to `hoare_vcg_conj_lift` and redo the proofs if that's preferable. If we do want to move it to somewhere in `Lib`, I'm not sure where it should go. It uses `pred_conj_def` and `wps`, so it can't live in too many files. `wps` is defined in `WPEx.thy`, and it does work there, but I'm not sure whether that's suitable.

lsf37: How much work do you reckon it would be to change it to `hoare_vcg_conj_lift`?

We might need to introduce a new theory where we put more of these `wp` derived methods. And maybe one that acts as a top-level include theory for everything `wp`-related. We're currently sort-of using `NonDetMonadVCG` for that, but it's not a good fit, because it contains too much content itself that you'd want to use for other tools, so we just end up putting everything in there. If we do this rearrangement, we should do it on master first and then merge it into `rt`.

That all sounds like quite a bit of work, so we should probably defer cleaning up this method, make an issue for it and for the `wp` entry point, and do this in separate PRs.

michaelmcinerney: It would probably be a day or so of work to change it to `hoare_vcg_conj_lift`.

I'll make an issue for what you mention here

michaelmcinerney: [VER-1416](https://sel4.atlassian.net/browse/VER-770)

I'm not sure I've described this well. Feel free to edit or I can give it another shot myself

lsf37: All good, have tweaked it a bit.

lsf37:
Thanks for adding the `run_tests` changes (and the two missing file, looks like no RISCV64-specific proofs in those). If the tests pass, this is ready to merge from my side after squash/rebase.

michaelmcinerney:
Now rebased over the merge of master into rt. I've increased the timeout for Refine to 9 hours, but hopefully this can be safely reduced back to 8 hours soon

PR #461, opened 2022-04-27
Import documentation: CRefine notes

corlewis:
We import some documentation with notes/issues on the refinement from
Haskell to C.

The file is a Markdown-ified version of previous documentation hosted at
UNSW.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -24,6 +24,7 @@ Current topics are:
 - [Compacting proofs](compacting-proofs.md)
 - [Architecture Split](arch-split.md) Why and How-To
 - [Haskell Assertions](haskell-assertions.md): how to use assertions in Haskell to use information from AInvs on Haskell and C levels
+- [CRefine Notes](crefine-notes.md)
 
 ## Plans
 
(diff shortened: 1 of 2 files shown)


corlewis:
I'm a bit confused by the Trailing Whitespace failure. It looks like the syntax for one of the headings is exactly the wrong length and that the test thinks that it is a conflict marker.

Should I switch to the more standard heading syntax of `#`, or ignore the test failure?

lsf37:
> I'm a bit confused by the Trailing Whitespace failure. It looks like the syntax for one of the headings is exactly the wrong length and that the test thinks that it is a conflict marker.

Nice one :-)
 
> Should I switch to the more standard heading syntax of `#`, or ignore the test failure?

Using `#` is probably better for now, because we'll otherwise get the same failure whenever we edit the file.

lsf37 (approved):
Happy with it as it is for now (when the checks pass). The file could do with a bit of cleanup and modernisation, but we can do that in a separate round.

corlewis:
> Using `#` is probably better for now, because we'll otherwise get the same failure whenever we edit the file.

Ok, I've switched everything over to `#` now. For completeness, as part of this I did change the level of many of the headings in an attempt to be more consistent.

ryybrr (approved):

Xaphiosis (approved):

Xaphiosis:
Nicely done, now it's in docs and can be seen by all, opening a path to potential updates in the future!

PR #463, opened 2022-04-29
Merge `master` into `rt`

michaelmcinerney:


michaelmcinerney:
I'm not sure how gitlint is failing if all these commit messages are from the master branch and the tests must have passed there

lsf37:
It looks like gitlint is failing on old `rt` commits (the selection logic doesn't work properly for following merge commits since we don't usually do these). We should fix up those old commits at some point, but it's unrelated to this merge.

lsf37:
This is looking good, I'm running a separate test board to make sure, because the PR checks don't work with merge commits. If they pass I'll push this directly to `rt`.

lsf37 (approved):
Local testboard has passed.

lsf37:
Merged in 3bd85ee0175712db51c3e111dee3c5eadd968c47

michaelmcinerney:
Thanks Gerwin. I was a bit worried about the tests but I'm glad it was all fine

PR #464, opened 2022-05-02
run_tests: HaskellKernel and C config

lsf37:
-  give more time for downloading and compiling dependencies for runs  where these are not cached.
 
-   make sure we're catching all dependencies that are declared for `design-spec` in the top-level Makefile. In particular, we want `c-config` to run at least once before either `ASpec` or `ExecSpec` run it, to make sure these two are not racing on config generation in `-j 2`.

### DIFF
--- a/spec/tests.xml
+++ b/spec/tests.xml
@@ -37,7 +37,7 @@
 
     <set>
         <!-- Build Haskell kernel code. -->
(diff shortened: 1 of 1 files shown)


lsf37:
This should fix the "Weekly Clean" and "External" builds.

Xaphiosis (approved):

PR #465, opened 2022-05-03
docs: split_simps and case over fun/primrec

lsf37:
Document that and why we should prefer `case` over `primrec`/`fun` for non-recursive definitions with pattern matching.

### DIFF
--- a/docs/Style.thy
+++ b/docs/Style.thy
@@ -1,4 +1,5 @@
 (*
+     Copyright 2022, Proofcraft Pty Ltd
      Copyright 2021, Data61, CSIRO (ABN 41 687 119 230)
 
      SPDX-License-Identifier: CC-BY-SA-4.0
@@ -8,8 +9,8 @@ chapter \<open>An Isabelle Syntax Style Guide\<close>
 
 theory Style
   imports
-    Main
     Style_pre
+    Lib.SplitRule
 begin
 
 text \<open>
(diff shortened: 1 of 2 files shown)


lsf37:
This is not strictly speaking syntax style only, but I think it still fits in the main style guide as definition style.

lsf37:
Looks like I need to tweak the docs ROOT file to get the Lib dependency in.

docs/Style.thy
```diff
@@ -248,6 +249,38 @@ text \<open>
...
 
 
+section \<open>Fun, primrec, or case? case!\<close>
+
+text \<open>
+  When defining a non-recursive function that uses pattern matching, there are multiple options:
+  @{command fun}, @{command primrec}, and @{text case}. @{text case} initially seems like the least
+  powerful of these, but combined with the @{attribute split_simps} attribute, it is the option that
+  provides the highest degree of proof automation. @{command fun} and @{command primrec} provide
+  simp rule for each of the pattern cases, but when a closed term appears in the proof, there is no
```
Xaphiosis, line 259: *a* simp rule

docs/Style.thy
```diff
@@ -248,6 +249,38 @@ text \<open>
...
+definition some_f :: "nat \<Rightarrow> nat" where
+  "some_f x \<equiv> case x of 0 \<Rightarrow> 1 | Suc n \<Rightarrow> n+2"
+
+lemmas some_f_simps[simp] = some_f_def[split_simps nat.split]
+
+(* Pattern simplified automatically: *)
+lemma "some_f 0 = 1"
+  by simp
+
+(* But automated case split still available after unfolding: *)
```
Xaphiosis, line 279: *is* still available

docs/Style.thy
```diff
@@ -248,6 +249,38 @@ text \<open>
...
+  @{command fun}, @{command primrec}, and @{text case}. @{text case} initially seems like the least
+  powerful of these, but combined with the @{attribute split_simps} attribute, it is the option that
+  provides the highest degree of proof automation. @{command fun} and @{command primrec} provide
+  simp rule for each of the pattern cases, but when a closed term appears in the proof, there is no
+  way to automatically perform a case distinction -- you need to do a manual @{method cases} or
+  @{method case_tac} invocation. @{text case} does provide automatic splitting via split rules, and
+  @{attribute split_simps} adds the missing simp rules.
+
+  So the recommended pattern is to use @{text case} in the definition and directly after, via
+  attribute, add the simp rules to the global simp set as @{command fun}/@{command primrec} would
```
Xaphiosis, line 265: Sentence is backwards, confusing. Suggest something like:
"The recommended pattern is to use @{text case} in the definition and supply the simp rules directly after it, adding them to the global simp set as ... would do."

michaelmcinerney (approved):
If Raf's comments can be addressed, this looks good, and is a nice addition to the docs

lsf37:
Fixed up Raf's comments and added the dependency. Tests should now pass.

Xaphiosis (approved):

corlewis (approved):
Thanks for the documentation, this looks like a really useful pattern that I hadn't consciously seen before.

PR #468, opened 2022-05-09
cherry pick OptionMonad updates from rt branch

lsf37:
This should bring `master` fully up to date with all OptionMonad improvements from the MCS work.

There are further `lib/` changes which we might want to add to `master`, but those are probably better done in separate PRs.

### DIFF
--- a/lib/Monad_WP/OptionMonad.thy
+++ b/lib/Monad_WP/OptionMonad.thy
@@ -40,7 +40,7 @@ lemma opt_mapE:
   "\<lbrakk> (f |> g) s = Some v; \<And>v'. \<lbrakk>f s = Some v'; g v' = Some v \<rbrakk> \<Longrightarrow> P \<rbrakk> \<Longrightarrow> P"
   by (auto simp: in_opt_map_eq)
 
-lemma opt_map_left_Some:
+lemma opt_map_red:
   "f x = Some y \<Longrightarrow> (f |> g) x = g y"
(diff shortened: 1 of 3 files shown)


lsf37:
I haven't really changed anything in these commits, only removed some unrelated changes for 2 of them.

lib/Monad_WP/OptionMonad.thy
```diff
@@ -48,14 +68,115 @@ lemma opt_map_upd_Some:
...
-lemmas opt_map_upd[simp] = opt_map_upd_None opt_map_upd_Some
+lemma opt_map_Some_upd_Some:
+  "f(x \<mapsto> v) ||> g = (f ||> g)(x \<mapsto> g v)"
+  by (simp add: opt_map_upd_Some)
+
+lemmas opt_map_upd[simp]
+  = opt_map_upd_None opt_map_upd_Some opt_map_Some_upd_Some
+
+lemma opt_map_upd_triv:
+  "t k = Some x \<Longrightarrow> (t |> f)(k := f x) = t |> f"
```
Xaphiosis, line 79: not sure where "t k" terminology came from

lsf37: probably from the application area. `k` makes sense as a generic key, `t` I'm not sure about. `g` would be slightly better, but if we do that we should do it in both branches.

Xaphiosis (approved):

PR #469, opened 2022-05-09
crefine: remove duplicate lemma

corlewis:
The lemma map_to_ko_atI2 was exactly the same as map_to_ko_atI.


lsf37 (approved):
Let's hope they started out different and converged over time. Otherwise it'd be a bit sad :)

michaelmcinerney (approved):

PR #473, opened 2022-05-10
crefine: remove duplicate lemmas from CLevityCatch

corlewis:


lsf37 (approved):
Very nice!

Xaphiosis:
Which meaning of duplicate applies here? Ones already proved in Refine?
The reason I ask is because I'm also moving things around, and for example `user_getreg_rv` shows up twice in CRefine, and I was fearing a conflict on that one but don't see one here.

Xaphiosis (approved):

corlewis:
> Which meaning of duplicate applies here? Ones already proved in Refine? The reason I ask is because I'm also moving things around, and for example `user_getreg_rv` shows up twice in CRefine, and I was fearing a conflict on that one but don't see one here.

I think that `projectKO_user_data_device` was already proved in Refine, but most of the ones removed from `crefine/RISCV64/CLevityCatch.thy` were duplicates of lemmas in `crefine/Move_C.thy`.

PR #474, opened 2022-05-10
cparser: annotate AST printing for top-level decls

Xaphiosis:
In order to compare ASTs for appearance/disappearance/modification of
declarations, it is easier to have the annotations obvious to any
external tool for ease of parsing.

Annotations take the form:
"##<decl_type>: <name>", e.g. "##Function: ctzl"

### DIFF
--- a/tools/c-parser/standalone-parser/main.sml
+++ b/tools/c-parser/standalone-parser/main.sml
@@ -378,15 +378,34 @@ end
 fun print_ast cse ast = let
   open Absyn_Serial
 
+  fun decl_header decl
+    = case decl of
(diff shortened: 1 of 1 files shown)


Xaphiosis:
Primarily want feedback on whether anyone else parses or uses these ASTs in a way my change would inconvenience, but other ideas welcome as well.

mbrcknl (approved):
Looks good to me, and I'm not aware of anything that this would disturb.

lsf37 (approved):
Let's merge it and see if anything breaks, but I don't think the preprocess test will have any problems.

PR #476, opened 2022-05-12
crefine: split fastpath, rearrange Refine-based theory imports

Xaphiosis:
Several parts of CRefine did not or should not depend on anything
C-related, but the import hierarchy (and theory content) did not reflect
this. Namely:
* Move_C and ArchMove_C were intended to hold items that could be moved
  to Refine yet used `kernel_m` locale and imported the C spec.
* IsolatedThreadAction indicates how to rearrange statements in the
  design spec and has nothing to do with the C spec or framework.
* Fastpath_C contained the design spec of the fastpath, the design spec
  rewrite proofs, and the C refinement. Having to rebuild nearly all of
  CRefine to work on rewrite proofs wasted time.

In the new import hierarchy:
* Move_C imports only Refine; ArchMove_C builds on Refine_C
* IsolatedThreadAction imports only ArchMove_C
* The fastpath proofs are split into the spec definition (Fastpath_Defs)
  and rewrite proofs (Fastpath_Equiv), which don't depend on anything
  C-related, with their C refinement remaining in Fastpath_C.

While it is possible to separate out the fastpath definitions and rewire
proofs into a separate image or even move them to Refine, development
experience indicates keeping them alongside their C refinement remains
more convenient for the proof engineer involved.


Xaphiosis:
@mbrcknl : I know you're busy, so no need for a thorough review, mostly curious if you like this idea.

corlewis:
> Move_C imports only Refine; ArchMove_C builds on Refine_C

I think that should that be `ArchMove_C builds on Move_C`.

Xaphiosis:
The RISCV64 failure is due to my screwing up conflict resolution during rebase. I'll check that separately but leave the rest of the regression running. Sigh.

lsf37 (approved):
Very nice to consolidate this more!

mbrcknl (approved):
I've somewhat superficially looked over this, and it looks good! I do like the idea, and it seems you've also done some nice clean-up.

PR #477, opened 2022-05-26
lib: add syntax to crunch_ignore multiple instances

corlewis:
Add support to crunch_ignore so that multiple instances can be added to and removed from simultaneously.

There is an example in [lib/test/Crunch_Test_NonDet.thy](https://github.com/seL4/l4v/compare/master...corlewis:master#diff-a70563cedfb16fff3aebf5def8732ad4a8133223084b0330635b1b4b907a41f7) of this being used.

### DIFF
--- a/lib/Crunch.ML
+++ b/lib/Crunch.ML
@@ -117,21 +117,25 @@ val crunch_ignoreP =
     Outer_Syntax.local_theory
          @{command_keyword "crunch_ignore"}
         "add to and delete from list of things that crunch should ignore in finding prerequisites"
-        ((Scan.optional (P.$$$ "(" |-- P.name --| P.$$$ ")") "" -- Scan.optional
+        ((Scan.optional (P.$$$ "(" |-- P.list1 P.name --| P.$$$ ")") [""] -- Scan.optional
           (P.$$$ "(" |-- read_sections [add_sect, del_sect] --| P.$$$ ")")
-          []
-        )
-        >> (fn (crunch_instance, wpigs) => fn lthy =>
+          [])
+         >> (fn (crunch_instances, wpigs) => fn lthy =>
                let fun const_name const = dest_Const (read_const lthy const) |> #1;
                    val add = wpigs |> filter (fn (s,_) => s = #1 add_sect) |> map #2 |> constants
                                    |> map const_name;
                    val del = wpigs |> filter (fn (s,_) => s = #1 del_sect) |> map #2 |> constants
                                    |> map const_name;
-                   val crunch_ignore_add_del = (case get_crunch_instance crunch_instance (Proof_Context.theory_of lthy) of
-                     NONE => error ("Crunch has not been defined for " ^ crunch_instance)
-                   | SOME x => snd x);
+                   fun crunch_ignore_add_del inst =
+                     (case get_crunch_instance inst (Proof_Context.theory_of lthy) of
+                        NONE => error ("Crunch has not been defined for " ^ inst)
+                      | SOME x => snd x);
+                   val crunch_ignore_add_dels =
+                     map (fn inst => crunch_ignore_add_del inst add del) crunch_instances
+                   val crunch_ignore_add_dels' =
+                     fold (curry (op #>)) (tl crunch_ignore_add_dels) (hd crunch_ignore_add_dels)
                in
-                  Local_Theory.raw_theory (crunch_ignore_add_del add del) lthy
+                  Local_Theory.raw_theory crunch_ignore_add_dels' lthy
                end));
 
 end;
--- a/lib/test/Crunch_Test_NonDet.thy
+++ b/lib/test/Crunch_Test_NonDet.thy
@@ -26,13 +26,14 @@ definition
     crunch_foo1 13
   od"
 
+crunch_ignore (valid, empty_fail, no_fail) (add: bind)
+
 crunch (empty_fail) empty_fail: crunch_foo2
-  (ignore: modify bind)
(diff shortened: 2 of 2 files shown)


lib/test/Crunch_Test_NonDet.thy
```diff
@@ -26,13 +26,14 @@ definition
     crunch_foo1 13
   od"
 
+crunch_ignore (valid, empty_fail, no_fail) (add: bind)
```
Xaphiosis, line 29: wonderful demo of this working

corlewis: Oh yeah, I was going to ask for people's opinions on the syntax.

Is everyone happy for both parts of that invocation to use parentheses? While I'm here we could easily change the first to use a different pair of brackets if it might make it look nicer.

Xaphiosis: I'm fine to leave it as is. If we switched it to `[ ]`, would that make it any clearer what it does? I guess it would distinguish part 1 (props) from part 2 (add/del instructions). Do you have any feelings on this?

corlewis: I think I agree. It would be nice to distinguish the two parts, but all the options I've thought of either have other problems or don't seem to help that much.

I guess the 'nicest' way might be `crunch_ignore ‹valid, empty_fail, no_fail› (add: bind)`, but I'm not sure how standard that is and it barely feels worth changing.

Xaphiosis: Agreed. Can check in with Gerwin tomorrow to see if he has any thoughts, but looks like current syntax will be perfectly fine.

lib/Crunch.ML
```diff
@@ -117,21 +117,25 @@ val crunch_ignoreP =
...
-                   val crunch_ignore_add_del = (case get_crunch_instance crunch_instance (Proof_Context.theory_of lthy) of
-                     NONE => error ("Crunch has not been defined for " ^ crunch_instance)
-                   | SOME x => snd x);
+                   fun crunch_ignore_add_del inst =
+                     (case get_crunch_instance inst (Proof_Context.theory_of lthy) of
+                        NONE => error ("Crunch has not been defined for " ^ inst)
+                      | SOME x => snd x);
+                   val crunch_ignore_add_dels =
+                     map (fn inst => crunch_ignore_add_del inst add del) crunch_instances
+                   val crunch_ignore_add_del' =
```
Xaphiosis, line 135: I thought for a moment that `crunch_ignore_add_del'` isn't all that informative, but reading over the flow of the function it's ok.

Xaphiosis (approved):

lib/Crunch.ML
corlewis: Hmm, rereading it again, I think I prefer `crunch_ignore_add_dels'`. That's not really any more informative, but makes more sense to me.

Xaphiosis:
@corlewis : do you happen to know if attempts to do multiple add or del generate any warnings? (i.e. adding a constant to crunch_ignores that is already being ignored)

lib/test/Crunch_Test_NonDet.thy
lsf37: I don't think either of them (`[ .. ]` and `‹ .. ›`) are good, because both are associated with different existing concepts in Isabelle already (`[ .. ]` for attributes and `‹ .. ›` for changing syntax context, e.g. to antiquotation or inner syntax).

We could do what we do for `crunches`, though: `crunch_ignore valid, empty_fail, no_fail (...)`, or even `crunch_ignore ops for valid, empty_fail, no_fail` although that latter would be a larger break with current `crunch_ignore` syntax.

lsf37 (approved):
Concrete syntax discussion is still ongoing, but otherwise I'm happy with it (would also be fine with the parentheses as they are now).

corlewis:
> @corlewis : do you happen to know if attempts to do multiple add or del generate any warnings? (i.e. adding a constant to crunch_ignores that is already being ignored)

`crunch_ignore` currently doesn't generate any warnings. It shouldn't be too hard to add basic warnings like those ones.

Xaphiosis:
> > @corlewis : do you happen to know if attempts to do multiple add or del generate any warnings? (i.e. adding a constant to crunch_ignores that is already being ignored)
> 
> `crunch_ignore` currently doesn't generate any warnings. It shouldn't be too hard to add basic warnings like those ones.

I would appreciate the addition of those, though that can be a separate PR.

lib/test/Crunch_Test_NonDet.thy
corlewis: I prefer leaving it as is over those options. I can't find a nice way for both `crunches` and `crunch_ignore` to have similar syntaxes and still be nicely readable. Here's an example of a `crunches` command.
```
crunches readVCPUHardwareReg
  for (no_fail) no_fail[intro!, wp, simp]
  and (empty_fail) empty_fail[intro!, wp, simp]
  and (no_irq) no_irq[intro!, wp, simp]
  ```

The only option that would be consistent is something like your `crunch_ignore ops for valid, empty_fail, no_fail` suggestion, but then that doesn't have a nice way for removing things from the `crunch_ignore` set.

corlewis:
> Concrete syntax discussion is still ongoing, but otherwise I'm happy with it (would also be fine with the parentheses as they are now).

I think I'll merge this as is. If we come up with an alternative syntax then it should be part of a different PR anyway.

PR #478, opened 2022-05-27
lib: add warnings to crunch_ignore

corlewis:
These warn when attempting to add a constant to crunch_ignore that is
already being ignored, or when removing a constant that is not being ignored.

### DIFF
--- a/lib/Crunch.ML
+++ b/lib/Crunch.ML
@@ -10,7 +10,7 @@ structure CrunchTheoryData = Theory_Data
        ((Token.src list -> string -> string
          -> (string * ((Facts.ref * Token.src list), xstring) sum) list
          -> string list -> local_theory -> local_theory)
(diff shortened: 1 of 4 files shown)


corlewis:
This was accidentally merged and should still be reviewed.

Xaphiosis:
Had a look at this, and everything seems fine. I'm curious why you had to switch to `local_theory` to make this work.

lsf37:
I'm also happy with it (and I think `local_theory` is an improvement, so that's good as well from my side).

corlewis:
> Had a look at this, and everything seems fine. I'm curious why you had to switch to `local_theory` to make this work.

I needed access to the local theory so that when printing the warning I could add the automatic markup to the name of the const.

corlewis:
For completeness, the tests have now all passed with these commits.

PR #479, opened 2022-05-27
lib: add a crunch option for naming schemes

corlewis:
Some properties that crunch can be used for have different legacy naming schemes. This commit makes it possible for different instances of crunch to be configured for either prefix or suffix naming.

### DIFF
--- a/lib/crunch-cmd.ML
+++ b/lib/crunch-cmd.ML
@@ -109,6 +109,8 @@ signature CrunchInstance =
 sig
     (* The name of the property. *)
     val name : string;
+    (* Which naming scheme to use - true for prefix, false for suffix*)
+    val prefix_name_scheme : bool
     (* The type of any parameters of the property beyond a (optional) precondition
        and the function being crunched. *)
     type extra;
@@ -143,11 +145,12 @@ end
 signature CRUNCH =
 sig
     type extra;
-    (* Crunch configuration: theory, naming scheme, name space, wp rules, wps_rules,
-    constants to ignore, simp rules, constants to not ignore, unfold rules*)
-    type crunch_cfg = {ctxt: local_theory, prp_name: string, nmspce: string option,
-        wp_rules: (string * thm) list, wps_rules: thm list, igs: string list, simps: thm list,
-        ig_dels: string list, rules: thm list};
+    (* Crunch configuration: theory, property name, name space, wp rules, wps_rules,
+       constants to ignore, simp rules, constants to not ignore, unfold rules *)
+    type crunch_cfg =
+      {ctxt: local_theory, prp_name: string, nmspce: string option,
+       wp_rules: (string * thm) list, wps_rules: thm list, igs: string list,
+       simps: thm list, ig_dels: string list, rules: thm list};
 
     (* Crunch takes a configuration, a precondition, any extra information needed, a debug stack,
     a constant name, and a list of previously proven theorems, and returns a theorem for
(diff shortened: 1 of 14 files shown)


lsf37 (approved):
Looks good from my side. I think this is also already merged, so we should close the PR.

corlewis:
> Looks good from my side. I think this is also already merged, so we should close the PR.

No, this one was actually in a branch and hasn't been merged yet.

lib/crunch-cmd.ML
```diff
@@ -143,11 +145,12 @@ end
...
-    constants to ignore, simp rules, constants to not ignore, unfold rules*)
-    type crunch_cfg = {ctxt: local_theory, prp_name: string, nmspce: string option,
-        wp_rules: (string * thm) list, wps_rules: thm list, igs: string list, simps: thm list,
-        ig_dels: string list, rules: thm list};
+    (* Crunch configuration: theory, property name, name space, wp rules, wps_rules,
+       constants to ignore, simp rules, constants to not ignore, unfold rules *)
+    type crunch_cfg =
+      {ctxt: local_theory, prp_name: string, nmspce: string option,
+       wp_rules: (string * thm) list, wps_rules: thm list, igs: string list,
+       simps: thm list, ig_dels: string list, rules: thm list};
```
Xaphiosis, line 153: stared at this for a bit... it's only re-styling, right?

lib/crunch-cmd.ML
```diff
@@ -882,9 +892,12 @@ fun crunch_x atts extra prp_name wpigs consts ctxt =
...
         val ctxt'' = ctxt' delsimps simp_dels;
 
-        val (_, thms) = funkyfold (crunch {ctxt = ctxt'', prp_name = prp_name, nmspce = nmspce,
-              wp_rules = wp_rules, wps_rules = wps_rules, igs = igs, simps = simps,
-              ig_dels = ig_dels, rules = rules} pre' extra' [])
+        val crunch_cfg =
+          {ctxt = ctxt'', prp_name = prp_name, nmspce = nmspce, wp_rules = wp_rules,
+           wps_rules = wps_rules, igs = igs, simps = simps, ig_dels = ig_dels,
+           rules = rules}
+        val (_, thms) =
```
Xaphiosis, line 899: doesn't the following fit on one line? strange to stair-step it like this

proof/refine/ARM_HYP/Ipc_R.thy
```diff
@@ -1563,8 +1563,8 @@ lemma makeArchFaultMessage_corres:
     apply (rule corres_split_eqr[OF _ asUser_getRestartPC_corres])
       apply (rule corres_split_eqr[OF _ corres_machine_op])
          apply (rule corres_trivial, simp)
-        apply (rule corres_underlying_trivial[OF addressTranslateS1CPR_no_fail])
-       apply (wp+, auto)
+        apply (rule corres_underlying_trivial)
+        apply (wp+, auto)
```
Xaphiosis, line 1567: These changes to ARM_HYP Refine appear in a commit tagged "lib proof:", which is not a tag I remember seeing before

lib/crunch-cmd.ML
corlewis: Yeah, it's just a re-styling now.

My original implementation of this option changed the type, and when I undid the rest of those changes I decided that this style is more consistent with the rest of the file. Apologies for any time you spent staring at it.

proof/refine/ARM_HYP/Ipc_R.thy
Xaphiosis: Looking over the history, I see that `lib proof:` has existed before, but it's confusing, if we're going that way I'd vote for `lib+proof:`

lib/crunch-cmd.ML
corlewis: True, it does now that the cfg is a separate line.

proof/refine/ARM_HYP/Ipc_R.thy
corlewis: In my mind I've changed lib to make `no_fail` use prefix-naming, and then updated all of the different proofs for anything that broke. While it turns out that this was the only change needed, I wouldn't want to label it specifically `ARM_HYP refine` since it conceptually isn't specific to just that.

corlewis: Oh, I'd also be happy with lib+proof.

Xaphiosis (approved):
Thanks. Looks good now, don't forget to squash.

Xaphiosis:
Good to go from my side. Note to self (@Xaphiosis) and @lsf37 : when this gets merged, the aarch64 branch needs an update for Machine_AI whenever it's rebased else we'll get a failure.

PR #481, opened 2022-05-30
add sorry + FIXME count scripts

lsf37:
The sorry-count script used to live in ci-actions/aws-proofs and still needs to be removed there. The FIXME count script is new.

Both start counting at the repo root and report per major interesting directory (e.g. `spec/abstract` or `proof/refine`). The FiXME count includes a total over the entire repo.


Xaphiosis (approved):
As we discussed, would be nice to error check the `cd "$(git rev-parse --show-toplevel)"`

PR #482, opened 2022-05-30
lib: fix crunch to only apply attributes once

corlewis:
I wasn't sure whether to put more tags on this commit. I changed one line in `invariant-abstract` which I already knew was explicitly due to this issue, but I'm sure that there are others that I don't know about and I don't want to go looking for them. Should I add `ainvs` to the tag?

### DIFF
--- a/lib/crunch-cmd.ML
+++ b/lib/crunch-cmd.ML
@@ -271,7 +271,7 @@ fun maybe_thms thy name = get_thms thy name handle ERROR _ => []
 fun thy_maybe_thms thy name = Global_Theory.get_thms thy name handle ERROR _ => []
 
 fun add_thm thm atts name ctxt =
-  Local_Theory.notes [((Binding.name name, atts), [([thm], atts)])] ctxt |> #2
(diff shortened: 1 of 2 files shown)


Xaphiosis:
I'd split this commit. The crunch change is `lib`, but the `riscv ainvs` change is pretty much cleanup and has nothing directly to do with fixing crunch.
I would also encourage a little more text in the commit for lib commits. What we gain here is warnings no longer getting generated when adding attributes to crunch/crunches properties, which is obvious to us now but not to someone looking at the commit later.

Xaphiosis:
Also, huge thanks for hunting this one down, it was quite annoying.

corlewis:
> I'd split this commit. The crunch change is `lib`, but the `riscv ainvs` change is pretty much cleanup and has nothing directly to do with fixing crunch. I would also encourage a little more text in the commit for lib commits. What we gain here is warnings no longer getting generated when adding attributes to crunch/crunches properties, which is obvious to us now but not to someone looking at the commit later.

Thanks for the suggestions, commits have been rewritten.

lsf37 (approved):
Thanks for tracking this one down.

Xaphiosis (approved):

PR #484, opened 2022-06-03
clean up translator warnings

lsf37:
- makes warning/error message handling in the Haskell translator consistent
- then uses that to provide highlighting with colour for tty use

I have a few more ideas what could be done to make this more useful (provide verbosity switch and suppress some of the warnings so that useful warnings stand out), but I wanted to merge this first before I do more.


tools/haskell-translator/lhs_pars.py
```diff
@@ -1407,10 +1430,10 @@ def get_type_map(string):
...
     actual_map = bits[1].strip()
     if not (actual_map.startswith('{') and actual_map.endswith('}')):
-        print('Error in ' + filename)
-        print('Expected "A { blah :: blah etc }"')
-        print('However { and } not found correctly')
-        print('When parsing %s' % string)
+        error(f'in {filename}\n'
+              'Expected "A { blah :: blah etc }"\n'
+              'However { and } not found correctly\n'
+              f'When parsing {string}')
```
corlewis, line 1436: Very minor, but you're currently not consistent in whether you make all lines of multiline strings f-strings, or just the lines that have expressions.

corlewis (approved):
Very nice! The only suggestion I have is for future work; it would be nice if more of the warnings and errors had line numbers in them.

lsf37:
> Very nice! The only suggestion I have is for future work; it would be nice if more of the warnings and errors had line numbers in them.

That is a bigger refactor, because usually the line number is not available to the function that produces the warning. In the next round I'm shooting for at least providing the file name :-) (because it's never clear to me if it's the file above or below the warning -- you'd usually think above, but it is the one below..)

corlewis:
> > Very nice! The only suggestion I have is for future work; it would be nice if more of the warnings and errors had line numbers in them.
> 
> That is a bigger refactor, because usually the line number is not available to the function that produces the warning. In the next round I'm shooting for at least providing the file name :-) (because it's never clear to me if it's the file above or below the warning -- you'd usually think above, but it is the one below..)

Haha, I thought about mentioning file name but was assuming that you could get it from the line above!

tools/haskell-translator/lhs_pars.py
corlewis: Ah, right. Based on me just learning about f-strings and reading up on them, I think you would use double braces in that case.

lsf37: Yes, you could escape them, but I find that less readable than just not using an f-string when it's not necessary.

PR #485, opened 2022-06-03
fewer haskell translator warnings

lsf37:
This is the second round of improvements for error/warning reporting in the Haskell translator:

- suppress warnings for types we know about
- remove warnings for intentionally missing X64 CPP setup
- use a status line for file progress output when running in a terminal. This should reduce clutter when run interactively.
- provide file name for warnings and errors

### DIFF
--- a/tools/haskell-translator/lhs_pars.py
+++ b/tools/haskell-translator/lhs_pars.py
@@ -702,12 +702,28 @@ def newtype_transform(d):
         return named_newtype_transform(line, header, d)
 
 
+# parameterised types of which we know that they are already defined in Isabelle
+known_type_assignments = [
+    'domain_schedule',
+    'kernel',
+    'kernel_f',
+    'kernel_f f',  # there is a KernelF instance that gets transformed into this
+    'kernel_init',
+    'kernel_init_state',
+    'machine_data',
+    'machine_monad',
+    'ready_queue',
+    'user_monad'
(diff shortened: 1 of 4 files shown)


lsf37:
Oops, forgot the header for `msgs.py`, one sec.

tools/haskell-translator/lhs_pars.py
```diff
@@ -17,6 +17,7 @@
 from functools import reduce
 from msgs import error, warning
 
+
```
corlewis, line 20: Is this an intentional newline?

corlewis (approved):
This is a massive improvement! It's amazing how much nicer that status line makes the output, although for completeness its addition could also be mentioned in the commit message.

lsf37:
> This is a massive improvement! It's amazing how much nicer that status line makes the output, although for completeness its addition could also be mentioned in the commit message.

Indeed it should, will add.

tools/haskell-translator/lhs_pars.py
lsf37: yes, it's supposed to be its own section for (local) imports, and needs to be 2 lines away from the `class` that follows

lsf37: Although I should fish out the `braces` import that is in the middle of the system imports above. Will do.

PR #487, opened 2022-06-07
Monadic rewrite cleanup

Xaphiosis:
Next step will be trying to wrap my head around how these lemmas are supposed to be used in practice and maybe adding some documentation on the approach in general.

I'm mainly looking for feedback on:
* style/sections/etc
* names of things (I did not rename in this set of patches)
* derps
* future ideas (I have some, but the more the merrier)

### DIFF
--- a/lib/MonadicRewrite.thy
+++ b/lib/MonadicRewrite.thy
@@ -601,22 +601,6 @@ lemma monadic_rewrite_weaken_failure:
 
 text \<open>corres / refinement\<close>
 
-lemma corres_underlyingI:
-  assumes rv: "\<And>s t rv' t'. \<lbrakk>(s, t) \<in> R; P s; P' t; (rv', t') \<in> fst (c t)\<rbrakk>
-                            \<Longrightarrow> \<exists>(rv, s') \<in> fst (a s). (s', t') \<in> R \<and> r rv rv'"
-  and     nf: "\<And>s t. \<lbrakk>(s, t) \<in> R; P s; P' t; nf'\<rbrakk> \<Longrightarrow> \<not> snd (c t)"
-  shows  "corres_underlying R nf nf' r P P' a c"
-  unfolding corres_underlying_def using rv nf by (auto simp: split_def)
-
-lemma corres_underlyingE:
-  assumes cul: "corres_underlying R nf nf' r P P' a c"
-  and     xin: "(s, t) \<in> R" "P s" "P' t" "(rv', t') \<in> fst (c t)"
-  and      rl: "\<And>s' rv. \<lbrakk>nf' \<longrightarrow> \<not> snd (c t); (rv, s') \<in> fst (a s); (s', t') \<in> R; r rv rv'\<rbrakk> \<Longrightarrow> Q"
-  and      nf: "nf \<longrightarrow> \<not> snd (a s)"
-  shows   "Q"
-  using cul xin nf
-  unfolding corres_underlying_def by (fastforce simp: split_def intro: rl)
-
 lemma monadic_rewrite_corres:
   assumes cu: "corres_underlying R False nf' r P P' a' c"
   and     me: "monadic_rewrite False True Q a a'"
--- a/lib/MonadicRewrite.thy
+++ b/lib/MonadicRewrite.thy
@@ -1,10 +1,11 @@
 (*
+ * Copyright 2022, Proofcraft Pty Ltd
  * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
  *
  * SPDX-License-Identifier: BSD-2-Clause
  *)
 
-(* A theory of rewriting under refinement. *)
+(* Theory of monadic rewriting under preconditions, applicable under refinement. *)
 
 theory MonadicRewrite
 imports
@@ -14,24 +15,102 @@ imports
   LemmaBucket
 begin
 
-definition
-  monadic_rewrite :: "bool \<Rightarrow> bool \<Rightarrow> ('a \<Rightarrow> bool)
-                       \<Rightarrow> ('a, 'b) nondet_monad \<Rightarrow> ('a, 'b) nondet_monad \<Rightarrow> bool"
-where
- "monadic_rewrite F E P f g \<equiv> \<forall>s. P s \<and> (F \<longrightarrow> \<not> snd (f s))
-           \<longrightarrow> (E \<longrightarrow> f s = g s)
-            \<and> (\<not> E \<longrightarrow> fst (g s) \<subseteq> fst (f s) \<and> (snd (g s) \<longrightarrow> snd (f s)))"
+definition monadic_rewrite ::
+  "bool \<Rightarrow> bool \<Rightarrow> ('s \<Rightarrow> bool) \<Rightarrow> ('s, 'a) nondet_monad \<Rightarrow> ('s, 'a) nondet_monad \<Rightarrow> bool"
+  where
+   "monadic_rewrite F E P f g \<equiv>
+      \<forall>s. P s \<and> (F \<longrightarrow> \<not> snd (f s))
+          \<longrightarrow> (E \<longrightarrow> f s = g s) \<and> (\<not> E \<longrightarrow> fst (g s) \<subseteq> fst (f s) \<and> (snd (g s) \<longrightarrow> snd (f s)))"
 
+lemma monadic_rewrite_inst:
+  "monadic_rewrite F E P f g \<Longrightarrow> monadic_rewrite F E P f g"
+  by assumption
 
-(* FIXME: also in Retype_C *)
+lemma monadic_rewrite_refl_generic:
+  "monadic_rewrite F E P f f"
+  by (simp add: monadic_rewrite_def)
+
+lemmas monadic_rewrite_refl = monadic_rewrite_refl_generic[where P=\<top>]
+
+lemma monadic_rewrite_impossible:
+  "monadic_rewrite F E \<bottom> f g"
+  by (clarsimp simp: monadic_rewrite_def)
+
+lemma monadic_rewrite_imp:
+  "\<lbrakk> monadic_rewrite F E Q f g; \<And>s. P s \<Longrightarrow> Q s \<rbrakk> \<Longrightarrow> monadic_rewrite F E P f g"
+  by (auto simp add: monadic_rewrite_def)
+
+lemma monadic_rewrite_trans:
+  "\<lbrakk> monadic_rewrite F E P f g; monadic_rewrite F E Q g h \<rbrakk>
+   \<Longrightarrow> monadic_rewrite F E (P and Q) f h"
+  by (auto simp add: monadic_rewrite_def)
+
+lemma monadic_rewrite_trans_dup:
+  "\<lbrakk> monadic_rewrite F E P f g; monadic_rewrite F E P g h \<rbrakk>
+   \<Longrightarrow> monadic_rewrite F E P f h"
+  by (rule monadic_rewrite_imp, (rule monadic_rewrite_trans; assumption), simp)
+
+lemma monadic_rewrite_from_simple:
+  "P \<longrightarrow> f = g \<Longrightarrow> monadic_rewrite F E (\<lambda>_. P) f g"
+  by (simp add: monadic_rewrite_def)
+
+lemma monadic_rewrite_gen_asm:
+  "\<lbrakk> P \<Longrightarrow> monadic_rewrite F E Q f g \<rbrakk> \<Longrightarrow> monadic_rewrite F E ((\<lambda>_. P) and Q) f g"
+  by (auto simp add: monadic_rewrite_def)
+
+lemma monadic_rewrite_name_pre:
+  "\<lbrakk> \<And>s. P s \<Longrightarrow> monadic_rewrite F E ((=) s) f g \<rbrakk>
+   \<Longrightarrow> monadic_rewrite F E P f g"
+  by (auto simp add: monadic_rewrite_def)
+
+lemma monadic_rewrite_cases:
+  "\<lbrakk> P \<Longrightarrow> monadic_rewrite F E Q a b; \<not> P \<Longrightarrow> monadic_rewrite F E R a b \<rbrakk>
+   \<Longrightarrow> monadic_rewrite F E (\<lambda>s. (P \<longrightarrow> Q s) \<and> (\<not> P \<longrightarrow> R s)) a b"
+  by (cases P, simp_all)
+
+lemma monadic_rewrite_is_refl:
+  "x = y \<Longrightarrow> monadic_rewrite F E \<top> x y"
+  by (simp add: monadic_rewrite_refl)
+
+lemma monadic_rewrite_refl3:
+  "\<lbrakk> \<And>s. P s \<Longrightarrow> f s = g s \<rbrakk> \<Longrightarrow> monadic_rewrite F E P f g"
+  by (simp add: monadic_rewrite_def)
+
+lemmas monadic_rewrite_refl2 = monadic_rewrite_refl3[where P=\<top>]
+
+lemma monadic_rewrite_exists:
+  "(\<And>v. monadic_rewrite F E (Q v) m m')
+   \<Longrightarrow> monadic_rewrite F E ((\<lambda>s. \<forall>v. P v s \<longrightarrow> Q v s) and (\<lambda>s. \<exists>v. P v s)) m m'"
+  by (auto simp: monadic_rewrite_def)
+
+lemma monadic_rewrite_exists_v:
+  "\<lbrakk> \<And>v. monadic_rewrite E F (Q v) f g \<rbrakk>
+   \<Longrightarrow> monadic_rewrite E F (\<lambda>x. (\<exists>v. P v x) & (\<forall>v. P v x \<longrightarrow> Q v x)) f g"
+  by (rule monadic_rewrite_name_pre)
+     (fastforce elim: monadic_rewrite_imp)
+
+lemma monadic_rewrite_weaken:
+  "monadic_rewrite (F \<and> F') (E \<or> E') P f g
+   \<Longrightarrow> monadic_rewrite F' E' P f g"
+  by (auto simp: monadic_rewrite_def)
+
+lemma monadic_rewrite_weaken2:
+  "monadic_rewrite F E P f g
+   \<Longrightarrow> monadic_rewrite F' E' ((\<lambda>_. (F \<longrightarrow> F') \<and> (E' \<longrightarrow> E)) and P) f g"
+  apply (rule monadic_rewrite_gen_asm)
+  apply (rule monadic_rewrite_weaken[where F=F and E=E])
+  apply auto
+  done
+
+text \<open>bind/bindE\<close>
+
 lemma snd_bind:
   "snd ((a >>= b) s) = (snd (a s) \<or> (\<exists>(r, s') \<in> fst (a s). snd (b r s')))"
   by (auto simp add: bind_def split_def)
 
 lemma monadic_rewrite_bind:
-  "\<lbrakk> monadic_rewrite F E P f g; \<And>x. monadic_rewrite F E (Q x) (h x) (j x);
-           \<lbrace>R\<rbrace> g \<lbrace>Q\<rbrace> \<rbrakk>
-        \<Longrightarrow> monadic_rewrite F E (P and R) (f >>= (\<lambda>x. h x)) (g >>= (\<lambda>x. j x))"
+  "\<lbrakk> monadic_rewrite F E P f g; \<And>x. monadic_rewrite F E (Q x) (h x) (j x); \<lbrace>R\<rbrace> g \<lbrace>Q\<rbrace> \<rbrakk>
+   \<Longrightarrow> monadic_rewrite F E (P and R) (f >>= (\<lambda>x. h x)) (g >>= (\<lambda>x. j x))"
   apply (cases E)
    apply (clarsimp simp: monadic_rewrite_def snd_bind imp_conjL)
    apply (drule spec, drule(1) mp, clarsimp)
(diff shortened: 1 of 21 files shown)


Xaphiosis:
Found some `lemmas` statements I didn't notice before, moved those over too. Hopefully fine for reviewing now. If I didn't mess anything up, without renames and attribute declarations proofs should go through as normal.

lib/MonadicRewrite.thy
```diff
@@ -14,24 +15,100 @@ imports
...
+
+lemma monadic_rewrite_trans:
+  "\<lbrakk> monadic_rewrite F E P f g; monadic_rewrite F E Q g h \<rbrakk>
+   \<Longrightarrow> monadic_rewrite F E (P and Q) f h"
+  by (auto simp add: monadic_rewrite_def)
+
+lemma monadic_rewrite_trans_dup:
+  "\<lbrakk> monadic_rewrite F E P f g; monadic_rewrite F E P g h \<rbrakk>
+   \<Longrightarrow> monadic_rewrite F E P f h"
+  by (auto simp add: monadic_rewrite_def)
```
lsf37, line 45: Minor nitpick: it'd be neat to do the proof of this one in terms of `monadic_rewrite_trans` so that it's clear that the first is fundamental and this one here is derivable.

lib/MonadicRewrite.thy
```diff
@@ -14,24 +15,100 @@ imports
...
+
+lemma monadic_rewrite_refl:
+  "monadic_rewrite F E \<top> f f"
+  by (simp add: monadic_rewrite_def)
+
+lemma monadic_rewrite_impossible:
+  "monadic_rewrite F E \<bottom> f g"
+  by (clarsimp simp: monadic_rewrite_def)
+
+lemma monadic_rewrite_trans:
```
lsf37, line 37: Not for this PR, but in general: since we have reflexivity and transitivity, I think we should also show symmetry (when `E` is true) and anti-symmetry (when `E` is false). Not sure we'll ever really need it, but it's good to state what the basic properties are as a relation, and it might allow some more manoeuvring in proofs.

lib/MonadicRewrite.thy
```diff
@@ -14,24 +15,100 @@ imports
...
+   "monadic_rewrite F E P f g \<equiv>
+      \<forall>s. P s \<and> (F \<longrightarrow> \<not> snd (f s))
+          \<longrightarrow> (E \<longrightarrow> f s = g s) \<and> (\<not> E \<longrightarrow> fst (g s) \<subseteq> fst (f s) \<and> (snd (g s) \<longrightarrow> snd (f s)))"
 
+lemma monadic_rewrite_inst:
+  "monadic_rewrite F E P f g \<Longrightarrow> monadic_rewrite F E P f g"
+  by assumption
+
+lemma monadic_rewrite_refl:
+  "monadic_rewrite F E \<top> f f"
```
lsf37, line 30: Also an idea for later: instead of \top I'd have expected any P here. Of course with weakening this is equivalent, and maybe we actually want P to be as concrete as possible for "final" rules in terms of rewriting. But in terms of basic properties, I'd do a `P` version and potentially use an instance with \top for rewriting.

lib/MonadicRewrite.thy
```diff
@@ -14,24 +15,100 @@ imports
...
+
+lemma monadic_rewrite_exists_v:
+  "\<lbrakk> \<And>v. monadic_rewrite E F (Q v) f g \<rbrakk>
+   \<Longrightarrow> monadic_rewrite E F (\<lambda>x. (\<exists>v. P v x) & (\<forall>v. P v x \<longrightarrow> Q v x)) f g"
+  by (rule monadic_rewrite_name_pre)
+     (fastforce elim: monadic_rewrite_imp)
+
+lemma monadic_rewrite_weaken:
+  "monadic_rewrite (F \<and> F') (E \<or> E') P f g
+   \<Longrightarrow> monadic_rewrite F' E' P f g"
```
lsf37, line 92: also for later: the name weaken is a bit strange for this rule. I get how it weakens, but for corres and others we called the `monadic_rewrite_imp` weaken. Probably better in the long run to be consistent there.

lsf37:
I haven't looked yet at typical uses of everything -- based on the rules we have, my assumption would be that we mainly decompose over bind and other control structures (generating schematics) until we hit atoms or chunks we know how to refine/rewrite (instantiating schematics). The main complication seems to be that sometimes we want to do more radical rewriting than just in atoms, maybe changing order of things etc.

I don't know how to help much with the latter, but for the former, we could start a wp-like method and rule set. The rule set would get all splitting rules and all atoms we know. We could distinguish those two sets, but I don't think we really need to (at least we don't need that in wp).

We'd then write a method that applies one rule, potentially making the precondition schematic if it isn't yet (like wp_pre). This would allow single-stepping. We can use that method with `+` if we want to proceed blindly, but I suspect we don't want to do that very often.

This is not a huge step forward, but it eliminates the need to remember the names of all splitting rules and basic atoms at least.

We could also add a small convenience method that applies corres_rewrite to a corres goal, either by explicitly stating to which term we want to rewrite something or by supplying a rule that already has a corres_rewrite proven. Not sure if that is very helpful, but it might clean up some instantiations etc.

lsf37:
Raf also mentioned the idea of applying a rewrite until some condition is true or some rule matches. I think that could be useful as well, maybe with the single-step method as the primitive stepping function.

Xaphiosis:
Squashed and rebased. So far no one other than you has responded. Any thoughts on where to keep these discussions after the PR is merged?

Regarding the strategy for approaching automation, I'm still hazy, same as you, before I examine more carefully how these are actually used. The atomic-steps you describe sound a lot like `corres` or `wp`, while the more complex rewrites (and my zip-until idea) sound more like `ccorres_rewrite`. I have thoughts like:
* zip along until you find a matching rule using this set of rules (A)
* when you get to a rewrite match, do a rewrite from this set of rules (B)
* resolve any `wp` rules in some way (rewriting under preconditions) (C)
* keep zipping through until you get to the end, then collapse the remainder with some kind of symmetry

Then you could give instructions like: "zip through everything (A), and you must use every rule I gave you in (B) (otherwise stop or fail?), and leave the (C) to another tool/interaction (maybe integrated at some distant future stage)".
This is so hazy as to be vapour at the moment, but that's my current thinking direction, and it sounds a bit like `ccorres_rewrite` with hopefully an ability to match larger blocks.

lib/MonadicRewrite.thy
Xaphiosis: Done as requested.

lib/MonadicRewrite.thy
Xaphiosis: I'm afraid you've lost me here due to gaps in my background. What is anti-symmetry and where can I read about it?
Fine to add it once I understand.

lib/MonadicRewrite.thy
Xaphiosis: I'll add it, but will make it `monadic_rewrite_refl_gen` for now (better name welcome) because `monadic_rewrite_refl` is used a lot in its `\top` form and I don't want that rename turbulence on this PR. I suspect it's more useful in the `\top` form anyway.

lib/MonadicRewrite.thy
Xaphiosis: What would you suggest? I'm not sure I follow exactly; these two weaken rules mess around with the failure flags.

lib/MonadicRewrite.thy
```diff
@@ -14,24 +15,100 @@ imports
...
-where
- "monadic_rewrite F E P f g \<equiv> \<forall>s. P s \<and> (F \<longrightarrow> \<not> snd (f s))
-           \<longrightarrow> (E \<longrightarrow> f s = g s)
-            \<and> (\<not> E \<longrightarrow> fst (g s) \<subseteq> fst (f s) \<and> (snd (g s) \<longrightarrow> snd (f s)))"
+definition monadic_rewrite ::
+  "bool \<Rightarrow> bool \<Rightarrow> ('s \<Rightarrow> bool) \<Rightarrow> ('s, 'a) nondet_monad \<Rightarrow> ('s, 'a) nondet_monad \<Rightarrow> bool"
+  where
+   "monadic_rewrite F E P f g \<equiv>
+      \<forall>s. P s \<and> (F \<longrightarrow> \<not> snd (f s))
+          \<longrightarrow> (E \<longrightarrow> f s = g s) \<and> (\<not> E \<longrightarrow> fst (g s) \<subseteq> fst (f s) \<and> (snd (g s) \<longrightarrow> snd (f s)))"
```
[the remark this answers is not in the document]
Xaphiosis: I'm not sure that's exactly the case. Attempting a proof:
```
lemma monadic_rewrite_antisym:
  "monadic_rewrite F True P f g = (monadic_rewrite F False P f g ∧ monadic_rewrite F False P g f)"
  unfolding monadic_rewrite_def
  apply (rule iffI)
  subgoal
    apply clarsimp
    apply (cases F; clarsimp)
```
This leaves us in this state:
```
⋀s. ⟦∀s. P s ∧ ¬ snd (f s) ⟶ f s = g s; P s; ¬ snd (g s); F⟧ ⟹ fst (f s) ≤ fst (g s) ∧ ¬ snd (f s)
```
Which as far as I can see doesn't have a proof. Thoughts?

lib/MonadicRewrite.thy
Xaphiosis: Done

lsf37:
> Squashed and rebased. So far no one other than you has responded. Any thoughts on where to keep these discussions after the PR is merged?

The PR will stay around when closed so things will still be there for reference, but it probably makes sense to put the idea comments into an issue so we can keep discussing there.

> Regarding the strategy for approaching automation, I'm still hazy, same as you, before I examine more carefully how these are actually used. The atomic-steps you describe sound a lot like `corres` or `wp`, while the more complex rewrites (and my zip-until idea) sound more like `ccorres_rewrite`. I have thoughts like:
> 
> * zip along until you find a matching rule using this set of rules (A)
> * when you get to a rewrite match, do a rewrite from this set of rules (B)
> * resolve any `wp` rules in some way (rewriting under preconditions) (C)
> * keep zipping through until you get to the end, then collapse the remainder with some kind of symmetry
> 
> Then you could give instructions like: "zip through everything (A), and you must use every rule I gave you in (B) (otherwise stop or fail?), and leave the (C) to another tool/interaction (maybe integrated at some distant future stage)". This is so hazy as to be vapour at the moment, but that's my current thinking direction, and it sounds a bit like `ccorres_rewrite` with hopefully an ability to match larger blocks.

Interesting, and much more ambitious than mine. If we compare this to term rewriting the set (A) is selecting the sub term you want to rewrite, (B) is the rewrite set (it'll be hard to demand that all rules of a set have been used in general, i.e. too much possible backtracking probably, but I can see a few straightforward cases), and (C) are solving side conditions on the rewrite rules.

In the Isabelle simplifier, to apply a conditional rewrite it first has to solve the side condition. In our case that might be too hard, although it might make sense for some trivial and almost trivial ones. The problem with deferring side condition proofs is that the side condition might not be provable and therefore the wrong rule has been applied. If we leave that to the user, we'll still have to be very careful which rewrite rules we supply.

For the set (A): that set basically implements a rewrite strategy, i.e. it could go for left-most occurrence in a bind, right-most occurrence in a bind, deepest, shallowest, etc. So it might make sense to have several sets of these to select a strategy (in addition to demanding that a specific rewrite is successful).

Does your workflow start with a `corres` goal or a `monadic_rewrite` goal (i.e. are we using rewrite rules or are we proving rewrite rules)? It looks to me like the digging-into-subterm strategy would be basically the same for both cases, and for both cases previously proved rewrites would be useful, so maybe the same tool could be useful for both.

The more I think about it, the more `monadic_rewrite` is just `corres` with identity as state relation and `(=)` as return value relation. That might actually be a useful theorem to prove to make use of existing corres rules as rewrites.

lib/MonadicRewrite.thy
lsf37: Anti-symmetry for a relation `<=_r` is the rule`[| x <=_r y; y <=_r x |] ==> x = y`

lib/MonadicRewrite.thy
lsf37: For this PR we shouldn't do anything yet, because it'd be a lot of renaming. Not even sure that it really is worth doing, it just bugs me a bit.

The point of inconsistency is that in corres and wp, the "weaken" rules refer to the preconditions + postcondition + return value relation. Here the "imp" rule does that, and the "weaken" rule refers to flags instead.

lib/MonadicRewrite.thy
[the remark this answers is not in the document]
lsf37: Hm, there might be something going in the wrong direction in my thinking. Let me check in Isabelle.

Xaphiosis:
> The PR will stay around when closed so things will still be there for reference, but it probably makes sense to put the idea comments into an issue so we can keep discussing there.

OK, I'll create an issue somewhere internally for monadic rewrite thoughts and link this PR.
 
> Interesting, and much more ambitious than mine. If we compare this to term rewriting the set (A) is selecting the sub term you want to rewrite, (B) is the rewrite set (it'll be hard to demand that all rules of a set have been used in general, i.e. too much possible backtracking probably, but I can see a few straightforward cases), and (C) are solving side conditions on the rewrite rules.

I don't think of it being very ambitious, more a case of shoving steps together:
`(A?+), B1, (A?+), B2, (A?+), ..., refl` and get left with `C`. I did not envision backtracking when working this high up, I'm still thinking in terms of "I have this thing on the screen, I want to rewrite these bits in it with these rules, go!" as opposed to "I have these generic sets or rules and this current state, do your best". I really like your rewrite rule analogy, that's a neat insight.
 
> In the Isabelle simplifier, to apply a conditional rewrite it first has to solve the side condition. In our case that might be too hard, although it might make sense for some trivial and almost trivial ones. The problem with deferring side condition proofs is that the side condition might not be provable and therefore the wrong rule has been applied. If we leave that to the user, we'll still have to be very careful which rewrite rules we supply.

When talking with Thomas last time (or you?), that person cautioned about having rules that only apply if the side-condition is solvable. You'd have to be in a position where multiple potential monadic_rewrite rules could apply, and only one of them is the one you'd want. Do you really see that happening?

> For the set (A): that set basically implements a rewrite strategy, i.e. it could go for left-most occurrence in a bind, right-most occurrence in a bind, deepest, shallowest, etc. So it might make sense to have several sets of these to select a strategy (in addition to demanding that a specific rewrite is successful).

That's also a good observation. I'll have to do an examination of how these are used. The bind_head/bind_tail/trans/refl etc stuff seems completely arbitrary to me at the moment, and the introduction/handling of alternative is even stranger.

> Does your workflow start with a `corres` goal or a `monadic_rewrite` goal (i.e. are we using rewrite rules or are we proving rewrite rules)? It looks to me like the digging-into-subterm strategy would be basically the same for both cases, and for both cases previously proved rewrites would be useful, so maybe the same tool could be useful for both.

I think my initial idea is to start with a monadic_rewrite goal, i.e. you have to jump from a corres goal in some way (usually "I want to rewrite (this part of?) the concrete side"). When and how to jump is an even bigger question than the automation once you do jump.

> The more I think about it, the more `monadic_rewrite` is just `corres` with identity as state relation and `(=)` as return value relation. That might actually be a useful theorem to prove to make use of existing corres rules as rewrites.

It is, but those `E` and `F` flags... don't they allow more flexibility, or is that already included in corres_underlying machinery (knowledge hazy here)? If it's subsumed, I could prove it, and that would be a useful back-and-forth, driving consolidation, especially if corres automation/tech could improve. Good idea.

lsf37:
> > The more I think about it, the more `monadic_rewrite` is just `corres` with identity as state relation and `(=)` as return value relation. That might actually be a useful theorem to prove to make use of existing corres rules as rewrites.
> 
> It is, but those `E` and `F` flags... don't they allow more flexibility, or is that already included in corres_underlying machinery (knowledge hazy here)? If it's subsumed, I could prove it, and that would be a useful back-and-forth, driving consolidation, especially if corres automation/tech could improve. Good idea.

`E` is not there in `corres_underlying`, but `F` is. A `monadic_rewrite` with `E = true` would imply `corres_underlying` with the right flags, but not the other way around.

lsf37:
> > In the Isabelle simplifier, to apply a conditional rewrite it first has to solve the side condition. In our case that might be too hard, although it might make sense for some trivial and almost trivial ones. The problem with deferring side condition proofs is that the side condition might not be provable and therefore the wrong rule has been applied. If we leave that to the user, we'll still have to be very careful which rewrite rules we supply.
> 
> When talking with Thomas last time (or you?), that person cautioned about having rules that only apply if the side-condition is solvable. You'd have to be in a position where multiple potential monadic_rewrite rules could apply, and only one of them is the one you'd want. Do you really see that happening?

I was more thinking out loud working out the relationship to "real" rewriting. It all depends on how hard those side conditions are to prove. If they are mostly trivial, we could demand side conditions first and gain more simp-like automation where you throw in a set of rewrite rules and they all get applied whenever they fit. If the side conditions are potentially hard, that doesn't work, and the consequence is that we have to carefully manually select which rules apply. I suspect you're right and we will have to remain on the select-carefully side, i.e. provide one specific rewrite rule to apply.

lib/MonadicRewrite.thy
```diff
@@ -102,86 +234,25 @@ proof -
...
-  done
-
-lemma monadic_rewrite_drop_modify:
-  "monadic_rewrite F E (\<lambda>s. f s = s) (modify f >>= v) (v ())"
-  by (simp add: monadic_rewrite_def bind_def simpler_modify_def)
+  = monadic_rewrite_symb_exec_pre[OF _ _ _ eq_UNIV_imp_helper, where P=\<top>,
+                                  simplified, THEN monadic_rewrite_trans]
 
 lemma monadic_rewrite_symb_exec_r:
   "\<lbrakk> \<And>s. \<lbrace>(=) s\<rbrace> m \<lbrace>\<lambda>r. (=) s\<rbrace>; no_fail P' m;
```
michaelmcinerney, line 252: Use invariant notation here too?

lib/MonadicRewrite.thy
```diff
@@ -255,29 +326,117 @@ lemma monadic_rewrite_symb_exec_l'_preserve_names:
...
+  = monadic_rewrite_symb_exec_l'_preserve_names[where P'="\<top>" and F=True, simplified]
 
 lemmas monadic_rewrite_symb_exec_l
-    = monadic_rewrite_symb_exec_l''[where F=True and P'=\<top>, simplified]
-      monadic_rewrite_symb_exec_l''[where F=False, simplified simp_thms]
+  = monadic_rewrite_symb_exec_l''[where F=True and P'=\<top>, simplified]
+    monadic_rewrite_symb_exec_l''[where F=False, simplified simp_thms]
+
+lemma monadic_rewrite_symb_exec_l_known:
+  "\<lbrakk> \<And>s. \<lbrace>(=) s\<rbrace> m \<lbrace>\<lambda>r. (=) s\<rbrace>; empty_fail m;
```
michaelmcinerney, line 341: Invariant notation?

lib/MonadicRewrite.thy
```diff
@@ -255,29 +326,117 @@ lemma monadic_rewrite_symb_exec_l'_preserve_names:
...
+lemmas monadic_rewrite_named_if
+  = monadic_rewrite_if[where Q="(=) s" and R="(=) s", simplified] for s
+
+lemma monadic_rewrite_if_lhs:
+  "\<lbrakk> P \<Longrightarrow> monadic_rewrite F E Q b a; \<not> P \<Longrightarrow> monadic_rewrite F E R c a \<rbrakk>
+   \<Longrightarrow> monadic_rewrite F E (\<lambda>s. (P \<longrightarrow> Q s) \<and> (\<not> P \<longrightarrow> R s)) (If P b c) a"
+  by (cases P, simp_all)
+
+lemma monadic_rewrite_if_known:
+  "monadic_rewrite F E ((\<lambda>s. C = X) and \<top>) (if C then f else g) (if X then f else g)"
```
michaelmcinerney, line 373: Is the `and \<top>` doing anything here or can it be removed?

lib/MonadicRewrite.thy
```diff
@@ -403,72 +522,88 @@ lemma monadic_rewrite_gets_the_walk:
...
 
-lemma monadic_rewrite_weaken2:
-  "monadic_rewrite F E P f g
-     \<Longrightarrow> monadic_rewrite F' E' ((\<lambda>_. (F \<longrightarrow> F') \<and> (E' \<longrightarrow> E)) and P) f g"
-  apply (rule monadic_rewrite_gen_asm)
-  apply (rule monadic_rewrite_weaken[where F=F and E=E])
-  apply auto
+lemma monadic_rewrite_gets_the_bind:
+  assumes mr: "(\<And>v. monadic_rewrite F E (Q v) (g v) m)"
+  shows   "monadic_rewrite F E (\<lambda>s. f s \<noteq> None \<and> Q (the (f s)) s) (gets_the f >>= (\<lambda>x. g x)) m"
```
michaelmcinerney, line 532: Nitpick: did you want the extra spaces after the `shows`?

lib/MonadicRewrite.thy
```diff
@@ -403,72 +522,88 @@ lemma monadic_rewrite_gets_the_walk:
...
-  by (cases P, simp_all)
+lemma monadic_rewrite_weaken_failure:
+  "\<lbrakk> monadic_rewrite True True P f f'; no_fail P' f; no_fail Q' f' \<rbrakk>
+   \<Longrightarrow> monadic_rewrite F E (P and P' and Q') f f'"
+  by (clarsimp simp: monadic_rewrite_def no_fail_def)
 
-lemma monadic_rewrite_to_eq:
-  "monadic_rewrite False True \<top> f g ==> f = g"
-  by (simp add: monadic_rewrite_def fun_eq_iff)
+lemma monadic_rewrite_is_valid:
```
michaelmcinerney, line 597: Did you want this one a bit earlier with the other rules for WP proofs?

lib/MonadicRewrite.thy
```diff
@@ -403,72 +522,88 @@ lemma monadic_rewrite_gets_the_walk:
...
+  "\<lbrakk> monadic_rewrite False False P' f f'; \<lbrace>P\<rbrace> do x <- f; g x od \<lbrace>Q\<rbrace> \<rbrakk>
+   \<Longrightarrow> \<lbrace>P and P'\<rbrace> do x <- f'; g x od \<lbrace>Q\<rbrace>"
+  by (fastforce simp: monadic_rewrite_def valid_def bind_def)
+
+text \<open>corres / refinement\<close>
 
 lemma corres_underlyingI:
```
michaelmcinerney, line 604: Would this rule and the following be better in `Corres_UL.thy` or something? Or is that bad because of imports?

michaelmcinerney (approved):
The cleanup looks good! I'll try to think a bit about the ideas mentioned in this PR, though I have limited experience doing long monadic rewrite proofs

michaelmcinerney:
You might find some of the `forward_inv_step` work (in `NonDetMonadVCG.thy`) and maybe `repeat_unless` (in `Eisbach_Methods.thy`) to be interesting. This was intended to help with forward reasoning in Hoare triple proofs, by just stepping over the first function in a bind while keeping the precondition the same. 

If we can get forward stepping or some kind of automation working with some rules to apply when possible, this could work nicely. I've often found it difficult to get the `bind` in the right spot for a rule to apply, but with some backtracking (like with the `,` trick we saw in the last few months) that might be doable. 

I'm not familiar with `ccorres_rewrite`, so I didn't follow that part of the discussion, but if you know of a good application of that, I'd be happy to take a look

lib/MonadicRewrite.thy
Xaphiosis: thanks!

lib/MonadicRewrite.thy
Xaphiosis: Spurred on by your suggestions, I went over the file looking for inv notation. Next push these should be resolved.

lib/MonadicRewrite.thy
Xaphiosis: I honestly have no idea. It can definitely be removed from the lemma, but as to how it is used in practice during monadic rewrite, I have not a clue.

lib/MonadicRewrite.thy
Xaphiosis: Thanks, a bunch of these are very old.

lib/MonadicRewrite.thy
Xaphiosis: Moved

lib/MonadicRewrite.thy
Xaphiosis: Very good point. I'll have a look.

Xaphiosis: Done

Xaphiosis:
> You might find some of the `forward_inv_step` work (in `NonDetMonadVCG.thy`) and maybe `repeat_unless` (in `Eisbach_Methods.thy`) to be interesting. This was intended to help with forward reasoning in Hoare triple proofs, by just stepping over the first function in a bind while keeping the precondition the same.

I'm aware of `repeat_unless`, but looking at `forward_inv_step`, I still maintain that for forward proof on `wp` goals we should not be one-stepping, but instead breaking off a whole chunk before where a specific expression matches, so you end up with two pieces, one of which is boring/predictable. I may be wrong, and I haven't written this tech yet... some day, I guess. Your suggestion might definitely be useful here though, thank you. The "iterate over the skeleton looking for rules to apply" paradigm happens a lot.

> If we can get forward stepping or some kind of automation working with some rules to apply when possible, this could work nicely. I've often found it difficult to get the `bind` in the right spot for a rule to apply, but with some backtracking (like with the `,` trick we saw in the last few months) that might be doable.

That is approximately the idea, and as you say the exact parenthesising of `bind` can cause problems for applying rules. Can you say more what you mean by "the `,` trick"?

> I'm not familiar with `ccorres_rewrite`, so I didn't follow that part of the discussion, but if you know of a good application of that, I'd be happy to take a look

You can find ccorres_rewrite as part of CRefine. Roughly speaking it is a zipper over SIMPL where it tries to apply a bunch of obvious "simplification" rules that simp can't because we're in a deep embedding.

michaelmcinerney:
> That is approximately the idea, and as you say the exact parenthesising of `bind` can cause problems for applying rules. Can you say more what you mean by "the `,` trick"?

I suppose it's not a trick, but I was unaware that `,` will cause backtracking, and so can be useful when we do have rules that only apply when the `do` bundle is broken up in a very particular way. Maybe `,` can be part of the automation to help with this issue

lib/MonadicRewrite.thy
corlewis: No constructive suggestions, but I agree with Gerwin that the name `monadic_rewrite_weaken` leads to me expecting that the precondition is what will be weakened.

lib/MonadicRewrite.thy
```diff
@@ -54,34 +133,87 @@ lemma monadic_rewrite_bind:
...
+                      split: sum.split)
+  apply simp
   done
 
+(* FIXME: poorly named, super-specific (could do this with maybe one bind?), used in Ipc_C *)
+lemma monadic_rewrite_do_flip:
+  "monadic_rewrite E F P (do c \<leftarrow> j; a \<leftarrow> f; b \<leftarrow> g c; return (a, c) od)
+                         (do c \<leftarrow> j; b \<leftarrow> g c; a \<leftarrow> f; return (a, c) od)
+   \<Longrightarrow> monadic_rewrite E F P (do c \<leftarrow> j; a \<leftarrow> f; b \<leftarrow> g c; h a c od)
+                            (do c \<leftarrow> j; b \<leftarrow> g c; a \<leftarrow> f; h a c od)"
```
corlewis, line 186: With this being so specific, would it make sense for it to stay in `crefine`? I guess the other goal is to minimise duplication, with the only arch-generic place to then put it being `Move_C` which doesn't make sense if we don't actually want to move it.

lib/MonadicRewrite.thy
```diff
@@ -403,86 +522,85 @@ lemma monadic_rewrite_gets_the_walk:
...
+     \<lbrace>P'\<rbrace> f \<lbrace>Q\<rbrace>, -;
+     F \<longrightarrow> no_fail P'' f \<rbrakk>
+   \<Longrightarrow> \<lbrace>P and P' and P''\<rbrace> g \<lbrace>Q\<rbrace>, -"
+  by (simp add: validE_R_def validE_def monadic_rewrite_refine_valid)
 
-lemmas monadic_rewrite_named_if
-    = monadic_rewrite_if[where Q="(=) s" and R="(=) s", simplified] for s
+lemma monadic_rewrite_is_valid:
+  "\<lbrakk> monadic_rewrite False False P' f f'; \<lbrace>P\<rbrace> do x <- f; g x od \<lbrace>Q\<rbrace> \<rbrakk>
+   \<Longrightarrow> \<lbrace>P and P'\<rbrace> do x <- f'; g x od \<lbrace>Q\<rbrace>"
```
corlewis, line 583: This name doesn't really make sense to me, and it feels like another of those very specific lemmas which aren't obviously needed.

lib/MonadicRewrite.thy
```diff
@@ -507,39 +625,27 @@ next
...
-lemma monadic_rewrite_refine_validE_R:
-  "monadic_rewrite F E P f g
-    \<Longrightarrow> \<lbrace>P'\<rbrace> f \<lbrace>Q\<rbrace>, -
-    \<Longrightarrow> F \<longrightarrow> no_fail P'' f
-    \<Longrightarrow> \<lbrace>P and P' and P''\<rbrace> g \<lbrace>Q\<rbrace>, -"
-  by (simp add: validE_R_def validE_def monadic_rewrite_refine_valid)
+lemma monadic_rewrite_corres2:
+  "\<lbrakk> monadic_rewrite False E Q a a';
+     corres_underlying R False F r P P' a' c \<rbrakk>
+   \<Longrightarrow> corres_underlying R False F r (P and Q) P' a c"
```
corlewis, line 637: Doesn't have to be as part of this PR, but for consistency with the previous lemma should this be `monadic_rewrite_corres_lhs`?

corlewis (approved):
Very nice cleanup so far!

lib/MonadicRewrite.thy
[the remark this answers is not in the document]
lsf37: To report back: anti-symmetry doesn't hold if F is true, because of the way monadic_rewrite is defined (it does for F = false).

`monadic_rewrite` (for E=false) is not actually `no_fail P f --> refined_by P f g` where `refined_by` would be the subset part of the `monadic_rewrite` definition. Instead it makes a weaker assumption on `f`, which leads to a stronger overall statement. The weaker assumption is that `f` does not have to satisfy all of `no_fail P`; instead absence of failure in `monadic_rewrite` is only relevant for states where `g` would otherwise not refine `f`.

My gut feeling is that the slightly weaker `no_fail P f --> refined_by P f g` is what we actually want, because it has nicer properties and relates nicely to existing concepts. There are more `f` and `g` for which `no_fail P f --> refined_by P f g` holds than for `monadic_rewrite True False f g`, but I doubt there is any real difference in practice.

lsf37:
> > > The more I think about it, the more `monadic_rewrite` is just `corres` with identity as state relation and `(=)` as return value relation. That might actually be a useful theorem to prove to make use of existing corres rules as rewrites.
> > 
> > It is, but those `E` and `F` flags... don't they allow more flexibility, or is that already included in corres_underlying machinery (knowledge hazy here)? If it's subsumed, I could prove it, and that would be a useful back-and-forth, driving consolidation, especially if corres automation/tech could improve. Good idea.
> 
> `E` is not there in `corres_underlying`, but `F` is. A `monadic_rewrite` with `E = true` would imply `corres_underlying` with the right flags, but not the other way around.

To report back on this bit as well: `corres_underlying` is not equivalent to `monadic_rewrite`, because `corres_underlying` is missing the implication that failure of `g` should imply failure of `f`. We use `corres_underlying` mostly with the requirement that `g` should not fail at all, so that absence doesn't matter for our normal refinement proofs, but I think it would actually be nicer to adjust `corres_underlying` to include that implication (just so that it is closer to actual refinement).

Xaphiosis:
OK, I'm going to merge this. Corey, your comments are good too and I'll try collate the TODOs for another round, likely after my "how is it used" exploration round, because I can't currently answer questions like the `gen_asm` being on the other side to usual.

PR #489, opened 2022-06-15
`cstate_relation`

michaelmcinerney:
This updates `cstate_relation`, and fixes breakages in `Wellformed_C.thy`.

### DIFF
--- a/proof/crefine/RISCV64/StateRelation_C.thy
+++ b/proof/crefine/RISCV64/StateRelation_C.thy
@@ -163,6 +163,16 @@ abbreviation
   where
   "map_to_ntfns hp \<equiv> projectKO_opt \<circ>\<^sub>m hp"
 
+abbreviation map_to_scs ::
+  "(obj_ref \<rightharpoonup> Structures_H.kernel_object) \<Rightarrow> obj_ref \<rightharpoonup> sched_context"
+  where
+  "map_to_scs hp \<equiv> projectKO_opt \<circ>\<^sub>m hp"
+
+abbreviation map_to_replies ::
+  "(obj_ref \<rightharpoonup> Structures_H.kernel_object) \<Rightarrow> obj_ref \<rightharpoonup> reply"
+  where
+  "map_to_replies hp \<equiv> projectKO_opt \<circ>\<^sub>m hp"
+
 abbreviation
   map_to_ptes :: "(machine_word \<rightharpoonup> Structures_H.kernel_object) \<Rightarrow> machine_word \<rightharpoonup> pte"
   where
@@ -267,12 +277,14 @@ where
      = (tsType_CL (fst ts') = scast ThreadState_Inactive)"
 | "cthread_state_relation_lifted (Structures_H.IdleThreadState) ts'
      = (tsType_CL (fst ts') = scast ThreadState_IdleThreadState)"
-| "cthread_state_relation_lifted (Structures_H.BlockedOnReply) ts'
-     = (tsType_CL (fst ts') = scast ThreadState_BlockedOnReply)"
-| "cthread_state_relation_lifted (Structures_H.BlockedOnReceive oref cg) ts'
+| "cthread_state_relation_lifted (Structures_H.BlockedOnReply replyobj) ts'
+     = (tsType_CL (fst ts') = scast ThreadState_BlockedOnReply
+        \<and> option_to_0 replyobj = replyObject_CL (fst ts'))"
+| "cthread_state_relation_lifted (Structures_H.BlockedOnReceive oref cg replyobj) ts'
      = (tsType_CL (fst ts') = scast ThreadState_BlockedOnReceive
         \<and> oref = blockingObject_CL (fst ts')
-        \<and> cg = to_bool (blockingIPCCanGrant_CL (fst ts')))"
+        \<and> cg = to_bool (blockingIPCCanGrant_CL (fst ts'))
+        \<and> option_to_0 replyobj = replyObject_CL (fst ts'))"
 | "cthread_state_relation_lifted (Structures_H.BlockedOnSend oref badge cg cgr isc) ts'
      = (tsType_CL (fst ts') = scast ThreadState_BlockedOnSend
         \<and> oref = blockingObject_CL (fst ts')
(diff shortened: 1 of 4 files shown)


proof/crefine/RISCV64/StateRelation_C.thy
```diff
@@ -351,18 +363,78 @@ definition
...
+    "dyn_array_list_rel h rel [] p = True"
+  | "dyn_array_list_rel h rel (a # as) p
+      = (\<exists>v. h p = Some v \<and> rel a v \<and> dyn_array_list_rel h rel as (p +\<^sub>p 1))"
+
+definition sc_ptr_to_crefill_ptr :: "obj_ref \<Rightarrow> refill_C ptr" where
+  "sc_ptr_to_crefill_ptr p \<equiv> Ptr (p + of_nat sizeof_sched_context_t)"
+
+definition refill_buffer_relation ::
+  "(obj_ref \<rightharpoonup> kernel_object) \<Rightarrow> heap_raw_state \<Rightarrow> cghost_state \<Rightarrow> bool"
+  where
```
lsf37, line 413: Could use the type abbreviation `kheap` for the first parameter. We don't use it very often outside ASpec, but it would probably make the type fit into one line.

proof/crefine/RISCV64/StateRelation_C.thy
```diff
@@ -351,18 +363,78 @@ definition
...
+
+definition crefill_relation :: "refill \<Rightarrow> refill_C \<Rightarrow> bool" where
+  "crefill_relation ar cr \<equiv> rTime ar = rTime_C cr \<and> rAmount ar = rAmount_C cr"
+
+primrec dyn_array_list_rel ::
+  "('c :: c_type) typ_heap \<Rightarrow> ('a \<Rightarrow> 'c \<Rightarrow> bool) \<Rightarrow> 'a list \<Rightarrow> 'c ptr \<Rightarrow> bool"
+  where
+    "dyn_array_list_rel h rel [] p = True"
+  | "dyn_array_list_rel h rel (a # as) p
+      = (\<exists>v. h p = Some v \<and> rel a v \<and> dyn_array_list_rel h rel as (p +\<^sub>p 1))"
```
lsf37, line 406: indentation is off in this one, the `|` should go into column 0, and the term strings start at column 2 (under the `where`)

proof/crefine/RISCV64/StateRelation_C.thy
```diff
@@ -351,18 +363,78 @@ definition
...
+  "(obj_ref \<rightharpoonup> kernel_object) \<Rightarrow> heap_raw_state \<Rightarrow> cghost_state \<Rightarrow> bool"
+  where
+  "refill_buffer_relation ah ch gs \<equiv>
+     let abs_sc_hp = map_to_scs ah;
+         crefill_hp = clift ch;
+         lens_hp = refill_buffer_lengths gs
+     in
+       dom abs_sc_hp = dom lens_hp
+     \<and> dom crefill_hp
+       = \<Union> ((\<lambda>p. set (map (\<lambda>i. sc_ptr_to_crefill_ptr p +\<^sub>p int i) [0..< the (lens_hp p)])) ` (dom abs_sc_hp))
```
lsf37, line 421: Looking at what Isabelle echoes back for the definition term, there is a slightly nicer way to write this (parsing to the exact same term):

```
∧ dom crefill_hp
       = (⋃p : dom abs_sc_hp. set  (map (λi. sc_ptr_to_crefill_ptr p +⇩p int i) [0..< the (lens_hp p)]))
```

proof/crefine/RISCV64/StateRelation_C.thy
lsf37: Actually, ignore me, that would be the ASpec `kernel_object`, not the Haskell `kernel_object`

proof/crefine/RISCV64/StateRelation_C.thy
```diff
@@ -351,18 +363,78 @@ definition
...
+       option_to_ptr (replyTCB areply) = replyTCB_C creply
+     \<and> option_to_ptr (replyPrev areply)
+        = ((Ptr \<circ> callStackPtr_CL \<circ> call_stack_lift \<circ> replyPrev_C) creply :: reply_C ptr)
+     \<and> isHead (replyNext areply) = to_bool ((isHead_CL \<circ> call_stack_lift \<circ> replyNext_C) creply)
+     \<and> (if bound (replyNext areply)
+        then if isHead (replyNext areply)
+             then theHeadScPtr (replyNext areply)
+             else theReplyNextPtr (replyNext areply)
+        else 0)
+       = (callStackPtr_CL \<circ> call_stack_lift \<circ> replyNext_C) creply"
```
lsf37, line 437: Stupid question: what is `call_stack_lift`? I can't find a definition.

lsf37:
Nice, I think we're almost there, apart from my confusion about `call_stack_lift`, I had only cosmetics.

proof/crefine/RISCV64/StateRelation_C.thy
michaelmcinerney: I can't open up CRefine just yet to double check and point you to the exact location, but `call_stack` is a type defined via a C `block` statement and it's in the same file as the others, like `endpoint_lift`. Can you find it now?

lsf37: Right, thanks for that, and of course that makes perfect sense. All good.

lsf37 (approved):

spec/cspec/RISCV64/Kernel_C.thy
```diff
@@ -26,7 +26,7 @@ declare [[populate_globals=true]]
 context begin interpretation Arch . (*FIXME: arch_split*)
 
 type_synonym cghost_state = "(machine_word \<rightharpoonup> vmpage_size) * (machine_word \<rightharpoonup> nat)
-    * ghost_assertions"
+    * ghost_assertions * (machine_word \<rightharpoonup> nat)"
```
corlewis, line 29: While you're here it would be nice to fix up this indentation. If it fits it would just be moving the whole type to the second line. If not, I guess something like this?
```
type_synonym cghost_state =
  "(machine_word ⇀ vmpage_size) * (machine_word ⇀ nat) * ghost_assertions *
   (machine_word ⇀ nat)"
```

proof/crefine/RISCV64/StateRelation_C.thy
corlewis: The indentation also seems a bit off here; the `∧` shouldn't be in the same column as the `in`. My preference would be for either the same column as the previous conjunct, or at the end of the previous line.

corlewis (approved):

lsf37:
It'd be good to merge this one tomorrow. I forgot that the seL4 PR seL4/seL4#872 needs the constants defined here for CKernel to work, so currently the MCS CKernel proofs are broken until the PR is merged.

spec/cspec/RISCV64/Kernel_C.thy
michaelmcinerney: Done!

PR #490, opened 2022-06-21
update to ghc 9.0.2

lsf37:
This is mostly to reduce build time/cache size for docker images.

Main change with ghc 9.0.2 is that `fail` now requires the class `MonadFail` instead of just `Monad` and a few small surface syntax tweaks. As far as I can see, there is no impact on the Isabelle side.

### DIFF
--- a/spec/haskell/.gitignore
+++ b/spec/haskell/.gitignore
@@ -8,6 +8,8 @@
 
 **.lhs-boot
 /dist/
+/dist-newstyle/
+/cabal.project.local*
 
 # cabal dependencies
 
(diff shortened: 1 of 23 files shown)


lsf37:
`stack` is great they said, `stack` makes you Haskell builds reproducible and easy they said.

Of course it works locally for me and not here. If anyone has an idea how to specify those dependencies so that they are not circular, please help :)

lsf37:
Switching to a more recent version than what I had in my original commits allows `stack` to build `cabal-install` again. Hopefully.

PR #491, opened 2022-07-01
Merge `master` into `rt`

michaelmcinerney:


michaelmcinerney:
It seems the automatic merge has been too invasive for `MonadicRewrite.thy`, so I will (soon) update the PR to take `MonadicRewrite.thy` verbatim from `master`, and then make a separate commit adding the rules for `monadic_rewrite` which are new for `rt`

michaelmcinerney:
Actually, the automatic merge didn't mangle `MonadicRewrite.thy` at all. The only difference is that there are two new rules in the version on the `rt` branch. I think I will make a PR for `master` which adds these, and which adds a couple other rules I would like to have. Then I can make another small PR for the `rt` branch to again match them up

lsf37:
> Actually, the automatic merge didn't mangle `MonadicRewrite.thy` at all. The only difference is that there are two new rules in the version on the `rt` branch. I think I will make a PR for `master` which adds these, and which adds a couple other rules I would like to have. Then I can make another small PR for the `rt` branch to again match them up

If possible, you should cherry-pick the `rt` commit that introduces the 2 rules for your master PR, and make a separate commit for any further additions. That way the next small merge should be conflict free.

lsf37 (approved):
Nice! It looks like Ryan's getPageAddr update merged relatively smoothly.

michaelmcinerney:
> If possible, you should cherry-pick the `rt` commit that introduces the 2 rules for your master PR, and make a separate commit for any further additions. That way the next small merge should be conflict free.

The commit with the monadic rewrite rule for `corres` is in a commit with another rule, and I couldn't track down the  `monadic_rewrite_sym` rule (its first appearance is in a merge conflict commit, and I couldn't find it earlier than that). Should I leave it as is?

lsf37:
> > If possible, you should cherry-pick the `rt` commit that introduces the 2 rules for your master PR, and make a separate commit for any further additions. That way the next small merge should be conflict free.
> 
> The commit with the monadic rewrite rule for `corres` contains another rule, and I couldn't track down the `monadic_rewrite_sym` rule (its first appearance is in a merge conflict commit, and I couldn't find it earlier than that). Should I leave it as is?

Yes, if it can't be extracted as a commit, that's fine, we'll just deal with the potential merge conflict when it happens.

PR #492, opened 2022-07-04
Some monadic rewrite rules

michaelmcinerney:


lsf37 (approved):
Looks good from my side.

lib/MonadicRewrite.thy
```diff
@@ -32,6 +32,10 @@ lemma monadic_rewrite_refl_generic:
 
 lemmas monadic_rewrite_refl = monadic_rewrite_refl_generic[where P=\<top>]
 
+lemma monadic_rewrite_sym:
+  "monadic_rewrite False True P f g \<Longrightarrow> monadic_rewrite False True P g f"
+  by (simp add: monadic_rewrite_def)
```
Xaphiosis, line 37: This is really high up, given the `F E` flags are explicitly set, so it's a sym rule only for that case. Does it still hold if the `F` flag is left generic?

lib/MonadicRewrite.thy
```diff
@@ -178,6 +182,12 @@ lemma monadic_rewrite_named_bindE:
   apply simp
   done
 
+lemma monadic_rewrite_drop_return:
+  "monadic_rewrite F E P f g \<Longrightarrow> monadic_rewrite F E P (do y \<leftarrow> return r; f od) g"
```
Xaphiosis, line 186: suggest `return x` or something that hints at a value... `r` is usually a relation/set.

lsf37: also happier with `return x`, but I'd like to point out that a set should be `R` not `r` (a relation could be either).

lib/MonadicRewrite.thy
michaelmcinerney: I'm happy to move this. Where would be best, do you think?

Unfortunately, I couldn't see any generalisation of this for either the `F` or `E` flag

Xaphiosis: Well, if this is the only way symmetry works, then leave it where it is. If we have doubts, or there's something we're paving over, maybe dump it at the end of this section after `monadic_rewrite_weaken2`?

michaelmcinerney: I've taken another look and I'm sure that we can't generalise this any further, so I think I might leave it where it is

michaelmcinerney:
Looking at the output of the `lemmas` command more closely, it wasn't what I wanted. But what I did want is very straightforward and is now schematic in the flags, too

Xaphiosis:
> Looking at the output of the `lemmas` command more closely, it wasn't what I wanted. But what I did want is very straightforward and is now schematic in the flags, too

Good.
That would mean the `sym` rule is now unused? Do we still need/want it?

michaelmcinerney:
> Good. That would mean the `sym` rule is now unused? Do we still need/want it?

That's right, it's now currently unused in `master`. It's used in several places for MCS, so it might be useful. I'd be happy to remove it, but maybe it would be good to have this file in sync for both branches

Xaphiosis:
> > Good. That would mean the `sym` rule is now unused? Do we still need/want it?
> 
> That's right, it's now currently unused in `master`. It's used in several places for MCS, so it might be useful. I'd be happy to remove it, but maybe it would be good to have this file in sync for both branches

If it's used on MCS then we can leave it. The question is whether it's used well on MCS, or more like a hammer looking for nails. Probably can do that another time though.

PR #494, opened 2022-07-04
Setup instructions

michaelmcinerney:
Instructions mainly for macOS, and a small update for Python

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -49,6 +49,39 @@ sudo apt-get install \
     mlton-compiler haskell-stack
 ```
 
+### macOS packages
+
+These instructions use Homebrew, which can be installed from [their website][homebrewwebsite].
+The main packages that are needed are:
+
+```
+brew install git libxml2 ncurses librsvg dtc cmake ninja texlive rysnc python ccache \
+    zstd haskell-stack
+```
+
+The installation of mlton on Apple silicon is currently not well supported. One
+approach would be to compile mlton on another architecture and transfer it.
+
+To install the cross-compilers, run
+
+```
+brew install --cask gcc-arm-embedded
+
+brew install arm-none-eabi-gcc
+
+brew tap messense/macos-cross-toolchains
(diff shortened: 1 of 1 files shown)


docs/setup.md
```diff
@@ -49,13 +49,42 @@ sudo apt-get install \
...
+brew tap messense/macos-cross-toolchains
+brew install x86_64-unknown-linux-gnu aarch64-unknown-linux-gnu
+
+brew tap riscv/riscv
+brew install riscv-tools
+```
+
+Note that CMake will require the x86 compiler before it can be invoked.
+
+The instructions in the sections below should apply for macOS.
```
lsf37, line 77:
```suggestion
The instructions in the sections below should apply for both, Linux and macOS.
```

docs/setup.md
```diff
@@ -49,13 +49,42 @@ sudo apt-get install \
     mlton-compiler haskell-stack
 ```
 
+### macOS packages
+
+First install Homebrew from [their website][homebrewwebsite]. Most of the dependencies
+can be installed with
+
+``` brew install git libxml2 ncurses librsvg dtc cmake ninja texlive rysnc python ccache zstd
+```
```
lsf37, line 58: the command needs to be on a separate line from the code fence (```)

`haskell-stack` seems to be missing from the list, which brew also has.

michaelmcinerney: Should I include `haskell-stack` in the list here, or should I mention in the Haskell section that macOS users can just perform `brew install haskell-stack`?

docs/setup.md
michaelmcinerney: Actually, I'd prefer it without the comma, if you don't mind

docs/setup.md
lsf37: include it in the package list (same as for Linux)

docs/setup.md
lsf37: Happy to leave the comma up to you :-)

PR #495, opened 2022-07-05
Sorrying till `CSpace_C.thy`

michaelmcinerney:
This sorries until the beginning of `CSpace_C.thy`. In fact, only one sorry is added, namely `setCurThread_isolatable`, which is probably no longer true (if people think it is preferable, I can simply remove this lemma).

Many lemmas involving `setObject` and `getObject` have been updated now that they involve the reader monad.

### DIFF
--- a/proof/crefine/Move_C.thy
+++ b/proof/crefine/Move_C.thy
@@ -1151,19 +1151,6 @@ lemma ctes_of_valid_strengthen:
   apply fastforce
   done
 
-lemma finaliseCap_Reply:
-  "\<lbrace>Q (NullCap,NullCap) and K (isReplyCap cap)\<rbrace> finaliseCapTrue_standin cap is_final \<lbrace>Q\<rbrace>"
-  apply (rule NonDetMonadVCG.hoare_gen_asm)
-  apply (wpsimp simp: finaliseCapTrue_standin_def isCap_simps)
-  done
-
-lemma cteDeleteOne_Reply:
-  "\<lbrace>st_tcb_at' P t and cte_wp_at' (isReplyCap o cteCap) slot\<rbrace> cteDeleteOne slot \<lbrace>\<lambda>_. st_tcb_at' P t\<rbrace>"
-  apply (simp add: cteDeleteOne_def unless_def split_def)
-  apply (wp finaliseCap_Reply isFinalCapability_inv getCTE_wp')
-  apply (clarsimp simp: cte_wp_at_ctes_of)
-  done
-
 lemma cancelSignal_st_tcb':
   "\<lbrace>\<lambda>s. t\<noteq>t' \<and> st_tcb_at' P t' s\<rbrace> cancelSignal t ntfn \<lbrace>\<lambda>_. st_tcb_at' P t'\<rbrace>"
   apply (simp add: cancelSignal_def Let_def)
@@ -1172,26 +1159,14 @@ lemma cancelSignal_st_tcb':
   apply clarsimp
   done
 
-lemma cancelIPC_st_tcb_at':
-  "\<lbrace>\<lambda>s. t\<noteq>t' \<and> st_tcb_at' P t' s\<rbrace> cancelIPC t \<lbrace>\<lambda>_. st_tcb_at' P t'\<rbrace>"
-  apply (simp add: cancelIPC_def Let_def getThreadReplySlot_def locateSlot_conv)
-  apply (wp sts_pred_tcb_neq' getEndpoint_wp cteDeleteOne_Reply getCTE_wp' | wpc)+
-          apply (rule hoare_strengthen_post [where Q="\<lambda>_. st_tcb_at' P t'"])
-           apply (wp threadSet_st_tcb_at2)
-           apply simp
-          apply (clarsimp simp: cte_wp_at_ctes_of capHasProperty_def)
-         apply (wp cancelSignal_st_tcb' sts_pred_tcb_neq' getEndpoint_wp gts_wp' | wpc)+
-  apply clarsimp
-  done
-
 lemma suspend_st_tcb_at':
-  "\<lbrace>\<lambda>s. (t\<noteq>t' \<longrightarrow> st_tcb_at' P t' s) \<and> (t=t' \<longrightarrow> P Inactive)\<rbrace>
-  suspend t
-  \<lbrace>\<lambda>_. st_tcb_at' P t'\<rbrace>"
+  "\<lbrace>\<lambda>s. (t \<noteq> t' \<longrightarrow> st_tcb_at' P t' s) \<and> (t = t' \<longrightarrow> P Inactive)\<rbrace>
+   suspend t
+   \<lbrace>\<lambda>_. st_tcb_at' P t'\<rbrace>"
   apply (simp add: suspend_def unless_def)
   unfolding updateRestartPC_def
-  apply (cases "t=t'")
-   apply (simp | wp cancelIPC_st_tcb_at' sts_st_tcb')+
+  apply (cases "t = t'")
+   apply (wpsimp wp: cancelIPC_st_tcb_at' sts_st_tcb' hoare_drop_imp)+
   done
 
 lemma typ_at'_no_0_objD:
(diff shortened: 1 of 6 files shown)


proof/crefine/RISCV64/SR_lemmas_C.thy
```diff
@@ -1268,22 +1226,22 @@ lemma setObject_eq:
...
    apply simp
-  apply (erule ssubst [OF ob])
-  done
+  by metis
 
 lemma getObject_eq:
   fixes ko :: "'a :: pspace_storable"
   assumes x: "\<And>q n ko. loadObject p q n ko =
-                (loadObject_default p q n ko :: 'a kernel)"
+                (loadObject_default p q n ko :: ('a :: pspace_storable) kernel_r)"
```
lsf37, line 1234:
```suggestion
                (loadObject_default p q n ko :: 'a :: pspace_storable kernel_r)"
```

michaelmcinerney: I think this appears in a few places, so I can fix all of these up as you suggest

proof/crefine/RISCV64/IsolatedThreadAction.thy
```diff
@@ -1319,7 +1269,9 @@ lemma threadGet_isolatable:
 
 lemma setCurThread_isolatable:
   "thread_actions_isolatable idx (setCurThread t)"
-  by (simp add: setCurThread_def modify_isolatable)
+sorry (* Maybe no longer true, now that setCurThread contains the ready_qs_runnable assertion,
+which depends on the thread state, and could be modified by partial_overwrite *)
+(*  by (simp add: setCurThread_def modify_isolatable) *)
```
lsf37, line 1274: Interesting. If you're fairly sure it is no longer true, we should turn this into an `oops` instead of `sorry`. (Also, for either, please tag with a `FIXME RT` comment, so we can count the overall to-do items in CRefine)

It could become fairly annoying for `seCurThread` to not be re-orderable. Do you think it'd be possible to remove the assertion?

proof/crefine/RISCV64/SR_lemmas_C.thy
lsf37: Yes, please. Too many parentheses :-)

proof/crefine/RISCV64/IsolatedThreadAction.thy
lsf37: (It's perfectly fine to leave this as `oops` for now and continue sorrying the rest first. Maybe that'll also show how much impact its absence really has)

michaelmcinerney: Sure, I will `oops` this and see what happens. It's been a bit too long since I did anything to do with `ready_qs_runnable`, but I think its use is fairly limited and we might be able to improve things somehow

lsf37 (approved):

proof/crefine/RISCV64/SR_lemmas_C.thy
```diff
@@ -240,23 +240,6 @@ lemma cap_get_tag_ZombieCap:
...
 
-
-
-lemma cap_get_tag_ReplyCap:
-  assumes cr: "ccap_relation cap cap'"
-  shows "(cap_get_tag cap' = scast cap_reply_cap) =
-  (cap =
-      ReplyCap (ctcb_ptr_to_tcb_ptr (Ptr (cap_reply_cap_CL.capTCBPtr_CL (cap_reply_cap_lift cap'))))
-               (to_bool (capReplyMaster_CL (cap_reply_cap_lift cap')))
-               (to_bool (capReplyCanGrant_CL (cap_reply_cap_lift cap'))))"
```
corlewis, line 251: Out of curiosity, what has changed that we now don't need a lemma like this for the new ReplyCaps?

proof/crefine/RISCV64/IsolatedThreadAction.thy
```diff
@@ -661,14 +648,21 @@ lemma setObject_assert_modify:
...
+   apply (fastforce simp: obj_at_simps)
+  apply (clarsimp simp: setObject_def split_def exec_gets updateObject_default_def assert_opt_def
+                        assert_def alignCheck_assert
+                 split: option.splits)
+  apply (clarsimp simp: magnitudeCheck_assert2 assert_def gets_the_def gets_def
+                        get_def bind_def return_def assert_opt_def fail_def obj_at'_def
+                 split: option.splits)
   done
 
+
```
corlewis, line 665: Was this new line intentional?

proof/crefine/RISCV64/IsolatedThreadAction.thy
```diff
@@ -982,24 +918,38 @@ lemma tcbSchedEnqueue_tcbPriority[wp]:
...
-  apply (simp add: setThreadState_def setSchedulerAction_def)
-  apply (wp hoare_pre_cont[where a=rescheduleRequired])
-  apply (rule_tac Q="\<lambda>_. ?P and st_tcb_at' ((=) st) t" in hoare_post_imp)
-   apply (clarsimp split: if_split)
-   apply (clarsimp simp: obj_at'_def st_tcb_at'_def projectKOs)
-  apply (wp threadSet_pred_tcb_at_state)
-  apply simp
-  done
+crunches unbindMaybeNotification, blockedCancelIPC, replyRemoveTCB, cancelSignal
+  for tcbPriority_obj_at'[wp]: "obj_at' (\<lambda>tcb. P (tcbPriority tcb)) t'"
```
corlewis, line 922: It would be nice to move these new hoare triple lemmas about the design spec to `Move_C`. We're not completely consistent about this, but the intention would be to move them to Refine (and eventually maybe [DInvs](https://sel4.atlassian.net/browse/VER-1299)).

corlewis (approved):

proof/crefine/RISCV64/StateRelation_C.thy
```diff
@@ -651,7 +651,7 @@ abbreviation
 abbreviation
   "tcb_cte_array_relation astate cstate
     \<equiv> cvariable_array_map_relation (map_to_tcbs (ksPSpace astate))
-        (\<lambda>x. 6) cte_Ptr (hrs_htd (t_hrs_' cstate))"
+        (\<lambda>x. unat tcbCNodeEntries) cte_Ptr (hrs_htd (t_hrs_' cstate))"
```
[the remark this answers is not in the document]
michaelmcinerney: Yeah I only just learnt that you can in C give a name to the number of fields in a struct. I like it

proof/crefine/Move_C.thy
```diff
@@ -1206,26 +1181,11 @@ lemma ko_at'_not_NULL:
...
-  apply wp
-       apply (rule hoare_pre_cont)
-      apply (wpsimp simp: if_apply_def2
-                    wp: hoare_drop_imps hoare_vcg_disj_lift threadSet_tcbState_st_tcb_at')+
-  done
-
-lemma sts_running_ksReadyQueuesL2Bitmap[wp]:
-  "\<lbrace>\<lambda>s. P (ksReadyQueuesL2Bitmap s)\<rbrace>
-   setThreadState Structures_H.thread_state.Running t
-   \<lbrace>\<lambda>_ s. P (ksReadyQueuesL2Bitmap s)\<rbrace>"
```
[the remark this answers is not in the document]
michaelmcinerney: I think this was a duplicate

proof/crefine/RISCV64/SR_lemmas_C.thy
michaelmcinerney: Oh maybe I was too hasty in removing this. I'll check

michaelmcinerney: Yeah this shouldn't have been removed. I've now added something back and will include it in a push soon(ish)

proof/crefine/RISCV64/IsolatedThreadAction.thy
michaelmcinerney: Nope! Removed!

proof/refine/RISCV64/KHeap_R.thy
```diff
@@ -949,6 +949,11 @@ proof -
                   elim: cte_wp_atE' canonical_address_add)
 qed
 
+lemma obj_at'_is_canonical:
+  "\<lbrakk>pspace_canonical' s; obj_at' P t s\<rbrakk> \<Longrightarrow> canonical_address t"
```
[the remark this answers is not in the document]
michaelmcinerney: I'm fairly sure we don't have `canonical_address` for ARM. Having a look at the places where this rule is used, `simp` might be most useful, so I think I'll do that unless there are any objections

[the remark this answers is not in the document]
corlewis: Ah, looks like you're right and it's currently only for RISCV64 and X64. That probably explains why I'm not familiar with it.

proof/crefine/RISCV64/IsolatedThreadAction.thy
michaelmcinerney: I've added a `FIXME RT: move` for this and mentioned DInvs

proof/refine/RISCV64/KHeap_R.thy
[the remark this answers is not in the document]
lsf37: Yes, it's a 64bit concept. I would not recommend `[simp]` for this, because that fires too often. You will often have `canonical` in the goal, but no `obj_at` or `pspace_canonical`

PR #496, opened 2022-07-05
docs: add mlton for x86 Apple machines

lsf37:
`mlton` installation is only tricky on M1 currently. Intel machines have a brew package.


michaelmcinerney (approved):
👍

Issue #497, opened 2022-07-05
fix CI for SimplExportAndRefine on MCS


lsf37:
We're currently attempting to build `SimplExportAndRefine` and `CKernel` for the MCS config of seL4 from the master branch of l4v. This was convenient because it kept the CI workflow simple.

It also used to work fine, but now breaks since we have added annotations to the C code that need concepts only available on the `rt` branch.

This is issue for fixing the CI test so proof updates are deployed again to the manifest. An intermediate step could be to just disable this step. Longer term, we want to switch l4v to the `rt` branch. This might be as simple as supplying a branch argument to the test, but I'm sure complications will arise as they always do.

lsf37:
@mbrcknl do you have time to look into this, or should I have a go later this week? We're currently not deploying any more to the verification manifest, so we should not leave it too long.

mbrcknl:
I'll take a look now.

mbrcknl:
@lsf37 Thinking about how to do this...

Do you think it would be worth pulling the `mcs-export` job out into a separate workflow? Just thinking that proof-deploy.yml is the wrong place for it, since proof-deploy is supposed to be a final check on a manifest we already believe is deployable. And therefore it shouldn't contain things not related to the deployability of a manifest.

Taking it out would make things more complicated, since I'd still need separate workflows for the MCS export and decompilation to stay within the 6 hour budget. But it would limit the damage if something like this comes up again.

I think the way I would do it would be to add a new workflow (say binary-mcs.yml) which only runs on commits to `rt`. It would do the MCS export, and then trigger the decompilation workflow in much the same way that proof-deploy.yml triggers it for master.

At this stage, I probably wouldn't want to add it to the proof.yml workflow, since I wouldn't want it to run on every PR.

Would that make sense?

lsf37:
Yes, that sounds eminently reasonable.

mbrcknl:
@lsf37 Thinking about this some more...

IIUC, proof-deploy runs if there's a push to l4v master, or a manual bump of the devel.xml manifest. That's useful for BV, since it roughly corresponds to the cases BV should check for non-MCS.

But currently, the preprocess test only checks non-MCS configurations. So a manual bump is only needed when there is a change to the non-MCS parts of seL4 C code. Changes that only touch MCS parts will be automatically deployed to the default manifest. Do I read that right?

[...]

PR #498, opened 2022-07-06
ci: temporarily remove mcs-export from proof-deploy

mbrcknl:
This removes the mcs-export matrix job from the proof-deploy workflow,
as the first step towards solving seL4/l4v#497. This should unblock
verification manifest deployments.

The mcs-export job was added to the proof-deploy workflow to perform
SimplExportAndRefine for binary verification targets. It took a short
cut, using the master branch of l4v to perform SimplExportAndRefine for
MCS configurations, since there were no differences between rt and
master that were relevant to SimplExportAndRefine. This is no longer the
case, because MCS seL4 C code now contains C parser annotations that use
symbols only available in the rt branch of l4v.

We intend to add an equivalent job that uses the rt branch of l4v for
MCS SimplExportAndRefine, but are still working out the best way to do
that.

Signed-off-by: Matthew Brecknell <matt@kry10.com>

### DIFF
--- a/.github/workflows/proof-deploy.yml
+++ b/.github/workflows/proof-deploy.yml
@@ -59,44 +59,10 @@ jobs:
         name: logs-${{ matrix.arch }}
         path: logs.tar.xz
 
-  mcs-export:
-    name: MCS
-    needs: code
-    runs-on: ubuntu-latest
-    strategy:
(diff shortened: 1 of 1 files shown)


lsf37 (approved):
👍

PR #499, opened 2022-07-07
ci: add mcs-export workflow

mbrcknl:
This adds adds an mcs-export workflow, as a further step towards solving
issue #497. This replaces the mcs-export matrix job that was removed
from the proof-deploy workflow in #498.

This workflow has some similarities to the proof-deploy workflow:
- Both workflows check out a manifest, run some proofs, and trigger
  a binary verification workflow.
- Both workflows are triggered by an update to an l4v branch.
- Both workflows can be triggered by manual updates to the development
  verification-manifest, but not automatic updates due to a successful
  preprocess test.

The differences are:
- proof-deploy updates the default verification-manifest, whereas the
  mcs-export workflow does not.
- proof-deploy uses l4v master, whereas mcs-export uses the rt branch.

When MCS CRefine proofs are sufficiently advanced, the proof-deploy
workflow will need to be updated to check MCS proofs before updating the
default verification-manifest. At that point, this workflow could be
reintegrated into the proof-deploy workflow.

Signed-off-by: Matthew Brecknell <matt@kry10.com>

### DIFF
--- a/.github/workflows/mcs-export.yml
+++ b/.github/workflows/mcs-export.yml
@@ -1,0 +1,129 @@
+# Copyright 2021 Proofcraft Pty Ltd
+#
+# SPDX-License-Identifier: BSD-2-Clause
+
+# MCS SimplExportAndRefine for binary verification.
+#
+# The first step in preparing inputs to binary verification is to run
+# SimplExportAndRefine. This workflow performs SimplExportAndRefine for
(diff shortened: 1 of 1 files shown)


mbrcknl:
Switched to draft while thinking about https://github.com/seL4/l4v/issues/497.

PR #500, opened 2022-07-08
Sorry till end of `TcbQueue_C.thy`

michaelmcinerney:
This adds one (subgoal) sorry and gets everything else at least working until the end of `TcbQueue_C.thy`.

### DIFF
--- a/proof/crefine/RISCV64/TcbQueue_C.thy
+++ b/proof/crefine/RISCV64/TcbQueue_C.thy
@@ -775,7 +775,8 @@ proof -
         apply simp
        apply (fastforce simp: inj_eq split: if_split_asm)
       apply (clarsimp simp: last_conv_nth distinct_remove1_take_drop nth_eq_iff_index_eq inj_eq split: if_split_asm)
-       apply arith
+       apply (metis One_nat_def assms(1) last_tl length_tl less_nat_zero_code list.size(3)
+                    tcb_ptr_to_ctcb_ptr_eq tcb_queue_relationE')
       apply (simp add: nth_append min_def nth_eq_iff_index_eq)
       apply clarsimp
       apply arith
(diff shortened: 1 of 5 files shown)


proof/crefine/RISCV64/TcbQueue_C.thy
```diff
@@ -1392,21 +1394,23 @@ lemma rf_sr_tcb_update_no_queue:
...
       apply simp
-      apply (rule cendpoint_relation_upd_tcb_no_queues, assumption+)
+      apply (rule cnotification_relation_upd_tcb_no_queues, assumption+)
        subgoal by fastforce
       subgoal by fastforce
-     apply (erule iffD1 [OF cmap_relation_cong, OF refl refl, rotated -1])
-     apply simp
-     apply (rule cnotification_relation_upd_tcb_no_queues, assumption+)
+     subgoal by (clarsimp simp: map_comp_update projectKO_opt_sc typ_heap_simps'
+                                refill_buffer_relation_def)
```
corlewis, line 1409: Out of curiosity, what is this `subgoal` for? Is it a long running command?

corlewis (approved):
Looks good to me, and some nice cleanup as well!

proof/crefine/RISCV64/TcbQueue_C.thy
michaelmcinerney: The commented out line will break that subgoal into two. It's related to the references of `thread`. I think we don't want `thread` to be the idle thread, and it's more complicated for MCS because of the idle sc. I spent a while trying to figure it out but seeing that it's not currently mentioned in the proof code at the moment, maybe it won't be used in the end.

lsf37: yes that is fine for long(ish) running commands.

I'm not entirely sure about the indentation on this one. I think so far we mostly have `by` on the same line as `subgoal` when we don't need to break, but have it in the next line indented by 2 if things get bigger. This policy comes from Isar proofs where you would do the same fore `have` commands and similar if they are proved with `by`.

proof/crefine/RISCV64/TcbQueue_C.thy
```diff
@@ -1424,11 +1428,13 @@ lemma tcb_queue_relation_not_in_q:
     = tcb_queue_relation' nxtFn prvFn hp xs start end"
   by (rule tcb_queue_relation'_cong, auto)
 
+(* FIXME RT: unused? This might require an additional assumption that thread is not
+   the idle thread, and some further reasoning involving valid_idle' *)
 lemma rf_sr_tcb_update_not_in_queue:
```
lsf37, line 1433: It is indeed entirely unused. I would vote for removing it (and the derived `..2` variant)

lsf37 (approved):
:+1: Very few sorries for a sorry-ing run, but I'm not going to complain about proofs that work :-)

PR #502, opened 2022-07-15
riscv haskell+proofs: fix PageTablePTE encoding

lsf37:
According to the RISC-V spec, PageTablePTEs must have the access,
dirty, and user bits set to 0. This means that

- there is no user attribute that can be set on PageTablePTEs
  (removed from Haskell spec)
- the encoding for PageTablePTEs in C must have 0 in these fields
  instead of 1.

See PR seL4/seL4#880 for discussion and corresponding C changes.

Test with seL4/seL4#880

### DIFF
--- a/proof/crefine/RISCV64/Arch_C.thy
+++ b/proof/crefine/RISCV64/Arch_C.thy
@@ -1514,7 +1514,7 @@ lemma list_3_collapse:
   done
 
 lemma pde_case_isPageTablePDE:
-  "(case pte of PageTablePTE _ _ _ \<Rightarrow> P | _ \<Rightarrow> Q)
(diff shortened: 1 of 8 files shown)


Xaphiosis (approved):
These have no impact on actual functionality, but we've been carrying around the fact that these non-existent flags are false? How did you notice this?

lsf37:
> These have no impact on actual functionality, but we've been carrying around the fact that these non-existent flags are false? How did you notice this?

I didn't notice it :-) It turns out that quemu 7 notices it and fails sel4test, which apparently [ahmedcharles](https://github.com/ahmedcharles) was using, leading to https://github.com/seL4/seL4/pull/880

PR #503, opened 2022-07-18
Sorry `Detype_C`

michaelmcinerney:
This sorries `Detype_C.thy`, adding one sorry related to `refill_buffer_relation`.

The size of the scheduling context structs is updated in the abstract spec and the Haskell.

An assert was added to `deleteObjects` to help resolves an assumption in the `delete_locale` locale.

### DIFF
--- a/proof/crefine/RISCV64/Detype_C.thy
+++ b/proof/crefine/RISCV64/Detype_C.thy
@@ -238,12 +238,11 @@ qed
 lemma valid_untyped_pspace_no_overlap':
  assumes vuc: "s \<turnstile>' UntypedCap d ptr bits idx"
   and    idx: "idx< 2^ bits"
-  and psp_al: "pspace_aligned' s" "pspace_distinct' s"
+  and    psp: "pspace_aligned' s" "pspace_distinct' s" "pspace_bounded' s"
   shows  "pspace_no_overlap' (ptr + of_nat idx) bits s"
 
 proof -
-  note blah[simp del] =  atLeastAtMost_iff atLeastatMost_subset_iff atLeastLessThan_iff
-          Int_atLeastAtMost atLeastatMost_empty_iff split_paired_Ex
+  note [simp del] = atLeastAtMost_simps
 
   from vuc have al: "is_aligned ptr bits" and vu: "valid_untyped' d ptr bits idx s" and p0: "ptr \<noteq> 0"
     and wb: "bits < word_bits"
@@ -259,14 +258,13 @@ proof -
     done
 
   show "pspace_no_overlap' (ptr + of_nat idx) bits s"
-    using vuc idx psp_al
+    using vuc idx psp
     apply -
     apply (clarsimp simp:valid_cap'_def valid_untyped'_def pspace_no_overlap'_def)
     apply (drule_tac x = x in spec)
     apply (frule(1) pspace_alignedD')
     apply (frule(1) pspace_distinctD')
-    apply (clarsimp simp:ko_wp_at'_def obj_range'_def p_assoc_help)
-    done
+    by (force simp:ko_wp_at'_def obj_range'_def p_assoc_help pspace_bounded'_def)
   qed
 
 lemma cmap_relation_disjoint:
@@ -277,7 +275,7 @@ lemma cmap_relation_disjoint:
   and     ht: "s' \<Turnstile>\<^sub>c x"
   and     tp: "\<And>ko v. proj ko = Some v \<Longrightarrow> koType TYPE('a) = koTypeOf ko"
   and     xv: "x \<notin> Ptr ` {ptr..+2 ^ bits}"
-  and    sof: "size_td (typ_info_t TYPE('b)) \<le> 2 ^ objBits (undefined :: 'a)"
+  and    sof: "\<And>ko v. proj ko = Some v \<Longrightarrow> size_td (typ_info_t TYPE('b)) \<le> 2 ^ objBitsKO ko"
   shows  "{ptr_val x..+size_of TYPE('b)} \<inter> {ptr..+2 ^ bits} = {}"
 proof -
   from vuc have al: "is_aligned ptr bits"
@@ -286,8 +284,7 @@ proof -
     and [simp]: "(ptr && ~~ mask bits) = ptr"
     by (auto elim!: valid_untyped_capE)
 
-  note blah[simp del] =  atLeastAtMost_iff atLeastatMost_subset_iff atLeastLessThan_iff
-          Int_atLeastAtMost atLeastatMost_empty_iff split_paired_Ex
+  note [simp del] = atLeastAtMost_simps
 
   let ?ran = "{ptr..ptr + 2 ^ bits - 1}"
   let ?s = "(s\<lparr>ksPSpace := (ksPSpace s) |` (- ?ran)\<rparr>)"
@@ -305,7 +302,7 @@ proof -
   have oran' [simp]: "?oran' = ?oran"
   proof (rule upto_intvl_eq)
     from invs have "pspace_aligned' s" ..
-    with ks show "is_aligned (ptr_val x) (objBitsKO ko)"  ..
+    with ks show "is_aligned (ptr_val x) (objBitsKO ko)"  by (fastforce intro: pspace_alignedD')
   qed
 
   from xv have "ptr_val x \<in> (- ?ran)" apply (simp only: ran' Compl_iff)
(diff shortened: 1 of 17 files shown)


proof/refine/ARM/KHeap_R.thy
```diff
@@ -5089,6 +5089,19 @@ lemma is_active_sc'2_cross:
...
+  apply (frule state_relation_release_queue_relation)
+  apply (clarsimp simp: valid_release_q_def obj_at'_def release_queue_relation_def)
+  apply (drule_tac x=p in bspec, blast)
+  apply (clarsimp simp: vs_all_heap_simps)
+  apply (frule_tac t=p in st_tcb_at_coerce_concrete[rotated, where P=runnable], simp, simp)
+   apply (clarsimp simp: pred_tcb_at_def obj_at_def)
+  apply (clarsimp simp: pred_tcb_at'_def obj_at'_def sts_rel_runnable)
+  done
+
+
```
michaelmcinerney, line 5104: Remove extra line

spec/abstract/Structures_A.thy
```diff
@@ -551,10 +551,10 @@ lemma MIN_BUDGET_le_MAX_PERIOD':
 
 definition "min_sched_context_bits = 8"
 
-(* RT : size of sched_context struct in C, excluding refills
-   numbers are from MCS C: (9 * sizeof(word_t)) + (3 * sizeof(ticks_t)) for aarch32 *)
+(* RT : size of sched_context struct in C. Currently hard coded from sched_context_C_size_of.
+   See GitHub issue 501. *)
 definition sizeof_sched_context_t :: nat where
-  "sizeof_sched_context_t = (9 * 4) + (3 * 8)"
+  "sizeof_sched_context_t = 96"
```
lsf37, line 558: Isn't the size architecture dependent? At least I'd imagine it would be different for 32 bit and 64 bit machines. This seems to work out, i.e. `size_of(word_t)` = 8 in RSCV64, which gives you exactly the 96 instead of the 60 currently defined here (and `size_of(ticks_t)` is stable over 32/64 for now).

So I think we should keep the old comment -- the comment is correct, just not correctly implemented. It's still a hard-coded number, so linking to the github issue is also good. For the definition, we could say:

```
sizeof_sched_context_t = (9 * word_size) + (3 * 8)"
```

This should evaluate to the correct value on all architectures we currently support.

spec/haskell/src/SEL4/Object/Structures.lhs
```diff
@@ -216,9 +216,9 @@ list of pointers to waiting threads;
 >     scRefillCount :: Int,
 >     scRefills :: [Refill]}
 
-> -- numbers from MCS C: (9 * sizeof(word_t)) + (3 * sizeof(ticks_t)) for aarch32
+> -- currently copied directly from sizeof_sched_context_t
 > schedContextStructSize :: Int
-> schedContextStructSize = (9 * 4) + (3 * 8)
+> schedContextStructSize = 96
```
lsf37, line 221: Should use the same construction as proposed for Strucutures_A

proof/crefine/RISCV64/Detype_C.thy
```diff
@@ -324,25 +322,14 @@ proof -
...
-      done
-    thus "{ptr_val x..+size_of TYPE('b)} \<subseteq> ?oran'" using sof
-      apply -
-      apply (rule intvl_start_le)
-      apply (simp add: size_of_def)
-      done
+  apply (rule disjoint_subset[rotated])
+  apply (rule intvl_start_le)
+  apply (clarsimp simp: size_of_def)
+  by (fastforce intro: sof po tp)
```
lsf37, line 328: Indent looks off by 2 (it should be indented by 2 from `thus`)

proof/crefine/RISCV64/Detype_C.thy
```diff
@@ -1025,10 +1012,11 @@ proof -
...
 
 lemma cnotification_relation_restrict:
-  assumes vuc: "s \<turnstile>' capability.UntypedCap d ptr bits idx"
-  and    invs: "invs' s"
-  and      rl: "\<forall>(p :: machine_word) P. ko_wp_at' P p s \<and> (\<forall>ko. P ko \<longrightarrow> live' ko) \<longrightarrow> p \<notin> {ptr..ptr + 2 ^ bits - 1}"
-  and    meps: "map_to_ntfns (ksPSpace s) p = Some ntfn"
+  assumes  vuc: "s \<turnstile>' capability.UntypedCap d ptr bits idx"
+  and     invs: "invs' s"
+  and sym_refs: "sym_refs (state_refs_of' s)"
+  and       rl: "\<forall>(p :: machine_word) P. ko_wp_at' P p s \<and> (\<forall>ko. P ko \<longrightarrow> live' ko) \<longrightarrow> p \<notin> {ptr..ptr + 2 ^ bits - 1}"
```
lsf37, line 1018: I know you've just swapped the line, but if we're touching it anyway, we could remove the unnecessary parentheses and type annotation for `p`.

lsf37 (approved):
Cool. Happy with it, but we should adjust the `sched_context_size` definition.

spec/abstract/Structures_A.thy
michaelmcinerney: OK great. The proofs will still go through for RISCV, but I think I'll have to revert the changes in Refine for ARM. I'll check that out

proof/crefine/RISCV64/Detype_C.thy
```diff
@@ -277,7 +275,8 @@ lemma cmap_relation_disjoint:
   and     ht: "s' \<Turnstile>\<^sub>c x"
   and     tp: "\<And>ko v. proj ko = Some v \<Longrightarrow> koType TYPE('a) = koTypeOf ko"
   and     xv: "x \<notin> Ptr ` {ptr..+2 ^ bits}"
-  and    sof: "size_td (typ_info_t TYPE('b)) \<le> 2 ^ objBits (undefined :: 'a)"
+  and    sof: "\<And>ko v. proj ko = Some v \<and> koType TYPE('a) = koTypeOf ko \<Longrightarrow>
+                      size_td (typ_info_t TYPE('b)) \<le> 2 ^ objBitsKO ko"
```
corlewis, line 279: Very minor, but considering the `tp` assumption above is the second conjunct here needed?

proof/crefine/RISCV64/Detype_C.thy
```diff
@@ -400,8 +387,8 @@ proof -
     case subset2 thus ?thesis .
   next
     case subset1
-    note blah[simp del] =  atLeastAtMost_iff atLeastatMost_subset_iff atLeastLessThan_iff
-          Int_atLeastAtMost atLeastatMost_empty_iff split_paired_Ex
+    note [simp del] = atLeastAtMost_simps
+
```
corlewis, line 391: This whitespace is new, might not be intentional.

proof/crefine/RISCV64/Detype_C.thy
```diff
@@ -1668,16 +1650,17 @@ proof -
...
     apply ((subst lift_t_typ_region_bytes,
-            rule cm_disj cm_disj_tcb cm_disj_cte cm_disj_user cm_disj_device
-            , assumption +,
-            simp_all add: objBits_simps' bit_simps
-                          heap_to_user_data_restrict heap_to_device_data_restrict)[1])+ \<comment> \<open>waiting ...\<close>
+           rule cm_disj cm_disj_tcb cm_disj_cte cm_disj_user cm_disj_device, assumption+,
+           simp_all add: objBits_simps' bit_simps scBits_simps
+                         heap_to_user_data_restrict heap_to_device_data_restrict)[1],
+           (rule_tac y="2^7" in order_trans, simp,
+            fastforce intro!: power_increasing_iff[THEN iffD2] scBits_at_least_7)?)+ \<comment> \<open>waiting ...\<close>
```
corlewis, line 1663: Wow, it can't have been fun getting this working!
It would be very nice to split this up into smaller, more understandable steps. However, considering that it was pretty bad already I wouldn't suggest putting too much effort into it.
However, could we make the comment a bit more specific? I'm assuming it means that this is a long running step, but I always find it more useful if comments like that provide a rough idea of how long the command is expected to take.

corlewis (approved):

proof/crefine/RISCV64/Detype_C.thy
michaelmcinerney: I've seen some comments that say "takes long" or "60 seconds" take very little time, so I think some of the older comments are a bit out of whack. Sometimes I just remove them, but if they're vague, it's hard to know what to change them to if it might improve soon. But here I think it's only about 10 seconds, so I could put "10 seconds or so"

proof/crefine/RISCV64/Detype_C.thy
michaelmcinerney: I originally wrote this as `!!v :: 'a. size_td (type_info_t TYPE('b)) \<le> objBits v`, but had a hard time wrestling with `objBits` and `objBitsT` and `objBitsKO`, or whatever the variations are. I will see if this second conjunct can be removed but I'm not overly hopeful that it'll be easy or nice

PR #507, opened 2022-07-26
C heap tactic

simonjwinwood:
This tactic deals with heap updates in the context of lift_t, turning UMM heap_updates into function updates to the lift_t, or removing them completely if the update doens't interact with the lift_t'd type.


lsf37:
Closing, because it is unfinished and unlikely to be finished soon. I don't think there is anything wrong with the idea, it just needs more work. If anyone wants to pick it up, please feel free to re-open.

PR #511, opened 2022-07-29
Style for `ccorres` statements

michaelmcinerney:


corlewis (commented):
The actual content looks good to me. However, I can't be sure without opening it up myself, but do you need all of the fake constants? I would have thought you would just need `ccorres`, `UNIV`, `c_guard`, `ptr_select`, and maybe `ptr` and `cond`. Couldn't the `ccorres` arguments be bound variables?

Also, you shouldn't need the axiomatizations unless you explicitly refer to the definition lemma.

michaelmcinerney:
> The actual content looks good to me. However, I can't be sure without opening it up myself, but do you need all of the fake constants. I would have thought you would just need `ccorres`, `UNIV`, `c_guard`, `ptr_select`, and maybe `ptr` and `cond`. Couldn't the `ccorres` arguments be bound variables?
> 
> Also, you shouldn't need the axiomatizations unless you explicitly refer to the definition lemma.

Very good point. None of the constants is needed. I've removed them, and adjusted things in the fake lemmas to make the schematic variable names nicer

docs/Style.thy
```diff
@@ -323,6 +323,59 @@ lemma my_hoare_triple_lemma:
...
+  oops
+
+lemma long_ccorres_example:
+  "ccorres rrel xf
+     long_abs_guard long_conc_guard hs
+     long_abs_fn long_conc_fn"
+  oops
+
+lemma longer_ccorres_example:
+  "ccorres rrel xf
```
Xaphiosis, line 366: something to consider: what about complex return relations? I think they should nearly all fit, I haven't found anything more complex than:

`ccorres  (cfault_rel2 ¤ dc) (liftxf errstate id (K ()) ret__unsigned_long_')`

docs/Style.thy
```diff
@@ -323,6 +323,59 @@ lemma my_hoare_triple_lemma:
...
+     long_conc_guard hs
+     long_abs_fn
+     long_conc_fn"
+  oops
+
+text \<open>
+  The concrete guard will often be simply @{term UNIV}, or an intersection of terms of the form
+  @{term "\<lbrace>\<acute>pointer = cond\<rbrace>"}, which supersedes the set-builder notation. Note that the
+  semantically redundant form @{term "UNIV \<inter> \<lbrace>\<acute>pointer = cond\<rbrace>"} should no longer be necessary,
+  but may still be required in some places.\<close>
```
Xaphiosis, line 377: I'd tweak this a little bit. We want to indicate that we don't want people using it unless they are forced to by legacy tactics/methods or related issues.

docs/Style.thy
```diff
@@ -323,6 +323,59 @@ lemma my_hoare_triple_lemma:
...
+  Often the guards will require more than one line, and in this case, hs should occur on the same
+  line as the concrete guard wherever possible.
+
+  If the statement must use more than one line, all lines after the first should be indented by
+  2 spaces from @{term ccorres}.
+
+  Here are some examples.\<close>
+
+lemma short_ccorres_example:
+  "ccorres rrel xf short_abs_guard short_conc_guard hs short_abs_fn short_conc_fn"
```
Xaphiosis, line 356: This is off-topic slightly, but `conc` always makes me think of "concurrent". `concr` might be less ambiguous, but up to you.

docs/Style.thy
michaelmcinerney: Sure, I'll put a `longer_rrefl` and `longer_xf` on separate lines in this example

Xaphiosis (approved):

lsf37 (approved):
Looks good from my side!

Xaphiosis:
@michaelmcinerney : any reason this wasn't merged?

michaelmcinerney:
> @michaelmcinerney : any reason this wasn't merged?

No good reason. I've been meaning to ask whether it's fine to merge. I'll try today

PR #512, opened 2022-08-02
Sorry until the end of `Interrupt_C.thy`

michaelmcinerney:


proof/crefine/RISCV64/Invoke_C.thy
```diff
@@ -1254,9 +1194,11 @@ lemma decodeCNodeInvocation_ccorres:
                                                          performInvocation_def bindE_assoc)
                                         apply (ctac add: setThreadState_ccorres)
                                           apply (rule ccorres_rhs_assoc2, rule ccorres_split_throws)
+                                           apply (rule ccorres_stateAssertE_fwd)
                                            apply (ctac(no_vcg) add: invokeCNodeRotate_ccorres)
                                              apply (rule ccorres_alternative2)
                                              apply (rule ccorres_return_CE, simp+)[1]
+
```
michaelmcinerney, line 1201: Remove

proof/crefine/RISCV64/Invoke_C.thy
```diff
@@ -1499,26 +1443,29 @@ lemma updateCap_ct_active'[wp]:
...
-     newType = APIObjectType apiobject_type.Untyped \<longrightarrow> us \<le> 61\<rbrakk>
+     newType = APIObjectType apiobject_type.Untyped \<longrightarrow> us \<le> 61;
+     newType = APIObjectType apiobject_type.SchedContextObject \<longrightarrow> sc_size_bounds us\<rbrakk>
            \<Longrightarrow> APIType_capBits newType us < 64"
   apply (case_tac newType)
-  apply (clarsimp simp: invokeUntyped_proofs_def APIType_capBits_def objBits_simps' bit_simps
+  apply (auto simp: invokeUntyped_proofs_def APIType_capBits_def objBits_simps' bit_simps'
+                    scBits_simps untypedBits_defs
                  split: apiobject_type.splits)+
+
```
michaelmcinerney, line 1464: Remove

proof/crefine/RISCV64/Invoke_C.thy
```diff
@@ -2412,8 +2363,8 @@ lemma invokeUntyped_Retype_ccorres:
           apply (drule canonical_address_neq_mask[where sz=sz])
            apply (rule sz_bound[folded maxUntypedSizeBits_def])
           apply clarsimp
-          apply (wp updateFreeIndex_forward_invs' sch_act_simple_lift
-                    updateFreeIndex_cte_wp_at hoare_vcg_const_Ball_lift
+          apply_trace (wp updateFreeIndex_forward_invs' sch_act_simple_lift
```
michaelmcinerney, line 2366: Remove `apply_trace`

michaelmcinerney:
Missed a few bits of cleanup, but should be all good now

proof/crefine/RISCV64/Invoke_C.thy
```diff
@@ -75,24 +75,43 @@ lemma setDomain_ccorres:
...
 
 lemma active_runnable':
   "active' state \<Longrightarrow> runnable' state"
-  by (fastforce simp: runnable'_def)
+  by fastforce
+
+lemma tcbSchedEnqueue_ksReadyQueues_distinct:
+  "\<lbrace>\<lambda>s. \<forall>d p. distinct (ksReadyQueues s (d, p)) \<and> thread \<notin> set (ksReadyQueues s (d,p))\<rbrace>
+   tcbSchedEnqueue thread
+   \<lbrace>\<lambda>_ s. \<forall>d p. distinct (ksReadyQueues s (d, p))\<rbrace>"
```
corlewis, line 103: It's rare, but `ksReadyQueues_distinct` lemmas are sometimes useful without the forall. If it's easy, then I would prefer the following, with a `hoare_vcg_all_lift` where it's called.
```suggestion
  "\<lbrace>\<lambda>s. distinct (ksReadyQueues s (d, p)) \<and> thread \<notin> set (ksReadyQueues s (d,p))\<rbrace>
   tcbSchedEnqueue thread
   \<lbrace>\<lambda>_ s. distinct (ksReadyQueues s (d, p))\<rbrace>"
```

proof/crefine/RISCV64/Invoke_C.thy
```diff
@@ -260,7 +281,8 @@ done
 
 lemma invokeCNodeCancelBadgedSends_ccorres:
   "ccorres (cintr \<currency> dc) (liftxf errstate id (K ()) ret__unsigned_long_')
-     (invs' and sch_act_simple and valid_cap' cap and K (isEndpointCap cap)) (UNIV \<inter> {s. ccap_relation cap (cap_' s)}) []
+     (invs' and sch_act_simple and valid_cap' cap and K (isEndpointCap cap))
+     (UNIV \<inter> {s. ccap_relation cap (cap_' s)}) []
```
corlewis, line 285: Do we want to remove the `UNIV` from lemmas that are already being modified?

proof/crefine/RISCV64/Invoke_C.thy
```diff
@@ -2413,7 +2362,7 @@ lemma invokeUntyped_Retype_ccorres:
            apply (rule sz_bound[folded maxUntypedSizeBits_def])
           apply clarsimp
           apply (wp updateFreeIndex_forward_invs' sch_act_simple_lift
-                    updateFreeIndex_cte_wp_at hoare_vcg_const_Ball_lift
+                    Untyped_R.updateFreeIndex_cte_wp_at hoare_vcg_const_Ball_lift
```
corlewis, line 2365: This qualification being needed made me check and it looks like we should be able to remove the lemma in `Untyped_R` and use the `Finalise_R` version. It's definitely not worth cleaning up now, but we might want to leave a FIXME.

corlewis (approved):

proof/crefine/RISCV64/Invoke_C.thy
lsf37: Yes, I'd also agree that this form is more useful. The post condition binds `d` and `p` so we're not introducing schematics in the precondition when we're leaving out the quantifier here, and it makes the lemma more often applicable.

proof/crefine/RISCV64/Invoke_C.thy
lsf37: I'd vote for removing it if we touch the lemma.

lsf37 (approved):
👍

proof/crefine/RISCV64/Invoke_C.thy
michaelmcinerney: Done. And yes, just one use of `hoare_vcg_all_lift` needed

proof/crefine/RISCV64/Invoke_C.thy
michaelmcinerney: The qualification was actually easy to remove after all, so I have just done that

PR #513, opened 2022-08-08
Sorry until the end of `Ipc_C.thy`

michaelmcinerney:


proof/crefine/RISCV64/Ipc_C.thy
```diff
@@ -2009,15 +2145,15 @@ lemma doFaultTransfer_ccorres [corres]:
     (UNIV \<inter> \<lbrace>\<acute>sender = tcb_ptr_to_ctcb_ptr sender\<rbrace>
           \<inter> \<lbrace>\<acute>receiver = tcb_ptr_to_ctcb_ptr receiver\<rbrace>
           \<inter> \<lbrace>\<acute>receiverIPCBuffer = Ptr (option_to_0 buffer)\<rbrace>
-          \<inter> \<lbrace>\<acute>badge = badge\<rbrace>) []
+          \<inter> \<lbrace>\<acute>badge___unsigned_long = badge\<rbrace>) []
```
corlewis, line 2148: Didn't we make it so that we didn't need to make these sorts of changes? Has something gone wrong here to cause this?

corlewis (approved):
Great job fixing up the Timeout fault!

We might want to come back to Ipc_C in the future if/when we make SchedContext_C and Reply_C, to move some of the new sorried lemmas introduced.

proof/crefine/RISCV64/Ipc_C.thy
lsf37: Resolved on MM, to summarise: the name `badge` is used with two compatible but different types in the code in MCS, so we do need the long name to distinguish. We discussed if we can change one of these types, but one comes from the bitfield generator, and the other use makes sense as it is, so we should probably leave it.

lsf37 (approved):
Cool. Looks like this one will be one with a few fairly substantial new proofs.

PR #516, opened 2022-08-27
lib: add named_eta and no_name_eta methods

Xaphiosis:
These allow selective eta-contraction in the goal based on the bound
variable's name. The `no_name_eta` method speficially targets
abstractions where the variable has no name, which can come up in
complicated unification scenarios.

These nameless abstractions can cause symbolic execution lemmas to no
longer pick up on the name of the bound variable in do-notation,
requiring multiple rename_tac invocations.

Co-authored-by: Gerwin Klein <gerwin.klein@proofcraft.systems>

### DIFF
--- a/lib/NICTATools.thy
+++ b/lib/NICTATools.thy
@@ -21,6 +21,7 @@ imports
   ShowTypes
   Locale_Abbrev
   Value_Type
+  Named_Eta
 begin
 
 section "Detect unused meta-forall"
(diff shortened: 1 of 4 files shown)


Xaphiosis:
@corlewis @michaelmcinerney Adding you two both to experience the magic and comment on whether the docs/comments etc make sense.

lsf37 (approved):
Looks good. Thanks for improving the docs!

corlewis (approved):
I didn't dig into the ML at all, but it looks good and the documentation and tests are really nice!

PR #517, opened 2022-09-09
lib: add rules_tac and related multi-thm instantiators

Xaphiosis:
In `Rules_Tac`, add a `rules_tac` which is `rule_tac` but with the ability to instantiate the same variable name in multiple theorems.

Also add the specialised `single_instantiate_tac` which allows using the above mechanism to instantiate a specific variable name in a specific set of theorems (e.g. "rv" in a set of symbolic-execution lemmas).

Note1: Rules_Tac.thy and Rules_Tac_Tests.thy are a bit small. We should consider whether we want every little method and tactic to show up with its own file in lib, or whether grouping them by some kind of topic or size (e.g. Small_Methods.thy) is more appropriate

Note2: this PR is used by the monadic_rewrite overhaul and will need to be merged before it.

### DIFF
--- a/lib/ROOT
+++ b/lib/ROOT
@@ -84,6 +84,7 @@ session Lib (lib) = Word_Lib +
     ML_Goal_Test
     Value_Type
     Named_Eta
+    Rules_Tac
 
     (* should really be a separate session, but too entangled atm: *)
     NonDetMonadLemmaBucket
@@ -163,6 +164,7 @@ session LibTest (lib) in test = Refine +
     Locale_Abbrev_Test
     Value_Type_Test
     Named_Eta_Test
+    Rules_Tac_Test
   (* use virtual memory function as an example, only makes sense on ARM: *)
   theories [condition = "L4V_ARCH_IS_ARM"]
     Corres_Test
--- a/lib/Rules_Tac.thy
+++ b/lib/Rules_Tac.thy
@@ -1,0 +1,55 @@
+(*
+ * Copyright 2022, Proofcraft Pty Ltd
+ *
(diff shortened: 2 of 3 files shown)


lsf37 (approved):
Looks good!

> Note1: Rules_Tac.thy and Rules_Tac_Tests.thy are a bit small. We should consider whether we want every little method and tactic to show up with its own file in lib, or whether grouping them by some kind of topic or size (e.g. Small_Methods.thy) is more appropriate

I think it's substantial enough for its own theory file. Also produces easier concurrency in the build.

It took me a while to understand what the point of `single_instantiate_tac` is -- I didn't understand it from the Rules_Tac file but only later from the examples later. I don't really have any good suggestions for improvements to the comment, though.

corlewis:
> > Note1: Rules_Tac.thy and Rules_Tac_Tests.thy are a bit small. We should consider whether we want every little method and tactic to show up with its own file in lib, or whether grouping them by some kind of topic or size (e.g. Small_Methods.thy) is more appropriate
> 
> I think it's substantial enough for its own theory file. Also produces easier concurrency in the build.

How about a `Small_Methods` directory? It would let us have these methods and tactics separate and keep the concurrency, while also making them easier to find and helping stop `lib` from becoming even larger.

lsf37:
> How about a `Small_Methods` directory? It would let us have these methods and tactics separate and keep the concurrency, while also making them easier to find and helping stop `lib` from becoming even larger.

We could think about that as a separate step, yes. The lib top-level directory is getting a bit large and it'd be nice to have a bit more content grouping.

PR #518, opened 2022-09-09
lib: add if_then_else, has_concl, case_concl Eisbach_Methods

Xaphiosis:
These allow for pattern-matching on the conclusion and reacting to whether the match succeeded.


lsf37 (approved):
👍 Nice addition.

(sign-off line is missing in the commit message)

PR #520, opened 2022-09-09
lib: add rev_all_new + then_all_new methods, REV_ALL_NEW combinator

Xaphiosis:
rev_all_new and REV_ALL_NEW act like `;` and THEN_ALL_NEW, but with the second method is applied to the results of the first in the order they were produced, making it safe for WP reasoning.

Note: I can remove then_all_new, it's not used. It's kinda useful to play around with when developing tactics, but it's also easy to recreate via copy-pasta.

### DIFF
--- a/lib/Eisbach_Methods.thy
+++ b/lib/Eisbach_Methods.thy
@@ -1,4 +1,5 @@
 (*
+ * Copyright 2022, Proofcraft Pty Ltd
  * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
  *
  * SPDX-License-Identifier: BSD-2-Clause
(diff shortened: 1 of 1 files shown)


Xaphiosis:
@corlewis pointed out that this has issues when tac2 creates more goals than it solves, and upon thinking about it, I think it'll also have issues when tac2 solves more than 1 goal. This needs a redo and re-think. It should still be doable, but needs more careful arithmetic taking into account the difference in number of subgoals between start and end of tac2.

Xaphiosis:
There are some fascinating issues here. Doing the following test indicates that the behaviour of both `THEN_ALL_NEW` and `REV_ALL_NEW` exhibits the same issue and even yield identical results (!), but ";" exhibits the expected behaviour of successfully applying m2 to every goal produced by m1 in `(m1;m2)`.

```
declare [[goals_limit=25]]
lemma
  assumes a: "⋀x. P x"
  assumes b: "⋀x. Q x"
  assumes goal_multiplicity2: "⋀n. ⟦ P (n+10);P (n+20) ⟧ ⟹ P n"
  assumes goal_multiplicity5: "⋀n. ⟦ P (n+1);P (n+2);P (n+3);P (n+4);P (n+5) ⟧ ⟹ P n"
  assumes goal_multiplicity2x: "⋀n. ⟦ P n;Q n ⟧ ⟹ P n"
  shows "P 0"
proof -

  (* generating using semicolon *)
  have "P 0"
  apply (rule goal_multiplicity2; rule goal_multiplicity5
         (* invoked only on P goals, results in 20 subgoals *)
         ; (print_concl, rule goal_multiplicity2x))
    by (repeat 20 ‹rule a b›)

  (* generating using then_all_new *)
  have "P 0"
    apply (
      then_all_new ‹rule goal_multiplicity2›
        ‹then_all_new ‹rule goal_multiplicity5›
        (* invoked only on P goals, results in 20 subgoals, but order is different to ";"! *)
        ‹print_concl, rule goal_multiplicity2x››)
    (* The generated goals are not right: goal_multiplicity5 and goal_multiplicity2x are never
       applied to the second output of goal_multiplicity2, so P (0+20) remains at the end. *)
    by (repeat 20 ‹rule a b›)

  (* generating using rev_all_new *)
  have "P 0"
    apply (
      rev_all_new ‹rule goal_multiplicity2›
        ‹rev_all_new ‹rule goal_multiplicity5›
        (* invoked only on P goals, results in 20 subgoals, but order is different to ";"! *)
        ‹print_concl, rule goal_multiplicity2x››) (* THIS IS COMPLETELY WRONG *)
    (* The generated goals are not right: goal_multiplicity5 and goal_multiplicity2x are never
       applied to the second output of goal_multiplicity2, so P (0+20) remains at the end. *)
    by (repeat 20 ‹rule a b›)

  oops
```

It's worth nothing that `;` does indeed to `THEN_ALL_NEW`, but *not the one in tactical.ML*! It uses its own version (see method.ML) instead. I have to check whether there is an issue in how I'm using that, or whether `THEN_ALL_NEW` behaves completely differently to `THEN_ALL_NEW`. Argh.

Xaphiosis:
Can confirm this is some kind of tactics vs methods vs me problem, the following version does work as expected, but of course I can't access the internals in `method.ML` it uses:

```
method_setup then_all_new2 = ‹
  Method.text_closure -- Method.text_closure >> (fn (m1,m2) => fn ctxt => fn facts =>
let
  val m = (Method.Combinator (Method.no_combinator_info, Method.Then_All_New, [m1,m2]))
  val tac = method_evaluate m ctxt facts
in
  SIMPLE_METHOD tac facts
end)›
```

Xaphiosis:
This one had issues. Re-implementation is done, testing is in progress, will open another PR when it's ready.

PR #521, opened 2022-09-15
lib: add fwd_all + fwd_all_new methods, combinators

Xaphiosis:
fwd_all and ALLGOALS_FWD act like `all`, but supplied method is applied to goals in first-to-last order, taking into account goals solved and generated.

fwd_all_new and FWD_ALL_NEW act like `;` and THEN_ALL_NEW, but with the second method is applied to the results of the first in the order they were produced, making it safe for WP reasoning.

Note: had to drop the WP rule tests as they don't work in Eisbach_Methods

### DIFF
--- a/lib/Eisbach_Methods.thy
+++ b/lib/Eisbach_Methods.thy
@@ -1,4 +1,5 @@
 (*
+ * Copyright 2022, Proofcraft Pty Ltd
  * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
  *
  * SPDX-License-Identifier: BSD-2-Clause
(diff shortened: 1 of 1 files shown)


lib/Eisbach_Methods.thy
```diff
@@ -574,4 +575,156 @@ method if_concl for pattern::bool methods m =
...
+
+in
+
+fun ALLGOALS_FWD tac = tac_in_order tac 1;
+
+end
+
+infix 1 FWD_ALL_NEW;
+
+(* this is like THEN_ALL_NEW but expressed with goal restriction and ALLGOALS_REV *)
```
corlewis, line 615: Should this comment say `ALLGOALS_FWD`?

lib/Eisbach_Methods.thy
```diff
@@ -574,4 +575,156 @@ method if_concl for pattern::bool methods m =
...
+\<close>
+
+method_setup fwd_all =
+  \<open>Method.text_closure >> (fn m => fn ctxt => fn facts =>
+   let
+     fun tac i st' =
+       Goal.restrict i 1 st'
+       |> method_evaluate m ctxt facts
+       |> Seq.map (Goal.unrestrict i)
+
```
corlewis, line 632: Tiny nitpick, but I think I would prefer this without the empty line.

lib/Eisbach_Methods.thy
```diff
@@ -574,4 +575,156 @@ method if_concl for pattern::bool methods m =
...
+lemma "True"
+  apply (succeeds \<open>fwd_all_new succeed succeed\<close>)
+  apply (fails \<open>fwd_all_new succeed fail\<close>)
+  apply (fails \<open>fwd_all_new fail succeed\<close>)
+  apply (fails \<open>fwd_all_new fail fail\<close>)
+  apply (rule TrueI)
+  done
+
+end (* fwd_all_new and fwd_all experiment *)
```
corlewis, line 728: nitpick: it's currently only really a `fwd_all_new` experiment

corlewis (approved):
I haven't played around with it, but this all looks good to me and there's been plenty of times in the past that I've wanted something like it!

Unfortunately, I tried to think of an alternative syntax similar to `;` but nothing came to mind for me that I liked. If we could find something then I think that would make it a lot easier to read and write.

lib/Eisbach_Methods.thy
```diff
@@ -574,4 +575,156 @@ method if_concl for pattern::bool methods m =
...
+      goals usually need to be solved in the order they were generated. The @{text fwd_all}
+      and @{text fwd_all_new} methods apply on goals in order, taking into account goals solved
+      and added.\<close>
+
+ML \<open>
+
+fun restrict_goal i n tac =
+  (PRIMITIVE (Goal.restrict i n)) THEN
+  tac THEN
+  (PRIMITIVE (Goal.unrestrict i));
```
lsf37, line 588: We should probably also put this in a `local` block or at least a struct, otherwise it's in the global name space and might get confused with the same function in the Isabelle sources.

lib/Eisbach_Methods.thy
```diff
@@ -574,4 +575,156 @@ method if_concl for pattern::bool methods m =
...
+
+  (* check also that we can solve everything it spits out *)
+  have "P 0"
+    by (
+      fwd_all_new
+        \<open>fwd_all_new \<open>rule goal_multiplicity2\<close>
+          \<open>print_concl, fwd_all_new \<open>rule goal_multiplicity5\<close>
+          \<open>rule goal_multiplicity2x\<close>\<close>\<close>
+        \<open>rule a b\<close>)
+  oops
```
lsf37, line 702: Since we're in experiment, we could prove the lemma with `show "P 0" by (rule a)` to avoid the explicit oops.

lib/Eisbach_Methods.thy
lsf37: True, we should add test cases for `fwd_all` as well.

lsf37 (approved):
Cool, also very happy with that. (Minor tweaks in the comments)

lib/Eisbach_Methods.thy
Xaphiosis: I think we'll need a struct, because we use it in the `method_setup`

lib/Eisbach_Methods.thy
Xaphiosis: thank you!

lib/Eisbach_Methods.thy
lsf37: Good point, yes, let's use a struct then.

lib/Eisbach_Methods.thy
Xaphiosis: I did the oops on purpose, thinking there is no benefit to exporting more lemmas into the context. Still want me to change it?

lib/Eisbach_Methods.thy
Xaphiosis: I'm not sure what kind of struct to use honestly. If the only thing that does in it is restrict_goal? Should I put the other combinators there? Any bikeshed help on this one?

lib/Eisbach_Methods.thy
Xaphiosis: Done

lib/Eisbach_Methods.thy
Xaphiosis: Added one experiment, not sure what else should go in there.

lib/Eisbach_Methods.thy
Xaphiosis: I used `local` and folded the def again into the method. It's short enough.

lib/Eisbach_Methods.thy
lsf37: When you close `experiment` it won't be exported into the context. (I think that's the main reason Makarius added it, but maybe there is more).

lsf37: Result of investigation for `experiment`: it likely will be in the context, but inaccessible. A version that might not produce anything could be `notepad`. In any case, no big deal, it's also fine to just leave the `oops`.

Xaphiosis: I changed it to be one big `notepad`, that way nothing gets exported, but I did have to add some `fix` to get it to work and not spit warnings.

PR #523, opened 2022-09-28
Add `hoare_from_abs` rule

michaelmcinerney:


michaelmcinerney:
I didn't put much thought into the names of the new lemmas. Maybe `valid_cross` or `hoare_cross` might be slightly better than `hoare_from_abs`. But I imagine that someone would have to search for a theorem with `valid` and `corres_underlying` to find it, anyway. These lemmas are visible in AInvs, Refine, and CRefine, at least

lib/ExtraCorres.thy
```diff
@@ -203,4 +204,40 @@ lemma wp_from_corres_unit:
...
+  "(s, s') \<in> sr \<Longrightarrow> P s \<Longrightarrow> ex_abs_underlying sr P s'"
+  by (auto simp add: ex_abs_underlying_def)
+
+lemma corres_underlying_split_ex_abs:
+  assumes ac: "corres_underlying srel nf nf' r' G G' a c"
+  assumes valid: "\<lbrace>G\<rbrace> a \<lbrace>P\<rbrace>" "\<lbrace>G' and ex_abs_underlying srel G\<rbrace> c \<lbrace>P'\<rbrace>"
+  assumes bd: "\<forall>rv rv'. r' rv rv' \<longrightarrow>
+                        corres_underlying srel nf nf' r (P rv) (P' rv') (b rv) (d rv')"
+  shows "corres_underlying srel nf nf' r G G' (a >>= (\<lambda>rv. b rv)) (c >>= (\<lambda>rv'. d rv'))"
+  using ac bd valid
```
Xaphiosis, line 220: at this point you can `use assms` since you're using all of them

lib/ExtraCorres.thy
```diff
@@ -203,4 +204,40 @@ lemma wp_from_corres_unit:
...
+  apply (clarsimp simp: Bex_def Ball_def valid_def ex_abs_underlying_def)
+  by meson
+
+lemma hoare_from_abs:
+  assumes abs_valid: "\<lbrace>P and R\<rbrace> f \<lbrace>\<lambda>rv. Q rv and S\<rbrace>"
+  assumes cross: "\<And>s s'. \<lbrakk>(s, s') \<in> srel; P' s'; R' s'\<rbrakk> \<Longrightarrow> P s"
+                 "\<And>s s' r r'. \<lbrakk>(s, s') \<in> srel; rrel r r'; Q r s; S s\<rbrakk> \<Longrightarrow> Q' r' s'"
+  assumes corres: "corres_underlying srel False True rrel G G' f f'"
+  shows "\<lbrace>P' and G' and R' and ex_abs_underlying srel (G and R)\<rbrace> f' \<lbrace>Q'\<rbrace>"
+  apply (insert assms)
```
Xaphiosis, line 231: consider `using assms`, it should have the same effect in front of a fastforce

lib/ExtraCorres.thy
```diff
@@ -203,4 +204,40 @@ lemma wp_from_corres_unit:
...
+  by (fastforce simp: valid_def ex_abs_underlying_def corres_underlying_def)
+
+lemma hoare_from_abs_inv:
+  assumes abs_valid: "f \<lbrace>P\<rbrace>"
+  assumes cross: "\<And>s s'. (s, s') \<in> srel \<Longrightarrow> P s = P' s'"
+  assumes corres: "corres_underlying srel False True rrel G G' f f'"
+  shows "\<lbrace>P' and G' and ex_abs_underlying srel G\<rbrace> f' \<lbrace>\<lambda>_. P'\<rbrace>"
+  by (insert assms)
```
Xaphiosis, line 239: this is a bit of a strange formulation, probably this would be more "canonical":
```
using assms
by (rule ... ,
    fastforce+)  (or ; fastforce if the reverse-order doesn't mess it up)
```

lib/ExtraCorres.thy
```diff
@@ -203,4 +204,40 @@ lemma wp_from_corres_unit:
...
+                 "\<And>s s' r r'. \<lbrakk>(s, s') \<in> srel; rrel r r'; Q r s; S s\<rbrakk> \<Longrightarrow> Q' r' s'"
+  assumes corres: "corres_underlying srel False True rrel G G' f f'"
+  shows "\<lbrace>P' and G' and R' and ex_abs_underlying srel (G and R)\<rbrace> f' \<lbrace>Q'\<rbrace>"
+  apply (insert assms)
+  by (fastforce simp: valid_def ex_abs_underlying_def corres_underlying_def)
+
+lemma hoare_from_abs_inv:
+  assumes abs_valid: "f \<lbrace>P\<rbrace>"
+  assumes cross: "\<And>s s'. (s, s') \<in> srel \<Longrightarrow> P s = P' s'"
+  assumes corres: "corres_underlying srel False True rrel G G' f f'"
```
Xaphiosis, line 237: Can you say anything about why you need `False True` here? Coming from monadic_rewrite I'm interested in the flags and strength of formulations.

Xaphiosis:
I'm fine with the names. Small comments, but nothing semantic.

Xaphiosis (approved):

lib/ExtraCorres.thy
```diff
@@ -203,4 +204,40 @@ lemma wp_from_corres_unit:
...
+lemma ex_absI[intro!]:
+  "(s, s') \<in> sr \<Longrightarrow> P s \<Longrightarrow> ex_abs_underlying sr P s'"
+  by (auto simp add: ex_abs_underlying_def)
+
+lemma corres_underlying_split_ex_abs:
+  assumes ac: "corres_underlying srel nf nf' r' G G' a c"
+  assumes valid: "\<lbrace>G\<rbrace> a \<lbrace>P\<rbrace>" "\<lbrace>G' and ex_abs_underlying srel G\<rbrace> c \<lbrace>P'\<rbrace>"
+  assumes bd: "\<forall>rv rv'. r' rv rv' \<longrightarrow>
+                        corres_underlying srel nf nf' r (P rv) (P' rv') (b rv) (d rv')"
+  shows "corres_underlying srel nf nf' r G G' (a >>= (\<lambda>rv. b rv)) (c >>= (\<lambda>rv'. d rv'))"
```
corlewis, line 219: It depends on the intention of how these rules are used, but for backwards reasoning purpose  a better order for these assumptions would be if `bd` was before `valid`.

lib/ExtraCorres.thy
```diff
@@ -203,4 +204,40 @@ lemma wp_from_corres_unit:
...
+  apply (clarsimp simp: corres_underlying_def bind_def)
+  apply (clarsimp simp: Bex_def Ball_def valid_def ex_abs_underlying_def)
+  by meson
+
+lemma hoare_from_abs:
+  assumes abs_valid: "\<lbrace>P and R\<rbrace> f \<lbrace>\<lambda>rv. Q rv and S\<rbrace>"
+  assumes cross: "\<And>s s'. \<lbrakk>(s, s') \<in> srel; P' s'; R' s'\<rbrakk> \<Longrightarrow> P s"
+                 "\<And>s s' r r'. \<lbrakk>(s, s') \<in> srel; rrel r r'; Q r s; S s\<rbrakk> \<Longrightarrow> Q' r' s'"
+  assumes corres: "corres_underlying srel False True rrel G G' f f'"
+  shows "\<lbrace>P' and G' and R' and ex_abs_underlying srel (G and R)\<rbrace> f' \<lbrace>Q'\<rbrace>"
```
corlewis, line 230: A bit more complicated here, but for backwards reasoning I think the most useful order of assumptions would be `corres`, `cross2`, `abs_valid`, and then `cross1`.
This one probably doesn't matter much in practice, since I imagine the `cross` assumptions would generally be instantiated directly with existing rules.

corlewis (approved):
The names seem reasonable to me as well.

lib/ExtraCorres.thy
michaelmcinerney: This didn't work, unfortunately, and I can't find anything much nicer than what was there originally

michaelmcinerney: After rearranging some of the assumptions in other rules and using some extra manual instantiation, I got this working

lib/ExtraCorres.thy
michaelmcinerney: I was a bit lazy with this. These can be made `nf nf'` as long as we have `nf ==> no_fail (P and R and G) f`. I think this is optimal

michaelmcinerney:
I have responded to the comments, and also added a type signature for `ex_abs_underlying`

lsf37:
This looks ready to merge now. @michaelmcinerney can you please squash that one commit?

michaelmcinerney:
I tried to squash that one commit, but I'm not sure what's happened, with all the other merge stuff

lsf37:
hm, yes, that looks not good :-) Let me check.

lsf37:
Ok, I think I fixed it. (Basically started a new branch from master, cherry-picked your two commits and squashed them)

michaelmcinerney:
Thanks!

PR #524, opened 2022-10-05
Monadic_Rewrite overhaul

Xaphiosis:
This is the final significant PR for the Monadic_Rewrite overhaul funded by the seL4 Foundation. There is one item outstanding: re-indenting `Fastpath_Equiv` and `Fastpath_C` but I don't want the noise of that in this PR, and I want to hack together a little indenter improvement to fix up mis-indented comments / subgoal afterwards, for which there's no time right now.

Generally:
* we now have useful tactics to work with most simple cases of monadic rewriting, plus other quality-of-life such as `wp_pre` understanding `monadic_rewrite_guard_imp`
* the rules have names that make more sense and are at least consistent with each other
* a number of rules were redundant or had assumptions that were too strong
* monadic_rewrite proofs have generally gotten shorter, but the extreme specificity of walking over the code structure makes automation prohibitive in a fair few cases, more than I initially estimated

The hope is the availability of this tech will improve willingness to use it for the simple cases which used to be a huge pain previously.

The overhaul includes work that was merged as previous PRs:
* #487 
* #516 
* #517
* #518
* #521 

Unhappinesses that were not solved (and I don't know how to solve at a fundamental level):
* fragility of name bindings inside a monad: these are just lambdas and follow eta-contraction and other random rules
    * for symbolic execution at the head, we have done well to preserve the names in the monad even in adverse circumstances (see `no_name_eta` for example)
    * unfortunately rules that "step through", like `monadic_rewrite_bind_tail` can only make use of the name capturing trick (which is deep in Isabelle and used for stuff like `allE` preserving names) on the first subgoal, with later subgoals getting `rv` or `x` or whatever is the name in the rule
    * this means that targeted tactics like `monadic_rewrite_l` will damage all names until the point where the supplied rule matches
    * there might be a way to hack past this in the future, but it will always be a hack, because we do not have an actual name binding, just a lambda in a shallow embedding, so we'd have to dive in and try extract it, then apply a rename that sticks to other goals... yuck. Try capture the name in some kind of deeper embedding? no idea how that can work; etc.

### DIFF
--- a/lib/MonadicRewrite.thy
+++ b/lib/MonadicRewrite.thy
@@ -40,7 +40,7 @@ lemma monadic_rewrite_impossible:
   "monadic_rewrite F E \<bottom> f g"
   by (clarsimp simp: monadic_rewrite_def)
 
-lemma monadic_rewrite_imp:
+lemma monadic_rewrite_guard_imp:
   "\<lbrakk> monadic_rewrite F E Q f g; \<And>s. P s \<Longrightarrow> Q s \<rbrakk> \<Longrightarrow> monadic_rewrite F E P f g"
   by (auto simp add: monadic_rewrite_def)
 
@@ -52,7 +52,7 @@ lemma monadic_rewrite_trans:
 lemma monadic_rewrite_trans_dup:
   "\<lbrakk> monadic_rewrite F E P f g; monadic_rewrite F E P g h \<rbrakk>
    \<Longrightarrow> monadic_rewrite F E P f h"
-  by (rule monadic_rewrite_imp, (rule monadic_rewrite_trans; assumption), simp)
+  by (rule monadic_rewrite_guard_imp, (rule monadic_rewrite_trans; assumption), simp)
 
 lemma monadic_rewrite_from_simple:
   "P \<longrightarrow> f = g \<Longrightarrow> monadic_rewrite F E (\<lambda>_. P) f g"
@@ -91,18 +91,18 @@ lemma monadic_rewrite_exists_v:
   "\<lbrakk> \<And>v. monadic_rewrite E F (Q v) f g \<rbrakk>
    \<Longrightarrow> monadic_rewrite E F (\<lambda>x. (\<exists>v. P v x) & (\<forall>v. P v x \<longrightarrow> Q v x)) f g"
   by (rule monadic_rewrite_name_pre)
-     (fastforce elim: monadic_rewrite_imp)
+     (fastforce elim: monadic_rewrite_guard_imp)
 
-lemma monadic_rewrite_weaken:
+lemma monadic_rewrite_weaken_flags:
   "monadic_rewrite (F \<and> F') (E \<or> E') P f g
    \<Longrightarrow> monadic_rewrite F' E' P f g"
   by (auto simp: monadic_rewrite_def)
 
-lemma monadic_rewrite_weaken2:
+lemma monadic_rewrite_weaken_flags':
   "monadic_rewrite F E P f g
    \<Longrightarrow> monadic_rewrite F' E' ((\<lambda>_. (F \<longrightarrow> F') \<and> (E' \<longrightarrow> E)) and P) f g"
   apply (rule monadic_rewrite_gen_asm)
-  apply (rule monadic_rewrite_weaken[where F=F and E=E])
+  apply (rule monadic_rewrite_weaken_flags[where F=F and E=E])
   apply auto
   done
 
@@ -162,7 +162,7 @@ lemmas monadic_rewrite_bindE_head
 lemma monadic_rewrite_bind2:
   "\<lbrakk> monadic_rewrite F E P f g; \<And>x. monadic_rewrite F E (Q x) (h x) (j x); \<lbrace>R\<rbrace> f \<lbrace>Q\<rbrace> \<rbrakk>
    \<Longrightarrow> monadic_rewrite F E (P and R) (f >>= (\<lambda>x. h x)) (g >>= (\<lambda>x. j x))"
-  apply (rule monadic_rewrite_imp)
+  apply (rule monadic_rewrite_guard_imp)
    apply (rule monadic_rewrite_trans)
     apply (erule(1) monadic_rewrite_bind_tail)
    apply (erule monadic_rewrite_bind_head)
@@ -173,7 +173,7 @@ lemma monadic_rewrite_named_bindE:
   "\<lbrakk> monadic_rewrite F E ((=) s) f f';
     \<And>rv s'. (Inr rv, s') \<in> fst (f' s) \<Longrightarrow> monadic_rewrite F E ((=) s') (g rv) (g' rv) \<rbrakk>
    \<Longrightarrow> monadic_rewrite F E ((=) s) (f >>=E (\<lambda>rv. g rv)) (f' >>=E g')"
-  apply (rule monadic_rewrite_imp)
+  apply (rule monadic_rewrite_guard_imp)
    apply (erule_tac R="(=) s" and Q="\<lambda>rv s'. (Inr rv, s') \<in> fst (f' s)" in monadic_rewrite_bindE)
     apply (rule monadic_rewrite_name_pre)
     apply clarsimp
@@ -193,9 +193,9 @@ lemma monadic_rewrite_add_return:
 (* FIXME: poorly named, super-specific (could do this with maybe one bind?), used in Ipc_C *)
 lemma monadic_rewrite_do_flip:
   "monadic_rewrite E F P (do c \<leftarrow> j; a \<leftarrow> f; b \<leftarrow> g c; return (a, c) od)
-                         (do c \<leftarrow> j; b \<leftarrow> g c; a \<leftarrow> f; return (a, c) od)
-   \<Longrightarrow> monadic_rewrite E F P (do c \<leftarrow> j; a \<leftarrow> f; b \<leftarrow> g c; h a c od)
-                            (do c \<leftarrow> j; b \<leftarrow> g c; a \<leftarrow> f; h a c od)"
+                         (do c \<leftarrow> j; b \<leftarrow> g c; a \<leftarrow> f; return (a, c) od) \<Longrightarrow>
+   monadic_rewrite E F P (do c \<leftarrow> j; a \<leftarrow> f; b \<leftarrow> g c; h a c od)
+                         (do c \<leftarrow> j; b \<leftarrow> g c; a \<leftarrow> f; h a c od)"
   apply (drule_tac h="\<lambda>(a, b). h a b" in monadic_rewrite_bind_head)
   apply (simp add: bind_assoc)
   done
@@ -228,8 +228,8 @@ lemma monadic_rewrite_symb_exec_pre:
   assumes inv: "\<And>s. g \<lbrace> (=) s\<rbrace>"
        and ef: "empty_fail g"
        and rv: "\<lbrace>P\<rbrace> g \<lbrace>\<lambda>y s. y \<in> S\<rbrace>"
-       and h': "\<And>y. y \<in> S \<longrightarrow> h y = h'"
-  shows "monadic_rewrite True True P (g >>= h) h'"
+       and h': "\<And>rv. rv \<in> S \<longrightarrow> h rv = h'"
+  shows "monadic_rewrite True True P (g >>= (\<lambda>rv. h rv)) h'"
 proof -
   have P: "\<And>s v. \<lbrakk> P s; v \<in> fst (g s) \<rbrakk> \<Longrightarrow> split h v = h' s"
     apply clarsimp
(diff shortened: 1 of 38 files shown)


Xaphiosis:
@mbrcknl this might be of limited interest to you, and you might have other interesting ideas for the future, but I know you're very busy :)

@michaelmcinerney : you wanted to learn a bit more about tactical stuff works, and a bunch of these renames will bite on the next merge with MCS, so you should be aware in order to yell at me then

lsf37:
For some reason the GitHub tests didn't trigger apart from the DCO. I'm very confused by this -- shouldn't be possible and the tests seem to be working fine on the previous PRs (from forks as well as internal branches). Was there anything special in the UI when you opened this PR?

lib/MonadicRewrite.thy
```diff
@@ -76,12 +77,11 @@ lemma monadic_rewrite_is_refl:
   "x = y \<Longrightarrow> monadic_rewrite F E \<top> x y"
   by (simp add: monadic_rewrite_refl)
 
-lemma monadic_rewrite_refl3:
+(* precondition implies reflexivity *)
+lemma monadic_rewrite_pre_imp_refl:
   "\<lbrakk> \<And>s. P s \<Longrightarrow> f s = g s \<rbrakk> \<Longrightarrow> monadic_rewrite F E P f g"
```
lsf37, line 82: `refl` is not quite right for this one (old name, was never really right). Reflexivity requires things to be syntactically equal, i.e. `x = x`, otherwise it's just equality (`x = y`) not reflexivity. Maybe `monadic_rewrite_pre_imp_eq`?

proof/crefine/ARM/IsolatedThreadAction.thy
```diff
@@ -690,12 +686,10 @@ lemma transferCaps_simple_rewrite:
...
   including no_pre
+  supply empty_fail_getReceiveSlots[wp] (* FIXME *)
   apply (rule monadic_rewrite_gen_asm)
-  apply (rule monadic_rewrite_imp)
-   apply (rule monadic_rewrite_trans)
-    apply (simp add: transferCaps_simple, rule monadic_rewrite_refl)
-   apply (rule monadic_rewrite_symb_exec2, (wp empty_fail_getReceiveSlots)+)
-   apply (rule monadic_rewrite_refl)
+  apply (simp add: transferCaps_simple)
+  apply (monadic_rewrite_symb_exec_l_drop, rule monadic_rewrite_refl)
```
lsf37, line 692: This is so much nicer than before, and lots of these everywhere. Very nice!

lib/MonadicRewrite.thy
Xaphiosis: this one is actually used. Going to be a pain to rename and squash the changes in, but it's probably a good point.

Xaphiosis: > For some reason the GitHub tests didn't trigger apart from the DCO. I'm very confused by this -- shouldn't be possible and the tests seem to be working fine on the previous PRs (from forks as well as internal branches). Was there anything special in the UI when you opened this PR?

Sorry, it was struggling from the beginning. I waited for ages for the whitespace test to fail, and never saw the proofs run. After I fixed the whitespace issue and pushed, I only saw DCO run and that's it.

lsf37 (approved):
Nice work! I don't have many detailed comments, because I've seen most of this before already. This should make future monadic_rewrite proofs a lot smoother to write.

lib/MonadicRewrite.thy
lsf37: The rename turned out not too bad, I think.

lib/MonadicRewrite.thy
```diff
@@ -613,77 +674,227 @@ lemma monadic_rewrite_weaken_failure:
...
+method monadic_rewrite_step =
+  determ \<open>rule monadic_rewrite_bind_tail monadic_rewrite_bindE_tail\<close>
+
+method monadic_rewrite_solve_head methods m =
+  (rule monadic_rewrite_bind_head monadic_rewrite_bindE_head)?,
+  solves \<open>m, (rule monadic_rewrite_refl)?\<close>
+
+(*
+The most common way of performing monadic rewrite is by doing a pass over the LHS/RHS via some
+a setup step, such as monadic_rewrite_gen_asm + monadic_rewrite_trans (or transverse).
```
corlewis, line 732: I think this got mixed up a bit during a rewrite, it was originally `some kind of setup step`
```suggestion
The most common way of performing monadic rewrite is by doing a pass over the LHS/RHS via
a setup step, such as monadic_rewrite_gen_asm + monadic_rewrite_trans (or transverse).
```

proof/crefine/ARM/IsolatedThreadAction.thy
```diff
@@ -633,6 +633,18 @@ lemma getHWASID_isolatable:
             | simp add: page_directory_at_partial_overwrite)+
   done
 
+lemma getTCB_threadGet:
+  "do
+     tcbobj \<leftarrow> getObject t;
+     x \<leftarrow> f (atcbVCPUPtr (tcbArch tcbobj));
+     g x
+   od = do
+     vcpu_ptr \<leftarrow> threadGet (atcbVCPUPtr o tcbArch) t;
```
corlewis, line 643: I'm a bit confused by `atcbVCPUPtr` appearing in `ARM`.

proof/crefine/ARM/IsolatedThreadAction.thy
```diff
@@ -673,11 +685,11 @@ lemma transferCaps_simple_rewrite:
    (transferCaps mi caps ep r rBuf)
    (return (mi \<lparr> msgExtraCaps := 0, msgCapsUnwrapped := 0 \<rparr>))"
   including no_pre
+  supply empty_fail_getReceiveSlots[wp] (* FIXME *)
```
corlewis, line 689: Just confirming, is the FIXME here that `empty_fail_getReceiveSlots` should be globally `wp`?

corlewis (approved):
Very nice, and the many helpful comments and commit messages are greatly appreciated!

proof/crefine/ARM/Fastpath_Equiv.thy
```diff
@@ -417,48 +410,33 @@ lemma schedule_rewrite_ct_not_runnable':
...
+            \<open>wpsimp simp: isHighestPrio_def'
+                    wp: hoare_vcg_imp_lift hoare_vcg_disj_lift threadGet_wp''
+                        scheduleSwitchThreadFastfail_False_wp\<close>)
+   (* fastpath: no scheduleChooseNewThread *)
+   apply (monadic_rewrite_l monadic_rewrite_if_l_False
+            \<open>wpsimp simp: isHighestPrio_def'
+                    wp: hoare_vcg_imp_lift hoare_vcg_disj_lift threadGet_wp''
+                        scheduleSwitchThreadFastfail_False_wp\<close>)
+   (* remove no-ops *)
+   apply (repeat 10 monadic_rewrite_symb_exec_l) (* until switchToThread *)
```
michaelmcinerney, line 429: I've seen `repeat n blah` a few times now. I'm not sure `repeat_unless` is always the best way to go, but would it be possible to repeat something until a certain term is seen (`switchToThread` in this case)?

proof/crefine/ARM/Fastpath_Equiv.thy
```diff
@@ -553,236 +527,229 @@ lemma fastpath_callKernel_SysCall_corres:
...
-                   apply (rule ext)
-                   apply (simp add: badgeRegister_def msgInfoRegister_def
-                                   ARM.badgeRegister_def
-                                   ARM.msgInfoRegister_def
-                            split: if_split)
-                  apply simp
-                 apply (wp | simp cong: if_cong bool.case_cong
-                           | rule getCTE_wp' gts_wp' threadGet_wp
-                                 getEndpoint_wp)+
+                   apply (rule_tac P="\<lambda>s. ksSchedulerAction s = ResumeCurrentThread
```
michaelmcinerney, line 724: I think the indentation got a little bit messed up here

michaelmcinerney (approved):
Very cool. 

I think the `handleFaultReply'` proof is the only big proof in MCS that will be majorly affected, and we might want to have a look at that one together at some point, anyway

proof/crefine/ARM/IsolatedThreadAction.thy
Xaphiosis: This was a derp, those variables are blue. Thanks for catching this!

proof/crefine/ARM/IsolatedThreadAction.thy
Xaphiosis: That is correct

lib/MonadicRewrite.thy
```diff
@@ -368,19 +408,39 @@ lemma monadic_rewrite_if:
...
 
-lemma monadic_rewrite_if_rhs:
+lemma monadic_rewrite_if_r:
   "\<lbrakk> P \<Longrightarrow> monadic_rewrite F E Q a b; \<not> P \<Longrightarrow> monadic_rewrite F E R a c \<rbrakk>
    \<Longrightarrow> monadic_rewrite F E (\<lambda>s. (P \<longrightarrow> Q s) \<and> (\<not> P \<longrightarrow> R s)) a (If P b c)"
   by (cases P, simp_all)
 
+lemma monadic_rewrite_if_r_True:
+  "\<lbrakk> P \<Longrightarrow> monadic_rewrite F E Q a b \<rbrakk>
+   \<Longrightarrow> monadic_rewrite F E ((\<lambda>_. P) and Q) a (If P b c)"
```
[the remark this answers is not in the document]
Xaphiosis: I'm not sure... it seems like when we talk about `If` from the perspective of a construct (i.e. there's no further detail on the contents of the predicate and branches) we stick to `If` because it's shorter. I could change it, but not sure it helps much.

proof/crefine/ARM/Fastpath_Equiv.thy
Xaphiosis: Definitely possible, but you have to match on the whole conclusion pattern, something like using `repeat_unless` with `has_concl` on `monadic_rewrite ?F ?E ?P (switchToThread ?t >>= ?c1) (?c2)`. Do you think it's worth it?

proof/crefine/ARM/Fastpath_Equiv.thy
Xaphiosis: Yup. I'll clear it up in a separate pass.

proof/crefine/ARM/Fastpath_Equiv.thy
michaelmcinerney: Ah, probably not worth it, then

lib/MonadicRewrite.thy
```diff
@@ -224,141 +219,186 @@ lemma monadic_rewrite_modify_noop:
...
+(* perform symbolic execution on RHS, proving a specific value is returned *)
+lemmas monadic_rewrite_symb_exec_r_known
+  = monadic_rewrite_symb_exec_r_known_E
+    monadic_rewrite_symb_exec_r_known_nE
 
-(* FIXME merge into below upon change-over desribed above *)
-lemmas monadic_rewrite_symb_exec_l'_TT
-  = monadic_rewrite_symb_exec_l'_preserve_names[where P'="\<top>" and F=True, simplified]
+lemma monadic_rewrite_symb_exec_l_drop_F:
+  "\<lbrakk> monadic_rewrite True E P g h; \<And>P. \<lbrace>P\<rbrace> m \<lbrace>\<lambda>_. P\<rbrace>; empty_fail m \<rbrakk>
```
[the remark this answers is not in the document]
Xaphiosis: Thank you, I fixed up the subsequent ones too.

proof/crefine/ARM/IsolatedThreadAction.thy
```diff
@@ -708,6 +708,7 @@ lemma copyMRs_simple:
   apply (simp add: upto_enum_def mapM_Nil)
   done
 
+thm monadic_rewrite_symb_exec_l_known_F
```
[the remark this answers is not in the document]
Xaphiosis: I'll try do that surgery. To remember for future events, @lsf37 what do you think our policy should be on this situation?

Xaphiosis:
All PR issues should now be addressed. I rebased, and did a re-indent pass on Fastpath_Equiv. Because of the rebase I'll wait for tests to succeed (am running them in parallel on a faster machine).
For a better indentation script, I have to wait until the newer Isabelle and the Scala 3 stuff clears.

proof/crefine/ARM/IsolatedThreadAction.thy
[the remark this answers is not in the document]
lsf37: I'm not too fussed about it. It's nicer if it looks like it was never introduced, but we don't have to be absolutely perfect with history. Main goal is that it makes sense and is useful.

lsf37 (approved):
Good to go from my side!

PR #527, opened 2022-10-13
MCS: synchronise proof directories with master

corlewis:
Several proof directories on the rt branch had diverged from master in unexpected ways. This commit manually brings them back in sync.

The specific directories are access-control, bisim, dpolicy, drefine, and infoflow.

### DIFF
--- a/proof/access-control/CNode_AC.thy
+++ b/proof/access-control/CNode_AC.thy
@@ -123,7 +123,7 @@ lemma set_original_integrity_autarch:
 
 lemma set_original_integrity:
   "\<lbrace>integrity aag X st and cdt_change_allowed' aag slot\<rbrace>
-      set_original slot orig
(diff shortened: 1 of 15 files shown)


michaelmcinerney (approved):
👍

lsf37 (approved):
Good to bring them in sync again. A lot of these look like wrong indentation got fixed on the MCS side. No problem reverting this here, but we should probably add a todo to fix it in master instead, so it gets carried over in the next merge.

corlewis:
> Good to bring them in sync again. A lot of these look like wrong indentation got fixed on the MCS side. No problem reverting this here, but we should probably add a todo to fix it in master instead, so it gets carried over in the next merge.

I actually thought it was the other way around, the indentation was fixed on master and had been previously accidentally reverted on MCS.

PR #529, opened 2022-10-19
Merge master into rt

corlewis:
The final commit in this PR is new and updates the proofs for the changes from https://github.com/seL4/l4v/pull/526.


lsf37 (approved):
Happy with that (if it passes the proof checks :-)). I assume the gitlint failures are the same as the ones for the original PR.

corlewis:
> Happy with that (if it passes the proof checks :-)). I assume the gitlint failures are the same as the ones for the original PR.

Yes, that's right, as is the theory linter failure.

PR #530, opened 2022-10-30
progress on aarch64 invariants

lsf37:
This PR clears all sorries up to the theory `ArchFinalise_AI` and in addition clears `ArchEmptyFail`, `ArchInterrupt`, `ArchTcb`, `ArchIPC`, and `ArchVSpaceEntries`.

In the spec, we add an additional read operation to model the atomic ASID pool field update that happens C, and we correct the definition of `arch_is_frame_type` to properly take into account the presence of VCPU objects.

### DIFF
--- a/proof/invariant-abstract/AARCH64/ArchFinalise_AI.thy
+++ b/proof/invariant-abstract/AARCH64/ArchFinalise_AI.thy
@@ -50,7 +50,7 @@ lemma invs_arm_asid_table_unmap:
   "invs s
    \<and> is_aligned base asid_low_bits
    \<and> (\<forall>asid_low. vmid_for_asid s (asid_of (asid_high_bits_of base) asid_low) = None)
-   \<and> tab = arm_asid_table (arch_state s)
+   \<and> tab = asid_table s
      \<longrightarrow> invs (s\<lparr>arch_state := arch_state s\<lparr>arm_asid_table := tab(asid_high_bits_of base := None)\<rparr>\<rparr>)"
   apply (clarsimp simp: invs_def valid_state_def valid_arch_caps_def)
   apply (strengthen valid_asid_map_unmap valid_vspace_objs_unmap_strg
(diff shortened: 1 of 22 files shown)


lsf37:
Theory linter failure expected for 2 introduced sorries (105 removed).

Xaphiosis (approved):
Already reviewed internally, re-applying my stamp of approval here.

proof/invariant-abstract/AARCH64/ArchSchedule_AI.thy
```diff
@@ -37,17 +37,11 @@ lemma arch_stt_invs [wp,Schedule_AI_asms]:
 
 lemma arch_stt_tcb [wp,Schedule_AI_asms]:
   "\<lbrace>tcb_at t'\<rbrace> arch_switch_to_thread t' \<lbrace>\<lambda>_. tcb_at t'\<rbrace>"
```
michaelmcinerney, line 39: nitpick: use invariant notation for the lemmas in this file?

michaelmcinerney (approved):
Looks great!

proof/invariant-abstract/AARCH64/ArchVSpaceEntries_AI.thy
```diff
@@ -31,7 +31,7 @@ lemmas obj_valid_vspace_simps[simp]
         [split_simps Structures_A.kernel_object.split
                      arch_kernel_obj.split]
 
-abbreviation
+locale_abbrev
```
corlewis, line 34: Would I be able to get a quick reminder about what `locale_abbrev` is?

proof/invariant-abstract/AARCH64/ArchFinalise_AI.thy
```diff
@@ -117,32 +182,35 @@ lemma set_asid_pool_unmap:
...
 lemma set_pt_tcb_at:
   "\<lbrace>\<lambda>s. P (ko_at (TCB tcb) t s)\<rbrace> set_pt a b \<lbrace>\<lambda>_ s. P (ko_at (TCB tcb) t s)\<rbrace>"
   by (wpsimp simp: set_pt_def obj_at_def wp: set_object_wp)
 
 lemma set_vcpu_tcb_at_arch: (* generalise? this holds except when the ko is a vcpu *)
   "\<lbrace>\<lambda>s. P (ko_at (TCB tcb) t s)\<rbrace>
-      set_vcpu p v
-        \<lbrace>\<lambda>_ s. P (ko_at (TCB tcb) t s)\<rbrace>"
+   set_vcpu p v
+   \<lbrace>\<lambda>_ s. P (ko_at (TCB tcb) t s)\<rbrace>"
```
corlewis, line 213: Very unimportant, but with this lemma already changing it could also be changed to invariant notation.

proof/invariant-abstract/AARCH64/ArchTcb_AI.thy
```diff
@@ -21,23 +21,19 @@ lemma activate_idle_invs[Tcb_AI_asms]:
   by (simp add: arch_activate_idle_thread_def)
 
 
-lemma empty_fail_getRegister [intro!, simp, Tcb_AI_asms]:
-  "empty_fail (getRegister r)"
-  by (simp add: getRegister_def)
-
+declare getRegister_empty_fail [Tcb_AI_asms]
```
corlewis, line 24: Does this mean that this is now proven before this file? If so, could it be removed from the locale completely, and then this declaration wouldn't be needed?

corlewis (approved):
Very nice progress!

proof/invariant-abstract/AARCH64/ArchVSpaceEntries_AI.thy
lsf37: abbreviations defined in locales usually don't contract (are input-only). That makes sense for a general setting where the abbreviation might not work outside the locale. `locale_abbrev` does contract inside the locale.

proof/invariant-abstract/AARCH64/ArchTcb_AI.thy
lsf37: It's only proved before this file in AARCH64 (in some crunch somewhere), but it's still in the interface, because other architectures have not necessarily seen that proof before at this point.

proof/invariant-abstract/AARCH64/ArchSchedule_AI.thy
lsf37: can do

proof/invariant-abstract/AARCH64/ArchTcb_AI.thy
corlewis: Oh right, of course. Sorry, that should have been obvious.
I guess that this is a basic enough lemma that it would be nice to have this consistent on all the architectures and then out of the interface, but that would be pretty low priority.

proof/invariant-abstract/AARCH64/ArchVSpaceEntries_AI.thy
corlewis: Thanks!

proof/invariant-abstract/AARCH64/ArchTcb_AI.thy
lsf37: Hm, I'm not entirely sure about that. If we do explicitly prove it earlier in the other architectures, we should have some interface somewhere that documents that we want it to exist at a certain point, otherwise we might delete it again thinking it can be done as part of some later `crunch` only to then fail deep in some other seemingly unrelated proof which could be different per platform.

I guess all that is fixable with a comment, so maybe not so bad. But I'd still prefer fewer "works by carefully constructed accident on all architectures" lemmas.

corlewis: I definitely agree that it should be forced to exist, but the interface could be a more lightweight `requalify_facts` instead of being part of the `Tcb_AI` locale. This would be both less text and also require less duplicated boilerplate between all of the architectures.

The downside is that it wouldn't explicitly state what the lemma should look like. In this case `empty_fail` doesn't have preconditions so that shouldn't matter, but I can imagine other cases where lemmas could be accidentally formulated differently on different architectures.

lsf37: `requalify_facts` is a good idea, that would indeed fit everything we need at that point. I'll probably do that in a separate PR.

lsf37:
Review feedback addressed + rebased. I'll do the `getRegister_empty_fail` refactor separately.

PR #533, opened 2022-11-07
Isabelle2022

corlewis:
The update this time was thankfully quite straightforward. Feel free to skip the first and last commits; the first is copied from the AFP and the last is temporary and is just so that the tests work during this PR.


proof/crefine/ARM/Ipc_C.thy
```diff
@@ -2349,7 +2349,7 @@ lemma transferCapsLoop_ccorres:
                  \<inter> \<lbrace>\<acute>destSlot = (if slots = [] then NULL else cte_Ptr (hd slots))
                                 \<and> length slots \<le> 1 \<and> slots \<noteq> [0]\<rbrace>)"
   defines "is_the_ep \<equiv> \<lambda>cap. isEndpointCap cap \<and> ep \<noteq> None \<and> capEPPtr cap = the ep"
-  defines "stable \<equiv> \<lambda>scap excap. excap \<noteq> scap \<longrightarrow> excap = maskedAsFull scap scap"
+  defines "stable_masked \<equiv> \<lambda>scap excap. excap \<noteq> scap \<longrightarrow> excap = maskedAsFull scap scap"
```
corlewis, line 2352: There is now an alternative constant called stable that is now visible in HOL, so I just changed the name of the local definition in this lemma. Suggestions welcome for alternative names.

lsf37: I'm fine with `stable_masked`, it describes (some of) what is going on.

lsf37 (approved):
Cool, looks like this was a very smooth update!

tools/c-parser/umm_heap/Arrays.thy
```diff
@@ -273,11 +273,14 @@ next
   show "set_array \<circ> map_array f = (`) f \<circ> set_array"
     by (rule map_array_set_img)
 next
-  show "card_order (BNF_Cardinal_Arithmetic.csum natLeq (card_of UNIV))"
-    by (simp add: card_of_card_order_on card_order_csum natLeq_card_order)
+  show "card_order (BNF_Cardinal_Arithmetic.csum natLeq (card_suc (card_of UNIV)))"
```
Xaphiosis, line 276: can you comment on why we went to `card_of ()` to `card_suc (card_of ())`?

tools/autocorres/autocorres.ML
```diff
@@ -114,7 +114,7 @@ fun named_option parser name elem_desc=
   expect "\"=\"" elem_desc parser)
 
 (* Generic parser for "NAME = STRING ..." *)
-val named_opt = named_option (Scan.repeat Parse.text)
+val named_opt = named_option (Scan.repeat Parse.embedded)
```
Xaphiosis, line 117: for my own knowledge: did `Parse.text` get renamed to `Parse.embedded`, or is there nuance on this one?

lib/Word_Lib/Word_Lemmas.thy
```diff
@@ -622,16 +622,13 @@ lemma shiftr1_0_or_1:"(x::('a::len) word) >> 1 = 0 \<Longrightarrow> x = 0 \<or>
...
-  apply (simp add: drop_bit_take_bit drop_bit_Suc)
-  done
+  by (auto simp add: bit_0 shiftr_def drop_bit_Suc ac_simps elim: evenE)
 
 lemma shiftr1_0_imp_only_lsb:"((x::('a::len) word) + 1) >> 1 = 0 \<Longrightarrow> x = 0 \<or> x + 1 = 0"
   by (metis One_nat_def shiftr1_0_or_1 word_less_1 word_overflow)
 
 lemma shiftr1_irrelevant_lsb': "\<not> (bit (x::('a::len) word) 0) \<Longrightarrow> x >> 1 = (x + 1) >> 1"
-  by (metis shiftr1_irrelevant_lsb)
+  using shiftr1_irrelevant_lsb [of x] by simp
```
Xaphiosis, line 631: suggest `using shiftr1_irrelevant_lsb[of x] by simp`
I used to advocate for the `rule [modifier]` style vs `rule[modifier]`, but after some years I came to the realisation that was not such a good idea

Xaphiosis (approved):
Looks good, and a small update too.
For your consideration: squashing both CRefine commits, not sure why ARM is separate.

tools/c-parser/umm_heap/Arrays.thy
corlewis: I have to admit that this change is mostly based off similar changes to BNF declarations in HOL, but the relevant NEWS item is
```
BNFs now require a strict cardinality bound (<o instead of \<le>o).
    Minor INCOMPATIBILITY for existing manual BNF declarations.
```

tools/autocorres/autocorres.ML
corlewis: Looking at the commit history, `Parse.text` was removed as part of a change which `discontinued old-style {* verbatim *} tokens`. That change then replaced uses of `Parse.text` with `Parse.embedded`, so I just did the same thing.

lib/Word_Lib/Word_Lemmas.thy
corlewis: `Word_Lib` is copied in from the AFP, so outside of breaking compatibility changes I think we would prefer to keep it the same.

I would agree with that change in general however. I prefer modifiers without the whitespace, since for me it makes it easier to take things in at a glance.

corlewis:
> Looks good, and a small update too. For your consideration: squashing both CRefine commits, not sure why ARM is separate.

Sure, happy to squash them. Being separate is mostly an artefact of doing ARM first and committing it for ease of testing.

lib/Word_Lib/Word_Lemmas.thy
lsf37: Currently it'd be Ok to tweak still, I haven't synced our other changes with upstream yet and need to do that anyway. (But yes, all of the Word_Lib changes are verbatim from the AFP)

Xaphiosis: Yeah, ok, let's keep it the same then, I wasn't quite sure what was going on, but if we're pulling from upstream, we're pulling from upstream.

tools/autocorres/autocorres.ML
Xaphiosis: Thank you!

tools/c-parser/umm_heap/Arrays.thy
Xaphiosis: Thanks. I don't really know much about BNF declarations in Isabelle. All good.

PR #535, opened 2022-11-11
Merge into rt branch

corlewis:
In particular, this pulls in the monadic rewrite overhaul and the Isabelle2022 update.

If I've done things correctly, the second last commit is the only one with new work that would need to be reviewed.


lsf37 (approved):
Looking good, I've checked the fixup commit, otherwise these are all from the `master` branch or will be dropped. Nice work!

corlewis:
Oh damn, I just remembered that I'll need to update ARM refine as well. That will hopefully be practically the same as RISCV64, but will take some processing time .

michaelmcinerney (approved):
Lovely!

corlewis:
I closed this PR because the tests weren't firing correctly, I think due to the fact that it was being merged from a fork. It is now https://github.com/seL4/l4v/pull/536.

PR #536, opened 2022-11-11
Merge into rt branch

corlewis:
In particular, this pulls in the monadic rewrite overhaul and the Isabelle2022 update.

If I've done things correctly, the second last commit is the only one with new work that would need to be reviewed.


corlewis:
I've redone this PR, this time from an in-repo branch, so that the push tests with the correct version of Isabelle would run.

There is also one new commit (https://github.com/seL4/l4v/pull/536/commits/d1cf6ea0b1e81a19752044646c9229158a9a8d52) that reverts a merge mistake in lib and updates the proofs for ARM. For the latter, there were some additional lemmas in Invariants_AI and KHeap_AI required so that the proof of `dmo_invs_lift` still worked.

michaelmcinerney (approved):

spec/design/m-skel/ARM/MachineTypes.thy
```diff
@@ -82,7 +82,7 @@ axiomatization
 where
   irq_oracle_max_irq: "\<forall> n. (irq_oracle n) <= ARM.maxIRQ"
 
-text {* The values the timer device will return (how much time passed since last query) *}
+text \<open> The values the timer device will return (how much time passed since last query) \<close>
```
Xaphiosis, line 85: minor, but do we want the spaces around `<` and `>` for `text` blocks?

Xaphiosis (approved):
I mainly reviewed the post-merge fixup commit. Don't forget to squash the squashes.

lib/Lib.thy
```diff
@@ -2761,6 +2761,19 @@ lemma distinct_map_enum:
...
+  they match goals up to equality.
+
+  Why does this lemma exist when lemmas like @{thm subst} already exist?
+  Because patterns like `?P ?x` usually have multiple unifiers with the
+  conclusion of a rule, whereas `(\<And>x. ?P x x)` only has one. This means
+  you can "inline" `equify` like this: `my_cool_thm[THEN equify]`.
+\<close>
+lemma equify:
+  "(\<And>x. P x x) \<Longrightarrow> x = x' \<Longrightarrow> P x x'"
+  by simp
```
Xaphiosis, line 2775: is this an MCS-only thing? I'm not seeing it on master. It looks like it has Ed-style comment indentation (including the extra `>` on its own line which we don't do anymore)

spec/design/m-skel/ARM/MachineTypes.thy
corlewis: I don't have any strong opinions, happy to remove the spaces if you think that would be more consistent.

lsf37 (approved):
Good from my side, I have nothing further to add after Raf's comments.

spec/design/m-skel/ARM/MachineTypes.thy
lsf37: Minor, but I'd be for removing the spaces for more consistency.

lib/Lib.thy
corlewis: Yeah, it was added on MCS and then I accidentally removed it during the merge.

I can change the indentation as part of the merge if you'd prefer. Would it be something like this?
```suggestion
―‹Helps transform facts with multiple instances of a variable so that
  they match goals up to equality.

  Why does this lemma exist when lemmas like @{thm subst} already exist?
  Because patterns like `?P ?x` usually have multiple unifiers with the
  conclusion of a rule, whereas `(⋀x. ?P x x)` only has one. This means
  you can "inline" `equify` like this: `my_cool_thm[THEN equify]`.›
```

Xaphiosis: Probably for a `lemma` we should stick to `text` or `(* *)` ... so for this one, I'd be fine with only moving the `>` to the previous line so it isn't on its own.

PR #537, opened 2022-11-29
cparser: annotate AST printing for top-level decls

Xaphiosis:
In order to compare ASTs for appearance/disappearance/modification of declarations, it is easier to have the annotations obvious to any external tool for ease of parsing.

Annotations take the form:
"##<decl_type>: <name>", e.g. "##Function: ctzl"


lsf37:
I think you have already added this patch, the merge diff is only one line. Yup, here it is: https://github.com/seL4/l4v/pull/474

PR #538, opened 2022-11-29
c-parser: fix default target for standalone parser

lsf37:
The default (=first) Makefile target for the standalone parser was `all`, which gains additional dependencies in the included Makefile. We want `make` in this directory to just build the standalone parser, so we set `stp_all` as the default.


Xaphiosis (approved):
fingers crossed

PR #539, opened 2022-11-30
haskell-translator: suppress known-good warnings

lsf37:
MCS has a few more instances of type assignments that are handled manually, and the lambda `\usage` confuses the C preprocessor (potential unicode escape sequence). Switch these off, because we know they are fine.

### DIFF
--- a/tools/haskell-translator/lhs_pars.py
+++ b/tools/haskell-translator/lhs_pars.py
@@ -690,6 +690,7 @@ def newtype_transform(d):
 known_type_assignments = [
     'domain_schedule',
     'kernel',
+    'kernel_r',
     'kernel_f',
     'kernel_f f',  # there is a KernelF instance that gets transformed into this
     'kernel_init',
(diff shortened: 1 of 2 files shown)


tools/haskell-translator/lhs_pars.py
```diff
@@ -706,7 +710,8 @@ def typename_transform(line, header, d):
         [oldtype] = line.split()
     except:
         if header not in known_type_assignments:
-            warning('type assignment with parameters not supported %s' % d.body, filename)
+            warning('type assignment with parameters not supported %s, (%s)' %
+                    (d.body, header), filename)
```
Xaphiosis, line 714: not following, what do you want printed here?

Xaphiosis (approved):

tools/haskell-translator/lhs_pars.py
lsf37: `header` is the thing it's called on, which turned out to be more useful for debugging than just `d.body` which it printed before.

PR #540, opened 2023-01-08
factor out and generalise definition of pred_conj and friends

lsf37:
The idea is to use the pre-defined boolean algebra class to have `pred_conj` (`_ and _`) not only on one-argument predicates `'s => bool`, but on predicates (functions to bool) with any number of arguments. In particular, this eliminates the need for a second `And` constant and specialised rules for that.

In addition, it turns out that we defined this setup multiple times separately in each monad development instead of once in a library.

The new `pred_conj` ends up as just an abbreviation for `inf_class.inf` from boolean algebras (to be precise, from the semilattice class that is used to construct the boolean algebra class).

Contains:

- lib update
- proof update
- a fix for a problem in AutoCorresDoc ROOT that for some reason hasn't triggered until now

### DIFF
--- a/lib/Fun_Pred_Syntax.thy
+++ b/lib/Fun_Pred_Syntax.thy
@@ -1,0 +1,210 @@
+(*
+ * Copyright 2022, Proofcraft Pty Ltd
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ *)
+
+(* Syntax for using multi-argument functions as predicates, e.g "P and Q" where P and Q are
+   functions to bool, taking one or more parameters. *)
+
+theory Fun_Pred_Syntax
+imports Main
+begin
+
+section \<open>Definitions\<close>
+
+text \<open>
+  Functions are already instances of Boolean algebras and provide all the standard laws one
+  would like to have. Default simplifications are automatic. Search for @{const inf}/
+  @{const sup}/@{const uminus} to find further laws and/or unfold via the definitions below.
+
+  The abbreviations here provide special syntax for the function instance of Boolean
+  algebras only, leaving other instances (such as @{typ bool}) untouched.\<close>
+
+abbreviation pred_conj :: "('a \<Rightarrow> 'b::boolean_algebra) \<Rightarrow> ('a \<Rightarrow> 'b) \<Rightarrow> ('a \<Rightarrow> 'b)" where
+  "pred_conj \<equiv> inf"
+
+abbreviation pred_disj :: "('a \<Rightarrow> 'b::boolean_algebra) \<Rightarrow> ('a \<Rightarrow> 'b) \<Rightarrow> ('a \<Rightarrow> 'b)" where
+  "pred_disj \<equiv> sup"
+
+abbreviation pred_neg :: "('a \<Rightarrow> 'b::boolean_algebra) \<Rightarrow> ('a \<Rightarrow> 'b)" where
+  "pred_neg \<equiv> uminus"
+
+
+text \<open>
+  Lifted True/False: ideally, we'd map these to top/bot, but top/bot are constants and there are
+  currently too many rules and tools that expect these conditions to beta-reduce and match against
+  True/False directly.\<close>
+
+abbreviation (input) pred_top :: "'a \<Rightarrow> bool" where
+  "pred_top \<equiv> \<lambda>_. True"
+
+abbreviation (input) pred_bot :: "'a \<Rightarrow> bool" where
+  "pred_bot \<equiv> \<lambda>_. False"
+
+
+text \<open>Version with two arguments for compatibility. Can hopefully be eliminated at some point.\<close>
+
+abbreviation (input) pred_top2 :: "'a \<Rightarrow> 'b \<Rightarrow> bool" where
+  "pred_top2 \<equiv> \<lambda>_ _. True"
+
+abbreviation (input) pred_bot2 :: "'a \<Rightarrow> 'b \<Rightarrow> bool" where
+  "pred_bot2 \<equiv> \<lambda>_ _. False"
+
+
+section \<open>Syntax bundles\<close>
+
+bundle fun_pred_syntax
+begin
+  (* infixl instead of infixr, because we want to split off conjuncts from the left *)
+  notation pred_conj (infixl "and" 35)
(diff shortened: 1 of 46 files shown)


lib/Monad_WP/TraceMonad.thy
```diff
@@ -941,54 +941,17 @@ where
...
-abbreviation(input)
-  toptop :: "'a \<Rightarrow> 'b \<Rightarrow> bool" ("\<top>\<top>")
-where
-  "\<top>\<top> \<equiv> \<lambda>_ _. True"
-
+text \<open> Abbreviations for trivial postconditions (taking three arguments): \<close>
 abbreviation(input)
   toptoptop :: "'a \<Rightarrow> 'b \<Rightarrow> 'b \<Rightarrow> bool" ("\<top>\<top>\<top>")
 where
   "\<top>\<top>\<top> \<equiv> \<lambda>_ _ _. True"
```
Xaphiosis, line 948: silly consistency question: why can we get rid of toptop, but not toptoptop?

lsf37: All monad Hoare logics introduced toptop, but only the TraceMonad uses toptoptop, and I didn't want to introduce something into the wider context that we should instead get rid of. 

I haven't figured out yet how to get rid of toptop (and therefore toptoptop) without too much impact on how wp works, that's why it still exists at all. But removing it should be the next step in this particular quest.

lib/Fun_Pred_Syntax.thy
```diff
@@ -0,0 +1,187 @@
...
+lemma pred_top_comp[simp]:
+  "\<top> \<circ> f = \<top>"
+  by (simp add: comp_def)
+
+lemma pred_bot_comp[simp]:
+  "\<bottom> \<circ> f = \<bottom>"
+  by (simp add: comp_def)
+
+(* We would get these for free if we could instantiate pred_top/pred_bot to top/bot directly: *)
+lemmas pred_top_left_neutral[simp] =
```
Xaphiosis, line 120: trivial: recommend newline above this line

Xaphiosis: actually, looking further above, it seems like I'd recommend it for most of the comments that outline sections of lemmas... have you considered `text` or actual sections of some kind?

lib/Fun_Pred_Syntax.thy
```diff
@@ -0,0 +1,187 @@
...
+abbreviation pred_conj :: "('a \<Rightarrow> 'b::boolean_algebra) \<Rightarrow> ('a \<Rightarrow> 'b) \<Rightarrow> ('a \<Rightarrow> 'b)" where
+  "pred_conj \<equiv> inf"
+
+abbreviation pred_disj :: "('a \<Rightarrow> 'b::boolean_algebra) \<Rightarrow> ('a \<Rightarrow> 'b) \<Rightarrow> ('a \<Rightarrow> 'b)" where
+  "pred_disj \<equiv> sup"
+
+abbreviation pred_neg :: "('a \<Rightarrow> 'b::boolean_algebra) \<Rightarrow> ('a \<Rightarrow> 'b)" where
+  "pred_neg \<equiv> uminus"
+
+(* Lifted True/False: ideally, we'd map these to top/bot, but top/bat are constants and there are
```
Xaphiosis, line 29: top/bat -> top/bot

lib/Fun_Pred_Syntax.thy
lsf37: actual (sub)sections might make sense, yes, I'll have a look at that

lib/Fun_Pred_Syntax.thy
```diff
@@ -0,0 +1,187 @@
...
+theory Fun_Pred_Syntax
+imports Main
+begin
+
+(* Functions are already instances of Boolean algebras and provide all the standard laws one
+   would like to have. Default simplifications are automatic. Search for inf/sup/uminus to
+   find further laws and/or unfold via the definitions below.
+
+   The abbreviations here provide special syntax for the function instance of Boolean
+   algebras only, leaving other instances (such as "bool") untouched. *)
```
Xaphiosis, line 18: I'll be honest, I don't really follow what is happening here and the comment doesn't help me that much either, likely due to `boolean_algebra` being a new thing for me.
Reading the header of this theory, "Syntax for using functions as predicates", it would imply that any functions can be used as predicates?
You're excluding "bool". Is that to avoid being able to say "True and True" thus duplicating the normal logical operators?

Xaphiosis (approved):
As we discussed, the commit message "lib+proofs+sys-init+tools: proof updates for Fun_Pred" should be Fun_Pred_Syntax

I'd recommend tweaking the comments that delineate sections to not be directly above a single lemma, as it looks like that comment only applies to that lemma.

Other than that, this is a very interesting generalisation!

lib/Fun_Pred_Syntax.thy
lsf37: > You're excluding "bool". Is that to avoid being able to say "True and True" thus duplicating the normal logical operators?

Yes, precisely, I don't want people to start using these for `True and True` for bool, just because it's theoretically possible.

> Reading the header of this theory, "Syntax for using functions as predicates", it would imply that any functions can be used as predicates?

Any predicate-like function type, yes. I.e. `'s => bool` as before or any `'a => 'b => 'c => 'd => .. => bool` (where and of the type variables can themselves be functions or anything else) or any `'x => 'y` where `'y` is a boolean algebra. As long as it ends in bool or a in a boolean algebra, it'll work.

Basically, `'s => bool` is what we'd usually call predicates (functions from something to bool), and this generalises it to any number of arguments where the arguments can be anything and the result doesn't have to be strictly bool, but is bool-like enough to satisfy the laws of booleans (= boolean_algebra).

Xaphiosis: Would it make any sense to update the header to reflect that a bit better, if this theory provides syntax for predicate-like functions with an arbitrary number of arguments? I'm ok either way, but speaking from a perspective of having been confused after reading it.

lsf37: Yup, can add a bit more to the header.

Xaphiosis: Thumbs up.

Issue #543, opened 2023-01-09
Reduce Monad lib dependencies


lsf37:
To factor AutoCorres out from l4v, the monad libraries will need to move with them. They currently depend on most of l4v/lib, which should not be necessary.

This issue is for reducing these dependencies.

PR #554, opened 2023-01-12
Reduce Monad library dependencies

lsf37:
This PR is not quite the full story yet, but makes some significant inroads into issue #543.

Contained in this PR:

- introduces Monad_Lib, which contains the minimal requirements for the monad definitions themselves
- changes monad defs to use that
- splits up the gigantic NonDetMonadLemmaBucket into multiple theories
- small proof adjustments (mostly theory import changes) for the changes above
- replace `non_fail` names with `no_fail` names. We had a few of the former from before the naming scheme took hold.

I've tried to not clean up the lemmas while moving them. I think I only failed on that for `Injection_Handler.thy`, but even there there are only style changes, no content.

That means, proof content before and after this PR should be exactly the same. The proof adjustments where necessary to get lemmas into scope that would now be out of scope with the changed dependencies and any tweaks for things that are in scope a bit earlier than before (like `K_def[simp]`).

There is more work to do after this, for instance splitting up NonDetMonadVCG into Hoare logic, EmptyFail, NoFail, and NoThrow, which now already exist as separate theories, but still build on NonDetMonadVCG. But that is for another PR.


lsf37:
For reviewing this, there's probably no need to look at the proofs themselves -- they just got moved around. Maybe check if the headers/explanations in the new files make any sense (and exist), and if you are feeling adventurous if each file contains the right kind of lemmas, i.e. that I moved things to the right places.

lsf37:
- [ ] I should probably rename NoThrow to No_Throw and NoFail to No_Fail while we still have a chance.

Xaphiosis:
Typo in commit: "Some of the no_fail lemmas had he wrong name."
Thumbs up to rename of NoThrow/NoFail.
Good job on remembering it's 2023. Will look deeper now.

lib/CutMon.thy
```diff
@@ -0,0 +1,88 @@
...
+
+lemma cutMon_validE_drop:
+  "\<lbrace>P\<rbrace> f \<lbrace>Q\<rbrace>,\<lbrace>E\<rbrace> \<Longrightarrow> \<lbrace>P\<rbrace> cutMon R f \<lbrace>Q\<rbrace>,\<lbrace>E\<rbrace>"
+  by (simp add: validE_def cutMon_valid_drop)
+
+lemma snd_cutMon:
+  "snd (cutMon P f s) = (P s \<longrightarrow> snd (f s))"
+  by (simp add: cutMon_def fail_def split: if_split)
+
+
```
Xaphiosis, line 53: extra empty lines: on purpose?

lib/CutMon.thy
```diff
@@ -0,0 +1,88 @@
...
+     cutMon_walk_if[THEN arg_cong[where f=empty_fail], THEN iffD2,
+                    OF empty_fail_If]
+
+lemma cutMon_assert_opt:
+  "cutMon P (gets_the f >>= g)
+      = gets_the (\<lambda>s. if P s then f s else None) >>= g"
+  by (simp add: cutMon_def gets_the_def exec_gets
+                bind_assoc fun_eq_iff assert_opt_def
+         split: if_split)
+
```
Xaphiosis, line 86: extra empty lines: on purpose?

lib/Lib.thy
```diff
@@ -46,25 +47,6 @@ lemma Collect_eq:
 (* FIXME: move next to HOL.iff_allI *)
 lemma iff_impI: "\<lbrakk>P \<Longrightarrow> Q = R\<rbrakk> \<Longrightarrow> (P \<longrightarrow> Q) = (P \<longrightarrow> R)" by blast
 
-(* Long ago, I, fun_app, the verification master of darkness, unleashed an unspeakable evil
-upon the world. But a foolish proof engineer wielding an input abbreviation stepped forth
-to oppose me. Before the final blow was struck, I tore open a hole in a number of refinement
-proofs, and flung him into a broken proof state, where my evil is law. *)
```
Xaphiosis, line 52: hilarious. who wrote that?

lib/MonadEq_Lemmas.thy
```diff
@@ -0,0 +1,278 @@
+(*
+ * Copyright 2023, Proofcraft Pty Ltd
+ * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ *)
+
+(* Lemmas that support the monad_eq method. Either things that go into the [monad_eq] set
```
Xaphiosis, line 8: potential idea for factoring a method out to its own file: say what it does?

lib/MonadEq_Lemmas.thy
```diff
@@ -0,0 +1,278 @@
...
+   or that are needed for such lemmas. Usually about "snd (f s)" or "x \<in> fst (f s)" where f
+   is a nondet monad.
+
+   If you are planning to use the monad_eq method, this is the theory you should import. *)
+
+theory MonadEq_Lemmas
+  imports MonadEq
+begin
+
+lemma snd_return [monad_eq]:
```
Xaphiosis, line 18: is it worth it to tweak the style in this theory to remove the space before `[monad_eq]`? should be a simple regexp

lib/MonadEq_Lemmas.thy
```diff
@@ -0,0 +1,278 @@
...
+lemma not_snd_bindI2:
+  "\<lbrakk> \<not> snd ((a >>= b) s); (rv, s') \<in> fst (a s) \<rbrakk> \<Longrightarrow> \<not> snd (b rv s')"
+  by (fastforce simp: bind_def)
+
+lemma in_returns [monad_eq]:
+  "(r, s) \<in> fst (return r s)"
+  "(Inr r, s) \<in> fst (returnOk r s)"
+  by (simp add: in_monad)+
+
+
```
Xaphiosis, line 56: extra empty line?

lib/MonadEq_Lemmas.thy
```diff
@@ -0,0 +1,278 @@
...
+  apply (clarsimp simp: fail_def return_def assert_def)
+  done
+
+lemma not_snd_assert :
+  "(\<not> snd (assert P s)) = P"
+  by (metis snd_assert)
+
+lemma snd_assert_opt [monad_eq]:
+  "snd (assert_opt f s) = (f = None)"
+    by (monad_eq simp: assert_opt_def split: option.splits)
```
Xaphiosis, line 99: indent

lib/Monad_Commute.thy
```diff
@@ -0,0 +1,195 @@
...
+    Monad_Equations
+    Monad_Lists (* for mapM_x *)
+begin
+
+
+definition monad_commute where
+  "monad_commute P a b \<equiv>
+   \<forall>s. (P s \<longrightarrow> ((do x\<leftarrow>a;y\<leftarrow>b; return (x, y) od) s) = ((do y\<leftarrow>b;x\<leftarrow>a; return (x, y) od) s))"
+
+
```
Xaphiosis, line 21: extra newline?

lib/Monad_Commute.thy
```diff
@@ -0,0 +1,195 @@
...
+lemma monad_eq:
+  "a s = b s \<Longrightarrow>  (a >>= g) s = (b >>= g) s"
+  by (auto simp:bind_def)
+
+lemma monad_commute_simple:
+  "\<lbrakk> monad_commute P a b; P s \<rbrakk> \<Longrightarrow> (do x\<leftarrow>a;y\<leftarrow>b; g x y od) s = (do y\<leftarrow>b;x\<leftarrow>a; g x y od) s"
+  apply (clarsimp simp:monad_commute_def)
```
Xaphiosis, line 28: if we care, this file has a lot of `[a-z]:[a-z]` that could use an extra space after `:`

lib/Monad_Commute.thy
```diff
@@ -0,0 +1,195 @@
+(*
+ * Copyright 2023, Proofcraft Pty Ltd
+ * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ *)
+
+(* The monad_commute predicate + lemmas. *)
```
Xaphiosis, line 8: no idea what monad_commute is for ... again if we factor it out, should we say anything?

lib/Monad_Lib.thy
```diff
@@ -0,0 +1,64 @@
+(*
+ * Copyright 2023, Proofcraft Pty Ltd
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ *)
+
+theory Monad_Lib
```
Xaphiosis, line 7: what's the guideline for what goes here?

lib/Monad_WP/WhileLoopRules.thy
```diff
@@ -562,7 +562,7 @@ lemma whileLoop_terminates_liftE:
    apply clarsimp
   apply (subgoal_tac "case (Inr r) of Inl _ \<Rightarrow> True | Inr r \<Rightarrow>
       whileLoop_terminates (\<lambda>r s. (\<lambda>r s. case r of Inl _ \<Rightarrow> False | Inr v \<Rightarrow> C v s) (Inr r) s)
-           (\<lambda>r. (lift (\<lambda>r. liftE (B r)) (Inr r)) >>= (\<lambda>x. return (theRight x))) r s")
+           (\<lambda>r. (lift (\<lambda>r. liftE (B r)) (Inr r)) >>= (\<lambda>x. return (projr x))) r s")
```
Xaphiosis, line 565: confirm: should `theRight` changes be in this PR?

tools/asmrefine/GraphProof.thy
```diff
@@ -2180,7 +2180,6 @@ lemma stepwise_graph_refine_Basic:
   apply (frule_tac i=i and tr=tr in exec_trace_step_cases)
   apply (frule_tac i=j and tr=tr' in exec_trace_step_cases)
   apply (clarsimp simp: exec_graph_step_def exec_graph_invariant_Cons split: graph_function.split_asm)
-  apply (simp add: K_def)
```
Xaphiosis, line 2183: does this indicate that the K_def [simp] declaration now leaks everywhere? I saw you put a FIXME to get rid of that declaration... what did you have in mind?

lib/Oblivious.thy
```diff
@@ -0,0 +1,151 @@
...
+
+lemma select_f_oblivious [simp]:
+  "oblivious f (select_f v)"
+  by (simp add: oblivious_def select_f_def)
+
+lemma oblivious_select:
+  "oblivious f (select S)"
+  by (simp add: oblivious_def select_def)
+
+
```
Xaphiosis, line 150: extra newline?

lib/NoThrow.thy
```diff
@@ -0,0 +1,163 @@
...
+   Includes some monad equations that have no_throw as a main assumption.  *)
+
+theory NoThrow
+  imports
+    WhileLoopRules
+    MonadEq_Lemmas
+begin
+
+(* Alternative definition of no_throw; easier to work with than unfolding validE. *)
+lemma no_throw_def': "no_throw P A = (\<forall>s. P s \<longrightarrow> (\<forall>(r, t) \<in> fst (A s). (\<exists>x. r = Inr x)))"
```
Xaphiosis, line 18: what are your thoughts on having def' lemmas near their actual definitions so that people know that the def' versions exist?

lib/NoThrow.thy
```diff
@@ -0,0 +1,163 @@
...
+  apply atomize
+  apply (clarsimp simp: no_throw_def' handleE'_def bind_def)
+  apply (clarsimp simp: validE_def valid_def)
+  apply (erule allE, erule (1) impE)
+  apply (drule (1) bspec)
+  apply (clarsimp simp: return_def split: sum.splits)
+  apply force
+  done
+
+lemma no_throw_handle: "\<lbrakk> \<And>a. no_throw Y (B a); \<lbrace> X \<rbrace> A \<lbrace> \<lambda>_ _. True \<rbrace>,\<lbrace> \<lambda>_. Y \<rbrace> \<rbrakk> \<Longrightarrow> no_throw X (A <handle> B)"
```
Xaphiosis, line 76: this file doesn't have the greatest style, but I know it was copy/pasted... hmm

lib/Monad_WP/NonDetMonadVCG.thy
```diff
@@ -797,6 +800,18 @@ lemma empty_fail_return [simp, wp]:
   "empty_fail (return x)"
   by (simp add: empty_fail_def return_def)
 
+lemma empty_fail_bindE:
+  "\<lbrakk> empty_fail f; \<And>rv. empty_fail (g rv) \<rbrakk>
+       \<Longrightarrow> empty_fail (f >>=E g)"
```
Xaphiosis, line 805: indent

lib/Monad_Lists.thy
```diff
@@ -0,0 +1,673 @@
...
+lemma mapME_Cons:
+  "mapME m (x # xs) = (doE y \<leftarrow> m x; ys \<leftarrow> (mapME m xs); returnOk (y # ys) odE)"
+  by (simp add: mapME_def sequenceE_def Let_def)
+
+lemma mapME_Nil : "mapME f [] = returnOk []"
+  unfolding mapME_def by (simp add: sequenceE_def)
+
+lemmas mapME_simps = mapME_Nil mapME_Cons
+
+
```
Xaphiosis, line 25: extra newline?

lib/Monad_Lists.thy
```diff
@@ -0,0 +1,673 @@
...
+  unfolding zipWithM_x_def zipWith_def
+  by (subst zip_append [OF ls], simp, rule sequence_xappend1)
+
+lemma zipWithM_x_Cons:
+  assumes ls: "length xs = length ys"
+  shows "(zipWithM_x f (x # xs) (y # ys)) = (f x y >>=  (\<lambda>_. zipWithM_x f xs ys))"
+  unfolding zipWithM_x_def zipWith_def
+  by (simp, rule sequence_x_Cons)
+
+
```
Xaphiosis, line 109: hmm... if you're using double empty lines to delineate things, consider adding comment/text/section or the careful spacing might not be respected by future people

lib/Monad_Lists.thy
```diff
@@ -0,0 +1,673 @@
...
+      apply (wp Cons.hyps minvp)
+     apply simp
+    apply (fold validE_R_def)
+    apply simp
+    apply (wp invr est)
+    apply simp
+    done
+qed clarsimp
+
+
```
Xaphiosis, line 625: again, not clear on the double-empty

Xaphiosis (approved):
I had some comments and questions, but other than that looks good.

lib/MonadEq_Lemmas.thy
lsf37: The method already has its own file that describes what the method does. This new file just has the supporting lemmas for the NonDetMonad for this method. I think this is right as written.

lib/MonadEq_Lemmas.thy
lsf37: I've tried not to do that to keep the diff to actual moves. I failed with that plan for Injection_Handler, but for the rest it should have worked. We could do a new PR with a bunch of cleanup, though.

lib/Monad_Commute.thy
lsf37: You shouldn't punish people doing cleanup by demanding more cleanup :-)

I didn't write this predicate and I'm not sure where it's used, but I thought the name and definition are obvious enough: it says when two monads commute. Without tracking down where it's used, I'm not sure what else I would say about it.

It does look like yet another instance of the general oblivious /isoatedThreadAction etc idea, so eventually we should replace all these with just one concept, but this PR is just about dependencies and factoring things out.

lib/Monad_Lib.thy
lsf37: Good point, that should go into a comment. This theory should be the minimal set of constants and lemmas to support the monad definitions.

lib/Monad_WP/WhileLoopRules.thy
lsf37: They shouldn't, but they aren't :-)

`theRight` is in Lib, and these theories don't see `Lib` any more, so they need to use the primitive concept that people outside l4v would use. Once the other PR is merged, we still need `projr` here, but we'd see `theRight` in goals via the abbreviation declared in Lib.

tools/asmrefine/GraphProof.thy
lsf37: It does not fully leak everywhere, yes. It did leak almost everywhere before already, though (it was in most monad definition files, but not all), so I would say this is better than before, because it's at least consistent.

Removing it is a bigger operation and shouldn't be in here. It's not entirely clear yet if it's actually better to remove it or not, but I would hope that it is (depends on how much breaks).

lib/NoThrow.thy
lsf37: In general I'd agree, but here I would say that the definition of `no_throw` should eventually go into this file instead. 

If we do move it, it should be in a separate PR, I think.

lib/NoThrow.thy
lsf37: Yes, these are all just moves, I've tried hard not to touch style or other cleanup. We should do those separately in this case, otherwise this gets too big.

lib/Monad_WP/NonDetMonadVCG.thy
lsf37: Hm. I've explicitly tried not to do that, but the imports into this file are pretty small. On the other hand the entire file needs massive cleanup and should eventually be much smaller, so I'm not sure it makes sense to mix that into this one.

lsf37:
So for the double newlines in general: I'm pretty sure most of them were there before and were just moved over (I moved entire chunks/sections as far as possible), and I guess some just crept in. I think removing those is probably fine compared to the other style things that I didn't want to touch in this PR, because they won't clash with anything. I'll have a look at that.

lib/MonadEq_Lemmas.thy
lsf37: Have added a comment that points to the MonadEq theory

lib/MonadEq_Lemmas.thy
lsf37: Ok, I've changed my mind and applied it directly. All those spaces are gone now.

lib/Monad_Commute.thy
lsf37: This one actually makes sense, separating the definition from the rest.

lib/Monad_Commute.thy
lsf37: since I've been messing with the diff already for `[monad_eq]`, I think I'll apply a search&replace for this one as well.

lib/NoThrow.thy
lsf37: I caved in and cleaned up the worst of it..

lib/Monad_Lists.thy
lsf37: If there was sectioning here before it's pretty much gone by now, so I've removed all of these.

lib/Lib.thy
lsf37: Apparently that was Callum. Who would have thought..

PR #555, opened 2023-01-13
modernise sum type discriminators+selectors

lsf37:
Changes isLeft, isRight, theLeft, and theRight to abbreviations and updates the proofs accordingly.

Proof changes are small, the larger one in capDL-api is mostly fixing broken indentation of the original proof.

### DIFF
--- a/lib/LemmaBucket.thy
+++ b/lib/LemmaBucket.thy
@@ -100,10 +100,10 @@ lemmas mapM_x_accumulate_checks
 lemma isRight_rel_sum_comb2:
   "\<lbrakk> (f \<oplus> r) v v'; isRight v' \<rbrakk>
        \<Longrightarrow> isRight v \<and> r (theRight v) (theRight v')"
-  by (clarsimp simp: isRight_def)
+  by (cases v; clarsimp)
 
 lemma isRight_case_sum: "isRight x \<Longrightarrow> case_sum f g x = g (theRight x)"
-  by (clarsimp simp add: isRight_def)
+  by (cases x; clarsimp)
 
 lemma enumerate_append:"enumerate i (xs @ ys) = enumerate i xs @ enumerate (i + length xs) ys"
   apply (induct xs arbitrary:ys i)
--- a/proof/crefine/ARM/Fastpath_C.thy
+++ b/proof/crefine/ARM/Fastpath_C.thy
@@ -1811,9 +1811,10 @@ proof -
           apply (simp add: from_bool_0 if_1_0_0 cong: if_cong)
           apply (rule ccorres_cond_true_seq)
           apply (rule ccorres_split_throws)
-           apply (fold dc_def)[1]
            apply (rule ccorres_call_hSkip)
-             apply (rule slowpath_ccorres, simp+)
+             apply (erule disjE; simp flip: dc_def; rule slowpath_ccorres)
+            apply simp
+           apply simp
           apply (vcg exspec=slowpath_noreturn_spec)
          apply (rule ccorres_rhs_assoc)+
          apply csymbr+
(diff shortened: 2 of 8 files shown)


lsf37:
(This currently sits on top of #554, but is otherwise independent)

proof/crefine/ARM/Fastpath_C.thy
```diff
@@ -1811,9 +1811,10 @@ proof -
           apply (simp add: from_bool_0 if_1_0_0 cong: if_cong)
           apply (rule ccorres_cond_true_seq)
           apply (rule ccorres_split_throws)
-           apply (fold dc_def)[1]
            apply (rule ccorres_call_hSkip)
-             apply (rule slowpath_ccorres, simp+)
+             apply (erule disjE; simp flip: dc_def; rule slowpath_ccorres)
+            apply simp
+           apply simp
```
Xaphiosis, line 1817: what happened here? this looks like something isn't working as well as it used to

lib/LemmaBucket.thy
```diff
@@ -100,10 +100,10 @@ lemmas mapM_x_accumulate_checks
 lemma isRight_rel_sum_comb2:
   "\<lbrakk> (f \<oplus> r) v v'; isRight v' \<rbrakk>
        \<Longrightarrow> isRight v \<and> r (theRight v) (theRight v')"
-  by (clarsimp simp: isRight_def)
+  by (cases v; clarsimp)
```
Xaphiosis, line 103: wouldn't isRight_def still work here?

Xaphiosis:
It's interesting to see that unifying with upstream generalisations causes some interesting side effects.
I'm a bit puzzled by the complications in the fastpath proof.
Also, that comment about having to use isLeft instead of isRight due to the underlying `\not isl` might trip up someone one day.
Thoughts?

lib/LemmaBucket.thy
lsf37: I think now it might. I fixed up that one before I got the idea for `isRight_def`.

proof/crefine/ARM/Fastpath_C.thy
lsf37: I couldn't see what the state was before, but I suspect that the simplifier now does slightly more than before. It ended up in a disjunction `isRight .. \/ ~ isEndpoint ...`. Which is really `~ isLeft ... \/ ~ isEndpoint ..`, which probably used to be `~ (isLeft .. /\ isEndpoint)`. The simplifier can't make progress from the `.. \/ ..` point and needs an explicit `disjE`, whereas before it seems to have used the conjunction first and simplified some `IF` condition somewhere.

This proof is the only case where that happened and the fix wasn't too bad, so I didn't investigate further. It's possible that the goal was close to some simp depth limit or similar and is now over. Not clear.

lsf37:
> It's interesting to see that unifying with upstream generalisations causes some interesting side effects.

In this case, they're not really generalisations, just slightly different default simp rules from upstream, so a few side effects look reasonable to me.

> Also, that comment about having to use isLeft instead of isRight due to the underlying `\not isl` might trip up someone one day.
> Thoughts?

I agree that it might trip people up in some circumstances, but I don't have a good solution. Writing `~ isLeft x` everywhere instead of `isRight x` would definitely be too ugly for our proofs. If we define our own constant `isRight` instead of an abbreviation, we lose the simplification that we get from the default setup and deviate from the philosophy of what upstream is trying to do (eliminating `isr` by using only `isl`). So it's not ideal, but I think it's the maximum we can get if we want the upstream definition.

This was only an issue in the fastpath proof which is trying a bit much unification magic anyway (`?P (isRight ?x) (theRight ?x)` is not a higher-order pattern and has has ?x multiple times etc -- it seems to work fine where we want to use it, though, so it's not wrong or anything, just a bit ambitious.)

We could add another comment to the `isRight` abbreviation. I doubt people will look there, but it's the most prominent place I can think of for this issue.

lib/LemmaBucket.thy
lsf37: (It's the more stable proof, so I don't think we should change it back)

lsf37:
Have added a comment to the definition of `isRight` to warn for matching. Matching `isRight` will not always fail, it might well work in some circumstances, so that's why it just says "consider", not something stronger. It's just a more complex term than matching on `isLeft` (and `isLeft` is still guaranteed to be there after simplification, but `isRight` might not be -- I think that's what actually happened in the fastpath proof: a `~ isRight x` got simplified into `isLeft x` and the rule expected `~ isRight x`).

Xaphiosis (approved):

PR #556, opened 2023-01-20
separate session for ML_Utils

lsf37:
Rename `ml_helpers` to `ML_Utils` and make it a separate session, because it has no other Lib dependencies.

Part of #543.


lib/ML_Utils/ROOT
```diff
@@ -0,0 +1,18 @@
+(*
+ * Copyright 2023, Proofcraft Pty Ltd
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ *)
+
+chapter Lib
+
+session ML_Utils (lib) = Word_Lib +
```
lsf37, line 9: This should actually be just HOL.

lib/Apply_Trace.thy
```diff
@@ -8,8 +8,7 @@
  * Provides an alternate refinement function which takes an additional stateful journaling operation. *)
 theory Apply_Trace
 imports
-  Main
-  "MLUtils"
+  ML_Utils.MLUtils
```
Xaphiosis, line 11: Do we not want these to be named consistently? `ML_Utils` vs `MLUtils` on the same line is a bit blah

Xaphiosis: Other than that, looks good to me.

Xaphiosis (approved):

lib/Apply_Trace.thy
lsf37: Didn't want to introduce too much noise, but you do have a point. I'll add that rename.

PR #557, opened 2023-01-20
docs: proof style on using `by`

Xaphiosis:
This is my stab at writing down the result of the discussions of the recent past and the hopeful resulting consensus.

### DIFF
--- a/docs/Style.thy
+++ b/docs/Style.thy
@@ -298,6 +298,64 @@ text \<open>
   approach it with a breadth-first approach. Since the default isabelle strategy is depth-first,
   prefers (or defers) will be needed, e.g. corres proofs.\<close>
 
+subsection \<open>Using by\<close>
+
+text \<open>
+  When all subgoals of a proof can be solved in one apply statement, use `by`.\<close>
+
+lemma
+  "True"
+  by simp
+
+lemma
+  "X"
+  apply (subgoal_tac "True")
+   prefer 2
+   subgoal by blast
+  apply (subgoal_tac "True")
+   prefer 2
+   subgoal
+     by blast \<comment> \<open>for tactic invocations that would overflow the line\<close>
+  oops
+
+text \<open>
+  When all subgoals of a proof can be solved in two apply statements, use `by` to indicate the
+  intermediate state is not interesting.\<close>
+
+lemma
+  "True \<and> True"
+  by (rule conjI) auto
+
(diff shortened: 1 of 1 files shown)


lsf37 (approved):
My personal impatience would suggest using `by` already for 1 sec wait time, but it's rare that something is just between 1 and 2 sec, so it probably doesn't matter anyway.

docs/Style.thy
```diff
@@ -298,6 +298,64 @@ text \<open>
...
+
+lemma
+  "X"
+  apply (subgoal_tac "True")
+   prefer 2
+   subgoal by blast
+  apply (subgoal_tac "True")
+   prefer 2
+   subgoal
+     by blast \<comment> \<open>for longer tactic invocations\<close>
```
corlewis, line 318: It wasn't immediately obvious to me that longer here was referring to the textual length of the tactic (instead of how long it took to execute). How about something like the following?
```suggestion
     by blast \<comment> \<open>for tactic invocations that do not fit within a single line\<close>
```

corlewis (approved):

docs/Style.thy
Xaphiosis: point taken, but it's also for stuff that would overflow the line if we tried to put it all on the same line, you know like `have "fdlksfsd dlkfjsdlkf jskldj" by (simp add: long list here)` can still be wrapped as:
```
have "fdlksfsd dlkfjsdlkf jskldj"
  by (simp add: long list here)
```
thoughts on a phrase that would work better but cover both thoughts?

corlewis: Hmm, it's not completely clear to me what distinction you are making, but would something like for tactic invocations that would overflow the line work?

Xaphiosis: I am trying to distinguish when the whole assembly (have ... + by ...) won't fit on one line due to the tactic being too long textually, as opposed to the tactic itself not fitting on one line.
I think your latest version does meet that requirement, so I'll change it.

Xaphiosis: done. makes more sense now?

corlewis: I think so, but we should probably also update the similar comment in the following example for using by to replace two apply statements.

Xaphiosis: thanks! hopefully done that too now. if you see anything else, let me know

PR #558, opened 2023-01-20
Style guide: left- vs right-wrapping of operators

Xaphiosis:
This is a first stab at stating what's going on with the two styles, and could really use more than one round of feedback.
Please ignore the `by` commit, PR is in progress for that one.

The part about `Inner-syntax indentation is not automatable in the general case, so our goal is to help ...` isn't clear where it should go, help welcome.

### DIFF
--- a/docs/Style.thy
+++ b/docs/Style.thy
@@ -219,6 +219,138 @@ lemma test_lemma3:
          case_tac h; simp)
   done
 
+section \<open>Right vs left operator-wrapping\<close>
+
+text \<open>
+  When a term is too long, there is a general consensus to wrap it at operators. However, consensus
+  has never emerged on whether the operator should then end up at the end of the line (right
+  operator wrapping), or start of the next one (left operator wrapping).
+  Some people have a tendency towards right-wrapping, others towards left-wrapping. They
+  each have advantages in specific contexts, thus both appear in the l4v proofs and are permitted
+  style.\<close>
+
+term \<open>A \<and> B \<longrightarrow> C \<and> D\<close> \<comment> \<open>no wrapping when A..D are small terms\<close>
+
+term \<open>A \<and>
+      B \<longrightarrow>
+        C \<and>
+        D\<close> \<comment> \<open>right-wrapping when A..D are large terms\<close>
+
+term \<open>A
+      \<and> B
+      \<longrightarrow> C
+         \<and> D\<close> \<comment> \<open>left-wrapping when A..D are large terms\<close>
+
+text \<open>
+  While both styles are permitted, do not mix them in the same lemma. If a lemma already uses
+  one style and you aren't doing a major rewrite, stick to the existing style.\<close>
+
+lemma
+  shows "\<lbrakk> A; B; C\<rbrakk> \<Longrightarrow>
+         D" \<comment> \<open>right-wrapping OK\<close>
+  and "\<lbrakk> A; B; C\<rbrakk>
+       \<Longrightarrow> D" \<comment> \<open>left-wrapping OK\<close>
+  oops \<comment> \<open>mixing styles: NOT OK\<close>
+
+text \<open>
+  Some operators and syntax only have ONE style. As seen in other sections:
+  * the `|` in `case`  never appears on the right
+  * `;` is always on the right when wrapping lists of assumptions
+  * `shows .. and ... and` wraps with `and` on the left
+  * `|` in method invocations always goes on the left
+  * commas and (semi)colons, owing to our natural language origins, always end up on the right\<close>
+
+lemma
+  shows
+    "\<lbrakk> A
+     ; B \<rbrakk> \<comment> \<open>wrong: always on right\<close>
+     \<comment> \<open>ok: \<Longrightarrow> can be either left or right\<close>
+     \<Longrightarrow> C" and \<comment> \<open>wrong: `shows/and` only on left!\<close>
+    "D"
+  and "E" \<comment> \<open>ok: on left\<close>
+proof -
+  have "True \<and> True"
+    by (rule conjI,
+        blast,
+        blast) \<comment> \<open>ok\<close>
+  have "True \<and> True"
+    by (rule conjI
+        , blast
+        , blast) \<comment> \<open>NOT OK: commas go on right\<close>
+  have "True \<and> True"
+    by (rule conjI;
+        blast) \<comment> \<open>ok\<close>
+  have "True \<and> True"
+    by (rule conjI
+        ; blast) \<comment> \<open>NOT OK: semicolons go on right\<close>
+  have "True \<and> True"
+    by (rule conjI
+        | blast)+ \<comment> \<open>ok\<close>
+  have "True \<and> True"
+    by (rule conjI |
+        blast)+ \<comment> \<open>NOT OK: `|` goes on the left\<close>
+  oops
+
+text \<open>
+  The general principle of "nothing indented less than what it belongs to" is in effect for both
+  wrapping styles. Remember, the goal of the exercise is to make it as clear to read for others as
+  you can. Sometimes, scanning the left side of the screen to see the overall term can help,
+  while other times putting @{text \<Longrightarrow>} on the right will save space and prevent subsequent lines
+  from wrapping.\<close>
+
+text \<open>
+  Inner-syntax indentation is not automatable in the general case, so our goal is to help
+  ease of comprehension as much as possible, i.e.
+  @{term "A \<and> B \<or> C \<longrightarrow> D \<or> E \<and> F"} is bearable if A..F are short, but if they are large terms,
+  please avoid doing either of these:\<close>
+
+term "
+  A \<and>
+  B \<or>
+  C \<longrightarrow>
+  D \<or>
+  E \<and>
+  F" \<comment> \<open>avoid: requires building a parse tree in one's head\<close>
+
+term "
+  A
+  \<and> B
+  \<or> C
+  \<longrightarrow> D
+  \<or> E
+  \<and> F" \<comment> \<open>can be marginally easier to scan, but still avoid due to mental parsing difficulty\<close>
+
+text \<open>Instead, help out the reader like this:\<close>
+
+term "
+  A \<and>
+  B \<or>
+  C \<longrightarrow>
+    D \<or>
+      E \<and>
+      F"
+
+term "
+  A
+  \<and> B
+  \<or> C
+  \<longrightarrow> D
+     \<or> E
+       \<and> F"
+
+text \<open>AVOID indentation that misrepresents the parse tree and confuses the reader:\<close>
+
+term "
+  A
+  \<and> B
+    \<or> C" \<comment> \<open>NOT OK: implies this parses as @{text "A \<and> (B \<or> C)"}\<close>
+
+term "
+  A \<and>
+  B \<longrightarrow>
+    B \<or>
+  A" \<comment> \<open>NOT OK: implies this parses as @{text "((A \<and> B) \<longrightarrow> B) \<or> A"}\<close>
+
 section \<open>Other\<close>
 
 text \<open>


docs/Style.thy
```diff
@@ -219,6 +219,102 @@ lemma test_lemma3:
...
   done
 
+section \<open>Right vs left operator-wrapping\<close>
+
+text \<open>
+  When a term is too long, there is a general consensus to wrap it at operators. However, consensus
+  has never emerged on whether the operator should then end up at the end of the line (right
+  operator wrapping), or start of the next one (left operator wrapping).
+  Mathematicians have a tendency towards right-wrapping, programmers towards left-wrapping, they
+  each have advantages in specific contexts, thus both appear in the l4v proofs and are permitted
```
lsf37, line 229: Not sure it's maths vs programming, I've seen strong supporters of both styles in both communities.

docs/Style.thy
```diff
@@ -219,6 +219,102 @@ lemma test_lemma3:
...
+  style.\<close>
+
+term \<open>A \<and> B \<longrightarrow> C \<and> D\<close> \<comment> \<open>no wrapping\<close>
+
+term \<open>A \<and>
+      B \<longrightarrow>
+        C \<and>
+        D\<close> \<comment> \<open>right-wrapping\<close>
```
lsf37, line 237: Might be a bit confusing, because you'd only end up with that if all of `A`, `B`, `C`, `D` are very large. Otherwise, you'd try to keep multiple things on one line even if others need to line-break. (This is true in both styles, but it looks stranger in right-style to me)

docs/Style.thy
```diff
@@ -219,6 +219,102 @@ lemma test_lemma3:
...
+  E \<and>
+  F" \<comment> \<open>avoid: requires building a parse tree in one's head\<close>
+
+term "
+  A
+  \<and> B
+  \<or> C
+  \<longrightarrow> D
+  \<or> E
+  \<and> F" \<comment> \<open>can be marginally easier to scan, but still avoid due to mental parsing difficulty\<close>
```
lsf37, line 315: Should provide a "do" case in addition to the "do not" cases.

docs/Style.thy
lsf37: Maybe just adding a comment that this form would be used if all of them are large terms would avoid the confusion already.

docs/Style.thy
```diff
@@ -219,6 +219,102 @@ lemma test_lemma3:
...
+        C \<and>
+        D\<close> \<comment> \<open>right-wrapping\<close>
+
+term \<open>A
+      \<and> B
+      \<longrightarrow> C
+         \<and> D\<close> \<comment> \<open>left-wrapping\<close>
+
+text \<open>
+  While both styles are permitted, but do not mix them in the same lemma. If a lemma already uses
```
michaelmcinerney, line 245: Remove the "but" on this line

docs/Style.thy
```diff
@@ -219,6 +219,102 @@ lemma test_lemma3:
...
+  have "True \<and> True"
+    by (rule conjI
+        , blast
+        , blast) \<comment> \<open>NOT OK: commas go on right\<close>
+  have "True \<and> True"
+    by (rule conjI;
+        blast) \<comment> \<open>ok\<close>
+  have "True \<and> True"
+    by (rule conjI
+        ; blast) \<comment> \<open>also ok, can be either, but prefer ; on right if it's a simple method
```
michaelmcinerney, line 284: I was told that the semicolon must also go at the end of the line

michaelmcinerney (approved):

docs/Style.thy
lsf37: I certainly much prefer that, but thought I lost that argument :-). If there is a majority for `;` at the end of line, I'm all for it.

Xaphiosis: I guess it's down to what we want to permit. I examined the situation now that I'm awake again, and it looks like a bit over 100 uses of a leading `;` in `l4v`, so it looks like this one can be arbitrarily permitted or scrapped. What do people think?

In my looking around I also discovered that `|` is rare at end of lines, so maybe we should say that it should go on the left while we're at it?

Then a candidate proof would look something like:
```
apply (a, b,
       aa; bb;
       (c
        | d
        | e
        | f)+)
```
Is this acceptable? Consider variants of all of these methods being short/long. Also, are there any other proof operators I'm forgetting? Obviously `?` and `+` are unary, thus go at the end and don't wrap.

docs/Style.thy
Xaphiosis: Ah yes, with both "while" and "but" in there, it's redundant. Thanks

docs/Style.thy
Xaphiosis: made an attempt at these, less confident about what the right-wrap style would be for large terms in this case

docs/Style.thy
Xaphiosis: Tried to add some commentary here.

docs/Style.thy
```diff
@@ -219,6 +219,120 @@ lemma test_lemma3:
...
+  * commas and colons, owing to our natural language origins, always end up on the right\<close>
+
+lemma
+  shows
+    "\<lbrakk> A
+     ; B \<rbrakk> \<comment> \<open>wrong: always on right\<close>
+     \<comment> \<open>ok: can be either\<close>
+     \<Longrightarrow> C" and \<comment> \<open>wrong: `shows/and` only on left!\<close>
+    "D"
+  and "E" \<comment> \<open>ok: on left\<close>
```
corlewis, line 269: It's not clear to me what the middle `ok: can be either` is referring to.

docs/Style.thy
corlewis: It's probably mostly due to just being what I'm used to, but I'm happy with `|` going on the left and `;` being preferred on the right. I know that some people have made the argument that the two are used similarly and should have the same style, but I think I can invent enough of a rationalisation that I am happy with them being different. In particular, to me the first tactic is more important than the second with `;`, while with `|` they're of equivalent importance.

docs/Style.thy
```diff
@@ -219,6 +219,120 @@ lemma test_lemma3:
...
+    E \<and>
+    F"
+
+term "
+  A
+  \<and> B
+    \<or> C
+  \<longrightarrow> D
+     \<or> E
+       \<and> F"
```
corlewis, line 333: I'm struggling to give an alternative suggestion, but these two examples don't quite seem right to me. At the very least, to be consistent should the right-wrap style have more indentation around the nested conj's and disj's?

docs/Style.thy
lsf37: Happy with bland on that one :-)

docs/Style.thy
lsf37: I'd be fine with that, yes.

docs/Style.thy
lsf37: I think both could be permitted. Depends on how used you are to reading `/\` and `\/` as a list and how clear the operator binding is. If you want to emphasise you'd probably write

```
term "
   A /\
   B \/
   C -->
     D \/
       E /\
       F"
```

I don't think I'd write

```
term "
   A /\
   B \/
     C -->
       D \/
         E /\
         F"
```

because that would indicate `A /\ B \/ (C --> D \/ E /\ F)` instead of `(A /\ B \/ C) --> (D \/ E /\ F)`

I guess you could go with:

```
term "
   A /\
   B \/
     C
   -->
   D \/
     E /\
     F"
```
 if you really wanted to emphasise the precedence. But my normal preference would be to keep it flat unless it becomes unclear.

docs/Style.thy
Xaphiosis: changed to ```― ‹ok: `⟹` can be either left or right›```
does that work?

docs/Style.thy
```diff
@@ -219,6 +219,126 @@ lemma test_lemma3:
...
+       \<Longrightarrow> D" \<comment> \<open>left-wrapping OK\<close>
+  oops \<comment> \<open>mixing styles: NOT OK\<close>
+
+text \<open>
+  Some operators and syntax only have ONE style. As seen in other sections:
+  * the `|` in `case`  never appears on the right
+  * `;` is always on the right when wrapping lists of assumptions
+  * `shows .. and ... and` wraps with `and` on the left
+  * `|` in method invocations always goes on the left
+  * commas and (semi)colons, owing to our natural language origins, always end up on the right\<close>
```
Xaphiosis, line 261: Hey @mbrcknl , you're the only other "`;` can go on left" enthusiast besides me, and I'm getting convinced that the style should be semicolons on right only, and indeed `;` on left doesn't see much use in the repo. Do you have thoughts on standardising this?

docs/Style.thy
Xaphiosis: So, uh, what do I put down for the "do" example for right-wrap?
I think the last one least (where `-->` isn't indented, and neither is `D` after the implication). The difference is, I think, that if you're dumping the operators on the far right with a ragged edge, the extra indentation isn't really going to help for the leading `\/`.
I like Gerwin's first example the most here, but I didn't write it myself because it's hard to quantify why I would have written it that way.

For the left-wrap, it makes more sense to show the spine. I guess you could flatten it as well, but with long terms you'd probably indent it, but flatten it if it's on one line.

mbrcknl: The second example (operators on left) is also a bit troubling, as @corlewis notes:
```
A
/\ B
   \/ C
--> D
    \/ E
       /\ F
```
...because it suggests `(A /\ (B \/ C)) --> (D \/ (E /\ F))`, which gives the wrong precedence between `/\` and `\/` in the antecedent. The issue seems to occur when you have a higher precedence operator (`/\`) before a lower precedence operator (`\/`), regardless of whether you're putting operators on the left or right. We naturally want to indent *below* something, not above it.

So unless we want to indent the `A /\ B` above the `\/ C`, I suggest we make it acceptable to flatten the LHS as long as it's safe for the reader to associate to the left without paying much attention to the operators.

For operators-on-the-right, I think that means you get @lsf37's first example:
```
A /\
B \/
C -->
  D \/
    E /\
    F
```

For operators-on-the-left, I think it would be: 
```
A
/\ B
\/ C
--> D
    \/ E
       /\ F
```

I tried to make more precise rules for which lines should be indented, but it got tricky, so I think it's probably best to keep it imprecise, i.e. indent as a hint to the reader that they need to break left associativity.

Does that make any sense?

mbrcknl (approved):
This is good! I like how you've made the purpose of the style guidelines clear.

docs/Style.thy
Xaphiosis: Thanks Matt!
Yes, it makes sense, and as you both say I've made a booboo (in fact, did the thing I said not to do, which is mislead the reader). I thought about it for a bit, and how to phrase a rule that we'd want in this case... and I'm also not clear on what that rule should be. As Matt says, this might be a place to not make a rule.
I'm also happy with both the right and left examples that Matt gave to go in as examples, @corlewis @lsf37 are you two in agreement?

lsf37: Yes, agreement. A high-level guideline we could put into text form is to say that the layout should not mislead about operator precedence.

Xaphiosis:
OK, it's a bit late but I think I've addressed all of the comments. I've cleaned up the ambiguity in the indentation as @mbrcknl wrote, added a note to avoid misleading about misrepresenting the parse tree.
@lsf37 and @corlewis are you ok with this now? If people are happy, I'll squash+rebase+merge.

docs/Style.thy
```diff
@@ -219,6 +219,138 @@ lemma test_lemma3:
...
+
+lemma
+  shows
+    "\<lbrakk> A
+     ; B \<rbrakk> \<comment> \<open>wrong: always on right\<close>
+     \<comment> \<open>ok: \<Longrightarrow> can be either left or right\<close>
+     \<Longrightarrow> C" and \<comment> \<open>wrong: `shows/and` only on left!\<close>
```
corlewis, line 268: Very minor, but this comment having an ! implies to me that it is more important than the other comments.

corlewis (approved):
I'm very happy with this, it's a great addition to the style guide.

docs/Style.thy
lsf37: Happy with these now!

lsf37 (approved):

PR #559, opened 2023-01-24
More fine-grained lib sessions

lsf37:
No proof changes, just moving files around and updating ROOT files + theory imports.

- introduce `Eisbach_Tools` session: Eisbach_Methods, Apply_Debug, etc. 
- rename `Monads_WP` to `Monads`
- introduce `Monads` session: everything previously in Monads_WP plus some of the theories factored out from NonDetMonadLemmaBucket, but not all of them.

Not entirely sure on the naming of `Eisbach_Tools`. Could also be just `Tactic_Tools` or something like that, because the session also contains `Conjuncts`, `ProvePart`, `Simp_No_Conditional`, and `Trace_Schematic_Insts` which are not strictly speaking Eisbach, just mostly used in Eisbach.

The "real" distinction to the other random tactic and Eisbach theories we have in Lib is that these are the ones that are needed by `Monads` and `AutoCorres`.


Xaphiosis:
What does "Not at monad definition itself, more tool setup for WPFix." mean?

Xaphiosis (approved):

lsf37:
> What does "Not at monad definition itself, more tool setup for WPFix." mean?

`Datatype_Schematic` doesn't have anything to do with monads as such and is not needed to define monads, but it's needed to set up the WPFix tool. It doesn't seem to be needed anywhere else, so it's Ok for it to be in the Monads session, but it should be in the `wp` directory, not the root `Monads` directory.

Xaphiosis:
> > What does "Not at monad definition itself, more tool setup for WPFix." mean?
> 
> `Datatype_Schematic` doesn't have anything to do with monads as such and is not needed to define monads, but it's needed to set up the WPFix tool. It doesn't seem to be needed anywhere else, so it's Ok for it to be in the Monads session, but it should be in the `wp` directory, not the root `Monads` directory.

Yeah I think the language there is a bit impenetrable, but I guess we'll live with it.

PR #560, opened 2023-01-24
reduce AutoCorres dependencies

lsf37:
This doesn't fully implement #545 yet, but it's worth a check point.

- remove all dependencies from C parser to Lib session
- reduce dependencies of AutoCorres session to theories that can be moved out of Lib

The only Lib imports in AutoCorres are now:

- `CLib.LemmaBucket_C` which can be moved to `CParser` session
- `Lib.Monad_Equations` and `Lib.More_NonDetMonadVCG` which can be moved to `Monads` session

- `CLib.LemmaBucket_C` itself depends on `Lib.CLib` which has no further dependencies. Naming and destination of `CLib.thy` still unclear. Ideas welcome.

### DIFF
--- a/lib/clib/LemmaBucket_C.thy
+++ b/lib/clib/LemmaBucket_C.thy
@@ -6,13 +6,9 @@
 
 theory LemmaBucket_C
 imports
-  "Lib.Lib"
-  "Word_Lib.WordSetup"
+  Lib.CLib
   CParser.TypHeapLib
-  "CParser.ArrayAssertion"
 begin
-
-declare word_neq_0_conv [simp del]
 
 lemma Ptr_not_null_pointer_not_zero: "(Ptr p \<noteq> NULL)=(p\<noteq>0)"
  by simp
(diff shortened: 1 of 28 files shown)


lsf37:
This time there are some minor proof changes, mostly in AutoCorres itself, because it doesn't see the full `Lib` context any more.

Xaphiosis:
"lib+spec+tools: adjust theory imports for TypHeapLib" commit changes `proofs`

lsf37:
> "lib+spec+tools: adjust theory imports for TypHeapLib" commit changes `proofs`
> 
> For this kind of commit, I wonder how much point there's to adding the tags... not sure, but pointing out the missing bit

Hm, let me look at this, it might actually be sqash gone wrong.

lib/clib/LemmaBucket_C.thy
```diff
@@ -6,13 +6,13 @@
...
-  "Word_Lib.WordSetup"
+  Lib.CLib
   CParser.TypHeapLib
-  "CParser.ArrayAssertion"
 begin
 
-declare word_neq_0_conv [simp del]
+context
+notes word_neq_0_conv [simp del]
+begin
```
Xaphiosis, line 15: how many of the proofs actually need this? if only one or two, might've made sense to note/supply it there

Xaphiosis (approved):

lib/clib/LemmaBucket_C.thy
lsf37: I could have a look at that. I initially wanted to not add too much noise, but there might not actually be any.

lsf37: Done. Disturbance was indeed minimal.

PR #561, opened 2023-01-24
introduce Basics session

lsf37:
This session currently contains only one theory (CLib), which we want to include both in Lib and later independently in CParser/AutoCorres.

The second commit uses this to move LemmaBucket_C into the CParser session without introducing a Lib dependency.


Xaphiosis (approved):
I can see and understand this now, but am wondering if we should have more documentation for the future... like what Basics was intended to be. Is the commit message that started it sufficient, do you think?

lsf37:
> I can see and understand this now, but am wondering if we should have more documentation for the future... like what Basics was intended to be. Is the commit message that started it sufficient, do you think?

You're right and I was thinking the same thing. I'll add a TODO item for putting README files into these new directories.

PR #563, opened 2023-01-24
lib FIXME cleanup

lsf37:
Processing "FIXME lib" issues introduced in the last few PRs.

These are mostly lemma moving, lemma renaming, and removing duplicates.

Some attempts at name consolidation (`unless_wp` instead of `hoare_unless_wp`, to be in line with `return_wp` etc), but only extremely limited, because this is a very deep rabbit hole and too much renaming will probably conflict with the rt branch at some point.

### DIFF
--- a/lib/NonDetMonadLemmaBucket.thy
+++ b/lib/NonDetMonadLemmaBucket.thy
@@ -51,8 +51,6 @@ lemma lifted_if_collapse: (* FIXME lib: move to Lib *)
   "(if P then \<top> else f) = (\<lambda>s. \<not>P \<longrightarrow> f s)"
   by auto
 
-lemmas list_case_return2 = list_case_return (* FIXME lib: eliminate *)
-
 (* We use isLeft, because isLeft=isl is the primitive concept; isRight=\<not>isl matches on isl. *)
 lemma valid_isLeft_theRight_split:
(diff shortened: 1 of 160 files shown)


lib/NonDetMonadLemmaBucket.thy
```diff
@@ -42,15 +42,11 @@ lemma return_bindE:
   "isRight v \<Longrightarrow> return v >>=E f = f (theRight v)"
   by (cases v; clarsimp simp: return_returnOk)
 
-lemma list_case_return: (* FIXME lib: move to Lib *)
+lemma list_case_return: (* not in Lib, because "return" is not in scope there *)
```
Xaphiosis, line 45: sanity check: `Monads` is part of what's going to get factored out of `lib`, right? So NonDetMonadLemmaBucket, despite its name, shouldn't go there?

lsf37: Yes, Monads gets moved out, but NonDetMonadLemmaBucket stays in. It's mostly left over as convenience to not disturb the l4v proof context too much, but we could actually try to eliminate it completely. I'd vote for doing that later, though.

Xaphiosis (approved):
Looks good, nice to see cleanup and consolidation.

PR #566, opened 2023-01-31
finish AArch64 abstract invariants

lsf37:
Finished AInvs proof and switches off `quick_and_dirty` setting for proper proof checking.

Still plenty of FIXMEs open, which will be addressed in a separate PR.

### DIFF
--- a/proof/invariant-abstract/AARCH64/ArchCNodeInv_AI.thy
+++ b/proof/invariant-abstract/AARCH64/ArchCNodeInv_AI.thy
@@ -389,10 +389,10 @@ lemma swap_of_caps_valid_arch_caps [CNodeInv_AI_assms]:
      apply (rule conjI; clarsimp)
     apply (intro allI impI)
     apply (elim exE conjE)
-    sorry (* FIXME AARCH64
-    subgoal for _ _ _ _ p
+    subgoal for \<dots> p
       by (case_tac "p=b"; case_tac "p=a";
-            fastforce simp add: valid_table_caps_def empty_table_caps_of)
+          clarsimp simp add: valid_table_caps_def empty_table_caps_of;
+          fastforce)
    apply (simp add: unique_table_caps_def split del: if_split)
    apply (erule allfEI[where f="id (a := b, b := a)"])
    apply (erule allfEI[where f="id (a := b, b := a)"])
@@ -402,7 +402,7 @@ lemma swap_of_caps_valid_arch_caps [CNodeInv_AI_assms]:
   apply (erule allfEI[where f="id (a := b, b := a)"])
   apply (clarsimp split del: if_split split: if_split_asm dest!:vs_cap_ref_to_table_cap_ref
                       dest!: weak_derived_table_cap_ref)
-  done *)
+  done
 
 
 lemma cap_swap_asid_map[wp, CNodeInv_AI_assms]:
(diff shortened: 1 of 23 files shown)


lsf37:
(Theory linter complaint is fixed in #567)

michaelmcinerney (approved):
I think my understanding of the specifics of a lot of the vspace stuff is still pretty limited, but the proofs do look very good.

proof/invariant-abstract/AARCH64/ArchCNodeInv_AI.thy
```diff
@@ -389,10 +389,10 @@ lemma swap_of_caps_valid_arch_caps [CNodeInv_AI_assms]:
      apply (rule conjI; clarsimp)
     apply (intro allI impI)
     apply (elim exE conjE)
-    sorry (* FIXME AARCH64
-    subgoal for _ _ _ _ p
+    subgoal for \<dots> p
```
corlewis, line 392: Huh, I don't think that I've seen this before. Does that work backwards like `rename_tac` does?

proof/invariant-abstract/AARCH64/ArchKernelInit_AI.thy
```diff
@@ -58,45 +59,44 @@ lemma pptr_base_num:
   "pptr_base = 0x8000000000"
   by (simp add: pptr_base_def pptrBase_def canonical_bit_def)
 
-(* IRQ nodes occupy 11 bits of address space in this RISCV example state:
-   6 for irq number, 5 for cte_level_bits. *)
+(* IRQ nodes occupy 14 bits of address space in this AARCH64 example state:
+   9 for irq number, 5 for cte_level_bits. *)
```
corlewis, line 63: They're not used in that many places, but is it worth making constants for these numbers, to make it easier if we ever need to change them again?

spec/abstract/AARCH64/VCPUAcc_A.thy
```diff
@@ -292,6 +259,42 @@ definition vcpu_switch :: "obj_ref option \<Rightarrow> (unit,'z::state_ext) s_m
               od))
      od"
 
+
```
corlewis, line 262: The extra whitespace here might have been unintended.

proof/invariant-abstract/AARCH64/ArchDetSchedSchedule_AI.thy
```diff
@@ -497,18 +528,19 @@ lemma handle_hyp_fault_valid_sched[wp]:
   by (cases fault; wpsimp wp: handle_fault_valid_sched simp: valid_fault_def)
 
 lemma handle_reserved_irq_valid_sched:
-  "\<lbrace>valid_sched and invs and (\<lambda>s. irq \<in> non_kernel_IRQs \<longrightarrow>  scheduler_act_sane s \<and> ct_not_queued s)\<rbrace>
-  handle_reserved_irq irq \<lbrace>\<lambda>rv. valid_sched\<rbrace>"
-  unfolding handle_reserved_irq_def irq_vppi_event_index_def when_def
-  sorry (* FIXME AARCH64 double-when causing problems
-  unfolding handle_reserved_irq_def by (wpsimp simp: non_kernel_IRQs_def) *)
+  "\<lbrace>valid_sched and invs and (\<lambda>s. irq \<in> non_kernel_IRQs \<longrightarrow> scheduler_act_sane s \<and> ct_not_queued s)\<rbrace>
+    handle_reserved_irq irq \<lbrace>\<lambda>rv. valid_sched\<rbrace>"
```
corlewis, line 532: Indentation

corlewis (approved):

proof/invariant-abstract/AARCH64/ArchCNodeInv_AI.thy
lsf37: Yes! @Xaphiosis pointed me to that one.

spec/abstract/AARCH64/VCPUAcc_A.thy
lsf37: I don't quite remember, but I think this one is actually intended for a change. It start a new section in the file.

proof/invariant-abstract/AARCH64/ArchKernelInit_AI.thy
lsf37: AFAICS it is only this one proof in which they are unfolded. Actually, only the sum (14) is unfolded, the `irq` type and `cte_level_bits` already abstract the actual numbers. It's just the comment that is trying to explain what is going on that mentions all of them. Might make sense to avoid the numbers in the comment as well, though. @Xaphiosis what do you think?

Xaphiosis: It would be limited benefit, but we could define `irq_node_bits` to be `14` (via Value_Type) and add that comment above it, then not repeat `14` in this proof. Since we already have cte_level_bits, do we have a ready definition for the `5`? Is it `LENGTH(undefined::irq)`?
If we have defs for both of the pieces and can derive a number, then we can make a definition which will then update itself should any of the pieces change... which is probably never.
My feeling is, reluctantly, let's put in a definition, but leave it in ArchKernelInit_A.

lsf37: 5 is `cte_level_bits`. I don't think we need a value type, we only use the 14 in `mask` in this one lemma, so that means a constant definition right above this lemma would be enough (it shouldn't leak outside `ArchKernelInit_AI`, it's not used anywhere else. We'd need to fiddle with the use site to unfold the constant accordingly so the lemma fits. Should be doable if we want it.

Xaphiosis: Sorry I meant 9 not 5 for LENGTH(undefined::irq)

corlewis: For completeness and not as an argument about it being worth a constant, it looks like there's one use of the irq number in this lemma as well.

lsf37: Done. Slightly spookily, none of the other lemmas needed adjustments. Now I need to investigate why that is.

lsf37: Right, we don't need any of this, because the `irq_node` is the last object allocated in the init state, we don't need its size to prove non-overlap with other objects. More things to remove..

lsf37:
(rebased and fixed up review feedback)

PR #568, opened 2023-02-01
Stuff from `rt`

michaelmcinerney:
This adds definitions and lemmas that were added to the `rt` branch, including many things related to while loops at both Refine and CRefine levels

### DIFF
--- a/lib/ExtraCorres.thy
+++ b/lib/ExtraCorres.thy
@@ -238,4 +238,135 @@ lemma hoare_from_abs_inv:
   using assms
   by (fastforce intro: hoare_from_abs[where R=\<top> and S=\<top> and R'=\<top> and Q="\<lambda>_. P" , simplified])
 
+lemma in_whileLoop_corres:
+  assumes body_corres:
+    "\<And>r r'. rrel r r' \<Longrightarrow>
+             corres_underlying srel False nf' rrel (P and C r) (P' and C' r') (B r) (B' r')"
+  and body_inv: "\<And>r. \<lbrace>P and C r\<rbrace> B r \<lbrace>\<lambda>_. P\<rbrace>"
+                "\<And>r'. \<lbrace>P' and C' r'\<rbrace> B' r' \<lbrace>\<lambda>_. P'\<rbrace>"
+  and cond: "\<And>r r' s s'. \<lbrakk>rrel r r'; (s, s') \<in> srel; P s; P' s'\<rbrakk> \<Longrightarrow> C r s = C' r' s'"
+  and result: "(rv', t') \<in> fst (whileLoop C' B' r' s')"
+  shows "\<forall>s r. (s, s') \<in> srel \<and> rrel r r' \<and> P s \<and> P' s'
+                \<longrightarrow> (\<exists>rv t. (rv, t) \<in> fst (whileLoop C B r s) \<and> (t, t') \<in> srel \<and> rrel rv rv')"
+  apply (rule in_whileLoop_induct[OF result])
+   apply (force simp: cond whileLoop_def)
+  apply clarsimp
+  apply (frule (1) corres_underlyingD2[OF body_corres]; (fastforce simp: cond)?)
+  apply clarsimp
+  apply (frule use_valid[OF _ body_inv(1)])
+   apply (fastforce dest: cond)
+  apply (frule use_valid[OF _ body_inv(2)])
+   apply fastforce
+  apply (fastforce simp: whileLoop_def intro: whileLoop_results.intros(3) dest: cond)
+  done
+
+lemma corres_whileLoop:
+  assumes cond: "\<And>r r' s s'. \<lbrakk>rrel r r'; (s, s') \<in> srel; P s; P' s'\<rbrakk> \<Longrightarrow> C r s = C' r' s'"
+  and body_corres:
+    "\<And>r r'. rrel r r' \<Longrightarrow>
+             corres_underlying srel False nf' rrel (P and C r) (P' and C' r') (B r) (B' r')"
+  and body_inv: "\<And>r. \<lbrace>P and C r\<rbrace> B r \<lbrace>\<lambda>_. P\<rbrace>"
+                "\<And>r'. \<lbrace>P' and C' r'\<rbrace> B' r' \<lbrace>\<lambda>_. P'\<rbrace>"
+  and rel: "rrel r r'"
+  and nf': "\<And>r'. no_fail (P' and C' r') (B' r')"
+  and termin: "\<And>r' s'. \<lbrakk>P' s'; C' r' s'\<rbrakk> \<Longrightarrow> whileLoop_terminates C' B' r' s'"
+  shows "corres_underlying srel False nf' rrel P P' (whileLoop C B r) (whileLoop C' B' r')"
+  apply (rule corres_no_failI)
+   apply (simp add: no_fail_def)
+   apply (intro impI allI)
+   apply (erule_tac I="\<lambda>_ s. P' s"
+                    and R="{((r', s'), r, s). C' r s \<and> (r', s') \<in> fst (B' r s)
+                                              \<and> whileLoop_terminates C' B' r s}"
+                    in not_snd_whileLoop)
+    apply (clarsimp simp: validNF_def)
+    apply (rule conjI)
+     apply (intro hoare_vcg_conj_lift_pre_fix; wpsimp?)
+       using body_inv
+       apply (fastforce simp: valid_def)
+      apply (clarsimp simp: valid_def)
+     apply (insert termin)[1]
+     apply wpsimp
+    apply (fastforce intro: no_fail_pre nf')
+   apply (fastforce intro: wf_subset[OF whileLoop_terminates_wf[where C=C']])
+  apply clarsimp
+  apply (frule in_whileLoop_corres[OF body_corres body_inv]; (fastforce dest: cond)?)
+  apply (fastforce intro: assms)
+  done
+
+lemma whileLoop_terminates_cross:
+  assumes body_corres:
+    "\<And>r r'. rrel r r' \<Longrightarrow>
+             corres_underlying srel False nf' rrel (P and C r) (P' and C' r') (B r) (B' r')"
+  and cond: "\<And>r r' s s'. \<lbrakk>rrel r r'; (s, s') \<in> srel; P s; P' s'\<rbrakk> \<Longrightarrow> C r s = C' r' s'"
+  and body_inv: "\<And>r. \<lbrace>P and C r\<rbrace> B r \<lbrace>\<lambda>_. P\<rbrace>"
+                "\<And>r'. \<lbrace>P' and C' r'\<rbrace> B' r' \<lbrace>\<lambda>_. P'\<rbrace>"
+  and abs_termination: "\<And>r s. P s \<Longrightarrow> whileLoop_terminates C B r s"
+  and ex_abs: "ex_abs_underlying srel P s'"
+  and rrel: "rrel r r'"
+  and P': "P' s'"
+  shows "whileLoop_terminates C' B' r' s'"
+proof -
+  have helper: "\<And>s. P s \<Longrightarrow> \<forall>r' s'. rrel r r' \<and> (s, s') \<in> srel \<and> P s \<and> P' s'
+                                     \<longrightarrow> whileLoop_terminates C' B' r' s'"
+       (is "\<And>s. _ \<Longrightarrow> ?I r s")
+    apply (rule_tac P="?I" in whileLoop_terminates.induct)
+      apply (fastforce intro: abs_termination)
+     apply (fastforce simp: whileLoop_terminates.intros dest: cond)
+    apply (subst whileLoop_terminates.simps)
+    apply clarsimp
+    apply (frule (1) corres_underlyingD2[OF body_corres], fastforce+)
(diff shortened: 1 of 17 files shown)


michaelmcinerney:
I'd be happy to write a more comprehensive commit message for the first commit, and to remove the `rt` from the commit message of the last commit

michaelmcinerney:
The lemma `ccorresE_gets_the` in `Corres_UL_C.thy` requires `no_ofail`, which isn't visible at that point, as is. I'm not sure of the best approach to fix this, but it can be fixed by adding `Monads.OptionMonadWP` to the list of imports in that file

lib/ExtraCorres.thy
```diff
@@ -238,4 +238,224 @@ lemma hoare_from_abs_inv:
...
+                                        \<and> whileLoop_terminates C' B' r s}"
+               in not_snd_whileLoop)
+    apply (insert assms)[1]
+    apply blast
+   apply (rename_tac conc_r s)
+   apply (clarsimp simp: validNF_def)
+   apply (intro conjI)
+    apply (intro hoare_vcg_conj_lift_pre_fix; (solves wpsimp)?)
+      apply (prop_tac "\<exists>abs_r. rrel abs_r conc_r")
+       apply (insert assms)[1]
```
Xaphiosis, line 407: insert with [1] shouldn't need the [1] as insert does not affect other subgoals... I claim these should all have the [1] removed in this lemma and any other lemmas that use this pattern

lib/clib/Corres_UL_C.thy
```diff
@@ -56,6 +56,20 @@ lemma exec_handlers_use_hoare_nothrow:
...
 
+lemma exec_handlers_use_hoare_nothrow_hoarep:
+  assumes ce: "E \<turnstile>\<^sub>h \<langle>c # hs, s'\<rangle> \<Rightarrow> (n, t)"
+  and   asms: "s' \<in> R'"
+  and valid': "E \<turnstile> R' c Q'"
+  shows "E \<turnstile> \<langle>c, Normal s'\<rangle> \<Rightarrow> t \<and> isNormal t"
+  apply (insert assms)
+  apply (drule hoare_sound)
+  apply (clarsimp simp: cvalid_def HoarePartialDef.valid_def)
+  apply (erule exec_handlers.cases)
```
Xaphiosis, line 68: fastforce should be able to handle everything from here, possibly the erule as well when given as `elim!`

lib/clib/CCorresLemmas.thy
```diff
@@ -1036,4 +1033,122 @@ lemma ccorres_Guard_True_Seq:
...
+     subgoal by (erule ccorresE[OF body_ccorres])
+                (fastforce simp: no_fail_def nf[simplified no_fail_def] dest: EHOther)+
+    apply clarsimp
+    apply (prop_tac "G s'")
+     apply (fastforce dest: use_valid intro: body_inv)
+    apply (prop_tac "inter_t \<in> G'")
+     apply (fastforce dest: hoarep_exec[rotated] intro: body_inv)
+    apply (drule_tac x=s' in spec)
+    apply (elim impE)
+     apply (drule_tac s'=inter_t and r1="xf t'" in ccorresE_gets_the[OF cond_ccorres]; assumption?)
```
Xaphiosis, line 1122: good

lib/Monads/NonDetMonad.thy
```diff
@@ -722,6 +722,38 @@ where
...
+
+definition
+  whileM :: "('s, bool) nondet_monad \<Rightarrow> ('s, 'a) nondet_monad \<Rightarrow> ('s, unit) nondet_monad" where
+  "whileM C B \<equiv> do
+    c \<leftarrow> C;
+    whileLoop (\<lambda>c s. c) (\<lambda>_. do B; C od) c;
+    return ()
+  od"
+
+definition
```
Xaphiosis, line 736: this one could use a comment for future people to understand

Xaphiosis:
The proofs in `lib/ExtraCorres.thy` look a bit old and written by a less experienced person, could use a refresh. We already discussed the `insert` pattern, for instance.
I have a few comments here and there, but nothing serious. Glad to see more improvements to lib/clib.

Xaphiosis:
> I'd be happy to write a more comprehensive commit message for the first commit, and to remove the `rt` from the commit message of the last commit

Good idea, thumbs up to both.

lib/Monads/NonDetMonadVCG.thy
```diff
@@ -556,6 +556,17 @@ lemma hoare_case_option_wp:
...
   by (cases v) auto
 
+(* Might be useful for forward reasoning, when P is known. *)
+lemma hoare_when_cases:
+  "(\<And>s. \<not>B \<Longrightarrow> P s \<Longrightarrow> Q s)
+    \<Longrightarrow> (B \<Longrightarrow> \<lbrace> P \<rbrace> f \<lbrace> \<lambda>r. Q \<rbrace>)
+    \<Longrightarrow> \<lbrace> P \<rbrace> when B f \<lbrace> \<lambda>r. Q \<rbrace>"
+  by (cases B; simp add: valid_def return_def)
+
+lemma bind_dummy_ret_val:
```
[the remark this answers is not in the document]
michaelmcinerney: Yes

lib/Monads/NonDetMonad.thy
```diff
@@ -722,6 +722,38 @@ where
 notation (output)
   whileLoopE  ("(whileLoopE (_)//  (_))" [1000, 1000] 1000)
 
+
+section "More Monadic Combinators"
```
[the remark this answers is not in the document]
lsf37: I agree with the Misc/More/Extra concern (the name is probably my fault), but I do think they should go into their own section, because there is a foundational difference between these combinators and the ones we had before.

For all of these new ones, conditions can themselves be monads (which is why they are `ifM` etc). We have explicitly avoided that so far, because it is annoying to reason about constructs where tests can have side effects. In `MCS` we finally hit loops and conditions in C where there was no good way to avoid that any more, so I gave up on my resistance to adding them to the library, but I think it's worth pointing out that these are different to the other ones.

So the question is what is a good section title? Maybe `Combinators that have conditions with side effects`?

lib/ExtraCorres.thy
lsf37: `insert` does affect all subgoals

lib/Monads/NonDetMonad.thy
lsf37: I think I wrote that one. What comment would you like? It's `ifM`..

lsf37: (perfectly happy to add something, just not sure which bit is unclear, since I wrote it :-))

lib/Monads/NonDetMonadVCG.thy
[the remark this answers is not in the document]
lsf37: We should probably call this `bind_K_bind_eq` or `bind_eq_K_bind` or something like that. `K_bind` is the name for the bind without return value.

lib/ExtraCorres.thy
Xaphiosis: for real? that's crazy! I take that comment back then, but it's still a proof smell I'd rather not keep around :/

lib/Monads/NonDetMonad.thy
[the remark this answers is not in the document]
Xaphiosis: Ah, yes, that is what comments/titles should be for. Without deep examination I didn't figure out they should be a special class. Thumbs up.

lib/ExtraCorres.thy
michaelmcinerney: I'm sure those old proofs of mine could be substantially improved. But one other reason for restricting things with an `(insert blah)[1]` (apart from `fastforce` instantiation things incorrectly) is that with many assumptions, it really clogs the proof state. These low-level proofs are sometimes not that pleasant and it seems nice to have at least a few points in the proof where someone can put their cursor and see a proof state that makes *some* sense

lsf37: Yes, I'd usually go for `using assms` these days, but that doesn't work if it's followed by `drule`, because it wants to do rule chaining. No real nice option for this specific case that I can see.

lib/Monads/NonDetMonad.thy
Xaphiosis: I think the section title / comment for combinators with conditions that have side effects will take care of this, as it then becomes clear what's being done here.

lib/ExtraCorres.thy
Xaphiosis: In this particular proof, there's lots of `apply (insert assms)[1]` followed by a method that solves the subgoal. In most cases, you can either give assms to the method (simp/fastforce/etc) or use `using`. Maybe even something like `subgoal using assms by fastforce` or `subgoal using assms by - blast` ... I'm not seeing the case with the assms followed by drule in this proof at least.

lsf37:
https://github.com/seL4/l4v/pull/569 is now merged, so if you rebase over that the tests might work now.

michaelmcinerney:
Rebased over #569 and now testing. I'll make a few more changes, including updating the proofs in `ExtraCorres.thy`

michaelmcinerney:
I have made the new proofs in `ExtraCorres.thy` nicer. There are failures in `LibTest`, which I'll look at now

lib/Monads/NonDetMonadVCG.thy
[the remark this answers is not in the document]
michaelmcinerney: This is actual in `master` (in Refine) with the name `bind_dummy_ret_val`. I think it was just that `rt` moved it to `Lib`, and maybe it's also used in AInvs for `rt` now too. Should I do the renaming?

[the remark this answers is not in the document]
michaelmcinerney: I suppose I should at least remove the duplicates from Refine. But I just wanted to make sure that we're happy for this PR to touch things outside of lib and clib.

And happy to do the renaming still, but I quite like the current name: the lemma is only applicable when the return value is a dummy, and it's not claiming equality between `K_bind` and `bind`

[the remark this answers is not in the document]
michaelmcinerney: Any thoughts?

lib/ExtraCorres.thy
```diff
@@ -238,4 +238,135 @@ lemma hoare_from_abs_inv:
   using assms
   by (fastforce intro: hoare_from_abs[where R=\<top> and S=\<top> and R'=\<top> and Q="\<lambda>_. P" , simplified])
 
+lemma in_whileLoop_corres:
+  assumes body_corres:
+   "\<And>r r'. rrel r r' \<Longrightarrow>
```
Xaphiosis, line 243: nitpick: indent under `assumes` should be 2 spaces

lib/ExtraCorres.thy
```diff
@@ -238,4 +238,135 @@ lemma hoare_from_abs_inv:
...
+    apply (fastforce simp: rrel2)
+   apply clarsimp
+   apply (rule_tac P="\<lambda>s'. \<exists>s. (s, s') \<in> srel \<and> (P and C abs_r) s \<and> P' s' \<and> C' conc_r s'"
+                   in no_fail_pre)
+    apply (insert cond body_corres)
+    apply (fastforce intro: corres_u_nofail simp: pred_conj_def)
+   apply fastforce
+  apply (fastforce intro: wf_subset[OF whileLoop_terminates_wf[where C=C']])
+  done
+
```
Xaphiosis, line 371: I like these proofs much better now, thank you!

lib/Monad_Lists.thy
```diff
@@ -22,25 +22,31 @@ lemma mapME_Nil : "mapME f [] = returnOk []"
...
+    "\<And>m x xs y ys. zipWithM_x m (x # xs) (y # ys) = do m x y; zipWithM_x m xs ys od"
     by (simp add: zipWithM_x_def sequence_x_def zipWith_def)
-  have IH: "\<lbrace>P\<rbrace> zipWithM_x m as bs \<lbrace>\<lambda>rv. P\<rbrace>"
-    by fact
-  show ?case
-    by (simp add: zipWithM_x_Cons) (wp IH x)
+  have zipWithM_x_Nil:
+    "\<And>m xs. zipWithM_x m xs [] = return ()"
+    by (simp add: zipWithM_x_def sequence_x_def zipWith_def)
+   show ?case
```
Xaphiosis, line 40: `show` should align with `have`, not be under it

lib/Monads/NonDetMonadVCG.thy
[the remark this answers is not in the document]
Xaphiosis: Fine to remove the duplicates in this PR as a separate commit. No real opinion on the name, I'm surprised we need this lemma at all.

michaelmcinerney:
I think all comments, apart from the renaming of `bind_dummy_ret_val`, have now been addressed

michaelmcinerney:
And now with improvements to the commit messages, as promised

Xaphiosis:
> And now with improvements to the commit messages, as promised

Content-wise, I'm happy with it now.  Commit-wise as well after you squash stuff. The `Clean up corres rules for whileLoop` commit doesn't have a squash or a `lib:` prefix, so I'm not sure where it belongs, please don't forget about it.

Xaphiosis (approved):

lsf37 (approved):

PR #569, opened 2023-02-01
remove opt_mapE from global `[elim!]` set 

lsf37:
- removes `opt_mapE` from `[elim!]`, but leaves other reader option monad elims in the set. The reasoning is that `opt_mapE` is mostly used for projections that you want to keep stable in assumptions, but the others are used for option monads, which you usually want to break up.

- minimal proof fixes in RISCV64 and AARCH64

### DIFF
--- a/lib/Monads/OptionMonad.thy
+++ b/lib/Monads/OptionMonad.thy
@@ -416,7 +416,7 @@ lemma oreturnOkE:
   by simp
 
 lemmas omonadE [elim!] =
-  opt_mapE obindE oreturnE ofailE othrowE oreturnOkE oassertE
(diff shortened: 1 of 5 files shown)


lsf37:
Might need additional small fixes, have only test ARM, AARCH64 fully so far. RISCV64 CRefine still running, and hopefully the other sessions behave like ARM.

lsf37:
Ok, looks like everything has passed fine. The tag ins the second commit should be `riscv` not `riscv64`, I'll fix that up once reviews are in.

michaelmcinerney (approved):
That's great. Worked out really nicely

PR #570, opened 2023-02-02
Split up NonDetMonadVCG and clean up parts

lsf37:
This turned into a bit of a monster cleanup project.

The main objective was to resolve the FIXMEs to split up `NonDetMonadVCG` into parts that make sense independently (all the different validity predicates, removing no_fail and empty_fail from it, removing equations that should go into other theories etc).

The smaller theories then prompted a style and proof contraction cleanup. This probably produces a lot of diff, but is minor. 

However, during that `empty_fail` in particular got a larger face lift, merging in `EmptyFailLib` from the Lib session, moving in a complete set of `empty_fail` theorems for all basic monad operations, and making the attributes on those lemmas consistent. The policy is: everything terminal is `[simp, intro!, wp]`, everything conditional (set `empty_fail_cond`) is only `[intro!, wp]`. The latter is to avoid the simplifier digging into large terms only to then give up after 10 bind statements, because the maximum conditional simp depth is reached.

Since `empty_fail_bind` is such a conditional rule, some of the existing proofs now either need `simp add: empty_fail_cond` or need to be solved by `fastforce`, `wp`, or `safe`.

The rest of the changes are proof fixes for this. In the `NonDetMonadVCG` cleanup the order of `wp` rules changes slightly which means `assert_wp` is now applied a few more times, which lead to some occasional goals with more implications. Otherwise some proofs got a tiny bit shorter.

There are currently many small commits esp int he proof fixes, because I did these incrementally. Happy to squash all that or not, whichever people prefer.


proof/crefine/RISCV64/IpcCancel_C.thy
```diff
@@ -2499,6 +2501,8 @@ lemma scheduleTCB_ccorres_valid_queues'_pre_simple:
              when (\<not> runnable \<and> curThread = thread \<and> action = ResumeCurrentThread) rescheduleRequired
           od)
          (Call scheduleTCB_'proc)"
+  supply empty_fail_cond[simp]
+  supply empty_fail_cond[simp]
```
lsf37, line 2505: Remove double declaration

proof/invariant-abstract/ARM/ArchTcb_AI.thy
```diff
@@ -210,6 +209,16 @@ lemma as_user_ipc_tcb_cap_valid4[wp]:
...
   done
 
+(* Pull invs out of a complex goal and prove it only once. Use as (strengthen invs_strengthen)+,
+   best in combination with simp and potentially conj_cong *)
+lemma invs_strengthen:
+  "invs s \<Longrightarrow> P \<longrightarrow> invs s"
+  "invs s \<and> Q s \<Longrightarrow> P \<longrightarrow> invs s \<and> Q s"
+  "invs s \<and> Q s \<Longrightarrow> P \<longrightarrow> Q s \<and> invs s"
+  "invs s \<and> Q s \<Longrightarrow> P \<longrightarrow> (invs and Q) s"
+  "invs s \<and> Q s \<Longrightarrow> P \<longrightarrow> (Q and invs) s"
```
lsf37, line 219: I just realised that this does not only pull out `invs`, but also `Q`. Seems to be working just fine in the long-running monster tactic below (making it faster), but I was hoping that this might be more generally useful. I might have another look at this if it is still faster when `Q` is not pulled out.

lsf37: Doesn't make a noticeable time difference, so I've made it more general and moved it to `Invariants_AI`

lsf37:
In terms of reviewing this: the proof updates for the different architectures are very similar (and should all be boring), so it's probably fine to pick just one and only skim the others.

lib/Monads/Empty_Fail.thy
```diff
@@ -0,0 +1,158 @@
...
+ * SPDX-License-Identifier: BSD-2-Clause
+ *)
+
+theory Empty_Fail
+  imports
+    NonDetMonad
+    WPSimp
+begin
+
+section \<open>Monads that are wellformed wrt failure\<close>
```
Xaphiosis, line 14: nitpick: "w.r.t." is probably nicer in docs

lib/Monads/In_Monad.thy
```diff
@@ -0,0 +1,143 @@
...
+  by (simp add: alternative_def)
+
+(* FIXME lib: remove unlessE_whenE unless_when here and replace with in_unless lemmas *)
+lemmas in_monad = inl_whenE in_whenE in_liftE in_bind in_bindE_L
+                  in_bindE_R in_returnOk in_throwError in_fail
+                  in_assertE in_assert in_return in_assert_opt
+                  in_get in_gets in_put in_when unlessE_whenE
+                  unless_when in_modify gets_the_in_monad
+                  in_alternative
+
```
Xaphiosis: extra empty line

lib/Monads/In_Monad.thy
```diff
@@ -0,0 +1,143 @@
...
+lemma exec_put:
+  "(put s' >>= m) s = m () s'"
+  by (simp add: bind_def put_def)
+
+lemma bind_execI:
+  "\<lbrakk> (r'',s'') \<in> fst (f s); \<exists>x \<in> fst (g r'' s''). P x \<rbrakk> \<Longrightarrow>
+  \<exists>x \<in> fst ((f >>= g) s). P x"
+  by (force simp: in_bind split_def bind_def)
+
+
```
Xaphiosis: extra empty line

lib/Monads/No_Throw.thy
```diff
@@ -14,6 +14,18 @@ theory No_Throw
     MonadEq_Lemmas
 begin
 
+section "Basic exception reasoning"
+
+text \<open>
+  The following predicates @{text no_throw} and @{text no_return} allow
+  reasoning about functions in the exception monad that either never
+  throw an exception or never return normally.
+\<close>
```
Xaphiosis, line 22: wrap > to previous line?

lib/Monads/No_Fail.thy
```diff
@@ -8,9 +8,192 @@
...
+wpc_setup "\<lambda>m. \<lbrace>P\<rbrace> m \<lbrace>Q\<rbrace>!" wpc_helper_validNF
+
+
+bundle no_pre = hoare_pre [wp_pre del] no_fail_pre [wp_pre del]
+
+bundle classic_wp_pre = hoare_pre [wp_pre del]
+    all_classic_wp_combs[wp_comb del] all_classic_wp_combs[wp_comb]
+
+
+
```
Xaphiosis: dunno if we care or I just notice these, but triple line between sections, maybe extra newlines above

proof/capDL-api/TCB_DP.thy
```diff
@@ -1169,7 +1169,7 @@ lemma restart_cdl_current_domain:
   apply (simp add:restart_def)
   apply (wp alternative_wp)
     apply (simp add:cancel_ipc_def)
-    apply (wpsimp wp: hoare_pre_cont[where a="revoke_cap_simple sl" for sl])+
+    apply (wpsimp wp: hoare_pre_cont[where f="revoke_cap_simple sl" for sl])+
```
Xaphiosis, line 1172: did this sneak into the wrong commit? it is capDL-api, but there was a commit for hoare_pre_cont variable renamed I think. also again below

proof/invariant-abstract/ARM/ArchTcb_AI.thy
```diff
@@ -238,8 +237,9 @@ lemma tc_invs[Tcb_AI_asms]:
   apply (simp add: split_def set_mcpriority_def cong: option.case_cong)
   apply (rule hoare_vcg_precond_imp)
    apply wp
-      (* takes long: *)
-      apply ((simp only: simp_thms
+      (* takes long: (30 sec) *)
+      apply ((simp only: simp_thms cong: conj_cong
```
Xaphiosis, line 241: very old style, might be worth reindenting the `|`, but not if you're just blending in

proof/crefine/ARM/IsolatedThreadAction.thy
```diff
@@ -912,7 +912,7 @@ lemma setThreadState_no_sch_change:
   "\<lbrace>\<lambda>s. P (ksSchedulerAction s) \<and> (runnable' st \<or> t \<noteq> ksCurThread s)\<rbrace>
       setThreadState st t
    \<lbrace>\<lambda>rv s. P (ksSchedulerAction s)\<rbrace>"
-  (is "NonDetMonad.valid ?P ?f ?Q")
+  (is "NonDetMonadVCG.valid ?P ?f ?Q")
```
Xaphiosis, line 918: While I do understand, it is a bit sad we have to now type a longer prefix, because we often use find_theorems with `-NondetMonad.valid`.

proof/invariant-abstract/ARM_HYP/ArchTcb_AI.thy
```diff
@@ -241,7 +240,8 @@ lemma tc_invs[Tcb_AI_asms]:
   apply (rule hoare_vcg_precond_imp)
    apply wp
       (* takes long: *)
-      apply ((simp only: simp_thms
+      apply ((simp only: simp_thms cong: conj_cong
+        | (strengthen invs_strengthen)+
```
Xaphiosis, line 244: On ARM you tweaked the comment as well

proof/invariant-abstract/ARM/ArchTcb_AI.thy
Xaphiosis: actually, changed my mind, then you'd have to do the same on ARM_HYP, and it's probably not worth it

Xaphiosis:
OK, I've gone through all the commits now. My thoughts are first of all, excellent work on getting all of this done and cleaned up to this level. All I have really are nitpicks and bikeshedding:
* longer name prefix for `valid` is a little bit inconvenient
* I like the `text < ... >` format where the `>` is not on a line on its own, and you got most of them but not all. I pointed out one or two but stopped.
* I like the way section/subsection have two empty lines before it, and other stuff that goes together does not have extra empty lines. Again, you got a bunch of these, but there's another bunch that you didn't change. Again, I pointed out a few and then stopped. This and the previous point I mostly want for `lib`, as it will live a long time and be an example for others. My eyes can help hunt them down if you want.

Now for the commits. I would prefer to not have 40+ commits on this one:
* the `lib` ones are fine, I think, even if they're small they have an individual story
* ones like resolving the x64 fixme move are also good and should stay
* the ones where you do a themed pass over all of proofs and tag it `proofs:` look good
* it's the ones with unexciting updates for the monad tweaks can be squashed, at the very least we don't need a crefine commit for every arch, then again for refine, ainvs, ... access for arm and riscv, infoflow for arm and riscv, etc.
* at the extreme end all of the "fixes for monad refactor" commits could be squashed into a single one tagged "proof:" as their contents is fairly ordinary
* I have a preference for "proof updates" over "proof fixes" ... fixes I prefer to reserve for when we actually break something properly, but your call

I don't really have a lot to say about the new structure. Nothing jumps out as being wrong, in a bad place, or named poorly. Whether the import graph is going to have any problems, I guess we'll find out when/if people use it, but it's better than what we had before.

If it is maybe possible for the sanity of the victims for slightly smaller PRs to be hit by on a Fri afternoon, that would be cool, but you did explain what happened and this is probably the most sane outcome.

lib/Monads/NonDetMonadLemmas.thy
```diff
@@ -232,6 +285,7 @@ lemma whileLoopE_cond_fail:
   apply (auto intro: whileLoop_cond_fail)
   done
 
+
```
michaelmcinerney: nitpick: extra newline

lib/Monads/NonDetMonadVCG.thy
```diff
@@ -525,413 +351,390 @@ lemma hoare_impI:
...
+  case x of Inl a \<Rightarrow> f x a | Inr b \<Rightarrow> g x b
+  \<lbrace>R x\<rbrace>"
+  by (cases x; simp)
+
+lemmas hoare_vcg_precond_imp = hoare_weaken_pre (* FIXME lib: eliminate *)
 
-lemma select_wp: "\<lbrace>\<lambda>s. \<forall>x \<in> S. Q x s\<rbrace> select S \<lbrace>Q\<rbrace>"
-  by (simp add: select_def valid_def)
+lemmas hoare_seq_ext = seq_ext[rotated]
+lemmas hoare_vcg_seqE = seqE[rotated]
```
michaelmcinerney, line 689: I think we've discussed before how the names for some of the error monad names are a little different from the normal names. Is this something we want to look into now or soon?

lib/Monads/NonDetMonadVCG.thy
```diff
@@ -1476,123 +1135,106 @@ lemma hoare_drop_impE_R:
...
-  apply (simp add: lift_def)
-  done
+subsection \<open>Strongest postcondition rules\<close>
 
-(*
- * \<lbrace>?P\<rbrace> ?m1 -, \<lbrace>?E\<rbrace> \<Longrightarrow> \<lbrace>?P\<rbrace> liftME ?f1 ?m1 -, \<lbrace>?E\<rbrace>
- *)
-lemmas [wp_split] = validE_validE_E [OF liftME_wp, simplified, OF validE_E_validE]
+lemma get_sp:
+  "\<lbrace>P\<rbrace> get \<lbrace>\<lambda>a s. s = a \<and> P s\<rbrace>"
```
michaelmcinerney, line 1174: Are these strongest postcondition rules a little easier to read if the return value has name `rv` or something?

lib/Monads/No_Fail.thy
```diff
@@ -17,58 +17,15 @@ begin
 subsection "Non-Failure"
 
 text \<open>
-  With the failure flag, we can formulate non-failure separately
-  from validity. A monad @{text m} does not fail under precondition
-  @{text P}, if for no start state in that precondition it sets
-  the failure flag.
+  With the failure flag, we can formulate non-failure separately from validity.
+  A monad @{text m} does not fail under precondition @{text P}, if for no start
+  state in that precondition it sets the failure flag.
```
michaelmcinerney, line 22: This is splitting hairs, but is it slightly better to say "no start state satisfying the precondition", rather than "in that precondition", seeing that these preconditions aren't actually sets?

lib/Monads/NonDetMonad.thy
```diff
@@ -112,124 +102,92 @@ text \<open>
...
+definition select :: "'a set \<Rightarrow> ('s,'a) nondet_monad" where
   "select A \<equiv> \<lambda>s. (A \<times> {s}, False)"
 
-definition
-  alternative :: "('s,'a) nondet_monad \<Rightarrow> ('s,'a) nondet_monad \<Rightarrow>
-                  ('s,'a) nondet_monad"
-  (infixl "\<sqinter>" 20)
-where
+definition alternative ::
+  "('s,'a) nondet_monad \<Rightarrow> ('s,'a) nondet_monad \<Rightarrow> ('s,'a) nondet_monad" (infixl "\<sqinter>" 20) where
```
michaelmcinerney, line 110: More boring nitpicks, but do we want a space after the comma in tuples? It's a bit inconsistent

michaelmcinerney (approved):
Looks very nice! And very grateful for all the style improvements, too, and nicer proofs

lib/Monads/Empty_Fail.thy
```diff
@@ -0,0 +1,158 @@
...
+
+text \<open>
+  Usually, well-formed monads constructed from the primitives above will have the following
```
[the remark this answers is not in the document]
lsf37: Oh, well spotted! This used to be in a file where that sentence made more sense.

lsf37:
Rebased and squashed as discussed, before I go through the more detailed review feedback.

proof/capDL-api/TCB_DP.thy
lsf37: it probably did, yes. Will see if I can extract it.

lib/Monads/Det.thy
```diff
@@ -0,0 +1,70 @@
+(*
+ * Copyright 2023, Proofcraft Pty Ltd
+ * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ *)
+
+theory Det
```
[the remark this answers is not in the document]
lsf37: `det_wp` is a different calculus, so should probably stay in its separate file. There is a question wether it shouldn't be included in the Monad session, but for now I think we should keep it to the AutoCorres dependencies and then move things over on demand.

lib/Monads/NonDetMonadVCG.thy
lsf37: Not for this PR, but probably soon-ish, yes.

lib/Monads/NonDetMonadVCG.thy
lsf37: Yes, I should make that part of the style update.

lib/Monads/NonDetMonad.thy
lsf37: We do want a space

lsf37:
Ok, review feedback mostly done. I'm expecting failures, because I renamed a whole bunch of random return value names in Hoare rules to `rv`, and if people didn't use `rename_tac`, that is going to break proofs.

PR #572, opened 2023-02-07
Remove is_thread_control and thread_control_target

corlewis:


Xaphiosis (approved):

lsf37 (approved):
:+1:

proof/invariant-abstract/Tcb_AI.thy
```diff
@@ -1112,31 +1089,24 @@ lemma decode_set_space_inv[wp]:
...
-  "\<lbrace>\<lambda>s. P (obj_ref_of cap)\<rbrace> decode_set_space args cap slot extras \<lbrace>\<lambda>rv s. P (thread_control_target rv)\<rbrace>,-"
-  apply (rule hoare_pre)
-   apply (simp   add: decode_set_space_def whenE_def unlessE_def
-           split del: if_split)
-   apply (wp | simp only: thread_control_target.simps)+
-  done
-
-(* FIXME: move *)
-lemma boring_simp[simp]:
-  "(if x then True else False) = x" by simp
```
corlewis, line 1138: This lemma is unrelated to the rest of the PR but I thought that it was safe to remove. However, it turns out that it is used in one lemma in X64 CRefine.

What are people's thoughts on this lemma? If we want to keep it then I might revert it in this PR and leave moving and renaming it to future cleanup.

lsf37: It's worth keeping, but should be renamed and moved up. Ok to do some other time.

It's questionable if it is actually safe to be `[simp]`.

I have another PR coming up for `if_option_eq` that is similar and rewrites `((if P then None else Some x) = None) = P` and I thought that was safe, but it turns out that if `P` is itself an equation, it can lead to non-termination. So now I'm not so sure any more if it should be `[simp]`. The one here is probably fine, because it should be rare enough and looks like it comes out of terms generated by the C parser.

PR #573, opened 2023-02-08
`rt` merge

michaelmcinerney:


michaelmcinerney:
I was hoping for a nice way of squashing the commit that fixes the merge conflict into the "merge commit", or whatever it's called, but I couldn't figure out how to do that.

Apart from that, the only commit that should be of interest is the final commit in the branch. Because `opt_pred` is now a definition rather than an abbreviation, it needs to be unfolded in quite a few places.

proof/refine/ARM/IpcCancel_R.thy
```diff
@@ -3821,6 +3832,11 @@ lemma cancelBadgedSends_filterM_helper':
 lemmas cancelBadgedSends_filterM_helper
     = spec [where x=Nil, OF cancelBadgedSends_filterM_helper', simplified]
 
+lemma list_case_return:
+  "(case xs of [] \<Rightarrow> return v | y # ys \<Rightarrow> return (f y ys))
+    = return (case xs of [] \<Rightarrow> v | y # ys \<Rightarrow> f y ys)"
+  by (simp split: list.split)
+
```
lsf37, line 3839: This lemma should be available in Lib or Monads somewhere already

michaelmcinerney: I think this was in Lib before the big change to NonDetMonadLemmaBucket. I can't find it now.

There's also hoare_case_option_wp2 and distinct_tuple_helper which were originally in Lib and I think are now gone, but resurrected in this PR. Maybe I missed the Refine lemmas when I did the PR the other day. It would slow me down even more to add them to Lib on master first, but I could do that if you'd like

michaelmcinerney: I'm not sure how easy it would be to rebase the work I've done for a merge over a change to `master`. Sounds bad. But what I could do is merge this PR, make another small PR for Lib on `master`, cherry-pick that Lib PR to `rt`, and then just delete these lemmas in a separate PR

lsf37: `list_case_return` is in `lib/NonDetMonadLemmaBucket` now (because `Lib` doesn't include the monads any more). I think that should be available at that point in the proof, so you should be able to just delete it. If not, then something bad has happened in the merge, because it would mean we lost some lemmas.

`distinct_tuple_helper` is also in `lib/NonDetMonadLemmaBucket` (for the same reason).

`hoare_case_option_wp2` seems to be `rt` only at this point. It's Ok to leave that in just `rt` for now, we can sync up more a bit later.

michaelmcinerney: Oh I did mess up that file in the merge. What's the best way of correcting that? Should I redo the merge, with the fixing of the conflicts, and then cherry-pick the last big commit from this branch (deleting the lemmas from Refine that will then be in Lib) and then test?

lsf37: Do you have `git config --global rerere.enabled true` enabled? If yes, then it should remember previous conflict resolutions and redoing the merge is probably the easiest way.

If not, it might be easier to try to do this one file only by looking at both diffs manually. @Xaphiosis or @corlewis might have better tips on that.

michaelmcinerney: I think I must have that enabled, because I do remember it resolving previous conflicts automatically. I wonder whether I originally messed this file up and it then applied that resolution, or whether it was the second time around. But I'll try again and see if it's any better

lsf37: Oh, yes, that actually makes complete sense. If it remembers a previous mistake it'll just re-apply that.

lsf37: `git rerere forget FILE-WITH-BAD-MERGE` is probably what you need

michaelmcinerney: It miraculously didn't automatically apply a previous resolution for this file, so I think it should be fine this time around. I'll locally run AInvs and Refine and update the PR when they've passed

proof/refine/RISCV64/Syscall_R.thy
```diff
@@ -550,6 +550,11 @@ crunches setThreadState
 
 context begin interpretation Arch . (*FIXME: arch_split*)
 
+lemma hoare_case_option_wp2:
+  "\<lbrakk> \<lbrace>P\<rbrace> f None \<lbrace>Q\<rbrace>; \<And>x.  \<lbrace>P' x\<rbrace> f (Some x) \<lbrace>Q' x\<rbrace> \<rbrakk>
+   \<Longrightarrow> \<lbrace>case_option P P' v\<rbrace> f v \<lbrace>\<lambda>rv s. case v of None \<Rightarrow> Q rv s | Some x \<Rightarrow> Q' x rv s\<rbrace>"
+  by (cases v) auto
```
lsf37, line 556: Please add `FIXME RT: move`

lsf37 (approved):
Thanks for doing that, it looks like a lot of work!

One of the reasons why it looks like we need so many more `opt_pred_def` and `opt_mapE` in these proofs is because they tend to use very specific rules like `opt_map_red` and unfolding specific definitions, instead of using `in_omonad` or `monad_defs` which tend to absorb such changes. Not point in changing this now, but if you find yourself wanting to use these specific lemmas for unfolding, consider using the lemma sets instead.

For the location of the `case_wp2` lemma: adding a FIXME is fine. If the checks succeed in this current state, you could also add it to More_NonDetMonadVGC without re-running everything. (You'll still have to rebuild to get back to any current proof, though).

proof/refine/ARM/Syscall_R.thy
```diff
@@ -550,6 +550,11 @@ crunches setThreadState
 
 context begin interpretation Arch . (*FIXME: arch_split*)
 
+lemma hoare_case_option_wp2:
+  "\<lbrakk> \<lbrace>P\<rbrace> f None \<lbrace>Q\<rbrace>; \<And>x.  \<lbrace>P' x\<rbrace> f (Some x) \<lbrace>Q' x\<rbrace> \<rbrakk>
+   \<Longrightarrow> \<lbrace>case_option P P' v\<rbrace> f v \<lbrace>\<lambda>rv s. case v of None \<Rightarrow> Q rv s | Some x \<Rightarrow> Q' x rv s\<rbrace>"
+  by (cases v) auto
```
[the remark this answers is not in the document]
michaelmcinerney: I might move this up to `More_NonDetMonagVCG`. I could then cherry-pick that commit onto `master`, if that's best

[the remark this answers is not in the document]
michaelmcinerney: Or would it be better if it's in `NonDetMonadVCG`, just after `hoare_case_option_wp`? I don't mind rebuilding things. I'll need to go through everything again to merge the newer Lib work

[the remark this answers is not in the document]
lsf37: In that case, yes, it would fit well there.

PR #574, opened 2023-02-09
Word_Lib: enable "eval" for word quantifiers

lsf37:
Enable use of "eval" and "value" for formulas that quantify over word values. The code generator will exhaustively run all possible values.

For small word sizes, this works in very reasonable time. E.g. try

    lemma "∀(x::8 word) y. x + y = (x AND y) + (x OR y)"
      by eval

or

    value "∀(x::8 word) y z. y mod z = 0 ⟶ (x * y) div z = x * (y div z)"

(shows `False`)

Note that as usual for "eval" and "value" terms have to be close, i.e. you need to use object logic quantifiers.


Xaphiosis (approved):
nice

PR #577, opened 2023-02-10
lib: add some lemmas from MCS work to OptionMonadWP.thy

michaelmcinerney:


michaelmcinerney:
I'd be happy to separate out the style changes from the new rules, if that's preferred

lsf37 (approved):
Looks good. I have no strong feelings about separating out the style changes. I think for this one it's fine to keep them.

Xaphiosis (approved):
For anything bigger than this, I'd split off the style updates into a separate commit, but this one is so small that I agree with Gerwin.

lib/Monads/OptionMonadWP.thy
```diff
@@ -167,67 +188,114 @@ lemma owhile_NF [wp]:
...
+
+lemma no_ofail_owhen[wp]:
+  "(P \<Longrightarrow> no_ofail Q f) \<Longrightarrow> no_ofail (if P then Q else \<top>) (owhen P f)"
+  by (simp add: no_ofail_def owhen_def)
+
+lemma no_ofail_ounless [wp]:
+  "(\<not>P \<Longrightarrow> no_ofail Q f) \<Longrightarrow> no_ofail (if P then \<top> else Q) (ounless P f)"
+  by (simp add: no_ofail_def ounless_def)
+
+lemma no_ofail_oassert [simp, wp]:
```
Xaphiosis, line 280: if you are doing a style improvement run, the style improvement run should also extend to the lemmas you're pasting in, no? at least consistently removing the spaces before `[...]` annotations would be nice.

PR #578, opened 2023-02-13
More stuff from `rt`

michaelmcinerney:


michaelmcinerney:
I'm not quite sure I'm following the conventions for everything (for example, the lemmas for `empty_fail` and how things are added to the `wp` set). In particular, I'm not sure whether the lemma in the error monad is necessary or whether there's a better approach in the places where it's used for MCS

Xaphiosis (approved):

Xaphiosis:
> I'm not quite sure I'm following the conventions for everything (for example, the lemmas for `empty_fail` and how things are added to the `wp` set). In particular, I'm not sure whether the lemma in the error monad is necessary or whether there's a better approach in the places where it's used for MCS

Not sure what the MCS situation is, but your PR looks fine to me.

PR #580, opened 2023-02-15
Merge `master` into `rt` round two

michaelmcinerney:


proof/invariant-abstract/ARM/ArchEmptyFail_AI.thy
```diff
@@ -37,7 +37,7 @@ context Arch begin global_naming ARM
 crunch (empty_fail) empty_fail[wp, EmptyFail_AI_assms]: possible_switch_to,set_thread_state_act
   (simp: kernel_object.splits option.splits arch_cap.splits cap.splits endpoint.splits
          bool.splits list.splits thread_state.splits split_def catch_def sum.splits
-         Let_def wp: zipWithM_x_empty_fail)
+         Let_def wp: empty_fail_zipWithM_x)
```
lsf37, line 40: I suspect you can simply remove the lemma from `wp: ..`, because it should already be declared `[wp]`, but not worth another re-run.

proof/refine/ARM/Ipc_R.thy
```diff
@@ -5653,21 +5654,21 @@ lemma replyRemoveTCB_sym_heap_scReplies [wp]:
          apply (clarsimp simp: isHead_to_head)
          apply (intro conjI; intro allI impI)
           apply clarsimp
-          apply (subgoal_tac "replyPrevs_of s r = Some ya")
+          apply (subgoal_tac "replyPrevs_of s rv = Some ya")
```
lsf37, line 5657: Ideally there should be a `rename_tac` for the `rv` at the point where it is introduced, but more importantly also for the `ya` (you didn't introduce that one here, but these can really be very annoying in maintenance, so we should fix them wherever we see them).

proof/refine/ARM/Ipc_R.thy
```diff
@@ -5653,21 +5654,21 @@ lemma replyRemoveTCB_sym_heap_scReplies [wp]:
...
           apply (clarsimp simp: isHead_to_head)
-          apply (subgoal_tac "replyPrevs_of s r = Some rPtr")
+          apply (subgoal_tac "replyPrevs_of s rv = Some rPtr")
            apply (insert replyNexts_Some_replySCs_None)
            apply (simp add: sym_refs_replyNext_replyPrev_sym[symmetric])
            apply (clarsimp simp: obj_at'_real_def ko_wp_at'_def projectKOs opt_map_def)
           apply (clarsimp simp: obj_at'_real_def ko_wp_at'_def projectKOs opt_map_def)
          apply clarsimp
-         apply (subgoal_tac "replyPrevs_of s r = Some ya")
+         apply (subgoal_tac "replyPrevs_of s rv = Some ya")
```
lsf37, line 5671: Same `rename_tac` situation as above (potentially these could be the same rename_tac, hard to see in the diff)

proof/refine/RISCV64/Ipc_R.thy
```diff
@@ -5665,21 +5666,21 @@ lemma replyRemoveTCB_sym_heap_scReplies [wp]:
          apply (clarsimp simp: isHead_to_head)
          apply (intro conjI; intro allI impI)
           apply clarsimp
-          apply (subgoal_tac "replyPrevs_of s r = Some ya")
+          apply (subgoal_tac "replyPrevs_of s rv = Some ya")
```
lsf37, line 5669: add `rename_tac`

lsf37 (approved):
Looks good, minor quibbles about `rename_tac` only. 

`hoare_vcg_if_lift2` seems a lot more popular in the update than on master. Not really a problem, but I'm wondering where it's coming from. Is that because there are more implications showing up in the goal from the stronger default `assert_wp` rule?

michaelmcinerney:
> `hoare_vcg_if_lift2` seems a lot more popular in the update than on master. Not really a problem, but I'm wondering where it's coming from. Is that because there are more implications showing up in the goal from the stronger default `assert_wp` rule?

Maybe because we have more rules which have preconditions phrased in terms of `if ... then ... else ...`?

PR #583, opened 2023-02-22
Unescape verification-manifest.xml before saving it

wom-bat:
The current xmllint tools don't like %0A to mean newline so are crashing.


mbrcknl (approved):
Thanks for finding this!

PR #584, opened 2023-02-24
MCS: remove is_thread_control and thread_control_target 

corlewis:
Cherry-pick https://github.com/seL4/l4v/pull/572 into rt, updated as required.


lsf37 (approved):
Looking good!

PR #585, opened 2023-02-27
docs: add instructions for generating `Kernel_Config.thy`

michaelmcinerney:


README.md
```diff
@@ -164,8 +164,16 @@ statements.
 Running the Proofs
 ------------------
 
-If Isabelle is set up correctly, a full test for the proofs in this repository
-can be run with the command
+In order to open the AInvs session, the file `Kernel_Config.thy` must be generated.
+This can be done by first setting the environment variable `L4V_ARCH` to `RISCV64`
```
Xaphiosis, line 168: I don't get it. This is the "running the proofs" section. Can't `run_tests` handle building AInvs from a fresh repo checkout? What does `RISCV64` have to do with anything?
It seems you are trying to say that opening AInvs in Isabelle/JEdit won't work until `Kernel_Config.thy` is generated for the platform you're trying to load, and building ASpec is one way of triggering that generation.
These instructions aren't going to be very useful as-is for someone confused about why AInvs isn't working, and might be somewhat confusing if they aren't working on RISCV64.

Xaphiosis: @lsf37 what's the real trigger for `Kernel_Config.thy` by the way?

README.md
```diff
@@ -164,8 +164,16 @@ statements.
...
+and running
+
+    make ASpec
+
+from the directory `l4v/spec`.
+
+Once `Kernel_Config.thy` has been generated and Isabelle is set up correctly,
+a full test for the proofs in this repository can be run with the command
 
     ./run_tests
```
lsf37, line 178: Hm, this is not quite right. The full proofs can be run without doing any of that. Just a simple `./run_tests` is still fine, because it does include the `make` command, so I don't think we should change that paragraph.

Actually, reading the paragraphs below it now, it does already say that you need to use `make` for any sessions that require generated content. They just don't say which sessions these are or how to find out. So we could enumerate them. AFAIR, the sessions that directly need generated content are `ASpec`, `ExecSpec`, and `CKernel`, so if you want to build any session that depends on these, you need to use either `make` for the session you want to build (which should take care of anything it depends on) or you use `./run_tests <session>`.

README.md
lsf37: `Kernel_Config.thy` is generated by the target `c-config` in `spec/Makefile`. `design-spec` depends on that, and `ASpec` depends on `design-spec`.

lsf37:
Ok, so overall I think there are two issues we're trying to fix with improved docs:

- we're currently not mentioning `L4V_ARCH` at all and that is probably confusing
- we're currently not saying which sessions need something generated or how to find out

The `L4V_ARCH` part should probably go into the first paragraph. The other part might need a bit more explaining. Maybe I'll have a go at that if that's fine.

michaelmcinerney:
> The `L4V_ARCH` part should probably go into the first paragraph. The other part might need a bit more explaining. Maybe I'll have a go at that if that's fine.

Sure, please do! I have no idea what I'm doing

README.md
Xaphiosis: Then if we're saying what the things to invoke `make` on are, would `c-kernel` and `c-config` be the right targets then, rather than ASpec/CKernel?

lsf37: That would be the minimal command, but I'd keep the description to the proof sessions, because we might add more dependencies to them, but we are unlike to change the names of the main sessions.

PR #587, opened 2023-02-27
Word_Lib improvements

lsf37:
- define `sgn` and `abs`
- prove equalities between boolean and arithmetic operations (in rings, so applies to `int` and `word`). Examples include
  ```
  x + y = (x OR y) + (x AND y)
  x + y = (x XOR y) + 2 * (x AND y)
  x - y = (x AND NOT y) - (NOT x AND y)
  x XOR y = (x OR y) - (x AND y)
  x OR y = (x AND NOT y) + y
  ```
  etc. This should be a fairly complete set.
- prove all inequalities between binary boolean functions on words (e.g. `a && b <= a || b` etc)
- add laws about min, max, and NOT 
- some cleanup in `More_Word` (because some of these are now subsumed by ring lemmas) and doc improvements
- a lemma to turn `a < b` into a bitwise expression on abstract word lengths

Provides only additional rules or changes proofs, but should not change the proof context in any destructive way.

### DIFF
--- a/lib/Word_Lib/Boolean_Inequalities.thy
+++ b/lib/Word_Lib/Boolean_Inequalities.thy
@@ -1,0 +1,139 @@
+(*
+ * Copyright 2023, Proofcraft Pty Ltd
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ *)
+
+theory Boolean_Inequalities
+  imports Word_EqI
+begin
+
+section \<open>All inequalities between binary Boolean operations on @{typ "'a word"}\<close>
+
+text \<open>
+  Enumerates all binary functions resulting from Boolean operations on @{typ "'a word"}
+  and derives all inequalities of the form @{term "f x y \<le> g x y"} between them.
+  We leave out the trivial @{term "0 \<le> g x y"}, @{term "f x y \<le> -1"}, and
+  @{term "f x y \<le> f x y"}, because these are already readily available to the simplifier and
+  other methods.
+
+  This leaves 36 inequalities. Some of these are subsumed by each other, but we generate
+  the full list to avoid too much manual processing.
+
+  All inequalities produced here are in simp normal form.\<close>
+
+context
+  includes bit_operations_syntax
+begin
+
+(* The following are all 16 binary Boolean operations (on bool):
+   all_bool_funs \<equiv> [
+     \<lambda>x y. False,
+     \<lambda>x y. x \<and> y,
+     \<lambda>x y. x \<and> \<not>y,
+     \<lambda>x y. x,
+     \<lambda>x y. \<not>x \<and> y,
+     \<lambda>x y. y,
+     \<lambda>x y. x \<noteq> y,
+     \<lambda>x y. x \<or> y,
+     \<lambda>x y. \<not>(x \<or> y),
+     \<lambda>x y. x = y,
+     \<lambda>x y. \<not>y,
+     \<lambda>x y. x \<or> \<not>y,
+     \<lambda>x y. \<not>x,
+     \<lambda>x y. \<not>x \<or> y,
+     \<lambda>x y. \<not>(x \<and> y),
+     \<lambda>x y. True
+   ]
+
+  We can define the same for word operations:
+*)
+definition all_bool_word_funs :: "('a::len word \<Rightarrow> 'a word \<Rightarrow> 'a word) list" where
+  "all_bool_word_funs \<equiv> [
+     \<lambda>x y. 0,
+     \<lambda>x y. x AND y,
+     \<lambda>x y. x AND NOT y,
+     \<lambda>x y. x,
+     \<lambda>x y. NOT x AND y,
+     \<lambda>x y. y,
+     \<lambda>x y. x XOR y,
+     \<lambda>x y. x OR y,
+     \<lambda>x y. NOT (x OR y),
+     \<lambda>x y. NOT (x XOR y),
+     \<lambda>x y. NOT y,
+     \<lambda>x y. x OR NOT y,
+     \<lambda>x y. NOT x,
+     \<lambda>x y. NOT x OR y,
+     \<lambda>x y. NOT (x AND y),
+     \<lambda>x y. -1
+   ]"
+
+
+text \<open>
+  The inequalities on @{typ "'a word"} follow directly from implications on propositional
+  Boolean logic, which @{method simp} can solve automatically. This means, we can simply
+  enumerate all combinations, reduce from @{typ "'a word"} to @{typ bool}, and attempt to
+  solve by @{method simp} to get the complete list.\<close>
+local_setup \<open>
+let
+  (* derived from Numeral.mk_num, but returns a term, not syntax. *)
+  fun mk_num n =
+    if n > 0 then
+      (case Integer.quot_rem n 2 of
+        (0, 1) => \<^const>\<open>Num.One\<close>
+      | (n, 0) => \<^const>\<open>Num.Bit0\<close> $ mk_num n
+      | (n, 1) => \<^const>\<open>Num.Bit1\<close> $ mk_num n)
+    else raise Match
+
+  (* derived from Numeral.mk_number, but returns a term, not syntax. *)
+  fun mk_number n =
+    if n = 0 then \<^term>\<open>0::nat\<close>
+    else if n = 1 then \<^term>\<open>1::nat\<close>
+    else \<^term>\<open>numeral::num \<Rightarrow> nat\<close> $ mk_num n;
+
+  (* generic form of the goal statement *)
+  val goal = @{term "\<lambda>n m. (all_bool_word_funs!n) x y \<le> (all_bool_word_funs!m) x y"}
+  (* instance of the goal statement for a pair (i,j) of Boolean functions *)
+  fun stmt (i,j) = HOLogic.Trueprop $ (goal $ mk_number i $ mk_number j)
+
+  (* attempt to prove an inequality between functions i and j *)
+  fun le_thm ctxt (i,j) = Goal.prove ctxt ["x", "y"] [] (stmt (i,j)) (fn _ =>
+    (asm_full_simp_tac (ctxt addsimps [@{thm all_bool_word_funs_def}])
+     THEN_ALL_NEW resolve_tac ctxt @{thms word_leI}
+     THEN_ALL_NEW asm_full_simp_tac (ctxt addsimps @{thms word_eqI_simps bit_simps})) 1)
+
+  (* generate all combinations for (i,j), collect successful inequality theorems,
+     unfold all_bool_word_funs, and put into simp normal form. We leave out 0 (bottom)
+     and 15 (top), as well as reflexive thms to remove trivial lemmas from the list.*)
+  fun thms ctxt =
+    map_product (fn x => fn y => (x,y)) (1 upto 14) (1 upto 14)
+    |> filter (fn (x,y) => x <> y)
+    |> map_filter (try (le_thm ctxt))
+    |> map (Simplifier.simplify ctxt o Local_Defs.unfold ctxt @{thms all_bool_word_funs_def})
+in
+  fn ctxt =>
+    Local_Theory.notes [((Binding.name "word_bool_le_funs", []), [(thms ctxt, [])])] ctxt |> #2
+end
+\<close>
+
+(* Sanity checks: *)
+lemma
+  "x AND y \<le> x" for x :: "'a::len word"
+  by (rule word_bool_le_funs(1))
+
+lemma
+  "NOT x \<le> NOT x OR NOT y" for x :: "'a::len word"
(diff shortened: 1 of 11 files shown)


lib/Word_Lib/Word_EqI.thy
```diff
@@ -15,8 +15,9 @@ begin
 
 text \<open>
   Some word equalities can be solved by considering the problem bitwise for all
-  @{prop "n < LENGTH('a::len)"}, which is different to running @{text word_bitwise}
-  and expanding into an explicit list of bits.
+  @{prop "n < LENGTH('a::len)"}. This is similar to to running @{text word_bitwise}
+  and expanding into an explicit list of bits, but works on words of abstract size
+  instead of concrete word sizes only.
```
Xaphiosis, line 20: This makes it sound like it works on abstract instead of concrete-only, which is a strange way of saying it works on both?

lsf37: Hm, could be clearer, yes. What I'm trying to say is that `word_bitwise` works on concrete only, so there is a need for a proof method that can do more than that. `word_bitwise` can solve a larger set of goals, but only concrete. `word_eqI` can currently solve a smaller set of goals, but can do both abstract and concrete. Maybe I should just say that :-)

Xaphiosis: Something like that would be clearer, yes :)

lib/Word_Lib/Word_Lemmas.thy
```diff
@@ -25,6 +26,32 @@ context
   includes bit_operations_syntax
 begin
 
+lemma word_max_le:
```
Xaphiosis, line 29: missing `or` in the name?

lib/Word_Lib/More_Word.thy
```diff
@@ -216,16 +213,12 @@ lemma of_int_uint:
...
+  by (fact disjunctive_add2)
 
 corollary word_plus_and_or_coroll2:
   "(x AND w) + (x AND NOT w) = x"
   for x w :: \<open>'a::len word\<close>
-  apply (subst disjunctive_add)
-   apply (simp add: bit_simps)
-  apply (simp flip: bit.conj_disj_distrib)
-  done
+  by (fact and_plus_not_and)
```
Xaphiosis, line 221: silly question: what does the `fact` tactic do?

lib/Word_Lib/More_Bit_Ring.thy
```diff
@@ -0,0 +1,131 @@
...
+lemma plus_eq_and_or:
+  "x + y = (x OR y) + (x AND y)"
+  using add_commute local.add.semigroup_axioms or_eq_and_not_plus semigroup.assoc
+  by (fastforce simp: and_plus_not_and)
+
+lemma xor_eq_or_minus_and:
+  "x XOR y = (x OR y) - (x AND y)"
+  by (metis (no_types) bit.de_Morgan_conj bit.xor_def2 bit_and_iff bit_or_iff disjunctive_diff)
+
+lemma nxor_eq_and_plus_not_or:
```
Xaphiosis, line 89: nxnor? not_xnor would fit better with the names of the others, and xor lemmas aren't going to be in common use anyway

lib/Word_Lib/More_Bit_Ring.thy
```diff
@@ -0,0 +1,131 @@
...
+
+lemma minus_eq_xor_minus_not_and:
+  "x - y = (x XOR y) - 2 * (NOT x AND y)"
+  by (metis (no_types) bit.compl_eq_compl_iff bit.xor_compl_left not_diff_distrib
+                       plus_eq_xor_plus_carry)
+
+lemma minus_eq_and_not_minus_xor:
+  "x - y = 2 * (x AND NOT y) - (x XOR y)"
+  by (metis and.commute minus_diff_eq minus_eq_xor_minus_not_and xor.commute)
+
```
Xaphiosis, line 127: These are really cool! Are they all from hacker's delight?

lib/Word_Lib/Sgn_Abs.thy
```diff
@@ -0,0 +1,131 @@
...
+lemma word_abs_msb:
+  "\<bar>w\<bar> = (if msb w then -w else w)" for w::"'a::len word"
+  by (simp add: abs_word_def msb_word_iff_sless_0 word_sless_eq)
+
+lemmas word_abs_numeral[simp] = word_abs_msb[where w="numeral w" for w]
+
+
+subsection \<open>Properties\<close>
+
+(* Many of these are from linordered_idom, but need <s instead of < and occasionally
```
Xaphiosis, line 79: grepping for linordered_idom doesn't produce anything, either I'm blind or it's a possible typo

lib/Word_Lib/Boolean_Inequalities.thy
```diff
@@ -0,0 +1,141 @@
...
+  solve by @{method simp} to get the complete list.\<close>
+local_setup \<open>
+let
+  (* derived from Numeral.mk_num, but returns a term, not syntax. *)
+  fun mk_num n =
+    if n > 0 then
+      (case Integer.quot_rem n 2 of
+        (0, 1) => \<^const>\<open>Num.One\<close>
+      | (n, 0) => \<^const>\<open>Num.Bit0\<close> $ mk_num n
+      | (n, 1) => \<^const>\<open>Num.Bit1\<close> $ mk_num n)
```
Xaphiosis, line 85: how does this terminate? it recurses with the same argument

lib/Word_Lib/Boolean_Inequalities.thy
```diff
@@ -0,0 +1,141 @@
...
+
+  (* generic form of the goal statement *)
+  val goal = @{term "\<lambda>n m. (all_bool_word_funs!n) x y \<le> (all_bool_word_funs!m) x y"}
+  (* instance of the goal statement for a pair (i,j) of Boolean functions *)
+  fun stmt (i,j) = HOLogic.Trueprop $ (goal $ mk_number i $ mk_number j)
+
+  (* attempt to prove an inequality between functions i and j *)
+  fun le_thm (i,j) = Goal.prove @{context} ["x", "y"] [] (stmt (i,j)) (fn _ =>
+    (asm_full_simp_tac (@{context} addsimps [@{thm all_bool_word_funs_def}]) THEN_ALL_NEW
+     resolve_tac @{context} @{thms word_leI} THEN_ALL_NEW
```
Xaphiosis, line 102: in Isabelle sources things like `THEN_ALL_NEW` go on the left

lib/Word_Lib/Boolean_Inequalities.thy
```diff
@@ -0,0 +1,141 @@
...
+  Enumerates all binary functions resulting from Boolean operations on @{typ "'a word"}
+  and derives all inequalities of the form @{term "f x y \<le> g x y"} between them.
+  We leave out the trivial @{term "0 \<le> g x y"}, @{term "f x y \<le> -1"}, and
+  @{term "f x y \<le> f x y"}, because these are already readily available to the simplifier and
+  other methods.
+
+  This leaves 36 inequalities. Some of these are subsumed by each other, but we generate
+  the full list to avoid too much manual processing.
+
+  All inequalities produced here are in simp normal form.\<close>
```
Xaphiosis, line 23: This is a really cool idea. How did you come up with it?

Xaphiosis:
Ok, I've reviewed all the commits in this PR now.

lib/Word_Lib/More_Bit_Ring.thy
lsf37: Mostly yes, there is a page of equations listed somewhere, and I just proved them all.

lib/Word_Lib/Sgn_Abs.thy
lsf37: If you grep in Isabelle there should be plenty. It's one of the algebraic type classes.

lib/Word_Lib/Boolean_Inequalities.thy
lsf37: There was a table of all boolean operators in the book and a description that you can read off `<=` from that. The description was implication (without saying the name), and from that it wasn't a big step to generating them all instead of trying to write them down.

I've been thinking about generating all 3-argument boolean operators as well (same principle), but it's probably overkill for now.

michaelmcinerney (approved):
Looks great!

lib/Word_Lib/Sgn_Abs.thy
Xaphiosis: Yes, I think `ag` and repo-in-repo messed up my checking. All good.

lib/Word_Lib/Boolean_Inequalities.thy
```diff
@@ -0,0 +1,141 @@
...
+     and 15 (top), as well as reflexive thms to remove trivial lemmas from the list.*)
+  val thms =
+    map_product (fn x => fn y => (x,y)) (1 upto 14) (1 upto 14)
+    |> filter (fn (x,y) => x <> y)
+    |> map_filter (try le_thm)
+    |> map (Simplifier.simplify @{context} o
+            Local_Defs.unfold @{context} @{thms all_bool_word_funs_def})
+in
+  fn ctxt => Local_Theory.notes [((Binding.name "word_bool_le_funs", []), [(thms, [])])] ctxt |> #2
+end
```
corlewis, line 116: It's a bit pedantic, but I think that it would be better style to pass around `ctxt` instead of using the `@{context}` antiquotation in the other functions.

lib/Word_Lib/Boolean_Inequalities.thy
```diff
@@ -129,12 +129,10 @@ lemma
   "x XOR y \<le> NOT x OR NOT y" for x :: "'a::len word"
   by (rule word_bool_le_funs)
 
-declare word_bool_le_funs[intro!]
-
```
corlewis: Very minor, but if the intention is to not have these rules available to the simplifier then you might as well remove this `declare` from the previous commit.

corlewis (approved):
Very nice!

lib/Word_Lib/More_Word.thy
lsf37: My memory is hazy, but I think it's pretty much equivalent to `rule`, but does a more elaborate lookup of facts from the local proof context. I.e. you can say `by fact` and get things you have proved with `have` without name etc.

lib/Word_Lib/More_Bit_Ring.thy
lsf37: you mean `not_xor`? The thing is that `not xor` is the same as `equals` (usually written `≡`), and I almost defined that as an operator, but then we'd have to have all kinds of basic rules and setup for it and the symbol is already taken etc, so I left it, but I still had the feeling that `not_xor` is a long name for a thing that is actually simpler than `xor`.

But you're right, `nxor` doesn't really make anything better. Will change to `not_xor`.

lib/Word_Lib/Boolean_Inequalities.thy
lsf37: I find both in the existing code (both ours and Isabelle), but I'm happy to change it, I don't really mind that one either way.

lib/Word_Lib/Boolean_Inequalities.thy
lsf37: Done.

lib/Word_Lib/Boolean_Inequalities.thy
lsf37: Also done.

lib/Word_Lib/Boolean_Inequalities.thy
```diff
@@ -0,0 +1,139 @@
...
+  (* generic form of the goal statement *)
+  val goal = @{term "\<lambda>n m. (all_bool_word_funs!n) x y \<le> (all_bool_word_funs!m) x y"}
+  (* instance of the goal statement for a pair (i,j) of Boolean functions *)
+  fun stmt (i,j) = HOLogic.Trueprop $ (goal $ mk_number i $ mk_number j)
+
+  (* attempt to prove an inequality between functions i and j *)
+  fun le_thm ctxt (i,j) = Goal.prove ctxt ["x", "y"] [] (stmt (i,j)) (fn _ =>
+    (asm_full_simp_tac (@{context} addsimps [@{thm all_bool_word_funs_def}])
+     THEN_ALL_NEW resolve_tac @{context} @{thms word_leI}
+     THEN_ALL_NEW asm_full_simp_tac (@{context} addsimps @{thms word_eqI_simps bit_simps})) 1)
```
corlewis, line 103: Out of curiosity, did you intentionally leave these `@{context}` antiquotations?

lsf37: Thanks for spotting those, I missed them!

lsf37: Now I have them all :-)

PR #588, opened 2023-02-28
lib: connection between exs_valid and wp conjugate

lsf37:
There is nothing new under the sky: draw connection between conjugate wp in the literature and our exs_valid definition.

Add exs_valid_alt lemma, which is one of the main rules that is different between wp and conjugate wp (or vs and).

This is mainly documentation and part of the quest to anchor the l4v concepts in standard literature concepts.


lib/Monads/NonDetMonad_Sat.thy
```diff
@@ -28,6 +28,29 @@ definition ex_exs_validE ::
   "\<lbrace>P\<rbrace> f \<exists>\<lbrace>Q\<rbrace>, \<lbrace>E\<rbrace> \<equiv> \<lbrace>P\<rbrace> f  \<exists>\<lbrace>\<lambda>rv. case rv of Inl e \<Rightarrow> E e | Inr v \<Rightarrow> Q v\<rbrace>"
```
Xaphiosis, line 30: nitpick: why two lines here, but one line before `subsection` below? suggest other way around

Xaphiosis: oh, that's the local style. sad face.

lib/Monads/NonDetMonad_Sat.thy
```diff
@@ -28,6 +28,29 @@ definition ex_exs_validE ::
...
 
 
+text \<open>
+  Seen as predicate transformer, this is the so-called conjugate wp in the literature, i.e. with
```
Xaphiosis, line 32: I'm confused. What's seen as [a] predicate transformer?

lsf37: The definitions above the sentence (`exs_valid` and `exs_validE`). I can make that more explicit.

lib/Monads/NonDetMonad_Sat.thy
lsf37: You're right, we should change that style, it's a bit strange.

lib/Monads/NonDetMonad_Sat.thy
Xaphiosis: Yes please

lib/Monads/NonDetMonad_Sat.thy
lsf37: Style updated.

lib/Monads/NonDetMonad_Sat.thy
lsf37: Done!

lib/Monads/NonDetMonad_Sat.thy
```diff
@@ -20,13 +20,36 @@ definition exs_valid ::
...
 
-
 text \<open>The above for the exception monad\<close>
 definition ex_exs_validE ::
   "('a \<Rightarrow> bool) \<Rightarrow> ('a, 'e + 'b) nondet_monad \<Rightarrow> ('b \<Rightarrow> 'a \<Rightarrow> bool) \<Rightarrow> ('e \<Rightarrow> 'a \<Rightarrow> bool) \<Rightarrow> bool"
   ("\<lbrace>_\<rbrace> _ \<exists>\<lbrace>_\<rbrace>, \<lbrace>_\<rbrace>") where
   "\<lbrace>P\<rbrace> f \<exists>\<lbrace>Q\<rbrace>, \<lbrace>E\<rbrace> \<equiv> \<lbrace>P\<rbrace> f  \<exists>\<lbrace>\<lambda>rv. case rv of Inl e \<Rightarrow> E e | Inr v \<Rightarrow> Q v\<rbrace>"
 
+text \<open>
+  Seen as predicate transformer, @{const exs_valid} is the so-called conjugate wp in the literature,
```
Xaphiosis, line 30: to make sure, it's called "conjugate wp" and is not "conjugate *of* wp"? showing my ignorance here

Xaphiosis (approved):

lib/Monads/NonDetMonad_Sat.thy
lsf37: I've only heard the former, but I suspect the latter actually makes more sense. Both will be clear when you know what it is and not say much when you don't. The comment is mostly there to draw the connection for people coming from a more theoretical background (and to show the connection between `valid` and `exs_valid`, that part hopefully comes across without the terminology).

PR #589, opened 2023-03-01
github: provide nl-unescape script to BV trigger job

lsf37:
Need to check out the ci-actions repo first (where the nl-unescape.sh script is located).

This should get the trigger action working to the degree that `xmllint` doesn't complain any more.


wom-bat (approved):
LGTM

PR #590, opened 2023-03-01
cspec: adjust Kernel_Config generation

lsf37:
seL4/seL4#975 slightly changed how the config headers are generated. They now need a (short) `ninja` build step and they produce less spaces in the header file.

This PR adjusts for that, but does not read the generated json file yet, which should be more robust in the future than parsing the C header.

### DIFF
--- a/spec/cspec/c/gen-config-thy.py
+++ b/spec/cspec/c/gen-config-thy.py
@@ -193,7 +193,7 @@ def parse_gen_config(gen_config_file: str) -> Dict[str, str]:
     """
     Parse gen_config.h and return a dictionary of the values.
     """
-    end_comment_re = re.compile(r'  /\*.*\*/')
+    end_comment_re = re.compile(r'/\*.*\*/')
 
     config = {}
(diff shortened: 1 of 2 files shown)


lsf37:
Appears to be working correctly, the failure before was very quick ;-)

Xaphiosis (approved):

nspin:
Oops, I meant to preserve the per-line format of `gen_config.h` in https://github.com/seL4/seL4/pull/975, but, as you've discovered, I accidentally introduced a change:

`#define X    /* empty */` to `#define X  /* empty */`

Sorry about that.

lsf37:
> Oops, I meant to preserve the per-line format of `gen_config.h` in [seL4/seL4#975](https://github.com/seL4/seL4/pull/975), but, as you've discovered, I accidentally introduced a change:
> 
> `#define X /* empty */` to `#define X /* empty */`
> 
> Sorry about that.

No worries, I shouldn't be relying on existence of spaces :-)

PR #591, opened 2023-03-02
cspec: provide mechanism for adding dts overlays

lsf37:
Add mechanism for adding overlay.dts files to the l4v build for all architectures apart from X64 (which does not use dts files).

For example, place a file `overlays/ARM/overlay.dts` into the tree and the build will pick it up as custom overlay file with the correct proof session dependencies.

If no file is provided, an empty default overlay file is used.

This mechanism is useful for a multi-kernel setup that uses overlays to restrict the available memory regions per CPU node.

### DIFF
--- a/spec/cspec/c/overlays/README.md
+++ b/spec/cspec/c/overlays/README.md
@@ -1,0 +1,22 @@
+<!--
+     Copyright 2024, Proofcraft Pty Ltd
+
+     SPDX-License-Identifier: CC-BY-SA-4.0
+-->
+
+# DTS Overlays
+
+This directory contains device tree overlay files that can be used to override
+default platform parameters such as available memory regions.
+
+The default files in the repository are all empty.
+
+To provide an override, place a file `overlay.dts` into the respective
+architecture directory, e.g. `ARM/overlay.dts`.
+
+The l4v build system will pick up this overlay file when it generates the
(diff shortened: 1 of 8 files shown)


spec/cspec/c/overlays/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+To provide an override, place a file `overlay.dts` into the respective
+architecture directory, e.g. `ARM/overlay.dts`.
+
+The l4v build system will pick up this overlay file when it generates the
+kernel configuration data and preprocessed kernel code. It will rebuild proof
+sessions according to their dependencies.
+
+The `X64` build does not support overlays, and therefore does not provide a
+default.
```
Xaphiosis, line 22: I like this. Question: should any other part of the docs or READMEs mention this so that people can find it?

Xaphiosis (approved):

spec/cspec/c/overlays/README.md
lsf37: I couldn't find anything that seemed appropriate, but if you have a preference, happy to add it.

Xaphiosis: The place I can think of is in `spec/cspec/c/README.md` after the paragraph "The target architecture may be specified by setting..."

PR #592, opened 2023-03-02
Various updates for binary verification

mbrcknl:
(I've lumped several things into one PR, but am happy to split them up if you'd prefer to review separately.)

Related PRs: seL4/ci-actions#256, seL4/seL4#982

## Add a script to export kernel build artifacts

Previously, the ci-actions `vm-steps.sh` saved a copy of the C graph-lang for the l4v workflow to upload as an artifact. Going forward, I would like to ensure that binary verification works on binaries built by the same toolchain used in the l4v proofs. But I don't think it makes sense to pollute the ci-actions `vm-steps.sh` script with even more specific details of the files that should be exported. This script allows `vm-steps.sh` to delegate most of that work to l4v, which already needs to know how to build the kernel.

## Move decompilation workflow to graph-refine

Since I'm planning to merge graph-refine updates to master in the near future, I think it makes sense to move decompilation workflows there. Note that the repository-dispatch won't work until that graph-refine merge is complete, but since decompilation is currently broken in l4v, we don't really lose anything. In the meantime, I can manually trigger the decompilation workflows I need.

## c-kernel: Support pre-built standalone C parser

This makes it possible to use the l4v kernel build makefile with a pre-built standalone C parser. This is useful if you want to perform kernel builds in a Docker container.

This isn't strictly needed for binary verification CI, since BV CI will primarily use the kernel builds generated via the aws-proofs `vm-steps.sh` and `export-kernel-builds.py`. However, I have some tools in graph-refine for performing kernel builds with different compiler versions, in case that's useful. This change allows me to bundle those tools into Docker images with a pre-built standalone C parser.


.github/workflows/proof.yml
```diff
@@ -21,7 +21,7 @@ on:
       - aarch64
 
 jobs:
-  all:
+  proofs:
     name: All
```
lsf37, line 25: This is the action for pull request checks. Do we want to trigger binary verification from those? I'd have thought those workflows are too expensive for that.

mbrcknl: I don't really mind either way.

On one hand, I was thinking that decompilation is possibly cheap enough to run in the same cases we run Isabelle proofs, so we could just try it and see. Decompilation duration is about the same order of magnitude as the AWS proofs. Although decompilation will do the work every time, whereas Isabelle is smart enough to use cached images when possible. On the back end, we can use the tags to control which jobs actually run graph-refine. Initially, I would allow tags `l4v/weekly-clean` and `l4v/proof-deploy/*`, so we wouldn't run graph-refine on PRs unless I intervene to add the job to the queue.

On the other hand, we can manually trigger decompilation on the artifacts produced by PR proof runs, as long as the proof run exports the build artifacts. So we wouldn't lose much by not triggering decompilation here.

What do you think?

lsf37: If it's on the same order as the AWS proofs that means it'll roughly double the PR wait time, because the decompilation can only start after the proof workflow is finished and has uploaded artifacts. I'd says that's too long to always trigger.

But a manual trigger for BV on PRs would make sense, so that we can kick it off if we're doing anything we think might interfere with decompilation (most stuff on Refine and above won't really have any influence on BV outcomes).

We should in fact probably add such a trigger to seL4 pull requests, because they are much more likely to break BV. But let's do that once the infrastructure works here.

.github/workflows/binary-verification.yml
```diff
@@ -0,0 +1,60 @@
+# Copyright 2023 Kry10 Limited
+#
+# SPDX-License-Identifier: BSD-2-Clause
+
+# Can be called from a proof run to trigger binary verification.
+# Issues a repository-dispatch event on the graph-refine repository,
+# but only if there are c-graph-lang artifacts available.
+
+on:
```
lsf37, line 9: We should give a `name: ..` line. Or does that interfere with being called from other workflows?

.github/workflows/proof.yml
lsf37: Ok, so I vote for keeping the artifact upload, but I would remove the automatic trigger for this workflow.

mbrcknl: Decompilation will be a separate workflow (on the graph-refine repo), triggered from an l4v proof run using repository-dispatch. So it won't delay l4v proof results.

However, having decompilation triggered automatically might mean you can run graph-refine sooner, if you decide you want that. Currently, it would not be possible to manually trigger a decompilation workflow until the l4v proof run is finished, so there would always be a delay between the l4v workflow finishing and the human remembering to start decompilation. We could get around that by making the repository-dispatch conditional on a label on the l4v workflow, but I currently don't have an implementation for that.

So I think the choice is mainly about how much of GitHub's compute budget we want to spend on decompilation, since they've made it free for us (for now!).

Note that strictly speaking, it would be possible to run decompilation in parallel with l4v proofs, but doing that makes some things difficult: It's harder to ensure that l4v and the decompiler are using the same kernel build. And it's harder to figure out how to trigger graph-refine, since it depends on both decompilation and l4v. The choice to have decompilation triggered from l4v is largely to make those problems go away.

W.r.t. triggers on seL4 pull requests, my thought was that we would use the existing mechanisms to trigger l4v proof runs, and then the l4v proof workflow would trigger decompilation. We could use a label on the seL4 PR to signal whether the l4v proof run should trigger decompilation, but the overall process would still need to go via the l4v proof workflow.

spec/cspec/c/export-kernel-builds.py
```diff
@@ -0,0 +1,180 @@
...
+                f'Error: Kernel build for {config_name} failed', file=sys.stderr)
+            sys.exit(1)
+
+        # Copy the preprocessed source from the l4v build, so we can later compare
+        # it to the exported builds if binary verification fails. There's no point
+        # comparing here, because there will always be differences due to lines
+        # inserted by the preprocessor and config system.
+        if self.l4v_paths.c_pp.is_file():
+            shutil.copyfile(self.l4v_paths.c_pp, export_dir /
+                            'kernel_all.c_pp.l4v')
```
lsf37, line 112: I think @Xaphiosis has a way of stripping those lines if that helps.

spec/cspec/c/export-kernel-builds.py
```diff
@@ -0,0 +1,180 @@
...
+    c_pp: Path
+    c_functions: Path
+
+
+def get_l4v_paths(l4v_arch: str) -> L4vPaths:
+    l4v_c = Path(__file__).resolve().parent
+    l4v = l4v_c.parent.parent.parent
+    kernel_mk = l4v_c / 'kernel.mk'
+    c_pp = l4v_c / f'build/{l4v_arch}/kernel_all.c_pp'
+    c_functions = l4v / f'proof/asmrefine/export/{l4v_arch}/CFunDump.txt'
```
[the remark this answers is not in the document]
mbrcknl: And now I'm wondering why I didn't use it more completely in the lines you just quoted... :-/

spec/cspec/c/kernel.mk
```diff
@@ -8,29 +8,39 @@
...
+endif
+
 ifndef L4V_REPO_PATH
   L4V_REPO_PATH := $(realpath $(dir $(lastword ${MAKEFILE_LIST}))../../..)
 endif
 
-ifndef SOURCE_ROOT
-  SOURCE_ROOT := $(realpath ${L4V_REPO_PATH}/../seL4)
+ifndef KERNEL_SOURCE_ROOT
+  KERNEL_SOURCE_ROOT := $(realpath ${L4V_REPO_PATH}/../seL4)
```
lsf37, line 20: Any specific reason for the rename? (It's going to conflicts with other build system PRs I have in the pipeline)

spec/cspec/c/kernel.mk
```diff
@@ -8,29 +8,39 @@
...
 endif
 
 ifndef CONFIG_DOMAIN_SCHEDULE
   CONFIG_DOMAIN_SCHEDULE := ${CSPEC_DIR}/c/config_sched.c
 endif
 
+# Default to the setting in the CONFIG file.
+ifdef CONFIG_OPTIMISATION
+  KERNEL_CMAKE_OPTIMISATION := -DKernelOptimisation=${CONFIG_OPTIMISATION}
+endif
```
lsf37, line 42: which `CONFIG` file is the comment referring to?

spec/cspec/c/kernel.mk
mbrcknl: No strong reason. I think it was just inspired by `KERNEL_BUILD_ROOT`. I'll change it back.

spec/cspec/c/kernel.mk
mbrcknl: The `CONFIG` variable a few lines up. I'll make that more explicit, by changing it to `${CONFIG}`.

spec/cspec/c/kernel.mk
```diff
@@ -42,7 +52,7 @@ ifndef TOOLPREFIX
     endif
   endif
   ifdef TRY_TOOLPREFIX
-    TOOLPREFIX := $(firstword $(strip $(foreach TRY,${TRY_TOOLPREFIX},$(if $(shell which ${TRY}gcc),${TRY},))))
+    TOOLPREFIX := $(firstword $(strip $(foreach TRY,${TRY_TOOLPREFIX},$(if $(shell command -v ${TRY}gcc),${TRY},))))
```
lsf37, line 55: Happy to change it, but can you elaborate on what went wrong with `which`?

spec/cspec/c/kernel.mk
```diff
@@ -127,6 +119,59 @@ ${KERNEL_CONFIG_ROOT}/.cmake_done: ${KERNEL_DEPS} gen-config-thy.py
...
+	${OBJDUMP} -dz $^ > $@
+
+${KERNEL_BUILD_ROOT}/kernel.elf.symtab: ${KERNEL_BUILD_ROOT}/kernel.elf
+	${OBJDUMP} -t $^ > $@
+
+# Override STANDALONE_C_PARSER_EXE to use a pre-built standalone C parser.
+ifndef STANDALONE_C_PARSER_EXE
+  STANDALONE_C_PARSER_DIR := ${L4V_REPO_PATH}/tools/c-parser/standalone-parser
+  STANDALONE_C_PARSER_EXE := ${STANDALONE_C_PARSER_DIR}/${L4V_ARCH}/c-parser
+endif
```
lsf37, line 143: This doesn't seem to override as the comment says, but instead provide a default. Am I reading that correctly?

lsf37: Or is the comment trying to say that the user can override this variable to use a different standalone parser?

spec/cspec/c/kernel.mk
mbrcknl: On most platforms, `which` writes messages to stderr when it doesn't find anything, and not every platform has a `-s` silent option. In this part of the makefile, `which` was used in a way where it's likely to not find several things. The messages were a bit distracting when I was trying to figure out some other build problem, since it took me a couple of minutes to realise most of the messages were harmless.

lsf37 (approved):
Cool. Happy for you to merge after the comments are addressed.

spec/cspec/c/kernel.mk
mbrcknl: Oh, I didn't consider that reading of the comment. In my head, it said "If you want to use a pre-built standalone C parser, then set STANDADLONE_C_PARSER_EXE so this default isn't used."

I will try to improve that.

spec/cspec/c/export-kernel-builds.py
mbrcknl: Yeah, that would be cool.

spec/cspec/c/kernel.mk
mbrcknl: > Or is the comment trying to say that the user can override this variable to use a different standalone parser?

Our replies crossed paths, but yes, that is how it read in my head. :-)

I do acknowledge the ambiguity, which I will try to fix.

spec/cspec/c/kernel.mk
lsf37: Cool, one more thing I learned today :-). Definitely happy with that change.

.github/workflows/binary-verification.yml
mbrcknl: I think I just forgot to name it. Will fix.

.github/workflows/proof.yml
lsf37: > Decompilation will be a separate workflow (on the graph-refine repo), triggered from an l4v proof run using repository-dispatch. So it won't delay l4v proof results.

It wouldn't delay those results, but if we wanted to know if it succeeded or not, we'd still have to wait. If we don't want to know, we probably shouldn't run it. While the minutes are free, the number of concurrent runners is limited, which we'd be using up.

> Currently, it would not be possible to manually trigger a decompilation workflow until the l4v proof run is finished, so there would always be a delay between the l4v workflow finishing and the human remembering to start decompilation. We could get around that by making the repository-dispatch conditional on a label on the l4v workflow, but I currently don't have an implementation for that.

Yes, I was thinking of the label approach. It's Ok to leave that for now, it's more important to get coverage for a release. And I think I can implement the label relatively easily from my side if need be.

> W.r.t. triggers on seL4 pull requests, my thought was that we would use the existing mechanisms to trigger l4v proof runs, and then the l4v proof workflow would trigger decompilation. We could use a label on the seL4 PR to signal whether the l4v proof run should trigger decompilation, but the overall process would still need to go via the l4v proof workflow.

This works for pushes to master (after merge), but not for pull request. Proofs on pull requests in the seL4 repo are currently just `CRefine` and `SimplExportAndRefine` and run in the seL4 repo. But if the infrastructure is available to run it in l4v with repository dispatch, I can do the rest when we decide we want that feature.

spec/cspec/c/kernel.mk
```diff
@@ -109,8 +105,8 @@ ${KERNEL_BUILD_ROOT}/.cmake_done: ${KERNEL_DEPS} ${CONFIG_DOMAIN_SCHEDULE}
 		-DCROSS_COMPILER_PREFIX=${TOOLPREFIX} \
 		-DCMAKE_TOOLCHAIN_FILE=${SOURCE_ROOT}/gcc.cmake \
 		-DKernelDomainSchedule=${CONFIG_DOMAIN_SCHEDULE} \
-		-DUMM_TYPES=$(abspath ${UMM_TYPES}) \
-		-DCSPEC_DIR=${CSPEC_DIR} ${KERNEL_CMAKE_EXTRA_OPTIONS} \
+		-DUMM_TYPES=$(abspath ${UMM_TYPES}) -DCSPEC_DIR=${CSPEC_DIR} \
+		${KERNEL_CMAKE_OPTIMISATION} ${KERNEL_CMAKE_EXTRA_OPTIONS} \
```
lsf37, line 109: Just saw that we should make the same change to the config target. Some of the other options depend on optimisation options and we want the config build to be in sync with the main build. (The proofs will fail if they are not, so it doesn't need to be trusted, but of course we don't want them to fail).

mbrcknl: Right, I had assumed it wouldn't matter, but you're right that we should ensure consistency. I will update.

.github/workflows/proof.yml
mbrcknl: > While the minutes are free, the number of concurrent runners is limited, which we'd be using up.

That's a good point, which I hadn't considered. So I agree that we should remove it here.

> > Currently, it would not be possible to manually trigger a decompilation workflow until the l4v proof run is finished, so there would always be a delay between the l4v workflow finishing and the human remembering to start decompilation. We could get around that by making the repository-dispatch conditional on a label on the l4v workflow, but I currently don't have an implementation for that.
> 
> Yes, I was thinking of the label approach. It's Ok to leave that for now, it's more important to get coverage for a release. And I think I can implement the label relatively easily from my side if need be.

Sounds good.

> > W.r.t. triggers on seL4 pull requests, my thought was that we would use the existing mechanisms to trigger l4v proof runs, and then the l4v proof workflow would trigger decompilation. We could use a label on the seL4 PR to signal whether the l4v proof run should trigger decompilation, but the overall process would still need to go via the l4v proof workflow.
> 
> This works for pushes to master (after merge), but not for pull request. Proofs on pull requests in the seL4 repo are currently just `CRefine` and `SimplExportAndRefine` and run in the seL4 repo. But if the infrastructure is available to run it in l4v with repository dispatch, I can do the rest when we decide we want that feature.

Right, that makes sense, and now I see what you meant by your original comment. I'll make a PR to add the artifact upload to the seL4 proof workflow, but I'll leave out the trigger for now. The manual decompilation trigger will be able to retrieve kernel build artifacts from the seL4 proof workflow in the same way it gets them from l4v proof workflows.

spec/cspec/c/export-kernel-builds.py
mbrcknl: I might try to figure this out later, but I'll leave it as it is for now.

PR #593, opened 2023-03-03
`ccorres_call_getter_setter`

michaelmcinerney:
I wasn't sure whether we would like a comment to explain how this rule might be used. I could include something along the lines of what I have in the commit message, maybe also mentioning that `bind_assoc_reverse` might be used to rewrite the abstract side into the format in the rule.


Xaphiosis:
Honestly, future victims of verification need all the help they can get, so thumbs up to  comment and even mentioning bind_assoc_reverse as a hint

Xaphiosis:
Small nitpick: "to reason" is very broad, maybe "prove correspondence"?

lib/clib/Corres_UL_C.thy
```diff
@@ -820,6 +820,37 @@ lemma ccorres_call:
...
+lemma ccorres_call_getter_setter:
+  assumes cul: "ccorresG sr \<Gamma> (=) xf' P (i ` P') [] getter (Call f)"
+  and     gsr: "\<And>x x' s t rv rv'.
+                  \<lbrakk> (x, t) \<in> sr; rv = xf' t; (rv', x') \<in> fst (setter rv x) \<rbrakk>
+                  \<Longrightarrow> (x', g s t (clean s t)) \<in> sr"
+  and     res: "\<And>s t rv. rv = xf' t \<Longrightarrow> rv = xf (g s t (clean s t))"
+  and     ist: "\<And>x s. (x, s) \<in> sr \<Longrightarrow> (x, i s) \<in> sr"
+  and      ef: "\<And>val. empty_fail (setter val)"
+  shows "ccorresG sr \<Gamma> (=) xf P P' hs
+           (do val \<leftarrow> getter; setter val; return val od)
```
Xaphiosis, line 832: thank you for calling these getter and setter as opposed to f and g

lsf37: Although it does suggest that `getter` doesn't change the state, which in our application it does. I'm happy with leaving it as getter/setter, though.

lsf37 (approved):
Nice to finally have a rule for globals. I agree that adding a comment about how/when to use this one would be helpful.

michaelmcinerney:
Forced pushed after rebase.

Now with comment! I'm not totally sure on the right terminology here (e.g., "call a function" versus "call a procedure") so happy to change things around

lsf37:
👍 In C, "function" is the right terminology.

lib/clib/Corres_UL_C.thy
```diff
@@ -820,6 +820,15 @@ lemma ccorres_call:
   apply simp
   done
 
+text \<open>This rule is intended to be used in the case where the C calls a function and then uses the
```
Xaphiosis, line 823: Style for text comments is different, see style guide.
First paragraph great, second paragraph I'm confused by "allows us to establish a
nontrivial relation between the values we set, if we wish" - what is "values we set" here?

Xaphiosis:
@lsf37 : what if anything did we conclude in the past about tags for moving stuff to lib? In this case two commits are pointless, the `lib:` tag is good, but should it get anything else?

lib/clib/Corres_UL_C.thy
michaelmcinerney: We get a value and then set the value. It's easiest to see in the Haskell: it's the `val` return value that the `getter` gets, which is then passed to `setter` which takes it as an argument. Same sort of deal, but of course more complicated, in the C

lsf37:
> @lsf37 : what if anything did we conclude in the past about tags for moving stuff to lib? In this case two commits are pointless, the `lib:` tag is good, but should it get anything else?

I'd be happy for these to just be `lib:`, but I have no strong feelings either way on this one, since it's so small.

Xaphiosis (approved):

corlewis:
> @lsf37 : what if anything did we conclude in the past about tags for moving stuff to lib? In this case two commits are pointless, the `lib:` tag is good, but should it get anything else?

If we're making a new decision about things like this then I would argue for just a `lib` tag. It's definitely useful to know that a change has been made to `lib`, but I can't think of a use-case for knowing that something was moved out of `ainvs`. In practice, nothing has changed for anyone working in or on top of `ainvs`, the rule still exists and rebasing over it is highly unlikely to cause any problems.

Xaphiosis:
> > @lsf37 : what if anything did we conclude in the past about tags for moving stuff to lib? In this case two commits are pointless, the `lib:` tag is good, but should it get anything else?
> 
> If we're making a new decision about things like this then I would argue for just a `lib` tag. It's definitely useful to know that a change has been made to `lib`, but I can't think of a use-case for knowing that something was moved out of `ainvs`. In practice, nothing has changed for anyone working in or on top of `ainvs`, the rule still exists and rebasing over it is highly unlikely to cause any problems.

Happy to go with that unless we're dealing with something significantly bigger. Then you might still be right, but I'd want to discuss first.

PR #594, opened 2023-03-03
cspec: extract physBase from C headers

lsf37:
Extract the numeric value physBase_raw from the generated header gen_headers/plat/machine/devices_gen.h and provide it as the constant physBase in Kernel_Config.thy.

In C this will later match up with the value returned by physBase().

Test with: seL4/seL4#983

(Test will fail until we have adjusted the proofs for use of `physBase()`)

### DIFF
--- a/spec/cspec/c/gen-config-thy.py
+++ b/spec/cspec/c/gen-config-thy.py
@@ -208,6 +208,24 @@ def parse_gen_config(gen_config_file: str) -> Dict[str, str]:
     return config
 
 
+def parse_physBase(l4v_arch: str, devices_gen_h_file: str) -> str:
+    """
+    Parse devices_gen.h and return the physBase value as string in hex or
+    decimal form. Return None for architectures that don't use physBase.
+    """
+    if l4v_arch == 'X64':
+        return None
+
+    physBase_re = re.compile(r'#define PHYS_BASE_RAW (0x[0-9a-fA-F]+|[0-9]+)')
+    with open(devices_gen_h_file, 'r') as f:
+        for line in f:
+            line = line.strip()
+            m = physBase_re.match(line)
+            if m:
+                return m.group(1)
+    raise Exception(f'Could not find PHYS_BASE_RAW in {devices_gen_h_file}')
+
+
 def add_defaults(config: Dict[str, str]):
     """
     Defaults are boolean config keys that are mentioned in known_config_keys
@@ -215,7 +233,7 @@ def add_defaults(config: Dict[str, str]):
     config file.
     """
     for key, (_, name) in known_config_keys.items():
-        if key not in config and not name is None:
+        if key not in config and name is not None:
             config[key] = 0
 
 
@@ -235,25 +253,27 @@ begin
   GENERATED -- DO NOT EDIT! Changes will be overwritten.
 
   This file was generated from {}
+  and {}
   by the script {}.
 *)
 
 """
 
 
-def write_config_thy(config_thy_path, config: Dict[str, str]):
+def write_config_thy(header, config_thy_path, config: Dict[str, str], physBase=None):
     """
     Write a Kernel_Config.thy file for a given configuration dict.
     """
     file_name = path.realpath(path.join(config_thy_path, 'Kernel_Config.thy'))
     with open(file_name, 'w') as f:
-        f.write(theory_header.format(
-            date.today().year,
-            config_path,
-            path.realpath(__file__))
-        )
+        f.write(header)
 
-        names = []
+        if physBase:
+            f.write('(* This value is PHYS_BASE_RAW in the devices_gen.h header listed above. *)\n')
+            f.write('definition physBase :: machine_word where\n')
+            f.write(f'  "physBase \\<equiv> {physBase}"\n\n')
+
+        names = ['physBase'] if physBase else []
         for key, value in config.items():
             type = type_of(key)
             if type is nat or type is word:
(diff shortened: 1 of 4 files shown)


lsf37:
The proofs are unexpectedly robust, all CRefines apart from ARM_HYP work with `physBase` as a function..

Xaphiosis:
> The proofs are unexpectedly robust, all CRefines apart from ARM_HYP work with `physBase` as a function..

That feels a bit wrong...

spec/cspec/c/gen-config-thy.py
```diff
@@ -208,14 +208,32 @@ def parse_gen_config(gen_config_file: str) -> Dict[str, str]:
     return config
 
 
+def parse_physBase(L4V_ARCH: str, devices_gen_h_file: str) -> str:
```
Xaphiosis, line 211: if you want to follow python standards, `L4V_ARCH` as an argument name is a bit sus

spec/cspec/c/gen-config-thy.py
```diff
@@ -208,14 +208,32 @@ def parse_gen_config(gen_config_file: str) -> Dict[str, str]:
...
+    """
+    Parse devices_gen.h and return the physBase value as string in hex or
+    decimal form. Return None for architectures that don't use physBase.
+    """
+    if L4V_ARCH == 'X64':
+        return None
+
+    physBase_re = re.compile(r'#define physBase_raw (0x[0-9a-fA-F]+|[0-9]+)')
```
Xaphiosis, line 219: -> PHYS_BASE_RAW now

spec/cspec/c/gen-config-thy.py
```diff
@@ -208,14 +208,32 @@ def parse_gen_config(gen_config_file: str) -> Dict[str, str]:
...
+    with open(devices_gen_h_file, 'r') as f:
+        for line in f:
+            line = line.strip()
+            m = physBase_re.match(line)
+            if m:
+                return m.group(1)
+    raise Exception(f'Could not find physBase_raw in {devices_gen_h_file}')
```
Xaphiosis, line 226: -> PHYS_BASE_RAW now

spec/cspec/c/gen-config-thy.py
```diff
@@ -253,6 +271,10 @@ def write_config_thy(config_thy_path, config: Dict[str, str]):
             path.realpath(__file__))
         )
 
+        if physBase:
+            f.write('definition physBase :: machine_word where\n')
```
Xaphiosis, line 275: do we want to spit out some comment here in anticipation of future victims?

spec/cspec/c/gen-config-thy.py
```diff
@@ -278,13 +300,14 @@ def write_config_thy(config_thy_path, config: Dict[str, str]):
                 pass
 
         f.write('\n(* These definitions should only be unfolded consciously and carefully: *)\n')
-
+        if physBase:
+            f.write('hide_fact (open) physBase_def\n')
```
Xaphiosis, line 304: would it make sense to add physBase to names instead?

spec/cspec/c/gen-config-thy.py
lsf37: Happy to change. Any suggestions? Just lower case `l4v_arch`?

spec/cspec/c/gen-config-thy.py
lsf37: It's a generated file, so I don't think we want content comments in there. It'd be weird to have them in the generator. But we could add add something that points to where it's coming from, which also makes it less likely for explanations  to diverge.

spec/cspec/c/gen-config-thy.py
lsf37: Was planning to keep them separate for now, because I want to overhaul the config generation with the new json input. Eventually adding them to `names` (or whatever that will be then) here would make sense, yes.

lsf37: Actually, I'm generating `names` in this function, it's not from the config file. So I'll add your suggestion now already.

spec/cspec/c/gen-config-thy.py
lsf37: (have done that now)

spec/cspec/c/gen-config-thy.py
lsf37: Have added a comment that this is `PHYS_BASE_RAW` and included the header file as dependency in the "generated" comment at the beginning of the file.

spec/cspec/c/gen-config-thy.py
```diff
@@ -235,25 +253,27 @@ def add_defaults(config: Dict[str, str]):
...
-        )
+        f.write(header)
+
+        if physBase:
+            f.write('(* This value is PHYS_BASE_RAW in the devices_gen.h header listed above. *)\n')
+            f.write('definition physBase :: machine_word where\n')
+            f.write(f'  "physBase \\<equiv> {physBase}"\n\n')
 
-        names = []
+        names = ['physBase'] if physBase else []
```
Xaphiosis, line 276: too fast! I typed this very line, and when I tried to add a comment is said I should refresh first. Nice!

Xaphiosis (approved):

Xaphiosis:
Please hold off on merging, I want to fix up everything and then squash the fixups together. Currently testing ARM and working on X64 ... although those seem to work, or what?

lsf37:
> Edit: ok if X64 is fine, then I guess we're good to merge the whole pile of physBase wonders (overlays come later?)

Correct yes, only ARM_HYP was breaking. And agreed to merge the overlays after this one.

We'll have to merge this PR here together with the seL4 PR.

PR #595, opened 2023-03-04
ci bv: Use bv-trigger action

mbrcknl:
Use the bv-trigger action to trigger a binary verification run, rather than a called workflow.

Depends on seL4/ci-actions#258.


lsf37 (approved):
Yes, this is nicer, and will also be easier to call from other repos.

PR #596, opened 2023-03-07
simp setup for If and option

lsf37:
Terms of the form "(if P then None else Some _) = None" and all their
combinations can be simplified automatically. For the "Some" variants
we provide a safer form, e.g.:

    ((if P then Some x else None) = Some x) = P

because

    ((if P then Some x else None) = Some y) = (P /\ x = y)

adds an equation to the goal that the simplifier will pick up. That is
often wanted, but sometimes leads to non-termination.

Even the safer form can lead to non-termination if P is an equation, so
none of these are [simp] by default.

- `if_option_eq` is the safer set
- `if_option` is the less safe set that simplifies more

This resolves one of the FIXMEs int he AARCH64 proofs.

### DIFF
--- a/proof/invariant-abstract/AARCH64/ArchKHeap_AI.thy
+++ b/proof/invariant-abstract/AARCH64/ArchKHeap_AI.thy
@@ -9,9 +9,7 @@ theory ArchKHeap_AI
 imports KHeapPre_AI
 begin
 
-(* FIXME AARCH64: if_option is missing all cases for None/Some = if .. *)
-
-declare if_option_Some_eq[simp]
+declare if_option_eq[simp]
 
 context Arch begin global_naming AARCH64
 
--- a/lib/Lib.thy
+++ b/lib/Lib.thy
@@ -2522,7 +2522,25 @@ lemma if_option_None_eq:
   "((if P then Some x else None) = None) = (\<not>P)"
   by simp+
 
-lemmas if_option = if_option_None_eq if_option_Some if_option_Some2
+lemma lhs_sym_eq:
+  "(a = b) = x \<longleftrightarrow> (b = a) = x"
+  by auto
+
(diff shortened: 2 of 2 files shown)


proof/invariant-abstract/AARCH64/ArchKHeap_AI.thy
```diff
@@ -9,9 +9,7 @@ theory ArchKHeap_AI
 imports KHeapPre_AI
 begin
 
-(* FIXME AARCH64: if_option is missing all cases for None/Some = if .. *)
-
-declare if_option_Some_eq[simp]
+declare if_option_eq[simp]
```
Xaphiosis, line 12: What about the other arches? Do you want them to get this as well? I realise for this week it's a good idea to not do that, but if yes, we shouldn't forget.

lsf37: I tried that in a separate branch (in fact, I tried to enable it globally), but it broke too many existing proofs in annoying ways. So keeping it AARCH64 only is better I think.

We should probably actually consider not making it globally `[simp]` there either and use it only locally to not diverge too much. I could have a look at that separately.

Xaphiosis: That's an excellent question. I do not know which is better. AARCH64 uses a lot of option monad projections, but then again so did RISCV64 and it didn't have that declaration, so I think you're right. Not sure we want it this week, but worth taking a look.

lsf37: I'll run a session in the background and see what breaks

lsf37:
(rebased, cancelling proof tests because these all passed before and nothing relevant happened on master)

PR #598, opened 2023-03-07
resolve some of the AARCH64 Haskell FIXMEs

lsf37:
- move countTrailingZeros to Data.Word_Lib (make generic, also remove from ARM_HYP)
- defer reader monad to MCS
- eliminate isValidNativeRoot


Xaphiosis:
Commit messages:
* "Can define ..." -> "Define ...", also, is it "on" and architecture or "in" an architecture?
* "defer reader monad to MCS" should indicate this is FIXME-only change, else sounds more profound... "defer reader monad FIXMEs to MCS"?

Content of commits: thumbs up.

Xaphiosis (approved):

PR #603, opened 2023-03-08
lib: several lemmas involving `exs_valid` and the reader monad

michaelmcinerney:


lib/Monads/NonDetMonad_Sat.thy
```diff
@@ -137,4 +137,12 @@ lemma gets_exs_valid:
   "\<lbrace>(=) s\<rbrace> gets f \<exists>\<lbrace>\<lambda>r. (=) s\<rbrace>"
   by (rule exs_valid_gets)
 
+lemma exs_valid_assert_opt[wp]:
+  "\<lbrace>\<lambda>s. G \<noteq> None \<and> Q (the G) s\<rbrace> assert_opt G \<exists>\<lbrace>Q\<rbrace>"
```
lsf37, line 141: Should have realised that earlier in MM: the preferred form for option lemmas is to avoid `the`. It looks nicer now, because you can avoid the quantifier, but it can bite you later when the ~= None gets simplified separately from the rest and potentially dropped/cut off. So the canonical formulation for the precondition is `\Ex x. G = Some x /\ Q x` (I'm assuming the weaker forall/implies form is not true for exs_valid).

It'll lead to needing to use the lifting lemma for `Ex` at the application site, but we do have these.

lib/Monads/NonDetMonad_Sat.thy
```diff
@@ -137,4 +137,12 @@ lemma gets_exs_valid:
...
+  by (clarsimp simp: assert_opt_def exs_valid_def get_def assert_def bind_def' return_def)
+
+lemma gets_the_exs_valid[wp]:
+  "\<lbrace>\<lambda>s. h s \<noteq> None \<and> Q (the (h s)) s\<rbrace> gets_the h \<exists>\<lbrace>Q\<rbrace>"
```
lsf37, line 145: (same as above, should use exists)

lsf37 (approved):
Happy with that, modulo the `exists`/`the` tweak.

lib/Monads/OptionMonadWP.thy
```diff
@@ -299,7 +299,12 @@ lemma no_ofail_wp_comb2[wp_comb]:
   "\<lbrakk> no_ofail P f; no_ofail P' f \<rbrakk> \<Longrightarrow> no_ofail (\<lambda>s. P s \<and> P' s) f"
   by (simp add: no_ofail_def)
 
-
+(* Lemmas relating ovalid and valid *)
+lemma ovalid_gets_the:
+  "ovalid P f Q \<Longrightarrow> \<lbrace>P\<rbrace> gets_the f \<lbrace>Q\<rbrace>"
+  apply wpsimp
+  apply (fastforce dest: use_ovalid)
+  done
```
Xaphiosis, line 307: the silly style in this file seems to be two (or more?) lines before a "section" comment and zero after, meaning if we care there should be an extra line before and after your new "section"

PR #604, opened 2023-03-08
lib+ainvs: tweak imports for and of `MonadicRewrite`

michaelmcinerney:


lsf37:
I might have missed the end of the discussion: does removing `LemmaBucket` from `MonadicRewrite` mean that we could include `MonadicRewrite` in `Lib`, or do we still get circular dependencies if we do that?

lsf37 (approved):

Xaphiosis:
> I might have missed the end of the discussion: does removing `LemmaBucket` from `MonadicRewrite` mean that we could include `MonadicRewrite` in `Lib`, or do we still get circular dependencies if we do that?

Still circular. Corres_UL's dependencies import Lib, and since MonadicRewrite has corres-related rules, there's no avoiding that import.

Xaphiosis:
The commit message says what the commit does but nothing about the purpose, and I'm not a fan of that. You wanted to make monadic rewrite available in AInvs, and the commit message should reflect that.

Xaphiosis (approved):

PR #605, opened 2023-03-08
lib: add `notM` and `notM_wp`

michaelmcinerney:


lib/Monads/NonDetMonad.thy
```diff
@@ -606,6 +606,9 @@ notation (output)
 
 section "Combinators that have conditions with side effects"
 
+definition notM :: "('s, bool) nondet_monad \<Rightarrow> ('s, bool) nondet_monad" where
+  "notM m = do c \<leftarrow> m; return (\<not> c) od"
```
lsf37, line 610: Not sure if this is easier or more obscure, but since it's a unary function, we could just define `notM = liftM Not`

lsf37 (approved):
We should use these only when strictly necessary, but it's good to complete the operator set in any case.

Xaphiosis (approved):

lib/ExtraCorres.thy
```diff
@@ -369,4 +369,53 @@ lemma corres_whileLoop_abs:
...
+  and conc_valid: "\<lbrace>A'\<rbrace> test' \<lbrace>\<lambda>c s. (c \<longrightarrow> Q' s) \<and> (\<not> c \<longrightarrow> R' s)\<rbrace>"
+  shows "corres_underlying srel nf nf' rrel A A' (ifM test a b) (ifM test' a' b')"
+  unfolding ifM_def
+  apply (rule corres_guard_imp)
+    apply (rule corres_split_forwards[OF test abs_valid conc_valid])
+    apply clarsimp
+    apply (intro conjI impI)
+     apply (rule l)
+    apply (rule r)
+   apply fastforce
```
lsf37, line 389: Could this be a `(fastforce intro: l r)`? (or `intro!`)

michaelmcinerney: After the `corres_split_forwards` step, the rest of the subgoals can be solved with `fastforce intro: assms`. Is that best?

lib/ExtraCorres.thy
```diff
@@ -369,4 +369,53 @@ lemma corres_whileLoop_abs:
...
+  apply fastforce
+  done
+
+lemma orM_corres:
+  "\<lbrakk>corres_underlying srel nf nf' (=) A A' a a'; corres_underlying srel nf nf' (=) R R' b b';
+    \<lbrace>A\<rbrace> a \<lbrace>\<lambda>c s. \<not> c \<longrightarrow> R s\<rbrace>; \<lbrace>A'\<rbrace> a' \<lbrace>\<lambda>c s. \<not> c \<longrightarrow> R' s\<rbrace>\<rbrakk>
+   \<Longrightarrow> corres_underlying srel nf nf' (=) A A' (orM a b) (orM a' b')"
```
lsf37, line 396: interesting that we can see in the corres rule that `orM` is not commutative

lsf37 (approved):
Still happy with it.

lib/ExtraCorres.thy
lsf37: Yes, that looks nicer.

PR #606, opened 2023-03-08
crefine: simp rules for true and false

corlewis:
These rules allow the simplifier to solve almost all existing goals that involve the C constants `true` and `false`, without unfolding their definitions.

Apologies for making this PR hard to review, the relevant rules are in `Wellformed_C`. All of the other changes are removing definition unfolds that are now unnecessary, and fixing style and indentation in touched lemmas.

There almost definitely could be similar rules for `to_bool` and `from_bool` that would further improve the automation, but I don't think that that is worth exploring now.

### DIFF
--- a/proof/crefine/ARM/Arch_C.thy
+++ b/proof/crefine/ARM/Arch_C.thy
@@ -54,7 +54,7 @@ lemma performPageTableInvocationUnmap_ccorres:
          apply wp
         apply (simp del: Collect_const)
         apply (vcg exspec=unmapPageTable_modifies)
-       apply (simp add: to_bool_def)
+       apply simp
        apply (rule ccorres_return_Skip')
       apply (simp add: cap_get_tag_isCap_ArchObject[symmetric])
       apply (clarsimp simp: cap_lift_page_table_cap cap_to_H_def
@@ -796,7 +796,7 @@ lemma decodeARMPageTableInvocation_ccorres:
             apply vcg
            apply (rule conseqPre, vcg)
            apply (clarsimp simp: throwError_def return_def syscall_error_rel_def
-                                 syscall_error_to_H_cases exception_defs false_def)
+                                 syscall_error_to_H_cases exception_defs)
            apply (erule lookup_failure_rel_fault_lift[rotated])
            apply (simp add: exception_defs)
           apply simp
@@ -889,8 +889,8 @@ lemma checkVPAlignment_spec:
   apply (clarsimp simp: mask_eq_iff_w2p word_size)
   apply (rule conjI)
    apply (simp add: pageBitsForSize_def split: vmpage_size.split)
-  apply (simp add: from_bool_def vmsz_aligned'_def is_aligned_mask
-                   mask_def split: if_split)
+  apply (simp add: vmsz_aligned'_def is_aligned_mask mask_def
+            split: if_split)
   done
 
 definition
@@ -1032,9 +1032,9 @@ lemma createSafeMappingEntries_PDE_ccorres:
     apply (clarsimp simp: pde_get_tag_alt cpde_relation_pde_case
                           pde_tag_defs fst_throwError_returnOk
                           pde_range_relation_def ptr_range_to_list_def
-                          exception_defs isRight_def from_bool_def[where b=True]
+                          exception_defs isRight_def
                           syscall_error_rel_def syscall_error_to_H_cases)
-    apply (clarsimp simp: cpde_relation_def true_def false_def)
+    apply (clarsimp simp: cpde_relation_def)
    apply (rule ccorres_Cond_rhs)
     apply (simp del: Collect_const)
     apply (rule ccorres_rhs_assoc)+
@@ -1074,7 +1074,7 @@ lemma createSafeMappingEntries_PDE_ccorres:
                     apply (rule allI, rule conseqPre, vcg)
                     apply (clarsimp simp: if_1_0_0 return_def typ_heap_simps Let_def)
                     apply (simp add: isPageTablePDE_def isSectionPDE_def
-                                     cpde_relation_pde_case from_bool_def)
+                                     cpde_relation_pde_case)
                     apply (intro impI conjI disjCI2, simp_all add: array_assertion_shrink_right)[1]
                     apply (clarsimp simp: pde_tag_defs  split: if_split bool.split)
                     apply (frule pde_pde_section_size_0_1[simplified pde_tag_defs, simplified], simp)
@@ -1142,7 +1142,7 @@ lemma createSafeMappingEntries_PDE_ccorres:
     ARMSectionBits_def word_0_sle_from_less
     pageBits_def)
   apply (rule conjI)
-   apply (simp add: cpde_relation_def true_def false_def)
+   apply (simp add: cpde_relation_def)
   apply (simp add: upt_def split: if_split)
   done
 
@@ -1274,7 +1274,7 @@ lemma createSafeMappingEntries_PTE_ccorres:
          apply vcg
         apply (rule conseqPre, vcg)
         apply (clarsimp simp: fst_throwError_returnOk syscall_error_to_H_cases
-                              syscall_error_rel_def exception_defs false_def)
+                              syscall_error_rel_def exception_defs)
         apply (erule lookup_failure_rel_fault_lift[rotated])
         apply (simp add: exception_defs)
        apply simp
@@ -1372,7 +1372,7 @@ lemma createSafeMappingEntries_PTE_ccorres:
          apply vcg
         apply (rule conseqPre, vcg)
         apply (clarsimp simp: fst_throwError_returnOk syscall_error_rel_def
-                              syscall_error_to_H_cases exception_defs false_def)
+                              syscall_error_to_H_cases exception_defs)
         apply (erule lookup_failure_rel_fault_lift[rotated])
         apply (simp add: exception_defs)
        apply (wp injection_wp[OF refl])
@@ -1391,7 +1391,7 @@ lemma createSafeMappingEntries_PTE_ccorres:
                         from_bool_mask_simp[unfolded mask_def, simplified])
   apply (clarsimp simp: typ_heap_simps pte_range_relation_def
                         ptr_range_to_list_def upto_enum_word)
-  apply (simp add: cpte_relation_def true_def false_def pte_tag_defs)
+  apply (simp add: cpte_relation_def pte_tag_defs)
   using pte_get_tag_exhaust
   apply (auto simp: vmsz_aligned'_def largePagePTEOffsets_def pteBits_def
                     upt_def upto_enum_def upto_enum_step_def)[1]
@@ -1445,7 +1445,7 @@ lemma pteCheckIfMapped_ccorres:
   apply clarsimp
   apply (rule conseqPre, vcg)
   apply (clarsimp simp: typ_heap_simps' return_def)
-  apply (case_tac rv, simp_all add: to_bool_def isInvalidPTE_def pte_tag_defs pte_pte_invalid_def
+  apply (case_tac rv, simp_all add: isInvalidPTE_def pte_tag_defs pte_pte_invalid_def
                                     cpte_relation_def pte_pte_large_lift_def pte_get_tag_def
                                     pte_lift_def Let_def
                              split: if_split_asm)
@@ -1473,7 +1473,7 @@ lemma pdeCheckIfMapped_ccorres:
   apply clarsimp
   apply (rule conseqPre, vcg)
   apply (clarsimp simp: typ_heap_simps' return_def)
-  apply (case_tac rv, simp_all add: to_bool_def cpde_relation_invalid isInvalidPDE_def
+  apply (case_tac rv, simp_all add: cpde_relation_invalid isInvalidPDE_def
                              split: if_split)
   done
 
@@ -1557,7 +1557,7 @@ lemma performPageInvocationMapPTE_ccorres:
             apply (rule allI, rule conseqPre, vcg)
             apply (clarsimp simp:return_def)
            apply (rule wp_post_taut)
-          apply (simp add: to_bool_def)
+          apply simp
           apply (rule_tac P=\<top> and P'=UNIV in ccorres_from_vcg_throws)
           apply (rule allI, rule conseqPre, vcg)
           apply (clarsimp simp:return_def)
@@ -1818,12 +1818,12 @@ lemma performPageInvocationMapPDE_ccorres:
            apply (rule allI, rule conseqPre, vcg)
            apply (clarsimp simp:return_def)
           apply wp [1]
-         apply (simp add: to_bool_def)
+         apply simp
          apply (rule_tac P=\<top> and P'=UNIV in ccorres_from_vcg_throws)
          apply (rule allI, rule conseqPre, vcg)
          apply (clarsimp simp:return_def)
         apply (wp hoare_vcg_const_imp_lift) [1]
-       apply (clarsimp simp: to_bool_def)
+       apply clarsimp
        apply (rule hoare_strengthen_post)
         apply (rule_tac Q'="\<lambda>rv s. valid_pde_mappings' s
                 \<and> valid_pde_slots'2 mapping s
@@ -2173,12 +2173,12 @@ where
 
 lemma resolve_ret_rel_None[simp]:
   "resolve_ret_rel None y = (valid_C y = scast false)"
-  by (clarsimp simp: resolve_ret_rel_def o_def to_option_def false_def to_bool_def split: if_splits)
+  by (clarsimp simp: resolve_ret_rel_def o_def to_option_def to_bool_def split: if_splits)
 
 lemma resolve_ret_rel_Some:
   "\<lbrakk>valid_C y = scast true;  frameSize_C y = framesize_from_H (fst x); snd x = frameBase_C y\<rbrakk>
    \<Longrightarrow> resolve_ret_rel (Some x) y"
-  by (clarsimp simp: resolve_ret_rel_def o_def to_option_def true_def)
+  by (clarsimp simp: resolve_ret_rel_def o_def to_option_def)
 
 lemma resolveVAddr_ccorres:
   "ccorres resolve_ret_rel ret__struct_resolve_ret_C_'
@@ -2448,7 +2448,7 @@ lemma decodeARMFrameInvocation_ccorres:
         apply vcg
        apply (rule conseqPre, vcg)
        apply (clarsimp simp: throwError_def return_def syscall_error_rel_def
-                             syscall_error_to_H_cases exception_defs false_def)
+                             syscall_error_to_H_cases exception_defs)
        apply (erule lookup_failure_rel_fault_lift[rotated])
        apply (simp add: exception_defs)
       apply (wp injection_wp[OF refl])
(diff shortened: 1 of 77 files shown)


proof/crefine/ARM/Tcb_C.thy
```diff
@@ -2611,21 +2598,20 @@ lemma slotCapLongRunningDelete_ccorres:
...
+       apply (simp add: from_bool_def
                  split: bool.split)
        apply (auto simp add: longRunningDelete_def isCap_simps
                  split: capability.split)[1]
       apply simp
       apply (wp hoare_drop_imps isFinalCapability_inv)
      apply (clarsimp simp: Collect_const_mem guard_is_UNIV_def)
      apply (rename_tac rv')
-     apply (case_tac rv'; clarsimp simp: false_def true_def)
+     apply (case_tac rv'; clarsimp simp: false_def)
```
corlewis, line 2609: This `false_def` wasn't able to be trivially removed. There are some rules that are more specific than I would like that would do it, but if anyone has any better ideas then the goal is
```
SCAST(32 signed → 32) false = ret__unsigned_long_' s ⟹ ret__unsigned_long_' s = 0
```

proof/crefine/ARM/Fastpath_C.thy
```diff
@@ -1526,8 +1526,8 @@ lemma cap_reply_cap_ptr_new_np_updateCap_ccorres:
                    limited_and_simps cap_reply_cap_def
                    limited_and_simps1[OF lshift_limited_and, OF limited_and_from_bool]
                    shiftr_over_or_dist word_bw_assocs mask_def shiftl_shiftr3 word_size)
-  apply (cases m ; clarsimp)
-  apply (cases canGrant ; clarsimp)
+  apply (cases m ; clarsimp simp: true_def)
+  apply (cases canGrant ; clarsimp simp: true_def false_def)
```
corlewis, line 1530: I also wasn't able to remove these unfolds, which were previously implicitly done by `true_def` and `false_def` being added to the simp set in this file. The goal here is
```
to_bool (SCAST(32 signed → 32) true || (from_bool canGrant << 5 >> 4) && 1) ∧
canGrant = to_bool ((SCAST(32 signed → 32) true >> Suc 0) && 1 || from_bool canGrant)
```

proof/crefine/X64/Finalise_C.thy
```diff
@@ -1987,7 +1985,7 @@ lemma ccap_relation_PDPT_IsMapped:
      capPDPTIsMapped_CL (cap_pdpt_cap_lift cap) = from_bool (\<not> Option.is_none x2)"
   apply (frule cap_get_tag_isCap_unfolded_H_cap)
   apply (clarsimp simp: ccap_relation_def cap_to_H_def cap_pdpt_cap_lift_def cap_lift_def Let_def
-                        cap_tag_defs to_bool_def true_def false_def
+                        cap_tag_defs to_bool_def true_def
```
corlewis, line 1988: This is the final unfold that I wasn't able to remove. The goal is
```
¬ (cap_C.words_C cap.[0] >> 58) AND 1 = 0
    ⟹ (cap_C.words_C cap.[0] >> 58) AND 1 = signed true
```
Unfolding `true_def` makes it solvable by `word_bitwise`.

Xaphiosis:
Commit message: look at the explanation you give on this PR, then at the commit message. One has very useful info not present in the other.

proof/crefine/ARM_HYP/Wellformed_C.thy
```diff
@@ -271,8 +271,70 @@ definition
...
+lemma from_bool_vals[simp]:
+  "from_bool True = signed true"
+  "from_bool False = signed false"
+  by (simp add: from_bool_def)+
+
+lemma true_neq_false[simp]:
+  "true \<noteq> false"
+  "signed true \<noteq> signed false"
+  "unsigned true \<noteq> unsigned false"
+  by (simp add: true_def false_def)+
```
Xaphiosis, line 337: These get repeated across all `Wellformed_C.thy` files (except X64, does that not have a `Wellformed_C`?
I've got a feeling that these should not be duplicated like that. Can they go into `proof/crefine/lib` somehow? There's stuff in there that imports SR_Lemmas, so post-kernel-parse things should fit.

corlewis: Oh, and they are also repeated in X64's `Wellformed_C`. That file might have been hiding from you when scrolling through the list, it's right at the bottom.

Xaphiosis: Yup, missed it!
New file would work for me. First idea that comes to mind is `Boolean_C` which could accommodate the `to_bool_bf`, but it'll be easy to rename if people get a better idea

corlewis: > Yup, missed it! New file would work for me. First idea that comes to mind is `Boolean_C` which could accommodate the `to_bool_bf`, but it'll be easy to rename if people get a better idea

This is done now, thanks for the suggestion! This is a lot nicer with the reduced duplication.

proof/crefine/ARM_HYP/CSpace_C.thy
```diff
@@ -152,11 +153,7 @@ lemma Arch_maskCapRights_ccorres [corres]:
    apply (cases arch_cap)
        by (fastforce simp add: cap_get_tag_isCap isCap_simps  simp del: not_ex simp_thms(44))+
 
-(* FIXME: move to Wellformed_C (or move to_bool_bf out of Wellformed_C) *)
-lemma to_bool_mask_to_bool_bf:
-  "to_bool (x && 1) = to_bool_bf (x::word32)"
-  by (simp add: to_bool_bf_def to_bool_def)
-
+(* FIXME: move to Wellformed_C *)
```
corlewis, line 156: I wasn't sure if this FIXME was originally just about this generic `to_bool_bf` lemma or if it also included the following more seL4 specific lemmas. For now I've left it in, it would probably be nice to move the latter lemmas, although they're too specific to go in the new `Boolean_C` file.

proof/crefine/RISCV64/Finalise_C.thy
```diff
@@ -1732,22 +1732,6 @@ lemma cnotification_relation_udpate_arch:
   apply (safe ; case_tac "xa = p" ; clarsimp simp: option_map2_def map_option_case)
   done
 
-lemma sanitiseSetRegister_ccorres:
```
corlewis, line 1737: I removed this lemma on RISCV64 and X64 because I noticed that it was unused. I think it was originally a helper lemma in ARM_HYP (where it is still used) and was copied over accidentally.
If it's too unrelated to the rest of the PR I'm happy to either revert or to leave it as a separate commit.

Xaphiosis: separate commit at least. might also want to check other arches to see if we blundered there too

proof/crefine/ARM_HYP/CSpace_C.thy
Xaphiosis: I don't think `capAllowRead_CL` is available in `Wellformed_C` as it comes from parsing the kernel, so this FIXME no longer makes sense and I think can be removed.
Your moving both the lemma and to_bool_bf to Boolean_C I think addresses the issue sufficiently.

Xaphiosis: I thought about this some more. Leaving this FIXME for now is fine.

proof/crefine/ARM/Wellformed_C.thy
```diff
@@ -1,4 +1,5 @@
 (*
+ * Copyright 2023, Proofcraft Pty Ltd
```
Xaphiosis, line 2: I have a concern on these (`Wellformed_C`): can we claim copyright if we only remove a bunch of lemmas from a file? @lsf37 ?

proof/crefine/lib/Boolean_C.thy
```diff
@@ -0,0 +1,140 @@
+(*
+ * Copyright 2023, Proofcraft Pty Ltd
+ * Copyright 2014, General Dynamics C4 Systems
+ *
+ * SPDX-License-Identifier: GPL-2.0-only
+ *)
+
+(* Definitions and rules for C boolean constants *)
```
Xaphiosis, line 8: not sure if it's useful saying "parsed from the kernel" or something, as they're boolean constants defined in the seL4 C code itself:
```
enum _bool {
    false = 0,
    true = 1
};
```

proof/crefine/lib/Boolean_C.thy
```diff
@@ -0,0 +1,140 @@
...
+
+lemma true_neq_false[simp]:
+  "true \<noteq> false"
+  "signed true \<noteq> signed false"
+  "unsigned true \<noteq> unsigned false"
+  by (simp add: true_def false_def)+
+
+(* this is slightly weird, but the bitfield generator
+   masks everything with the expected bit length.
+   So we do that here too. *)
```
Xaphiosis, line 82: was this the original comment?

proof/crefine/ARM/Tcb_C.thy
Xaphiosis: This one looks fine to unfold false_def

proof/crefine/ARM/Fastpath_C.thy
Xaphiosis: yeah, this one is fine as well... I don't think there's much point adding rules for what happens if we right-shift `true`

proof/crefine/ARM_HYP/Wellformed_C.thy
Xaphiosis: Looks a lot nicer, and another small step forward to a more generic CRefine!
Maybe Gerwin has some crazy idea to make Boolean_C *really* generic by using locales, but I'm happy with it as is.

Xaphiosis:
Modulo two comments and that copyright thing I'm pretty happy with this PR (although I must admit I only skimmed some of the fixed theories, since you appear to be making it at least a bit better in all cases, and certainly not worse). I'll approve once we agree on the copyright thing.

proof/crefine/ARM/Wellformed_C.thy
lsf37: Let's leave it out if we end up just removing lines.

proof/crefine/lib/Boolean_C.thy
corlewis: Yeah, just copied across from `Wellformed_C`

proof/crefine/ARM/Wellformed_C.thy
corlewis: No problem, happy to get rid of it. How about all of the other files, where it's largely just removing rules passed to `simp` and restyling?

proof/crefine/lib/Boolean_C.thy
Xaphiosis: Hmm, could we tweak it to either be more content-oriented rather than feelings-of-writer-oriented and/or fitting on fewer lines?

proof/crefine/X64/Finalise_C.thy
corlewis: Good suggestion, the following solves it without unfolding `true` and without the `word_bitwise`
```
lemma and_one_neq_zero:
  "a AND 1 ≠ 0 ⟹ a AND 1 = 1"
  by (clarsimp simp: and_one_eq)
  ```

Comment (Xaphiosis) [reply proof/crefine/X64/Finalise_C.thy:1973]: Funny thing, that's probably a `<-->` equivalence.
You can also make a version for `a AND 1 /= 1 ==> a AND 1 = 0`, and shove it into lib somewhere (unless we have something like that already, or Gerwin cooked one up in his recent Word_Lib enhancements)

Comment (lsf37) [reply proof/crefine/X64/Finalise_C.thy:1973]: Word_Lib is a good place for `semiring_bit_operations` things. A lot of the things in Word_Lib are generalise to those now.

Comment (lsf37) [reply proof/crefine/X64/Finalise_C.thy:1973]: Oh, and simpset: yes, this is looking good to be `[simp]`

Comment (Xaphiosis) [reply proof/crefine/X64/Finalise_C.thy:1973]: Yeah, I thought about it for a bit, thumbs up to simpset.
Potential hiccup: if you had `a AND 1 ≠ 0` and wanted to show `a ≠ 0`, it might be easier to show that than from `a AND 1 = 1` ... but how often does that happen?
For `a AND 1` this is less of a concern.

File: proof/crefine/Move_C.thy [hunk 10]
```diff
@@ -11,6 +11,11 @@ theory Move_C
 imports Refine.Refine
 begin
 
+lemma and_one_neq_simps[simp]:
+  "a AND 1 \<noteq> 0 \<longleftrightarrow> a AND 1 = 1"
+  "a AND 1 \<noteq> 1 \<longleftrightarrow> a AND 1 = 0"
+  by (clarsimp simp: and_one_eq)+
```
corlewis, line 17: This is a temporary location for this lemma for reviewing purposes, once https://github.com/seL4/l4v/pull/587 is merged I'm planning to move it to `lib/Word_Lib/More_Bit_Ring.thy`.

proof/crefine/ARM/Arch_C.thy
```diff
@@ -3962,10 +3954,10 @@ lemma Arch_decodeInvocation_ccorres:
                   elim!: ccap_relationE split: if_split_asm)
    apply (clarsimp split: list.split)
   apply (clarsimp simp: cap_lift_asid_pool_cap cap_lift_page_directory_cap
-                        cap_to_H_def to_bool_def
-                        cap_page_directory_cap_lift_def
+                        cap_to_H_def cap_page_directory_cap_lift_def to_bool_def
                  elim!: ccap_relationE split: if_split_asm)
   done
```
corlewis, line 3959: Adding `and_one_neq_simps` did break a few proofs. It's a bit hard to diagnose what is going wrong, but as an example, in this case I think that it is due to the proof resulting in a state roughly saying
```to_bool (x && 1) ⟹ x && 1 ≠ 0```
Previously, the proof solved this by simplifying to
```⟦to_bool (x && 1); x && 1 = 1⟧ ⟹ False```
which was easily solvable with existing rules about `to_bool`.
However, now `and_one_neq_simps` is used to rewrite the goal to
```to_bool (x && 1) ⟹ x && 1 = 1```
which requires unfolding `to_bool_def` to solve.

I think I'm ok with the limited breakages that adding `and_one_neq_simps` to the simpset has caused, but I'd also be happy to rework it if people disagree. Suggestions are also welcome if anyone has ideas that could improve the automation around these constants and rules

File: proof/crefine/lib/Boolean_C.thy (hunk 8 above)
Comment (corlewis) [reply proof/crefine/lib/Boolean_C.thy]: Done

File: proof/crefine/ARM/Arch_C.thy (hunk 11 above)
Comment (Xaphiosis) [reply proof/crefine/ARM/Arch_C.thy:3959]: Well, that's somewhat unexpected. I think your observation of the breakage leans me towards also saying this breakage is ok. It wasn't the to_bool simps that were doing the heavy lifting here, it was the \noteq 0 plus clarsimp. Once that rule applies, there's not much else you can do other than unfold `to_bool`.

**Review by Xaphiosis — COMMENTED** [review]
I think after squashing, things are looking good.

**Comment by corlewis** [comment]
Alright, I've tested the latest version on ARM and assuming that the other architectures work then I think that this is ready to go. Let me know if you would prefer me to squash before looking at it again.

**Comment by lsf37** [comment]
I think squashing now is good, there's a good chance that it might be mergable after.

**Comment by corlewis** [comment]
Ok, the commits are now squashed and should be in a sensible state.

File: proof/crefine/ARM_HYP/Wellformed_C.thy (hunk 4 above)
Comment (lsf37) [reply proof/crefine/ARM_HYP/Wellformed_C.thy]: Happy with it as is.

File: proof/crefine/lib/Boolean_C.thy [hunk 12]
```diff
@@ -0,0 +1,138 @@
...
+  "to_bool (unsigned true)"
+  by (simp add: to_bool_def true_def)+
+
+lemma true_equals_simps[simp]:
+  "true = 1 \<longleftrightarrow> True"
+  "signed true = 1 \<longleftrightarrow> True"
+  "unsigned true = 1 \<longleftrightarrow> True"
+  "true = 0 \<longleftrightarrow> False"
+  "signed true = 0 \<longleftrightarrow> False"
+  "unsigned true = 0 \<longleftrightarrow> False"
```
lsf37, line 33: Not too sure about these as `[simp]` rules, because the conclusion doesn't have much the simplifier can match on. What breaks when you leave them out?

proof/crefine/lib/Boolean_C.thy
```diff
@@ -0,0 +1,138 @@
...
+  by (simp add: to_bool_def false_def)+
+
+lemma false_equals_simps[simp]:
+  "false = 0 \<longleftrightarrow> True"
+  "signed false = 0 \<longleftrightarrow> True"
+  "unsigned false = 0 \<longleftrightarrow> True"
+  "false = 1 \<longleftrightarrow> False"
+  "signed false = 1 \<longleftrightarrow> False"
+  "unsigned false = 1 \<longleftrightarrow> False"
+  by (simp add: false_def)+
```
lsf37, line 61: Same as above for `true`

proof/crefine/lib/Boolean_C.thy
```diff
@@ -0,0 +1,138 @@
...
+
+lemma false_eq_from_bool[simp]:
+  "(false = from_bool P) = (\<not> P)"
+  "(signed false = from_bool P) = (\<not> P)"
+  "(unsigned false = from_bool P) = (\<not> P)"
+  by (simp add: from_bool_def split: bool.splits)+
+
+lemma from_bool_vals[simp]:
+  "from_bool True = signed true"
+  "from_bool False = signed false"
```
lsf37, line 71: Do we also have `unsigned` versions of these? They might already be in the simp set.

lsf37 (approved):
Thanks for this massive amount of work, this has turned out quite nicely!

proof/crefine/lib/Boolean_C.thy
corlewis: These being `[simp]` rules was basically the starting point of this PR. The original purpose of this work was to make it so that simple goals involving `true` and `false` are handled completely automatically, so that people who don't know about them don't need to spend time figuring out that they are special. (To be even more precise, I was one of those people recently and spent too long trying to work out why a goal wasn't being solved before realising that I needed to give it `false_def`.)

I'm waiting for things to rebuild to confirm, but I would expect that putting them in the set as needed will require reverting everywhere that I've removed `true_def` and `false_def` and replacing them with `true_equals_simps` and `false_equals_simps`.

proof/crefine/lib/Boolean_C.thy
corlewis: These were moved from `IpcCancel_C` now that we have a central location for them to go in. I don't think `unsigned` versions of them exist.

proof/crefine/lib/Boolean_C.thy
lsf37: Sorry, all good, I once again misread `longleftrightarrow` for `longrightarrow`. Equivalences make complete sense here.

proof/crefine/lib/Boolean_C.thy
lsf37: Would the unsigned version just be `from_bool True = true`? In any case, it's fine to merge as is, this is something we could add in another PR if we wanted to.

corlewis: Hmm, I've now managed to confuse myself with this. I don't think that we want `from_bool True = true`, because that would only apply for `from_bool :: bool ⇒ 32 signed word`.

In some sense I don't think that it really matters, in that we're basically picking a form to simplify the `from_bool` into and the `signed`/`unsigned` is mostly to get the types right.

corlewis: Thinking about it a bit more, it's possible that quite a few of the lemmas in this file could have the repetition around `unsigned` and `signed` reduced. Because `true` has a  specific value and type, we can show that `signed true = unsigned true` holds and pick whichever form we prefer.

lsf37: Agreed, I think this is Ok.

Xaphiosis (approved):
suggest commit mesage tweak to have `word_lib+crefine` instead of `crefine+word_lib`, otherwise all good. Thanks for doing this!

PR #607, opened 2023-03-08
cspec: make remaining relative paths absolute

lsf37:
Previous commit 1a7eb921110da missed two instances.

(We were still getting BV artifact export errors that are masked in the PR tests, because they don't usually need to re-run SimplExportAndRefine)


Xaphiosis (approved):
Might be worth mentioning this is overlay-related in the commit message. This is over-caution on my side, but hashes changed when we moved from bitbucket->github, and to my mind, everything is temporary.

PR #608, opened 2023-03-09
aarch64 ainvs: avoid global [simp] for if_option_eq

lsf37:
Keep simp set more in line with other architectures.

Using `if_option_eq` or `if_option` locally turns out to work fine and a relatively small change.


Xaphiosis (approved):
That was fast

PR #609, opened 2023-03-10
use physBase definition from C config

lsf37:
- use physBase definition from C header in Platform.thy
- update proofs accordingly


lsf37:
(rebased)

Xaphiosis (approved):

PR #610, opened 2023-03-13
cspec: Allow kernel build export from older l4v

mbrcknl:
In `export-kernel-builds.py`, we allow using two different l4 checkouts:
- one in which to look for l4v kernel build artifacts, and
- one to provide the kernel build and export machinery.

This is occasionally useful for binary verification (BV), since it allows us to obtain kernel build artifacts in the format expected by BV, from older l4v versions that lack the export machinery.

Additionally, in the l4v kernel build (`kernel.mk`), we change the way device tree overlays are handled, to avoid the need to create files in the l4v checkout providing the build export machinery. This makes it possible to build using read-only l4v checkouts.

### DIFF
--- a/spec/cspec/c/export-kernel-builds.py
+++ b/spec/cspec/c/export-kernel-builds.py
@@ -46,29 +46,20 @@ import sys
 
 from pathlib import Path
 from tempfile import TemporaryDirectory
-from typing import List, NamedTuple, Optional
+from typing import NamedTuple, Optional, Sequence
 
 
-class L4vPaths(NamedTuple):
-    kernel_mk: Path
-    c_pp: Path
-    c_functions: Path
-
-
-def get_l4v_paths(l4v_arch: str) -> L4vPaths:
-    l4v_c = Path(__file__).resolve().parent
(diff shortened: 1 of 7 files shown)


mbrcknl:
@lsf37 Let me know what you think of the overlay changes. I realised that I don't actually need these, since the changes to `export-kernel-builds.py` are enough for me to be able to perform exports from older l4v checkouts. I thought I'd post them anyway, in case you like them, but I'm happy to take them out if you don't.

lsf37:
As discussed in person, I'm happy with the changes to the export script, but I think it'd be good to avoid complicating `kernel.mk` more than necessary with additional conditionals.

If that is still useful, we could add a rule that creates an empty default overlay (using `touch`) if none exists -- this should then work with older versions of `l4v` as well.

lsf37:
@mbrcknl should we close this PR and archive it for now? Happy for it to be re-opened when we get back to it, but currently it looks like we're not going to make progress on it.

mbrcknl:
Sure, I'll close it for now.

PR #612, opened 2023-03-16
ci: Fix broken external workflow

### LINKED ISSUE #611: check scheduled workflow runs
It seems like we overlooked something when merging the binary verification trigger:

```
The workflow is not valid. .github/workflows/external.yml (Line: 47, Col: 12): Job 'binary-verification' depends on unknown job 'proofs'.
```

mbrcknl:
A previous commit added a new job which depended on a job that didn't exist. We rename the `all` job to `proofs` for consistency with other workflows.

Fixes #611.


lsf37 (approved):
:+1:

PR #614, opened 2023-03-21
tests: add kernel-config session

lsf37:
Make the C kernel config extraction visible as a separate test session in `run_tests` so that `run_tests` can do concurrency control for it.


Xaphiosis (approved):
Useful, thanks!

PR #616, opened 2023-03-26
Rename `bind_assoc_reverse` to `bind_assoc_return_reverse`

michaelmcinerney:
This also improves the style of this lemma


lsf37 (approved):

corlewis (approved):

Xaphiosis (commented):
Checked the style with @lsf37 ... this is fine to use in lib/ where the extra newline would be a waste of space and there's instances of this style already, but also worth pointing out that in specs we don't do this (the `do` ends up on a different line from the first statement within it unless the `od` is on the same line).

PR #617, opened 2023-03-27
physBase abstraction

corlewis:
Refactor the proofs such that they do not unfold the value of these constants, but
instead only work with characteristic properties. Prove these characteristic properties once
per architecture in a separate theory file. If this file passes, the rest of the proofs are now guaranteed
to work for that architecture.

### DIFF
--- a/proof/crefine/ARM/ArchMove_C.thy
+++ b/proof/crefine/ARM/ArchMove_C.thy
@@ -1,4 +1,5 @@
 (*
+ * Copyright 2023, Proofcraft Pty Ltd
  * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
  * Copyright 2014, General Dynamics C4 Systems
  *
@@ -234,6 +235,22 @@ crunches insertNewCap, Arch_createNewCaps, threadSet, Arch.createObject, setThre
   (wp: crunch_wps setObject_ksPSpace_only
    simp: unless_def updateObject_default_def crunch_simps
    ignore_del: preemptionPoint)
+
+lemma addrFromPPtr_mask[simplified ARM.pageBitsForSize_simps]:
+  "n \<le> pageBitsForSize ARMSuperSection
+   \<Longrightarrow> addrFromPPtr ptr && mask n = ptr && mask n"
+  apply (simp add: addrFromPPtr_def)
+  apply (prop_tac "pptrBaseOffset AND mask n = 0")
+   apply (rule mask_zero[OF is_aligned_weaken[OF pptrBaseOffset_aligned]], simp)
+  apply (simp flip: mask_eqs(8))
+  done
+
+(* this could be done as
+   lemmas addrFromPPtr_mask_5 = addrFromPPtr_mask[where n=5, simplified]
+   but that wouldn't give a sanity check of the n \<le> ... assumption  disappearing *)
+lemma addrFromPPtr_mask_5:
+  "addrFromPPtr ptr && mask 5 = ptr && mask 5"
+  by (rule addrFromPPtr_mask[where n=5, simplified])
 
 end
 
--- a/proof/crefine/ARM_HYP/ArchMove_C.thy
+++ b/proof/crefine/ARM_HYP/ArchMove_C.thy
@@ -1,4 +1,5 @@
 (*
+ * Copyright 2023, Proofcraft Pty Ltd
  * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
  * Copyright 2014, General Dynamics C4 Systems
  *
@@ -633,6 +634,17 @@ crunches insertNewCap, Arch_createNewCaps, threadSet, Arch.createObject, setThre
   (wp: crunch_wps setObject_ksPSpace_only
    simp: unless_def updateObject_default_def crunch_simps
    ignore_del: preemptionPoint)
+
+(* this could be done as
+   lemmas addrFromPPtr_mask_6 = addrFromPPtr_mask[where n=6, simplified]
+   but that wouldn't give a sanity check of the n \<le> ... assumption  disappearing *)
+lemma addrFromPPtr_mask_6:
+  "addrFromPPtr ptr && mask 6 = ptr && mask 6"
+  by (rule addrFromPPtr_mask[where n=6, simplified])
+
+lemma ptrFromPAddr_mask_6:
+  "ptrFromPAddr ps && mask 6 = ps && mask 6"
+  by (rule ptrFromPAddr_mask[where n=6, simplified])
 
 end
 
--- a/proof/infoflow/ARM/Example_Valid_State.thy
+++ b/proof/infoflow/ARM/Example_Valid_State.thy
@@ -1,4 +1,5 @@
 (*
+ * Copyright 2023, Proofcraft Pty Ltd
  * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
  *
  * SPDX-License-Identifier: GPL-2.0-only
@@ -205,7 +206,8 @@ definition "High_pt_ptr = kernel_base + 0xC00"
 
 (* init_globals_frame \<equiv> {kernel_base + 0x5000,... kernel_base + 0x5FFF} *)
 
-definition "shared_page_ptr = kernel_base + 0x6000"
+definition "shared_page_ptr_virt = kernel_base + 0x6000"
+definition "shared_page_ptr_phys = addrFromPPtr shared_page_ptr_virt"
 
 definition "Low_pd_ptr = kernel_base + 0x20000"
 definition "High_pd_ptr = kernel_base + 0x24000"
@@ -233,14 +235,14 @@ lemmas s0_ptr_defs =
   Low_pd_ptr_def High_pd_ptr_def Low_pt_ptr_def High_pt_ptr_def Low_tcb_ptr_def
   High_tcb_ptr_def idle_tcb_ptr_def timer_irq_def Low_prio_def High_prio_def Low_time_slice_def
   Low_domain_def High_domain_def init_irq_node_ptr_def init_globals_frame_def init_global_pd_def
-  kernel_base_def shared_page_ptr_def
+  kernel_base_def shared_page_ptr_virt_def
 
 (* Distinctness proof of kernel pointers. *)
 
 distinct ptrs_distinct [simp]:
   Low_tcb_ptr High_tcb_ptr idle_tcb_ptr
   Low_pt_ptr High_pt_ptr
-  shared_page_ptr ntfn_ptr
+  shared_page_ptr_virt ntfn_ptr
   Low_pd_ptr High_pd_ptr
   Low_cnode_ptr High_cnode_ptr Silc_cnode_ptr irq_cnode_ptr
   init_globals_frame init_global_pd
@@ -507,7 +509,7 @@ definition
   Low_pt' :: "word8 \<Rightarrow> pte "
 where
   "Low_pt' \<equiv> (\<lambda>_. InvalidPTE)
-            (0 := SmallPagePTE shared_page_ptr {} vm_read_write)"
+            (0 := SmallPagePTE shared_page_ptr_phys {} vm_read_write)"
 
 definition
   Low_pt :: kernel_object
@@ -542,7 +544,7 @@ definition
 where
   "High_pt' \<equiv>
     (\<lambda>_. InvalidPTE)
-     (0 := SmallPagePTE shared_page_ptr {} vm_read_only)"
+     (0 := SmallPagePTE shared_page_ptr_phys {} vm_read_only)"
 
 
 definition
@@ -872,7 +874,7 @@ definition
   Sys1AgentMap :: "(auth_graph_label subject_label) agent_map"
 where
   "Sys1AgentMap \<equiv>
-   (\<lambda>p. if ptrFromPAddr shared_page_ptr \<le> p \<and> p < ptrFromPAddr shared_page_ptr + 0x1000
+   (\<lambda>p. if p \<in> ptr_range shared_page_ptr_virt pageBits
           then partition_label Low else partition_label IRQ0)
             \<comment> \<open>set the range of the shared_page to Low, default everything else to IRQ0\<close>
      (Low_cnode_ptr := partition_label Low,
@@ -901,12 +903,11 @@ lemma Sys1AgentMap_simps:
       "Sys1AgentMap Low_tcb_ptr = partition_label Low"
       "Sys1AgentMap High_tcb_ptr = partition_label High"
       "Sys1AgentMap idle_tcb_ptr = partition_label Low"
-      "\<And>p. \<lbrakk>ptrFromPAddr shared_page_ptr \<le> p; p < ptrFromPAddr shared_page_ptr + 0x1000\<rbrakk>
+      "\<And>p. p \<in> ptr_range shared_page_ptr_virt pageBits
           \<Longrightarrow> Sys1AgentMap p = partition_label Low"
   unfolding Sys1AgentMap_def
   apply simp_all
-  by (auto simp: ptrFromPAddr_def pptrBaseOffset_def
-          pptrBase_def Kernel_Config.physBase_def s0_ptr_defs)
+  by (auto simp: s0_ptr_defs ptr_range_def pageBits_def)
 
 definition
   Sys1ASIDMap :: "(auth_graph_label subject_label) agent_asid_map"
@@ -998,8 +999,7 @@ lemma thread_bounds_of_state_s0:
 
 lemma Sys1_wellformed':
   "policy_wellformed (pasPolicy Sys1PAS) False irqs x"
-  apply (clarsimp simp: Sys1PAS_def Sys1AgentMap_simps policy_wellformed_def
-                        Sys1AuthGraph_def)
+  apply (clarsimp simp: Sys1PAS_def policy_wellformed_def Sys1AuthGraph_def)
   done
 
 corollary Sys1_wellformed:
@@ -1009,8 +1009,7 @@ corollary Sys1_wellformed:
 
 lemma Sys1_pas_wellformed:
   "pas_wellformed Sys1PAS"
-  apply (clarsimp simp: Sys1PAS_def Sys1AgentMap_simps policy_wellformed_def
-                        Sys1AuthGraph_def)
+  apply (clarsimp simp: Sys1PAS_def policy_wellformed_def Sys1AuthGraph_def)
   done
 
 lemma domains_of_state_s0[simp]:
@@ -1029,12 +1028,10 @@ lemma Sys1_pas_refined:
   apply (clarsimp simp: pas_refined_def)
   apply (intro conjI)
        apply (simp add: Sys1_pas_wellformed)
-      apply (clarsimp simp: irq_map_wellformed_aux_def s0_internal_def Sys1AgentMap_simps Sys1PAS_def)
+      apply (clarsimp simp: irq_map_wellformed_aux_def s0_internal_def Sys1PAS_def)
       apply (clarsimp simp: Sys1AgentMap_def)
-      apply (clarsimp simp: ptrFromPAddr_def s0_ptr_defs cte_level_bits_def
-                            pptrBaseOffset_def pptrBase_def Kernel_Config.physBase_def)
-      apply (drule le_less_trans[OF irq_node_offs_min[simplified s0_ptr_defs cte_level_bits_def, simplified]])
-      apply simp
+      apply (clarsimp simp: s0_ptr_defs ptr_range_def pageBits_def cte_level_bits_def)
+      apply word_bitwise
      apply (clarsimp simp: tcb_domain_map_wellformed_aux_def
                            Sys1PAS_def Sys1AgentMap_def
                            default_domain_def minBound_word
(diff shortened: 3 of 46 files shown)


spec/abstract/RISCV64/Init_A.thy
```diff
@@ -37,12 +37,20 @@ definition canonical_user :: "vspace_ref" where
...
    each. This is not the layout the real kernel uses, but we are only trying to show that
-   the invariants are consistent. *)
+   the invariants are consistent.
+
+   The values we pick here for the size of these regions constrain pptr_base and kernel_elf_base
+   in real kernel configurations, so we pick relatively small values that are still large enough
+   to work for the example and that are reasonable lower bounds for real platforms:
+   4M physical memory (1 << 22) and one page (1 << pageBits) for the kernel elf region. *)
+definition kernel_window_bits :: nat where
+  "kernel_window_bits \<equiv> 22"
```
corlewis, line 47: This constant could probably be abstracted in a nicer way, but for now I just replaced all the relevant instances of `1 << 20` that I found with `1 << kernel_window_bits` and then unfolded `kernel_window_bits` whenever needed.

spec/abstract/RISCV64/Init_A.thy
```diff
@@ -54,18 +62,25 @@ definition init_arch_state :: arch_state
...
+  where
+  "toplevel_bits = pt_bits_left max_pt_level"
+
+definition elf_index :: pt_index
+  where
+  "elf_index = ucast (pt_index max_pt_level kernel_elf_base)"
 
-(* {pptr_base ..< pptr_base + (1 << 30)} is pt index 0x100 at max_pt_level,
-   {kernel_elf_base ..< kernel_elf_base + (1 << 20)} comes out to pt index 0x1FE.
+(* {pptr_base ..< pptr_base + (1 << 20)} is pt index 0x100 at max_pt_level,
```
corlewis, line 73: However, it looks like I missed at least one instance. I guess this comment should be changed to use `kernel_window_bits` as well.

spec/abstract/RISCV64/Init_A.thy
lsf37: I think that's good as is. We should probably add a comment that this is just an example size and chosen to be smallish but to fit at least a  `RISCVLargePage` + a minimum set of objects for the InfoFlow example.

lsf37: Ah, I should have looked into the file first :)

lsf37: Most of it is there, but the forward reference to infoflow and the constraints from that would be good to add.

proof/crefine/ARM/ArchMove_C.thy
```diff
@@ -235,6 +235,22 @@ crunches insertNewCap, Arch_createNewCaps, threadSet, Arch.createObject, setThre
...
+   apply (rule mask_zero[OF is_aligned_weaken[OF pptrBaseOffset_aligned]], simp)
+  apply (simp flip: mask_eqs(8))
+  done
+
+(* FIXME MK: this can be done as:
+   lemmas addrFromPPtr_mask_5 = addrFromPPtr_mask[where n=5, simplified]
+   but that doesn't give a sanity check of the n \<le> ... assumption  disappearing *)
+lemma addrFromPPtr_mask_5:
+  "addrFromPPtr ptr && mask 5 = ptr && mask 5"
+  by (rule addrFromPPtr_mask[where n=5, simplified])
```
lsf37, line 252: I think it's Ok to drop the fixme and just leave the rest of the comment there as explanation why it is done this way.

proof/crefine/ARM_HYP/ArchMove_C.thy
```diff
@@ -634,6 +634,38 @@ crunches insertNewCap, Arch_createNewCaps, threadSet, Arch.createObject, setThre
...
+   apply (rule mask_zero[OF is_aligned_weaken[OF pptrBaseOffset_aligned]], simp)
+  apply (simp flip: mask_eqs(7))
+  done
+
+(* FIXME MK: this can be done as:
+   lemmas addrFromPPtr_mask_6 = addrFromPPtr_mask[where n=6, simplified]
+   but that doesn't give a sanity check of the n \<le> ... assumption  disappearing *)
+lemma addrFromPPtr_mask_6:
+  "addrFromPPtr ptr && mask 6 = ptr && mask 6"
+  by (rule addrFromPPtr_mask[where n=6, simplified])
```
lsf37, line 662: Same as the `addrFromPPtr_mask_5` I'd drop the FIXME here. Also for the following `ptrFromPAddr_mask_6`

proof/crefine/ARM_HYP/ArchMove_C.thy
```diff
@@ -634,6 +634,38 @@ crunches insertNewCap, Arch_createNewCaps, threadSet, Arch.createObject, setThre
    simp: unless_def updateObject_default_def crunch_simps
    ignore_del: preemptionPoint)
 
+(* FIXME MK: consider move *)
+lemma addrFromPPtr_mask[simplified ARM_HYP.pageBitsForSize_simps]:
+  "n \<le> pageBitsForSize ARMSuperSection
+   \<Longrightarrow> addrFromPPtr ptr && mask n = ptr && mask n"
```
lsf37, line 640: Agreed, this and the next one should probably go relatively high up in AInvs somewhere. Maybe ArchInvariants_AI.

proof/infoflow/ARM/Example_Valid_State.thy
```diff
@@ -1635,14 +1621,18 @@ lemma valid_arch_caps_s0[simp]:
   apply (drule s0_caps_of_state)+
   by auto
 
+\<comment> \<open>FIXME MK: move, or kernel_base could be changed to be an abbreviation of pptrBase\<close>
+lemma kernel_base_aligned_pageBits:
+  "is_aligned kernel_base pageBits"
+  by (simp add: is_aligned_def kernel_base_def pageBits_def)
```
lsf37, line 1627: @corlewis, can you check how often we mention `kernel_base`? It'd be good to get rid of it completely and replace globally by `pptrBase`. It's from ages back when terminology hadn't settled yet and ASpec/Haskell/C all had different names for the same thing.

proof/infoflow/ARM/Example_Valid_State.thy
```diff
@@ -205,7 +206,8 @@ definition "High_pt_ptr = kernel_base + 0xC00"
 
 (* init_globals_frame \<equiv> {kernel_base + 0x5000,... kernel_base + 0x5FFF} *)
 
-definition "shared_page_ptr = kernel_base + 0x6000"
+definition "shared_page_ptr_virt = kernel_base + 0x6000"
+definition "shared_page_ptr_phys = addrFromPPtr shared_page_ptr_virt"
```
lsf37, line 210: I like how the proofs have become simpler with this.

proof/infoflow/ARM/Example_Valid_State.thy
```diff
@@ -1,5 +1,6 @@
 (*
  * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+ * Copyright 2023, Proofcraft Pty Ltd
```
lsf37, line 3: Usually I put the most recent ones at the top so that we get reverse chronological order. I'm not too hung up on the particular order, but probably better to do it consistently.

proof/invariant-abstract/RISCV64/ArchKernelInit_AI.thy
```diff
@@ -237,81 +275,197 @@ lemma kernel_mapping_slots_0x100[simp]:
...
+                simp: kernelELFPAddrBase_def pageBits_def is_aligned_def)
+
+(* FIXME MK: move *)
+lemma aligned_mask_le_mask_minus:
+  "\<lbrakk> is_aligned x m; m \<le> n; n < LENGTH('a)\<rbrakk> \<Longrightarrow> x && mask n \<le> mask n - mask m" for x :: "'a::len word"
+  by (metis and_mask_less' is_aligned_after_mask is_aligned_neg_mask_eq'
+           mask_2pm1 mask_sub neg_mask_mono_le word_less_sub_le)
+
+(* FIXME MK: rename, move *)
+lemma aligned_mask_plus_bounded:
```
lsf37, line 316: have done the rename already, but still should be moved to the Word_LIb staging area

proof/infoflow/ARM/Example_Valid_State.thy
corlewis: It's used enough that it might be a bit too much to replace it as part of this PR. Across ARM, ARM_HYP and X64 there's 39 instances of it in `spec` and 391 in `proof`.

Xaphiosis: I hit the same issue when I was doing kernel constant renames to match C (see "arm+arm-hyp: kernelBase and physMappingOffset renames" commit on l4v). `kernel_base` should definitely go, but I agree to leave it out of this PR.

- [ ] create an issue for kernel_base rename to pptrBase

proof/infoflow/RISCV64/Example_Valid_State.thy
```diff
@@ -1,5 +1,6 @@
 (*
  * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+ * Copyright 2023, Proofcraft Pty Ltd
```
Xaphiosis, line 2: copyright order again

proof/infoflow/RISCV64/Example_Valid_State.thy
```diff
@@ -1756,15 +1727,15 @@ lemma equal_kernel_mappings_s0[simp]:
...
+                        kernel_mapping_slots_def s0_ptr_defs misc elf_index_value)+
   done
 
 lemma valid_asid_map_s0[simp]:
   "valid_asid_map s0_internal"
   by (clarsimp simp: valid_asid_map_def s0_internal_def arch_state0_def)
 
 lemma valid_global_pd_mappings_s0_helper:
-  "\<lbrakk> pptr_base \<le> vref; vref < pptr_base + 0x40000000 \<rbrakk>
+  "\<lbrakk> pptr_base \<le> vref; vref < pptr_base + (1<<kernel_window_bits) \<rbrakk>
```
Xaphiosis, line 1738: spaces around `<<`?

proof/infoflow/RISCV64/Example_Valid_State.thy
```diff
@@ -1099,13 +1102,18 @@ lemma vs_lookup_s0_SomeD:
                      pool_for_asid_s0 asid_pools_of_s0 vspace_for_pool_def
               split: if_splits)+
 
+lemma pt_bits_left_max_minus_1_pageBitsForSize:
+  "pt_bits_left (max_pt_level - 1) = pageBitsForSize RISCVLargePage"
+  apply (clarsimp simp: pt_bits_left_def max_pt_level_def2)
+  done
```
Xaphiosis, line 1108: should this get moved somewhere higher?

proof/invariant-abstract/ARM/ArchAcc_AI.thy
```diff
@@ -2543,38 +2543,59 @@ lemma create_mapping_entries_valid_slots [wp]:
...
-                   pptrBase_def pageBits_def)
-  apply (erule is_aligned_weaken[rotated])
-  apply (simp add: is_aligned_def)
-  done
+(* FIXME MK: staging area for physBase abstraction *)
+
+(* FIXME MK: candidate interface lemma  *)
+lemma physBase_aligned[simplified pageBitsForSize_simps]:
+  "is_aligned physBase (pageBitsForSize ARMSuperSection)"
+  by (simp add: is_aligned_def Kernel_Config.physBase_def)
```
Xaphiosis, line 2551: This one needs to go into Kernel_Config_Lemmas, but where should the ones below go? ArchInvariants? Leave them here? ArchAcc is a strange file, I'm not sure of its origins.

proof/invariant-abstract/RISCV64/ArchKernelInit_AI.thy
```diff
@@ -208,22 +209,59 @@ lemma pt_walk_init_A_st[simp]:
...
+(* FIXME MK: move to Word_Lib *)
+lemma word_mask_and_neg_shift_eq_0:
+  "n < m \<Longrightarrow> - (1 << m) && mask n = 0"
+  by (metis NOT_mask_AND_mask add.inverse_neutral leD less_imp_le mask_AND_NOT_mask mask_eqs(10)
+            shiftl_1 t2n_mask_eq_if)
+
+(* Since the bottom 30 bits don't matter for 1GB pages, kernelELFPAddrBase and physBase don't have
+   any influence on the index in the page table, i.e. the virtual address kernel_elf_base is always
+   within the same 1GB page. *)
+lemma elf_index_value:
```
Xaphiosis, line 226: don't we call these `_def'`? haven't seen `_value` much

proof/invariant-abstract/RISCV64/ArchKernelInit_AI.thy
```diff
@@ -237,81 +275,197 @@ lemma kernel_mapping_slots_0x100[simp]:
...
+  by (fastforce intro: aligned_mask_plus_bounded is_page_aligned_kernelELFPAddrBase
+                simp: toplevel_bits_value bit_simps)
+
+lemma elf_window_4k:
+  "\<lbrakk> kernel_elf_base \<le> vref; vref < kernel_elf_base + (1 << pageBits) \<rbrakk> \<Longrightarrow>
+    table_index (pt_slot_offset max_pt_level riscv_global_pt_ptr vref) = elf_index"
+  using is_page_aligned_kernelELFPAddrBase
+  apply (simp add: table_index_riscv_global_pt_ptr elf_index_value toplevel_bits_value)
+  apply (unfold kernel_elf_base_def kernelELFBase_def bit_simps)
+  apply (simp add: add_ac)
```
Xaphiosis, line 332: unfold + simp can be combined into one simp?

proof/invariant-abstract/RISCV64/ArchKernelInit_AI.thy
```diff
@@ -237,81 +275,197 @@ lemma kernel_mapping_slots_0x100[simp]:
...
+  "0x100 \<le> elf_index"
+  by (simp add: elf_index_value)
+
+lemma kernel_mapping_slots_elf_index[simp]:
+  "elf_index \<in> kernel_mapping_slots"
   by (simp add: kernel_mapping_slots_def pptr_base_def bit_simps pt_bits_left_def level_defs
-                pptrBase_def canonical_bit_def)
+                pptrBase_def canonical_bit_def leq_elf_index)
+
+lemma rewrite_eq_minus_to_plus_eq:
```
Xaphiosis, line 351: this can be moved to Lib as well

proof/invariant-abstract/RISCV64/ArchKernelInit_AI.thy
```diff
@@ -237,81 +275,197 @@ lemma kernel_mapping_slots_0x100[simp]:
...
+
+lemma is_aligned_ptrFromPAddr_kernelELFPAddrBase:
+  "is_aligned
+     (ptrFromPAddr
+       (addr_from_ppn (ucast (kernelELFPAddrBase && ~~ mask toplevel_bits >> pageBits))))
+     (pt_bits_left max_pt_level)"
+  apply (simp add: addr_from_ppn_def ptrFromPAddr_def addrFromPPtr_def elf_index_value)
+  apply (simp add: pt_bits_left_def bit_simps level_defs ucast_ucast_mask toplevel_bits_value)
+  apply (simp add: pptrBase_def pptrBaseOffset_def paddrBase_def canonical_bit_def)
+  apply (rule is_aligned_add)
```
Xaphiosis, line 369: [rotated] can drop the prefer 2

spec/abstract/RISCV64/Init_A.thy
```diff
@@ -54,18 +62,25 @@ definition init_arch_state :: arch_state
      riscv_kernel_vspace = init_vspace_uses
    \<rparr>"
 
+definition toplevel_bits :: nat
+  where
+  "toplevel_bits = pt_bits_left max_pt_level"
+
+definition elf_index :: pt_index
+  where
```
Xaphiosis, line 70: for this and prev definition, please move `where` to prev line

spec/machine/RISCV64/Platform.thy
```diff
@@ -109,10 +109,13 @@ definition kernelELFPAddrBase :: machine_word
   where
   "kernelELFPAddrBase = physBase + 0x4000000"
 
+definition pptrTop :: machine_word
+  where
+  "pptrTop \<equiv> - (1 << 31)"
```
Xaphiosis, line 114: we should at some stage do a restyle of Platform.thy on all arches

- [ ] add issue

proof/crefine/ARM/ArchMove_C.thy
corlewis: It would be a bit over the top, but if we want to enforce that the assumption disappears without explicitly writing out the lemma we could do something like https://github.com/seL4/l4v/blob/rt/proof/refine/ARM/Invariants_H.thy#L3013

spec/abstract/RISCV64/Init_A.thy
lsf37: No, please don't, that breaks the style of the rest of the file. We should restyle the entire file which currently is consistent or we should keep it consistent.

Xaphiosis: Ah, my bad. Somehow missed the file this was in. As Gerwin said, ignore me.

proof/invariant-abstract/RISCV64/ArchKernelInit_AI.thy
corlewis: The best I managed without spending a lot more time on this is
```
  using is_page_aligned_kernelELFPAddrBase
  apply (simp add: table_index_riscv_global_pt_ptr elf_index_value toplevel_bits_value
                   kernel_elf_base_def kernelELFBase_def)
  apply (simp add: bit_simps add_ac)
```
The simplifier makes things more complicated if `bit_simps` is used too early.

proof/infoflow/RISCV64/Example_Valid_State.thy
corlewis: It's pretty specific, but I guess it would be useful if we have other situations where the different formulations meet. Any suggestions for a sensible location for it?

proof/invariant-abstract/RISCV64/ArchKernelInit_AI.thy
corlewis: It definitely should be moved, but I'm struggling to find a good location. So far the best I've got is either sticking it in `Lib.Lib` or in `Word_Lib.Word_Lemmas_Internal`. An alternative might be `Word_Lib.More_Bit_Ring` from https://github.com/seL4/l4v/pull/587, once that is merged.

proof/infoflow/RISCV64/Example_Valid_State.thy
corlewis: Sure, I'll change any of these that have crept in. I don't think we're completely consistent in putting the spaces in but it's definitely the dominant style.

proof/invariant-abstract/RISCV64/ArchKernelInit_AI.thy
lsf37: That sounds familiar, I think I explicitly tried to single-step the unfold because bit_simps and add_ac were interacting somehow

proof/invariant-abstract/RISCV64/ArchKernelInit_AI.thy
lsf37: Under `_def'` I would usually expect a different abstract representation and I wanted to make a distinction in the name to indicate you get a number when you use this (which we usually don't do). I thought I had seen `_val` or `_value` as a thing before, but it does indeed not seem to be mentioned anywhere.

If it's confusing, I'm fine with renaming to `def'`

proof/invariant-abstract/ARM/ArchAcc_AI.thy
lsf37: (recording offline discussion: should go into ArchInvariants for now)

proof/infoflow/RISCV64/Example_Valid_State.thy
lsf37: I would leave it in the example -- the lemma people should be using in abstract proofs is already available higher up, and if we move the concrete one then people might feel the urge to complete the set and then we'd have hard coded the number of levels and relationships instead of keeping them generic.

proof/invariant-abstract/RISCV64/ArchKernelInit_AI.thy
lsf37: It should not go into Word_Lib, it's not really much to do with words. I would actually be happy to keep it here, because it's just fixing a specific form so that a specific rewrite goes the way I need it to. The general case of such equations where the rest is solvable algebraically should already go away by simp (as does this). We could even make it a private lemma if we wanted to.

spec/machine/RISCV64/Platform.thy
lsf37: Much agreed.

proof/invariant-abstract/RISCV64/ArchKernelInit_AI.thy
Xaphiosis: Hmm, def' is used for both values and different abstract representation. On the one hand, you're right, on the other, n+1 flavoured soup. I'm fine keeping `_value` and expect the next person who needs to do this to go with `_val`, so we get the full set.

lsf37: Happy to oblige and call the next one `_val` ;-)

spec/machine/ARM/Arch_Kernel_Config_Lemmas.thy
```diff
@@ -0,0 +1,19 @@
...
+ * SPDX-License-Identifier: GPL-2.0-only
+ *)
+
+(* Architecture-specific lemmas constraining Kernel_Config definitions *)
+
+theory Arch_Kernel_Config_Lemmas
+imports
+  Kernel_Config_Lemmas
+  Setup_Locale
+  Platform
```
Xaphiosis, line 13: @lsf37 Platform imports Setup_Locale, so we can drop the Setup_Locale (and squash that change too)

Xaphiosis: I'll do that and pile it on top.

lsf37 (approved):
Quite happy with the outcome of this, a lot of proofs have become simpler and the abstraction was successful.

Xaphiosis (approved):

PR #621, opened 2023-03-30
github: add manual triggers for testing

lsf37:
The worklow_dispatch trigger adds a button in the GitHub UI that lets one trigger the workflow manually. Useful for testing the workflows.


lsf37:
@mbrcknl this should help us to reset the external test, because it runs so rarely.

Xaphiosis (approved):

PR #622, opened 2023-03-30
READMEs: use run_tests consistently in READMEs

lsf37:
Avoid mixing `isabelle`, `make`, and `run_tests` invocations. Standardise on `run_tests` and mention `L4V_ARCH` each time to indicate that you can and should set `L4V_ARCH`.


corlewis (approved):
Good idea!

PR #623, opened 2023-03-30
README: flatten installation instructions

lsf37:
- Provide one flattened set of instructions to install all dependencies, google repo, manifest checkout, and Isabelle installation. At the end of it, link to the description on how to run the proofs.

- Remove jEdit section from main README, since it's duplicated in `setup.md`.

- update Google repo link to a page that contains installation instructions

This doesn't do anything yet about `mlton` on Apple silicon. Next steps would be to provide a binary for that and to turn this page into a script instead.

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -33,9 +33,11 @@ The Haskell Stack package is unavailable on Bullseye and out-of-date on Buster,
 so you will need to install it from the [Haskell Stack
 website](https://docs.haskellstack.org/en/stable/).
 
+Continue with the [python setup step](#python) below.
+
 ### Linux Packages - Ubuntu
 
-These instructions are intended for Ubuntu LTS versions 18.04, 20.04, and 22.04.
+These instructions are intended for Ubuntu LTS versions 20.04, and 22.04.
 
 To run all proofs against the **ARMv7-A** architecture you will need to install
 the following packages:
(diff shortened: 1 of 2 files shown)


docs/setup.md
```diff
@@ -193,12 +227,18 @@ below to bind a key to the "Back" function. We recommend ``[ctrl]+[`]``.
 
 Run the following commands in the directory `verification/l4v/`:
 
-```
+```sh
 mkdir -p ~/.isabelle/jedit/macros
 cp misc/jedit/macros/goto-error.bsh ~/.isabelle/jedit/macros/.
 ```
```
corlewis, line 233: I've never actually done a comparison, how does our macro differ to the built-in Isabelle one that has since been released? Does it make sense to continue suggesting and maintaining our version?

corlewis (approved):
This looks like a really nice improvement.

The README in https://github.com/seL4/verification-manifest might need be updated as well, at the very least to point directly to this improved setup file.

docs/setup.md
lsf37: I still prefer ours because it jumps to the end of the file if there is no error, so the UX for updating proofs is much nicer.

lsf37:
> This looks like a really nice improvement.
> 
> The README in https://github.com/seL4/verification-manifest might need be updated as well, at the very least to point directly to this improved setup file.

Good point, I'll make a PR there as well.

docs/setup.md
```diff
@@ -99,21 +92,55 @@ pip3 install --user sel4-deps
...
+
+instead for the `init` line. The `-m devel.xml` gives you the `master` branch of
+the `l4v` repository instead of the last known-good version. To set up `git` for
+`ssh` make sure to use the `ssh://` protocol explicitly as above instead of just
+`git@github.com`, because the short form seems to confuse `repo`.
+
+[repo]: https://gerrit.googlesource.com/git-repo/+/HEAD/README.md
+[verification-manifest]: https://github.com/seL4/verification-manifest
+
 ## Isabelle Setup
```
scottbuckley, line 139: I notice that we are now building `HOL` instead of `HOL-Word` as part of these instructions. I believe this is only for newer versions of Isabelle, right? At some point the version we were using in timeprot was not up to date, so I had to remember the difference (in fact in my head there was a third way of building Word at some point).

Anyway, if you think this is an issue others might run into, we might want to document that here. If that's not likely, then no change is needed.

scottbuckley (approved):
This is great. Much easier to follow than the previous version I think.

My old flattened instructions - I will strip that down to only the changes I make for my own setup. Do you think I should move that to my personal account, or leave it in au-ts?

docs/setup.md
```diff
@@ -153,7 +180,7 @@ These commands perform the following steps:
 * build basic Isabelle images to ensure that
   the installation works. This may take a few minutes.
 
-Alternatively, it is possible to use the official Isabelle2021 release
+Alternatively, it is possible to use the official Isabelle2022 release
```
michaelmcinerney, line 183: Would it be better to say "the latest official version of Isabelle", or do we want to come back and check this as part of the Isabelle update every time?

michaelmcinerney (approved):
Looks great! Very helpful, even for people who've done this a fair few times, like me

docs/setup.md
lsf37: The problem is that it's not always the latest version. E.g. when Isabelle2023 comes out it'll likely take a few months until we're updated and during that time the sentence is wrong. Of course when we forget to update it like last time, then the sentence is wrong for a longer time ;-)

lsf37:
> This is great. Much easier to follow than the previous version I think.

Coo, that was the objective.
 
> My old flattened instructions - I will strip that down to only the changes I make for my own setup. Do you think I should move that to my personal account, or leave it in au-ts?

If there is still stuff in there that is useful for people in TS I'd leave it in au-ts, but if it's down to only personal setup then I guess moving it makes sense.

docs/setup.md
lsf37: It should be always safe to just build HOL. It'll just take a bit longer to build older sessions the first time because they also need to build HOL-Word, but that will happen automatically.

PR #624, opened 2023-03-31
docs/setup: mlton for M1 now available on homebrew

lsf37:
Installation instructions for Intel and Apple Silicon Mac are now equal


michaelmcinerney (approved):
Cool!

PR #629, opened 2023-04-27
refine: add obj_range'_disjoint

michaelmcinerney:
This also moves several lemmas required for obj_range'_disjoint to Invariants_H

### DIFF
--- a/proof/refine/ARM/Invariants_H.thy
+++ b/proof/refine/ARM/Invariants_H.thy
@@ -3393,4 +3393,50 @@ add_upd_simps "invs' (gsUntypedZeroRanges_update f s)"
   (obj_at'_real_def)
 declare upd_simps[simp]
 
+lemma neq_out_intv:
+  "\<lbrakk> a \<noteq> b; b \<notin> {a..a + c - 1} - {a} \<rbrakk> \<Longrightarrow> b \<notin> {a..a + c - 1}"
+  by simp
+
+lemma rule_out_intv:
+  "\<lbrakk> ksPSpace s a = Some obj; ksPSpace s b = Some obj'; pspace_distinct' s; a \<noteq> b \<rbrakk>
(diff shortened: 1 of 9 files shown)


lsf37 (approved):
Nice. Looking forward to arch splitting Refine ;-)

michaelmcinerney:
There are some errors in the ARM, ARM_HYP, and X64 tests. I was a bit lazy and was hoping they'd be the same as RISCV, but the object ranges aren't defined using `mask` there. I'll cancel the tests, and push a fix soon

michaelmcinerney:
Apologies for the force push. RISCV has had only a minor style change. ARM, ARM_HYP, and X64 all use the same proofs, which are a little different from the proofs for RISCV. 

The proofs of more than a couple lines that end with `by` unfortunately do take a few seconds to finish.

lsf37:
It looks like there's still a slight misalignment somewhere, we're now getting time outs in CRefine. I remember hitting those kinds of annoyances when we introduced ranges with masks. Should not be too bad, though, hopefully.

proof/refine/ARM/Invariants_H.thy
```diff
@@ -3393,4 +3393,50 @@ add_upd_simps "invs' (gsUntypedZeroRanges_update f s)"
   (obj_at'_real_def)
 declare upd_simps[simp]
 
+lemma neq_out_intv:
+  "\<lbrakk> a \<noteq> b; b \<notin> {a..a + c - 1} - {a} \<rbrakk> \<Longrightarrow> b \<notin> {a..a + c - 1}"
```
corlewis, line 3397: I guess this lemma is pretty specific, but do we want to move it to `lib`? That would at least reduce duplication while we haven't arch-split `Refine`.

corlewis (approved):

PR #630, opened 2023-05-01
c-parser: add `dom_lift_t_heap_update` and lemmas for `proj_d`

michaelmcinerney:


michaelmcinerney:
I am expecting to use `dom_lift_t_heap_update` in future work, so I have removed the comment saying that it is unused. And I think it fits nicely in `TypHeap.thy`, rather than `LemmaBucket_C.thy` as the comment suggests

lsf37 (approved):
And deduplicates as well. Happy with it.

corlewis (approved):

PR #636, opened 2023-05-26
AArch64 Refine+Orphanage sorrying run

Xaphiosis:
This PR is based on RISCV64 but adapts:
* hypervisor-related material from ARM_HYP
* FPU-related material from X64 and the iMX8 port
* new changes+improvements in ASpec/AInvs for AArch64

The sorrying run was more difficult than usual due to ARM_HYP being very old (`word32` anyone?), AARCH64 being 3 generations of improvements ahead (ARM_HYP->X64->RISCV64->AARCH64), and no other verification architecture using hypervisor features. In order to confirm that the specs and proofs still made sense, more checking was needed, meaning fewer sorries for the full proof run.

We were not required to look at Orphanage (ARM_HYP doesn't have it), but did it anyway.

Note that there is extra ghost state in the arch state to be able to tell the type of page table present at a given address (root PT vs normal PT). Putting it in the arch state is an attempt to make arch-splitting easier in the future. It also adds dummy definitions etc to other arches.

There's a lot of old tech and styles present in this one.


lsf37:
(Theory linter failure can be ignored, it naturally complains about a lot of new sorries, which is the point of the PR)

lsf37 (approved):
I have done another full reading pass and this is looking good. I can't find any open statements with obvious flaws -- the rest will have to be flushed out by proof.

PR #637, opened 2023-05-31
haskell: constrain run_tests to current L4V_ARCH

lsf37:
Provide L4V_ARCH targets in the Haskell Makefile and constrain run_tests to use only the current L4V_ARCH, to avoid building all architectures in all tests.

A manual invocation of just `make` will still build all architectures for easier checking.


Xaphiosis (approved):
I think this is a most sensible outcome.

PR #638, opened 2023-06-02
cspec: Use L4V_PLAT in build export script

mbrcknl:
CI is introducing an `L4V_PLAT` variable to support proof runs across more platform configurations. This commit incorporates `L4V_PLAT` into the paths generated by `export-kernel-builds.py`, to ensure that exported builds can be disambiguated.

See also https://github.com/seL4/ci-actions/pull/275.


lsf37 (approved):
this looks good and should work well with #639

PR #639, opened 2023-06-02
cspec: introduce L4V_PLAT

lsf37:
L4V_PLAT selects a platform variation within a L4V_ARCH. This mostly affects which seL4 cmake config file is loaded when building config data and the kernel C code. This in turn affects (and will rebuild) ASpec, ExecSpec, and CSpec.

Examples:

    L4V_ARCH=ARM L4V_FEATURES="" L4V_PLAT=""

will load `ARM_verified.cmake`

    L4V_ARCH=ARM L4V_FEATURES="" L4V_PLAT=imx8mm

will load `ARM_imx8mm_verified.cmake`, and

    L4V_ARCH=ARM L4V_FEATURES=MCS L4V_PLAT=imx8mm

will load `ARM_MCS_imx8mm_verified.cmake`.

When `L4V_PLAT` is empty, everything should work as it has so far.

See also https://github.com/seL4/ci-actions/pull/275 and #638

PR #640, opened 2023-06-05
github: auto-rebase platform branches

lsf37:
The action will abort when no clean rebase is possible, and force-push the rebased branch when the rebase over origin/master was clean.

The push will trigger proof runs on the rebased branches.

This required L4V_PLAT etc to be merged first, i.e. everything related to #639

### DIFF
--- a/.github/workflows/proof-deploy.yml
+++ b/.github/workflows/proof-deploy.yml
@@ -76,3 +76,29 @@ jobs:
       with:
         token: ${{ secrets.PRIV_REPO_TOKEN }}
         tag: "l4v/proof-deploy/${{ github.event_name }}"
+
+  rebase:
+    name: Rebase platform branches
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        branch: [imx8-fpu-ver, exynos5-ver]
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+      with:
+        ref: ${{ matrix.branch }}
+        path: l4v-${{ matrix.branch }}
+        fetch-depth: 0
+    - name: Rebase
+      run: |
+        cd l4v-${{ matrix.branch }}
+        git config --global user.name "seL4 CI"
+        git config --global user.email "ci@sel4.systems"
+        git rebase origin/master
+        git status
+    - name: Push
+      run: |
+        cd l4v-${{ matrix.branch }}
+        git push -f origin HEAD:${{ matrix.branch }}
--- a/.github/workflows/proof-deploy.yml
+++ b/.github/workflows/proof-deploy.yml
@@ -101,4 +101,4 @@ jobs:
     - name: Push
       run: |
         cd l4v-${{ matrix.branch }}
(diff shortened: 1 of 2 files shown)


lsf37:
For discussion at this point, I'm not sure this is ready to merge yet.

This current version has no safety-rails whatsoever: it'll just rebase and force-push if the rebase is clean. In particular, this "forgets" the old branch, which may be annoying when the proofs don't work and we want to compare what changed.

This could be alleviated by first pushing to a different branch, e.g. `$branch-rebased` and then force-pushing to `$branch` once the proofs have passed there, which would be slightly safer, but only has one bit of history.

This would require using a slightly different proof action setup on the branches (the force push from ``$branch-rebased` to `$branch`  would have to go in there, and we'd not want those main branches to re-run the proofs again, since they just ran on `$branch-rebased` with the same content.

Race conditions:
 - there could be another rebase + force push while the proofs are still running. I don't think that is a problem -- in both options (with `-rebase` and without) we the old commit still exists on GitHub and continues to run, and the new push will kick off a new run.
 - pushes to seL4 or other manifest repos a) between push to master and the rebase and b) in between push to master and force push to whichever branch does the proof run. The branch run will pick up the current `devel.xml` manifest at that time, which has a specific seL4 version, and branch heads for the other repos (HOL4, graph-refine, isabelle). For seL4, this is very likely, but not guaranteed to be the same version the manifest has in the `Freeze Code` step on `master` (there would have to be a push to that manifest in the time difference between `Freeze Code` gets there and the proof run triggers its own action, which should be less than half a minute). Slightly more likelihood of getting pushes to the master branches of one of the other repos, mostly because nobody but CI usually pushes to the manifest, but the time window is the same.

I think it's fine to ignore the race conditions for now. The branch might then not be in complete sync with the default.xml manifest the master branch will push, but it will be in sync again once the next rebase is triggered from the master branch and there is no interference from other repos. (We already have the same race conditions in the PR actions and have never observed them, but they are also not important there, because we're not aiming to be in sync with a default.xml manifest at any point).

lsf37:
Ok, I've talked myself into at least doing the `-rebased` staging thing, which is not in the current PR. This requires a bit more machinery in the branches themselves, e.g. like in this commit 68582a8a88402eae16fea245827bc6bb7127ad25.

I think this is now ready to go.

Xaphiosis:
This looks ok to me, and the `L4V_PLAT` looks fine as well. Are the respective branches ready for this change? i.e. do they know which `L4V_PLAT` they're supposed to use? Or do we enable the branches first, run them on the `L4V_PLAT`-unaware branches and then change the branches to become aware?

Xaphiosis (approved):

mbrcknl (approved):
Looks good.

lsf37:
> This looks ok to me, and the `L4V_PLAT` looks fine as well. Are the respective branches ready for this change? i.e. do they know which `L4V_PLAT` they're supposed to use? Or do we enable the branches first, run them on the `L4V_PLAT`-unaware branches and then change the branches to become aware?

The branches are all ready to go, this is the final bit to start off the automation. Merging now 🤞

PR #641, opened 2023-06-06
minor run_tests improvements

lsf37:
- echo L4V_FEATURES and L4V_PLAT when running tests
- fix outdated comment
- can leave out REFINE_QUICK_AND_DIRTY, because it's set in proofs/Makefile

### DIFF
--- a/run_tests
+++ b/run_tests
@@ -133,11 +133,6 @@ for arch in archs:
     elif "L4V_ARCH_IS_ARM" in os.environ:
         del os.environ["L4V_ARCH_IS_ARM"]
 
-    # Permit sorries in AARCH64 Refine during development
-    if arch == "AARCH64":
-        os.environ["REFINE_QUICK_AND_DIRTY"]="1"
-        print("Testing AARCH64 Refine in quick and dirty mode")
-
     sys.stdout.flush()
 
     # Arguments:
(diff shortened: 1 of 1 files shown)


corlewis (approved):
These changes look good to me, although I'm a bit confused about how `run_tests` works if it only gets `L4V_PLAT` from the environment. Would we want to be able to pass platforms in as an argument, possibly paired with the architectures to be tested?

lsf37:
> These changes look good to me, although I'm a bit confused about how `run_tests` works if it only gets `L4V_PLAT` from the environment. Would we want to be able to pass platforms in as an argument, possibly paired with the architectures to be tested?

We could do that in a second step, and do the same for L4V_FEATURES, but currently everything uses the environment variables only.

For L4V_ARCH, it makes sense to test for multiple architectures, but for PLAT and FEATURES you need to be on a specific branch each, so until we unify this a bit more, I don't think we'll want to test multiple values of these in one run.

corlewis:
> For L4V_ARCH, it makes sense to test for multiple architectures, but for PLAT and FEATURES you need to be on a specific branch each, so until we unify this a bit more, I don't think we'll want to test multiple values of these in one run.

Right, I didn't think about the different platforms being on different branches. That seems ok then, assuming that on each branch we only test the specific architecture that has been modified.

Issue #634, opened 2023-05-22
simpler `corres` and `corressimp` methods


lsf37:
Dan built two fairly sophisticated proof methods for refinement proofs, `corres` and `corressimp`, but we don't use them a lot.

One reason for that is that we're currently mostly updating existing corres proofs and not writing very many new ones. The other reason is that they don't fail in very useful ways. I.e. when they succeed, it's great, but when the methods fail it is very hard to understand why and where, and even harder to figure out how to make progress from there.

My guess is that if we simplify these methods so that they do much less, it will be easier to form a mental model of what they are doing and therefore to use them more productively. They'll probably solve a lot fewer goals automatically, but if it's clear how to make progress on the goals that they leave over, I think they'll be used a lot more.

Similar (but but not equal) to how `clarsimp` is a lot easier to use incrementally than `force` or `blast`, because it makes incremental progress and one can usually understand how to add things to make more progress.

For instance, one thing that Dan did for these methods is to change the underlying corres calculus such that it better supports automation. That new calculus definitely is better for automation, but since we don't use it manually, we don't really understand its subtleties.

My proposal would be as first step to write a completely naive `corres` method that only know how to split off `bind` with functions it already has rules for. Maybe, as a plus, it can attempt to immediately solve the `wp` goals that it produces, but that is already advanced.

Opinions and discussions welcome!

Xaphiosis:
Some extra thoughts:

`corres` has issues with single-sided rules, e.g. dealing with an assert on either side

Best-guess automation as with `crunch` really benefits from tracing and feedback, especially with a two-stage cascade of corres rules + wp[simp]. Things could go wrong at either stage, and being able to trace which corres rules were used is invaluable.

The way wpsimp et al only split off a chunk (bind) if they can do something with it is nice, we should keep that.

Graceful degradation is helpful, i.e. being able to get dumped in the middle of a failed wp discharging after the corres. For bonus points, being able to resume after human assistance would also be helpful, but that can get really weird given that corres can spit out combinations of corres and wp goals, and we wouldn't want the tactic to just keep eating away at goals as long as they are of either form. I guess one could limit the number of goals, but by that argument one could also do `(my_tactic+)[n_goals]`.

It sucks to have to specify `corres_split_eqr[OF my_corres_rule]` and similar. Would be nice to be able to only give it `my_corres_rule`. This would necessitate some kind of  basic mechanism to go from user-specified, to most specific, to more generic rules?

lsf37:
[...]

PR #642, opened 2023-06-09
lib: split out WP_Pre.pre_tac for wp_pre

lsf37:
Factor out pre_tac such that we can have separate theorem sets and methods for wp_pre, monadic_rewrite_pre, corres_pre, and potentially others in the future.

Leave everything in wp_pre that we expect to use wp or wp_simp on, in particular also no_fail.

Implements a tiny part of #634

### DIFF
--- a/lib/MonadicRewrite.thy
+++ b/lib/MonadicRewrite.thy
@@ -40,7 +40,7 @@ lemma monadic_rewrite_impossible:
   "monadic_rewrite F E \<bottom> f g"
   by (clarsimp simp: monadic_rewrite_def)
 
-lemma monadic_rewrite_guard_imp[wp_pre]:
+lemma monadic_rewrite_guard_imp:
   "\<lbrakk> monadic_rewrite F E Q f g; \<And>s. P s \<Longrightarrow> Q s \<rbrakk> \<Longrightarrow> monadic_rewrite F E P f g"
   by (auto simp add: monadic_rewrite_def)
 
@@ -720,6 +720,10 @@ wpc_setup "\<lambda>m. monadic_rewrite F E Q' (m >>=E c) m'" wpc_helper_monadic_
 
 text \<open>Tactics\<close>
 
+named_theorems monadic_rewrite_pre
+declare monadic_rewrite_guard_imp[monadic_rewrite_pre]
+method monadic_rewrite_pre = (WP_Pre.pre_tac monadic_rewrite_pre)?
+
 method monadic_rewrite_step =
   determ \<open>rule monadic_rewrite_bind_tail monadic_rewrite_bindE_tail\<close>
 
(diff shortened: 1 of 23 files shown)


lib/MonadicRewrite.thy
```diff
@@ -763,7 +766,7 @@ method monadic_rewrite_l_method methods action finalise =
 
 (* Step over RHS until action applies, then finalise. *)
 method monadic_rewrite_r_method methods action finalise =
-  monadic_rewrite_single_pass \<open>wp_pre, rule monadic_rewrite_trans[rotated]\<close>
+  monadic_rewrite_single_pass \<open>monadic_rewrite_pre, rule monadic_rewrite_trans[rotated]\<close>
```
Xaphiosis, line 769: wait, how is this gonna work if you haven't added any theorems to `monadic_rewrite_pre`?

lsf37: right, it's not going to work :-). I forgot to add the thm after declaring the theorem set

lib/Monads/wp/WP_Pre.thy
```diff
@@ -51,21 +50,20 @@ fun pre_tac trace ctxt pre_rules used_thms_ref i t = let
...
-    Args.context >> (fn _ => fn ctxt =>
-      Method.SIMPLE_METHOD' (tac (Config.get ctxt wp_trace) used_thms_ref ctxt))
+    Attrib.thms >> (fn thms => fn ctxt =>
+      Method.SIMPLE_METHOD' (pre_tac (Config.get ctxt wp_trace) ctxt thms used_thms_ref))
   end
 end
 \<close>
 
-method_setup wp_pre0 = \<open>WP_Pre.method\<close>
+method_setup pre_tac = \<open>WP_Pre.method\<close>
```
Xaphiosis, line 63: I think some docs might be useful for how this works and how people should use this

lib/Monads/wp/WP_Pre.thy
```diff
@@ -51,21 +50,20 @@ fun pre_tac trace ctxt pre_rules used_thms_ref i t = let
...
+
+named_theorems wp_pre
+method wp_pre0 declares wp_pre = pre_tac wp_pre
 method wp_pre = wp_pre0?
```
Xaphiosis, line 67: Speaking only for myself, I find having the same name for the method and named thms a bit confusing. It probably is the right way to go in terms of actual use, and only confusing during implementation.

Xaphiosis:
In the commit message, did you intend to write `wp_simp`? I'm not familiar with that name

lib/MonadicRewrite.thy
lsf37: Ok, added now.

lib/Monads/wp/WP_Pre.thy
lsf37: Usage is added now -- for how it works, I don't actually know. This version of the tactic is completely different to what I wrote in the beginning, or at least to what I remember planning to write. I just pulled out the named_theorem into a parameter.

lsf37:
> In the commit message, did you intend to write `wp_simp`? I'm not familiar with that name

Now fixed.

lib/Monads/wp/WP_Pre.thy
lsf37: It is a bit confusing in implementation, but I think it is better for the user, because there are fewer names to remember. It's also an old tradition for things like `simp`. It'd probably have been better back then (25 years ago or so) to introduce the convention of `simp_thms`, but I don't think we should do our own one now.

Xaphiosis (approved):

Xaphiosis:
Looks good. Minor suggestion for commit message: "in particular also" -> "in particular"

PR #643, opened 2023-06-13
Monadic rewrite: improve bound name handling, add docs/tests

Xaphiosis:
This is a first stab at some documentation / demo material to help understand what the `monadic_rewrite` top-level tactics do.

There is also a commit with a very long message that describes the changes to handling bound variables in monadic statements when iterating over them with `monadic_rewrite_bind_tail` and grabbing onto them with `monadic_rewrite_bind_head`. This was very, very hard to figure out. I still don't understand why bind_head must not have eta terms for example, and for the bind_tail it's a reasonable, but best-guess.

I have some issues confusing `eta` and `[lambda] abstraction` that could use some correcting, plus feedback on the (in)comprehensibility of the demo (I know Michael wanted to see some examples in the past).

This probably needs to be tweaked after the wp_pre PR goes in (#642)

### DIFF
--- a/lib/MonadicRewrite.thy
+++ b/lib/MonadicRewrite.thy
@@ -146,17 +146,22 @@ lemma monadic_rewrite_bindE:
   apply (case_tac x; simp add: lift_def monadic_rewrite_refl)
   done
 
+(* in order to preserve bound names in the tail, bind_head must avoid eta on both sides *)
+lemma monadic_rewrite_bind_head:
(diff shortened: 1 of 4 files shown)


lsf37 (approved):
Very nice to have an example + test file for `monadic_rewrite` and extremely nice to preserve names much better now.

Xaphiosis:
Rebased for `wp_pre -> monadic_rewrite_pre` change, ready to merge given no one seems interested in the docs.

PR #644, opened 2023-06-13
allow a second predicate-type guard in `wpc`

lsf37:
- document the `wpc_helper` predicate and setup for `wpc` (but not the proof method internals at this point)

- add facilities for a second guard of predicate type

The current setup works well for judgements with one guard of type predicate or set (valid, validE, etc), or with two guards where one is a predicate and the other is a set (ccorres), but not for judgements that have two predicate guards, i.e. plain corres. This commit adds a second such guard, which can be ignored for the judgements that don't need it in the same way that valid/validE currently ignore the set-type guard.

### DIFF
--- a/lib/EquivValid.thy
+++ b/lib/EquivValid.thy
@@ -572,10 +572,10 @@ lemmas pre_ev =
   hoare_pre
   equiv_valid_guard_imp
 
-subsection\<open>Tom instantiates wpc\<close>
+subsection\<open>wpc setup\<close>
 
 lemma wpc_helper_equiv_valid:
-  "equiv_valid D A B Q f \<Longrightarrow> wpc_helper (P, P') (Q, Q') (equiv_valid D A B P f)"
+  "equiv_valid D A B Q f \<Longrightarrow> wpc_helper (P, P', P'') (Q, Q', Q'') (equiv_valid D A B P f)"
   using equiv_valid_guard_imp
   apply (simp add: wpc_helper_def)
   apply (blast)
--- a/lib/MonadicRewrite.thy
+++ b/lib/MonadicRewrite.thy
@@ -710,12 +710,13 @@ lemmas corres_gets_the_bind
 text \<open>Tool integration\<close>
 
 lemma wpc_helper_monadic_rewrite:
-  "monadic_rewrite F E Q' m m'
-   \<Longrightarrow> wpc_helper (P, P') (Q, {s. Q' s}) (monadic_rewrite F E (\<lambda>s. s \<in> P') m m')"
+  "monadic_rewrite F E Q m m'
+   \<Longrightarrow> wpc_helper (P, P', P'') (Q, Q', Q'') (monadic_rewrite F E P m m')"
   by (auto simp: wpc_helper_def elim!: monadic_rewrite_guard_imp)
 
 wpc_setup "\<lambda>m. monadic_rewrite F E Q' m m'" wpc_helper_monadic_rewrite
 wpc_setup "\<lambda>m. monadic_rewrite F E Q' (m >>= c) m'" wpc_helper_monadic_rewrite
+wpc_setup "\<lambda>m. monadic_rewrite F E Q' (m >>=E c) m'" wpc_helper_monadic_rewrite
 
 text \<open>Tactics\<close>
 
--- a/lib/Monads/wp/WPC.thy
+++ b/lib/Monads/wp/WPC.thy
@@ -1,4 +1,5 @@
 (*
+ * Copyright 2023, Proofcraft Pty Ltd
  * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
  *
  * SPDX-License-Identifier: BSD-2-Clause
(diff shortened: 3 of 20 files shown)


sys-init/DuplicateCaps_SI.thy
```diff
@@ -256,7 +256,7 @@ lemma distinct_card':
 (* FIXME, move higher *)
 lemma distinct_length_filter':
   "distinct xs \<Longrightarrow> length [x\<leftarrow>xs. P x] = card {x \<in> set xs. P x}"
-  by (metis distinct_length_filter set_conj_Int_simp inf_commute)
+  by (metis distinct_card' distinct_filter set_filter)
```
Xaphiosis, line 259: unexpected change for this PR

lib/MonadicRewrite.thy
```diff
@@ -710,12 +710,13 @@ lemmas corres_gets_the_bind
...
 lemma wpc_helper_monadic_rewrite:
-  "monadic_rewrite F E Q' m m'
-   \<Longrightarrow> wpc_helper (P, P') (Q, {s. Q' s}) (monadic_rewrite F E (\<lambda>s. s \<in> P') m m')"
+  "monadic_rewrite F E Q m m'
+   \<Longrightarrow> wpc_helper (P, P', P'') (Q, Q', Q'') (monadic_rewrite F E P m m')"
   by (auto simp: wpc_helper_def elim!: monadic_rewrite_guard_imp)
 
 wpc_setup "\<lambda>m. monadic_rewrite F E Q' m m'" wpc_helper_monadic_rewrite
 wpc_setup "\<lambda>m. monadic_rewrite F E Q' (m >>= c) m'" wpc_helper_monadic_rewrite
+wpc_setup "\<lambda>m. monadic_rewrite F E Q' (m >>=E c) m'" wpc_helper_monadic_rewrite
```
Xaphiosis, line 719: good catch on this one

lib/EquivValid.thy
```diff
@@ -575,7 +575,7 @@ lemmas pre_ev =
 subsection\<open>Tom instantiates wpc\<close>
```
Xaphiosis, line 575: I know this isn't yours, but this section title ...

lib/Monads/wp/WPC.thy
```diff
@@ -213,67 +213,88 @@ end;
...
+  "\<lbrakk> wpc_test Q X' R S; P \<subseteq> Q; \<And>s. X s \<Longrightarrow> X' s \<rbrakk> \<Longrightarrow> wpc_test P X R S"
   by (simp add: wpc_test_def, blast)
 
-lemma wpc_helper_validF:
-  "wpc_test Q' R S \<Longrightarrow> wpc_helper (P, P') (Q, Q') (wpc_test P' R S)"
-  by (simp add: wpc_test_def wpc_helper_def, blast)
+(* Setup rule, establishes connection between wpc_helper and judgment wpc_test.
+   The precondition has the judgement with transformed (Q) guards, the conclusion has the
+   helper predicate with the judgement applied to the original (P) guards. The guard triple
+   arguments of wpc_helper must be in the form below (no arguments or patterns) for the method
```
Xaphiosis, line 258: "The guard triple arguments of ..." - not comprehending what this means

lib/Monads/wp/WPC.thy
```diff
@@ -213,67 +213,88 @@ end;
 val _ =
     Outer_Syntax.command
         @{command_keyword "wpc_setup"}
-        "Add wpc stuff"
+        "Add new WPC term and helper rule"
```
Xaphiosis, line 216: oh wow

lib/Monads/wp/WPC.thy
```diff
@@ -7,61 +8,64 @@
...
+(* Case splitting method producing independent guards (preconditions) for each case in a
+   datatype case split. The current setup can handle judgements such as valid, corres, or ccorres
+   with up to two independent predicate guards and one independent set typ guard. Unneeded guards
+   can be ignored in setup.
+
+   The helper predicate unifies the treatment of guards in the proof method. The P guards will be
+   transformed into Q guards in each branch of the case. The R is the judgement (valid, corres, etc).
+
+   The helper predicate encodes that the judgement supports a standard guard weakening rule.
+   From that rules for conjunction-lifting and forall-lifting follow below. These are then used
```
Xaphiosis, line 22: suggestion: `. From that` -> `, from which` (unless you mean something else here)

lib/Monads/wp/WPC.thy
```diff
@@ -7,61 +8,64 @@
...
-
 begin
 
-definition
-  wpc_helper :: "(('a \<Rightarrow> bool) \<times> 'b set)
-                 \<Rightarrow> (('a \<Rightarrow> bool) \<times> 'b set) \<Rightarrow> bool \<Rightarrow> bool" where
- "wpc_helper \<equiv> \<lambda>(P, P') (Q, Q') R. ((\<forall>s. P s \<longrightarrow> Q s) \<and> P' \<subseteq> Q') \<longrightarrow> R"
+(* Case splitting method producing independent guards (preconditions) for each case in a
+   datatype case split. The current setup can handle judgements such as valid, corres, or ccorres
+   with up to two independent predicate guards and one independent set typ guard. Unneeded guards
```
Xaphiosis, line 15: "set typ guard" -> "set type guard"?

Xaphiosis (approved):
Nice, improvement. Thoughts on moving `corres` wpc setup out of AutoCorres? I know you mentioned that issue yesterday.

I won't lie and say I now get how wpc works, but the explanation definitely helps going in that direction.

sys-init/DuplicateCaps_SI.thy
lsf37: ah, but it should be expected :-). The lemma `set_conj_Int_simp` for some reason was proved in `WPC.thy` and is unused in l4v apart from this one file. So this change is a consequence of cleaning up `WPC.thy`.

lib/EquivValid.thy
lsf37: Ha, I did not see that :-) I should probably adjust this.

lib/Monads/wp/WPC.thy
lsf37: To be honest, I don't really comprehend this any more either. Might have lost a word somewhere in editing that sentence. Will adjust.

lsf37:
> Nice, improvement. Thoughts on moving `corres` wpc setup out of AutoCorres? I know you mentioned that issue yesterday.

I was planning on leaving that for the `Corres_Method` PR because I was afraid of breaking more things if I move it now. Maybe I'll just make it an independent PR, because the current form of the corres method doesn't actually use `wpc` any more, so the setup should probably be directly in `Corres_UL`.

PR #646, opened 2023-06-19
Cong rules for ccorres_underlying

corlewis:
This demonstrates the required changes after adding new cong rules for `ccorres_underlying`, in particular one that does not rewrite the return relations and extraction functions.

In general, I would say that this helps with all of the easy cases due to having more stable return relations that are easier to unify with. However, there are some cases that are now more complicated due to needing to choose a different cong rule.

### DIFF
--- a/lib/clib/Corres_UL_C.thy
+++ b/lib/clib/Corres_UL_C.thy
@@ -1719,4 +1719,110 @@ lemma ccorres_grab_asm:
    ccorres_underlying sr G rr xf ar ax (P and K Q) P' hs f g"
   by (fastforce simp: ccorres_underlying_def)
 
+
+\<comment> \<open> An experimental cong rule for rewriting everywhere reasonable, with full context.
+    Can cause problems when there are schematic variables or when one of the return relations
+    takes a pair as a parameter. \<close>
+lemma ccorres_context_cong_helper':
+  assumes c: "ccorres_underlying sr \<Gamma> r xf ar axf P Q hs a c"
+  assumes "\<And>s. P s = P' s"
+  \<comment> \<open>Don't use membership equality when rewriting Q, as the LHS can be simplified into something
+     that is unable to unify with the RHS.
+  assumes "\<And>s s'. \<lbrakk> (s,s') \<in> sr; P' s \<rbrakk> \<Longrightarrow> s' \<in> Q = (s' \<in> Q')"\<close>
+  assumes "\<And>s s'. \<lbrakk> (s, s') \<in> sr; P' s \<rbrakk> \<Longrightarrow> Q = Q'"
+  assumes "hs = hs'"
+  assumes "\<And>s s'. \<lbrakk> (s, s') \<in> sr; P' s; s' \<in> Q' \<rbrakk> \<Longrightarrow> a s = a' s"
+  assumes "\<And>s s' s''. \<lbrakk> (s, s') \<in> sr; P' s; s' \<in> Q' \<rbrakk>  \<Longrightarrow> semantic_equiv \<Gamma> s' s'' c c'"
+  assumes "\<And>s s' t'.
+             \<lbrakk> (s, s') \<in> sr; P' s; s' \<in> Q'; \<Gamma> \<turnstile>\<^sub>h \<langle>c' # hs', s'\<rangle> \<Rightarrow> (size hs', Normal t') \<rbrakk> \<Longrightarrow>
+             xf t' = xf' t'"
+  assumes "\<And>x s t s' t'.
+             \<lbrakk> (s, s') \<in> sr; P' s; s' \<in> Q'; (t, t') \<in> sr; (x, t) \<in> fst (a' s);
+               \<Gamma> \<turnstile>\<^sub>h \<langle>c' # hs', s'\<rangle> \<Rightarrow> (size hs', Normal t') \<rbrakk> \<Longrightarrow>
+             r x (xf' t') = r' x (xf' t')"
+  assumes "\<And>s s' t' n.
+             \<lbrakk> (s, s') \<in> sr; P' s; s' \<in> Q'; n \<noteq> size hs'; \<Gamma> \<turnstile>\<^sub>h \<langle>c' # hs', s'\<rangle> \<Rightarrow> (n, Normal t') \<rbrakk> \<Longrightarrow>
+             axf t' = axf' t'"
+  assumes "\<And>x s t s' t' n.
+             \<lbrakk> (s, s') \<in> sr; P' s; s' \<in> Q'; (t, t') \<in> sr; (x, t) \<in> fst (a' s); n \<noteq> size hs';
+               \<Gamma> \<turnstile>\<^sub>h \<langle>c' # hs', s'\<rangle> \<Rightarrow> (n, Normal t') \<rbrakk> \<Longrightarrow>
+             ar x (axf' t') = ar' x (axf' t')"
+  shows   "ccorres_underlying sr \<Gamma> r' xf' ar' axf' P' Q' hs' a' c'"
+  using c
+  apply -
+  apply (rule ccorresI')
+  apply (erule (1) ccorresE)
+      apply (force simp: assms)
+     apply (force simp: assms)
+    apply (force simp: assms)
+   apply (clarsimp simp: assms)
+   apply (erule exec_handlers_semantic_equivD2)
+   apply (force simp: assms)
+  apply (fastforce simp: unif_rrel_def assms)
+  done
+
+lemma ccorres_context_cong_helper:
+  assumes "\<And>s. P s = P' s"
+  assumes "\<And>s s'. \<lbrakk> (s, s') \<in> sr; P' s \<rbrakk> \<Longrightarrow> Q = Q'"
+  assumes "hs = hs'"
+  assumes "\<And>s s'. \<lbrakk> (s, s') \<in> sr; P' s; s' \<in> Q' \<rbrakk> \<Longrightarrow> a s = a' s"
+  assumes "\<And>s s' s''. \<lbrakk> (s, s') \<in> sr; P' s; s' \<in> Q' \<rbrakk>  \<Longrightarrow> semantic_equiv \<Gamma> s' s'' c c'"
+  assumes "\<And>s s' t'.
+             \<lbrakk> (s, s') \<in> sr; P' s; s' \<in> Q'; \<Gamma> \<turnstile>\<^sub>h \<langle>c' # hs', s'\<rangle> \<Rightarrow> (size hs', Normal t') \<rbrakk> \<Longrightarrow>
+             xf t' = xf' t'"
+  assumes "\<And>x s t s' t'.
+             \<lbrakk> (s, s') \<in> sr; P' s; s' \<in> Q'; (t, t') \<in> sr; (x, t) \<in> fst (a' s);
+               \<Gamma> \<turnstile>\<^sub>h \<langle>c' # hs', s'\<rangle> \<Rightarrow> (size hs', Normal t') \<rbrakk> \<Longrightarrow>
+             r x (xf' t') = r' x (xf' t')"
+  assumes "\<And>s s' t' n.
+             \<lbrakk> (s, s') \<in> sr; P' s; s' \<in> Q'; n \<noteq> size hs'; \<Gamma> \<turnstile>\<^sub>h \<langle>c' # hs', s'\<rangle> \<Rightarrow> (n, Normal t') \<rbrakk> \<Longrightarrow>
+             axf t' = axf' t'"
+  assumes "\<And>x s t s' t' n.
+             \<lbrakk> (s, s') \<in> sr; P' s; s' \<in> Q'; (t, t') \<in> sr; (x, t) \<in> fst (a' s); n \<noteq> size hs';
+               \<Gamma> \<turnstile>\<^sub>h \<langle>c' # hs', s'\<rangle> \<Rightarrow> (n, Normal t') \<rbrakk> \<Longrightarrow>
+             ar x (axf' t') = ar' x (axf' t')"
+  shows   "ccorres_underlying sr \<Gamma> r xf ar axf P Q hs a c
+           = ccorres_underlying sr \<Gamma> r' xf' ar' axf' P' Q' hs' a' c'"
+  using assms
+  apply -
+  apply rule
+   apply (erule ccorres_context_cong_helper'; assumption)
+  apply (erule ccorres_context_cong_helper')
+  by (fastforce simp: semantic_equiv_sym exec_handlers_semantic_equiv[where a=c and b=c'])+
+
+lemmas ccorres_context_cong = ccorres_context_cong_helper[OF _ _ _ _ semantic_equivI]
+
+\<comment> \<open> Only rewrite guards, the handler stack and function bodies, with context.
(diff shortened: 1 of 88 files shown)


proof/crefine/lib/Corres_C.thy
```diff
@@ -1015,21 +1011,53 @@ lemma ccorres_zipWithM_x_while_genQ:
...
+    apply (rule wb)
+   apply simp+
   done
 
+\<comment> \<open>This context is used to temporarily remove ccorres_weak_cong, so that the following lemmas can
+    be constructed with simplified return relations.
+    We do not use ccorres_all_cong due to it causing unexpected eta-expansion.\<close>
+context
+notes ccorres_weak_cong[cong del]
+begin
```
corlewis, line 1030: I moved these lemmas so that I only needed the context boilerplate once. Are we ok with this, or would we prefer to keep the original structure with each lemma following the lemma it is constructed from?

lib/clib/Corres_UL_C.thy
```diff
@@ -1719,4 +1719,113 @@ lemma ccorres_grab_asm:
...
+          apply (fastforce simp: semantic_equiv_sym exec_handlers_semantic_equiv[where a=c and b=c'])+
+  done
+
+lemmas ccorres_context_cong = ccorres_context_cong_helper[OF _ _ _ _ semantic_equivI]
+
+\<comment> \<open> Only rewrite guards, the handler stack and function bodies, with context.
+    This is often more useful, as we generally want the return relations and extraction
+    functions to be stable while working with a ccorres_underlying statement. \<close>
+\<comment> \<open> FIXME: this lemmas command unifies the return relations in an unexpected way.
+lemmas ccorres_context_weak_cong = ccorres_context_cong[OF _ _ _ _ _ refl refl refl refl]\<close>
```
corlewis, line 1800: Is it worth keeping this FIXME in, or are we ok with writing out `ccorres_context_weak_cong` in full?

corlewis:
```
(* Unit would be more appropriate, but the record package will simplify xfdc to () *)
definition
  "xfdc (t :: cstate) ≡ (0 :: nat)"

lemma xfdc_equal [simp]:
  "xfdc t = xfdc s"
  unfolding xfdc_def by simp
```
Currently, the provided `xfdc_equal` lemma is not used by the simplifier, presumably due to the fact that it would loop. This means that we need to unfold `xfdc_def` for simple goals like `xfdc s = xfdc (f s)`.

If we go with making `ccorres_weak_cong` the default then I think it might now be ok to change `xfdc` to unit. This would not be rewritten inside `ccorres` due to the cong rule, but would resolve other goals by being simplified when it appears on its own.

proof/crefine/lib/Corres_C.thy
```diff
@@ -101,6 +101,24 @@ lemma exvalue_liftxf [simp]:
   unfolding liftxf_def by simp
 
 
+\<comment> \<open>FIXME: move?\<close>
```
corlewis, line 104: Maybe to `lib/clib/CCorresLemmas.thy`? That's where the required `inl_rrel` and `inr_rrel` lemmas are.

proof/crefine/lib/Corres_C.thy
lsf37: From my side that's fine, it's different, but no less sane than the original I think

lib/clib/Corres_UL_C.thy
lsf37: I'd actually prefer writing it out, it makes it easier to see what is going on.

lib/clib/Corres_UL_C.thy
```diff
@@ -1719,4 +1719,113 @@ lemma ccorres_grab_asm:
...
+\<comment> \<open> An experimental cong rule for rewriting everywhere reasonable, with full context.
+    Can cause problems when there are schematic variables or when one of the return relations
+    takes a pair as a parameter. \<close>
+lemma ccorres_context_cong_helper':
+  assumes c: "ccorres_underlying sr \<Gamma> r xf ar axf P Q hs a c"
+  assumes "\<And>s. P s = P' s"
+  \<comment> \<open>Don't add context for rewriting Q, as the LHS can be simplified into something that is
+     unable to unify with the RHS.
+  assumes "\<And>s s'. \<lbrakk> (s,s') \<in> sr; P' s \<rbrakk> \<Longrightarrow> s' \<in> Q = (s' \<in> Q')"\<close>
+  assumes "Q = Q'"
```
lsf37, line 1732: If we wanted to, we could actually still add a little bit of context:
```
assumes "\<And>s s'. \<lbrakk> (s,s') \<in> sr; P' s \<rbrakk> \<Longrightarrow> Q = Q'"
```
This would allow it to use predicates or parts of the state relation that don't depend on the state in the simplification of `Q`.

lsf37 (commented):
This looks very nice! So much removal of dc_def, we should have done that 10 years ago, our pain threshold is way too high :-)

The removals by far outweigh the few small adjustments that are needed. I'd be for rolling this out over all architectures and merging into rt soon so Michael can benefit from it in CRefine.

lib/clib/Corres_UL_C.thy
Xaphiosis: Yeah, sorry, my idea didn't work out here.

lib/clib/Corres_UL_C.thy
```diff
@@ -1719,4 +1719,113 @@ lemma ccorres_grab_asm:
...
+    functions to be stable while working with a ccorres_underlying statement. \<close>
+\<comment> \<open> FIXME: this lemmas command unifies the return relations in an unexpected way.
+lemmas ccorres_context_weak_cong = ccorres_context_cong[OF _ _ _ _ _ refl refl refl refl]\<close>
+lemma ccorres_context_weak_cong:
+  assumes "\<And>s. P s = P' s"
+  assumes "Q = Q'"
+  assumes "\<And>s s'. \<lbrakk> (s,s') \<in> sr; P' s; s' \<in> Q' \<rbrakk> \<Longrightarrow> a s = a' s"
+  assumes "\<And>s s' s''. \<lbrakk> (s,s') \<in> sr; P' s; s' \<in> Q' \<rbrakk>  \<Longrightarrow> \<Gamma>\<turnstile> \<langle>c,Normal s'\<rangle> \<Rightarrow> s'' = \<Gamma>\<turnstile> \<langle>c',Normal s'\<rangle> \<Rightarrow> s''"
+  shows   "ccorres_underlying sr \<Gamma> r xf ar axf P Q hs a c
+             = ccorres_underlying sr \<Gamma> r xf ar axf P' Q' hs a' c'"
```
Xaphiosis, line 1807: The indentation in these rules is off. The idea of putting `=` on the left is to not have to indent, so I'd suggest either:
```
a =
b
```
or
```
a
= b
```
here (and below)

lib/clib/Corres_UL_C.thy
```diff
@@ -1719,4 +1719,113 @@ lemma ccorres_grab_asm:
...
+    environment. \<close>
+lemma ccorres_all_cong:
+  "\<lbrakk> r=r'; xf=xf'; ar=ar'; axf=axf'; P=P'; Q=Q'; hs=hs'; m=m'; c=c' \<rbrakk> \<Longrightarrow>
+   ccorres_underlying srel \<Gamma> r xf ar axf P Q hs m c
+     = ccorres_underlying srel \<Gamma> r' xf' ar' axf' P' Q' hs' m' c'"
+  by (simp cong: ccorres_context_cong)
+
+\<comment> \<open> Only rewrite guards, the handler stack and function bodies, without context.
+    We make this the default behaviour, so that the the return relations and extraction
+    functions are stable under simplification. \<close>
```
Xaphiosis, line 1828: Really appreciate these docs.

proof/crefine/ARM/Finalise_C.thy
```diff
@@ -1787,12 +1771,6 @@ lemma Arch_finaliseCap_ccorres:
    apply (frule cap_get_tag_isCap_unfolded_H_cap, simp)
   done
 
-lemma ccte_relation_ccap_relation:
-  "ccte_relation cte cte' \<Longrightarrow> ccap_relation (cteCap cte) (cte_C.cap_C cte')"
-  by (clarsimp simp: ccte_relation_def ccap_relation_def
-                     cte_to_H_def map_option_Some_eq2
-                     c_valid_cte_def)
```
Xaphiosis, line 1794: are these getting removed because they're already proved elsewhere? (referring also to `empty_fail_getNotification`

proof/crefine/lib/Corres_C.thy
Xaphiosis: good plan

proof/crefine/lib/Corres_C.thy
```diff
@@ -988,14 +988,10 @@ lemma ccorres_mapM_x_while_gen':
...
    apply (rule ccorres_sequence_x_while_gen'[where xf_update=xf_update])
-        apply (clarsimp simp only: length_map nth_map rl)
-       apply (simp add: assms hi[simplified])+
+         apply (clarsimp simp only: length_map nth_map rl)
+        apply (simp add: assms hi[simplified])+
   done
 
-lemmas ccorres_mapM_x_while'
-    = ccorres_mapM_x_while_gen' [OF _ _ _ _ _ i_xf_for_sequence, folded word_bits_def,
-                                 where j=1, simplified]
```
Xaphiosis, line 997: there's a bunch of these getting removed... are they unused, already proved, bad ideas?

Xaphiosis: oh never mind, they got moved to the context below

proof/crefine/lib/Corres_C.thy
Xaphiosis: I think you can drop "this context is used to" from the doc. Good idea clumping them together. Would they make more sense towards the top of the file rather than at the very bottom though?

Xaphiosis:
I am really glad you decided to do this. The fragility of `dc_def` and `dc_simp` always sucked, and this is a very nice way of dealing with it!

proof/crefine/lib/Corres_C.thy
corlewis: They need to be after all of the lemmas that they are constructed from, so without much larger changes to the file structure this is the earliest they can go.

proof/crefine/lib/Corres_C.thy
corlewis: Moved

lib/clib/Corres_UL_C.thy
corlewis: Good idea, I hadn't thought of parts that didn't depend on the state.
I've done this now and it at least didn't break the one or two places these cong rules are currently being used.

lib/clib/Corres_UL_C.thy
corlewis: Whoops, fixed.

proof/crefine/ARM/Finalise_C.thy
corlewis: Yeah, exactly. Sorry for mixing them in with unrelated changes, I was just cleaning them up when I noticed them and they were adjacent to lemmas I was already touching.

What are our thoughts for bits of cleanup like this? Optimally they would go in their own commits and even PRs but with how long it takes to rebuild and make changes I feel like that would lead to it almost never happening.

corlewis:
> ```
> (* Unit would be more appropriate, but the record package will simplify xfdc to () *)
> definition
>   "xfdc (t :: cstate) ≡ (0 :: nat)"
> 
> lemma xfdc_equal [simp]:
>   "xfdc t = xfdc s"
>   unfolding xfdc_def by simp
> ```
> 
> Currently, the provided `xfdc_equal` lemma is not used by the simplifier, presumably due to the fact that it would loop. This means that we need to unfold `xfdc_def` for simple goals like `xfdc s = xfdc (f s)`.
> 
> If we go with making `ccorres_weak_cong` the default then I think it might now be ok to change `xfdc` to unit. This would not be rewritten inside `ccorres` due to the cong rule, but would ease other goals by being simplified when it appears on its own.

After some experimentation and off-line discussions we've concluded that changing `xfdc` to `unit` unfortunately still isn't possible. It works almost everywhere but there is at least one edge case where `clarify` can still rewrite it to `()`, even protected by a cong rule.

corlewis:
> > ```
> > (* Unit would be more appropriate, but the record package will simplify xfdc to () *)
> > definition
> >   "xfdc (t :: cstate) ≡ (0 :: nat)"
> > 
> > lemma xfdc_equal [simp]:
> >   "xfdc t = xfdc s"
> >   unfolding xfdc_def by simp
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Currently, the provided `xfdc_equal` lemma is not used by the simplifier, presumably due to the fact that it would loop. This means that we need to unfold `xfdc_def` for simple goals like `xfdc s = xfdc (f s)`.
> > If we go with making `ccorres_weak_cong` the default then I think it might now be ok to change `xfdc` to unit. This would not be rewritten inside `ccorres` due to the cong rule, but would ease other goals by being simplified when it appears on its own.
> 
> After some experimentation and off-line discussions we've concluded that changing `xfdc` to `unit` unfortunately still isn't possible. It works almost everywhere but there is at least one edge case where `clarify` can still rewrite it to `()`, even protected by a cong rule.

For posterity, I also tried to work with the current definition and improve the automation but couldn't find an approach that I liked.

The closest was along the lines of
```
definition
  "xfdc' = (0 :: nat)"

definition
  "xfdc ≡ λ(t :: cstate). xfdc'"

lemma xfdc_simp[simp]:
  "xfdc t = xfdc'"
  by (simp add: xfdc_def)

lemma xfdc'_simp[simp]:
  "(λt. xfdc') = xfdc"
  by (simp add: xfdc_def)
```
but this is a rather strange formulation and could still lead to unexpected intermediate goals. This possibly could have been further improved by making `xfdc` an abbreviation but at that point things started breaking in `ctac` and this became more complicated than I wanted to deal with.

proof/crefine/ARM/Finalise_C.thy
Xaphiosis: Definitely fine to have alongside other commits in one PR.
It's definitely possible to extract these into a separate commit without having to rebuild. One way I'd suggest is to create a branch identical to the current one. Then you can diff against that branch and be sure whatever git-fu you're doing doesn't affect anything.
Via `git add -p`, can add the removals separately, and then commit the rest of the change as a the same commit as before.

The other approach is to notice the lemmas you removed, remove those first as one commit, and then rebase your PR on top of that (and check with backup branch, should be empty diff).

I'm not sure how much it's worth it. If you've only reviewed a few lemmas, it's not going to be very hard, and might be good git practice. We've been ok with leaving small cleanups like this in the past, so your call.

corlewis:
Previous suggestions and other architectures have now been completed.

lsf37 (approved):
Very happy with this.

Xaphiosis (approved):

Xaphiosis:
Suggestion, if you have one-each per arch crefine commit, might as well squash it into one `crefine:` commit.

corlewis:
> Suggestion, if you have one-each per arch crefine commit, might as well squash it into one `crefine:` commit.

Thanks for the reminder, I meant to ask if that's the preference. If the tests still pass I'll do that squash and then merge.

PR #647, opened 2023-06-21
github: use explicit token to enable push triggers

lsf37:
The implicit GITHUB_TOKEN does not trigger further push actions in the same repo, but in this case we do want the push action to happen on the `-rebased` branches, so we use an explicit auth token instead.

### DIFF
--- a/.github/workflows/proof-deploy.yml
+++ b/.github/workflows/proof-deploy.yml
@@ -91,6 +91,9 @@ jobs:
         ref: ${{ matrix.branch }}
         path: l4v-${{ matrix.branch }}
         fetch-depth: 0
+        # needed to trigger push actions on the -rebased branch
+        # (implict GITHUB_TOKEN does not trigger further push actions).
+        token: ${{ secrets.CI_SSH }}
     - name: Rebase
       run: |
         cd l4v-${{ matrix.branch }}


Xaphiosis (approved):
Does this mean the entire repo now becomes vulnerable to the infinite loop problem, or is it limited to the rebased branches?

lsf37:
> Does this mean the entire repo now becomes vulnerable to the infinite loop problem, or is it limited to the rebased branches?

Only this specific job in this specific workflow, i.e. only the `-rebased` branches, yes.

Xaphiosis:
While I'm here, what is the intended outcome for having both `exynos5-ver` and `exynos5-ver-rebased` branches (same for `imx8-fpu`) in `l4v`? I suspect I created (one of?) the latter, but are they all intended to stay?

lsf37:
> While I'm here, what is the intended outcome for having both `exynos5-ver` and `exynos5-ver-rebased` branches (same for `imx8-fpu`) in `l4v`? I suspect I created (one of?) the latter, but are they all intended to stay?

They are all intended to stay. The `imx8-fpu-ver` branch is the main branch that is tested and working, the `-rebased` branch is the latest version under test and may or may not be working.

PR #649, opened 2023-06-23
add corres_cases method

lsf37:
Add safe datatype case distinction for corres -- wpc turns out to be  insufficient even after generalisation of the helper predicate.
    
- `corres_cases_left`: case distinction on abstract monad
- `corres_cases_right`: case distinction on concrete monad
- `corres_cases`: case distinction first of abstract/concrete monad
- `corres_cases_both`: simultaneous (quadratic) case distinction on both sides, with safe elimination of trivially contradictory cases

Main intended benefit is less thinking about safety of schematics, fewer mentions of goal parameter names, and fewer manual guard instantiations.

The PR also removes the old defunct wpc setup for corres and changes a few proofs in refine/ARM to demonstrate how the method is intended to be used.

### DIFF
--- a/lib/Corres_Cases.thy
+++ b/lib/Corres_Cases.thy
@@ -1,0 +1,323 @@
+(*
+ * Copyright 2023, Proofcraft Pty Ltd
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ *)
+
+theory Corres_Cases
+imports Corres_UL
+begin
+
+text \<open>
+  This file defines the following main methods for safe data type case distinctions on
+  corres/corres_underlying predicates.
+
+  \<^item> corres_cases_left: case distinction on abstract monad
+  \<^item> corres_cases_right: case distinction on concrete monad
+  \<^item> corres_cases: try corres_cases_left, then corres_cases_right
+  \<^item> corres_cases_both: simultaneous (quadratic) case distinction on both sides, with safe
+                       elimination of trivially contradictory cases.
+
+  The first 3 methods take no arguments, corres_cases_both takes an optional simp argument to,
+  for example, unfold relations that synchronise cases between the abstract and concrete side.
+
+  The case distinctions work if the entire monad is a "case" statement, or if the monad is a
+  @{const bind} or @{const bindE} term with a "case" statement in the head position.
+
+  There is an existing method for case distinctions (@{method wpc}), but this method is not
+  flexible enough for @{term corres}: consider the goal
+  @{text "\<And>x. corres r (?G x) ?G' (case x of None \<Rightarrow> a | Some y \<Rightarrow> b y) m"} -- if we perform
+  case distinction on @{term x}, then we can transform @{text "?G x"} into
+  @{text "\<lambda>x s. (x = None \<longrightarrow> ?Q1 x s) \<and> (\<forall>y. x = Some y \<longrightarrow> ?Q2 x y s)"},
+  but we cannot do the same with @{text "?G'"}, because @{text "?G'"} does not depend on @{text x}.
+  The best we can do is @{text "?G' = \<lambda>s. ?A s \<and> ?B s"}, which so far seems to be good enough
+  in our manual proofs.
+
+  The @{method wpc} method will try to treat both preconditions uniformly and fail on @{text "?G'"}.
+  Extending @{method wpc} to deal with guards in a non-uniform way would be possible, but would make
+  setup for new constants even more messy than it already is. Instead we re-use the general idea
+  here (in Eisbach instead of ML), and leave the @{method wpc} setup clean for other uses.
+\<close>
+
+section \<open>Helper functions and definitions\<close>
+
+(* The following three definitions are originally by Dan Matichuck from the Eisbach
+   Corres_Method example *)
+
+(* Retrieve a split rule for a target term that is expected to be a case statement. *)
+ML \<open>
+fun get_split_rule ctxt target =
+let
+  val (hdTarget, args) = strip_comb (Envir.eta_contract target)
+  val (constNm, _) = dest_Const hdTarget
+  val constNm_fds = String.fields (fn c => c = #".") constNm
+
+  val _ = if String.isPrefix "case_" (List.last constNm_fds) then ()
+          else raise TERM ("Not a case statement", [target])
+
+  val typeNm = (String.concatWith "." o rev o tl o rev) constNm_fds
+  val split = Proof_Context.get_thm ctxt (typeNm ^ ".split")
+  val vars = Term.add_vars (Thm.prop_of split) []
+
+  val datatype_name = List.nth (rev constNm_fds, 1)
+
+  fun T_is_datatype (Type (nm, _)) = (Long_Name.base_name nm = Long_Name.base_name datatype_name)
+    | T_is_datatype _ = false
+
+  val datatype_var =
+    case find_first (fn ((_, _), T') => T_is_datatype T') vars of
+      SOME (ix, _) => ix
+    | NONE => error ("Couldn't find datatype in thm: " ^ datatype_name)
+
+  val split' = Drule.infer_instantiate ctxt
+                                       [(datatype_var, Thm.cterm_of ctxt (List.last args))] split
+
+in SOME split' end
+handle TERM _ => NONE;
+\<close>
+
+(* The above function as an attribute. The term argument is expected to be a case statement. *)
+attribute_setup get_split_rule = \<open>Args.term >>
+  (fn t => Thm.rule_attribute [] (fn context => fn _ =>
+      case get_split_rule (Context.proof_of context) t of
+        SOME thm => thm
+      | NONE => Drule.free_dummy_thm))\<close>
+
+(* Apply a split rule to a goal. Example usage:
+
+   apply_split f "\<lambda>f. corres_underlying sr nf nf' r P P' f f'"
+
+   The first (free) f is expected to be a case statement and is used to extract the split rule.
+   The second term is expected to take this f as a parameter and provide the term context of the
+   case statement in the goal so the split rule is applied to the correct occurrence of the case
+   statement.
+*)
+method apply_split for f :: 'a and R :: "'a \<Rightarrow> bool" =
+    (match [[get_split_rule f]] in U: "(?x :: bool) = ?y" \<Rightarrow>
+      \<open>match U[THEN iffD2] in U': "\<And>H. ?A \<Longrightarrow> H (?z :: 'c)" \<Rightarrow>
+        \<open>match (R) in "R' :: 'c \<Rightarrow> bool" for R' \<Rightarrow>
+          \<open>rule U'[where H=R']\<close>\<close>\<close>)
+
+context
+begin
+
+(* This predicate provides an abstraction for guard/precondition terms for transformations
+   on those guards.
+
+   P and P' are the abstract and concrete preconditions before the transformation
+   Q and Q' are the abstract and concrete preconditions after the transformation
+
+   R is the predicate to be transformed.
+*)
+private definition corres_case_helper ::
+  "(('a \<Rightarrow> bool) \<times> ('b \<Rightarrow> bool)) \<Rightarrow> (('a \<Rightarrow> bool) \<times> ('b \<Rightarrow> bool)) \<Rightarrow> bool \<Rightarrow> bool" where
+  "corres_case_helper \<equiv> \<lambda>(P, P') (Q, Q') R. (\<forall>s. P s \<longrightarrow> Q s) \<longrightarrow>  (\<forall>s. P' s \<longrightarrow> Q' s) \<longrightarrow> R"
+
+
+(* The following lemmas enable us to lift preconditions of corres_case_helper over conjunction,
+   universal quantifiers, and implication. Note that there are strong versions for forall/implies
+   where both guards are treated uniformly, and weak versions, where forall/implies is dropped
+   in one guard, but not the other.
+
+   The collection of the lemmas below is used to process the term R in corres_case_helper and
+   create appropriately lifted guard/preconditions during that procedure. The names and general
+   idea are from the WPC theory.
+*)
+
+private lemma corres_case_helperI:
+  "corres_case_helper (P, P') (P, P') R \<Longrightarrow> R"
+  by (simp add: corres_case_helper_def)
+
+private lemma corres_case_conj_process:
+  "\<lbrakk> corres_case_helper (P, P') (A, A') R; corres_case_helper (P, P') (B, B') S \<rbrakk>
+   \<Longrightarrow> corres_case_helper (P, P') (\<lambda>s. A s \<and> B s, \<lambda>s. A' s \<and> B' s) (R \<and> S)"
+  by (clarsimp simp add: corres_case_helper_def)
+
+private lemma corres_case_all_process:
+  "\<lbrakk> \<And>x. corres_case_helper (P, P') (Q x, Q' x) (R x) \<rbrakk>
+   \<Longrightarrow> corres_case_helper (P, P') (\<lambda>s. \<forall>x. Q x s, \<lambda>s. \<forall>x. Q' x s) (\<forall>x. R x)"
+  by (clarsimp simp: corres_case_helper_def subset_iff)
+
+private lemma corres_case_all_process_weak:
+  "\<lbrakk> \<And>x. corres_case_helper (P, P') (Q x, Q') (R x) \<rbrakk>
+   \<Longrightarrow> corres_case_helper (P, P') (\<lambda>s. \<forall>x. Q x s, Q') (\<forall>x. R x)"
+  by (clarsimp simp: corres_case_helper_def subset_iff)
+
+private lemma corres_case_imp_process:
+  "\<lbrakk> S \<Longrightarrow> corres_case_helper (P, P') (Q, Q') R \<rbrakk>
+   \<Longrightarrow> corres_case_helper (P, P') (\<lambda>s. S \<longrightarrow> Q s, \<lambda>s. S \<longrightarrow> Q' s) (S \<longrightarrow> R)"
+  by (clarsimp simp add: corres_case_helper_def subset_iff)
+
+private lemma corres_case_imp_process_weak:
+  "\<lbrakk> S \<Longrightarrow> corres_case_helper (P, P') (Q, Q') R \<rbrakk>
+   \<Longrightarrow> corres_case_helper (P, P') (\<lambda>s. S \<longrightarrow> Q s, Q') (S \<longrightarrow> R)"
+  by (clarsimp simp add: corres_case_helper_def subset_iff)
+
+private lemmas corres_case_process =
+  corres_case_conj_process corres_case_all_process corres_case_imp_process
+
+private lemmas corres_case_process_weak =
+  corres_case_conj_process corres_case_all_process_weak corres_case_imp_process_weak
+
+(* Turn goals of the form
+
+   (\<forall>y. x = SomeConstr y \<longrightarrow> corres (?P x) P' (SomeConstr y) g) \<and>
+   (\<forall>y. x = OtherConstr y \<longrightarrow> corres (?P x) P' (OtherConstr y) g) \<and>
+   ...
+
+   into multiple goals of the form
+
+   \<And>y. x = SomeConstr y \<Longrightarrow> corres (?P1 x y) ?P'1 (SomeConstr y) g)
+   \<And>y. x = OtherConstr y \<Longrightarrow> corres (?P2 x y) ?P'2 (OtherConstr y) g)
+
+   with instantiations
+
+   ?P x = \<lambda>s. (\<forall>y. x = SomeConstr y \<longrightarrow> ?P1 x y s) \<and> (\<forall>y. x = OtherConstr y \<longrightarrow> ?P2 x y s)
+   ?P' = \<lambda>s. ?P'1 s \<and> ?P'2 s
+
+   We do this by first transforming the goal into a corres_case_helper goal, and then applying
+   the corresponding lifting rules. We first try to get both sides (?P and ?P') to have
+   quantifiers and implications to get a stronger statement, and fall back to the weaker \<and> for ?P'
+   shown above when that doesn't work (e.g. because ?P' might not depend on x).
+
+   When all lifting rules have applied, we transform the goal back into a corres goal using the
+   provided helper rule (e.g. corres_case_helper_corres_left below).
+*)
+private method corres_cases_body uses helper =
+  determ \<open>rule corres_case_helperI, repeat_new \<open>rule corres_case_process\<close>; rule helper
+          | rule corres_case_helperI, repeat_new \<open>rule corres_case_process_weak\<close>; rule helper\<close>
+
+
+(* Instances of corres_case_helper for left and right side of the corres predicate.
+   These lemmas bind the corres guards to the corres_case_helper guards. *)
+private lemma corres_case_helper_corres_left:
+  "corres_underlying sr nf nf' r Q Q' f f' \<Longrightarrow>
+    corres_case_helper (P, P') (Q, Q') (corres_underlying sr nf nf' r P P' f f')"
+  by (auto simp: corres_case_helper_def elim!: corres_guard_imp)
+
+private lemma corres_case_helper_corres_right:
+  "corres_underlying sr nf nf' r Q' Q f f' \<Longrightarrow>
+    corres_case_helper (P, P') (Q, Q') (corres_underlying sr nf nf' r P' P f f')"
+  by (auto simp: corres_case_helper_def elim!: corres_guard_imp)
+
+
+section \<open>Main method definitions\<close>
+
+(* Case distinction on abstract side *)
+method corres_cases_left =
+  determ \<open>
+    corres_pre,
+    (match conclusion in
+      "corres_underlying sr nf nf' r P P' (f >>= g) f'" for sr nf nf' r P P' f g f'
+         \<Rightarrow> \<open>apply_split f "\<lambda>f. corres_underlying sr nf nf' r P P' (f >>= g) f'"\<close>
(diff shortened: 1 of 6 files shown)


lib/Corres_Method.thy
```diff
@@ -5,22 +5,9 @@
  *)
 
 theory Corres_Method
-imports Corres_UL SpecValid_R
+imports Corres_Cases SpecValid_R
 begin
 
-(*TODO move this *)
-
-method_setup repeat_new =
```
Xaphiosis, line 13: this got removed in this commit, but I've not seen it moved... does that mean we have an existing `repeat_new` this was shadowing?

lib/Corres_Cases.thy
```diff
@@ -0,0 +1,319 @@
...
+imports Corres_UL
+begin
+
+text \<open>
+  This file defines the following main methods for safe data type case distinctions on
+  corres/corres_underlying predicates.
+
+  \<^item> corres_cases_left: case distinction on abstract monad
+  \<^item> corres_cases_right: case distinction on concrete monad
+  \<^item> corres_cases: case distinction first of abstract/concrete monad
```
Xaphiosis, line 17: I am not clear on what this means here or in the commit message: "case distinction first of abstract/concrete monad"
Does this mean it will apply it to the first side it is able to out of those two, starting with abstract side and if that fails moving on to concrete side? If so, this could be phrased a bit more clearly.

lib/Corres_Cases.thy
```diff
@@ -0,0 +1,319 @@
...
+  \<^item> corres_cases_both: simultaneous (quadratic) case distinction on both sides, with safe
+                       elimination of trivially contradictory cases.
+
+  The first 3 methods take no arguments, corres_cases_both takes an optional simp argument to,
+  for example, unfold relations that synchronises cases between abstract and concrete side.
```
Xaphiosis, line 22: "synchronise cases"?

lib/Corres_Cases.thy
```diff
@@ -0,0 +1,319 @@
...
+
+  The case distinctions work if the entire monad is a "case" statement, or if the monad is a
+  @{const bind} or @{const bindE} term with a "case" statement in the head position.
+
+  There is an existing method for case distinctions (@{method wpc}), but this method is not
+  flexible enough for @{term corres}. The reason for this is that @{method wpc}
+  will always treat abstract and concrete guards the same way. This doesn't always work, because
+  the abstract guard usually only depends on abstract values and the concrete guard only on
+  concrete values. If we make a case distinction on such an abstract value, the concrete guard
+  cannot to transformed into one that follows that case distinction, so @{method wpc} will fail.
```
Xaphiosis, line 32: "cannot be transformed"?

Xaphiosis: bit hazy on meaning of "the concrete guard cannot to transformed into one that follows that case distinction"

lib/Corres_Cases.thy
```diff
@@ -0,0 +1,319 @@
...
+  The best we can do in this case is give each concrete case a different guard (but still one that
```
Xaphiosis, line 33: "but one that still"?

lib/Corres_Cases.thy
```diff
@@ -0,0 +1,319 @@
...
+  fun T_is_datatype (Type (nm,_)) = (Long_Name.base_name nm) = (Long_Name.base_name datatype_name)
+    | T_is_datatype _ = false
+
+  val datatype_var =
+    case (find_first (fn ((_,_),T') => (T_is_datatype T')) vars) of
+      SOME (ix,_) => ix
+    | NONE => error ("Couldn't find datatype in thm: " ^ datatype_name)
+
+  val split' = Drule.infer_instantiate ctxt
+    [(datatype_var, Thm.cterm_of ctxt (List.last args))] split
```
Xaphiosis, line 75: trivial: indent under `Drule.` while we're here

lib/Corres_Cases.thy
```diff
@@ -0,0 +1,319 @@
...
+      case (get_split_rule (Context.proof_of context) t) of
+        SOME thm => thm
+      | NONE => Drule.free_dummy_thm))\<close>
+
+(* Apply a split rule to a goal. Example usage:
+
+   apply_split f "\<lambda>f. corres_underlying sr nf nf' r P P' f f'"
+
+   The corres_underlying term provides the context, and f is the term we expect to be
+   a case statement.  *)
```
Xaphiosis, line 94: does the f in `f` and f in `\lambda f. ?whatever f` have the same significance relative to the `f` mentioned in "and f is the term we expect ..."?

lib/Corres_Cases.thy
```diff
@@ -0,0 +1,319 @@
...
+   universal quantifiers, and implication. Note that there are strong versions for forall/implies
+   where both guards are treated uniformly, and weak versions, where forall/implies is dropped
+   in one guard, but not the other.
+*)
+
+private lemma corres_case_helperI:
+  "corres_case_helper (P, P') (P, P') R \<Longrightarrow> R"
+  by (simp add: corres_case_helper_def)
+
+private lemma corres_case_conj_process:
```
Xaphiosis, line 127: I'm not clear why these are all named `_process` ... is that because they're meant to be used for processing of the ... er, I'm not sure what.

Xaphiosis: Ah, R, the term to be transformed?

lib/Corres_Cases.thy
```diff
@@ -0,0 +1,319 @@
...
+private lemmas corres_case_process =
+  corres_case_conj_process corres_case_all_process corres_case_imp_process
+
+private lemmas corres_case_process_weak =
+  corres_case_conj_process corres_case_all_process_weak corres_case_imp_process_weak
+
+(* Turn goals of the form
+
+   (\<forall>y. x = SomeConstr y \<longrightarrow> corres (?P x) P' (SomeConstr y) g) \<and>
+   (\<forall>y. x = OtherConstr y \<longrightarrow> corres (?P x) P' (SomeConstr y) g) \<and>
```
Xaphiosis, line 161: is the second one supposed to have OtherConstr y and SomeConstr y?

lib/Corres_Cases.thy
```diff
@@ -0,0 +1,319 @@
...
+   \<And>y. x = SomeConstr y \<Longrightarrow> corres (?P1 x y) ?P'1 (SomeConstr y) g)
+   \<And>y. x = OtherConstr y \<Longrightarrow> corres (?P2 x y) ?P'2 (SomeConstr y) g)
+
+   with instantiations
+
+   ?P x = \<lambda>s. (\<forall>y. x = SomeConstr y \<longrightarrow> ?P1 x y s) \<and> (\<forall>y. x = OtherConstr y \<longrightarrow> ?P2 x y s)
+   ?P' = \<lambda>s. ?P'1 s \<and> ?P'2 s
+
+   We do this by first transforming the goal into a corres_case_helper goal, and then applying
+   the corresponding lifting rules. We first try if we can get both sides (?P and ?P') to have
```
Xaphiosis, line 175: "try if" -> "try to"?

lib/Corres_Cases.thy
```diff
@@ -0,0 +1,319 @@
...
+  by (auto simp: corres_case_helper_def elim!: corres_guard_imp)
+
+
+section \<open>Main method definitions\<close>
+
+(* Case distinction on abstract side *)
+method corres_cases_left =
+  determ \<open>
+    corres_pre,
+    (match conclusion in
```
Xaphiosis, line 206: how are you managing to do an Eisbach `match` with schematics around? I thought `match` forced variable focus

lib/Corres_Cases.thy
```diff
@@ -0,0 +1,319 @@
...
+
+(* Case distinction on abstract and concrete side with quadractic blowup, but attempt to solve
+   contradictory side conditions by simp. Cases that are solved by simp will produce \<top> as guard
+   so that no free schematics are introduced into later goals. *)
+method corres_cases_both uses simp =
+  (* corres_pre first, so that the ";" later only refers to corres goals, not the final implications *)
+  determ \<open>
+    corres_pre,
+    (corres_cases_left; corres_cases_right;
+     (solves \<open>rule corres_inst[where P=\<top> and P'=\<top>], simp add: simp\<close>)?)\<close>
```
Xaphiosis, line 239: This one is really interesting, and I think very clever.

lib/Corres_Cases.thy
```diff
@@ -0,0 +1,319 @@
...
+(* some example relation between abstract and concrete values *)
+definition
+  "none_rel x y \<equiv> (x = None) = (y = None)"
+
+lemma
+  "none_rel x y \<Longrightarrow>
+   corres_underlying srel nf nf' rrel G G' (case x of None \<Rightarrow> a | Some y \<Rightarrow> b)
+                                            (case y of None \<Rightarrow> a' | Some y \<Rightarrow> b' y)"
+  (* two cases remain, none_rel is untouched in the cases that remain, but unfolded in the
+     ones that were eliminated *)
```
Xaphiosis, line 313: cool

lib/Corres_Cases.thy
```diff
@@ -0,0 +1,319 @@
...
+  apply (corres_cases_both simp: none_rel_def)
+  oops
+
+end
```
Xaphiosis, line 317: I'm never sure when we cross the threshold into needing tests to be placed in a separate file. This is short, and the extra definition (`none_rel`) shouldn't leak, so all good, but I'm curious as to your opinion on where that threshold lies.

lib/Corres_Cases.thy
Xaphiosis: Would `_lift` make sense, given the description above says `enable us to lift`?

Xaphiosis (approved):
I am happy with the methods and the improvement they provide. Nice! I also like that they manage to do what they do concisely. If I tried to to it, it could've ended up an over-engineered mess. This approach ends up very neat.

I do have some difficulty following some of the textual descriptions, and wonder if I'm too ignorant to be the target audience, or whether the descriptions could become easier. A few Englishisms snuck in there as well.

lib/Corres_Cases.thy
lsf37: Oh, very well spotted, that should definitely be both `OtherConstr`

Xaphiosis: Ah, it was a bit confusing. Same on the lines below then!

lib/Corres_Method.thy
lsf37: yes, apparently `repeat_new` was added ages ago already and the move in here was forgotten/never noticed

lib/Corres_Cases.thy
lsf37: That is indeed what it is supposed to mean. Will rephrase.

lib/Corres_Cases.thy
lsf37: The lambda describes where the `f` is expected inside the bigger term and the `f` itself is the term we expect to be in the form of a case statement.

lib/Corres_Cases.thy
lsf37: The terminology is from `wpc`, which has almost the same lemmas, and calls them `wpc_helper_X_process`. `_lift` would probably be a better name, but then that connection is lost. I could add a comment that this is where the names come from.

lib/Corres_Cases.thy
lsf37: This is actually even a tiny bit more powerful than I thought it might be, because we have global simp rules that will solve trivial `corres` statements like `corres .. (return x) (return x)`, so these might also disappear, but only if they are solvable with a trivial guard. I only found that out later in the VSpace examples.

lib/Corres_Cases.thy
lsf37: This is pretty close to that line. The section used to be much shorter, but then I introduced the >>E and then corres_cases_both and then the test with schematics and now it's 3 times as long..

lib/Corres_Cases.thy
lsf37: Interesting, I adapted that from Dan's code without thinking about schematics. You're right that we'd usually expect them to be frozen. I *think* what is going on is that the pattern matches the schematics explicitly with `for .. P P' ..` which then makes them schematic again on the inside. All other schematics would remain frozen, but that is even  beneficial in this context, because we would not want to touch those.

Xaphiosis: My concern is that "freezing" a schematic *is* touching it, i.e. you cannot instantiate it after it is frozen, which is why we usually can't use `match` in wp proofs. This would indicate that there is a workaround available, but it will only work if the match can mention and re-generalise the schematics directly? Something to think about for sure.

If we end up with a schematic in the assumptions, probably that's a mistake. We really shouldn't see any schematics in the corres bodies... so that leaves the return relation. I don't immediately know of any cases where we'd want to do `corres_cases` with a schematic return relation. What do you think?

lib/Corres_Cases.thy
lsf37: I was trying to be concise, but that's probably a mistake. The situation is that for terms like

```
  "⋀x. corres .. (?G x) ?G' (case x of ..) m"
```

if you do case distinction on `x`, you can transform `?G` into e.g. `λx s. (x = None ⟶ ?G1 x s) ∧ (∀y. x = Some y ⟶ ?G2 x y s)`, but you can't do that with `?G'` because `?G'` doesn't depend on `x`. The best you can get for `?G'` is just `?G1' and ?G2'`.

lib/Corres_Cases.thy
Xaphiosis: Would help, yes, thank you

lib/Corres_Cases.thy
Xaphiosis: But if we have a bind, then `case ...  >>= return ...` would be `f` and not a case statement? Am I being obtuse here?

lib/Corres_Cases.thy
Xaphiosis: Ah, that makes sense. I think the original sentence is a bit over-compressed, explaining what "following" "that" case distinction means definitely helped me here.

lib/Corres_Cases.thy
lsf37: The freezing is only inside the scope of the `match`, and once the `match` is complete they unfreeze, so if we don't need to interact with them, they should be unchanged. That means schematic return relations should be fine.

For implementing `wp` that is not useful enough, because there we'd usually want them to stay schematic on the inside of the `match` scope, but you are right that explicitly matching them in patterns like `for .. P P' ..` might be a workaround for that. Definitely worth keeping in mind.

lib/Corres_Cases.thy
lsf37: The case for bind looks like this:

```
apply_split f "λf. corres_underlying sr nf nf' r P P' (f >>= g) f'"
```

so the `f` is still the case statement, but the larger context term has changed to include the bind.

Xaphiosis: Hmm, I find that a bit confusing. It's a lot of `f`.

lsf37: There are only two `f``, one bound, one free. The free one is expected to be `case ..` the bound one says where in the goal that `case ..` term is.

lib/Corres_Cases.thy
lsf37: On second reading, there is unfortunately nothing new here at all: the `for .. P P' ..` doesn't make the `P` schematic inside the `match`, they just become fixed variables. What's going on here is that the `match` scope is small, it only covers `apply_split`, which figures out and the split rule, but does not touch any schematics. The precondition processing happens after all that, when the `match` statement is concluded and the schematics are available again.

lib/Corres_Cases.thy
corlewis: It's not for this PR, but it might be worth auditing those simp rules at some point and thinking about whether they should have schematic or fixed preconditions.

It doesn't normally break proofs but I've had instances before where I was annoyed about schematics propagating further than expected due to a trivial `corres` statement not instantiating its preconditions to `⊤`.

lsf37:
(updated with Raf's review feedback)

lib/Corres_Cases.thy
lsf37: (have updated the explanation with a bit more detail)

lib/Corres_Cases.thy
lsf37: Yes, it's super annoying, and in the upcoming `corres` method they definitely cause trouble (Dan encountered the same and has a set of "simp rules that break everything"). I was too afraid of a big update to try to get rid of them properly, because I have the feeling they are used almost everywhere where we have a final `return` or `gets`.

Might be best to schedule that update for when we have arch split Refine, then we at least will have to do only one pass and not five.

lib/Corres_Cases.thy
```diff
@@ -0,0 +1,325 @@
...
+  corres/corres_underlying predicates.
+
+  \<^item> corres_cases_left: case distinction on abstract monad
+  \<^item> corres_cases_right: case distinction on concrete monad
+  \<^item> corres_cases: try corres_cases_left, then corres_cases_right
+  \<^item> corres_cases_both: simultaneous (quadratic) case distinction on both sides, with safe
+                       elimination of trivially contradictory cases.
+
+  The first 3 methods take no arguments, corres_cases_both takes an optional simp argument to,
+  for example, unfold relations that synchronise cases between abstract and concrete side.
```
Xaphiosis, line 22: between -> between the ?

lib/Corres_Cases.thy
```diff
@@ -0,0 +1,325 @@
...
+  The case distinctions work if the entire monad is a "case" statement, or if the monad is a
+  @{const bind} or @{const bindE} term with a "case" statement in the head position.
+
+  There is an existing method for case distinctions (@{method wpc}), but this method is not
+  flexible enough for @{term corres}: consider the goal
+  @{text "\<And>x. corres r (?G x) ?G' (case x of None \<Rightarrow> a | Some y \<Rightarrow> b y) m"} -- if we perform
+  case distinction on @{term x}, then we can transform @{text "?G x"} into
+  @{text "\<lambda>x s. (x = None \<longrightarrow> ?Q1 x s) \<and> (\<forall>y. x = Some y \<longrightarrow> ?Q2 x y s)"},
+  but we cannot do the same with @{text "?G'"}, because @{text "?G'"} does not depend on @{text x}.
+  The best we can do is @{text "?G' \<leadsto> \<lambda>s. ?A s \<and> ?B s"}, which so far seems to be good enough
```
Xaphiosis, line 33: what does the squiggly right arrow signify?

lib/Corres_Cases.thy
```diff
@@ -0,0 +1,325 @@
...
+  in our manual proofs.
+
+  The @{method wpc} method will try to treat both preconditions the same and fail on @{text "?G'"}.
+  Extending @{method wpc} so far that it can deal with both situations (uniform treatment of
```
Xaphiosis, line 37: Extending X so far that it -> Extending X until it

lib/Corres_Cases.thy
```diff
@@ -0,0 +1,325 @@
...
+  "corres_case_helper \<equiv> \<lambda>(P, P') (Q, Q') R. (\<forall>s. P s \<longrightarrow> Q s) \<longrightarrow>  (\<forall>s. P' s \<longrightarrow> Q' s) \<longrightarrow> R"
+
+
+(* The following lemmas enable us to lift preconditions of corres_case_helper over conjunction,
+   universal quantifiers, and implication. Note that there are strong versions for forall/implies
+   where both guards are treated uniformly, and weak versions, where forall/implies is dropped
+   in one guard, but not the other.
+
+   The collection of the lemmas below is used to process the term R in corres_case_helper and
+   created appropriately lifted guard/preconditions during that procedure. The names and general
```
Xaphiosis, line 125: created -> create

lib/Corres_Cases.thy
Xaphiosis: It seems I do not understand variable focus then. My impression was if you focus, *all* schematics freeze (become fixed variables), and then you can't apply schematic manipulation functions as in `wp` goals. I was not aware that if you apply a rule that does not affect the schematic, after the entire application (match + rule) is applied, the schematics unfreeze ... is that right?

Xaphiosis:
Last typos, otherwise I'm happy with it, the explanations are now ones I can follow. Thank you!

lib/Corres_Cases.thy
lsf37: > It seems I do not understand variable focus then. My impression was if you focus, _all_ schematics freeze (become fixed variables), and then you can't apply schematic manipulation functions as in `wp` goals. 

That's exactly right.

> I was not aware that if you apply a rule that does not affect the schematic, after the entire application (match + rule) is applied, the schematics unfreeze ... is that right?

The schematics always unfreeze after the match, not matter what you do. The effect is that inside the match you can't instantiate schematics that existed when you matched, but you can do everything else. The main thing that it is trying to make sure is that the focussed proof does not affect state outside the focussed proof (which instantiated schematics would).

lib/Corres_Cases.thy
lsf37: It's the Isabelle notation for instantiation (you can see it for instance in schematic_goal)

lib/Corres_Cases.thy
lsf37: Hm, "until" doesn't sound right to me. Extending `wpc` is not a process that goes on until we reach enough generality. I'm more trying to express that including the feature in `wpc` would stretch the concepts there by quite a bit and make things messy. It'd be completely possible, but it wouldn't be nice.

Xaphiosis: Your version nears German, so if you want that, maybe:
`Overextending X to deal with both situations?`

lib/Corres_Cases.thy
Xaphiosis: Learning all the time!

lib/Corres_Cases.thy
lsf37: Ok, have reformulated and shrunk that entire paragraph. No more stretching :-)

corlewis: As another alternative, maybe `Extending @{method wpc} enough so that it could deal with both situations`?

lib/Corres_Cases.thy
Xaphiosis: OK, that makes `match` less useless in schematic goals than I thought! It means that you could maybe use them to apply a rule to a `wp` goal to modify the head of the bind, for example, so long as it can deal with the schematics being frozen. That opens up a new avenue of thinking that I didn't think was possible, thank you so much for explaining.

lib/Corres_Cases.thy
corlewis: That's a new one for me as well!

lib/Corres_Cases.thy
corlewis: Sorry, looks like I should have refreshed first, it reads well to me now.
Keeping my pedantic hat on, I think my only other suggestion would be to split the sentence up
```
 even more messy than it already is. We instead re-use the general idea here (in Eisbach instead
  of ML), and leave the @{method wpc} setup clean for other uses.
  ```

File: lib/Corres_Cases.thy [hunk 19]
```diff
@@ -0,0 +1,323 @@
...
+    | NONE => error ("Couldn't find datatype in thm: " ^ datatype_name)
+
+  val split' = Drule.infer_instantiate ctxt
+                                       [(datatype_var, Thm.cterm_of ctxt (List.last args))] split
+
+in SOME split' end
+handle TERM _ => NONE;
+\<close>
+
+(* The above function as an attribute. The term argument is expected to be a case statement. *)
```
corlewis, line 79: This might be a failing of my expectations, but this comment took me quite a while to understand. My brain kept on trying to interpret 'function' as a verb, which doesn't make sense for any potential interpretation. It's probably fine to leave it as is, but I think something like `The following is the above function as an attribute` would fix it for anyone who has the same problem as me.

corlewis (approved):
Very cool work, the examples demonstrating the reduced need for manual instantiation are particularly encouraging.

My only remaining confusion is why `wpc` seems to work in crefine. Do we just avoid the cases where the guards don't depend on the bound variable being split? If so that suggests we might want something very similar to this for `ccorres` as well.

lsf37:
> My only remaining confusion is why `wpc` seems to work in crefine. Do we just avoid the cases where the guards don't depend on the bound variable being split? If so that suggests we might want something very similar to this for `ccorres` as well.

I was thinking that the problem doesn't occur in `ccorres`, because we don't have a normal `rv` thing going on on the C side, and I thought that eliminates the problem. But a) we do have things like symbolic execution, and b) that actually makes the problem worse, because the abstract guard will depend on abstract goal parameters, and the C guard will not, which is exactly the case I'm trying to solve.

So I think you're right and we should probably be doing the same thing for `ccorres`. Do we have cases where `wpc` mysteriously doesn't work on `ccorres`? (because that's probably because of exactly this issue).

lib/Corres_Cases.thy
lsf37: Should I replace it with `=` instead? If you both haven't seen it, it's likely that others haven't either.

corlewis:
> So I think you're right and we should probably be doing the same thing for `ccorres`. Do we have cases where `wpc` mysteriously doesn't work on `ccorres`? (because that's probably because of exactly this issue).

I don't know about one with `wpc`, but I have definitely had cases in `ccorres` where I wanted to split a concrete `IF .. THEN .. ELSE ..` and the abstract precondition didn't have the required bound variable.

lsf37:
> > So I think you're right and we should probably be doing the same thing for `ccorres`. Do we have cases where `wpc` mysteriously doesn't work on `ccorres`? (because that's probably because of exactly this issue).
> 
> I don't know about one with `wpc`, but I have definitely had cases in `ccorres` where I wanted to split a concrete `IF .. THEN .. ELSE ..` and the abstract precondition didn't have the required bound variable.

Right, that's another case we should treat. I should probably start collecting these in an issue.

PR #650, opened 2023-06-23
Improve `corres` rules `whileLoop`

michaelmcinerney:
Please see the commit message for a brief explanation

### DIFF
--- a/lib/ExtraCorres.thy
+++ b/lib/ExtraCorres.thy
@@ -254,17 +254,20 @@ lemma hoare_from_abs_inv:
 lemma in_whileLoop_corres:
   assumes body_corres:
     "\<And>r r'. rrel r r' \<Longrightarrow>
-             corres_underlying srel False nf' rrel (P and C r) (P' and C' r') (B r) (B' r')"
-  and body_inv: "\<And>r. \<lbrace>P and C r\<rbrace> B r \<lbrace>\<lambda>_. P\<rbrace>"
-                "\<And>r'. \<lbrace>P' and C' r'\<rbrace> B' r' \<lbrace>\<lambda>_. P'\<rbrace>"
-  and cond: "\<And>r r' s s'. \<lbrakk>rrel r r'; (s, s') \<in> srel; P s; P' s'\<rbrakk> \<Longrightarrow> C r s = C' r' s'"
-  and result: "(rv', t') \<in> fst (whileLoop C' B' r' s')"
-  shows "\<forall>s r. (s, s') \<in> srel \<and> rrel r r' \<and> P s \<and> P' s'
+             corres_underlying srel nf nf' rrel (P r and C r) (P' r' and C' r') (B r) (B' r')"
+  assumes body_inv:
+    "\<And>r. \<lbrace>P r and C r\<rbrace> B r \<lbrace>P\<rbrace>"
+    "\<And>r'. \<lbrace>P' r' and C' r'\<rbrace> B' r' \<lbrace>P'\<rbrace>"
+  assumes cond: "\<And>r r' s s'. \<lbrakk>rrel r r'; (s, s') \<in> srel; P r s; P' r' s'\<rbrakk> \<Longrightarrow> C r s = C' r' s'"
+  assumes result: "(rv', t') \<in> fst (whileLoop C' B' r' s')"
+  assumes nf: "\<And>r. nf \<Longrightarrow> no_fail (P r and C r) (B r)"
+  shows "\<forall>s r. (s, s') \<in> srel \<and> rrel r r' \<and> P r s \<and> P' r' s'
                 \<longrightarrow> (\<exists>rv t. (rv, t) \<in> fst (whileLoop C B r s) \<and> (t, t') \<in> srel \<and> rrel rv rv')"
   apply (rule in_whileLoop_induct[OF result])
    apply (force simp: cond whileLoop_def)
   apply clarsimp
-  apply (frule (1) corres_underlyingD2[OF body_corres]; (fastforce simp: cond)?)
+  apply (frule (1) corres_underlyingD2[OF body_corres];
+         (fastforce dest: nf simp: cond no_fail_def)?)
   apply clarsimp
   apply (frule use_valid[OF _ body_inv(1)])
    apply (fastforce dest: cond)
@@ -273,21 +276,22 @@ lemma in_whileLoop_corres:
   apply (fastforce simp: whileLoop_def intro: whileLoop_results.intros(3) dest: cond)
   done
 
-lemma corres_whileLoop:
-  assumes cond: "\<And>r r' s s'. \<lbrakk>rrel r r'; (s, s') \<in> srel; P s; P' s'\<rbrakk> \<Longrightarrow> C r s = C' r' s'"
-  and body_corres:
+lemma corres_whileLoop_ret:
+  assumes cond: "\<And>r r' s s'. \<lbrakk>rrel r r'; (s, s') \<in> srel; P r s; P' r' s'\<rbrakk> \<Longrightarrow> C r s = C' r' s'"
+  assumes body_corres:
     "\<And>r r'. rrel r r' \<Longrightarrow>
-             corres_underlying srel False nf' rrel (P and C r) (P' and C' r') (B r) (B' r')"
-  and body_inv: "\<And>r. \<lbrace>P and C r\<rbrace> B r \<lbrace>\<lambda>_. P\<rbrace>"
-                "\<And>r'. \<lbrace>P' and C' r'\<rbrace> B' r' \<lbrace>\<lambda>_. P'\<rbrace>"
-  and rel: "rrel r r'"
-  and nf': "\<And>r'. no_fail (P' and C' r') (B' r')"
-  and termin: "\<And>r' s'. \<lbrakk>P' s'; C' r' s'\<rbrakk> \<Longrightarrow> whileLoop_terminates C' B' r' s'"
-  shows "corres_underlying srel False nf' rrel P P' (whileLoop C B r) (whileLoop C' B' r')"
+             corres_underlying srel False nf' rrel (P r and C r) (P' r' and C' r') (B r) (B' r')"
+  assumes body_inv:
+    "\<And>r. \<lbrace>P r and C r\<rbrace> B r \<lbrace>P\<rbrace>"
+    "\<And>r'. \<lbrace>P' r' and C' r'\<rbrace> B' r' \<lbrace>P'\<rbrace>"
+  assumes rel: "rrel r r'"
+  assumes nf': "\<And>r'. no_fail (P' r' and C' r') (B' r')"
+  assumes termin: "\<And>r' s'. \<lbrakk>P' r' s'; C' r' s'\<rbrakk> \<Longrightarrow> whileLoop_terminates C' B' r' s'"
+  shows "corres_underlying srel False nf' rrel (P r) (P' r') (whileLoop C B r) (whileLoop C' B' r')"
   apply (rule corres_no_failI)
    apply (simp add: no_fail_def)
    apply (intro impI allI)
-   apply (erule_tac I="\<lambda>_ s. P' s"
+   apply (erule_tac I="\<lambda>r' s'. P' r' s'"
                     and R="{((r', s'), r, s). C' r s \<and> (r', s') \<in> fst (B' r s)
                                               \<and> whileLoop_terminates C' B' r s}"
                     in not_snd_whileLoop)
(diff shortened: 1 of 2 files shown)


michaelmcinerney:
Sorry, forced pushed to correct the commit message

lib/ExtraCorres.thy
```diff
@@ -293,82 +295,98 @@ lemma corres_whileLoop:
...
+    apply (frule (1) corres_underlyingD2[OF body_corres], (fastforce dest: nf simp: no_fail_def)+)
     apply (fastforce dest: use_valid intro: body_inv)
     done
 
   show ?thesis
     apply (insert assms helper)
-    apply (clarsimp simp: ex_abs_underlying_def)
+    apply (cases "C' r' s'")
+     apply (fastforce simp: ex_abs_underlying_def)
+    apply (simp add: whileLoop_terminates.intros(1))
```
Xaphiosis, line 331: does it make sense to fold these into a single `cases ...; fastforce simp: ...`?

lib/ExtraCorres.thy
```diff
@@ -293,82 +295,98 @@ lemma corres_whileLoop:
...
+   apply (rule no_fail_pre)
+    apply (rule_tac G="rrel abs_r conc_r" in no_fail_grab_asm)
+    apply (rule corres_u_nofail)
+     apply (fastforce dest: body_corres)
+    apply (fastforce dest: nf)
+   apply (fastforce simp: cond)
   apply (fastforce intro: wf_subset[OF whileLoop_terminates_wf[where C=C']])
   done
 
+lemmas corres_whileLoop_abs =
```
Xaphiosis, line 388: Out of curiosity, when would you reach for the _abs form of the whileLoop rule?

lib/ExtraCorres.thy
```diff
@@ -293,82 +295,98 @@ lemma corres_whileLoop:
...
+                                  \<and> P rv s \<and> (P' conc_r s' \<and> C' conc_r s') \<and> s' = new_s"
+                in no_fail_pre[rotated])
+    apply fastforce
+   apply (rule no_fail_ex_pre)
+   apply (rename_tac abs_r)
+   apply (rule no_fail_pre)
+    apply (rule_tac G="rrel abs_r conc_r" in no_fail_grab_asm)
+    apply (rule corres_u_nofail)
+     apply (fastforce dest: body_corres)
+    apply (fastforce dest: nf)
```
Xaphiosis, line 383: I suspect that the `rule corres_u_nofail` and two subsequent fastforces can be compressed, but if they demonstrate interesting steps, do leave as is.

Xaphiosis (approved):
Having spent a bit of time poking that compare time while loop, this improvement makes a lot of sense!

lib/ExtraCorres.thy
michaelmcinerney: The first cases is really where all the work is done: the second case is somewhat trivial (if the loop condition is false, the loop doesn't do anything, so it therefore does terminate). So I think it's nicer as these two steps

lib/ExtraCorres.thy
michaelmcinerney: I will compress these

lib/ExtraCorres.thy
michaelmcinerney: I think the `_abs` form would be better in almost all circumstances. 

I suppose the non-`_abs` form is more general (it seems possible for the abstract version to not terminate, where as the concrete version might terminate in all relevant situations). 

But the `_abs` form allows you to reason about `no_fail` and `whileLoop_terminates` on the abstract side, and that is often much more convenient, both because abstract functions are generally just nicer overall, and because the data structures are nicer (termination on the abstract side might be an Isabelle list shrinking in size, whereas on the concrete side we might need to argue about heaps of next pointers having no loops, or something).

lib/ExtraCorres.thy
```diff
@@ -241,17 +241,19 @@ lemma hoare_from_abs_inv:
...
-  and body_inv: "\<And>r. \<lbrace>P and C r\<rbrace> B r \<lbrace>\<lambda>_. P\<rbrace>"
-                "\<And>r'. \<lbrace>P' and C' r'\<rbrace> B' r' \<lbrace>\<lambda>_. P'\<rbrace>"
-  and cond: "\<And>r r' s s'. \<lbrakk>rrel r r'; (s, s') \<in> srel; P s; P' s'\<rbrakk> \<Longrightarrow> C r s = C' r' s'"
+             corres_underlying srel nf nf' rrel (P r and C r) (P' r' and C' r') (B r) (B' r')"
+  and body_inv: "\<And>r. \<lbrace>P r and C r\<rbrace> B r \<lbrace>\<lambda>r. P r\<rbrace>"
+                "\<And>r'. \<lbrace>P' r' and C' r'\<rbrace> B' r' \<lbrace>\<lambda>r'. P' r'\<rbrace>"
+  and cond: "\<And>r r' s s'. \<lbrakk>rrel r r'; (s, s') \<in> srel; P r s; P' r' s'\<rbrakk> \<Longrightarrow> C r s = C' r' s'"
   and result: "(rv', t') \<in> fst (whileLoop C' B' r' s')"
-  shows "\<forall>s r. (s, s') \<in> srel \<and> rrel r r' \<and> P s \<and> P' s'
+  and nf: "nf \<Longrightarrow> (\<And>r. no_fail (P r and C r) (B r))"
```
lsf37, line 249: Logically it's equivalent, but Isabelle convention usually is to extend the scope of meta-quantifiers to the entire goal where it makes sense, so that here the `r` becomes more of a goal parameter than an actually quantified variable.

```suggestion
  and nf: "\<And>r. nf \<Longrightarrow> no_fail (P r and C r) (B r)"
```

lib/ExtraCorres.thy
```diff
@@ -293,82 +295,98 @@ lemma corres_whileLoop:
...
-  and cond: "\<And>r r' s s'. \<lbrakk>rrel r r'; (s, s') \<in> srel; P s; P' s'\<rbrakk> \<Longrightarrow> C r s = C' r' s'"
-  and body_inv: "\<And>r. \<lbrace>P and C r\<rbrace> B r \<lbrace>\<lambda>_. P\<rbrace>"
-                "\<And>r'. \<lbrace>P' and C' r'\<rbrace> B' r' \<lbrace>\<lambda>_. P'\<rbrace>"
-  and abs_termination: "\<And>r s. P s \<Longrightarrow> whileLoop_terminates C B r s"
-  and ex_abs: "ex_abs_underlying srel P s'"
+             corres_underlying srel nf nf' rrel (P r and C r) (P' r' and C' r') (B r) (B' r')"
+  and cond: "\<And>r r' s s'. \<lbrakk>rrel r r'; (s, s') \<in> srel; P r s; P' r' s'\<rbrakk> \<Longrightarrow> C r s = C' r' s'"
+  and body_inv: "\<And>r. \<lbrace>P r and C r\<rbrace> B r \<lbrace>\<lambda>r. P r\<rbrace>"
+                "\<And>r'. \<lbrace>P' r' and C' r'\<rbrace> B' r' \<lbrace>\<lambda>r'. P' r'\<rbrace>"
+  and abs_termination: "\<And>r s. (P r s \<and> C r s) \<Longrightarrow> whileLoop_terminates C B r s"
```
lsf37, line 308: Unless there is a specific reason in terms of matching/unification (or otherwise), the rule should be in clarsimp normal form, i.e. more:

```suggestion
  and abs_termination: "\<And>r s.  \<lbrakk>P r s; C r s\<rbrakk> \<Longrightarrow> whileLoop_terminates C B r s"
```

lib/ExtraCorres.thy
```diff
@@ -293,82 +295,98 @@ lemma corres_whileLoop:
...
+  and ex_abs: "ex_abs_underlying srel (P r) s'"
   and rrel: "rrel r r'"
-  and P': "P' s'"
+  and P': "P' r' s'"
+  and nf: "nf \<Longrightarrow> (\<And>r. no_fail (P r and C r) (B r))"
```
lsf37, line 312: (same as in the other rule)

```suggestion
  and nf: "\<And>r. nf \<Longrightarrow> no_fail (P r and C r) (B r)"
```

lib/ExtraCorres.thy
```diff
@@ -293,82 +295,98 @@ lemma corres_whileLoop:
...
   shows "whileLoop_terminates C' B' r' s'"
 proof -
-  have helper: "\<And>s. P s \<Longrightarrow> \<forall>r' s'. rrel r r' \<and> (s, s') \<in> srel \<and> P s \<and> P' s'
-                                     \<longrightarrow> whileLoop_terminates C' B' r' s'"
+  have helper: "\<And>s. P r s \<and> C r s \<Longrightarrow> \<forall>r' s'. rrel r r' \<and> (s, s') \<in> srel \<and> P r s \<and> P' r' s'
```
lsf37, line 315: Not sure from the diff alone if this is the, but with my change above this one might turn into `[| P r s; C r s |] ==> ..` here.

lib/ExtraCorres.thy
lsf37: The only reason I'd compress this one is so that it becomes a one-line `by`, but in terms of maintenance for this proof it might actually be nicer to keep the distinction explicit here.

lib/ExtraCorres.thy
```diff
@@ -293,82 +295,98 @@ lemma corres_whileLoop:
...
   and rrel: "rrel r r'"
-  and rrel2: "\<forall>r'. \<exists>r. rrel r r'"
-  and body_inv: "\<And>r. \<lbrace>P and C r\<rbrace> B r \<lbrace>\<lambda>_. P\<rbrace>"
-                "\<And>r'. \<lbrace>P' and C' r'\<rbrace> B' r' \<lbrace>\<lambda>_. P'\<rbrace>"
-  and abs_termination: "\<And>r s. P s \<Longrightarrow> whileLoop_terminates C B r s"
-  shows "corres_underlying srel False nf' rrel P P' (whileLoop C B r) (whileLoop C' B' r')"
+  and body_inv: "\<And>r. \<lbrace>P r and C r\<rbrace> B r \<lbrace>\<lambda>r. P r\<rbrace>"
+                "\<And>r'. \<lbrace>P' r' and C' r'\<rbrace> B' r' \<lbrace>\<lambda>r'. P' r'\<rbrace>"
+  and abs_termination: "\<And>r s. (P r s \<and> C r s) \<Longrightarrow> whileLoop_terminates C B r s"
+  and nf: "nf \<Longrightarrow> (\<And>r. no_fail (P r and C r) (B r))"
```
lsf37, line 344: (same change as in the other rule)

lib/Monads/No_Fail.thy
```diff
@@ -225,4 +225,12 @@ lemma no_fail_condition:
   unfolding condition_def no_fail_def
   by clarsimp
 
+lemma no_fail_ex_pre:
+  "(\<And>x. no_fail (P x) f) \<Longrightarrow> no_fail (\<lambda>s. \<exists>x. P x s) f"
```
lsf37, line 229: Interesting. This is entirely analogous to `hoare_vcg_ex_lift`, and as such probably fine (I'd rename to `no_fail_ex_list`, though). But now, looking at it I'd have expected something like

```
"no_fail (P ?x) f ⟹ no_fail (λs. ∃x. P x s) f"
```
and the form we have here for `∀` instead. Have you tried that? Is it true?

If yes, we should potentially change `hoare_vcg_ex_lift` if that is also true in the schematic form. The `?x` might become inconvenient, not clear. We can leave that rabbit hole for a different PR/investigation, but I'd be curious if it's true at least.

lsf37: Ignore me, the scope of the `⋀` ends before the implication, so this is entirely equivalent with `P ?x` and much nicer to work with. And that's why `hoare_vcg_ex_lift` is not too strong either. All good.

I'd still rename this one to `no_fail_ex_lift`, though.

lsf37 (approved):
Cool, much nicer rules! Minor style nitpicks only.

lib/Monads/No_Fail.thy
```diff
@@ -225,4 +225,12 @@ lemma no_fail_condition:
...
+  by (clarsimp simp: no_fail_def)
+
+lemma no_fail_grab_asm:
+  "(G \<Longrightarrow> no_fail P f) \<Longrightarrow> no_fail (\<lambda>s. G \<and> P s) f"
```
corlewis, line 233: It's not a big deal, but with this seeming to be the first time one of these lemmas is being used for `no_fail`, is it worth trying to settle on a preferred form of them?
We've never been very consistent but I feel like the `gen_asm` variants are the most common pattern, with the stateless predicate at the end of the precondition and marked by a `K`.

lib/ExtraCorres.thy
```diff
@@ -241,17 +241,19 @@ lemma hoare_from_abs_inv:
 lemma in_whileLoop_corres:
   assumes body_corres:
     "\<And>r r'. rrel r r' \<Longrightarrow>
-             corres_underlying srel False nf' rrel (P and C r) (P' and C' r') (B r) (B' r')"
-  and body_inv: "\<And>r. \<lbrace>P and C r\<rbrace> B r \<lbrace>\<lambda>_. P\<rbrace>"
-                "\<And>r'. \<lbrace>P' and C' r'\<rbrace> B' r' \<lbrace>\<lambda>_. P'\<rbrace>"
-  and cond: "\<And>r r' s s'. \<lbrakk>rrel r r'; (s, s') \<in> srel; P s; P' s'\<rbrakk> \<Longrightarrow> C r s = C' r' s'"
+             corres_underlying srel nf nf' rrel (P r and C r) (P' r' and C' r') (B r) (B' r')"
+  and body_inv: "\<And>r. \<lbrace>P r and C r\<rbrace> B r \<lbrace>\<lambda>r. P r\<rbrace>"
+                "\<And>r'. \<lbrace>P' r' and C' r'\<rbrace> B' r' \<lbrace>\<lambda>r'. P' r'\<rbrace>"
```
corlewis, line 246: I'm undecided myself, but could using `rv` in these postconditions to distinguish from the bound `r` improve readability? I guess a potential concern is that it could also be confused with the final return value of the whole loop.
```suggestion
  and body_inv: "\<And>r. \<lbrace>P r and C r\<rbrace> B r \<lbrace>\<lambda>rv. P rv\<rbrace>"
                "\<And>r'. \<lbrace>P' r' and C' r'\<rbrace> B' r' \<lbrace>\<lambda>rv'. P' rv'\<rbrace>"
```

corlewis (approved):

lib/ExtraCorres.thy
michaelmcinerney: The situation is quite tricky with these `whileLoop`s. There is the (blue) initial value given to the `whileLoop` (you can see an example of this in the conclusion of the statement of `corres_whileLoop_ret`). Then there is the (green) return value given by the loop body. I think it would be nicer to stick with the `r` for here, to reinforce that connection. But happy to discuss it some more

lib/Monads/No_Fail.thy
michaelmcinerney: The new rules I have for `no_fail` have the exact format and naming approach of `hoare_grab_asm` and `hoare_ex_pre`. So do let me know if I need to change those as well

corlewis: Oh, to clarify, absolutely no problems with this lemma as it is, as you say it completely follows precedent. It just reminded me about how inconsistent we have been around `grab_asm` vs `gen_asm` vs `gen_asm'` vs `gen_asm2` vs ... and I thought that this could be an opportunity to take a more considered approach.

michaelmcinerney: Yeah for sure. I'll probably be doing lots of things to lib as a result of the release queue work, so now might be a good time to change those original rules as well

michaelmcinerney:
Also, most of these lemmas use the `assumes... shows` and I don't think they're properly indented. The autoindenter will indent will leave the `assumes` indented by 2, indent the subsequent `and`s by a further 2 spaces, and then have the `shows` indented by 2. Should I go with that?

lib/ExtraCorres.thy
michaelmcinerney: The other alternative for these `valid` bits is that I could just eta contract the postcondition. I left it written out to emphasise that these guards do talk about the return value, but maybe that's clear enough

Xaphiosis:
> Also, most of these lemmas use the `assumes... shows` and I don't think they're properly indented. The autoindenter will indent will leave the `assumes` indented by 2, indent the subsequent `and`s by a further 2 spaces, and then have the `shows` indented by 2. Should I go with that?

I thought it would do this:
```
assumes a: blah
    and dsad: blah2
```
hmm.

lib/Monads/No_Fail.thy
Xaphiosis: I know grab_asm takes from the right side in ccorres, and `wp` also seems to prefer the right, but I don't remember what the corres standard is.

Xaphiosis: The other thing to remember is that `and` (`pred_conj`) is `infixl` while `∧` is `infixr`. This means when you stick to the `and K P` way of doing things, you grab from the right, but if you do `∧ P s` this won't work, meaning you have to grab from the left. Caution is advised in standardisation efforts for this one, although more consistency would be welcome.

lsf37: I think the rule is correct as is, it's completely analogous to `hoare_grab_asm` and has the `G` at the left, which works with `/\` (for `and` you'd want it at the right as Raf is pointing out). Both forms will be useful, i.e. I don't think we can eliminate one in favour of the other.

lsf37:
> Also, most of these lemmas use the `assumes... shows` and I don't think they're properly indented. The autoindenter will indent will leave the `assumes` indented by 2, indent the subsequent `and`s by a further 2 spaces, and then have the `shows` indented by 2. Should I go with that?

I usually avoid the `and` and just repeat the `assumes`, which is also more stable against moving things around. The correct indent of that is:

```
lemma name:
  assumes x: "..."
  assumes y: "..."
  shows "..."
```

lib/ExtraCorres.thy
lsf37: I would usually prefer the eta-contracted version unless we're trying to do the bound-name preservation trick where Isabelle would replace `rv` with whatever name was in the goal (which is not the case here).

PR #653, opened 2023-06-26
lib+refine: rename Corres_Method to CorresK_Method

lsf37:
This also renames most of the `corres*` methods to `corresK*` methods, including `corressimp` -> `corresKsimp`.

This doesn't make anything prettier, it's just preparation for a new `corres` method (implementing #634), and will hopefully make the PR for that corres method nicer and easier to review.

### DIFF
--- a/lib/CorresK_Method.thy
+++ b/lib/CorresK_Method.thy
@@ -4,11 +4,20 @@
  * SPDX-License-Identifier: BSD-2-Clause
  *)
 
-theory Corres_Method
+theory CorresK_Method
 imports Corres_Cases SpecValid_R
 begin
 
-chapter \<open>Corres Methods\<close>
+(* Advanced Eisbach example for automating corres proofs via a new corresK calculus that improves
+   on some of properties that are problematic for automation in the original corres calculus.
+
+   See also section 7.3 in
+
+   Daniel Matichuk: Automation for proof engineering: Machine-checked proofs at scale,
+   PhD thesis, UNSW 2018. https://trustworthy.systems/publications/papers/Matichuk%3Aphd.abstract
+*)
+
+chapter \<open>CorresK Methods\<close>
 
 section \<open>Boilerplate\<close>
 
@@ -112,7 +121,7 @@ private definition "my_false s \<equiv> False"
 private
   lemma corres_my_falseE: "my_false x \<Longrightarrow> P" by (simp add: my_false_def)
 
-method no_schematic_prems = (fails \<open>erule corres_my_falseE\<close>)
+private method no_schematic_prems = (fails \<open>erule corres_my_falseE\<close>)
 
 private lemma hoare_pre: "\<lbrace>my_false\<rbrace> f \<lbrace>Q\<rbrace>" by (simp add: valid_def my_false_def)
 private lemma hoareE_pre: "\<lbrace>my_false\<rbrace> f \<lbrace>Q\<rbrace>,\<lbrace>Q'\<rbrace>" by (simp add: validE_def valid_def my_false_def)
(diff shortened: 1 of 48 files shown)


lib/CorresK_Method.thy
```diff
@@ -4,11 +4,11 @@
  * SPDX-License-Identifier: BSD-2-Clause
  *)
 
-theory Corres_Method
+theory CorresK_Method
```
Xaphiosis, line 7: We might want to write some kind of a tombstone for why CorresK exists, when people ask why there is a Corres_Method and a CorresK_Method.

lib/CorresK_Method.thy
```diff
@@ -267,8 +267,8 @@ lemma wpc2_corres_protect:
   "wpc2_protect B Q \<Longrightarrow> corres_protect Q"
   by (simp add: wpc2_protect_def corres_protect_def)
 
-method corresc_left = (corresc_left_raw; (drule wpc2_corres_protect[where B=False]))
-method corresc_right = (corresc_right_raw; (drule wpc2_corres_protect[where B=True]))
+method corresKc_left = (corresc_left_raw; (drule wpc2_corres_protect[where B=False]))
```
Xaphiosis, line 270: no rename for `corresc_left_raw` etc?

lib/CorresK_Method.thy
```diff
@@ -475,7 +475,7 @@ lemmas corres_rv_lifts =
 lemmas corres_rv_prove_simple =
   corres_rv_proveT[# \<open>thin_tac _, thin_tac _\<close>, simplified]
 
-method corres_rv =
+method corresK_rv =
   (((repeat_new \<open>rule corres_rv_trivials corres_rv_lifts\<close>)?);
     ((rule corres_rv_trivials corres_rv_defers corres_rv_noops |
      (succeeds \<open>rule corres_rv_defer_left corres_rvE_R_defer_left\<close>,
```
Xaphiosis, line 481: number of `corres_` that aren't renamed here, thoughts?

lib/CorresK_Method.thy
```diff
@@ -600,39 +600,39 @@ private method corres_apply =
...
+method corresK_once declares corresK_splits corres corresK corresc_simp =
   (no_schematic_concl,
-   (corres_alternate |
+   (corresK_alternate |
      (corres_fold_dc?,
      (corres_pre',
       #break "corres",
       ( (check_corresK, determ \<open>rule corresK\<close>)
       | corres_apply
       | corres_concrete_r
```
Xaphiosis, line 615: same here w.r.t. things not renamed

lib/CorresK_Test.thy
```diff
@@ -125,10 +125,10 @@ thm corres
 
 subsection \<open>The corresc method\<close>
```
Xaphiosis, line 126: rename?

lib/CorresK_Method.thy
lsf37: It's currently still used, so maybe a bit early for tombstone, but you're right that some kind of comment is in order.

lib/CorresK_Method.thy
lsf37: For this specific one (`corres_rv`), I have now left the lemmas and lemma sets the same, because the `corres_rv` definition and lemmas would also still make sense for plain `corres`, but I'm renaming the method, because we'd only expect that to work in the `corresK` setup.

lsf37:
Alright, should now be better, and hopefully also work properly. Ready for review again.

Xaphiosis (approved):
I think you got them all. I'm not 100% whether the `corres` in the named theorems will not cause breakage of `corresK` once the new `corres` tactic develops.

lsf37:
> I think you got them all. I'm not 100% whether the `corres` in the named theorems will not cause breakage of `corresK` once the new `corres` tactic develops.

It can in theory do that, the new `corres` method in #654 will be introducing a `corresK_unsafe` set which are all the rules that are in the shared set but should not be used for `corresK`.

Xaphiosis:
> Alright, should now be better, and hopefully also work properly. Ready for review again.

Looked over it, diff looks fine.

PR #654, opened 2023-06-28
a simpler corres method

lsf37:
This new `corres` method is similar to the `corresK` method and calculus, but much less ambitious. Its main purpose is to automate boilerplate proof steps in corres proofs and is specifically not trying to fully automate such proofs (although some few might be solved).

The idea is that the method will make some progress with obvious steps and leave over a proof state the user can operate on further.

I've tried to put in a lot of comments to document the design and some of the rationale, but there are no examples yet. 

Implements #634  (see there for parts of the design discussion)

This is on top of #653, which renames the old `Corres_Method` to `CorresK_Method`

### DIFF
--- a/lib/CorresK_Method.thy
+++ b/lib/CorresK_Method.thy
@@ -567,8 +567,6 @@ text \<open>Handles structured decomposition of corres goals\<close>
 named_theorems
   corresK_splits and (* rules that, one applied, must
                         eventually yield a successful corres or corresK rule application*)
-  corres_simp_del and (* bad simp rules that break everything *)
-  corres and (* solving terminal corres subgoals *)
   corresK (* calculational rules that are phrased as corresK rules *)
 
 context begin
@@ -611,9 +609,23 @@ private method corresK_apply =
 
 private method corresK_alternate = corresK_inst | corresK_rv
 
-
+(* Corres_Method and CorresK_Method share the [corres] set. Corres_Method is more resilient against
+   unsafe terminal rules that set, so we list those [corres] rules here that might be problematic
+   for corresK. Users shouldn't need to interact with this set, but if you have declared something
+   [corres] and want it to be used by the corres method only (not corresK), then additionally
+   declare it [corres_unsafeK]. *)
+named_theorems corres_unsafeK
+lemmas [corres_unsafeK] =
+  whenE_throwError_corres
+  corres_if2
+  corres_when
+  corres_whenE
+  corres_split_handle
+  corres_split_catch
+  corres_mapM_x
 
 method corresK_once declares corresK_splits corres corresK corresKc_simp =
+  use corres_unsafeK[corres del] in \<open>use in \<open>
   (no_schematic_concl,
    (corresK_alternate |
      (corresK_fold_dc?,
@@ -624,7 +636,7 @@ method corresK_once declares corresK_splits corres corresK corresKc_simp =
       | corresK_concrete_r
       | corresKc
       | (rule corresK_splits, corresK_once)
-      )))))
+      )))))\<close>\<close>
 
 
 method corresK declares corresK_splits corres corresK corresKc_simp =
@@ -1010,10 +1022,8 @@ method corresKsimp uses simp cong search wp
     (corresK corresKc_simp: simp
     | corresKwp wp: wp
     | (rule calculus_True_insts, solves \<open>clarsimp cong: cong simp: simp corres_protect_def\<close>)
-    | clarsimp cong: cong simp: simp simp del: corres_simp_del split del: if_split
+    | clarsimp cong: cong simp: simp simp del: corres_no_simp split del: if_split
     | (match search in _ \<Rightarrow> \<open>corresK_search search: search\<close>)))+)[1]
-
-declare corres_return[corres_simp_del]
 
 section \<open>Normalize corres rule into corresK rule\<close>
 
--- a/lib/Corres_Method.thy
+++ b/lib/Corres_Method.thy
@@ -1,0 +1,215 @@
+(*
+ * Copyright 2023, Proofcraft Pty Ltd
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ *)
+
+theory Corres_Method
+imports Corres_Cases ExtraCorres
+begin
+
+(* A proof method for automating simple steps in corres proofs.
+
+   While the method might solve some corres proofs completely, the purpose is to make simple
+   things more automatic, remove boilerplate, and to leave a proof state in which the user can make
+   more progress. The goal is not to provide full automation or deeper search.
+
+   The main idea is to repeatedly try to apply terminal [corres] rules after splitting off the head
+   of a bind/bindE statement on both sides of a corres goal. The method provides options for less
+   safe rules such as moving asserts to guards etc when the user knows that this is safe to do in
+   a particular instance.
+
+   See description at corres' method below for all parameters and options.
+*)
+
+section \<open>Goal predicates\<close>
+
+(* Succeed if the conclusion is a corres goal and also not purely schematic *)
+method is_corres = succeeds \<open>rule corres_inst\<close>, fails \<open>rule TrueI\<close>
+
+lemma no_fail_triv: "no_fail P f \<Longrightarrow> no_fail P f" .
+lemmas hoare_trivs = hoare_triv hoare_trivE hoare_trivE_R hoare_trivR_R no_fail_triv
+
+(* Succeed if the conclusion is a wp/no_fail goal and also not purely schematic*)
+method is_wp = succeeds \<open>rule hoare_trivs\<close>, fails \<open>rule TrueI\<close>
+
+lemmas hoare_post_False = hoare_pre_cont[where P="\<lambda>_. \<bottom>"]
+lemmas hoareE_post_False = hoare_FalseE[where Q="\<lambda>_. \<bottom>" and E="\<lambda>_. \<bottom>"]
+
+(* Succeed if the conclusion has a schematic post condition (assuming a wp goal). *)
+method is_hoare_schematic_post =
+  (* If the post condition matches both \<top> and \<bottom>, it must be schematic *)
+  succeeds \<open>wp_pre, rule hoare_post_False hoareE_post_False\<close>,
+  succeeds \<open>wp_pre, rule wp_post_taut wp_post_tautE\<close>
+
+section \<open>Main corres method\<close>
+
+named_theorems corres_splits
+method corres_split declares corres_splits = no_name_eta, rule corres_splits
+
+(* This method is called on non-corres, non-wp side conditions after a corres rule has been
+   applied. At that point, there should be no schematic variables in those side condition goals.
+   Despite that, we are still careful with simp etc here, in case the user does provide a corres
+   rule that generates a schematic in those side condition goals. *)
+method corres_cleanup methods m uses simp simp_del split split_del cong intro =
+    m
+  | assumption
+  | rule refl TrueI
+  | clarsimp simp del: corres_no_simp simp_del simp: simp split: split split del: split_del
+             cong: cong intro!: intro
+  (* enables passing in conjI for terminal goals: *)
+  | (rule intro;
+     corres_cleanup m simp: simp simp_del: simp_del split: split split_del: split_del
+                      cong: cong intro: intro)
+
+(* Apply a single corres rule and attempt to solve non-corres and non-wp side conditions.
+   We don't expect any wp side conditions, but check anyway for safety. If the rule is declared
+   as terminal rule, all side conditions must be solved and no corres or wp side conditions are
+   allowed. If the rule is declared as a regular corres rule, unsolved side conditions as well as
+   corres and wp side conditions will be left over unchanged. *)
+method corres_rule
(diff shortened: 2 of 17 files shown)


lsf37:
I might try to replace some of the existing `corresKsimp` invocations and maybe add a file with examples. Depending on how fast that goes, this would either be a separate PR or another commit or two on top of this one.

lsf37:
Have now deployed the method in a few places in RISCV64 Refine and fine tuned thing a bit from that experience.

Seems to be working relatively well, at least I was able to develop with it incrementally fairly nicely and then collapse things down. The real test will be if it works for others as well.

lib/ExtraCorres.thy
```diff
@@ -8,6 +8,15 @@ theory ExtraCorres
 imports Corres_UL
 begin
 
+(* FIXME: the S in this rule is mainly to make the induction work, we don't actually need it in
```
Xaphiosis, line 11: Not sure why this and the comments below are FIXMEs rather than a normal comment? The contents of the comment is fine.

proof/refine/RISCV64/ArchAcc_R.thy
```diff
@@ -498,14 +498,30 @@ lemma getPTE_wp:
...
-  "corres_inst_eq ptr ptr' \<Longrightarrow> \<forall>s s'. (s, s') \<in> state_relation \<longrightarrow> True \<longrightarrow>
-   (pspace_aligned s \<and> pspace_distinct s \<and> pt_at ptr s \<and> ptr = ptr') \<longrightarrow>
-    \<top> s' \<longrightarrow> page_table_at' ptr' s'"
-  by ( fastforce intro!: page_table_at_cross)
+(* only applies when ptr is available in abstract and concrete guard at the same time *)
+lemma pt_at_lift_eq:
+  "\<forall>s s'. (s, s') \<in> state_relation \<longrightarrow>
+          (pspace_aligned s \<and> pspace_distinct s \<and> pt_at ptr s) \<longrightarrow>
+          \<top> s' \<longrightarrow>
+          page_table_at' ptr s'"
```
Xaphiosis, line 506: off-topic, but why is the format for these rules so strange? Especially for `pt_at_lift_relation` the combination of meta and object implication doesn't look obvious to me

proof/refine/RISCV64/VSpace_R.thy
```diff
@@ -194,58 +173,28 @@ lemma hwASIDFlush_corres[corres]:
...
-                      P'="\<top>" and
-                      Q="invs and
-                         (\<lambda>s. asid_table s = asidTable \<circ> ucast)" in
-                      corres_split)
-         apply (simp add: dom_def)
-         apply (rule get_asid_pool_corres_inv'[OF refl, unfolded pred_conj_def, simplified])
-        apply (rule corres_when)
+  apply corres_pre
+    apply (corres_split, corres)
+      apply corres_cases_both
```
Xaphiosis, line 184: I'd enjoy a comment here saying which cases these are (so can tell where corres got to without looking at the proof state.
Otherwise, looks like the `corres` method and its cousins are very useful indeed.

lib/Corres_UL.thy
```diff
@@ -903,7 +903,13 @@ lemma corres_returnOk:
   apply wp
   done
 
-lemmas corres_returnOkTT = corres_trivial [OF corres_returnOk]
+lemma corres_returnOkTT:
+  "r x y \<Longrightarrow> corres_underlying sr nf nf' (r' \<oplus> r) \<top> \<top> (returnOk x) (returnOk y)"
```
Xaphiosis, line 907: did something change here, or are you writing it out?

lib/Corres_UL.thy
```diff
@@ -484,30 +472,23 @@ lemma corres_if3:
 
 text \<open>Some equivalences about liftM and other useful simps\<close>
 
-lemma snd_liftM [simp]:
+lemma snd_liftM[simp]: (* FIXME: move *)
```
Xaphiosis, line 475: if you're doing a clean up commit ...

lib/CorresK_Method.thy
```diff
@@ -611,9 +609,18 @@ private method corresK_apply =
 
 private method corresK_alternate = corresK_inst | corresK_rv
 
-
+(* declared [corres] in Corres_Method and safe there, but not here, because they are not strictly
+   terminal rules *)
+named_theorems corres_unsafeK
+lemmas [corres_unsafeK] =
```
Xaphiosis, line 615: I thought these changes (the ones to `CorresK_Method`) made sense when I first looked at this diff, but now I'm completely confused and the comment above seems cryptic.

lib/Corres_Method.thy
```diff
@@ -0,0 +1,212 @@
...
+
+lemma corres_get_trivial[corres_term]:
+  "corres_underlying sr nf nf' (\<lambda>s s'. (s,s') \<in> sr) \<top> \<top> get get"
+  by simp
+
+lemmas corres_underlying_stateAssert_stateAssert_trivial[corres_term] =
+  corres_underlying_stateAssert_stateAssert[where P=\<top> and P'=\<top>, simplified]
+
+lemmas corres_modify_tivial[corres_term] =
+  corres_modify[where P=\<top> and P'=\<top>, simplified]
```
Xaphiosis, line 188: can you check on the var names this one and `corres_underlying_stateAssert_stateAssert_trivial` spit out? I'm having flashbacks to my monadic rewrite cleanup and some of the horrors `simplified` produced there.

lib/Corres_Method.thy
```diff
@@ -0,0 +1,212 @@
...
+    | succeeds \<open>is_wp\<close>, fails \<open>is_hoare_schematic_post\<close>,
+      (wpsimp wp: wp wp_del: wp_del simp: simp simp_del: simp_del split: split split_del: split_del
+              cong: cong)+
+   )+)[1]
+
+(* Instance of the corres' method with default cleanup tactic *)
+method corres
+  uses simp term_simp simp_del split split_del cong intro wp wp_del fallback
+  declares corres corres_term corres_splits =
+  corres' \<open>fail\<close> simp: simp term_simp: term_simp simp_del: simp_del split: split
```
Xaphiosis, line 154: does this mean the default cleanup tactic is `fail`?

lib/Corres_Method.thy
```diff
@@ -0,0 +1,212 @@
...
+method is_corres = succeeds \<open>rule corres_inst\<close>, fails \<open>rule TrueI\<close>
+
+lemma no_fail_triv: "no_fail P f \<Longrightarrow> no_fail P f" .
+lemmas hoare_trivs = hoare_triv hoare_trivE hoare_trivE_R hoare_trivR_R no_fail_triv
+
+(* Succeed if the conclusion is a wp/no_fail goal and also not purely schematic*)
+method is_wp = succeeds \<open>rule hoare_trivs\<close>, fails \<open>rule TrueI\<close>
+
+lemmas hoare_post_False = hoare_pre_cont[where P="\<lambda>_. \<bottom>"]
+lemmas hoareE_post_False = hoare_FalseE[where Q="\<lambda>_. \<bottom>" and E="\<lambda>_. \<bottom>"]
```
Xaphiosis, line 37: that naming consistency...

lib/Corres_Method.thy
```diff
@@ -0,0 +1,212 @@
...
+
+(* Succeed if the conclusion has a schematic post condition (assuming a wp goal). *)
+method is_hoare_schematic_post =
+  (* If the post condition matches both \<top> and \<bottom>, it must be schematic *)
```
Xaphiosis, line 41: oh wow, that is an impressive idea

lib/Corres_Method.thy
```diff
@@ -0,0 +1,212 @@
...
+  succeeds \<open>wp_pre, rule wp_post_taut wp_post_tautE\<close>
+
+section \<open>Main corres method\<close>
+
+named_theorems corres_splits
+method corres_split declares corres_splits = no_name_eta, rule corres_splits
+
+(* This method is called on non-corres, non-wp side conditions after a corres rule has been
+   applied. At that point, there should be no schematic variables in those side condition goals.
+   Despite that,  we are still careful with simp etc here, in case the user does provide a corres
```
Xaphiosis, line 52: double space

lib/Corres_Method.thy
```diff
@@ -0,0 +1,212 @@
...
+   After preliminaries such as wpfix and corres_pre, repeatedly try to either solve the goal
+   outright or split off the head from a bind/bindE statement and apply a corres rule (only
+   split when a corres rule applies). If none of these works, try a corres rule from the "fallback"
+   argument. (These are for things like moving asserts to a guard, which we only want to do if no
+   other corres rule applies).
+
+   Attempt to solve side conditions with the corres_cleanup method. The cleanup method uses the
+   simp and term_simp arguments.
+
+   Attempt clarsimp on the head corres goal without rewriting guards and return relation when
```
Xaphiosis, line 94: without rewriting guards *or/nor* return relation

lib/Corres_Method.thy
```diff
@@ -0,0 +1,212 @@
...
+       but it also very likely that the first thing we want to do for such a goal is clarsimp. That
+       means, overall we might solve a few more goals, and not be detrimental to interactive proof
+       either. *)
+    | fails \<open>is_corres\<close>, fails \<open>is_wp\<close>,
+      clarsimp cong: cong simp del: simp_del simp: simp split del: if_split split_del split: split
+               intro!: intro
+    (* if (and only if) we get to the state where all corres goals and side conditions are solved,
+       attempt to solve all wp goals that were generated in order. We are not using then_all_new_fwd
+       here, because we should only start solving wp goals once *all* corres goals are solved,
+       because otherwise the goal will still have schematic post conditions. Fail if there is a
```
Xaphiosis, line 142: `, because ..., because otherwise` -> `because/since ..., otherwise`

lib/Corres_Method.thy
Xaphiosis: Ah wait, I think I get it, that's passed all the way down ... yeah, would still be useful to hint at why you want `fail` here in a/the comment.

Xaphiosis (approved):
I'm impressed you did this only with Eisbach. I'd have gone down rabbit holes of examining terms for schematics, which would result in faster code, but definitely less comprehensible. The documentation is solid as well.

There was one thing I didn't comment on directly because I'm used to it, but not sure if it's confusing to others or not: inline `etc` in the docs, e.g. "rules such as moving asserts to guards etc when the user" can be a tad confusing.

lib/Corres_UL.thy
lsf37: It changed slightly, the `lemmas` version had some state relation junk left over in it that didn't do anything. It didn't really hurt either, but it was a bit messy.

proof/refine/RISCV64/ArchAcc_R.thy
lsf37: These are left-overs from Dan's very early attempts at cross-over rules. They are used only exactly here and nowhere else. In the long term we should actually remove them, I just wanted to make sure the new method could deal with the same kinds of goals that come out of these (and there were actually a few subtleties there, because when you cross over things, the guards might not depend on the correct green variables and so one).

lib/ExtraCorres.thy
lsf37: Should probably split it up into two comments. One part is FIXME and should be fixed, the other part is just warning for future me not to prove the strong version again to realise again that it doesn't help and is not actually stronger.

lib/CorresK_Method.thy
lsf37: the two methods share the `[corres]` set, and my method is less susceptible to unsafe rules being in the `[corres]` set, because it has all the safeguards. And because that works, I'm promptly abusing it, which then breaks `corresK`. This `corres_unsafeK` set are those unsafe rules that `corresK` should remove before it does stuff.

This idea is that normal users will follow the instructions and not put new unsafe rules in, so it should never be an issue later on, just for initial setup. (If users do put in unsafe rules, the can still add them to this set to preserve `corresK`).

lib/Corres_Method.thy
lsf37: ah, yes, very good point. Probably better to write them out.

lib/Corres_Method.thy
lsf37: good point. `succeed` would always stop the cleanup and do nothing, `fail` lets all other tactics work.

lib/Corres_Method.thy
lsf37: yeah, we have this old old issue "wp lemmas could do with a mass rename"..

proof/refine/RISCV64/ArchAcc_R.thy
Xaphiosis: if so, then these ones could use a FIXME

michaelmcinerney (approved):
Looks very cool. I'm sure I don't understand many of the subtleties of the method setup, but `corressimp` is used a fair bit in MCS and if I am the one who handles the merge, then I'm sure it'll be a good opportunity to test this out a bit.

lib/CorresK_Method.thy
lsf37: Added more of that to the comment in the file.

lsf37:
Review feedback addressed and commits squashed.

PR #655, opened 2023-06-30
crefine: add no_name_eta to tactics

corlewis:
Add `no_name_eta` as a final step in various tactics used by `crefine`, in particular to eta-contract expanded terms. This has partially been superseded by https://github.com/seL4/l4v/pull/646, but still provides improved consistency and better bound variable names.

As potential negatives:
* there will probably be a performance impact. This hasn't been noticeable in interactive sessions but it would be good if we could measure it.
* this isn't perfect consistency, `rule` applications also eta-expand terms when there is unification with higher-order variables.
* a couple of simplification steps in existing proofs made implicit use of terms being eta-expanded. I don't think that this will be a problem for new proofs, it can be fairly easily worked around when encountered and over time could lead to further improvements if we more consistently use eta-contracted forms.

### DIFF
--- a/lib/clib/CCorres_Rewrite.thy
+++ b/lib/clib/CCorres_Rewrite.thy
@@ -17,7 +17,7 @@ lemma ccorres_com_eq_hom:
            elim: ccorres_semantic_equivD2)
 
 method ccorres_rewrite declares C_simp C_simp_pre C_simp_simps C_simp_throws
-  = simpl_rewrite hom: ccorres_com_eq_hom
+  = (simpl_rewrite hom: ccorres_com_eq_hom, no_name_eta)
 
 lemma hoarep_com_eq_hom:
   "com_eq_hom \<Gamma> (\<lambda>c. hoarep \<Gamma> {} F P c Q A)"
--- a/proof/crefine/ARM/Finalise_C.thy
+++ b/proof/crefine/ARM/Finalise_C.thy
@@ -277,10 +277,10 @@ lemma cancelAllIPC_ccorres:
   apply (cinit lift: epptr_')
    apply (rule ccorres_symb_exec_l [OF _ getEndpoint_inv _ empty_fail_getEndpoint])
     apply (rule_tac xf'=ret__unsigned_'
-                and val="case rv of IdleEP \<Rightarrow> scast EPState_Idle
+                and val="case ep of IdleEP \<Rightarrow> scast EPState_Idle
                             | RecvEP _ \<Rightarrow> scast EPState_Recv | SendEP _ \<Rightarrow> scast EPState_Send"
-                and R="ko_at' rv epptr"
-                 in ccorres_symb_exec_r_known_rv_UNIV[where R'=UNIV])
+                and R="ko_at' ep epptr"
+             in ccorres_symb_exec_r_known_rv_UNIV[where R'=UNIV])
        apply vcg
        apply clarsimp
        apply (erule cmap_relationE1 [OF cmap_relation_ep])
@@ -289,8 +289,8 @@ lemma cancelAllIPC_ccorres:
        apply (simp add: cendpoint_relation_def Let_def
                  split: endpoint.split_asm)
       apply ceqv
-     apply (rule_tac A="invs' and ko_at' rv epptr"
-                  in ccorres_guard_imp2[where A'=UNIV])
+     apply (rule_tac A="invs' and ko_at' ep epptr"
+              in ccorres_guard_imp2[where A'=UNIV])
       apply wpc
         apply (rename_tac list)
         apply (simp add: endpoint_state_defs
@@ -404,10 +404,10 @@ lemma cancelAllSignals_ccorres:
   apply (cinit lift: ntfnPtr_')
    apply (rule ccorres_symb_exec_l [OF _ get_ntfn_inv' _ empty_fail_getNotification])
     apply (rule_tac xf'=ret__unsigned_'
-                and val="case ntfnObj rv of IdleNtfn \<Rightarrow> scast NtfnState_Idle
+                and val="case ntfnObj ntfn of IdleNtfn \<Rightarrow> scast NtfnState_Idle
                             | ActiveNtfn _ \<Rightarrow> scast NtfnState_Active | WaitingNtfn _ \<Rightarrow> scast NtfnState_Waiting"
-                and R="ko_at' rv ntfnptr"
-                 in ccorres_symb_exec_r_known_rv_UNIV[where R'=UNIV])
+                and R="ko_at' ntfn ntfnptr"
+             in ccorres_symb_exec_r_known_rv_UNIV[where R'=UNIV])
        apply vcg
        apply clarsimp
        apply (erule cmap_relationE1 [OF cmap_relation_ntfn])
@@ -416,8 +416,8 @@ lemma cancelAllSignals_ccorres:
        apply (simp add: cnotification_relation_def Let_def
                  split: ntfn.split_asm)
       apply ceqv
-     apply (rule_tac A="invs' and ko_at' rv ntfnptr"
-                  in ccorres_guard_imp2[where A'=UNIV])
+     apply (rule_tac A="invs' and ko_at' ntfn ntfnptr"
+              in ccorres_guard_imp2[where A'=UNIV])
       apply wpc
         apply (simp add: notification_state_defs ccorres_cond_iffs)
         apply (rule ccorres_return_Skip)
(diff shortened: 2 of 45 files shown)


proof/crefine/ARM/Tcb_C.thy
```diff
@@ -2020,7 +2019,7 @@ shows
                         apply (clarsimp simp: min_def iffD2 [OF mask_eq_iff_w2p] word_size
                                               word_less_nat_alt
                                       split: if_split_asm dest!: word_unat.Rep_inverse')
-                       apply simp
+                       apply (simp add: pred_conj_def)
```
corlewis, line 2022: This is a case where the previous proof could automatically rewrite `pred_conj` due to the postcondition being eta-expanded.

proof/crefine/ARM/Ipc_C.thy
```diff
@@ -2951,9 +2951,10 @@ proof -
                    del: Collect_const)
        apply csymbr
        apply (rename_tac "lngth")
-       apply (simp add: mi_from_H_def mapME_def del: Collect_const cong: bind_apply_cong)
+       apply (unfold mapME_def)[1]
+       apply (simp add: mi_from_H_def del: Collect_const)
```
corlewis, line 2955: This was a bit annoying, in eta-contracted form the simplifier does not rewrite `mapME` inside of the abstract program. It's very possible that there is a nicer way of doing this but for a first draft I just went with `unfold`.

proof/crefine/ARM/VSpace_C.thy
```diff
@@ -2233,9 +2234,7 @@ lemma unmapPage_ccorres:
           \<comment> \<open>ARMSection\<close>
           apply (rule ccorres_Cond_rhs)
            apply (rule ccorres_rhs_assoc)+
-           \<comment> \<open>FIXME: The second csymbr rewrites the return relation of the goal.
-                      If this was avoided then folding the dc_def could be removed\<close>
-           apply (csymbr, csymbr, fold dc_def)
+           apply (csymbr, csymbr)
```
corlewis, line 2237: This was an unexpected but nice benefit of this change.

proof/crefine/ARM/Finalise_C.thy
```diff
@@ -778,13 +778,13 @@ lemma unbindMaybeNotification_ccorres:
...
+    apply (rule_tac P="ntfnBoundTCB ntfn \<noteq> None \<longrightarrow>
+                         option_to_ctcb_ptr (ntfnBoundTCB ntfn) \<noteq> NULL"
+             in ccorres_gen_asm)
     apply (rule_tac xf'=boundTCB_'
-                           and val="option_to_ctcb_ptr (ntfnBoundTCB rv)"
-                           and R="ko_at' rv ntfnptr and valid_bound_tcb' (ntfnBoundTCB rv)"
-                            in ccorres_symb_exec_r_known_rv_UNIV[where R'=UNIV])
+                and val="option_to_ctcb_ptr (ntfnBoundTCB ntfn)"
+                and R="ko_at' ntfn ntfnptr and valid_bound_tcb' (ntfnBoundTCB ntfn)"
+             in ccorres_symb_exec_r_known_rv_UNIV[where R'=UNIV])
```
Xaphiosis, line 787: does anything in this file have to do with no_name_eta? it seems like the result of a rename_tac

Xaphiosis: oh! the no_name_eta means more names are preserved, right?

proof/crefine/ARM/Ipc_C.thy
Xaphiosis: simp with a cong rule?

lib/clib/CCorres_Rewrite.thy
```diff
@@ -17,7 +17,7 @@ lemma ccorres_com_eq_hom:
            elim: ccorres_semantic_equivD2)
 
 method ccorres_rewrite declares C_simp C_simp_pre C_simp_simps C_simp_throws
-  = simpl_rewrite hom: ccorres_com_eq_hom
+  = (simpl_rewrite hom: ccorres_com_eq_hom, no_name_eta)
```
Xaphiosis, line 20: Interesting, do you think ccorres_rewrite can get into a situation that requires no_name_eta?

Xaphiosis:
In general, I approve. It would be useful in this case to run a test of ARM CRefine to see if there's any time penalty that is noticeable. If you run it on the machine at work without any `-j`, it shouldn't be doing anything else and give us an estimate for whether we should care or not.

proof/crefine/ARM/Finalise_C.thy
corlewis: Yep, exactly!
For completeness, there are also a few places where a name being preserved meant that a different `rva` was changed to `rv` and so I also added some `rename_tacs` to improve those.

proof/crefine/ARM/Ipc_C.thy
corlewis: I tried a couple, including a new one that I thought might help, but couldn't get anything to apply. I might try again now that I've finished a first pass through the proofs.

lib/clib/CCorres_Rewrite.thy
corlewis: I unfortunately don't have the example anymore, but that was actually what first sparked all of this work. A `ccorres_rewrite` left the goal eta-expanded and a following step then rewrote `dc` when I didn't want it to.

Xaphiosis: Unexpected! In this case even more of a thumbs up.

proof/crefine/ARM/Finalise_C.thy
Xaphiosis: That cheers me up. I'm so tired of `rv` and `x` appearing by default.

corlewis:
> In general, I approve. It would be useful in this case to run a test of ARM CRefine to see if there's any time penalty that is noticeable. If you run it on the machine at work without any `-j`, it shouldn't be doing anything else and give us an estimate for whether we should care or not.

Hmm, I get
```
Finished CRefine                   passed      ( 0:38:32 real,  3:37:02 cpu, 23.14GB)
```
vs
```
Finished CRefine                   passed      ( 0:42:36 real,  3:59:06 cpu, 22.00GB)
```
so it's a noticeable difference.

Xaphiosis:
That is a noticeable difference, but not terminal. I'd be willing to take it for less insanity in the future. Wonder what @lsf37 will say.

lsf37:
Hm, that's 20min compute time. Not huge in terms of real time, but I wonder if there is something else going on. Is the timing stable if you run it multiple times? I know, it takes forever to check, but if it's just noisy and the real difference is smaller, I'd feel a lot better about pulling this in.

corlewis:
Looks like there's a lot of noise, this time it's reversed
```
Finished CRefine                   passed      ( 0:40:15 real,  3:42:22 cpu, 21.60GB)
```
vs
```
Finished CRefine                   passed      ( 0:37:21 real,  3:25:27 cpu, 22.87GB)
```

corlewis:
I can run them some more times over the weekend and try to average it out, but I don't think that we're getting much useful information
```
Finished CRefine                   passed      ( 0:45:59 real,  4:34:58 cpu, 22.60GB)
```
vs
```
Finished CRefine                   passed      ( 0:45:41 real,  4:28:58 cpu, 23.22GB)
```

lsf37:
Ok, it does seem to be mostly noise. Happy for it to go ahead, then.

corlewis:
Alright, this is some very unscientific benchmarking but averaging out 34 runs gives me `CRefine (00:39:20 real, 03:42:35 cpu, 22.87GB)` without these changes and `CRefine (00:41:54 real, 03:52:56 cpu, 22.82GB)` with them.

lsf37:
Cool, that is close enough. Do you have a standard deviation?

corlewis:
`Standard Deviations (00:03:04 real,  00:22:12 cpu, 0.78GB)` without changes and `Standard Deviations (00:03:21 real,  00:24:35 cpu, 0.73GB)` with

lsf37:
Perfect, thank you for computing those.

corlewis:
Very little difference with these size numbers, but I realised that I had some floating point errors. It should have been `Standard Deviation (00:02:59 real,  00:22:08 cpu, 0.58GB)` without changes and `Standard Deviation (00:03:15 real,  00:24:26 cpu, 0.50GB)` with.

corlewis:
All architectures are now updated. For reviewing purposes, feel free to skip commits marked as 'copied from previous arches'.

proof/crefine/ARM/Ipc_C.thy
lsf37: I'm fine with the unfold here. Compared to the benefit of much fewer `rv` and `xa` etc, that is a small price to pay.

proof/crefine/ARM/VSpace_C.thy
lsf37: cool

lsf37 (approved):
Nice, I'm quite happy with this. Much easier to see what's going on when not everything is just `rv`.

PR #656, opened 2023-06-30
test and docs for `corres` method + slight tweaks for the same

lsf37:
- enforce (no_asm) more strongly for corres term rewriting in `corres`
- add an additional breakpoint for debugging `corres_cleanup`
- add a tutorial for the `corres` method, which also serves as a longer test case

### DIFF
--- a/lib/Corres_Method.thy
+++ b/lib/Corres_Method.thy
@@ -161,6 +161,10 @@ method corres
 
 section \<open>Corres rule setup\<close>
 
+(* Avoid using equations in the assumptions. subst_all gets around (no_asm_use) in some cases,
+   which we don't want. *)
+lemmas [corres_no_simp] = subst_all
+
 lemmas [corres_splits] =
   corres_split
   corres_splitEE
--- a/lib/Corres_Method.thy
+++ b/lib/Corres_Method.thy
@@ -42,6 +42,9 @@ method is_hoare_schematic_post =
   succeeds \<open>wp_pre, rule hoare_post_False hoareE_post_False\<close>,
   succeeds \<open>wp_pre, rule wp_post_taut wp_post_tautE\<close>
 
+(* Succeed if wpsimp or wp can safely be applied *)
+method is_safe_wp = is_wp, fails \<open>is_hoare_schematic_post\<close>
+
 section \<open>Main corres method\<close>
 
 named_theorems corres_splits
@@ -125,7 +128,7 @@ method corres'
     (* simplify head corres goal, e.g. for things like liftM unfolding if the user provides such
        a rule as "simp". Not clarsimp, because clarsimp will still perform hypsubst on assumptions
        and might through that rewrite guards *)
-    | succeeds \<open>is_corres\<close>,
+    | is_corres,
       simp (no_asm_use) cong: corres_weaker_cong cong split: split split del: if_split split_del
                         add: simp corres_simp del: corres_no_simp simp_del
     (* simplify any remaining side condition head goal (non-corres, non-wp). This is either a side
@@ -143,7 +146,7 @@ method corres'
        otherwise the goal will still have schematic post conditions. Fail if there is a
        free schematic postcondition despite all these measures.
        *)
-    | succeeds \<open>is_wp\<close>, fails \<open>is_hoare_schematic_post\<close>,
+    | is_safe_wp,
       (wpsimp wp: wp wp_del: wp_del simp: simp simp_del: simp_del split: split split_del: split_del
               cong: cong)+
    )+)[1]
--- a/lib/Corres_Method.thy
+++ b/lib/Corres_Method.thy
@@ -52,7 +52,8 @@ method corres_split declares corres_splits = no_name_eta, rule corres_splits
    Despite that, we are still careful with simp etc here, in case the user does provide a corres
    rule that generates a schematic in those side condition goals. *)
 method corres_cleanup methods m uses simp simp_del split split_del cong intro =
-    m
+  #break "corres_cleanup",
+  ( m
   | assumption
   | rule refl TrueI
   | clarsimp simp del: corres_no_simp simp_del simp: simp split: split split del: split_del
@@ -60,7 +61,7 @@ method corres_cleanup methods m uses simp simp_del split split_del cong intro =
   (* enables passing in conjI for terminal goals: *)
   | (rule intro;
      corres_cleanup m simp: simp simp_del: simp_del split: split split_del: split_del
-                      cong: cong intro: intro)
+                      cong: cong intro: intro))
 
 (* Apply a single corres rule and attempt to solve non-corres and non-wp side conditions.
    We don't expect any wp side conditions, but check anyway for safety. If the rule is declared
--- a/lib/test/Corres_Test.thy
+++ b/lib/test/Corres_Test.thy
@@ -1,0 +1,312 @@
+(*
+ * Copyright 2023, Proofcraft Pty Ltd
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ *)
+
+theory Corres_Test
+imports Lib.Corres_Method
+begin
+
+(* Test cases and tutorial/docs for Corres_Method *)
+
+
+section "Setup"
+
+(* Setting up some monads and lemmas to play with later *)
+experiment
+  fixes sr nf nf'
+
+  fixes f f' :: "('s, nat) nondet_monad"
+  assumes f: "corres_underlying sr nf nf' (=) \<top> \<top> f f'"
+
+  fixes Q g g' t
+  assumes g: "\<And>x x'::nat. x = t x' \<Longrightarrow> corres_underlying sr nf nf' (=) Q \<top> (g x) (g' x')"
+  assumes t: "\<And>x. t x = x"
+
+  fixes P
+  assumes Q: "\<lbrace>P\<rbrace> f \<lbrace>\<lambda>_. Q\<rbrace>"
+
+  fixes h h'
+  assumes h: "corres_underlying sr nf nf' (=) \<top> \<top> h h'"
+begin
+
+abbreviation "corres \<equiv> corres_underlying sr nf nf'"
+
+
+section "Examples"
+
+(* The purpose of the corres method is to make progres on easy corres steps, where things
+   "obviously" match up on the concrete and abstract side. You can provide basic terminal
+    corres rules like f and g to try. You can provide simp rules to rewrite corres goals
+    and to solve side conditions of terminal rules such as the rule for g above. Finally,
+    you can provide wp rules to solve or make progress on the final wp goals that a corres
+    proof produces. *)
+lemma "corres (=) P \<top> (do x \<leftarrow> f; g x od) (do y \<leftarrow> f'; g' y od)"
+  by (corres corres: f g wp: Q simp: t)
+
+(* All of these can be declared globally and will be picked up by the method *)
+context
+  notes [corres] = f g
+  notes [wp] = Q
+  notes [simp] = t
+begin
+
+lemma "corres (=) P \<top> (do x \<leftarrow> f; g x od) (do y \<leftarrow> f'; g' y od)"
+  by corres
+
+end
+
+(* During development, the rules needed are often not declared [corres] yet or the right
+   simp rules for side conditions etc have yet to be figured out. The following proof
+   demonstrates this process. *)
+lemma "corres (=) P \<top> (do x \<leftarrow> f; g x od) (do y \<leftarrow> f'; g' y od)"
+  (* We begin by invoking "corres" *)
+  apply corres
+    (* In this case, not much has happened yet, corres has only produced schematic preconditions.
+       However, we can see that f and f' are the heads of both sides, and searching with find_theorems
+       for a corres rule that mentions those two turns up the rule "f", which we provided to the corres
+       method. At this point we can either go back and add it to the previous line, or we
+       add a new invocation. The process is very similar to using wpsimp. *)
+    apply (corres corres: f)
+      (* We see that f has been split off, and we now have a goal for g. Same process as above finds
+         the corresponding rule. *)
+      apply (corres corres: g)
+      (* This solves the corres goal but leaves the side condition of the "g" rule. We can
+         now either solve it manually with "apply (simp add: t)" and then continue, or, if it really
+         is as simple as a few simp rules, we can tell the corres method to apply it directly *)
+      apply (corres simp: t)
+     (* We now have only wp goals and the final implication left. *)
+     apply (wp Q)
+    apply wp
+   apply simp
+  apply simp
+  done
+
+(* Once we have found this proof, we can roll it up, and merge eg. the "simp: t" into the corres
+   line before. *)
+lemma "corres (=) P \<top> (do x \<leftarrow> f; g x od) (do y \<leftarrow> f'; g' y od)"
+  apply corres
+    apply (corres corres: f)
+      apply (corres corres: g simp: t)
+     (* Adding "wp: Q" to the previous line does not help at this stage, because this wp goal
+        is produced in the (corres corres: f) line above. We could do
+        apply (corres corres: g simp: t wp: Q)+
+        above, which *would* solve the rest of the goals, but using + in an uncontrolled way
+        is not very stable and therefore not recommended style. *)
+     apply (wp Q)
+    apply wp
+   apply simp
+  apply simp
+  done
+
+(* Merging the g and f corres lines does enable us to prove the Q wp rule. *)
+lemma "corres (=) P \<top> (do x \<leftarrow> f; g x od) (do y \<leftarrow> f'; g' y od)"
+  apply corres
+    apply (corres corres: f g simp: t wp: Q)
+    (* This will still leave the final implication, because we have produced that implication
+       outside this subgoal. Merging the two corres invocations above will attempt the final
+       implications automatically as well. *)
+   apply simp
+  apply simp
+  done
+
+
+section "More controlled single-stepping"
+
+(* Sometimes invoking "corres" does too much or too little.
+   Too much can occur when the method applies a rule we didn't know is in the [corres] set and
+   which leaves us with a strange side condition to solve. Or we may have added an unsafe,
+   not-really-terminal rule to [corres] and now we are getting an unprovable goal. Too little
+   can occur when the method refuses to split off the head terms even though it looks like a
+   terminal corres rule should apply. For these cases, we can take apart some of the internal
+   steps like this: *)
+lemma "corres (=) P \<top> (do x \<leftarrow> f; g x od) (do y \<leftarrow> f'; g' y od)"
+  (* Controlled way to only introduce schematic preconditions and the final implication *)
+  apply corres_pre
+    (* Invoking "corres" would now fail. Maybe we are convinced that the "f" rule is declared
+       [corres] and we want to figure out why it does not apply. Invoking the corres_split method
+       will give us the goal the terminal corres rule is tried on: *)
+    apply corres_split
+       (* Trying out "rule f" does work now -- if it didn't we could debug that and find out why *)
+       apply (succeeds \<open>rule f\<close>)
+       (* Turns out we forgot to declare it, so we add it manually, and the corres method now
+          succeeds on the subgoal *)
+       apply (corres corres: f)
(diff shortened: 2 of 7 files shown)


lsf37:
Tagging @Xaphiosis just in case he wants to read through this (but completely fine to ignore)

michaelmcinerney (approved):
The tutorial is very helpful, thanks. I hadn't come across the `apply_debug` feature before.

lib/test/Corres_Test.thy
```diff
@@ -0,0 +1,306 @@
...
+   simp rules for side conditions etc have yet to be figured out. The following proofs
+   demonstrates this process. *)
+lemma "corres (=) P \<top> (do x \<leftarrow> f; g x od) (do y \<leftarrow> f'; g' y od)"
+  (* We begin by just invoking "corres" *)
+  apply corres
+    (* In this case, not much has happened yet, corres has only produced schematic preconditions.
+       However, we can see that f and f' are the heads of both sides, and searching with find_theorems
+       for a corres rule that mentions those two turns up the rule "f". We provided it to the corres
+       method. At this point we can either go back and add it to the previous line, or we just
+       add a new invocation. The process is very similar to using wpsimp. *)
```
corlewis, line 64: Tiny nitpick. I'm not certain that I've kept the meaning you intended, although that does suggest it would be good to change it in some way.
```suggestion
       for a corres rule that mentions those two turns up the rule "f", which we can provide to the corres
       method. At this point we can either go back and add it to the previous line, or we just
       add a new invocation. The process is very similar to using wpsimp. *)
```

lib/test/Corres_Test.thy
```diff
@@ -0,0 +1,306 @@
...
+    apply (corres corres: f)
+      (* We see that f has been split off, and we now have a goal for g. Same process as above finds
+         the corresponding rule. *)
+      apply (corres corres: g)
+      (* This solves the corres goal, but leaves over the side condition of the "g" rule. We can
```
corlewis, line 69:
```suggestion
      (* This solves the corres goal, but leaves the side condition of the "g" rule. We can
```

lib/test/Corres_Test.thy
```diff
@@ -0,0 +1,306 @@
...
+  finish
+  done
+
+(* Rewriting corres terms *)
+lemma "corres (=) P \<top> (do x \<leftarrow> f; g x od) (do y \<leftarrow> liftM t f'; g' y od)"
+  (* In this goal, corres will stop at liftM without finding a rule to apply. Unfolding
+     liftM_def exposes the bare f' to the toplevel and lets it apply the existing "f" rule.
+     The "t" rewrite happens to solve the now more complex side condition for g.
+     Unfolding liftM_def is generally preferred to the liftM corres simp rules, because
+     these transform schematic guards in ways that later hinder unification. *)
```
corlewis, line 171: This is more of a comment on the `corres` method and not for this PR, but this is a bit unintuitive for me. Normally I would try to avoid unfolding low-level definitions like `liftM` and instead use explicit rules that transform the various parameters and guards appropriately.

corlewis (approved):
Very nice documentation/tutorial! It's cool to see an actual use of `apply_debug` like that, it's not something I ever think of.

proof/refine/RISCV64/ArchAcc_R.thy
```diff
@@ -334,8 +334,9 @@ lemma corres_cross_over_pte_at:
   done
 
 lemma getObject_PTE_corres[corres]:
-  "corres pte_relation' (pte_at p and pspace_aligned and pspace_distinct) \<top>
-          (get_pte p) (getObject p)"
+  "p = p' \<Longrightarrow>
+   corres pte_relation' (pte_at p and pspace_aligned and pspace_distinct) \<top>
+          (get_pte p) (getObject p')"
```
Xaphiosis, line 339: why not use `[@lift_corres_args]` on original lemma?

lib/Corres_Method.thy
```diff
@@ -42,6 +42,9 @@ method is_hoare_schematic_post =
   succeeds \<open>wp_pre, rule hoare_post_False hoareE_post_False\<close>,
   succeeds \<open>wp_pre, rule wp_post_taut wp_post_tautE\<close>
 
+(* Succeed if wpsimp or wp can safely be applied *)
+method is_safe_wp = is_wp, fails \<open>is_hoare_schematic_post\<close>
```
Xaphiosis, line 46: I think the changes in this file are in the wrong commit.

lib/Corres_Method.thy
```diff
@@ -161,6 +161,10 @@ method corres
 
 section \<open>Corres rule setup\<close>
 
+(* Avoid using equations in the assumptions. subst_all gets around (no_asm_use) in some cases,
+   which we don't want. *)
+lemmas [corres_no_simp] = subst_all
```
Xaphiosis, line 169: I don't understand this one. I also don't understand how it is added in this PR, but its use seems to already appear before this PR.

lib/test/Corres_Test.thy
```diff
@@ -0,0 +1,306 @@
...
+
+  fixes Q g g' t
+  assumes g: "\<And>x x'::nat. x = t x' \<Longrightarrow> corres_underlying sr nf nf' (=) Q \<top> (g x) (g' x')"
+  assumes t: "\<And>x. t x = x"
+
+  fixes P
+  assumes Q: "\<lbrace>P\<rbrace> f \<lbrace>\<lambda>_. Q\<rbrace>"
+
+  fixes h h'
+  assumes h: "corres_underlying sr nf nf' (=) \<top> \<top> h h"
```
Xaphiosis, line 28: shouldn't this be `h h'`?

lib/test/Corres_Test.thy
```diff
@@ -0,0 +1,306 @@
...
+begin
+
+lemma "corres (=) P \<top> (do x \<leftarrow> f; g x od) (do y \<leftarrow> f'; g' y od)"
+  by corres
+
+end
+
+(* During development, the rules needed are often not declared [corres] yet or the right
+   simp rules for side conditions etc have yet to be figured out. The following proofs
+   demonstrates this process. *)
```
Xaphiosis, line 56: `proof demonstrates` or `proofs demonstrate`

lib/test/Corres_Test.thy
```diff
@@ -0,0 +1,306 @@
...
+lemma "corres (=) P \<top> (do x \<leftarrow> f; g x od) (do y \<leftarrow> f'; g' y od)"
+  (* We begin by just invoking "corres" *)
+  apply corres
+    (* In this case, not much has happened yet, corres has only produced schematic preconditions.
+       However, we can see that f and f' are the heads of both sides, and searching with find_theorems
+       for a corres rule that mentions those two turns up the rule "f". We provided it to the corres
+       method. At this point we can either go back and add it to the previous line, or we just
```
Xaphiosis, line 63: so much just

lib/test/Corres_Test.thy
Xaphiosis: agreed

lib/test/Corres_Test.thy
```diff
@@ -0,0 +1,306 @@
...
+  apply corres
+    apply (corres corres: f g simp: t wp: Q)
+    (* This will still leave the final implication, because we have produced that implication
+       outside this subgoal. Merging the two corres invocations above will attempt the final
+       implications automatically as well. *)
+   apply simp
+  apply simp
+  done
+
+
```
Xaphiosis, line 108: double empty lines usually go between sections of some kind, but there isn't anything like that between the setup and start of examples

lib/test/Corres_Test.thy
```diff
@@ -0,0 +1,306 @@
...
+(* More controlled single-stepping. *)
+
+(* Sometimes invoking "corres" does too much or too little.
+   Too much can occur when the method applies a rule we didn't know is in the [corres] set and
+   that leaves us with a strange side condition to solve. Or we may have added an unsafe,
```
Xaphiosis, line 113: that leaves -> which leaves?

lib/test/Corres_Test.thy
```diff
@@ -0,0 +1,306 @@
...
+   not-really-terminal rule to [corres] and now we are getting an unprovable goal. Too little
+   can occur when the method refuses to split off the head terms even though it looks like a
+   terminal corres rule should apply. For these cases, we can take apart some of the internal
+   steps like this: *)
+lemma "corres (=) P \<top> (do x \<leftarrow> f; g x od) (do y \<leftarrow> f'; g' y od)"
+  (* Controlled way to just introduce schematic preconditions and the final implication *)
```
Xaphiosis, line 119: I'd prefer a subset of the "just" in this file switch to "only"

lib/Corres_Method.thy
corlewis: I was confused by this as well. My understanding is that `subst_all` was causing simplification problems before this PR. The change here is to add it to the `corres_no_simp` set, which is then passed to `simp del` inside of the `corres` method (https://github.com/seL4/l4v/blob/555d1a49427171c7283be5a26bd369bef4d028d8/lib/Corres_Method.thy#L130C59)

lib/test/Corres_Test.thy
```diff
@@ -0,0 +1,306 @@
...
+       will give us the goal the terminal corres rule is tried on: *)
+    apply corres_split
+       (* Trying out "rule f" does work now -- if it didn't we could debug that and find out why *)
+       apply (succeeds \<open>rule f\<close>)
+       (* Turns out we just forgot to declare it, so we add it manually, and the corres method now
+          succeeds on the subgoal *)
+       apply (corres corres: f)
+      (* For the next goal, we have only g. Maybe we want to debug why corres doesn't solve the
+         application of the "g" rule automatically, or where the "x = t x" side condition comes from.
+         To do that, we can just manually apply the rule: *)
```
Xaphiosis, line 132: 4th "just" in this lemma, are you trying to make me cry? :)

lib/Corres_Method.thy
Xaphiosis: Ahh, so my brain blanked on this alternate syntax to `declare subst_all[corres_no_simp]`, thanks for the clarification!

lib/test/Corres_Test.thy
```diff
@@ -0,0 +1,306 @@
...
+lemma "corres (=) P \<top> (do x \<leftarrow> f; g x od) (do y \<leftarrow> f'; g' y od)"
+  (* There is also a corres_cleanup breakpoint for further data *)
+  apply_debug (trace) (tags "corres", "corres_cleanup") (corres corres: f g simp: t wp: Q)
+  continue (* guard implication *)
+    continue (* application of f *)
+      continue (* application of g, showing side condition *)
+  continue (* solve side condition (separate goal) *)
+     continue (* wpsimp+, which happens to solve all remaining goals *)
+  finish
+  done
```
Xaphiosis, line 163: very cool, thank you for these breakpoint examples

lib/test/Corres_Test.thy
Xaphiosis: It possibly suggests a comment to add above the liftM corres simp rules

lib/Corres_Method.thy
corlewis: Yeah, it isn't a syntax that I've consciously used before. Based on the following examples in this file it looks like a more compact way to simultaneously give multiple lemmas the same attribute, although we would have to be careful that we aren't modifying any of them and putting unnamed lemmas into attribute sets.

lib/test/Corres_Test.thy
```diff
@@ -0,0 +1,306 @@
...
+     of terminal corres steps, not to the corres terms themselves. Usually those simp rules
+     are fairly distinct and side-condition rules don't do anything to the corres terms, so
+     it's fine to just put them in the "simp:" section, but occasionally we want more control. *)
+  by (corres corres: f g simp: liftM_def term_simp: t wp: Q)
+
+(* Dealing with asserts and symbolic execution *)
+lemma "corres (=) P \<top> (do s \<leftarrow> get; assert (P s); x \<leftarrow> f; g x od) (do y \<leftarrow> f'; g' y od)"
+  (* Here we'd like to do symbolic execution on "get" and then use the unsafe rule
+     corres_assert_gen_asm_l for the assert. Often it is good enough to just locally
+     provide them as [corres], but just adding corres_symb_exec_l here for instance will
```
Xaphiosis, line 186: Provide "them"? Which them? Also "just locally provide ..., but just ..." ...

lib/test/Corres_Test.thy
```diff
@@ -0,0 +1,306 @@
...
+     go too far. It will try to execute all of get, assert, and f: *)
+  apply (corres corres: corres_symb_exec_l[where P=P])
+  (* unsolvable *)
+  oops
+
+lemma "corres (=) P \<top> (do s \<leftarrow> get; assert (P s); x \<leftarrow> f; g x od) (do y \<leftarrow> f'; g' y od)"
+  (* We can provide the same rule as a fallback rule. This means it will be tried only when
+     no other rule has worked. This lets f and corres_assert_gen_asm_l go first. *)
+  by (corres corres: corres_assert_gen_asm_l f g
+             fallback: corres_symb_exec_l[where P=P]
```
Xaphiosis, line 196: nice, I really wanted to be able to do this and am very glad we can

lib/test/Corres_Test.thy
```diff
@@ -0,0 +1,306 @@
...
+(* Providing something like fastforce can lead to non-termination or slowdown, because the method
+   will be tried for any side condition. If there is a distinctive goal pattern that can
+   distinguish when the cleanup method should be run, you can use "match" to restrict the method: *)
+lemma "corres (=) P \<top> (do x \<leftarrow> f; g x; h od) (do y \<leftarrow> f'; g' y; h od)"
+  by (corres' \<open>match conclusion in "x = t y" for x y \<Rightarrow> \<open>fastforce simp: t\<close>\<close> corres: f g h wp: Q)
+
+
+(* Form of [corres] rules *)
+
+(* The method expects terminal corres rule to instantiate return relation and guards.
```
Xaphiosis, line 242: rules?

lib/test/Corres_Test.thy
```diff
@@ -0,0 +1,306 @@
...
+   It also expects distinct variables for the abstract and concrete side and tries hard to
+   not accidentally mix these by rewriting corres terms with assumptions.
+
+   For instance, it would be tempting to write the "g" rule as follows:
+*)
+lemma g': "corres (=) Q \<top> (g x) (g' x)"
+  by (simp add: g t)
```
Xaphiosis, line 249: would adding an example using `[@lift_corres_args]` be of interest around here (if it's in scope)?

lib/test/Corres_Test.thy
```diff
@@ -0,0 +1,306 @@
...
+
+
+(* Form of [corres] rules *)
+
+(* The method expects terminal corres rule to instantiate return relation and guards.
+   It also expects distinct variables for the abstract and concrete side and tries hard to
+   not accidentally mix these by rewriting corres terms with assumptions.
+
+   For instance, it would be tempting to write the "g" rule as follows:
+*)
```
Xaphiosis, line 247: `*)` to prev line for consistency

lib/test/Corres_Test.thy
```diff
@@ -0,0 +1,306 @@
...
+   instead wherever possible, because it provides some protection against common pitfalls.
+
+   Occasionally it is useful to interleave with tactics that work on specific kinds of goals
+   only, e.g. a clarsimp on goals that are not corres goals. For this, the predicate methods
+   is_corres, is_wp, and is_safe_wp are available. These do not change the proof state, but they
+   fail when their predicate does not hold.
+
+   is_corres succeeds on corres goals only
+   is_wp succeeds on wp goals only (valid, validE, no_fail)
+   is_safe_wp succeeds only on wp goals without schematic post condition (where wpsimp is not safe)
```
Xaphiosis, line 286: `a` schematic postcondition?

lib/test/Corres_Test.thy
```diff
@@ -0,0 +1,306 @@
...
+
+   Boolean combinations of predicates and be achieved with "," "|" and "fails" for "and", "or", and
+   "not".
+*)
```
Xaphiosis, line 290: would `Boolean combinations of predicates can be obtained using "," (and), "|" (or) and "fails" (not).` be better? In any case the first `and` should probably be `can`.

lib/test/Corres_Test.thy
```diff
@@ -0,0 +1,306 @@
...
+
+(* Example for predicate methods *)
```
Xaphiosis, line 292: for -> of?

lib/test/Corres_Test.thy
```diff
@@ -0,0 +1,306 @@
...
+lemma "corres (=) P \<top>
+              (do x \<leftarrow> case z of None \<Rightarrow> f | Some x' \<Rightarrow> do return x'; f od; g x od)
+              (do y \<leftarrow> f'; g' y od)"
+  (* Do case distinction and apply the corres method only to the corres goals: *)
+  apply (corres_cases; (is_corres, corres corres: f g)?)
+         (* Find all safe wp goals and run wpsimp on them *)
+         apply (all \<open>(is_safe_wp, wpsimp wp: Q)?\<close>)
+     (* There should now only be non-corres and non-wp goals left -- fail if that is not the case *)
```
Xaphiosis, line 300: `Only non-corres and non-wp goals should remain`?

Xaphiosis (approved):
Good examples, definitely very helpful. There's some confusion about which commit some changes are in, and I left a bit of feedback on the text in comments along with  begging for a review of `just` frequency. Might also make sense to use sections in the test file, I find it hard to tell in places what is a section header and what isn't.

lib/Corres_Method.thy
Xaphiosis: Mm. Yes, I think it's because we normally either see `[simp]` as the attribute, or we see multiple lemmas. Seeing `[corres_no_simp]` and only one lemma threw me off, but it's the more stable notation if we want to add more to the set in future. Regarding the modifying, I think `declare` would have the same problem. I guess you could remove it from the set using the same modified thm.

lib/test/Corres_Test.thy
lsf37: Yes, a comment for the rule would make sense. Will do.

It's not really anything to do with the method, it's just that the rule is not a real wp-style rule, because it transforms the relation in the wrong way.

We're probably going to hit a few more such cases, which we haven't noticed so far, because we've always manually worked around them or just wondered why stuff didn't work and why we needed to instantiate things manually, but then moved on without investigating deeply.

proof/refine/RISCV64/ArchAcc_R.thy
lsf37: I'm mostly trying to establish this form as the canonical form for writing corres lemmas to reduce friction. We could also state other lemmas with object implication and then use them with `rule_format`, but we don't do that, because it would be annoying to use `[rule_fromat]` everywhere.

lib/Corres_Method.thy
lsf37: I can confirm both the syntax and the intention discussed in this thread. I'll mark it as resolved, but if you want me to do anything to help clear up confusion, let me know.

lib/test/Corres_Test.thy
lsf37: Well spotted.

lib/test/Corres_Test.thy
lsf37: I had the same idea, but it's not in scope yet (it's declared for the CorresK method, I think). We could refactor all that a bit more, but I'd prefer to leave that one for another time.

lib/test/Corres_Test.thy
lsf37: obtained is much better, thanks!

lib/test/Corres_Test.thy
lsf37: rephrased it a bit more

lsf37:
> Good examples, definitely very helpful. There's some confusion about which commit some changes are in, and I left a bit of feedback on the text in comments along with begging for a review of `just` frequency. 

I re-read the text and I wholeheartedly apologise for the unjust frequency of `just`. Not sure what was riding me, but it really was way too much. Frequency is drastically reduced now.

> Might also make sense to use sections in the test file, I find it hard to tell in places what is a section header and what isn't.

Yes, I should probably do that. Will try it out.

PR #657, opened 2023-07-04
monads: synchronise with rt branch

corlewis:
This pulls in changes made to `Monads` on the `rt` branch, in preparation for further refactoring.

### DIFF
--- a/lib/Monads/NonDetMonadVCG.thy
+++ b/lib/Monads/NonDetMonadVCG.thy
@@ -306,6 +306,8 @@ lemma hoare_gen_asm:
   "(P \<Longrightarrow> \<lbrace>P'\<rbrace> f \<lbrace>Q\<rbrace>) \<Longrightarrow> \<lbrace>P' and K P\<rbrace> f \<lbrace>Q\<rbrace>"
   by (fastforce simp add: valid_def)
 
+lemmas hoare_gen_asm_single = hoare_gen_asm[where P'="\<top>", simplified pred_conj_def simp_thms]
+
 lemma hoare_gen_asm_lk:
   "(P \<Longrightarrow> \<lbrace>P'\<rbrace> f \<lbrace>Q\<rbrace>) \<Longrightarrow> \<lbrace>K P and P'\<rbrace> f \<lbrace>Q\<rbrace>"
   by (fastforce simp add: valid_def)
--- a/lib/Monads/OptionMonadWP.thy
+++ b/lib/Monads/OptionMonadWP.thy
@@ -21,8 +21,10 @@ begin
    TODO: design a sensible syntax for them. *)
 
 (* Partial correctness. *)
-definition ovalid :: "('s \<Rightarrow> bool) \<Rightarrow> ('s \<Rightarrow> 'a option) \<Rightarrow> ('a \<Rightarrow> 's \<Rightarrow> bool) \<Rightarrow> bool" where
-  "ovalid P f Q \<equiv> \<forall>s r. P s \<and> f s = Some r \<longrightarrow> Q r s"
+definition ovalid :: "('s \<Rightarrow> bool) \<Rightarrow> ('s \<Rightarrow> 'a option) \<Rightarrow> ('a \<Rightarrow> 's \<Rightarrow> bool) \<Rightarrow> bool"
+  ("\<lblot>_\<rblot>/ _ /\<lblot>_\<rblot>") where
+  "\<lblot>P\<rblot> f \<lblot>Q\<rblot> \<equiv> \<forall>s r. P s \<and> f s = Some r \<longrightarrow> Q r s"
+
 (* Total correctness. *)
 definition ovalidNF :: "('s \<Rightarrow> bool) \<Rightarrow> ('s \<Rightarrow> 'a option) \<Rightarrow> ('a \<Rightarrow> 's \<Rightarrow> bool) \<Rightarrow> bool" where
   "ovalidNF P f Q \<equiv> \<forall>s. P s \<longrightarrow> (f s \<noteq> None \<and> (\<forall>r. f s = Some r \<longrightarrow> Q r s))"
@@ -57,6 +59,10 @@ lemmas owhile_add_inv = owhile_inv_def[symmetric]
 
 
 (* WP rules for ovalid. *)
+lemma ovalid_inv[wp]:
+  "ovalid P f (\<lambda>_. P)"
+  by (simp add: ovalid_def)
+
 lemma obind_wp[wp]:
   "\<lbrakk> \<And>r. ovalid (R r) (g r) Q; ovalid P f R \<rbrakk> \<Longrightarrow> ovalid P (obind f g) Q"
   by (simp add: ovalid_def obind_def split: option.splits, fast)
--- a/lib/Monads/WhileLoopRules.thy
+++ b/lib/Monads/WhileLoopRules.thy
@@ -383,6 +383,22 @@ lemma whileLoop_wp:
   \<lbrace> I r \<rbrace> whileLoop C B r \<lbrace> Q \<rbrace>"
   by (rule valid_whileLoop)
 
+lemma whileLoop_valid_inv:
+  "(\<And>r. \<lbrace> \<lambda>s. I r s \<and> C r s \<rbrace> B r \<lbrace> I \<rbrace>) \<Longrightarrow> \<lbrace> I r \<rbrace> whileLoop C B r \<lbrace> I \<rbrace>"
+  apply (fastforce intro: whileLoop_wp)
+  done
+
+lemma valid_whileLoop_cond_fail:
+  assumes pre_implies_post: "\<And>s. P r s \<Longrightarrow> Q r s"
+      and pre_implies_fail: "\<And>s. P r s \<Longrightarrow> \<not> C r s"
+    shows "\<lbrace> P r \<rbrace> whileLoop C B r \<lbrace> Q \<rbrace>"
+  using assms
+  apply (clarsimp simp: valid_def)
+  apply (subst (asm) whileLoop_cond_fail)
+   apply blast
+  apply (clarsimp simp: return_def)
+  done
+
 lemma whileLoop_wp_inv [wp]:
   "\<lbrakk> \<And>r. \<lbrace>\<lambda>s. I r s \<and> C r s\<rbrace> B r \<lbrace>I\<rbrace>; \<And>r s. \<lbrakk>I r s; \<not> C r s\<rbrakk> \<Longrightarrow> Q r s \<rbrakk>
    \<Longrightarrow> \<lbrace> I r \<rbrace> whileLoop_inv C B r I M \<lbrace> Q \<rbrace>"


lib/Monads/OptionMonadWP.thy
```diff
@@ -21,8 +21,15 @@ begin
    TODO: design a sensible syntax for them. *)
 
 (* Partial correctness. *)
-definition ovalid :: "('s \<Rightarrow> bool) \<Rightarrow> ('s \<Rightarrow> 'a option) \<Rightarrow> ('a \<Rightarrow> 's \<Rightarrow> bool) \<Rightarrow> bool" where
-  "ovalid P f Q \<equiv> \<forall>s r. P s \<and> f s = Some r \<longrightarrow> Q r s"
+definition ovalid :: "('s \<Rightarrow> bool) \<Rightarrow> ('s \<Rightarrow> 'a option) \<Rightarrow> ('a \<Rightarrow> 's \<Rightarrow> bool) \<Rightarrow> bool"
+  ("o\<lbrace>_\<rbrace>/ _ /\<lbrace>_\<rbrace>") where
```
michaelmcinerney, line 25: I don't think we were happy with this syntax. I can't think of anything nicer, and I know that our options for fancy symbols are limited. Is there maybe something that looks like `{|...|}`, but replacing the curly braces with some other pair that's open and closed?

corlewis: Oh, good point. I had only been thinking getting rt in sync with master but if we can think of an alternative then this would be the right time to change that syntax.

lsf37: Right, yes, there was that. Did we come up with something in the meeting? The main problem we had was that `o{| .. |}` looks like there is a left-over function composition somewhere. The same would unfortunately apply to `{| .. |}o`. 

We could make it `⦃ .. ⦄⇩o` (subscript `o`), but that is a lot to type.

Any other ideas?

corlewis: Yeah, I thought about subscript as well but it's such a pain to type.
Looking at unused brackets, we could use `⦉P⦊ f ⦉Q⦊`. Still not the easiest to type as they don't have a unique abbreviation, but presumably that can be changed.

lsf37: The more I think about it, the less I like ovalid as the concept that it is now (nothing to do with this PR as such, it just made me think about it). The post condition should just be a about the return value, because the state can't change (by type alone)-- it doesn't make sense to carry a state predicate in the post condition and have bind rules for it etc. Of course that's now a bit late to change in AutoCorres, but maybe that is still what we should do before it gets even more ingrained.

How much ovalid are we using in the proofs?

michaelmcinerney: I'm really not sure about the use of `ovalid` in `master`, but it's used here and there for MCS. Not nearly as much as it could be used: I've had to write some pretty low level rules for `ovalid` about `get_object`, etc., but that's really just come up very recently.

I like the idea that we shouldn't mention the state in the postcondition. Before I put more rules for reader monad concepts into Lib, it might be good to change this. And I do like Corey's idea for the syntax.

corlewis: I'm not sure if I understand what the alternative is, don't we want to be able to connect return values to the state? If that's the case then how could we have the state in the pre condition but not the post? What would `obind_wp` and `ogets_wp` look like?

Separately, I think I've changed my mind. For my purposes it would be good to get something like this PR merged quickly. Are we ok with merging it as is and discussing/modifying `ovalid` in parallel?

lsf37: Yes, ok, let's not block this PR on that. Corey's syntax suggestion is a bit better, but not ideal yet, because it doesn't suggest anything about option or reader monad. If you want to get this merged quickly, I'm fine with Corey's syntax for now until we can find something better, though.

For connecting the state and the return value: you're right, you'd want the state to be available for the post condition for that purpose. My first reaction would be to just have a post condition then, but it's possible that you want to be able to split off the state-only part so you can propagate it. Maybe that's where this all came from. But do we really need to propagate it?

In either case, that might suggest a different syntax where there is no precondition, just a state + rv condition and maybe an additional state condition. We can change that after this PR, though.

One more idea for the state in this one: since the state does not change, we could just name it and leave conditions on it to normal Isabelle preconditions. E.g. something like `P s ==> f {| = s. %r. Q s r |}` where the `= s` is a free `s` and the definition of `f {| = s. Q |}` would be `!r. f s = Some r --> Q s r`. I don't know if that would work nicely, but might be worth exploring a bit.

lib/Monads/NonDetMonadVCG.thy
```diff
@@ -709,6 +711,11 @@ lemma hoare_vcg_if_split:
   "\<lbrakk> P \<Longrightarrow> \<lbrace>Q\<rbrace> f \<lbrace>S\<rbrace>; \<not>P \<Longrightarrow> \<lbrace>R\<rbrace> g \<lbrace>S\<rbrace> \<rbrakk> \<Longrightarrow> \<lbrace>\<lambda>s. (P \<longrightarrow> Q s) \<and> (\<not>P \<longrightarrow> R s)\<rbrace> if P then f else g \<lbrace>S\<rbrace>"
   by simp
 
+lemma hoare_vcg_if_split_strong:
+ "\<lbrakk> P \<Longrightarrow> \<lbrace>Q\<rbrace> f \<lbrace>S\<rbrace>; \<not>P \<Longrightarrow> \<lbrace>R\<rbrace> g \<lbrace>S\<rbrace> \<rbrakk> \<Longrightarrow>
+  \<lbrace>\<lambda>s. if P then Q s else R s\<rbrace> if P then f else g \<lbrace>S\<rbrace>"
```
lsf37, line 716: I wouldn't call that rule `_strong`, it's not stronger than the existing rules and it doesn't really fit and automatic wp or sp calculus (for sp it should be something like an `S and S'`, because it is unlikely that both branches will have the same calculated post condition). How about `hoare_vcg_if_post_join`?

lib/Monads/OptionMonadWP.thy
```diff
@@ -21,8 +21,15 @@ begin
...
+  "o\<lbrace>P\<rbrace> f \<lbrace>Q\<rbrace> \<equiv> \<forall>s r. P s \<and> f s = Some r \<longrightarrow> Q r s"
+
+abbreviation (input)
+  oinvariant :: "('s,'a) lookup \<Rightarrow> ('s \<Rightarrow> bool) \<Rightarrow> bool" ("_ o\<lbrace>_\<rbrace>" [59,0] 60)
+where
+  "oinvariant f P \<equiv> o\<lbrace>P\<rbrace> f \<lbrace>\<lambda>_. P\<rbrace>"
+
```
lsf37, line 32: We shouldn't need this one, because this statement is always true

lib/Monads/WhileLoopRules.thy
```diff
@@ -383,6 +383,22 @@ lemma whileLoop_wp:
   \<lbrace> I r \<rbrace> whileLoop C B r \<lbrace> Q \<rbrace>"
   by (rule valid_whileLoop)
 
+lemma whileLoop_wp':
+  "(\<And>r. \<lbrace> \<lambda>s. I r s \<and> C r s \<rbrace> B r \<lbrace> I \<rbrace>) \<Longrightarrow> \<lbrace> I r \<rbrace> whileLoop C B r \<lbrace> I \<rbrace>"
```
lsf37, line 387: The name `_wp'` doesn't really work, because the rule forces the post condition to be the invariant, so it's not a `wp` rule any more. Maybe `whileLoop_inv`? (not sure if that one is taken already)

lib/Monads/WhileLoopRules.thy
```diff
@@ -383,6 +383,22 @@ lemma whileLoop_wp:
...
+  apply (fastforce intro: whileLoop_wp)
+  done
+
+lemma valid_whileLoop_cond_fail:
+  assumes pre_implies_post: "\<And>s. P r s \<Longrightarrow> Q r s"
+      and pre_implies_fail: "\<And>s. P r s \<Longrightarrow> \<not> C r s"
+    shows "\<lbrace> P r \<rbrace> whileLoop C B r \<lbrace> Q \<rbrace>"
+  apply (insert assms)
```
lsf37, line 395: I'd usually do `using assms` for this purpose. Not entirely sure why I prefer that, maybe just because `insert` usually affects all goals and so we tend to avoid it.

Since we're not referring to the names of the assumptions, we could also remove the names (also fine to leave them in as documentation if you want).

lib/Monads/NonDetMonadVCG.thy
```diff
@@ -306,6 +306,8 @@ lemma hoare_gen_asm:
   "(P \<Longrightarrow> \<lbrace>P'\<rbrace> f \<lbrace>Q\<rbrace>) \<Longrightarrow> \<lbrace>P' and K P\<rbrace> f \<lbrace>Q\<rbrace>"
   by (fastforce simp add: valid_def)
 
+lemmas hoare_gen_asm_single = hoare_gen_asm[where P'="\<top>", simplified pred_conj_def simp_thms]
```
lsf37, line 309: It's not quite the same, but would `hoare_assume_pre` already work for your purposes, or do you need to match against a specific `K` precondition?

lib/Monads/OptionMonadWP.thy
lsf37: > I'm really not sure about the use of `ovalid` in `master`, but it's used here and there for MCS. Not nearly as much as it could be used: I've had to write some pretty low level rules for `ovalid` about `get_object`, etc., but that's really just come up very recently.

This is actually quite encouraging. If it's not used much in MCS either yet, then there is probably not much reader monad use overall. For the reader/projection proofs in master I don't actually use any special syntax at all, I just write lemmas of the form `[| f s = Some r; P s |] ==> Q s r`. Since `obind` and `opt_map` decompose nicely in that `f = Some` form, there is no further machinery. I don't know if that's ideal, it's just what I ended up doing.

lsf37 (approved):
Modulo the naming changes, I'm happy for this to go ahead with Corey's syntax for now. We can change that again and discuss more what that calculus should look like separately.

corlewis:
Just clarifying that I wasn't the original author of any of these changes, I'm just pulling them into master from rt. I'm happy to take this opportunity to rename but will need to make another PR on rt to keep it consistent and to fix any uses of the lemmas.

lsf37:
> Just clarifying that I wasn't the original author of any of these changes, I'm just pulling them into master from rt. I'm happy to take this opportunity to rename but will need to make another PR on rt to keep it consistent and to fix any uses of the lemmas.

Yes, of course. Sorry I missed all these on the rt side. It's much easier to spot when it's just a very small PR :-)

lib/Monads/NonDetMonadVCG.thy
corlewis: I'm not sure what the original purpose for this one was. I can investigate if you want, currently it only seems to be used in one place.

corlewis:
Hmm, I just realised that it would both be better to do these renames first on `rt` to make sure that we're happy with them and also that I should be pulling from `current-rt` instead. We have made quite a few more changes to the `Monads` session on that branch.

I'll be back soon with an expanded PR here, so that we can easily notice any other changes we want, and then new PRs on those branches.

lib/Monads/NonDetMonadVCG.thy
corlewis: It looks like this isn't used in rt anymore, the lemma it was written for was redone at some point. Is it worth keeping or should I just remove it?

lib/Monads/WhileLoopRules.thy
corlewis: Hmm, it's not taken but there's a bit of a conceptual nameclash with the `whileLoop_inv` definition and 
```
lemma whileLoop_wp_inv [wp]:
  "⟦ ⋀r. ⦃λs. I r s ∧ C r s⦄ B r ⦃I⦄; ⋀r s. ⟦I r s; ¬ C r s⟧ ⟹ Q r s ⟧
   ⟹ ⦃ I r ⦄ whileLoop_inv C B r I M ⦃ Q ⦄"
```

lib/Monads/NonDetMonadVCG.thy
lsf37: I'd remove it

lib/Monads/WhileLoopRules.thy
lsf37: That does actually mesh, I think. The `wp_inv` variant is a wp rule, and the `inv` variant would not be, but is related. But the definition name would be confusing. How about `whileLoop_hoare_inv`  or `whileLoop_valid_inv`?

lib/Monads/NonDetMonadVCG.thy
corlewis: Ok, the only remaining use of this lemma can be replaced by `hoare_assume_pre`, so if we want to we could remove it.

However, to me it sort of makes sense to keep it. `hoare_gen_asm` is our 'standard' way of pulling assumptions out of valid and it would be annoying to have to remember to use `hoare_assume_pre` for the specific case where there isn't anything else in the precondition.

corlewis: Hmm, although really, what I might be arguing for instead is something along the lines of
```
lemmas hoare_gen_asms =
  hoare_gen_asm hoare_gen_asm' hoare_gen_asm_conj hoare_gen_asm_single
```

PR #658, opened 2023-07-06
MCS: monads cleanup

corlewis:
These are changes to the Monads directory that were suggested in https://github.com/seL4/l4v/pull/657.

### DIFF
--- a/lib/Monads/OptionMonadWP.thy
+++ b/lib/Monads/OptionMonadWP.thy
@@ -22,13 +22,8 @@ begin
 
 (* Partial correctness. *)
 definition ovalid :: "('s \<Rightarrow> bool) \<Rightarrow> ('s \<Rightarrow> 'a option) \<Rightarrow> ('a \<Rightarrow> 's \<Rightarrow> bool) \<Rightarrow> bool"
-  ("o\<lbrace>_\<rbrace>/ _ /\<lbrace>_\<rbrace>") where
-  "o\<lbrace>P\<rbrace> f \<lbrace>Q\<rbrace> \<equiv> \<forall>s r. P s \<and> f s = Some r \<longrightarrow> Q r s"
-
-abbreviation (input)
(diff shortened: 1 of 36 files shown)


michaelmcinerney (approved):

lib/Monads/OptionMonadWP.thy
```diff
@@ -22,13 +22,8 @@ begin
...
 definition ovalid :: "('s \<Rightarrow> bool) \<Rightarrow> ('s \<Rightarrow> 'a option) \<Rightarrow> ('a \<Rightarrow> 's \<Rightarrow> bool) \<Rightarrow> bool"
-  ("o\<lbrace>_\<rbrace>/ _ /\<lbrace>_\<rbrace>") where
-  "o\<lbrace>P\<rbrace> f \<lbrace>Q\<rbrace> \<equiv> \<forall>s r. P s \<and> f s = Some r \<longrightarrow> Q r s"
-
-abbreviation (input)
-  oinvariant :: "('s,'a) lookup \<Rightarrow> ('s \<Rightarrow> bool) \<Rightarrow> bool" ("_ o\<lbrace>_\<rbrace>" [59,0] 60)
-where
-  "oinvariant f P \<equiv> o\<lbrace>P\<rbrace> f \<lbrace>\<lambda>_. P\<rbrace>"
+  ("\<lblot>_\<rblot>/ _ /\<lblot>_\<rblot>") where
+  "\<lblot>P\<rblot> f \<lblot>Q\<rblot> \<equiv> \<forall>s r. P s \<and> f s = Some r \<longrightarrow> Q r s"
```
lsf37, line 26: Do we use this anywhere? (Because it really should alway be trivially true, so we should never need to state it)

corlewis: That's the definition of `ovalid`, this change is already removing `oinvariant` and replacing the one use of it with the new `ovalid_inv` lemma.

lsf37: Right, sorry I completely misread the diff. All perfect then, and ready to merge from my side!

PR #660, opened 2023-07-19
Monads refactor

corlewis:
Separate out the different monads into subdirectories and use a consistent naming scheme with prefixes. This is in preparation for the trace monad to be eventually expanded to match the nondet monad, which will lead to a lot of similar files.

For reviewing purposes, I think the main things to look at are the names and locations of the files in the `monads` directory and subdirectories. Almost all other changes are the result of `sed` commands.


corlewis:
In some following work I got a bit annoyed by the random order of files in `Monads/ROOT`, so I took the opportunity to reorder it. I didn't put that much thought into it, I was mainly wanting to be consistent and group related files. There's almost definitely further improvements that could be made if dependencies were also considered.

I'm happy to go with a different order or even revert and ignore it if people prefer.

lsf37 (approved):
Looking good. While going through the files, I was wondering if we should also go for `WhileLoopRules` -> `While_Loop_Rules`, but I'm not sure I care enough :-)

lsf37:
> In some following work I got a bit annoyed by the random order of files in `Monads/ROOT`, so I took the opportunity to reorder it. I didn't put that much thought into it, I was mainly wanting to be consistent and group related files. There's almost definitely further improvements that could be made if dependencies were also considered.
> 
> I'm happy to go with a different order or even revert and ignore it if people prefer.

I'm happy with the reordering.

corlewis:
> Looking good. While going through the files, I was wondering if we should also go for `WhileLoopRules` -> `While_Loop_Rules`, but I'm not sure I care enough :-)

Good idea, that is definitely more consistent.

Xaphiosis (approved):
Nicely done. The new names will take a while to get used to, but at they're consistent now.

PR #663, opened 2023-08-11
Update proofs for AArch64 seL4 PR 

lsf37:
Update proofs for the changes in seL4/seL4#801, in particular commit https://github.com/seL4/seL4/commit/b3cc85242563b9e08c243b62e181c4a8f075239c

### DIFF
--- a/proof/crefine/ARM_HYP/Retype_C.thy
+++ b/proof/crefine/ARM_HYP/Retype_C.thy
@@ -2332,6 +2332,9 @@ definition
                             | ARM_HYP_H.PageDirectoryObject \<Rightarrow> scast seL4_ARM_PageDirectoryObject
                             | ARM_HYP_H.VCPUObject \<Rightarrow> scast seL4_ARM_VCPUObject"
 
+(* always unfold StrictC'_mode_object_defs together with api_object_defs *)
+lemmas api_object_defs = api_object_defs StrictC'_mode_object_defs
+
 lemmas nAPIObjects_def = seL4_NonArchObjectTypeCount_def
 
 lemma nAPIOBjects_object_type_from_H:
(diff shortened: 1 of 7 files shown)


proof/crefine/ARM_HYP/Retype_C.thy
```diff
@@ -2332,6 +2332,9 @@ definition
                             | ARM_HYP_H.PageDirectoryObject \<Rightarrow> scast seL4_ARM_PageDirectoryObject
                             | ARM_HYP_H.VCPUObject \<Rightarrow> scast seL4_ARM_VCPUObject"
 
+(* always unfold StrictC'_mode_object_defs together with api_object_defs *)
+lemmas api_object_defs = api_object_defs StrictC'_mode_object_defs
```
Xaphiosis, line 2336: I'm not 100% on whether having this kind of def override so far into CRefine is a risky approach. Should be ok here, but made me stop and think.

Xaphiosis (approved):

corlewis (approved):
The changes themselves look fine to me but I'm a bit confused. Isn't the C change generally getting rid of the `PageDirectoryObj` and using `PageTableObj` instead?

Otherwise, a very minor comment is that I think the more standard commit tag would be `arm abstract+design` instead of `arm abstract/design`.

proof/crefine/ARM_HYP/Retype_C.thy
lsf37: In this case I think it's fine for ARM/ARM_HYP, because it recreates the old situation before it matters the first time. But generally I agree, it's a bit risky.

lsf37:
> The changes themselves look fine to me but I'm a bit confused. Isn't the C change generally getting rid of the `PageDirectoryObj` and using `PageTableObj` instead?

It doesn't actually get rid of `PageDirectoryObj`, it just moves it to the `mode` file which defines the top-level vspace object. For AArch64 that's going to be called `VSpaceObj`, and  for AArch32 it remains `PageDirectoryObj`, but now has the same enum position as `VSpaceObj` on AArch64.


> Otherwise, a very minor comment is that I think the more standard commit tag would be `arm abstract+design` instead of `arm abstract/design`.

:+1:

PR #664, opened 2023-08-14
Trace Monad refactor

corlewis:
This PR refactors the Trace Monad to have the same general structure and style as the Nondet Monad. In particular:
* the main Trace_Monad file now only contains the basic monad definitions and lemmas about them
* other definitions (like `valid`, `validI`, `no_fail`, `no_trace`, etc.) and their lemmas are in their own files.

The intentions for the commits in this PR and the reviewing suggestions are as follows:
* the first commit updates style and content to match Nondet Monad's and hopefully just needs minimal review
* the second moves content that also appears in Nondet Monad to match the structure of Nondet Monad and hopefully just needs minimal review
* the third moves content specific to the Trace Monad to different files and hopefully just needs the new locations reviewed
* the fourth makes changes to the content specific to the Trace Monad and should be reviewed as normal
* the fifth updates the rest of lib as needed and should be reviewed as normal.


lib/Monads/trace/Trace_Monad.thy
```diff
@@ -572,139 +460,141 @@ lemma "doE x \<leftarrow> returnOk 1;
...
+    s \<leftarrow> get;
+    put_trace [(Me,s)]
+  od"
+
+definition interference :: "('s,unit) tmonad" where
+  "interference \<equiv>
+  do
+    commit_step;
+    env_steps
+  od"
```
Xaphiosis, line 505: I'm reviewing commit-by-commit, so this might be addressed in further commits, but for these 3 defs, the `do ... od` indentation isn't right and I'm looking at the `restyle Trace_Monad.thy` commit. Bit confused.

lib/Monads/trace/Trace_Monad.thy
```diff
@@ -923,81 +784,66 @@ abbreviation (input)
...
-\<close>
-definition
-  validE_R :: "('s \<Rightarrow> bool) \<Rightarrow> ('s, 'e + 'a) tmonad \<Rightarrow>
-               ('a \<Rightarrow> 's \<Rightarrow> bool) \<Rightarrow> bool"
-   ("\<lbrace>_\<rbrace>/ _ /\<lbrace>_\<rbrace>, -")
-where
+  The following two instantiations are convenient to separate reasoning for exceptional and
+  normal case.\<close>
+(* Narrator: they are in fact not convenient, and are now considered a mistake that should have
+             been an abbreviation instead. *)
```
Xaphiosis: does this indicate this should be a FIXME?

lib/Monads/trace/Trace_Monad.thy
```diff
@@ -843,19 +714,21 @@ where
...
+  This section defines a Hoare logic for partial correctness for
+  the interference trace monad as well as the exception monad.
   The logic talks only about the behaviour part of the monad and ignores
   the failure flag.
 
   The logic is defined semantically. Rules work directly on the
   validity predicate.
 
-  In the nondeterministic state monad, validity is a triple of precondition,
+  In the interference trace monad, validity is a triple of precondition,
```
Xaphiosis: oh wow, nicely spotted

lib/Monads/trace/Trace_VCG.thy
```diff
@@ -29,11 +29,11 @@ text \<open>
   bool (a state predicate), the postcondition is a function from return value
   to state to bool. A triple is valid if for all states that satisfy the
   precondition, all result values and result states that are returned by
-  the monad satisfy the postcondition. Note that if the computation returns
-  the empty set, the triple is trivially valid. This means @{term "assert P"}
-  does not require us to prove that @{term P} holds, but rather allows us
-  to assume @{term P}! Proving non-failure is done via a separate predicate and
-  calculus (see Trace_No_Fail).\<close>
+  the monad satisfy the postcondition. Note that if the results of the
```
Xaphiosis, line 32: `results ... are` or `result ... is` ?

lib/Monads/trace/Trace_Monad.thy
```diff
@@ -645,7 +653,8 @@ subsection "Loops"
 text \<open>
   Loops are handled using the following inductive predicate;
   non-termination is represented using the failure flag of the
-  monad.\<close>
+  monad.
+FIXME: update comment about non-termination\<close>
```
Xaphiosis, line 657: was this FIXME intended to be left in?

lib/Monads/trace/Trace_RG.thy
```diff
@@ -108,11 +192,11 @@ lemma guar_cond_append:
   by (simp add: guar_cond_def trace_steps_append ball_Un conj_comms)
 
 lemma prefix_closed_bind:
-  "prefix_closed f \<Longrightarrow> (\<forall>x. prefix_closed (g x)) \<Longrightarrow> prefix_closed (f >>= g)"
+  "\<lbrakk>prefix_closed f; \<forall>x. prefix_closed (g x)\<rbrakk> \<Longrightarrow> prefix_closed (f >>= g)"
   apply (subst prefix_closed_def, clarsimp simp: bind_def)
   apply (auto simp: Cons_eq_append_conv split: tmres.split_asm
              dest!: prefix_closedD[rotated];
-    fastforce elim: rev_bexI)
+         fastforce elim: rev_bexI)
```
Xaphiosis, line 199: this indent doesn't seem right

lib/Monads/trace/Trace_Det.thy
```diff
@@ -0,0 +1,74 @@
+(*
+ * Copyright 2023, Proofcraft Pty Ltd
```
Xaphiosis, line 2: I'm a bit hazy on what we should do when a theory is new but contains only a subset of lines from another file. Should we be trying to pull in copyrights from the source file as well? @lsf37

lib/Monads/trace/Trace_In_Monad.thy
```diff
@@ -0,0 +1,141 @@
...
+  apply (clarsimp simp: bindE_def in_bind lift_def in_throwError)
+  apply (safe del: disjCI; strengthen subst[where P="\<lambda>x. x \<in> mres (f s)", mk_strg I _ E];
+         auto simp: in_throwError split: sum.splits)
+  done
+
+lemmas in_monad = inl_whenE in_whenE in_liftE in_bind in_bindE_L
+                  in_bindE_R in_returnOk in_throwError in_fail
+                  in_assertE in_assert in_return in_assert_opt
+                  in_get in_gets in_put in_when (* unlessE_whenE *)
+                  (* unless_when *) in_modify gets_the_in_monad
```
Xaphiosis, line 122: any idea why the `unless_*` are commented out?

Xaphiosis:
For commit messages, there are some that refer to `trace` as a thing on its own, e.g. `trace files`, but other commits say `trace monad`. Would it be possible to have `trace monad` in all cases? For example, I find `lib: update for trace refactor` confusing

lib/Monads/trace/Trace_Strengthen_Setup.thy
```diff
@@ -0,0 +1,47 @@
...
+  by (cases F, auto elim: hoare_post_impErr)
+
+lemma strengthen_validE_E_cong[strg]:
+  "(\<And>r s. st F (\<longrightarrow>) (S r s) (T r s))
+    \<Longrightarrow> st F (\<longrightarrow>) (\<lbrace>P\<rbrace> f -, \<lbrace>S\<rbrace>) (\<lbrace>P\<rbrace> f -, \<lbrace>T\<rbrace>)"
+  by (cases F, auto elim: hoare_post_impErr simp: validE_E_def)
+
+lemma strengthen_validI[strg]:
+  "(\<And>r s0 s. st F (\<longrightarrow>) (Q r s0 s) (Q' r s0 s))
+    \<Longrightarrow> st F (\<longrightarrow>) (\<lbrace>P\<rbrace>,\<lbrace>G\<rbrace> f \<lbrace>R\<rbrace>,\<lbrace>Q\<rbrace>) (\<lbrace>P\<rbrace>,\<lbrace>G\<rbrace> f \<lbrace>R\<rbrace>,\<lbrace>Q'\<rbrace>)"
```
Xaphiosis, line 42: I think the `⟹` is over-indented on all the strengthen lemmas if I'm reading the parens correctly.

lib/Monads/trace/Trace_No_Throw.thy
```diff
@@ -0,0 +1,30 @@
...
+ *)
+
+(* Lemmas about no_throw. Usually should have a conclusion "no_throw P m".
+   Includes some monad equations that have no_throw as a main assumption.  *)
+
+theory Trace_No_Throw
+imports
+  \<comment> \<open>Trace_WhileLoopRules
+  Trace_MonadEq_Lemmas\<close>
+  Trace_VCG \<comment> \<open>FIXME: this import replaces the above\<close>
```
Xaphiosis, line 15: forgotten FIXME?

lib/Monads/trace/Trace_Monad_Equations.thy
```diff
@@ -0,0 +1,246 @@
...
+definition repeat :: "('s, unit) tmonad \<Rightarrow> ('s, unit) tmonad" where
+  "repeat f = do n \<leftarrow> select UNIV; repeat_n n f od"
+
+definition env_step :: "('s,unit) tmonad" where
+  "env_step =
+  (do
+    s' \<leftarrow> select UNIV;
+    put_trace_elem (Env, s');
+    put s'
+  od)"
```
Xaphiosis, line 109: indent/parens

Xaphiosis (approved):
This is a superb overhaul of the mess we had before, and you were way more thorough than I thought anyone would have the energy for. Many thanks and words of praise!

It's also a huge diff, so it was difficult to review. I didn't see anything moved to somewhere where it shouldn't have been. I had a few things I noticed or had questions about, but even if this goes in as-is it's a huge improvement. Further improvements should be incremental after this PR goes in, especially when we get to finally using the trace monad in more developments.

lib/Monads/trace/Trace_Monad.thy
corlewis: Good catch, I missed these. I'll probably put this in the fourth commit though. I haven't been perfect with it but my intention for the first one was to be restyling updates taken from the nondet monad with changes to trace monad sections being left for later.

lib/Monads/trace/Trace_Monad.thy
corlewis: Maybe? This comment is copied verbatim from the nondet monad though, so I'd prefer not to change it as part of this work.

lib/Monads/trace/Trace_Monad.thy
corlewis: Sort of, or at least it still needs to be updated. I'll either need to stare at this more to work out how non-termination is represented here or get help from someone who understands how these loops are constructed better than me.

lib/Monads/trace/Trace_RG.thy
corlewis: The indent seems right to me, but the `auto ...; fastforce` method is obviously terrible. I'll see if I can rework it into something nicer.

lib/Monads/trace/Trace_Det.thy
corlewis: Thanks for spotting that, I think that's some copy paste mistakes that I made, since the copyrights for the new files aren't consistent. My intention was for all of the new files to inherit the copyrights from the files that I was copying the content from.

lib/Monads/trace/Trace_In_Monad.thy
corlewis: This comment in Nondet Monad gives a hint
```
(* FIXME lib: remove unlessE_whenE + unless_when here and replace with in_unless lemmas *)
```
It looks like someone removed them for the Trace Monad without adding in the new ones. I might have a quick look to see if they're easy to add.

corlewis:
> For commit messages, there are some that refer to `trace` as a thing on its own, e.g. `trace files`, but other commits say `trace monad`. Would it be possible to have `trace monad` in all cases? For example, I find `lib: update for trace refactor` confusing

Good point, I'll try to make that consistent.

lib/Monads/trace/Trace_No_Throw.thy
corlewis: I originally put that FIXME in as a reminder to keep the same import structure if those other theories were added for the trace monad, but at this stage I'd say that that's unlikely, so I'll remove it.

lib/Monads/trace/Trace_RG.thy
Xaphiosis: The `fastforce` is to the left of the `(`, which can't be right. The `dest!:` on the `auto` should align with the `simp`, not the `auto`, and then the `fastforce` goes after the semicolon, so should be on the same level as the `auto`.
That said, I agree with the method itself being terrible, so either approach works.

lib/Monads/trace/Trace_In_Monad.thy
corlewis: Good news, they were easy to add. You'll see them in the next PR in this chain, along with the equivalent for the nondet monad.

corlewis:
The last force push did involve some new changes, but they were just restyling the theory imports to be consistent and hopefully aren't controversial.

lib/Monads/trace/Trace_Monad.thy
```diff
@@ -766,56 +625,53 @@ text \<open>
...
+inductive_set whileLoop_results ::
+  "('r \<Rightarrow> 's \<Rightarrow> bool) \<Rightarrow> ('r \<Rightarrow> ('s, 'r) tmonad) \<Rightarrow> (('r \<times> 's) \<times> ((tmid \<times> 's) list \<times> ('s, 'r) tmres)) set"
+  for C B where
     "\<lbrakk> \<not> C r s \<rbrakk> \<Longrightarrow> ((r, s), ([], Result (r, s))) \<in> whileLoop_results C B"
   | "\<lbrakk> C r s; (ts, Failed) \<in> B r s \<rbrakk> \<Longrightarrow> ((r, s), (ts, Failed)) \<in> whileLoop_results C B"
   | "\<lbrakk> C r s; (ts, Incomplete) \<in> B r s \<rbrakk> \<Longrightarrow> ((r, s), (ts, Incomplete)) \<in> whileLoop_results C B"
   | "\<lbrakk> C r s; (ts, Result (r', s')) \<in> B r s; ((r', s'), (ts',z)) \<in> whileLoop_results C B  \<rbrakk>
        \<Longrightarrow> ((r, s), (ts'@ts,z)) \<in> whileLoop_results C B"
 
+\<comment> \<open>FIXME: there are less lemmas here than in NonDetMonad and I don't understand this well enough
```
lsf37, line 639: fewer :-)

lib/Monads/trace/Trace_Monad.thy
lsf37: We should leave the FIXME in. I don't think we have properly thought about how this should be defined yet. In particular, in the concurrent setting it does make sense to have a semantics for non-terminating loops, so the whole definition here might have to be rethought (or maybe a potentially non-terminating loop should get a separate definition).

lib/Monads/trace/Trace_RG.thy
lsf37: Also happy to leave this one for later, yes.

lib/Monads/trace/Trace_Monad.thy
lsf37: FIXME would indicate intent to change it. I think we probably don't want to change this, but we should also record that it probably wasn't the optimal decision to go with this in the first place.

lsf37 (approved):
Nice work!

We might have this somewhere already and I missed it: do we have a lemma that shows that the Hoare triples are equivalent to RG-statements with R=G=UNIV? (Or is that not true?) If it is true, that's something we should probably add, which then leads to the question if triples should be an abbreviation. All of that is for another PR, looking through this material just made me think of the connection.

lib/Monads/trace/Trace_Monad.thy
corlewis: To clarify, there actually is a FIXME on the definition in the following line, saying `(* FIXME lib: this should be an abbreviation *)`
See https://github.com/seL4/l4v/blob/master/lib/Monads/nondet/Nondet_VCG.thy#L66

PR #665, opened 2023-08-14
Minor Nondet Monad improvements

corlewis:
These are small changes that were made to the Trace Monad in https://github.com/seL4/l4v/pull/664 and copied across to the Nondet Monad here to keep them in sync.

PR #666, opened 2023-08-23
More Nondet Monad improvements

corlewis:
More improvements to Nondet Monad, following on from https://github.com/seL4/l4v/pull/665. In general this improves style and some proofs, removes duplicate lemmas, and stays in sync with coming changes to Trace Monad.

### DIFF
--- a/lib/Monads/nondet/Nondet_VCG.thy
+++ b/lib/Monads/nondet/Nondet_VCG.thy
@@ -6,16 +6,17 @@
  *)
 
 theory Nondet_VCG
-imports
-  Nondet_Lemmas
-  WPSimp
+  imports
+    Nondet_Lemmas
+    WPSimp
 begin
 
 section \<open>Hoare Logic\<close>
 
 subsection \<open>Validity\<close>
 
-text \<open>This section defines a Hoare logic for partial correctness for
+text \<open>
+  This section defines a Hoare logic for partial correctness for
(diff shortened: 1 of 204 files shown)


lib/Monads/nondet/Nondet_VCG.thy
```diff
@@ -6,16 +6,17 @@
  *)
 
 theory Nondet_VCG
-imports
-  Nondet_Lemmas
-  WPSimp
+  imports
```
Xaphiosis, line 9: when did we start indenting imports? I know the autoindenter likes to do it, but I'm not sure that that's our policy. @lsf37 ?

Xaphiosis (approved):
Other than the `import` indentation question, thumbs up from me.

lib/Monads/nondet/Nondet_VCG.thy
corlewis: I noticed that most of the files in `nondet` already had indented imports so I decided to go with the autoindenter and make them consistent. I don't know if we have an official policy, but I'd also be happy to go in the opposite direction and make them consistent with the rest of the repository.

lsf37: We probably shouldn't fight the auto indenter on this. We used to not indent them, but I'd be happy with indented import statements (and I think I have done that a few times myself now)

PR #667, opened 2023-08-24
Progress on completing the trace monad rule set 

corlewis:
This fills out the rule set by copying in as many definitions and rules as possible from the nondet monad.

Note, for review purposes the first commit has been separated into two. The first commit currently seen here copies in verbatim from nondet monad and can be ignored. The second commit updates as required for the trace monad and the third is completely new content. Both should be reviewed as normal, although there is quite a bit of crossover between the second commit here and the first commit in https://github.com/seL4/l4v/pull/666.

### DIFF
--- a/lib/Monads/Fun_Pred_Syntax.thy
+++ b/lib/Monads/Fun_Pred_Syntax.thy
@@ -173,6 +173,17 @@ lemmas pred_neg_bot_eq[simp] =
    entirely in the future *)
 
 
+subsection "Simplification Rules for Lifted And/Or"
+
+lemma bipred_disj_op_eq[simp]:
+  "reflp R \<Longrightarrow> ((=) or R) = R"
+  "reflp R \<Longrightarrow> (R or (=)) = R"
+  by (auto simp: reflp_def)
+
+lemma bipred_le_true[simp]: "R \<le> \<top>\<top>"
+  by clarsimp
+
+
 section \<open>Examples\<close>
 
 experiment
--- a/lib/Monads/trace/Trace_Empty_Fail.thy
+++ b/lib/Monads/trace/Trace_Empty_Fail.thy
@@ -1,0 +1,343 @@
+(*
+ * Copyright 2023, Proofcraft Pty Ltd
+ * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ *)
+
+theory Trace_Empty_Fail
+  imports
+    Trace_Monad
+    WPSimp
+begin
+
+section \<open>Monads that are wellformed w.r.t. failure\<close>
+
+text \<open>
+  Usually, well-formed monads constructed from the primitives in Nondet_Monad will have the following
+  property: if they return an empty set of results, they will have the failure flag set.\<close>
+definition empty_fail :: "('s,'a) nondet_monad \<Rightarrow> bool" where
+  "empty_fail m \<equiv> \<forall>s. fst (m s) = {} \<longrightarrow> snd (m s)"
+
+text \<open>Useful in forcing otherwise unknown executions to have the @{const empty_fail} property.\<close>
+definition mk_ef :: "'a set \<times> bool \<Rightarrow> 'a set \<times> bool" where
+  "mk_ef S \<equiv> (fst S, fst S = {} \<or> snd S)"
+
+
+subsection \<open>WPC setup\<close>
+
+lemma wpc_helper_empty_fail_final:
+  "empty_fail f \<Longrightarrow> wpc_helper (P, P', P'') (Q, Q', Q'') (empty_fail f)"
+  by (clarsimp simp: wpc_helper_def)
+
+wpc_setup "\<lambda>m. empty_fail m" wpc_helper_empty_fail_final
+
+
+subsection \<open>@{const empty_fail} intro/dest rules\<close>
+
+lemma empty_failI:
+  "(\<And>s. fst (m s) = {} \<Longrightarrow> snd (m s)) \<Longrightarrow> empty_fail m"
+  by (simp add: empty_fail_def)
+
+lemma empty_failD:
+  "\<lbrakk> empty_fail m; fst (m s) = {} \<rbrakk> \<Longrightarrow> snd (m s)"
+  by (simp add: empty_fail_def)
+
+lemma empty_fail_not_snd:
+  "\<lbrakk> \<not> snd (m s); empty_fail m \<rbrakk> \<Longrightarrow> \<exists>v. v \<in> fst (m s)"
+  by (fastforce simp: empty_fail_def)
+
+lemmas empty_failD2 = empty_fail_not_snd[rotated]
+
+lemma empty_failD3:
+  "\<lbrakk> empty_fail f; \<not> snd (f s) \<rbrakk> \<Longrightarrow> fst (f s) \<noteq> {}"
+  by (drule(1) empty_failD2, clarsimp)
+
+lemma empty_fail_bindD1:
+  "empty_fail (a >>= b) \<Longrightarrow> empty_fail a"
+  unfolding empty_fail_def bind_def
+  by (fastforce simp: split_def image_image)
+
+
+subsection \<open>Wellformed monads\<close>
+
+(*
+  Collect generic empty_fail lemmas here:
+   - naming convention is emtpy_fail_NAME.
+   - add lemmas with assumptions to [empty_fail_cond] set
+   - add lemmas without assumption to [empty_fail_term] set
+*)
+
+named_theorems empty_fail_term
+named_theorems empty_fail_cond
+
+lemma empty_fail_K_bind[empty_fail_cond]:
+  "empty_fail f \<Longrightarrow> empty_fail (K_bind f x)"
+  by simp
+
+lemma empty_fail_fun_app[empty_fail_cond]:
+  "empty_fail (f x) \<Longrightarrow> empty_fail (f $ x)"
+  by simp
+
+(* empty_fail as such does not need context, but empty_fail_select_f does, so we need to build
+   up context in other rules *)
+lemma empty_fail_If[empty_fail_cond]:
+  "\<lbrakk> P \<Longrightarrow> empty_fail f; \<not>P \<Longrightarrow> empty_fail g \<rbrakk> \<Longrightarrow> empty_fail (if P then f else g)"
+  by (simp split: if_split)
+
+lemma empty_fail_If_applied[empty_fail_cond]:
+  "\<lbrakk> P \<Longrightarrow> empty_fail (f x); \<not>P \<Longrightarrow> empty_fail (g x) \<rbrakk> \<Longrightarrow> empty_fail ((if P then f else g) x)"
+  by simp
+
+lemma empty_fail_put[empty_fail_term]:
+  "empty_fail (put f)"
+  by (simp add: put_def empty_fail_def)
+
+lemma empty_fail_modify[empty_fail_term]:
+  "empty_fail (modify f)"
+  by (simp add: empty_fail_def simpler_modify_def)
+
+lemma empty_fail_gets[empty_fail_term]:
+  "empty_fail (gets f)"
+  by (simp add: empty_fail_def simpler_gets_def)
+
+lemma empty_fail_select[empty_fail_cond]:
+  "S \<noteq> {} \<Longrightarrow> empty_fail (select S)"
+  by (simp add: empty_fail_def select_def)
+
+lemma empty_fail_select_f[empty_fail_cond]:
+  assumes ef: "fst S = {} \<Longrightarrow> snd S"
+  shows "empty_fail (select_f S)"
+  by (fastforce simp add: empty_fail_def select_f_def intro: ef)
+
+lemma empty_fail_bind[empty_fail_cond]:
+  "\<lbrakk> empty_fail a; \<And>x. empty_fail (b x) \<rbrakk> \<Longrightarrow> empty_fail (a >>= b)"
+  by (fastforce simp: bind_def empty_fail_def split_def)
+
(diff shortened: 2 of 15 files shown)


lib/Monads/Fun_Pred_Syntax.thy
```diff
@@ -173,6 +173,18 @@ lemmas pred_neg_bot_eq[simp] =
    entirely in the future *)
 
 
+subsection "Simplification Rules for Lifted And/Or"
+
+lemma bipred_disj_op_eq[simp]:
+  "reflp R \<Longrightarrow> ((=) or R) = R"
```
lsf37, line 179: Nice. If we add this one, we should probably add the commuted version as well to the simp set (`R or (=)`), so that it rewrites reliably.

(edit: just saw that this lemma is only moved; would still be nice to add the other one for completeness)

lib/Monads/trace/Trace_Monad.thy
```diff
@@ -46,6 +46,24 @@ text \<open>
...
 type_synonym ('s, 'a) tmonad = "'s \<Rightarrow> ((tmid \<times> 's) list \<times> ('s, 'a) tmres) set"
 
+text \<open>
+  Print the type @{typ "('s,'a) tmonad"} instead of its unwieldy expansion.
+  Needs an AST translation in code, because it needs to check that the state variable
+  @{typ 's} occurs three times. This comparison is not guaranteed to always work as expected
+  (AST instances might have different decoration), but it does seem to work here.\<close>
+print_ast_translation \<open>
+  let
+    fun tmonad_tr _ [t1, Ast.Appl [Ast.Constant @{type_syntax set},
```
lsf37, line 56: 👍

lsf37 (approved):
Nice!

lsf37:
Should probably start making issues for this, but I think the next step would be to start on a Trace_RG_VCG theory. We could try to lift the existing `hoare_vcg_*` theorems via equivalence, but we'd only get them for R=G=True. With a bit of luck most of them will hold for arbitrary R and G, which would be nice to have.

We probably also want `no_trace` and `prefix_closed` proved of all monad primitives we have, so that these can be discharged completely automatically when they occur. 

Looking at how equal all those existing proofs are to the nondet-monad it really is a bit of a shame that we can't do monads generically. I'm almost tempted to try some locales with way too general types that have large instantiations. One wouldn't get the definitions from the locales because of the too general types, but one could potentially get a lot of lemmas. They'd probably have large instantiations (all the definitions). Anyway. That one is lower priority -- the RG rules are not going to follow from the locales and those are the ones we really want to have in the end.

corlewis:
This PR now also copies `empty_fail` and its related proofs in from the nondet monad. While most of this was relatively easy to update for the trace monad, lemmas involving `bind` were a bit complicated and `mk_ef` was commented out, due to it requiring more thought about whether it makes sense to invent a (presumably empty) trace.

lsf37 (approved):
Thanks for adding those, that makes the setup more complete.

lib/Monads/trace/Trace_Empty_Fail.thy
```diff
@@ -0,0 +1,372 @@
...
+text \<open>
+  Usually, well-formed monads constructed from the primitives in Trace_Monad will have the following
+  property: if they return an empty set of completed results, there exists a trace corresponding to
+  a failed result.\<close>
+definition empty_fail :: "('s,'a) tmonad \<Rightarrow> bool" where
+  "empty_fail m \<equiv> \<forall>s. mres (m s) = {} \<longrightarrow> Failed \<in> snd ` (m s)"
+
+text \<open>Useful in forcing otherwise unknown executions to have the @{const empty_fail} property.\<close>
+\<comment> \<open>definition mk_ef :: "'a set \<times> bool \<Rightarrow> 'a set \<times> bool" where
+  "mk_ef S \<equiv> (fst S, fst S = {} \<or> snd S)"\<close>
```
Xaphiosis, line 25: this appears commented out?

lib/Monads/trace/Trace_Empty_Fail.thy
```diff
@@ -0,0 +1,372 @@
...
+  by (force simp: split_def mres_def vimage_def split: tmres.splits)
+
+
+subsection \<open>Wellformed monads\<close>
+
+(*
+  Collect generic empty_fail lemmas here:
+   - naming convention is emtpy_fail_NAME.
+   - add lemmas with assumptions to [empty_fail_cond] set
+   - add lemmas without assumption to [empty_fail_term] set
```
Xaphiosis, line 71: bit off-topic and beyond the scope of this PR, but would it be possible to write attributes that can check you're not putting the wrong thing into one of these sets, or automatically select which set to put it into?

lib/Monads/trace/Trace_Empty_Fail.thy
```diff
@@ -0,0 +1,372 @@
...
+
+lemma bind_FailedI2:
+  "\<lbrakk>\<forall>res\<in>mres (f s). Failed \<in> snd ` (g (fst res) (snd res)); mres (f s) \<noteq> {}\<rbrakk>
+   \<Longrightarrow> Failed \<in> snd ` (f >>= g) s"
+  by (force simp: bind_def mres_def image_def split_def)
+
+\<comment> \<open>FIXME: move\<close>
+lemma context_disjE:
+  "\<lbrakk>P \<or> Q; P \<Longrightarrow> R; \<lbrakk>\<not>P; Q\<rbrakk> \<Longrightarrow> R\<rbrakk> \<Longrightarrow> R"
+  by auto
```
Xaphiosis, line 134: move now, or move later?

lib/Monads/trace/Trace_Empty_Fail.thy
```diff
@@ -0,0 +1,372 @@
...
+  "empty_fail (assert P)"
+  unfolding assert_def by (simp add: empty_fail_term)
+
+lemma empty_fail_assert_opt[empty_fail_term]:
+  "empty_fail (assert_opt x)"
+  by (simp add: assert_opt_def empty_fail_term split: option.splits)
+
+\<comment> \<open>lemma empty_fail_mk_ef[empty_fail_term]:
+  "empty_fail (mk_ef o m)"
+  by (simp add: empty_fail_def mk_ef_def)\<close>
```
Xaphiosis, line 200: any reason this one's commented out without a FIXME or indication of what to do?

lib/Monads/trace/Trace_Monad.thy
```diff
@@ -678,6 +721,49 @@ definition whileLoopE ::
...
+definition notM :: "('s, bool) tmonad \<Rightarrow> ('s, bool) tmonad" where
+  "notM m = do c \<leftarrow> m; return (\<not> c) od"
+
+definition
+  whileM :: "('s, bool) tmonad \<Rightarrow> ('s, 'a) tmonad \<Rightarrow> ('s, unit) tmonad" where
+  "whileM C B \<equiv> do
+    c \<leftarrow> C;
+    whileLoop (\<lambda>c s. c) (\<lambda>_. do B; C od) c;
+    return ()
+  od"
```
Xaphiosis, line 736: not sure about whether this is the PR for it, but these are not the new definition style, the wrapping is (as I was reminded recently):
```
definition name ::
  very_long_type
  where
  ...
```

lib/Monads/trace/Trace_Monad_Equations.thy
```diff
@@ -10,41 +10,364 @@
...
+lemma modify_assert:
+  "(do v2 \<leftarrow> modify f; v1 \<leftarrow> assert v; c v1 od)
+    = (do v1 \<leftarrow> assert v; v2 \<leftarrow> modify f; c v1 od)"
+  by (simp add: simpler_modify_def return_def assert_def fail_def bind_def
+         split: if_split)
+
+lemma gets_fold_into_modify:
+  "do x \<leftarrow> gets f; modify (g x) od = modify (\<lambda>s. g (f s) s)"
+  "do x \<leftarrow> gets f; _ \<leftarrow> modify (g x); h od
+     = do modify (\<lambda>s. g (f s) s); h od"
```
Xaphiosis, line 286: there's a bunch of wonky indentation in this file, but I'm guessing these are moved/copied?

lib/Monads/trace/Trace_No_Throw.thy
```diff
@@ -25,4 +25,76 @@ definition no_throw :: "('s \<Rightarrow> bool) \<Rightarrow> ('s, 'e + 'a) tmon
 definition no_return :: "('a \<Rightarrow> bool) \<Rightarrow> ('a, 'b + 'c) tmonad \<Rightarrow> bool" where
   "no_return P A \<equiv> \<lbrace> P \<rbrace> A \<lbrace>\<lambda>_ _. False\<rbrace>,\<lbrace>\<lambda>_ _. True \<rbrace>"
 
+(* Alternative definition of no_throw; easier to work with than unfolding validE. *)
+lemma no_throw_def':
+  "no_throw P A = (\<forall>s. P s \<longrightarrow> (\<forall>(r, t) \<in> mres (A s). (\<exists>x. r = Inr x)))"
+  by (clarsimp simp: no_throw_def validE_def2 split_def split: sum.splits)
+
```
Xaphiosis, line 32: extra newline?

lib/Monads/trace/Trace_More_VCG.thy
```diff
@@ -11,10 +11,770 @@
...
+
+lemma valid_rv_split:
+  "\<lbrakk>\<lbrace>P\<rbrace> f \<lbrace>\<lambda>rv s. rv \<longrightarrow> Q s\<rbrace>; \<lbrace>P\<rbrace> f \<lbrace>\<lambda>rv s. \<not>rv \<longrightarrow> Q' s\<rbrace>\<rbrakk>
+  \<Longrightarrow>
+  \<lbrace>P\<rbrace> f \<lbrace>\<lambda>rv s. if rv then Q s else Q' s\<rbrace>"
+  by (fastforce simp: valid_def)
+
+lemma hoare_rv_split:
+  "\<lbrakk>\<lbrace>P\<rbrace> f \<lbrace>\<lambda>rv s. rv \<longrightarrow> (Q rv s)\<rbrace>; \<lbrace>P\<rbrace> f \<lbrace>\<lambda>rv s. (\<not>rv) \<longrightarrow> (Q rv s)\<rbrace>\<rbrakk>
+  \<Longrightarrow> \<lbrace>P\<rbrace> f \<lbrace>Q\<rbrace>"
```
Xaphiosis, line 716: this file also features wonky indents, assuming copy/move

lib/Monads/trace/Trace_Empty_Fail.thy
corlewis: Yes, I had commented it out due to it requiring more thought about how to define `mk_ef` for the trace monad. Thanks for getting me to look at it again though, I came up with a simple idea which while untested at least works for the basic `empty_fail_mk_ef` lemma.

lib/Monads/trace/Trace_Empty_Fail.thy
corlewis: Ooh, interesting idea. That does feel possible, although I don't have experience with these lemma sets to know how they are used in practice and what the end result would be for us as users.

lib/Monads/trace/Trace_Empty_Fail.thy
corlewis: Happy to move now but not sure where to. Any suggestions?

lib/Monads/trace/Trace_Empty_Fail.thy
corlewis: It is uncommented out now that I have given `mk_ef` a new definition for the trace monad.

lib/Monads/trace/Trace_Monad_Equations.thy
corlewis: Yeah, they were copied from nondet. Now that you've pointed them out I've taken the opportunity to fix the indentation up, both here and in the original nondet files. It's almost definitely still not perfect but should at least be an improvement.

lib/Monads/trace/Trace_No_Throw.thy
corlewis: This was also copied from nondet, and I think was signifying a subsection. I've made that explicit now.

lib/Monads/trace/Trace_Empty_Fail.thy
lsf37: It's certainly possible to do, but the attribute(s) would be manually written ML code like `[simp]` and it would still be hard to prevent the user from just adding theorems to the set directly (without making it much harder to use, at least).

In terms of maintainability trade-off, the simple named-thms sets probably win out, because we don't usually dynamically add new stuff to these sets.

lib/Monads/trace/Trace_Empty_Fail.thy
Xaphiosis: Is `lib/Lib.thy` in the right hierarchy for this?

lib/Monads/trace/Trace_Empty_Fail.thy
Xaphiosis: OK, if it's not going to get dynamically added to the wrong set by some future genius, then it's not worth it. Only wanted to point out that it's possible to programmatically check whether something has assumptions or not before the set add.

Xaphiosis (approved):
Nice work!

lib/Monads/trace/Trace_Empty_Fail.thy
corlewis: Unfortunately not, the session dependencies are in the wrong direction for that.
`Monads` currently depends only on `HOL`, `Eisbach_Tools`, and `ML_Utils`, which without further restructuring doesn't leave many places to move this.

corlewis: I'll move it to `Monads/Monad_Lib` for now, which at least makes it accessible to the other monads.

PR #669, opened 2023-09-14
crefine: change misleading proof step in CSpace_RAB_C

Xaphiosis:
Trying to figure this out was very educational, since ccorres_abstract was used without intending to abstract a variable, the xf' and lambda name were both red herrings (in fact, this proof only worked if xf' was instantiated with an *irrelevant* C local var name), and the body was not transformed.


lsf37 (approved):
Thanks for that.

PR #670, opened 2023-09-27
update to Isabelle 2023

lsf37:
Updates all proofs and ML code to Isabelle2023.

Changes are fairly minor this time. A few tweaks in ML where functions get more parameters (location info and similar). Mostly these are filled with `None` or default values unless something is specifically available to put in.

On the proof side, the only larger change is that the syntax precedence of the "mapsto" arrow has changed from
`f t (x |-> y)` to `(f t) (x |-> y)`, which is arguably clearer and easier to read in most cases.

Haven't noticed any major performance impact, but we'll see more comparable numbers when the AWS tests come in.

### DIFF
--- a/lib/Word_Lib/Word_Lemmas.thy
+++ b/lib/Word_Lib/Word_Lemmas.thy
@@ -496,7 +496,7 @@ next
       ultimately show ?thesis
         using \<open>y < LENGTH('a)\<close>
         by (auto simp add: drop_bit_eq_div word_less_nat_alt unat_div unat_word_ariths
-          shiftr_def shiftl_def)
+                           shiftr_def shiftl_def)
     next
       case False
       with \<open>y < n\<close> have *: \<open>unat x \<noteq> 2 ^ n div 2 ^ y\<close>
@@ -508,8 +508,9 @@ next
       also have \<open>\<dots> \<longleftrightarrow> unat x < 2 ^ n div 2 ^ y\<close>
         using * by (simp add: less_le)
       finally show ?thesis
-      using that \<open>x \<noteq> 0\<close> by (simp flip: push_bit_eq_mult drop_bit_eq_div
-        add: shiftr_def shiftl_def unat_drop_bit_eq word_less_iff_unsigned [where ?'a = nat])
+        using that \<open>x \<noteq> 0\<close>
+        by (simp flip: push_bit_eq_mult drop_bit_eq_div
+                 add: shiftr_def shiftl_def unat_drop_bit_eq word_less_iff_unsigned [where ?'a = nat])
     qed
   qed
 qed
@@ -716,7 +717,8 @@ lemma word_and_notzeroD:
 lemma shiftr_le_0:
   "unat (w::'a::len word) < 2 ^ n \<Longrightarrow> w >> n = (0::'a::len word)"
   by (auto simp add: take_bit_word_eq_self_iff word_less_nat_alt shiftr_def
-    simp flip: take_bit_eq_self_iff_drop_bit_eq_0 intro: ccontr)
+           simp flip: take_bit_eq_self_iff_drop_bit_eq_0
+           intro: ccontr)
 
 lemma of_nat_shiftl:
   "(of_nat x << n) = (of_nat (x * 2 ^ n) :: ('a::len) word)"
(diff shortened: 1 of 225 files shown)


lsf37:
(Setting this to draft for now, because we should merge the other larger PRs first and then rebase this one)

lsf37:
This is now ready for review. The last commit needs to be dropped before merging (it just makes the tests pick Isabelle2023 instead of Isabelle2022 for the PR).

corlewis (approved):
Looks good to me!
My only comment is that a lot of the commits could be squashed together, since I don't think we get any benefit from the mapsto syntax updates being per directory.

lib/Word_Lib/Word_Lemmas.thy
```diff
@@ -496,7 +496,7 @@ next
       ultimately show ?thesis
         using \<open>y < LENGTH('a)\<close>
         by (auto simp add: drop_bit_eq_div word_less_nat_alt unat_div unat_word_ariths
-                           shiftr_def shiftl_def)
+          shiftr_def shiftl_def)
```
michaelmcinerney, line 499: Indentation

lib/Word_Lib/Word_Lemmas.thy
```diff
@@ -716,7 +716,7 @@ lemma word_and_notzeroD:
 lemma shiftr_le_0:
   "unat (w::'a::len word) < 2 ^ n \<Longrightarrow> w >> n = (0::'a::len word)"
   by (auto simp add: take_bit_word_eq_self_iff word_less_nat_alt shiftr_def
-           simp flip: take_bit_eq_self_iff_drop_bit_eq_0 intro: ccontr)
+    simp flip: take_bit_eq_self_iff_drop_bit_eq_0 intro: ccontr)
```
michaelmcinerney, line 719: Indentation

lib/Word_Lib/Word_Lemmas.thy
```diff
@@ -1466,8 +1466,11 @@ lemma mask_shift_sum:
   "\<lbrakk> a \<ge> b; unat n = unat (p AND mask b) \<rbrakk>
    \<Longrightarrow> (p AND NOT(mask a)) + (p AND mask a >> b) * (1 << b) + n = (p :: 'a :: len word)"
   apply (simp add: shiftl_def shiftr_def flip: push_bit_eq_mult take_bit_eq_mask word_unat_eq_iff)
-  apply (subst disjunctive_add, clarsimp simp add: bit_simps)+
-  apply (rule bit_word_eqI)
+  apply (subst disjunctive_add)
+   apply (auto simp add: bit_simps)
```
michaelmcinerney, line 1470: Limit the auto to the current subgoal? And below

proof/access-control/Ipc_AC.thy
```diff
@@ -178,8 +178,8 @@ lemma send_upd_ctxintegrity:
      integrity aag X st s; st_tcb_at ((=) Running) thread s;
      get_tcb thread st = Some tcb; get_tcb thread s = Some tcb'\<rbrakk>
      \<Longrightarrow> integrity aag X st
-                   (s\<lparr>kheap := kheap s(thread \<mapsto>
-                                       TCB (tcb'\<lparr>tcb_arch := arch_tcb_context_set c' (tcb_arch tcb')\<rparr>))\<rparr>)"
+                   (s\<lparr>kheap := (kheap s)
+                               (thread \<mapsto> TCB (tcb'\<lparr>tcb_arch := arch_tcb_context_set c' (tcb_arch tcb')\<rparr>))\<rparr>)"
```
michaelmcinerney, line 182: I think I'd prefer the previous indentation

lsf37:
> Looks good to me!
> My only comment is that a lot of the commits could be squashed together, since I don't think we get any benefit from the mapsto syntax updates being per directory.

That makes sense, the current set is mostly how they were constructed. I'll squash them into larger sets.

michaelmcinerney (approved):
Looks good, just a few minor nitpicks

lib/Word_Lib/Word_Lemmas.thy
lsf37: Hm, these are noise from the AFP. I guess we should fix them, otherwise we'll accumulate them.

lib/Word_Lib/Word_Lemmas.thy
lsf37: I have the felling that'll break the proof. So great to have stuff in the AFP..

lib/Word_Lib/Word_Lemmas.thy
lsf37: Done. Thanks for spotting these.

lib/Word_Lib/Word_Lemmas.thy
lsf37: Ok, managed to fix that one and a few more of them.

PR #671, opened 2023-09-29
Remove uses of _tac methods from nondet

corlewis:
In particular, try to remove as many _tac methods that refer to system-generated variable names as possible. Any remaining uses are explicitly protected by a rename_tac.

### DIFF
--- a/lib/Monads/nondet/Nondet_While_Loop_Rules.thy
+++ b/lib/Monads/nondet/Nondet_While_Loop_Rules.thy
@@ -286,22 +286,35 @@ lemma fst_whileLoop_cond_false:
   using loop_result
   by (rule in_whileLoop_induct, auto)
 
+lemma whileLoop_terminates_results:
+  assumes non_term: "\<And>r. \<lbrace> \<lambda>s. I r s \<and> C r s \<and> \<not> snd (B r s) \<rbrace> B r \<exists>\<lbrace> \<lambda>r' s'. C r' s' \<and> I r' s' \<rbrace>"
+  shows
+    "\<lbrakk>whileLoop_terminates C B r s; (Some (r, s), None) \<notin> whileLoop_results C B; I r s; C r s\<rbrakk>
+     \<Longrightarrow> False"
+proof (induct rule: whileLoop_terminates.induct)
+  case (1 r s)
+  then show ?case
+    apply clarsimp
+    done
+next
+  case (2 r s)
+  then show ?case
+    apply (cut_tac non_term[where r=r])
+    apply (clarsimp simp: exs_valid_def)
+    apply (subst (asm) (2) whileLoop_results.simps)
+    apply clarsimp
+    apply (insert whileLoop_results.simps)
+    apply fast
+    done
+qed
+
 lemma snd_whileLoop:
   assumes init_I: "I r s"
-      and cond_I: "C r s"
-      and non_term: "\<And>r. \<lbrace> \<lambda>s. I r s \<and> C r s \<and> \<not> snd (B r s) \<rbrace> B r \<exists>\<lbrace> \<lambda>r' s'. C r' s' \<and> I r' s' \<rbrace>"
+    and cond_I: "C r s"
+    and non_term: "\<And>r. \<lbrace> \<lambda>s. I r s \<and> C r s \<and> \<not> snd (B r s) \<rbrace> B r \<exists>\<lbrace> \<lambda>r' s'. C r' s' \<and> I r' s' \<rbrace>"
   shows "snd (whileLoop C B r s)"
   apply (clarsimp simp: whileLoop_def)
-  apply (rotate_tac)
-  apply (insert init_I cond_I)
-  apply (induct rule: whileLoop_terminates.induct)
-   apply clarsimp
-  apply (cut_tac r=r in non_term)
-  apply (clarsimp simp: exs_valid_def)
-   apply (subst (asm) (2) whileLoop_results.simps)
-   apply clarsimp
-  apply (insert whileLoop_results.simps)
-  apply fast
+  apply (erule (1) whileLoop_terminates_results[OF non_term _ _ init_I cond_I])
   done
 
 lemma whileLoop_terminates_inv:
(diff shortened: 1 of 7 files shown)


lib/Monads/nondet/Nondet_While_Loop_Rules.thy
```diff
@@ -562,37 +583,45 @@ lemma snd_X_return[simp]:
...
   apply (rule prod_eqI)
    apply (rule set_eqI, rule iffI)
     apply clarsimp
-    apply (clarsimp simp: in_bind whileLoop_def liftE_def)
-    apply (rule_tac x="b" in exI)
-    apply (rule_tac x="projr a" in exI)
+    apply (clarsimp simp: in_liftE whileLoop_def)
+    \<comment> \<open>FIXME: existential is instantiated by 'subst isr_Inr_proj' ... 'rule refl' in two lines
+    apply (rule_tac x="projr a" in exI)\<close>
+    apply (rule exI)
```
corlewis, line 600: Are we ok with leaving a schematic from `exI` like this, to be instantiated a few steps later? This was the nicest way I could find to do it that didn't require referring to a bound variable.

lsf37: Yes, I think that is fine and should be more stable than mentioning the name. The price is that it's probably a bit harder to understand what is going on, but I think for a single exI that's Ok.

lib/Monads/nondet/Nondet_While_Loop_Rules.thy
```diff
@@ -498,32 +508,43 @@ lemma empty_fail_whileM[empty_fail_cond, intro!, wp]:
...
                          \<Longrightarrow> (Some (rt r, st s), None) \<in> whileLoop_results C' B'"
-  and refine: "\<And>r s r' s'. \<lbrakk> I r s; C r s; (r', s') \<in>  fst (B r s) \<rbrakk>
+    and refine: "\<And>r s r' s'. \<lbrakk> I r s; C r s; (r', s') \<in>  fst (B r s) \<rbrakk>
                             \<Longrightarrow> (rt r', st s') \<in> fst (B' (rt r) (st s))"
   shows "(Q, R) \<in> whileLoop_results C' B'"
   apply (subst vars1)
   apply (subst vars2)
   apply (insert base inv_init)
-  apply (induct rule: whileLoop_results.induct)
+proof (induct rule: whileLoop_results.induct)
```
lsf37, line 524: Same as above, the AFP (and proof style in general) will frown on `apply` followed by `proof` (we have a few instances of that in other proofs, they are not good examples to follow :-)).

This one is a bit more complex. My first instinct would be to try something with `defines`, but I can't quite see in the diff if that makes sense and which names `Q` and `R` mention on the right-hand side.

lib/Monads/nondet/Nondet_While_Loop_Rules_Completeness.thy
```diff
@@ -66,7 +68,8 @@ proof (rule iffI)
     by auto
 
   thus ?RHS
-    by (rule_tac validNF_whileLoop [where I=I and R=R], auto)
+    apply -
+    by (rule validNF_whileLoop[where I=I and R=R], auto)
```
lsf37, line 72:
```suggestion
    by - (rule validNF_whileLoop[where I=I and R=R], auto)
```

lib/Monads/nondet/Nondet_While_Loop_Rules_Completeness.thy
```diff
@@ -260,15 +265,23 @@ proof -
...
-      apply (induct rule: in_whileLoop_induct)
-       apply (rule_tac x="[(r,s)]" in exI)
-       apply clarsimp
-      apply clarsimp
-      apply (case_tac l)
-       apply clarsimp
-      apply (rule_tac x="(r, s)#l" in exI)
-      apply clarsimp
-      done
+    proof (induct rule: in_whileLoop_induct)
```
lsf37, line 268: when followed by `proof` I would usually indent `using` the same as `proof`. Maybe that's not very consistent. What do others think? @Xaphiosis?

lsf37 (approved):
Nice, these are pretty much all improvements. I don't have a very good solution for the second `apply` followed by `proof` bit. That one might need a bit more experimentation.

lib/Monads/nondet/Nondet_While_Loop_Rules.thy
```diff
@@ -286,21 +286,28 @@ lemma fst_whileLoop_cond_false:
...
-  apply (induct rule: whileLoop_terminates.induct)
-   apply clarsimp
-  apply (cut_tac r=r in non_term)
-  apply (clarsimp simp: exs_valid_def)
-   apply (subst (asm) (2) whileLoop_results.simps)
-   apply clarsimp
-  apply (insert whileLoop_results.simps)
-  apply fast
-  done
+proof (induct rule: whileLoop_terminates.induct)
```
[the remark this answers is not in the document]
corlewis: The `clarsimp` is needed to get to the term actually being inducted on, so I've gone with a separate lemma.

lib/Monads/nondet/Nondet_While_Loop_Rules_Completeness.thy
corlewis: I have no strong opinions, but the auto-indenter likes it this way. For clarity without the diff formatting getting in the way, this is
```
proof -
  {
    fix r' s'
    assume x: "(r', s') ∈ fst (whileLoop C B r s)"
    assume y: "Q r' s'"
    have ?thesis
      using x y
    proof (induct rule: in_whileLoop_induct)
      case (1 r s)
        apply -
 ```

lib/Monads/nondet/Nondet_While_Loop_Rules.thy
Xaphiosis: I'm pretty much done for the week so will have to continue this PR review next week, but: I very much object to `proof` in the middle of `apply` scripts. We have some historical leftovers (mainly from SimonW), but we should not introduce more of that.

corlewis: I've gone with making them `defines [simp]:`. Seeing how the previous proof unfolded them straight away, I'm assuming that they were introduced mainly to improve readability.
I did also have to parameterise them to get the induction to apply properly.

corlewis: Unfortunately, that didn't actually work, I've realised that the intention for how it was previously written was to make the rule easy to apply where it's used.
I've now split it into two lemmas; one helper that uses `defines` and does the induction, and a second that has the same lemma statement as before my changes.

lib/Monads/nondet/Nondet_While_Loop_Rules_Completeness.thy
lsf37: For now, let's not fight the auto indenter on this one. Some of my habits are from before-indenter-times and probably not worth preserving.

lsf37 (approved):

Xaphiosis:
I'm fine with this, no further review comments. Not sure what the motivation for it was, AFP requirements?

lsf37:
> I'm fine with this, no further review comments. Not sure what the motivation for it was, AFP requirements?

Yes, AFP requirements

PR #672, opened 2023-10-05
Monads: improve style and consistency of nondet and trace

corlewis:


lib/Monads/trace/Trace_VCG.thy
```diff
@@ -708,6 +734,58 @@ lemma hoare_post_comb_imp_conj:
...
+  "\<lbrace>R\<rbrace> f \<lbrace>\<lambda>rv s. (P \<longrightarrow> X rv s) \<and> (\<not>P \<longrightarrow> Y rv s)\<rbrace> \<Longrightarrow>
+   \<lbrace>R\<rbrace> f \<lbrace>\<lambda>rv s. if P then X rv s else Y rv s\<rbrace>"
+
+  "\<lbrace>R\<rbrace> f \<lbrace>\<lambda>rv s. (P \<longrightarrow> X rv s) \<and> (\<not>P \<longrightarrow> Y rv s)\<rbrace> \<Longrightarrow>
+   \<lbrace>R\<rbrace> f \<lbrace>\<lambda>rv. if P then X rv else Y rv\<rbrace>"
+  by (auto simp: valid_def split_def)
+
+lemma hoare_vcg_disj_lift_R:
+  assumes x: "\<lbrace>P\<rbrace>  f \<lbrace>Q\<rbrace>,-"
+  assumes y: "\<lbrace>P'\<rbrace> f \<lbrace>Q'\<rbrace>,-"
```
Xaphiosis, line 747: the alignment of `f` is strange, we don't add extra spaces like this anywhere else

lib/Monads/trace/Trace_Monad.thy
```diff
@@ -98,9 +98,8 @@ definition bind ::
     | Result (rv, s) \<Rightarrow> fst_upd (\<lambda>ys. ys @ xs) ` g rv s"
 
 text \<open>Sometimes it is convenient to write @{text bind} in reverse order.\<close>
-abbreviation(input) bind_rev ::
-  "('c \<Rightarrow> ('a, 'b) tmonad) \<Rightarrow> ('a, 'c) tmonad \<Rightarrow> ('a, 'b) tmonad" (infixl "=<<" 60)
-  where
+abbreviation (input) bind_rev ::
+  "('c \<Rightarrow> ('a, 'b) tmonad) \<Rightarrow> ('a, 'c) tmonad \<Rightarrow> ('a, 'b) tmonad" (infixl "=<<" 60) where
```
Xaphiosis, line 102: @lsf will have to weigh in here, but unless I'm mistaken, according to the current style guide I think once you wrap after the `::` you also are committed to `where` on a separate line.

Xaphiosis: (there are multiple occurrences in other files in this PR as well)

Xaphiosis (approved):
Two nitpicks, otherwise I agree this is an improvement.

lsf37 (approved):

lib/Monads/trace/Trace_Monad.thy
lsf37: I don't have a strong personal opinion on that one, but Raf is right, that is what the style guide doc says.

lib/Monads/trace/Trace_VCG.thy
corlewis: Fixed, both her and in `nondet` where I copied it from.

lib/Monads/trace/Trace_Monad.thy
corlewis: Huh, I missed that last time I looked at the style guide and in an attempt to be consistent have added a lot more like this in the last few PRs.
I'll go through `Trace_Monad` and `Nondet_Monad` and fix all of the definitions to follow this.

Xaphiosis: Where are we between that and changing the style guide? For now probably we should stick with it, but if it goes into AFP who knows what they want anyway. Normally I'm all for conserving vertical space, but in this case I'm not sure.

lsf37: The AFP doesn't care about style, that would be way too hard to enforce. I think it's probably better to keep the rule simple, even if it's not always optimal.

lib/Monads/nondet/Nondet_Monad.thy
```diff
@@ -106,7 +106,8 @@ definition select :: "'a set \<Rightarrow> ('s,'a) nondet_monad" where
   "select A \<equiv> \<lambda>s. (A \<times> {s}, False)"
 
 definition alternative ::
-  "('s, 'a) nondet_monad \<Rightarrow> ('s, 'a) nondet_monad \<Rightarrow> ('s, 'a) nondet_monad" (infixl "\<sqinter>" 20) where
+  "('s, 'a) nondet_monad \<Rightarrow> ('s, 'a) nondet_monad \<Rightarrow> ('s, 'a) nondet_monad"
+  (infixl "\<sqinter>" 20) where
```
corlewis, line 110: There isn't an official guideline that I can see for where the infix notation should go but this is what makes sense to me to and seems to be consistent with the other examples I can find.
Any objections, and if not should I add it to the style guide?

lsf37: The infix should usually go with the type (because the type gives you the arity the infix uses). But if it would have to line break then it makes sense to me to break after the type, yes.

corlewis: Right, I'll adjust some of them to do that then, where the line breaks aren't needed.
For completeness, what about this one?
```
inductive whileLoop_terminates ::
  "('r ⇒ 's ⇒ bool) ⇒ ('r ⇒ ('s, 'r) nondet_monad) ⇒ 'r ⇒ 's ⇒ bool"
  for C B where
  ```
In my mind we would also want the line break there, since the `for C B` isn't related to the type.

lsf37: Yes, breaking before `for` makes sense.

Xaphiosis: I agree with these outcomes.

PR #673, opened 2023-10-06
Some rules for CLib

michaelmcinerney:
This updates `ccorres_While` and `ccorres_call_getter_setter` to be more useful, adding and moving a few other things along the way

### DIFF
--- a/lib/clib/CCorresLemmas.thy
+++ b/lib/clib/CCorresLemmas.thy
@@ -5,7 +5,7 @@
  *)
 
 theory CCorresLemmas
-imports CCorres_Rewrite
+imports CCorres_Rewrite MonadicRewrite_C
 begin
 
 lemma ccorres_rel_imp2:
@@ -1176,5 +1176,39 @@ lemma ccorres_inr_rrel_Inr[simp]:
   "ccorres_underlying srel \<Gamma> (inr_rrel r \<circ> Inr) xf ar axf P Q hs m c
    = ccorres_underlying srel \<Gamma> r xf ar axf P Q hs m c"
   by (simp cong: ccorres_context_cong)+
+
+lemma add_remove_return:
+  "getter >>= setter = do (do val \<leftarrow> getter; setter val; return val od); return () od"
+  by (simp add: bind_assoc)
+
+lemma ccorres_call_getter_setter_dc:
+  assumes cul: "ccorresG sr \<Gamma> r' xf' P (i ` P') [] getter (Call f)"
+    and   gsr: "\<And>x x' s t rv.
+                 \<lbrakk> (x, t) \<in> sr; r' rv (xf' t); ((), x') \<in> fst (setter rv x) \<rbrakk>
+                 \<Longrightarrow> (x', g s t (clean s t)) \<in> sr"
+    and   ist: "\<And>x s. (x, s) \<in> sr \<Longrightarrow> (x, i s) \<in> sr"
+    and    ef: "\<And>val. empty_fail (setter val)"
+  shows "ccorresG sr \<Gamma> dc xfdc P P' hs
+           (getter >>= setter)
+           (call i f clean (\<lambda>s t. Basic (g s t)))"
+  apply (rule ccorres_guard_imp)
+    apply (rule monadic_rewrite_ccorres_assemble[rotated])
+     apply (rule monadic_rewrite_is_refl)
+     apply (rule add_remove_return)
+    apply (rule ccorres_seq_skip'[THEN iffD1])
+    apply (rule ccorres_split_nothrow_novcg)
+        apply (rule ccorres_call_getter_setter)
+            apply (fastforce intro: cul)
+           apply (fastforce intro: gsr)
+          apply (simp add: gsr)
+         apply (fastforce intro: ist)
+        apply (fastforce intro: ef)
+       apply (rule ceqv_refl)
+      apply (fastforce intro: ccorres_return_Skip)
+     apply wpsimp
+    apply (clarsimp simp: guard_is_UNIV_def)
+   apply wpsimp
+  apply fastforce
+  done
 
 end
--- a/lib/clib/CCorresLemmas.thy
+++ b/lib/clib/CCorresLemmas.thy
@@ -1040,49 +1040,6 @@ lemma ccorres_Guard_True_Seq:
    \<Longrightarrow> ccorres_underlying sr \<Gamma> r xf arrel axf A C hs a (Guard F \<lbrace>True\<rbrace> c ;; d)"
    by (simp, ccorres_rewrite, assumption)
 
-lemma exec_While_final_inv'':
-  "\<lbrakk> \<Gamma> \<turnstile> \<langle>b, x\<rangle> \<Rightarrow> s'; b = While C B; x = Normal s;
-    \<And>s. s \<notin> C \<Longrightarrow> I s (Normal s);
-    \<And>t t' t''. \<lbrakk> t \<in> C; \<Gamma>\<turnstile> \<langle>B, Normal t\<rangle> \<Rightarrow> Normal t'; \<Gamma>\<turnstile> \<langle>While C B, Normal t'\<rangle> \<Rightarrow> t'';
-                 I t' t'' \<rbrakk> \<Longrightarrow> I t t'';
-    \<And>t t'. \<lbrakk> t \<in> C; \<Gamma>\<turnstile> \<langle>B, Normal t\<rangle> \<Rightarrow> Abrupt t' \<rbrakk> \<Longrightarrow> I t (Abrupt t');
-    \<And>t. \<lbrakk> t \<in> C; \<Gamma> \<turnstile> \<langle>B, Normal t\<rangle> \<Rightarrow> Stuck \<rbrakk> \<Longrightarrow> I t Stuck;
-    \<And>t f. \<lbrakk> t \<in> C; \<Gamma>\<turnstile> \<langle>B, Normal t\<rangle> \<Rightarrow> Fault f \<rbrakk> \<Longrightarrow> I t (Fault f) \<rbrakk>
-   \<Longrightarrow> I s s'"
-  apply (induct arbitrary: s rule: exec.induct; simp)
-  apply (erule exec_elim_cases; fastforce simp: exec.WhileTrue exec.WhileFalse)
-  done
-
-lemma intermediate_Normal_state:
-  "\<lbrakk>\<Gamma> \<turnstile> \<langle>Seq c\<^sub>1 c\<^sub>2, Normal t\<rangle> \<Rightarrow> t''; t \<in> P; \<Gamma> \<turnstile> P c\<^sub>1 Q\<rbrakk>
-   \<Longrightarrow> \<exists>t'. \<Gamma> \<turnstile> \<langle>c\<^sub>1, Normal t\<rangle> \<Rightarrow> Normal t' \<and> \<Gamma> \<turnstile> \<langle>c\<^sub>2, Normal t'\<rangle> \<Rightarrow> t''"
-  apply (erule exec_Normal_elim_cases(8))
-  apply (insert hoarep_exec)
-  apply fastforce
-  done
-
-lemma While_inv_from_body:
-  "\<Gamma> \<turnstile> (G \<inter> C) B G \<Longrightarrow> \<Gamma> \<turnstile> G While C B G"
-  apply (drule hoare_sound)+
-  apply (rule hoare_complete)
-  apply (clarsimp simp: cvalid_def HoarePartialDef.valid_def)
-  by (erule exec_While_final_inv''[where I="\<lambda>s s'. s \<in> G \<longrightarrow> s' \<in> Normal ` G", THEN impE],
-      fastforce+)
-
-lemma While_inv_from_body_setter:
-  "\<lbrakk>\<Gamma> \<turnstile> G setter (G \<inter> {s. s \<in> C \<longrightarrow> s \<in> Cnd}); \<Gamma> \<turnstile> (G \<inter> Cnd) B G\<rbrakk>
-   \<Longrightarrow> \<Gamma> \<turnstile> (G \<inter> {s. s \<in> C \<longrightarrow> s \<in> Cnd}) While C (Seq B setter) G"
-  apply (drule hoare_sound)+
-  apply (rule hoare_complete)
-  apply (clarsimp simp: cvalid_def HoarePartialDef.valid_def)
-  by (rule exec_While_final_inv''
-            [where I="\<lambda>s s'. s \<in> G \<and> (s \<in> C \<longrightarrow> s \<in> Cnd) \<longrightarrow> s' \<in> Normal ` G", THEN impE],
-      (fastforce dest!: intermediate_Normal_state[where c\<^sub>1=B and P="G \<inter> Cnd" and Q=G]
-                 intro: hoare_complete
-                  simp: cvalid_def HoarePartialDef.valid_def)+)
-
-lemmas hoarep_Int_pre_fix = hoarep_Int[where P=P and P'=P for P, simplified]
-
 lemma ccorres_While:
   assumes body_ccorres:
     "\<And>r. ccorresG srel \<Gamma> (=) xf (G and (\<lambda>s. the (C r s))) (G' \<inter> C') hs (B r) B'"
--- a/lib/clib/CCorresLemmas.thy
+++ b/lib/clib/CCorresLemmas.thy
@@ -1040,76 +1040,197 @@ lemma ccorres_Guard_True_Seq:
    \<Longrightarrow> ccorres_underlying sr \<Gamma> r xf arrel axf A C hs a (Guard F \<lbrace>True\<rbrace> c ;; d)"
    by (simp, ccorres_rewrite, assumption)
 
+lemma ccorres_While_Normal_helper:
+  assumes setter_inv:
+    "\<Gamma> \<turnstile> {s'. \<exists>rv s. G rv s s'} setter {s'. \<exists>rv s. G rv s s' \<and> (cond_xf s' \<noteq> 0 \<longrightarrow> Cnd rv s s')}"
+  assumes body_inv: "\<Gamma> \<turnstile> {s'. \<exists>rv s. G rv s s' \<and> Cnd rv s s'} B {s'. \<exists>rv s. G rv s s'}"
+  shows "\<Gamma> \<turnstile> ({s'. \<exists>rv s. G rv s s' \<and> (cond_xf s' \<noteq> 0 \<longrightarrow> Cnd rv s s')})
+             While {s'. cond_xf s' \<noteq> 0} (Seq B setter)
+             {s'. \<exists>rv s. G rv s s'}"
+  apply (insert assms)
+  apply (rule hoare_complete)
+  apply (simp add: cvalid_def HoarePartialDef.valid_def)
+  apply (intro allI)
+  apply (rename_tac xstate xstate')
+  apply (rule impI)
+  apply (case_tac "\<not> isNormal xstate")
+   apply fastforce
+  apply (simp add: isNormal_def)
+  apply (elim exE)
+  apply (simp add: image_def)
+  apply (erule exec_While_final_inv''[where C="{s'. cond_xf s' \<noteq> 0}" and B="B;; setter"]; clarsimp)
+     apply (frule intermediate_Normal_state[OF _ _ body_inv])
+      apply fastforce
+     apply clarsimp
+     apply (rename_tac inter_t)
+     apply (frule hoarep_exec[OF _ _ body_inv, rotated], fastforce)
+     apply (frule_tac s=inter_t in hoarep_exec[rotated 2], fastforce+)[1]
+    apply (metis (mono_tags, lifting) HoarePartial.SeqSwap empty_iff exec_abrupt mem_Collect_eq)
+   apply (metis (mono_tags, lifting) HoarePartial.SeqSwap exec_stuck mem_Collect_eq)
+  apply (metis (mono_tags, lifting) HoarePartial.SeqSwap empty_iff exec_fault mem_Collect_eq)
+  done
+
 lemma ccorres_While:
   assumes body_ccorres:
-    "\<And>r. ccorresG srel \<Gamma> (=) xf (G and (\<lambda>s. the (C r s))) (G' \<inter> C') hs (B r) B'"
-  and cond_ccorres:
-    "\<And>r. ccorresG srel \<Gamma> (\<lambda>rv rv'. rv = to_bool rv') cond_xf G G' hs (gets_the (C r)) setter"
-  and nf: "\<And>r. no_fail (G and (\<lambda>s. the (C r s))) (B r)"
-  and no_ofail: "\<And>r. no_ofail G (C r)"
-  and body_inv: "\<And>r. \<lbrace>G and (\<lambda>s. the (C r s))\<rbrace> B r \<lbrace>\<lambda>_. G\<rbrace>"
-                "\<Gamma> \<turnstile> (G' \<inter> C') B' G'"
-  and setter_inv_cond: "\<Gamma> \<turnstile> G' setter (G' \<inter> {s'. cond_xf s' \<noteq> 0 \<longrightarrow> s' \<in> C'})"
-  and setter_xf_inv: "\<And>val. \<Gamma> \<turnstile> {s'. xf s' = val} setter {s'. xf s' = val}"
-  shows "ccorresG srel \<Gamma> (=) xf G (G' \<inter> {s'. xf s' = r}) hs
-           (whileLoop (\<lambda>r s. the (C r s)) B r)
-           (Seq setter (While {s'. cond_xf s' \<noteq> 0} (Seq B' setter)))"
+    "\<And>r. ccorresG srel \<Gamma> rrel xf (\<lambda>s. G r s \<and> C r s = Some True) (G' \<inter> C') [] (B r) B'"
+  assumes setter_ccorres:
+    "\<And>r. ccorresG srel \<Gamma> (\<lambda>rv rv'. rv = to_bool rv') cond_xf (G r) G' [] (gets_the (C r)) setter"
+  assumes nf: "\<And>r. no_fail (\<lambda>s. G r s \<and> C r s = Some True) (B r)"
+  assumes no_ofail: "\<And>r. no_ofail (G r) (C r)"
+  assumes body_inv:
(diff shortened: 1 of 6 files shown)


michaelmcinerney:
Force pushed to clean up the commits a bit. Should hopefully be good to go now

lib/clib/CCorresLemmas.thy
```diff
@@ -1040,119 +1040,211 @@ lemma ccorres_Guard_True_Seq:
...
-  apply (clarsimp simp: cvalid_def HoarePartialDef.valid_def)
-  by (rule exec_While_final_inv''
-            [where I="\<lambda>s s'. s \<in> G \<and> (s \<in> C \<longrightarrow> s \<in> Cnd) \<longrightarrow> s' \<in> Normal ` G", THEN impE],
-      (fastforce dest!: intermediate_Normal_state[where c\<^sub>1=B and P="G \<inter> Cnd" and Q=G]
-                 intro: hoare_complete
-                  simp: cvalid_def HoarePartialDef.valid_def)+)
-
-lemmas hoarep_Int_pre_fix = hoarep_Int[where P=P and P'=P for P, simplified]
+  apply (simp add: cvalid_def HoarePartialDef.valid_def)
+  apply (intro allI)
```
lsf37, line 1053: Would it mess up the state too much if that was compressed into a `clarsimp` instead (includes the `allI` and `impI`).

lib/clib/CCorresLemmas.thy
```diff
@@ -1040,119 +1040,211 @@ lemma ccorres_Guard_True_Seq:
...
+  apply (rename_tac xstate xstate')
+  apply (rule impI)
+  apply (case_tac "\<not> isNormal xstate")
+   apply fastforce
+  apply (simp add: isNormal_def)
+  apply (elim exE)
```
lsf37, line 1059: Also smells like it could be a `clarsimp` (or would have been resolved by an earlier clarsimp). I can't see the proof state, though, `clarsimp` might be doing a bit too much.

lib/clib/CCorresLemmas.thy
```diff
@@ -1040,119 +1040,211 @@ lemma ccorres_Guard_True_Seq:
...
+     apply (frule intermediate_Normal_state[OF _ _ body_inv])
+      apply fastforce
+     apply clarsimp
+     apply (rename_tac inter_t)
+     apply (frule hoarep_exec[OF _ _ body_inv, rotated])
+      apply fastforce
+     apply (frule_tac s=inter_t in hoarep_exec[rotated 2])
+       apply fastforce
+      apply fastforce
+     apply fastforce
```
lsf37, line 1075:
```suggestion
     apply (frule_tac s=inter_t in hoarep_exec[rotated 2]; fastforce)
```

lib/clib/CCorresLemmas.thy
```diff
@@ -1040,119 +1040,211 @@ lemma ccorres_Guard_True_Seq:
...
+  apply (simp add: cvalid_def HoarePartialDef.valid_def)
+  apply (intro allI)
+  apply (rename_tac xstate xstate')
+  apply (rule impI)
+  apply (case_tac "\<not> isNormal xstate")
+   apply fastforce
+  apply (simp add: isNormal_def)
+  apply (elim exE)
+  apply (simp add: image_def)
+  apply (erule exec_While_final_inv''[where C="{s'. cond_xf s' \<noteq> 0}" and B="B;; setter"])
```
lsf37, line 1061: Would a `; clarsimp` do anything useful here or just add noise?

lib/clib/CCorresLemmas.thy
```diff
@@ -1040,119 +1040,211 @@ lemma ccorres_Guard_True_Seq:
...
+    apply (prop_tac "\<exists>rv' s'. rrel rv' (xf inter_t) \<and> (rv', s') \<in> fst (B r s)
+                              \<and> (s', inter_t) \<in> srel")
+     apply (insert body_ccorres)[1]
+     apply (drule_tac x=r in meta_spec)
+     apply (erule (1) ccorresE)
+         apply fastforce
+        apply fastforce
+       using nf apply (fastforce simp: no_fail_def)
+      apply (fastforce dest!: EHOther)
+     apply fastforce
```
lsf37, line 1135: this bit looks like it could potentially compressed a bit without losing readability

lib/clib/CCorresLemmas.thy
```diff
@@ -1040,119 +1040,211 @@ lemma ccorres_Guard_True_Seq:
...
-    apply (prop_tac "xf inter_t = xf t'")
-     apply (fastforce dest: hoarep_exec[rotated] intro: setter_xf_inv)
+     apply (intro conjI)
+           apply fastforce
+          using no_ofail apply (fastforce elim!: no_ofailD)
+         apply fastforce
+        apply (fastforce dest: hoarep_exec[OF _ _ setter_rrel, rotated])
+       apply (fastforce dest: hoarep_exec[OF _ _ setter_inv_cond, rotated])
+      apply (fastforce dest: hoarep_exec[OF _ _ setter_inv_cond, rotated])
+     apply (fastforce dest: hoarep_exec[OF _ _ setter_inv_cond, rotated])
```
lsf37, line 1156: I could well be wrong, but this also looks like this might be possible to compress into a single fastforce. E.g. something like 
```suggestion
     using no_ofail
     apply (fastforce elim!: no_ofailD dest: hoarep_exec[OF _ _ setter_rrel, rotated hoarep_exec[OF _ _ setter_inv_cond, rotated])
```

The advantage would be that the single fastforce is not sensitive to the order of the conjuncts should those ever change.

lib/clib/CCorresLemmas.thy
```diff
@@ -1040,119 +1040,211 @@ lemma ccorres_Guard_True_Seq:
...
+    apply (drule_tac x=r in meta_spec)
+    apply (rule hoarep_conj_lift_pre_fix)
+     apply (rule hoarep_conj_lift_pre_fix)
+      apply (insert setter_inv_cond)[1]
+      apply (drule_tac x=s in meta_spec)
+      apply (drule_tac x=r in meta_spec)
+      apply (rule_tac Q'="{s' \<in> G'. cond_xf s' \<noteq> 0 \<longrightarrow> s' \<in> C'}" in conseqPost)
+        apply fastforce
+       apply fastforce
+      apply fastforce
```
lsf37, line 1182: I'd usually prefer `; fastforce` instead for situations like that.

lib/clib/CCorresLemmas.thy
```diff
@@ -1040,119 +1040,211 @@ lemma ccorres_Guard_True_Seq:
...
+     apply (rule hoarep_conj_lift_pre_fix)
+      apply (simp add: Collect_mono conseq_under_new_pre)
+     apply (simp add: Collect_mono conseq_under_new_pre)
```
lsf37, line 1185: Does this one work?
```suggestion
     apply (simp add: hoarep_conj_lift_pre_fix Collect_mono conseq_under_new_pre)
```

lib/clib/CCorresLemmas.thy
```diff
@@ -1040,119 +1040,211 @@ lemma ccorres_Guard_True_Seq:
...
+        apply (rule hoarep_conj_lift_pre_fix)
+         apply fastforce
+        apply (simp add: conseq_under_new_pre)
+       apply fastforce
+      apply fastforce
+     apply fastforce
+    apply (case_tac xstate; clarsimp)
+    apply (frule intermediate_Normal_state[OF _ _ setter_hoarep])
+      apply fastforce
+     apply fastforce
```
lsf37, line 1234: `; fastforce`

lib/clib/CCorresLemmas.thy
```diff
@@ -1177,4 +1269,38 @@ lemma ccorres_inr_rrel_Inr[simp]:
...
+     apply (rule monadic_rewrite_is_refl)
+     apply (rule add_remove_return)
+    apply (rule ccorres_seq_skip'[THEN iffD1])
+    apply (rule ccorres_split_nothrow_novcg)
+        apply (rule ccorres_call_getter_setter)
+            apply (fastforce intro: cul)
+           apply (fastforce intro: gsr)
+          apply (simp add: gsr)
+         apply (fastforce intro: ist)
+        apply (fastforce intro: ef)
```
lsf37, line 1297: Would feel tempted to try this one in a single `fastforce`. May well fail, but would be nicer.

lsf37 (approved):
Cool. The content looks good, I just have a bunch of potential small proof compression queries. The idea would be not just to make it a bit shorter, but also slightly less sensitive to the precise order of assumptions etc in rules.

lib/clib/Corres_UL_C.thy
```diff
@@ -288,6 +288,43 @@ lemma ccorres_from_vcg_nofail:
...
+  apply (simp add: HoarePartialDef.valid_def cvalid_def)
+  apply (intro impI allI)
+  apply clarsimp
+  apply (frule (5) ccorres_empty_handler_abrupt)
+  apply (erule (4) ccorresE)
+   apply (erule (1) EHOther)
+  apply (rule image_eqI)
+   apply simp
+  apply (fastforce simp: unif_rrel_simps valid_def)
+  done
```
Xaphiosis, line 307: This proof appears nearly identical to `ccorres_to_vcg`. Does it mean that it becomes an instance of this rule? No sense repeating the proof in that case, suggest changing proof of `ccorres_to_vcg` to be an instantiation.
Also, this proof looks a bit ancient (have a feeling that fastforce/force/auto can take it from the `image_eqI` line for example), if you have time maybe try shorten it a bit?

lib/clib/Corres_UL_C.thy
```diff
@@ -288,6 +288,43 @@ lemma ccorres_from_vcg_nofail:
...
+
+lemma ccorres_to_vcg_with_prop':
+  "\<lbrakk>ccorres_underlying srel \<Gamma> rrel xf arrel axf P P' [] a c; no_fail Q a; \<lbrace>P\<rbrace> a \<lbrace>R\<rbrace>\<rbrakk>
+   \<Longrightarrow> \<Gamma> \<turnstile> {s'. P s \<and> Q s \<and> s' \<in> P' \<and> (s, s') \<in> srel}
+           c {t'. \<exists>rv t. (t, t') \<in> srel \<and> rrel rv (xf t') \<and> R rv t}"
+  apply (frule (2) ccorres_to_vcg_with_prop[where s=s])
+  apply (fastforce elim: conseq_under_new_pre)
+  done
```
Xaphiosis, line 315: I'd be left scratching my head as to when to use this rule vs any of the other ones.
Suggestion: since it's a two-line proof without interesting bits, use a `by`

lib/clib/Corres_UL_C.thy
```diff
@@ -288,6 +288,43 @@ lemma ccorres_from_vcg_nofail:
...
+
+lemma ccorres_to_vcg_gets_the:
+  "\<lbrakk>ccorres_underlying srel \<Gamma> rrel xf arrel axf P P' [] (gets_the r) c; no_ofail P r\<rbrakk>
+   \<Longrightarrow> \<Gamma> \<turnstile> (P' \<inter> {s'. (s, s') \<in> srel \<and> P s})
+           c {t'. (s, t') \<in> srel \<and> P s \<and> (r s \<noteq> None) \<and> rrel (the (r s)) (xf t')}"
+  apply (frule ccorres_to_vcg_with_prop[where R="\<lambda>_. P" and s=s])
+    apply (erule no_ofail_gets_the)
+   apply wpsimp
+  apply (clarsimp simp: gets_the_def gets_def bind_def return_def get_def assert_opt_def
+                        fail_def conseq_under_new_pre
```
Xaphiosis, line 325: Don't know if it helps here, but reminder that `simpler_gets_def` exists and I've yet to see when we want gets_def

lib/clib/Corres_UL_C.thy
Xaphiosis: Also same comment as Gerwin on the while rule, `clarsimp` includes doing `allI` and `impI` already, so that `simp, intro, clarsimp` might be compressible.

lib/clib/CCorresLemmas.thy
```diff
@@ -1177,4 +1177,38 @@ lemma ccorres_inr_rrel_Inr[simp]:
...
+       apply (rule ceqv_refl)
```
Xaphiosis, line 1298: out of curiosity (not suggesting you change it), but why not `apply ceqv`?

lib/clib/CCorresLemmas.thy
```diff
@@ -1040,76 +1040,211 @@ lemma ccorres_Guard_True_Seq:
...
+                setter
+                {s'. s' \<in> G' \<and> (cond_xf s' \<noteq> 0 \<longrightarrow> s' \<in> C')}"
+  assumes setter_rrel:
+    "\<And>r s. \<Gamma> \<turnstile> {s'. s' \<in> G' \<and> (s, s') \<in> srel \<and> G r s \<and> rrel r (xf s')}
+                setter
+                {s'. rrel r (xf s')}"
+  shows
+    "ccorresG srel \<Gamma> rrel xf (G r) (G' \<inter> {s'. rrel r (xf s')}) hs
+       (whileLoop (\<lambda>r s. the (C r s)) B r)
+       (setter;; (While {s'. cond_xf s' \<noteq> 0} (B';; setter)))"
```
Xaphiosis, line 1104: I'm a bit puzzled by this rule (ok, besides getting dizzy trying to follow all the `*rel`s). In a "typical" C loop, you have an initialiser, a loop body with a test at the front, and then an increment/advance at the end. This pattern is so common it got enshrined as the `for` loop. In this rule, however, you have `setter; while (cond) {B; setter;}` which would indicate a much more specific type of while loop, what we'd see with a do-while loop: `do {B; setter} while (cond)`.

Is this what's happening on the C side here where you're trying to apply this rule? Did they use a do-while there? I'm not sure what the C parser does with a do-while honestly.

If that's what we're looking at, then the name `ccorres_While` might be a slightly misleading name, and possibly the C code while loop might be a bit strange.

lsf37: The `do .. while` loop potentially does one more `B` than the `while` loop, so it's not quite the same.

If I understood correctly, the `setter` part is state change that is related to the condition, but not the body, and is usually not state change that affects the state relation. A better name would be welcome, but we haven't found one yet. Come to think of it, we did talk about needing a comment for this rule, specifically describing the role of the `setter` bit. We should really add that.

You have a point that the name `ccorres_While` is a bit too general, because it does expect a specific form of while loop on the C side.

michaelmcinerney: I haven't tried looked into exactly what the C parser is doing (I doubt I would understand it anyway) but this format seems to be produced whenever there is a separate function which is used as the loop condition in a plain `while` statement. I could check, but I think when there isn't a separate function for the loop condition, then there would be no `setter` as I have in this rule.

The reason I do want to have a separate name (possibly inlined function in the C) for the loop condition is that it makes writing a `ccorres` lemma much easier than if I were to try to paste a big blob of Simpl into a `ccorres` statement. One of the `while` loops in `refill_budget_check` does not use a separate function, and the Simpl version of the loop condition there is very involved and ugly.

Would it be reasonable to always use a C function for the loop condition?

lsf37: Actually, you're right, I suspect the C parser is always going to produce this pattern if the loop condition has a potential side effect. (But will not if it doesn't). The pattern might be harder to extract if the body is not just a function call, because it'll probably just put a semicolon at the end, which is hard to match against if there are other semicolons in the body, but it should still be there in principle.

So what `setter` really is is the potentially side-effectful expression in the loop condition position in C. So a better name would actually be `condition` after all. We could look into the C parser to change associativity of that final semicolon so that the pattern here always fits for those kinds of loops, which would make the rule applicable to all while loops that have a condition with side effect.

michaelmcinerney: I would be quite happy if we always had the loop body in the C as its own function call, but I wouldn't be surprised if it's not in keeping with the style the C has. If it weren't its own function call, then could always just use one of the rules about associativity of the Simpl in a `ccorres` rule to change things around, right? I wouldn't be averse to changing what the C parser does here, though.

But thinking more generally, are we happy to always require our C `while` loops to use a standalone function as the loop condition?

lib/clib/CCorresLemmas.thy
michaelmcinerney: I use `erule exec_While_final_inv''` a few lines later, and I'm wanting to get the conclusion in a particular form at this point, so that I can match perfectly. The predicate I'm wanting to match is a bit clunky. Would it be better to instantiate the predicate manually, or to use this sequence of finicky proof steps and than just match?

lib/clib/CCorresLemmas.thy
michaelmcinerney: Sure, that gets rid of a few of the `fastforce`s

lib/clib/CCorresLemmas.thy
michaelmcinerney: This spins. It even spins with `blah; fastforce?`. It seems like the best I can do is `blah; clarsimp` followed by two `fastforce`s, which only gets rid of one line.

lib/clib/CCorresLemmas.thy
michaelmcinerney: I can do ` using nf subgoal by (fastforce elim: ccorresE simp: no_fail_def dest!: EHOther)` but it takes about 5 seconds, compared with the 1 to 2 seconds that this chunk currently takes. Is that preferable?

lib/clib/CCorresLemmas.thy
michaelmcinerney: I couldn't get this or something like it to work. The conjuncts are from the statement `helper` itself, so if they did change, it would be from some fairly major surgery on this rule itself. Is it worth looking into some other approach to this?

I did try getting rid of this `helper` bit, but it's useful, again because of the rather specific predicate that `exec_While_final_inv''` uses.

lib/clib/CCorresLemmas.thy
michaelmcinerney: Sadly not. I can get `apply (fastforce intro: hoarep_conj_lift_pre_fix simp add: Collect_mono conseq_under_new_pre)` to work, but it takes a good few seconds.

lib/clib/CCorresLemmas.thy
michaelmcinerney: That spins too. Here I can do `blah; assumption?` and then just one `fastforce`, which is again just one line better.

lib/clib/CCorresLemmas.thy
michaelmcinerney: Can't get that to work, either.

lib/clib/CCorresLemmas.thy
michaelmcinerney: `ceqv` isn't available in this theory file. Should we investigate making it available?

lsf37: I suspect that would be too much reshuffling. Given that the use is trivial here, I think leaving it is Ok.

lib/clib/CCorresLemmas.thy
lsf37: > I would be quite happy if we always had the loop body in the C as its own function call, but I wouldn't be surprised if it's not in keeping with the style the C has. 

Yes, that would definitely not be normal, having the loop body as a single function call is definitely an exception.


> If it weren't its own function call, then could always just use one of the rules about associativity of the Simpl in a `ccorres` rule to change things around, right? 

Yes, that would be possible, but somewhat annoying to use.


> I wouldn't be averse to changing what the C parser does here, though.
 
Happy to look into that (in a separately from this PR).

> But thinking more generally, are we happy to always require our C `while` loops to use a standalone function as the loop condition?

No, definitely not. The default situation is that the loop condition does not affect the state and is an otherwise arbitrary expression. Most loops are of that form. Loop conditions that do change the state (as we have here) should be the exception. Even there I don't think we can demand that the condition should always be just one function call. For the cases we are seeing in MCS, it makes sense to have them as a single function call, but I wouldn't say that in general.

lib/clib/CCorresLemmas.thy
lsf37: Hm, depends on how big the instantiation is. If it's reasonably small, that would probably be preferable, because it would rely less on the exact term structure of the goal. If it's as big as stating another lemma it's not really worth it, though.

lib/clib/CCorresLemmas.thy
lsf37: That is interesting, that means it depends on the order the goals are solved in. How about `(blah, fastforce+)[1]` or Raf's `fwd_all_new` (i.e. fwd_all_new <blah> fastforce`).

lib/clib/CCorresLemmas.thy
lsf37: If it's that much slower then the expanded form is better. Could try a `; fastforce?` but that is probably reaching for straws.

lib/clib/CCorresLemmas.thy
lsf37: It's Ok, if it doesn't work then let's leave it as is. Sometimes it's not really worth the time.

lib/clib/CCorresLemmas.thy
lsf37: Hm. Even with `intro!: ..`? If not, we could fall back to `rule ..; simp add: ...`

lib/clib/CCorresLemmas.thy
lsf37: Might also be the order again, so `(.., fastforce+)[1]` might have a chance.

lib/clib/CCorresLemmas.thy
lsf37: Alright, was worth a shot :-)

lib/clib/Corres_UL_C.thy
michaelmcinerney: Sorry, I meant to make a comment about these rules. There are some variations on this rule: `ccorres_to_vcg'`, which appears in CLib and is pretty much `ccorres_to_vcg` in clarsimp normal form, and then there's also `ccorres_to_vcg_nf` in CRefine, which uses `no_fail` rather than `snd _`. Should I consolidate these? I could prove the `with_prop` version, and get any other variant that's required from it

lib/clib/CCorresLemmas.thy
michaelmcinerney: Both of those work here. Is one preferable?

lib/clib/CCorresLemmas.thy
michaelmcinerney: `; fastforce?` spins. I suppose I'll leave it as it was for now, unless we come up with some other ideas

lib/clib/CCorresLemmas.thy
michaelmcinerney: `intro!` does help, actually, and the original `simp`s weren't so fast, so this is probably no slower. I'll go with that. Thanks

lib/clib/Corres_UL_C.thy
michaelmcinerney: I might be able to get rid of this one, actually. I can use the stronger form just above, I'm pretty sure.

The `with_prop` versions are probably useful only in some sort of a `ccorres` rule for some sort of loop, where we'd like the `\exists rv t` bit and then want to carry a loop invariant through, too. But you get the property almost for free, so I thought I could prove this and then get `ccorres_to_vcg` from having the trivial postcondition

lib/clib/CCorresLemmas.thy
michaelmcinerney: I tried getting a manual instantiation working, but it got very annoying and I think I might give up on that.

lsf37: No worries, it was worth a try.

lib/clib/CCorresLemmas.thy
lsf37: Inside a proof method, I'd prefer `fwd_all_new`, because it's more precise, but in normal proof scripts the `(_, _+)[1]` is probably more familiar to people.

lib/clib/Corres_UL_C.thy
michaelmcinerney: I was able to get rid of this rule. So now there's just a question of whether to clean up the other variants of this rule, `ccorres_to_vcg'` and `ccorres_to_vcg_nf`

Xaphiosis: Cool. I'd say yes, if any rules become easily provable due to the new one you've added, then cleaning up their proofs is preferable.

lib/clib/Corres_UL_C.thy
Xaphiosis: I honestly don't know which of these is the better form. Maybe @lsf37 does. Thumbs up for consolidation if you have time for it.

michaelmcinerney: I'd like to hear Gerwin's opinion, but I'm pretty sure that having `ccorres_to_vcg_with_prop` as in this PR, and then replacing the current `ccorres_to_vcg` with

```
lemma ccorres_to_vcg:
  "⟦ccorres_underlying srel Γ rrel xf arrel axf P P' [] a c; no_fail Q a⟧
   ⟹ Γ ⊢ {s'. P s ∧ Q s ∧ s' ∈ P' ∧ (s, s') ∈ srel}
           c {t'. ∃(rv, t) ∈ fst (a s). (t, t') ∈ srel ∧ rrel rv (xf t')}"
  apply (frule (1) ccorres_to_vcg_with_prop[where R="⊤⊤"])
   apply wpsimp
  apply fastforce
  done
```

would be the best way to go.

The original `ccorres_to_vcg` isn't in clarsimp normal form. Then `ccorres_to_vcg'` requires `no_fail` for the abstract function with the precondition `\top`, which is stronger than what I require with this new formulation. And then `ccorres_to_vcg_nf` requires the precondition for `no_fail` to be no stronger than the abstract guard of the `ccorres` rule, which is not necessary.

michaelmcinerney: That's not quite right. I suppose it's that `ccorres_to_vcg'` requires the abstract function to not fail on the given abstract state. But non-failure is probably best dealt with via an actual `no_fail` assumption, which this rule doesn't do.

michaelmcinerney: It seems like the variants of `ccorres_to_vcg` were actually unused in CRefine. Updates in the last commit

lsf37: Cool. I'm happy with the version in the last commit.

Xaphiosis (approved):
I think this is good now, don't forget to squash. I'd also prefer it if you squashed the "clib: some proof compression" commit into prior commits, not really something we need in the history.

PR #675, opened 2023-10-07
Cpp wrapper

adbrucker:
A small update to the  cpp-script that allows to use the same script on Linux/Windows (with the cpp from gcc) and macOS. This makes it easier to share the same configuration across a team using different operating systems.


lsf37:
Thanks for drawing our attention to this script -- I don't think it's actually necessary any more, at least I haven't used it any more in ages and I'm one of the main Darwin users of the l4v repo.

I'm tempted to remove it from the repo. Did you find the tool necessary for other applications of the parser?

adbrucker:
The situation is the opposite for me: I seem to the only non-macOS/Darwin user. I just checked with the macOS/Darwin user in my team. And, indeed, the script does not seem to be necessary any longer. Hence, removing it from the repo seems to be  a sensible action.

lsf37:
Perfect, then I'll close this PR and make another one to remove the script.

PR #678, opened 2023-10-10
CKernel, CSpec, and CBaseRefine sessions for AArch64

lsf37:
- set up base sessions for CRefine
- disable SimplExportAndRefine for AArch64 (no BV infrastructure for this architecture yet)
- switch off `quick_and_dirty` for Refine session (proof finished)
- switch on `quick_and_dirty` for CRefine session (proof in development)

### DIFF
--- a/spec/cspec/AARCH64/Kernel_C.thy
+++ b/spec/cspec/AARCH64/Kernel_C.thy
@@ -1,0 +1,110 @@
+(*
+ * Copyright 2023, Proofcraft Pty Ltd
+ *
+ * SPDX-License-Identifier: GPL-2.0-only
+ *)
+
+theory Kernel_C
+imports
+  "ExecSpec.MachineTypes"
+  "CLib.CTranslationNICTA"
+  "AsmRefine.CommonOps"
+begin
+
+external_file
+  "../c/build/$L4V_ARCH/kernel_all.c_pp"
+
+context begin interpretation Arch .
+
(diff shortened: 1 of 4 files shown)


spec/cspec/AARCH64/Kernel_C.thy
```diff
@@ -0,0 +1,110 @@
...
+(* workaround for the fact that the C parser wants to know the vmpage sizes*)
+(* create appropriately qualified aliases *)
+context begin interpretation Arch . global_naming vmpage_size
+requalify_consts ARMSmallPage ARMLargePage ARMHugePage
+end
+
+definition
+  ctcb_size_bits :: nat
+where
+  "ctcb_size_bits \<equiv> 11"
```
Xaphiosis, line 79: I checked against the latest Kernel_C on aarch64 crefine branch and it's 10, because I got it wrong in this version. Consider pulling in that one? 04ea4e5468c1f3d9d20b41b31fb816335a3f4c0a

Xaphiosis: (probably I squashed that change in at some point, but I definitely pushed it out)

lsf37: Will do. We should really make that symbolic at some point, but today is not that day.

Xaphiosis (approved):

PR #680, opened 2023-10-12
MCS: merge master into rt

corlewis:
This includes the update to Isabelle2023. As usual with merges, the last commit is probably the only one that it makes sense to review.


lsf37 (approved):
Good work. There's a whole lot more `mapsto` and `corresKsimp` going on in `rt` than I thought.

Since the test is still set to Isabelle2022 we should probably do a PR on `mcs.xml` in the manifest repo first and the re-run the tests here to double-check before we merge.

michaelmcinerney (approved):
Looks good!

Yeah that is a lot of `corresKsimp`, almost all mine, I think. It would be interesting to go through and see how quickly they could be replaces with the new `corres` method.

proof/invariant-abstract/CSpace_AI.thy
```diff
@@ -233,7 +233,7 @@ lemma preemption_point_inv:
   apply (rule hoare_seq_ext_skipE, wpsimp)
   apply (rule valid_validE)
   apply (rule OR_choiceE_weak_wp)
-  apply (rule alternative_valid; (solves wpsimp)?)
+  apply (rule alternative_wp[where P=P and P'=P for P, simplified]; (solves wpsimp)?)
```
corlewis, line 236: I forgot to point this out when I made this PR. `alternative_valid` was removed on master due to not being used, but did have a couple of uses in `rt` as part of some forward reasoning proofs.

Are we ok with what I've done here, which is basically inlining it in the two places it was used? The obvious to me other options would be to either add `alternative_valid` back to `Nondet_VCG`, or to rework these proofs to be backwards.

michaelmcinerney: I'd be OK with what you've done here, but wouldn't mind if these proofs were reworked to be backwards. I'm pretty sure they'd be my proofs originally, so I'd be happy to do that. I think these were for `preemption_point`, which is not the nicest function to deal with, so it might not be so straightforward.

lsf37: I'd be fine with either. Definitely no problems with it being reworked as backwards proof. Could also be in a separate PR.

PR #681, opened 2023-10-12
Remove more uses of _tac methods from Monads

corlewis:


lsf37 (approved):
Excellent, thank you.

PR #683, opened 2023-10-17
MCS CRefine sorried for RISCV64

michaelmcinerney:
This sorries the last files in the CRefine session, thereby completing the setup of the proof infrastructure for MCS CRefine. 

In addition, this enables testing of the CRefine session via our continuous integration setup.

### DIFF
--- a/proof/Makefile
+++ b/proof/Makefile
@@ -15,6 +15,11 @@ ifeq "$(L4V_ARCH)" "AARCH64"
   export REFINE_QUICK_AND_DIRTY=1
 endif
 
+# Allow sorry command in RISCV64 CRefine during development:
+ifeq "$(L4V_ARCH)" "RISCV64"
+  export CREFINE_QUICK_AND_DIRTY=1
+endif
+
 #
 # Setup heaps.
 #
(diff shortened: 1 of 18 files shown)


Xaphiosis:
Looking at the commits for now:

The "update copyright" commit (no tags, no explanantion), at least the CRefine parts look like they should be folded into the sorrying run. Not sure about the lib part.

For the "rt crefine: finish sorrying run", I know I lost the battle about indicating in the tags that this is only for RISCV64, but maybe I could get a comment in the body of the commit, as it's quite a large commit?

For the run_tests and Makefile changes, why are they tagged "rt:", these are not rt specific. The first should say "run_tests: enable CRefine for RISCV64". The second is misleading, as it's RISCV64 only, and also not rt-specific; I'd suggest Gerwin-styled "riscv64 proofs: turn on quick_and_dirty for CRefine".

Xaphiosis:
`rt: update the sorrying-run to seL4 revision 4214775` commit is also crefine-only and not tagged as such

lsf37 (approved):
Alright, the commit messages are cleaned up and the content has been good the entire time already, so that is also fine.

Thanks for diligently adding new lemmas for all of the new MCS functions, this makes the sorry-count much more useful.

PR #684, opened 2023-10-18
copyright update for MCS file changes

lsf37:
Add Proofcraft copyright header to the files with substantial changes where we had forgotten to do so.


michaelmcinerney (approved):
Thanks Gerwin

PR #686, opened 2023-10-20
docs: add a guide for commit messages

lsf37:
Add a guide for how to write commit messages and commit message tags to make the messages more consistent and to help new people on the project get started more quickly.

### DIFF
--- a/docs/commit-messages.md
+++ b/docs/commit-messages.md
@@ -1,0 +1,392 @@
+<!--
+     Copyright 2023, Proofcraft Pty Ltd
+
+     SPDX-License-Identifier: CC-BY-SA-4.0
+-->
+
+# Commit Messages
+
+## Objective
+
+The l4v repository is relatively large, active, and long-lived. It has a public
+history of about one decade, and an additional decade of unreleased private
+history. It will hopefully live on for another 20 years.
+
+This means that the commit history is important. Examples of questions the commit
+history should be able to answer reasonably quickly and painlessly are:
+
+- > Is this written in a strange way for good reasons, or were we just in a hurry?
+  > What was the reason? Does it still apply?
+
+- > When did we change doing things this way and why?
+
+- > Has this always been broken or was there a seemingly unrelated change that broke it?
+
+- > How long did it take to finish this proof?
+
+- > How much change was necessary to do this proof?
+
+- > Where did this library lemma come from?
+
+## General
+
+The [seL4 repository guidelines][git-conventions] apply to the `l4v` repository,
+with the following exceptions and additions:
+
+- header can be max 78 characters
+- body is wrapped at 78 characters
+- we use tags in the header to indicate which part of the repository
+  the commit applies to
+
+We are using 78 for the header because of the tags, which take up some space. If
+you can manage to stay within 50 characters anyway, that is appreciated, but it's
+not always practical. Using a body wrap of 72 is also allowed, since that is the
+default for other seL4 repositories.
+
+We use tags, because the repository is relatively large and most commits only
+apply to small parts of it. The tags make it easy to identify relevance of a
+commit at a glance while scanning through a large number of commits.
+
+The general guidelines prescribe capitalising the commit header. We do not
+capitalise the tag or after the tag, but we do capitalise the (rare) cases where
+there is no tag in the commit header.
+
+## Header and Body
+
+There is good general advice on [commit message writing][commit-messages]
+available on the internet and it is as relevant to proofs as it is to code.
+You should read it, it's not long and it's good advice.
+
+Repeating some high-level points:
+
+- Use imperative voice (e.g. `proof: rename lemma foo to bar`)
+- The header should be a short summary, focusing on "what"
+- The body should explain what is going on in more detail (also in imperative
+  voice), but much more importantly *why* it is going on (is `bar` more
+  consistent than `foo`? Is the name `foo` needed for something else? Does `bar`
+  just better describe what is going on?).
+- You are trying to explain things to your future self or a future colleague
+  5-10 years from now. You can assume knowledge of the repository in general,
+  but you should not assume specific context that is obvious to you right now,
+  but that will not be known to a different person 5 years from now.
+
+## Tags
+
+- We use tags to indicate which part of the repository the commit applies to,
+  and if it is architecture-specific, then to which architecture it applies to.
+
+- We do not usually use branch tags, because git branches are ephemeral and we
+  are using rebase branches for most of our work. The one exception is the `rt`
+  branch, which has been alive for over half a decade. For this branch, we allow
+  merge commits (from branch `master` into `rt` only), and we want to be able to
+  reconstruct a rebase history from that at the end of the branch's life.
+
+  This means, we do use the tag `rt` for commits that only make sense on this
+  branch. If you could apply the commit to the master branch directly (e.g. you're
+  adding a lemma to a library), it should not get the tag. Otherwise it should.
+
+### Tag Examples
+
+The main tags we use are mostly the directory names of the main proof something
+is in, e.g. `refine`, `crefine`, `sys-init`, `camkes`. For some of these, there
+are abbreviations, mainly `aspec` for the abstract spec and `ainvs` for the
+abstract invariants.
+
+For large directories that have logical sub parts, we use slashes, e.g.
+`lib/monads`. Not so much because the change is in that directory, but because
+we want to see that it's a library change and applies to the monad part of the
+library.
+
+If the change applies to many proofs, for instance large lemma renaming commits,
+we use tags such as `proof` and `spec`.
+
+We combine tags with `+` if a change applies to multiple parts, e.g.
+`clib+crefine` or `lib+proof`.
+
+If something is specific to an architecture we preface the tag with the
+architecture, e.g. `arm refine:` or `aarch64 aspec+ainvs:`. The current
+architecture tags are: `arm`, `arm-hyp`, `x64`, `riscv`, `aarch64`.
+Please use these spellings only.
+
+More tag examples:
+
+- `trivial`: for small trivial changes like fixing a typo, where no proofs or
+  specs have changed, i.e.\ that would not need a proof CI run.
+- `cleanup:` for cleanups that do not change proof content
+- `github:` for GitHub CI changes
+- `run_tests`: for changes to the top-level `run_tests` file
+- `isabelle20xx`: for easily identifying commits related to Isabelle updates
+
+For more consistency there is an order between tags. More abstract/general
+things should come first, e.g. `lib` < `aspec` < `haskell` < `ainvs` < `refine`
+`orphanage` < `crefine`. Or `dspec` < `drefine` and `access` < `infoflow`. Specs
+< proofs and `infoflow` < refinement proofs. This is not a total order, it's Ok
+to use your own judgement.
+
+Because `lib` has many subdirectories and separate parts, it's fine to use
+session names as tags there to shorten things a bit, e.g. `clib`, `monads`,
+`word_lib` instead of `lib/clib`, `lib/monads`, or `lib/word_lib`. This makes
+sense when the tags are session names.
+
+See also the longer example list of [good tags](#good-tags) below.
+
+## Tips and Tools
+
+### Looking at history
+
+The main tools to interact with the git history are browsing it on GitHub and
+various forms of `git log`:
+
+```sh
+git log --oneline  # show only headings
+git log            # show commit info with author, date, message
+git log --stat     # additionally show which files have changed
+git log --p        # additionally show full diff
+```
+
+For all of these, you can supply a path to restrict the log to a certain file
+or directory in the repo. You can also supply a branch, or a commit range like
+`master..rt` to restrict the output.
+
+Especially `git log --oneline` is useful for quickly getting an overview. Example
+output:
+
+```
+b3c6df48a clib: improve ccorres_While
+49ff8457f clib+crefine: improve and consolidate variants of ccorres_to_vcg
+8c433c085 clib: add some rules for hoarep
+82b954817 clib: improve ccorres_call_getter_setter
+8c59df449 lib/monads: remove more uses of _tac methods
+563232397 run_tests+proof: exclude SimplExportAndRefine for AARCH64
+1cce5b3ff proof: switch AArch64 quick_and_dirty from Refine to CRefine
+402a8342d run_tests: enable CBaseRefine for AARCH64
+32a672412 aarch64 cspec: add Kernel_C.thy to base CKernel image on
+b2cd1ce4a aarch64 asmrefine: copy ArchSetup from RISCV64
+67c0109b7 lib/monads: avoid clarsimp as initial Isar method
+bd5026438 lib/monads: fix remaining document preparation issues
+4d0086567 lib/monads: add new Trace_* files to ROOT
+598e19dd6 lib/monads: coherent document structure
+4cbfb4ab0 lib/monads: minor style + warning cleanup
+b2dd5db6d lib/monads: fix document preparation issues
+03a045309 lib/monads: add AFP document setup
+d0bab9c79 misc/scripts: remove Darwin cpp wrapper
+```
+
+You can very quickly see that C verification has been active recently, that
+new tests were added, that AARCH64 refinement proofs have been done, and that there was
+some work to do with the AFP and the monad library. You can see that nothing
+has happened with the system initialiser or other user-level proofs, and that there
+are no changes that should affect, for instance, the C parser.
+
+You only see such things quickly when the messages are consistent and follow the
+same kind of pattern. It's not so important what the pattern is. It is important
+that it is consistent.
+
+Note in e.g. `proof: switch AArch64 quick_and_dirty from Refine to CRefine` that
+the architecture tag is used only when the change is specific to files for that
+architecture. In this commit, the overall ROOTS file is changed, so it shouldn't
+get the architecture tag.
+
+### What tag should I pick?
+
+If you're uncertain what tag to pick for a certain file or directory, the
+easiest way to figure it out is to do
+
+```sh
+git log --oneline <that-file>
+```
+
+Do your tags the same way people have done before. This will make the pattern
+consistent and should be reasonably easy to read even if it's not perfect. Look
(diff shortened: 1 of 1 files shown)


lsf37:
The list of currently used tags is generated from the repo and probably still contains some few that are inconsistent or unwanted. Help appreciated in weeding these out.

docs/commit-messages.md
```diff
@@ -0,0 +1,386 @@
...
+arm ainvs
+arm aspec
+arm crefine
+arm haskell
+arm infoflow
+arm refine
+arm+arm-hyp crefine
+arm+arm-hyp machine
+arm+arm_hyp crefine
+arm+arm_hyp machine
```
lsf37, line 240: the underscore version of `arm_hyp` is wrong

docs/commit-messages.md
```diff
@@ -0,0 +1,386 @@
...
+arm-hyp refine
+asmrefine
+aspec
+aspec+access
+aspec+ainvs
+aspec+design+haskell
+aspec+haskell
+autocorres
+bisim
+c-kernel
```
lsf37, line 259: do we want `c-kernel` or `ckernel`. Neither is a directory. The session is CKernel, so I'd be for `ckernel` as the tag.

docs/commit-messages.md
```diff
@@ -0,0 +1,386 @@
...
+license
+misc
+misc/regression
+misc/scripts
+misc/stats
+proof
+proof+autocorres
+proof/Makefile
+proof/ROOT
+proofs
```
lsf37, line 328: should probably only use `proof`, not `proofs`

docs/commit-messages.md
```diff
@@ -0,0 +1,386 @@
...
+- > Where did this library lemma come from?
+
+## General
+
+The [seL4 repository guidelines][git-conventions] apply to the `l4v` repository,
+with the following exceptions and additions:
+
+- header can be max 78 characters
+- body is wrapped at 78 characters
+- we use tags in the header to in indicate which part of the repository
```
michaelmcinerney, line 38: Remove extra `in`

docs/commit-messages.md
```diff
@@ -0,0 +1,386 @@
...
+For large directories that have logical sub parts, we use slashes, e.g.
+`lib/monads`. Not so much because the change is in that directory, but because
+we want to see that it's a library change and applies to the monad part of the
+library.
+
+If the change applies to many proofs, for instance large lemma renaming commits,
+we use tags such as `proof` and `spec`.
+
+We combine tags with `+` if a change applies to multiple parts, e.g.
+`clib+crefine` or `lib+proof`.
```
michaelmcinerney, line 103: `clib` is a subdirectory of `lib`, so maybe we should either use `lib/clib` or make an exception for `clib`. `clib` is used quite a few times below

docs/commit-messages.md
```diff
@@ -0,0 +1,386 @@
...
+aarch64 aspec
+aarch64 aspec+ainvs
+aarch64 aspec+design
+aarch64 aspec+haskell
+aarch64 aspec+machine
+aarch64 cspec
+aarch64 design
+aarch64 design+machine
+aarch64 haskell
+aarch64 haskell+aspec
```
michaelmcinerney, line 217: This would go against the ordering you gave above, but maybe that doesn't matter too much

docs/commit-messages.md
```diff
@@ -0,0 +1,386 @@
...
+arm+arm_hyp machine
+arm-hyp aspec
+arm-hyp aspec+design
+arm-hyp ainvs
+arm-hyp crefine
+arm-hyp design
+arm-hyp haskell
+arm-hyp haskell+refine
+arm-hyp machine
+arm-hyp proofs+ROOT
```
michaelmcinerney, line 249: `ROOT` seems like it should go before pretty much anything else, or be in its own commit, but I'm not sure

docs/commit-messages.md
michaelmcinerney: `ckernel` sounds good to me

docs/commit-messages.md
```diff
@@ -0,0 +1,386 @@
...
+aspec+ainvs
+aspec+design+haskell
+aspec+haskell
+autocorres
+bisim
+c-kernel
+c-parser
+c-parser+autocorres
+c-parser+crefine
+c-parser+crefine+clib
```
michaelmcinerney, line 263: This again seems to violate the ordering

docs/commit-messages.md
```diff
@@ -0,0 +1,386 @@
...
+riscv64+aarch64 ainvs
+run_tests
+run_tests+proof
+spec+proof
+style
+sys-init
+tools
+tools/asmrefine
+trivial
+word_lib
```
michaelmcinerney, line 367: `word_lib` might be another exception to a subdirectory of `lib`

michaelmcinerney (approved):
Just some small nitpicks. But this looks very good and is very helpful. Especially for newcomers. Thanks Gerwin

corlewis:
One thing I'm not sure about for the tags, are we more interested in the intention behind the commits or the specific files changed?
For example, let's say that we're making a change to `lib` and are then updating the rest of the proofs as required. However, it turns out that the only change needed is in `proof/refine/ARM`. In this case, which tag would be more useful, `proof` or `arm refine`?
I can see arguments for either, the latter is more useful for seeing which part of the repository have or have not changed, but it also implies more relevance than actually exists and could be distracting in the git history.

docs/commit-messages.md
```diff
@@ -0,0 +1,386 @@
...
+## Header and Body
+
+There is good general advice on [commit message writing][commit-messages]
+available on the internet and it is as relevant to proofs as it is to code.
+You should read it, it's not long and it's good advice.
+
+Repeating some high-level points:
+
+- Use imperative voice (e.g. `proofs: rename lemma foo to bar`)
+- The header should be a short summary, focussing on "what"
```
corlewis, line 62: It's debated but I think that `focusing` is more standard.

docs/commit-messages.md
```diff
@@ -0,0 +1,386 @@
...
+  5-10 years from now. You can assume knowledge of the repository in general,
+  but you should not assume specific context that is obvious to you right now,
+  but that will not be known to a different person 5 years from now.
+
+## Tags
+
+- We use tags to indicate which part of the repository the commit applies to,
+  and if it is architecture-specific, then to which architecture it applies to.
+
+- We do no usually use branch tags, because git branches are ephemeral and we
```
corlewis, line 77:
```suggestion
- We do not usually use branch tags, because git branches are ephemeral and we
```

docs/commit-messages.md
```diff
@@ -0,0 +1,386 @@
...
+git log --stat     # additionally show which files have changed
+git log --p        # additionally show full diff
+```
+
+For all of these, you can supply a path to restrict the log to a certain file
+or directory in the repo. You can also supply a branch, or a commit range like
+`master..rt` to restrict the output.
+
+Esp `git log --oneline` is useful for quickly getting an overview. Example
+output:
```
corlewis, line 143: This abbreviation is pretty common but it still feels better to me to use the full word
```suggestion
Especially `git log --oneline` is useful for quickly getting an overview. Example
output:
```

docs/commit-messages.md
```diff
@@ -0,0 +1,386 @@
...
+b2dd5db6d lib/monads: fix document preparation issues
+03a045309 lib/monads: add AFP document setup
+d0bab9c79 misc/scripts: remove Darwin cpp wrapper
+```
+
+You can very quickly see that C verification has been active recently, that
+new tests were added, that AARCH64 proofs have been done, and that there was
+some work to do with the AFP and the monad library. You can see that nothing
+has happened with system initialiser or other user-level proofs, and that there
+are no changes that should affect for instance the C parser.
```
corlewis, line 170:
```suggestion
has happened with the system initialiser or other user-level proofs, and that there
are no changes that should affect, for instance, the C parser.
```

lsf37:
> One thing I'm not sure about for the tags, are we more interested in the intention behind the commits or the specific files changed? For example, let's say that we're making a change to `lib` and are then updating the rest of the proofs as required. However, it turns out that the only change needed is in `proof/refine/ARM`. In this case, which tag would be more useful, `proof` or `arm refine`? I can see arguments for either, the latter is more useful for seeing which part of the repository have or have not changed, but it also implies more relevance than actually exists and could be distracting in the git history.

Generally we want the intention. If it's cheap it's also good to be specific, though, because like you said it enables one to quickly see what has *not* changed. So in this specific case, I'd go for `arm refine`, but I'm open to arguments against.

docs/commit-messages.md
lsf37: I'd agree, `ROOT` should go first. If it makes a logical unit and is small enough I would not say it has to be separate commits, but depending on what it is, it probably often makes sense.

docs/commit-messages.md
lsf37: No, that's good, that's exactly the examples I was wanting to weed out. The list here is just what historically happened, the guide is supposed to be what we want to happen.

corlewis (approved):
Thanks for writing this up, it should be very helpful in us keeping a useful git history!

docs/commit-messages.md
lsf37: This is an interesting point and maybe we should discuss that one a bit. `CLib` is both a session in the global session name space and a `clib` is a subdirectory of `lib`. We don't say `proof/refine` etc either, so maybe we should just get rid of the slashes altogether. I've used them in the past for things that I thought too obscure/small for people to know or find quickly. I'd be open to just not doing that.

corlewis: `Monads` is a case where I've been using slashes recently, for example `lib/monads/nondet`. `Monads` is also a global session, although a much smaller one, so I can live with or without the `lib` part of the tag. The suffix is useful here though, I often do want to disambiguate which monad I am working on and it's easier to just put it in the tag instead of trying to organically mention it in the header.

docs/commit-messages.md
lsf37: you're right, single `s` is the way to go. Feels weird to me, but maybe it's from `cussing` :-)

docs/commit-messages.md
lsf37: I was about to remove the `proofs` tag and standardise on `proof`, but now I'm not so sure any more. We (ok, mostly I) use that tag when many proofs are affected and it doesn't make sense to enumerate them all. `proofs` sounds better to me for that. Any opinions?

Xaphiosis: I prefer `proof` because it's shorter, and in my mind stands for `proof update`, so I'd vote for `proof`. Note: if someone's mind thinks `update proofs` they'll come to the opposite conclusion.

docs/commit-messages.md
Xaphiosis: Right, word_lib is an odd one, because it's synced with AFP ... now that our monads are going upstream, they might hit similar naming awkwardness, since we'd want to mark what affects upstream.

docs/commit-messages.md
Xaphiosis: The slashes are visually a bit confusing, because they represent both paths and `or`. I guess in tags people shouldn't be saying `or`.
I'm fine with an exception for clib, even though it's physically placed under lib.

For consideration: as per my comment on `word_lib`, now that monads are getting upstreamed it might make sense to also make an exception for `monads` in that case. There is a bunch of work upcoming for concurrency, and so we'll be seeing `lib/monads/nondet` a fair bit, which could become `monads/nondet`.

docs/commit-messages.md
```diff
@@ -0,0 +1,378 @@
...
+- Use imperative voice (e.g. `proofs: rename lemma foo to bar`)
+- The header should be a short summary, focusing on "what"
+- The body should explain what is going on in more detail (also in imperative
+  voice), but much more importantly *why* it is going on (is `bar` more
+  consistent than `foo`? Is the name `foo` needed for something else? Does `bar`
+  just better describe what is going on?).
+- You are trying to explain things to your future self or a future colleague
+  5-10 years from now. You can assume knowledge of the repository in general,
+  but you should not assume specific context that is obvious to you right now,
+  but that will not be known to a different person 5 years from now.
```
Xaphiosis, line 70: I don't know if more emphasis is needed on not being lazy and/or doing header-only commits because they're "obvious". In theory yes, in practice would it really help ...

docs/commit-messages.md
```diff
@@ -0,0 +1,378 @@
...
+`clib+crefine` or `lib+proof`.
+
+If something is specific to an architecture we preface the tag with the
+architecture, e.g. `arm refine:` or `aarch64 aspec+ainvs:`. The current
+architecture tags are: `arm`, `arm-hyp`, `x64`, `riscv`, `aarch64`.
+Please use these spellings only.
+
+More tag examples:
+
+- `trivial`: for small trivial changes like fixing a typo
```
Xaphiosis, line 112: trivial has the extra role that no proofs/specs should be changed, i.e. a guarantee that a proof run will have the same result, so in theory it doesn't need to be regression tested (and in the past, regression tests would ignore testing these, not sure what's the situation now)

docs/commit-messages.md
```diff
@@ -0,0 +1,378 @@
...
+output:
+
+```
+b3c6df48a clib: improve ccorres_While
+49ff8457f clib+crefine: improve and consolidate variants of ccorres_to_vcg
+8c433c085 clib: add some rules for hoarep
+82b954817 clib: improve ccorres_call_getter_setter
+8c59df449 lib/monads: remove more uses of _tac methods
+563232397 run_tests+proof: exclude SimplExportAndRefine for AARCH64
+1cce5b3ff proof: switch AArch64 quick_and_dirty from Refine to CRefine
```
Xaphiosis, line 152: I think this one identifies are relevant distinction, because it's <inside the build system> <change an aarch64 setting>, as opposed to `aarch64: change something in the build system`, the latter I would want to be seen as incorrect

docs/commit-messages.md
```diff
@@ -0,0 +1,378 @@
...
+4d0086567 lib/monads: add new Trace_* files to ROOT
+598e19dd6 lib/monads: coherent document structure
+4cbfb4ab0 lib/monads: minor style + warning cleanup
+b2dd5db6d lib/monads: fix document preparation issues
+03a045309 lib/monads: add AFP document setup
+d0bab9c79 misc/scripts: remove Darwin cpp wrapper
+```
+
+You can very quickly see that C verification has been active recently, that
+new tests were added, that AARCH64 proofs have been done, and that there was
```
Xaphiosis, line 167: refine proofs, not all proofs :)

docs/commit-messages.md
```diff
@@ -0,0 +1,378 @@
...
+
+If you're uncertain what tag to pick for a certain file or directory, the
+easiest way to figure it out is to do
+
+```sh
+git log --oneline <that-file>
+```
+
+Do your tags the same way people have done before. This will make the pattern
+consistent and should be reasonably easy to read even if it's not perfect.
```
Xaphiosis, line 186: I'd also say if you're not sure, take a look at a few commits, because sometimes someone invented a new flavour in the most recent commit and you don't want to derail the train, but gently guide it back into balance.

docs/commit-messages.md
```diff
@@ -0,0 +1,378 @@
...
+arm-hyp design
+arm-hyp haskell
+arm-hyp haskell+refine
+arm-hyp machine
+arm-hyp refine
+asmrefine
+aspec
+aspec+access
+aspec+ainvs
+aspec+design+haskell
```
Xaphiosis, line 251: this points out historical order: `aspec < haskell/design < ainvs < refine < crefine`, with specs going before proofs, which I think should also be spelled out (did I miss that?)

Xaphiosis: Oh yeah, I missed that. Might be worth expanding a little bit?

docs/commit-messages.md
```diff
@@ -0,0 +1,378 @@
...
+
+For large directories that have logical sub parts, we use slashes, e.g.
+`lib/monads`. Not so much because the change is in that directory, but because
+we want to see that it's a library change and applies to the monad part of the
+library.
+
+If the change applies to many proofs, for instance large lemma renaming commits,
+we use tags such as `proof` and `spec`.
+
+We combine tags with `+` if a change applies to multiple parts, e.g.
```
Xaphiosis, line 102: There's another thing to consider on the line below, when to *not* combine lib and proof changes. I sometimes want to `git show` the commit to see what was the relevant library change, and not see every update that could go into a `proof: update for <whatever lib changes>`.
This is not a clear-cut line, because it means there is a broken commit between the two, but it's very useful and our proofs can't really be bisected anyway due to the horrible runtime. I'm interested in people's thoughts on this.

docs/commit-messages.md
```diff
@@ -0,0 +1,378 @@
...
+with the following exceptions and additions:
+
+- header can be max 78 characters
+- body is wrapped at 78 characters
+- we use tags in the header to indicate which part of the repository
+  the commit applies to
+
+We are using 78 for the header because of the tags, which take up some space. If
+you can manage to stay within 50 characters anyway, that is appreciated, but it's
+not always practical.
```
Xaphiosis, line 43: we also don't kick people out for wrapping at 72 for the body, which is the default on many editors (including mine)

Xaphiosis:
I think this is excellent, especially given the very short production timeframe (and no comma/just avalanche!
Added thoughts for discussion now that I had a sec, and I second Gerwin's opinion about intention and tagging above.

docs/commit-messages.md
lsf37: I'd be happy with that.

docs/commit-messages.md
lsf37: I'd leave this outside the commit message guide, because it's more a question of what commits one should make. If the proof upadte is small, I think combining them makes sense, but if the proof update is large I also like the split version better, because it enables you to figure out more easily what content has actually changed in the spec.

docs/commit-messages.md
lsf37: The paragraph is probably long enough. I don't think more emphasis would achieve much.

lsf37:
Have addressed the feedback and am planning to merge when the tests are through. Still happy to discuss further changes, the content is not set in stone.

PR #689, opened 2023-10-26
produce _def and _val theorems in value_type

lsf37:
Produce a `_def` theorem for the value type that provides direct symbolic access to the right-hand side of the `value_type` definition. This avoids having to unfold those terms in subsequent proofs.

The numeric value as as term is still available as `<type-name>_val`.

`value_type` is now restricted to types `nat` and `int` on the right-hand side (`int` being cast to `nat` automatically). Other types can still be used, but have to be cast to nat manually. It turns out that so far we have only needed `int` and `nat`, so this works out without changes to any of the `value_type` definitions.

### DIFF
--- a/lib/Value_Type.thy
+++ b/lib/Value_Type.thy
@@ -10,7 +10,7 @@ keywords "value_type" :: thy_decl
 begin
 
 (*
-   Define a type synonym from a term that evaluates to a numeral.
+   Define a type synonym from a term of type nat or int that evaluates to a (positive) numeral.
 
    Examples:
 
@@ -41,6 +41,8 @@ fun force_nat_numeral (Const (@{const_name numeral}, Type ("fun", [num, _])) $ n
   | force_nat_numeral (Const (@{const_name "Groups.zero"}, _)) = @{term "0::nat"}
   | force_nat_numeral t = raise TERM ("force_nat_numeral: number expected", [t])
 
+fun cast_to_nat t = if type_of t = @{typ int} then @{term nat} $ t else t
+
 fun make_type binding v lthy =
   let
     val n = case get_term_numeral v of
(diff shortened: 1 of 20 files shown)


lsf37:
I haven't yet used the new `_def` theorems to eliminate previous manual instance of that, because my first search&replace attempt turned out to be too aggressive. I might add another commit to this PR for that, but the current state at least works and is worth looking at.

lib/Value_Type.thy
```diff
@@ -41,6 +41,8 @@ fun force_nat_numeral (Const (@{const_name numeral}, Type ("fun", [num, _])) $ n
   | force_nat_numeral (Const (@{const_name "Groups.zero"}, _)) = @{term "0::nat"}
   | force_nat_numeral t = raise TERM ("force_nat_numeral: number expected", [t])
 
+fun cast_to_nat t = if type_of t = @{typ int} then @{term nat} $ t else t
```
Xaphiosis, line 44: This has an interesting implication for negative numbers. What happens there? Maybe we should have the comment above say "a positive numeral"?

lib/test/Value_Type_Test.thy
```diff
@@ -5,11 +5,13 @@
...
 theory Value_Type_Test
-imports Lib.Value_Type
+  imports
+    Lib.Value_Type
+    "Word_Lib.WordSetup"
 begin
 
 (*
-   Define a type synonym from a term that evaluates to a numeral.
+   Define a type synonym from a term of type nat or int that evaluates to a numeral.
```
Xaphiosis, line 14: same comment as the comment about negative values

lib/Value_Type.thy
lsf37: I don't think a negative number is a numeral (it's unary minus applied to a numeral), so it'll likely just throw an exception in the code that forces it to be a numeral.

lib/test/Value_Type_Test.thy
lsf37: Happy to add that

Xaphiosis:
I think this is a solid start (going via _val for now). Content looks good. For `aarch64` branch, we'll have to do the update as well. Maybe after pre-VSpace_C, but before VSpace_C. Thoughts?

Commits:
* `lib: produce _def and _val thes in value_type` : thms?
* is it worth repeating that it's for nat/int in the commit message?
* `proof: proof updates for value_type _def -> _val` - you can probably skip the second `proof` here
* `Rename _def into _val occurrences for value_type name. Does not make
full use of value_type _def theorems yet.` - this is hard to understand, perhaps the word "generated" could make an appearance to help? "Rename occurrences of _def to _val for value_type-generated names." ?

Xaphiosis (approved):
@michaelmcinerney should possibly be aware this is in the queue

lsf37:
> I think this is a solid start (going via _val for now). Content looks good. For `aarch64` branch, we'll have to do the update as well. Maybe after pre-VSpace_C, but before VSpace_C. Thoughts?

Let's not block anything on this one. I'm happy to do the rebase whenever it goes in and the aarch64 branch is ready for it.
 
> Commits:
> 
> * `lib: produce _def and _val thes in value_type` : thms?
> * is it worth repeating that it's for nat/int in the commit message?
> * `proof: proof updates for value_type _def -> _val` - you can probably skip the second `proof` here
> * `Rename _def into _val occurrences for value_type name. Does not make full use of value_type _def theorems yet.` - this is hard to understand, perhaps the word "generated" could make an appearance to help? "Rename occurrences of _def to _val for value_type-generated names." ?

Thanks, will go through these

* [ ] `lib: produce _def and _val thes in value_type` : thms
* [ ] repeat that it's for nat/int in the commit message
* [ ] `proof: proof updates for value_type _def -> _val` - skip the second `proof`
* [ ] `Rename _def into _val occurrences for value_type name. Does not make full use of value_type _def theorems yet.` -  "Rename occurrences of _def to _val for value_type-generated names." ?

lib/Value_Type.thy
```diff
@@ -51,12 +53,31 @@ fun make_type binding v lthy =
     lthy |> Typedecl.abbrev (binding, [], Mixfix.NoSyn) typ |> #2
   end
 
-fun make_def binding v lthy =
+(* From method eval in HOL.thy: *)
+fun eval_tac ctxt =
```
corlewis, line 57: Very minor, but maybe make it clear that this is directly copied?
```suggestion
(* Copied from method eval in HOL.thy: *)
fun eval_tac ctxt =
```

lib/Value_Type.thy
```diff
@@ -51,12 +53,31 @@ fun make_type binding v lthy =
...
+  let val conv = Code_Runtime.dynamic_holds_conv
+  in
+    CONVERSION (Conv.params_conv ~1 (Conv.concl_conv ~1 o conv) ctxt) THEN'
+    resolve_tac ctxt [TrueI]
+  end
+
+(* This produces two theorems: one symbolic _def theorem and one numeric _val theorem.
+   The _def theorem is a definition, via Specification.definition.
+   The _val theorem is proved from that definition using "eval_tac" via the code generator. *)
+fun make_def binding t v lthy =
```
corlewis, line 67: Just as minor, but should this be `make_defs` now?

corlewis (approved):
I haven't used `value_type` myself so can't comment on the practicalities of the change, but this looks good to me.

As another nitpick; in the first commit message - "The numeric value as 'a' term"

lib/Value_Type.thy
lsf37: Have added "positive" for more clarity.

PR #695, opened 2023-11-07
Sporadic implies active

michaelmcinerney:
Please see the commit messages for what's here.

Test with https://github.com/seL4/seL4/pull/1127

### DIFF
--- a/proof/invariant-abstract/ARM/ArchDetSchedAux_AI.thy
+++ b/proof/invariant-abstract/ARM/ArchDetSchedAux_AI.thy
@@ -274,10 +274,18 @@ lemma perform_asid_control_invocation_pred_map_sc_refill_cfgs_of:
   by (wpsimp wp: delete_objects_pred_map_sc_refill_cfgs_of
            comb: hoare_drop_imp)
 
-lemma perform_asid_control_invocation_inactive_implies_zero_budget:
+lemma perform_asid_control_invocation_implies_zero_budget:
   "perform_asid_control_invocation aci
-   \<lbrace>\<lambda>s. pred_map inactive_scrc (sc_refill_cfgs_of s) p
+   \<lbrace>\<lambda>s. pred_map Q (sc_refill_cfgs_of s) p
         \<longrightarrow> pred_map (\<lambda>cfg. scrc_budget cfg = 0) (sc_refill_cfgs_of s) p\<rbrace>"
+  unfolding perform_asid_control_invocation_def
+  by (wpsimp wp: delete_objects_pred_map_sc_refill_cfgs_of
+           comb: hoare_drop_imp)
+
+lemma perform_asid_control_invocation_sporadic_implies:
+  "perform_asid_control_invocation aci
+   \<lbrace>\<lambda>s. pred_map (\<lambda>cfg. scrc_sporadic cfg) (sc_refill_cfgs_of s) p
+        \<longrightarrow> pred_map Q (sc_refill_cfgs_of s) p\<rbrace>"
   unfolding perform_asid_control_invocation_def
   by (wpsimp wp: delete_objects_pred_map_sc_refill_cfgs_of
            comb: hoare_drop_imp)
@@ -303,7 +311,8 @@ lemma perform_asid_control_invocation_valid_sched:
                                    perform_asid_control_invocation_sc_at_pred_n
                                    perform_asid_control_invocation_valid_idle
                                    perform_asid_control_invocation_pred_map_sc_refill_cfgs_of
-                                   perform_asid_control_invocation_inactive_implies_zero_budget
+                                   perform_asid_control_invocation_implies_zero_budget
+                                   perform_asid_control_invocation_sporadic_implies
                                    hoare_vcg_all_lift
                              simp: ipc_queued_thread_state_live live_sc_def)+
   done
--- a/proof/invariant-abstract/DetSchedAux_AI.thy
+++ b/proof/invariant-abstract/DetSchedAux_AI.thy
@@ -258,10 +258,21 @@ lemma retype_region_active_sc_props[wp]:
   apply (clarsimp simp: active_sc_def default_sched_context_def)
   done
 
-lemma retype_region_inactive_implies_zero_budget[wp]:
+lemma retype_region_implies_zero_budget[wp]:
   "retype_region ptr numObjects o_bits type dev
-   \<lbrace>\<lambda>s. pred_map inactive_scrc (sc_refill_cfgs_of s) p
+   \<lbrace>\<lambda>s. pred_map Q (sc_refill_cfgs_of s) p
         \<longrightarrow> pred_map (\<lambda>cfg. scrc_budget cfg = 0) (sc_refill_cfgs_of s) p\<rbrace>"
+  unfolding retype_region_def
+  apply wp
+  apply (clarsimp simp del: fun_upd_apply simp add: vs_all_heap_simps foldr_fun_upd_value)
+  apply (case_tac type; simp add: default_object_def)
+  apply (clarsimp simp: active_sc_def default_sched_context_def)
+  done
+
+lemma retype_region_sporadic_implies[wp]:
+  "retype_region ptr numObjects o_bits type dev
+   \<lbrace>\<lambda>s. pred_map (\<lambda>cfg. scrc_sporadic cfg) (sc_refill_cfgs_of s) p
+        \<longrightarrow> pred_map Q (sc_refill_cfgs_of s) p\<rbrace>"
   unfolding retype_region_def
   apply wp
   apply (clarsimp simp del: fun_upd_apply simp add: vs_all_heap_simps foldr_fun_upd_value)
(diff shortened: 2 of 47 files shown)


proof/refine/ARM/Schedule_R.thy
```diff
@@ -3920,6 +3920,18 @@ lemma refillUnblockCheck_corres:
       apply (clarsimp simp: obj_at_simps valid_refills'_def opt_map_red opt_pred_def)
   done
 
+lemma sporadic_implies_active_cross:
+  "\<lbrakk>(s, s') \<in> state_relation; active_scs_valid s; sc_at scPtr s; ko_at' sc scPtr s';
+    scSporadic sc\<rbrakk>
+   \<Longrightarrow> is_active_sc' scPtr s'"
```
lsf37, line 3926: :+1:

lsf37 (approved):
This is great. And special thanks for the informative commit message :-)

proof/invariant-abstract/ARM/ArchDetSchedAux_AI.thy
```diff
@@ -274,14 +274,22 @@ lemma perform_asid_control_invocation_pred_map_sc_refill_cfgs_of:
   by (wpsimp wp: delete_objects_pred_map_sc_refill_cfgs_of
            comb: hoare_drop_imp)
 
-lemma perform_asid_control_invocation_inactive_implies_zero_budget:
+lemma perform_asid_control_invocation_implies_zero_budget:
   "perform_asid_control_invocation aci
-   \<lbrace>\<lambda>s. pred_map inactive_scrc (sc_refill_cfgs_of s) p
+   \<lbrace>\<lambda>s. pred_map Q (sc_refill_cfgs_of s) p
         \<longrightarrow> pred_map (\<lambda>cfg. scrc_budget cfg = 0) (sc_refill_cfgs_of s) p\<rbrace>"
```
lsf37, line 280: Not suggesting to change these, just out of curiosity: would it make sense to pull the implication into the pred_map? We don't seem to be doing this in other proofs either, but it looks shorter to me (and equivalent).

michaelmcinerney: Yeah I think we could do that, and without too much of a change. Maybe it would make the lifting rules a little more annoying to deal with.

corlewis: I'm definitely not suggesting changing anything at this point, but does us doing this for quite a few of our conditions indicate that we actually want finer-grained projections?

michaelmcinerney: Oh when I saw Gerwin's comment, I thought he was referring to the statement of the new conjunct in `active_scs_valid`.

I didn't pay an awful lot of attention to these lemmas for `retype_region` etc. I think they're basically saying "if an active sched contexts exists after some deletion process occurred, then we didn't actually touch it" or "if we did retype it, it's now a default object". I am having a better look now to see if they can be improved or generalised.

@corlewis what sort of finger-grained projections were you thinking of? I think the `sc_refill_cfgs_of` projection works pretty well. It was used an awful lot for `valid_refills`

proof/refine/RISCV64/Syscall_R.thy
```diff
@@ -1011,36 +1011,36 @@ crunches refillNew, refillUpdate, commitTime
   and ex_nonz_cap_to'[wp]: "ex_nonz_cap_to' ptr"
   (simp: crunch_simps wp: crunch_wps)
 
+lemma scSBadge_update_invs'[wp]:
+  "updateSchedContext scPtr (scBadge_update f) \<lbrace>invs'\<rbrace>"
+  apply (wpsimp wp: updateSchedContext_invs')
+  apply (intro conjI impI allI)
+    apply (erule (1) live_sc'_ko_ex_nonz_cap_to')
```
Xaphiosis, line 1018: Is this erule very sensitive to things not being clarsimped before? Or maybe it's not in simp-normal form? Otherwise it looks tempting to give everything after the wpsimp to a fastforce. (similar temptation for `scSporadic_update_invs'`

proof/invariant-abstract/DetSchedAux_AI.thy
```diff
@@ -665,11 +665,17 @@ lemma valid_sched_tcb_state_preservation_gen:
     "\<And>P. \<lbrace>\<lambda>s. (\<forall>p. pred_map active_scrc (sc_refill_cfgs_of s) p \<longrightarrow> pred_map P (sc_refill_cfgs_of s) p) \<and> I s\<rbrace>
             f \<lbrace>\<lambda>rv s. \<forall>p. pred_map active_scrc (sc_refill_cfgs_of s) p \<longrightarrow> pred_map P (sc_refill_cfgs_of s) p\<rbrace>"
   assumes sc_refill_cfg3:
-    "\<And>P. \<lbrace>\<lambda>s. (\<forall>p. pred_map (\<lambda>cfg. \<not> active_scrc cfg) (sc_refill_cfgs_of s) p
+    "\<lbrace>\<lambda>s. (\<forall>p. pred_map (\<lambda>cfg. \<not> active_scrc cfg) (sc_refill_cfgs_of s) p
```
Xaphiosis, line 668: maybe I'm confused by the diff, but is the implication on the next line now incorrectly indented?

spec/haskell/src/SEL4/Object/SchedContext.lhs
```diff
@@ -525,6 +525,7 @@ This module uses the C preprocessor to select a target architecture.
 > schedContextZeroRefillMax :: PPtr SchedContext -> Kernel ()
 > schedContextZeroRefillMax scPtr = do
 >     updateSchedContext scPtr $ (\sc -> sc { scRefillMax = 0 })
+>     updateSchedContext scPtr $ (\sc -> sc { scSporadic = False })
```
Xaphiosis, line 528: This is a bit odd, we now have `schedContextZeroRefillMax` setting `scSporadic`, which minimises turbulence in the rest of the proof, but doesn't seem to make a lot of sense to read. This file is a bit sparse on commentary, which doesn't make it easier to grasp the meaning for the uninitiated.

Xaphiosis:
Nice work, this looks like an improvement, and the commit messages are more informative than the comments in the specs, thus helping a lot with understanding what you did in this PR!

Xaphiosis (approved):

proof/invariant-abstract/ARM/ArchDetSchedAux_AI.thy
michaelmcinerney: I did have a quick look to see whether it might be straightforward to get rid of the doubling up of `pred_map` in these sorts of implications. It's quite annoying, actually. I think these rules would work fairly well with `hoare_vcg_imp_lift'` or `hoare_drop_imp`, especially with the schematic predicates that are in some of these rules.

As for whether we should get rid of this sort of thing in the statement of the invariants, I would say that that would make sense, too. But I'm growing to rather dislike `pred_map` and its many often contorted variants. My vote would be to rewrite a lot of these invariants with `opt_map` in some way.

Should I leave all this as is for now?

proof/invariant-abstract/DetSchedAux_AI.thy
michaelmcinerney: Ah, this was incorrect. Fixed, thanks

spec/haskell/src/SEL4/Object/SchedContext.lhs
michaelmcinerney: This function has drifted over the last few years, so the name is no longer a good one. This function is now probably better thought of now as zeroing (in a general sense) all of the fields for the "sporadic" parts of the sched context. But I'm not really sure what "sporadic" means any more...

Would it be better to call it `sched_context_zero_sporadic_fields` or something? I can't think of anything better

Xaphiosis: Yeah, your comment and the better name makes more sense (I'm guessing you'd want a `camelCase` name since this is Haskell). If it's inconvenient to do in this PR, can maybe add a rename to another day, but the helpful comment would still be welcome.

michaelmcinerney: Actually, maybe `sched_context_zero_sporadic_fields` isn't such a good name: presumably we could have the `sc_sporadic` field be `False`, so that sched context isn't "sporadic", but we'd still like to call this function to make the sc inactive. So it really does seems like it makes it inactive (which is why it was called `zero_refill_max` to begin with) and then we've just got a bunch of other things we'd like to zero out in order to keep the invariants happy. Still not sure what to call it, then.

I think it should be renamed, but considering that the name was off before this work was started, it would be better to redo the naming in another PR.

And yes, `camelCase` for the Haskell, but we do have the analogue of this function in the abstract spec as well.

lsf37: I think making it inactive is the right view -- that requires resetting the refills and the sporadic flag at the same time, because it's not properly inactive if the sporadic flag is still set.

proof/refine/RISCV64/Syscall_R.thy
michaelmcinerney: I was able to compress this into one `fastforce` after the `wpsimp` in both proofs.

PR #696, opened 2023-11-14
More progress on completing the trace monad rule set

corlewis:
This extends the rule sets for the various monad properties (`no_trace`, `no_fail`, etc.) to cover all of the monad primitives that we have. It does not touch the RG rule set.

### DIFF
--- a/lib/Monads/trace/Trace_Monad.thy
+++ b/lib/Monads/trace/Trace_Monad.thy
@@ -44,7 +44,6 @@ text \<open>
   pair of result and state when the computation did not fail.\<close>
 type_synonym ('s, 'a) tmonad = "'s \<Rightarrow> ((tmid \<times> 's) list \<times> ('s, 'a) tmres) set"
 
-
 text \<open>
   Print the type @{typ "('s,'a) tmonad"} instead of its unwieldy expansion.
   Needs an AST translation in code, because it needs to check that the state variable
@@ -67,6 +66,17 @@ print_ast_translation \<open>
 text \<open>Returns monad results, ignoring failures and traces.\<close>
 definition mres :: "((tmid \<times> 's) list \<times> ('s, 'a) tmres) set \<Rightarrow> ('a \<times> 's) set" where
   "mres r = Result -` (snd ` r)"
+
+text \<open>True if the monad has a computation resulting in Failed.\<close>
+definition failed :: "((tmid \<times> 's) list \<times> ('s, 'a) tmres) set \<Rightarrow> bool" where
+  "failed r \<equiv> Failed \<in> snd ` r"
+
+lemma failed_simps[simp]:
+  "failed {(x, y)} = (y = Failed)"
+  "failed (r \<union> r') = (failed r \<or> failed r')"
+  "\<not> failed {}"
+  by (auto simp: failed_def)
+
 
 text \<open>
   The definition of fundamental monad functions @{text return} and
--- a/lib/Monads/trace/Trace_No_Trace.thy
+++ b/lib/Monads/trace/Trace_No_Trace.thy
@@ -44,21 +44,17 @@ lemma no_trace_is_triple[wp_trip]:
 
 subsection \<open>Rules\<close>
 
-lemma no_trace_prim:
-  "no_trace get"
-  "no_trace (put s)"
-  "no_trace (modify f)"
-  "no_trace (return v)"
-  "no_trace fail"
-  "no_trace (returnOk v)"
-  by (simp_all add: no_trace_def get_def put_def modify_def bind_def
-                    return_def select_def fail_def returnOk_def)
+(*
+  Collect generic no_trace lemmas here:
+   - naming convention is no_trace_NAME.
+   - add lemmas with assumptions to [no_trace_cond] set
+   - add lemmas without assumption to [no_trace_terminal] set
+*)
 
-lemma no_trace_select:
-  "no_trace (select S)"
-  by (clarsimp simp add: no_trace_def select_def)
+named_theorems no_trace_cond
+named_theorems no_trace_terminal
 
-lemma no_trace_bind:
+lemma no_trace_bind[wp_split]:
   "\<lbrakk>no_trace f; \<forall>rv. no_trace (g rv)\<rbrakk>
    \<Longrightarrow> no_trace (do rv \<leftarrow> f; g rv od)"
   apply (subst no_trace_def)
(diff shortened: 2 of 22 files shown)


lib/Monads/trace/Trace_Prefix_Closed.thy
```diff
@@ -0,0 +1,301 @@
...
+  apply (drule(1) prefix_closedD)+
+  apply fastforce
+  done
+
+context begin
+(* We want to enforce that prefix_closed_terminal only contains lemmas that have no
+   assumptions. The following thm statement should fail if this is not true. *)
+private lemmas check_no_assumptions_internal = iffD1[OF refl, where P="prefix_closed f" for f]
+thm prefix_closed_terminal[atomized, THEN check_no_assumptions_internal]
+end
```
corlewis, line 295: Huh, I forgot that I added this check here. Are we ok with this approach, and if so should I add it to the other monad files that use the same attribute pattern?

It won't protect us from rules being incorrectly added to the sets past this point, but it should catch any mistakes we make adding new rules in this file.

lib/Monads/trace/Trace_Monad.thy
```diff
@@ -68,6 +67,17 @@ text \<open>Returns monad results, ignoring failures and traces.\<close>
 definition mres :: "((tmid \<times> 's) list \<times> ('s, 'a) tmres) set \<Rightarrow> ('a \<times> 's) set" where
   "mres r = Result -` (snd ` r)"
 
+text \<open>True if the monad has a computation resulting in Failed.\<close>
+definition failed :: "((tmid \<times> 's) list \<times> ('s, 'a) tmres) set \<Rightarrow> bool" where
+  "failed r \<equiv> Failed \<in> snd ` r"
```
lsf37, line 72: I like it. I'm almost tempted to try it out on the nondet monad as well. But let's resist temptation for now, it'd lead to much more change on that side.

lib/Monads/trace/Trace_No_Trace.thy
```diff
@@ -44,34 +44,318 @@ lemma no_trace_is_triple[wp_trip]:
...
+
+lemma no_trace_apply[no_trace_cond]:
+  "no_trace (f (g x)) \<Longrightarrow> no_trace (f $ g x)"
+  by simp
+
+\<comment> \<open>FIXME: make proof nicer\<close>
+lemma no_trace_liftM_eq[simp]:
+  "no_trace (liftM f m) = no_trace m"
+  apply (clarsimp simp: no_trace_def bind_def liftM_def return_def split_def
+                 split: tmres.split_asm)
```
lsf37, line 138: Could this proof work by combining `no_trace` results about `bind` and `return` instead of unfolding? (Doesn't have to be in this PR)

lib/Monads/trace/Trace_Prefix_Closed.thy
lsf37: That's a neat idea! I wonder if there is a simpler version, but I can't come up with anything.

Happy to also add this in the other theories that could benefit from similar checks.

lib/Monads/trace/Trace_No_Trace.thy
```diff
@@ -44,34 +44,318 @@ lemma no_trace_is_triple[wp_trip]:
...
+
+lemma no_trace_andM[no_trace_cond]:
+  "\<lbrakk> no_trace A; no_trace B \<rbrakk> \<Longrightarrow> no_trace (andM A B)"
+  unfolding andM_def
+  by (wpsimp wp: no_trace_terminal no_trace_cond)
+
+\<comment> \<open>FIXME: This is true but is it useful to even state?
+          The parallel composition of traceless functions doesn't make much sense\<close>
+lemma no_trace_parallel[no_trace_cond]:
+  "\<lbrakk> no_trace f; no_trace g \<rbrakk> \<Longrightarrow> no_trace (parallel f g)"
```
lsf37, line 354: I think it's useful to have -- you want trivial goals to go away automatically, and this will enable that. It's unlikely to occur, but I could imagine it popping up in some kind of contradiction or things like that.

lib/Monads/trace/Trace_Prefix_Closed.thy
```diff
@@ -0,0 +1,301 @@
...
+  Collect generic prefix_closed lemmas here:
+   - naming convention is prefix_closed_NAME.
+   - add lemmas with assumptions to [prefix_closed_cond] set
+   - add lemmas without assumption to [prefix_closed_terminal] set
+*)
+
+named_theorems prefix_closed_terminal
+named_theorems prefix_closed_cond
+
+lemmas [prefix_closed_terminal] = no_trace_terminal[THEN no_trace_prefix_closed]
```
lsf37, line 58: Neat, that saves a lot of typing.

lib/Monads/trace/Trace_Prefix_Closed.thy
```diff
@@ -0,0 +1,301 @@
...
+
+lemma prefix_closed_bind[wp_split]:
+  "\<lbrakk>prefix_closed f; \<And>x. prefix_closed (g x)\<rbrakk> \<Longrightarrow> prefix_closed (f >>= g)"
+  apply (subst prefix_closed_def, clarsimp simp: bind_def)
+  apply (auto simp: Cons_eq_append_conv split: tmres.split_asm
+             dest!: prefix_closedD[rotated];
+         fastforce elim: rev_bexI)
```
lsf37, line 65: Sometimes it helps to give auto deeper depth parameters e.g. `auto (6,3) simp: ..`. Not actually sure if that is less dodgy than `auto ..; fastforce ..` or more, because people tend to not know these, but I wanted to put it out there. (Also not sure if it'll actually help -- the first number is the depth given to `blast` inside `auto`, default is 4, and the second is another depth parameter for another tactic that I don't actually know (default = 2)).

lsf37 (approved):
I'm glad we're doing this. These rules would be so much pain to discover and add incrementally during the proofs, and it would lead to a complete mess.

lib/Monads/trace/Trace_Prefix_Closed.thy
lsf37: To appease the thy linter, we could write `private lemmas check = ..` instead of `thm ..`

lib/Monads/trace/Trace_Prefix_Closed.thy
lsf37: (The other one to try, which you might already have done anyway is `force` instead of `auto`)

lib/Monads/trace/Trace_Prefix_Closed.thy
corlewis: If you're curious, this isn't completely new, I remembered that I had done something very similar in [rt](https://github.com/seL4/l4v/blob/rt/proof/refine/ARM/Invariants_H.thy#L3017).

For your second point, I think I have a very small preference for `thm ..` over `private lemmas check = ..`. It's pretty easy to ignore the thy linter failure here and it shouldn't reoccur in future PRs, as it feels very unlikely that these specific lines will need updating. On the other hand, it's minor but `private lemmas` aren't completely private. They can't actually be used anywhere else but they still pollute the namespace and can lead to mildly confusing `find_theorems` results.

lib/Monads/trace/Trace_No_Trace.thy
corlewis: Unfortunately not, at least not without new `bind` and `return` results. That would work for the following `no_trace_liftM` lemma but not for this one with its equality.

To be honest, I probably would have skipped this equality and just done `no_trace_liftM`, except for the fact that I was copying a lot of the structure of this file from `Trace_No_Fail` and `Trace_Empty_Fail` which both have `liftM_eq` equivalents.

lib/Monads/trace/Trace_No_Trace.thy
corlewis: Good point. I'll remove the FIXME then but might leave a comment since I imagine we might otherwise go through the same conversation at some point in the future..

lib/Monads/trace/Trace_Prefix_Closed.thy
corlewis: Yeah, I was pretty happy that this worked.

lib/Monads/trace/Trace_Prefix_Closed.thy
corlewis: I'll give both of those a go, it would be nice to avoid `auto ..; fastforce ..`

corlewis: It turns out that `auto 4 4 ..` does work, so I guess that just leaves the question of which is less dodgy.

lib/Monads/trace/Trace_Prefix_Closed.thy
lsf37: Ok, that is a valid point. Let's stick with `thm`, the linter won't be seeing this file very often.

lib/Monads/trace/Trace_No_Trace.thy
lsf37: Ok, no worries.

lib/Monads/trace/Trace_Prefix_Closed.thy
lsf37: I would say `auto 4 4` is marginally less dodgy than `auto; fastforce`, but it's a toss-up. Happy to leave it up to your taste, you're the author :-)

PR #697, opened 2023-12-04
Further improvements to ccorres_While

michaelmcinerney:
Please see the commit message.


michaelmcinerney:
I suppose I should update the commit message to refer not to "the setter" but rather to "`cond`", given the change I made to the rule (if people are happy with the `cond` naming)

lsf37 (approved):
Nice, I like `cond` better than `setter`, and the comment is good. Should give is enough information to figure out what is going on when we encounter this the next time.

Xaphiosis (approved):
Thumbs up from me. It's a tricky rule, I think you've done a good job making it as understandable as possible.

PR #701, opened 2023-12-10
provide a NUM_DOMAINS test override

lsf37:
Setting the environment variable `INPUT_NUM_DOMAINS` will cause the build to override the `KernelNumDomains` setting in the config file with the provided setting. `INPUT_NUM_DOMAINS` is the name we get from the GitHub workflow input with the name `NUM_DOMAINS`.

The second commit switches the test to the following `NUM_DOMAINS` settings:

- PRs: default as in config file, no matrix
- push to master: with NUM_DOMAINS = 1 and default (= '')
- weekly test: with NUM_DOMAINS = 1, 7, and default

The default is 16 in the current config files. 1 leads to structural code changes is the setting most likely to break. 7 is for checking that the proofs also work with a value that is not a power of 2.

This needs seL4/ci-actions#293 to be merged first.

### DIFF
--- a/.github/workflows/proof-deploy.yml
+++ b/.github/workflows/proof-deploy.yml
@@ -37,6 +37,7 @@ jobs:
       fail-fast: false
       matrix:
         arch: [ARM, ARM_HYP, AARCH64, RISCV64, X64]
+        num_domains: ['1', '']
     # test only most recent push:
     concurrency: l4v-regression-${{ github.ref }}-${{ strategy.job-index }}
     steps:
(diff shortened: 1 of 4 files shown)


Xaphiosis:
I'm somewhat confused, where does NUM_DOMAINS and INPUT_NUM_DOMAINS meet up? At the github level, NUM_DOMAINS is set, but then in scripts, INPUT_NUM_DOMAINS is read.

Xaphiosis (approved):

lsf37:
> I'm somewhat confused, where does NUM_DOMAINS and INPUT_NUM_DOMAINS meet up? At the github level, NUM_DOMAINS is set, but then in scripts, INPUT_NUM_DOMAINS is read.

This is a [GitHub thing](): workflow input names `<n>` are mapped to environment variables `INPUT_<N>` on the test runner machine.

PR #703, opened 2024-01-08
aarch64 cspec+crefine: pt_type ghost state + table array sizes

lsf37:
- add ghost state corresponding to gsPTTypes in Haskell and ASpec
- add ghost type comments
- style update for old definitions since we need to touch most of these
- define vs/pt_array_len for use in C annotations
- make NormalPT_T/VSRootPT_T names available for use in C annotations

Currently the vs/pt_array_len declarations introduce a dependency of CSpec on ASpec. We could remove that if we moved the definitions of pt_index_len and vs_index_bits higher, so that they become visible in ExecSpec.

This PR is required for merging seL4/seL4#1156

### DIFF
--- a/spec/cspec/AARCH64/Kernel_C.thy
+++ b/spec/cspec/AARCH64/Kernel_C.thy
@@ -18,6 +18,8 @@ context begin interpretation Arch .
 
 requalify_types
   machine_state
+  pt_array_len
+  vs_array_len
 
 end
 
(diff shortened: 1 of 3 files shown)


spec/ROOT
```diff
@@ -93,6 +93,7 @@ session CSpec in "cspec" = CKernel +
 
 session CKernel in "cspec/$L4V_ARCH" = CParser +
   sessions
+    "ASpec"
```
Xaphiosis, line 96: Does this dependency cost anything time-wise on other arches, or only if they actually pull in theories from ASpec?
There's a concurrency issue with CKernel normally being built in parallel with ASpec, and the xml files will probably try to do that, which could lead to unreliable results.

spec/cspec/AARCH64/Kernel_C.thy
```diff
@@ -6,6 +6,7 @@
 
 theory Kernel_C
 imports
+  "ASpec.Structures_A" (* FIXME AARCH64: needed for pt_index_len/vs_index_len -- maybe move these to spec/machine? *)
```
Xaphiosis, line 9: how difficult would this be to do? I've got mixed feelings about the "CKernel depends on ASpec" change.

spec/ROOT
lsf37: The other arches are unaffected time-wise, because they don't actually draw in any new theories, they only declare the session dependency. You're right with the concurrency issues, though, I hadn't thought about these. If ASpec was not in the path before and it needs regeneration of stuff, there might be interference or CRefine might not see the right version.

spec/cspec/AARCH64/Kernel_C.thy
lsf37: I'll give this a go. The plan was to clean this up anyway, might as well do it now.

lsf37:
Ok, new approach: the page table size and array types are now defined in `machine/Platforms.thy` and therefore available much earlier. Some of the sanity checks have to remain where they are, because `ptTranslationBits` isn't available yet in `machine`, but I think that's Ok.

I haven't fully checked all sessions yet, but CKernel works, and the PR checks should show the rest.

spec/machine/AARCH64/Platform.thy
```diff
@@ -101,5 +101,64 @@ definition irqVTimerEvent :: irq where
...
+
+text \<open>Sanity check:\<close>
+lemma length_vs_index_len[simp]:
+  "LENGTH(vs_index_len) = vs_index_bits"
+  by (simp add: len_of_vs_index_len type_definition.card[OF type_definition_vs_index_len])
+
+
+section \<open>C array sizes corresponding to page table sizes\<close>
+
+value_type pt_array_len = "(2::nat) ^ LENGTH(pt_index_len)"
```
Xaphiosis, line 159: silly question: why do we not have a `pt_index_bits` linked to `LENGTH(pt_index_len)`?

Xaphiosis:
> Ok, new approach: the page table size and array types are now defined in `machine/Platforms.thy` and therefore available much earlier. Some of the sanity checks have to remain where they are, because `ptTranslationBits` isn't available yet in `machine`, but I think that's Ok.
> 
> I haven't fully checked all sessions yet, but CKernel works, and the PR checks should show the rest.

I think this is a decent compromise, though the C array size part might be a surprise for people to see in `Platform.thy`. I'm not sure what we'll need to do if more of these are needed (e.g. in future platform port automation work), the lack of visibility of spec values when parsing the C kernel has been a source of grumpiness for a while.

Xaphiosis:
Oh also, you'll need to update the commit tags once this works now that it's changing machine. Or split that move out into a separate commit.

Xaphiosis (approved):

spec/machine/AARCH64/Platform.thy
lsf37: We have no `pt_index_bits` defined. `vs_index_bits` only exists because of the funky definition as a separate type with potentially the same size, but different name (so that we can tell pt/vs apart even on configs where they are equal).

lsf37: (And we can't use `LENGTH(vs_index_len)` because the code generator can't figure out the value with that manual definition)

PR #704, opened 2024-01-10
Rules for Lib

michaelmcinerney:


michaelmcinerney:
The errors should hopefully be fixed now: they all seemed to be caused by the reordering of the assumptions in `no_fail_bind`. The changes are all in the last commit, which I will squash into the appropriate commit, once approved

lib/Corres_UL.thy
```diff
@@ -481,6 +481,18 @@ lemma corres_if3:
                                     (if G then a else b) (if G' then c else d)"
   by simp
 
+lemma corres_if_strong:
+  "\<lbrakk>\<And>s s'. \<lbrakk>(s, s') \<in> sr; R s; R' s'\<rbrakk> \<Longrightarrow> G = G';
+    \<lbrakk>G; G'\<rbrakk> \<Longrightarrow> corres_underlying sr nf nf' r P P' a c;
+    \<lbrakk>\<not> G; \<not> G'\<rbrakk> \<Longrightarrow> corres_underlying sr nf nf' r Q Q' b d \<rbrakk>
+   \<Longrightarrow> corres_underlying sr nf nf' r
+         (R and (if G then P else Q)) (R' and (if G' then P' else Q'))
+         (if G then a else b) (if G' then c else d)"
```
lsf37, line 490: It feels like we have loads of variations of `if` rules already, but I don't think we have anything where both guard conditions are in the assumptions. I was initially going to suggest deriving some of the other variations from this rule, but the proofs are all one-liners anyway, so I'm fine with it as it is.

lib/Corres_UL.thy
```diff
@@ -740,6 +752,11 @@ lemma corres_assert_assume:
   by (auto simp: bind_def assert_def fail_def return_def
                  corres_underlying_def)
 
+lemma corres_assert_assume_l:
+  "corres_underlying sr nf nf' rrel P Q (f ()) g
+  \<Longrightarrow> corres_underlying sr nf nf' rrel (P and (\<lambda>s. P')) Q (assert P' >>= f) g"
```
lsf37, line 757: Thanks for moving this.

lib/ExtraCorres.thy
```diff
@@ -251,6 +249,88 @@ lemma hoare_from_abs_inv:
...
+    apply (erule nf)
+    done
+  show ?thesis
+  apply (rule corres_guard_imp)
+    apply (subst return_bind[symmetric], rule corres_split[OF P])
+      apply (rule z)
+     apply wp
+    apply (rule y)
+   apply simp+
+  done
```
lsf37, line 307: indent is off here, the `apply` should be +2

lib/ExtraCorres.thy
```diff
@@ -251,6 +249,88 @@ lemma hoare_from_abs_inv:
...
+qed
+
+lemmas corres_symb_exec_r_conj_ex_abs_forwards =
+  corres_symb_exec_r_conj_ex_abs[where P'=P' and Q'=P' for P', simplified]
+
+lemma gets_the_corres_ex_abs:
+ "\<lbrakk>no_ofail P a; no_ofail (P' and ex_abs_underlying sr P) b\<rbrakk> \<Longrightarrow>
+   corres_underlying sr False True r P P' (gets_the a) (gets_the b)
+   = (\<forall> s s'. P s \<and> P' s' \<and> (s, s') \<in> sr \<longrightarrow> r (the (a s)) (the (b s')))"
```
lsf37, line 316: Phrasing corres rules as equalities is usually dangerous, because `simp` doesn't play nice with the schematics we usually have in those goals. If it's important to have an equality in this one, I'd vote for providing an additional rule with `[THEN iffD2]` that can be used as an intro rule in the goal.

lsf37 (approved):
Looks good from my side!

lib/Corres_UL.thy
michaelmcinerney: I think this is the only rule for `if` that also allows us to use the assumptions here to show equality of `G` and `G'`

lib/ExtraCorres.thy
michaelmcinerney: It seems like `gets_the_corres` is used either with `subst`, or with a `[THEN iffD1/2]`. Are you wanting a `lemmas gets_the_corres'` or something that does these `[THEN iffD1/2]` things for us?

This is a generalisation of `gets_the_corres`, which we have on the `rt` branch, and which is one of the few rules which hasn't made its way over to `master`. Should I do anything with that rule, too? Should I put that into `master` too, as part of this PR?

lib/Corres_UL.thy
michaelmcinerney: Oh sorry, I misread your comment. We're saying the same thing, oops

lib/ExtraCorres.thy
lsf37: I would call this rule (with equality) `gets_the_corres_ex_abs'`, because it's not in standard corres rule form (which are basically all `intro` rules for `corres` with different patterns). And yes, my `gets_the_corres_ex_abs` would just be a `lemmas .. = gets_the_corres_ex_abs'[THEN iffD2]` (or iffD1, I can never remember the direction), so people don't have to do that themselves and don't get tempted to just chuck the equality into `simp`.

Bringing the `gets_the_corres` rule over is a good idea in any case, though. If this one here is more general, we should possibly derive the other one from it, unless the proofs are trivial.

lib/Corres_UL.thy
lsf37: All good, I should probably not even have said anything in the end, but I was too lazy to delete it again :-)

lib/ExtraCorres.thy
michaelmcinerney: I've pushed one commit for these things. `iffD2` was the right one to make the `corres_underlying` appear in the conclusion that that they're `intro` rules. And I do use the other direction for another lemma (`gets_the_corres_relation` just below) so yes, we should keep the equality too.

The proof for these was a lot of unfolding of basic definitions, so I did use the more general one to proof the less general one.

lib/ExtraCorres.thy
```diff
@@ -251,6 +249,100 @@ lemma hoare_from_abs_inv:
...
+  \<Longrightarrow> r (the (f s)) (the (f' s'))"
+  apply (rule_tac P=P and a=f and b=f' and P'=P'
+               in gets_the_corres_ex_abs'[THEN iffD1, rule_format];
+         fastforce?)
+  apply (rule no_ofail_gets_the_eq[THEN iffD2])
+  apply (fastforce intro: corres_u_nofail)
+  done
+
+
+\<comment> \<open>Some @{term corres_underlying }rules for @{term whileLoop}\<close>
```
corlewis, line 344: Very minor
```suggestion
\<comment> \<open>Some @{term corres_underlying} rules for @{term whileLoop}\<close>
```

PR #705, opened 2024-01-12
Suppress simplifier warnings in `corres_rewrite`

lsf37:
- add a method for setting context visibility to `false` for Eisabach methods
- use it in `simpl_rewrite` to suppress simplifier warnings in `ccorres_rewrite`

This should make `ccorres_rewrite` quite about `Collect_const` being in the simpset or not.


Xaphiosis (approved):
This might be also be useful in `ccorres_rewrite` and the duplicate simp rule warning

PR #717, opened 2024-02-18
pull Simpl update from AFP 2023

lsf37:
Pulls in license change and content update of the Simpl session from the AFP. The license changes from LGPL to BSD-3-Clause.

Flow-on breakage seems to be low, the main things that break are:

- the implementation of `block` has changed slightly -- to fully unfold a `block` (and therefore a `call`) you now also need `block_exn_def`
- the tuple `params`, a field of the `proc_info` type has an additional element, which means ML code accessing that needs to be updated accordingly to deal with triples instead of pair.


lsf37:
- [ ] commit msg of the first commit should mention `Simpl` not just `c-parser`.

corlewis (approved):
Looks good!

It's very minor, but I'm not sure that we gain anything from all of the Simpl updates being separate commits.

lsf37:
> Looks good!
> 
> It's very minor, but I'm not sure that we gain anything from all of the Simpl updates being separate commits.

We can probably squash all of them, yes. I mostly kept adding as I discovered failures, but then left it, because it's easy to see what is affected. I don't have a strong preference either way.

Xaphiosis (approved):
These are quite large diffs to SIMPL, it's interesting that there's not more fallout for us. Wonder if we'll get any improvements in the term printing speed at some point.

lsf37:
> These are quite large diffs to SIMPL, it's interesting that there's not more fallout for us. Wonder if we'll get any improvements in the term printing speed at some point.

I have not actually checked -- this might already contain some.

PR #719, opened 2024-02-19
monads/wp: update checks for when to trace wp

corlewis:
This changes it so that setting `wp_trace_instantiation` to true is enough to see the trace, instead of also needing to set `wp_trace`.


lsf37 (approved):

Xaphiosis (approved):
good idea

PR #720, opened 2024-02-20
Some more rules for `Lib`

michaelmcinerney:
Currently separated into many smaller commits, but I could consolidated them into just a few commits if that's preferred: one commit for reader monad lemmas, one for lists, then one for `CLib`, `Heap_List.thy`, `Corres_UL.thy`, and `MonadicRewrite.thy`

### DIFF
--- a/lib/Lib.thy
+++ b/lib/Lib.thy
@@ -1913,6 +1913,37 @@ lemma length_takeWhile_gt:
   apply simp
   done
 
+lemma map_fst_filter_zip[simp]:
+ "map fst (filter (\<lambda>(_, x). P x) (zip ls (map f ls))) = filter (\<lambda>x. P (f x)) ls"
+  by (induction ls; clarsimp)
+
+lemma map_fst_dropWhile_zip[simp]:
+ "map fst (dropWhile (\<lambda>(_, x). P x) (zip ls (map f ls))) = dropWhile (\<lambda>x. P (f x)) ls"
+  by (induction ls; clarsimp)
+
+lemma map_fst_takeWhile_zip[simp]:
+ "map fst (takeWhile (\<lambda>(_, x). P x) (zip ls (map f ls))) = takeWhile (\<lambda>x. P (f x)) ls"
+  by (induction ls; clarsimp)
+
+lemma suffix_tl:
+  "suffix xs ys \<Longrightarrow> suffix (tl xs) ys"
+  apply (clarsimp simp: suffix_def)
+  apply (metis append.assoc append_Cons append_Nil list.exhaust_sel list.sel(2))
+  done
+
+lemma nonempty_proper_suffix_split:
+  "\<lbrakk>sfx \<noteq> []; sfx \<noteq> queue; suffix sfx queue\<rbrakk> \<Longrightarrow> (\<exists>xs ys. xs \<noteq> [] \<and> queue = xs @ hd sfx # ys)"
+  apply (clarsimp simp: suffix_def)
+  apply (fastforce simp: neq_Nil_conv)
+  done
+
+lemma nonempty_proper_suffix_split_distinct:
+  "\<lbrakk>sfx \<noteq> []; sfx \<noteq> queue; suffix sfx queue; distinct queue\<rbrakk>
+   \<Longrightarrow> \<exists>xs ys. xs \<noteq> [] \<and> queue = xs @ hd sfx # ys \<and> hd queue \<noteq> hd sfx"
+  apply (frule (2) nonempty_proper_suffix_split)
+  apply (cases sfx; fastforce)
+  done
+
 lemma hd_drop_conv_nth2:
   "n < length xs \<Longrightarrow> hd (drop n xs) = xs ! n"
   by (rule hd_drop_conv_nth) clarsimp
@@ -2031,6 +2062,22 @@ lemma contrapos_imp:
 lemma filter_eq_If:
   "distinct xs \<Longrightarrow> filter (\<lambda>v. v = x) xs = (if x \<in> set xs then [x] else [])"
   by (induct xs) auto
+
+lemma filter_last_equals_butlast:
+  "\<lbrakk>distinct q; q \<noteq> []\<rbrakk> \<Longrightarrow> filter ((\<noteq>) (last q)) q = butlast q"
+  apply (induct q rule: length_induct)
+  apply (rename_tac list)
+  apply (case_tac list; simp)
+  apply (fastforce simp: filter_id_conv)
+  done
+
+lemma filter_middle_distinct:
+  "\<lbrakk>distinct q; q = ys @ t # zs\<rbrakk> \<Longrightarrow> filter ((\<noteq>) t) q = ys @ zs"
+  apply (induct q rule: length_induct)
+  apply (rename_tac list)
+  apply (case_tac list; simp)
+  apply (subst filter_True | fastforce)+
+  done
 
 lemma (in semigroup_add) foldl_assoc:
 shows "foldl (+) (x+y) zs = x + (foldl (+) y zs)"
@@ -2561,6 +2608,18 @@ lemma ranD:
 lemma fun_upd_swap:
   "a \<noteq> c \<Longrightarrow>  hp(c := d, a := b) = hp(a := b, c := d)"
   by fastforce
+
+lemma list_not_head:
+  "Suc 0 < length ls \<Longrightarrow> ls \<noteq> [hd ls]"
+  by (cases ls; clarsimp)
+
+lemma list_not_last:
+  "Suc 0 < length ls \<Longrightarrow> ls \<noteq> [last ls]"
+  by (cases ls; clarsimp)
+
+lemma length_tail_nonempty:
+  "Suc 0 < length list \<Longrightarrow> tl list \<noteq> []"
+  by (cases list, simp, simp)
 
 text \<open>Prevent clarsimp and others from creating Some from not None by folding this and unfolding
   again when safe.\<close>
--- a/lib/ListLibLemmas.thy
+++ b/lib/ListLibLemmas.thy
@@ -397,6 +397,20 @@ lemma set_list_insert_before:
   apply fastforce
   done
 
+lemma takeWhile_dropWhile_insert_list_before:
+  "\<lbrakk>distinct ls; sorted (map f ls); sfx \<noteq> []; suffix sfx ls;
+    \<forall>pfx v. pfx @ sfx = ls \<and> v \<in> set pfx \<longrightarrow> f v \<le> f new; f new < f (hd sfx)\<rbrakk> \<Longrightarrow>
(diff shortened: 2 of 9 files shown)


lib/MonadicRewrite.thy
```diff
@@ -760,6 +760,10 @@ lemma monadic_rewrite_corres_r:
 lemmas corres_gets_the_bind
   = monadic_rewrite_corres_l[OF monadic_rewrite_bind_head[OF monadic_rewrite_gets_the_gets]]
 
+lemma oblivious_monadic_rewrite:
```
Xaphiosis, line 763: this is in the corres/refinement section... isn't there a better place for it, even a new section?
btw, this should work for any F, right?

lib/clib/Corres_UL_C.thy
```diff
@@ -309,6 +309,14 @@ lemma ccorres_to_vcg:
   apply fastforce
   done
 
+lemma ccorres_to_vcg_Normal:
+  "\<lbrakk>ccorres_underlying srel \<Gamma> rrel xf arrel axf P P' [] a c; no_fail Q a\<rbrakk>
+   \<Longrightarrow> \<Gamma> \<turnstile> {s'. P s \<and> Q s \<and> s' \<in> P' \<and> (s, s') \<in> srel} c UNIV"
```
Xaphiosis, line 314: I don't understand this one... is this equivalent to saying that it isn't UNIV? where does this fall out of?

lib/Lib.thy
```diff
@@ -1913,6 +1913,37 @@ lemma length_takeWhile_gt:
...
+ "map fst (filter (\<lambda>(_, x). P x) (zip ls (map f ls))) = filter (\<lambda>x. P (f x)) ls"
+  by (induction ls; clarsimp)
+
+lemma map_fst_dropWhile_zip:
+ "map fst (dropWhile (\<lambda>(_, x). P x) (zip ls (map f ls))) = dropWhile (\<lambda>x. P (f x)) ls"
+  by (induction ls; clarsimp)
+
+lemma map_fst_takeWhile_zip:
+ "map fst (takeWhile (\<lambda>(_, x). P x) (zip ls (map f ls))) = takeWhile (\<lambda>x. P (f x)) ls"
+  by (induction ls; clarsimp)
```
Xaphiosis, line 1926: these three look tempting to stick into the simpset

Xaphiosis:
Nice. I'm really behind the times with all the reader and option monad stuff, so can't really comment on those.
Something worth deciding on now: I don't know if one commit per lemma added is the right way to go. It might be more useful to group them, have one commit per group and list the lemmas in there (so they still show up in git log)? The commits don't really say where something was added either, so the knowledge they provide is kinda limited anyway. @lsf37 ?

lib/clib/Corres_UL_C.thy
michaelmcinerney: I have used this rule with the `ccorres` rules for `While` - when the C loop condition is just `UNIV`, the `ccorres_While` rule will ask you to show a `hoarep` goal with the postcondition just `UNIV`, which just means that the execution will result in a `Normal` state. This new rule allows us to do this fairly easily from a `ccorres` rule for the loop body.

lib/MonadicRewrite.thy
michaelmcinerney: It will work with both flags schematic, actually. Thanks.

I could make a new section, but I'm not sure what to call it. Maybe "Interaction with `oblivious`"? Or maybe I could expand the current section to "corres / refinement / oblivious"?

Xaphiosis: I don't think I'd want it in this section, even renamed. My suggestion is to create a section after this one (before Tool Integration), something like: "Interaction with other rewrite techniques: oblivious" (and then if more come along, we can do oblivious/whatever). Thoughts?

lib/clib/Corres_UL_C.thy
Xaphiosis: Fascinating. Does it really have to be a proper subset then?

lib/MonadicRewrite.thy
michaelmcinerney: Sure, I'll go with your suggestion

lib/clib/Corres_UL_C.thy
michaelmcinerney: Sorry, what do you mean by a proper subset?

Xaphiosis: Upon closer examination there seems to be some font rendering issue happening on my side that made the `c UNIV` look like `\<subset> UNIV`. Fortunately I can see the `c` now, and apologise for the noise.

lib/Lib.thy
lsf37: +1, they do.

lib/ListLibLemmas.thy
```diff
@@ -423,6 +437,16 @@ lemma list_remove_none: "x \<notin> set list \<Longrightarrow> list_remove list
   apply clarsimp+
   done
 
+lemma list_remove_append:
+  "list_remove (xs @ ys) t = list_remove xs t @ list_remove ys t"
+  by (induct xs) auto
```
lsf37, line 442: This would also be a good `[simp]` rule

lib/MonadicRewrite.thy
lsf37: +1 for a new section, I think calling it "Interaction with oblivious" would be fine.

lib/Monads/reader_option/Reader_Option_VCG.thy
```diff
@@ -124,6 +134,16 @@ lemma owhile_ovalid[wp]:
    apply auto
   done
 
+lemma case_options_weak_ovalid:
+  "\<lbrakk> \<lblot>P\<rblot> f \<lblot>Q\<rblot>; \<And>x. \<lblot>P'\<rblot> g x \<lblot>Q\<rblot> \<rbrakk>
+   \<Longrightarrow> \<lblot>P and P'\<rblot> case opt of None \<Rightarrow> f | Some x \<Rightarrow> g x \<lblot>Q\<rblot>"
```
lsf37, line 139: Curious that we need this one. Does `wpc` work for `ovalid`? Do we still need to set that one up?

lsf37 (approved):
Looking good! I would indeed squash the commits as proposed.

lib/Monads/reader_option/Reader_Option_VCG.thy
michaelmcinerney: I'm not sure I've tried `wpc` with `ovalid`. Actually, this lemma ended up not being used in the work I thought I wanted it for. Should I just delete it?

lsf37: Yes, let's scrap it. If `wpc` isn't set up yet, we should probably do that in this theory, but some other time.

michaelmcinerney: Great. Should I squash and force push so that you can do one last check of the commits?

lsf37: yes, please

Xaphiosis:
> Looking good! I would indeed squash the commits as proposed.

What do you think of the current result? I don't think the "many" is useful, and I was hoping that the lemma names would instead make it into the longer commit message... but if they're gone at this point, putting them back requires effort.

lib/MonadicRewrite.thy
Xaphiosis: Sigh. This is probably the only interaction with oblivious we're going to see, hence my suggestion for a more general section name, but this can work I guess.

michaelmcinerney: Happy to call it whatever. If you two could decide between yourselves and let me know, that would be appreciated

lsf37:
I would still tweak the commit messages a bit. I.e. I'd drop the `many` from the `lib:` commits and indicate where the lemmas are added (first one says "reader monad", which is fine), but in the other commits I can't tell from looking at the commit message what it is that is changing. Mentioning the theory name would work, or the corresponding concept when the theory name is too long.

lib/MonadicRewrite.thy
Xaphiosis: Gerwin guesses that there might be more rules for oblivious, I'm not so sure. Fine to leave it like it is.

michaelmcinerney:
> I would still tweak the commit messages a bit. I.e. I'd drop the `many` from the `lib:` commits and indicate where the lemmas are added -- first one says "reader monad", which is fine, but in the other commits I can't tell from looking at the commit message what it is that is changing. Mentioning the theory name would work, or the corresponding concept when the theory name is too long.

So would `lib: add lemmas about lists to Lib.thy and ListLibLemmas.thy` be ok? The other commits seems specific enough to me. But could we not also just use `git log —name-only` to see the files which are modified, if we wanted to?

lsf37:
> So would `lib: add lemmas about lists to Lib.thy and ListLibLemmas.thy` be ok? 

The `.thy` doesn't carry information, so I'd go for `lib: add list lemmas to Lib and ListLibLemmas`

> The other commits seems specific enough to me. 

They mention the lemma name, but they don't say where something changes apart from `lib` which is very large. In this case the lemma names do kinda give you enough information (`corres_stateAssert` would correctly assumed to be in `Corres_UL` etc), so yes, you are right. We can leave those as is.

> But could we not also just use `git log —name-only` to see the files which are modified, if we wanted to?

Yes, one could -- it's not about enumerating all changed files, but to make it possible to quickly see what areas changed when you just scan the commit headers.

michaelmcinerney:
> The `.thy` doesn't carry information, so I'd go for `lib: add list lemmas to Lib and ListLibLemmas`

`Lib` is the name of the session as well as the name of the file to which I'm adding lemmas, but I will go with your suggestion.
 
> Yes, one could -- it's not about enumerating all changed files, but to make it possible to quickly see what areas changed when you just scan the commit headers.

I personally can't see the benefit of that, but if that's the case, then could the commit messages guide be updated for the case where we're adding to `lib` but not to a subdirectory like `monads`?

lsf37:
> > The `.thy` doesn't carry information, so I'd go for `lib: add list lemmas to Lib and ListLibLemmas`
> 
> `Lib` is the name of the session as well as the name of the file to which I'm adding lemmas, but I will go with your suggestion.

It is the name of the session, but the tag already indicates that, so I don't think it would be confused.

> > Yes, one could -- it's not about enumerating all changed files, but to make it possible to quickly see what areas changed when you just scan the commit headers.
> 
> I personally can't see the benefit of that, but if that's the case, then could the commit messages guide be updated for the case where we're adding to `lib` but not to a subdirectory like `monads`?

The [example section](https://github.com/seL4/l4v/blob/master/docs/commit-messages.md#looking-at-history) is trying to express that, but we can make it more explicit for `lib` specifically, yes.

PR #721, opened 2024-02-21
Enlarge the trace monad rule set for RG logic

corlewis:
This generally follows the structure and naming conventions of the existing rule set for Hoare logic, in `Trace_VCG.thy` and `Trace_More_VCG.thy` (and the very similar ruleset and files in the nondet directory).

I took the opportunity to make some changes as I went along. In particular:
* `validIE_R` and `validIE_L` are abbreviations instead of definitions.
* some lemmas were relocated to better follow the structure of the file and to improve reuse.
* some lemmas were renamed in an attempt to be more consistent.

It would be good to propagate the last two back to the Hoare versions, to keep things consistent and allow us to update the rest of the proofs now. The first one might not be worth it however, as it would require too many other changes to work.

Lemmas for lifting over combinators in the guarantee have been left out for now as it is unclear if they would be useful. They shouldn't be too hard to add if we come across a use case.

### DIFF
--- a/lib/Monads/trace/Trace_More_RG.thy
+++ b/lib/Monads/trace/Trace_More_RG.thy
@@ -1,0 +1,637 @@
+(*
+ * Copyright 2023, Proofcraft Pty Ltd
+ * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ *)
+
+(* Partial correctness RG logic lemmas over the trace monad. RG quintuples, lifting lemmas, etc.
+   If it doesn't contain a RG quintuple it likely doesn't belong in here. *)
+
+theory Trace_More_RG
+  imports
+    Trace_RG
+begin
+
+lemma rg_take_disjunct:
+  "\<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>\<lambda>rv s0 s. P' rv s0 s \<and> (False \<or> P'' rv s0 s)\<rbrace>
+   \<Longrightarrow> \<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>P''\<rbrace>"
+  by (erule rg_strengthen_post, simp)
+
+lemma rg_post_add:
+  "\<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> S \<lbrace>G\<rbrace>,\<lbrace>\<lambda>r s0 s. Q' r s0 s \<and> Q r s0 s\<rbrace> \<Longrightarrow> \<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> S \<lbrace>G\<rbrace>,\<lbrace>Q\<rbrace>"
+  by (erule rg_strengthen_post, simp)
+
+lemma rg_post_addE:
+  "\<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>\<lambda>_ s0 s. R s0 s \<and> Q s0 s\<rbrace>,\<lbrace>T\<rbrace> \<Longrightarrow> \<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>\<lambda>_ s0 s. Q s0 s\<rbrace>,\<lbrace>T\<rbrace>"
+  by (erule rg_strengthen_postE; simp)
+
+lemma rg_pre_add:
+  "(\<forall>s0 s. P s0 s \<longrightarrow> P' s0 s) \<Longrightarrow> \<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>Q\<rbrace> \<longleftrightarrow> \<lbrace>P and P'\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>Q\<rbrace>"
+  apply (subst iff_conv_conj_imp)
+  by(intro conjI impI; rule rg_weaken_pre, assumption, clarsimp)
+
+lemma rg_pre_addE:
+  "(\<forall>s0 s. P s0 s \<longrightarrow> R s0 s) \<Longrightarrow> \<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>Q\<rbrace>,\<lbrace>S\<rbrace> \<longleftrightarrow> \<lbrace>P and R\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>Q\<rbrace>,\<lbrace>S\<rbrace>"
+  apply (subst iff_conv_conj_imp)
+  by(intro conjI impI; rule rg_weaken_preE, assumption, clarsimp)
+
+lemma rg_name_pre_state:
+  "\<lbrakk> \<And>s0 s. P s0 s \<Longrightarrow> \<lbrace>\<lambda>_. (=) s\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>Q\<rbrace>; prefix_closed f \<rbrakk> \<Longrightarrow> \<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>Q\<rbrace>"
+  by (clarsimp simp: validI_def)
+
+lemma rg_name_pre_stateE:
+  "\<lbrakk>\<And>s0 s. P s0 s \<Longrightarrow> \<lbrace>\<lambda>_. (=) s\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>Q\<rbrace>,\<lbrace>E\<rbrace>; prefix_closed f\<rbrakk> \<Longrightarrow> \<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>Q\<rbrace>,\<lbrace>E\<rbrace>"
+  by (clarsimp simp: validIE_def2)
+
+lemma rg_vcg_if_lift_strong:
+  "\<lbrakk> \<lbrace>P'\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>P\<rbrace>; \<lbrace>\<lambda>s0 s. \<not> P' s0 s\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>\<lambda>rv s0 s. \<not> P rv s0 s\<rbrace>; \<lbrace>Q'\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>Q\<rbrace>; \<lbrace>S'\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>S\<rbrace> \<rbrakk> \<Longrightarrow>
+   \<lbrace>\<lambda>s0 s. if P' s0 s then Q' s0 s else S' s0 s\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>\<lambda>rv s0 s. if P rv s0 s then Q rv s0 s else S rv s0 s\<rbrace>"
+
+  "\<lbrakk> \<lbrace>P'\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>P\<rbrace>; \<lbrace>\<lambda>s0 s. \<not> P' s0 s\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>\<lambda>rv s0 s. \<not> P rv s0 s\<rbrace>; \<lbrace>Q'\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace> Q\<rbrace>; \<lbrace>S'\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>S\<rbrace> \<rbrakk> \<Longrightarrow>
+   \<lbrace>\<lambda>s0 s. if P' s0 s then Q' s0 s else S' s0 s\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>\<lambda>rv s0 s. (if P rv s0 s then Q rv else S rv) s0 s\<rbrace>"
+  by (wpsimp wp: rg_vcg_imp_lift' | assumption | fastforce)+
+
+lemma rg_vcg_imp_lift_pre_add:
+  "\<lbrakk> \<lbrace>P and Q\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>\<lambda>rv s0 s. Q' rv s0 s\<rbrace>; f \<lbrace>R\<rbrace>,\<lbrace>G\<rbrace>,\<lbrace>\<lambda>s0 s. \<not> Q s0 s\<rbrace> \<rbrakk>
+   \<Longrightarrow> \<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>\<lambda>rv s0 s. Q s0 s \<longrightarrow> Q' rv s0 s\<rbrace>"
+  apply (rule rg_weaken_pre)
+   apply (rule rg_vcg_imp_lift')
+    apply fastforce
+   apply fastforce
+  apply clarsimp
+  done
+
+lemma rg_pre_tautI:
+  "\<lbrakk> \<lbrace>A and P\<rbrace>,\<lbrace>R\<rbrace> a \<lbrace>G\<rbrace>,\<lbrace>B\<rbrace>; \<lbrace>A and not P\<rbrace>,\<lbrace>R\<rbrace> a \<lbrace>G\<rbrace>,\<lbrace>B\<rbrace> \<rbrakk> \<Longrightarrow> \<lbrace>A\<rbrace>,\<lbrace>R\<rbrace> a \<lbrace>G\<rbrace>,\<lbrace>B\<rbrace>"
+  by (fastforce simp: validI_def)
+
+lemma rg_lift_Pf_pre_conj:
+  assumes P: "\<And>x. \<lbrace>\<lambda>s0 s. Q x s0 s\<rbrace>,\<lbrace>R\<rbrace> m \<lbrace>G\<rbrace>,\<lbrace>P x\<rbrace>"
+  assumes f: "\<And>P. \<lbrace>\<lambda>s0 s. P (g s0 s) \<and> P' s0 s\<rbrace>,\<lbrace>R\<rbrace> m \<lbrace>G\<rbrace>,\<lbrace>\<lambda>_ s0 s. P (f s0 s)\<rbrace>"
+  shows "\<lbrace>\<lambda>s0 s. Q (g s0 s) s0 s \<and> P' s0 s\<rbrace>,\<lbrace>R\<rbrace> m \<lbrace>G\<rbrace>,\<lbrace>\<lambda>rv s0 s. P (f s0 s) rv s0 s\<rbrace>"
+  apply (clarsimp simp: validI_def validI_prefix_closed[OF f] guar_by_rg[OF f])
+  apply (rule use_validI[OF _ P], simp)
+  apply (rule use_validI[OF _ f], simp+)
+  done
+
+lemmas rg_lift_Pf4 = rg_lift_Pf_pre_conj[where P'="\<top>\<top>", simplified]
+lemmas rg_lift_Pf3 = rg_lift_Pf4[where f=f and g=f for f]
+lemmas rg_lift_Pf2 = rg_lift_Pf3[where P="\<lambda>f _. P f" for P]
+lemmas rg_lift_Pf = rg_lift_Pf2[where Q=P and P=P for P]
+
+lemmas rg_lift_Pf3_pre_conj = rg_lift_Pf_pre_conj[where f=f and g=f for f]
+lemmas rg_lift_Pf2_pre_conj = rg_lift_Pf3_pre_conj[where P="\<lambda>f _. P f" for P]
+lemmas rg_lift_Pf_pre_conj' = rg_lift_Pf2_pre_conj[where Q=P and P=P for P]
+
+lemma rg_if_r_and:
+  "\<lbrace>P\<rbrace>,\<lbrace>R'\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>\<lambda>r. if R r then Q r else Q' r\<rbrace>
+   = \<lbrace>P\<rbrace>,\<lbrace>R'\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>\<lambda>r s0 s. (R r \<longrightarrow> Q r s0 s) \<and> (\<not>R r \<longrightarrow> Q' r s0 s)\<rbrace>"
+  by (fastforce simp: validI_def)
+
+lemma rg_convert_imp:
+  "\<lbrakk> \<lbrace>\<lambda>s0 s. \<not> P s0 s\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>\<lambda>rv s0 s. \<not> Q s0 s\<rbrace>; \<lbrace>P'\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>S\<rbrace> \<rbrakk>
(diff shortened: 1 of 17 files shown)


lib/Monads/trace/Trace_RG.thy
```diff
@@ -101,30 +106,77 @@ definition validI ::
...
+
+abbreviation validIE_E ::
+  "('s \<Rightarrow> 's \<Rightarrow> bool) \<Rightarrow> 's rg_pred \<Rightarrow> ('s, 'a + 'b) tmonad \<Rightarrow> 's rg_pred \<Rightarrow> ('a \<Rightarrow> 's \<Rightarrow> 's \<Rightarrow> bool) \<Rightarrow>
+    bool"
+  ("(\<lbrace>_\<rbrace>,/ \<lbrace>_\<rbrace>)/ _ /(\<lbrace>_\<rbrace>,/ -,/ \<lbrace>_\<rbrace>)") where
+  "\<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,-,\<lbrace>E\<rbrace> \<equiv> \<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>\<lambda>_ _ _. True\<rbrace>,\<lbrace>E\<rbrace>"
+
+text \<open>
+  The following abbreviations are convenient to separate reasoning about postconditions and the
+  guard.\<close>
```
lsf37, line 150:
```suggestion
  The following abbreviations are convenient to separate reasoning about postconditions and the
  guarantee condition.\<close>
```

lib/Monads/trace/Trace_RG.thy
```diff
@@ -101,30 +106,77 @@ definition validI ::
...
+\<comment> \<open>FIXME MC: name\<close>
+abbreviation validI' ::
```
lsf37, line 152: maybe `validI_no_guarantee`? Since it's an abbreviation, we won't have to type (or see) this name very often, so it's fine if it's a bit longer.

lib/Monads/trace/Trace_RG.thy
```diff
@@ -230,6 +314,135 @@ lemma bind_twp[wp_split]:
...
+  apply (simp add: bindE_def validIE_def)
+  apply (erule bind_twp[rotated])
+  apply (clarsimp simp: lift_def throwError_def split: sum.splits)
+  apply (subst validI_valid_no_trace_eq)
+   apply wpsimp+
+  done
+
+\<comment> \<open>FIXME MC: improve naming for bind_twp vs seq_ext vs rg_seq_ext
+          existing proofs mention hoare_seq_ext a lot and seq_ext a few times\<close>
+\<comment> \<open>FIXME MC: names of seq_ext and seq_extE clash with valid equivalents\<close>
```
lsf37, line 329: I'd recommend `(* .. *)` comments for FIXME, because then you're not constrained by LaTeX syntax (these here will crash on the underscore, I think).

lsf37: The `seq_ext` names are old and a bit non-standard. If we can eliminate them now in favour of variations of `bind_wp`, I'd be up for it.

lib/Monads/trace/Trace_RG.thy
```diff
@@ -230,6 +314,135 @@ lemma bind_twp[wp_split]:
...
+
+lemmas rg_seq_ext = bind_twp
+lemmas rg_seq_extE = bindE_twp
+
+lemma rg_TrueI[simp]:
+  "\<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> f -,\<lbrace>\<lambda>_ _. \<top>\<rbrace> = prefix_closed f"
+  by (simp add: validI_def)
+
+lemma rgE_TrueI[simp]:
+  "\<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> f -,\<lbrace>\<lambda>_ _. \<top>\<rbrace>,\<lbrace>\<lambda>_ _. \<top>\<rbrace> = prefix_closed f"
```
lsf37, line 341: These two look slightly dangerous as `[simp]`, because they could leave `P` and `R` as free schematics in the goal.

lib/Monads/trace/Trace_RG.thy
```diff
@@ -230,6 +314,135 @@ lemma bind_twp[wp_split]:
...
+
+lemma rg_conj:
+  "\<lbrakk>\<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>Q\<rbrace>; \<lbrace>P'\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>Q'\<rbrace>\<rbrakk> \<Longrightarrow> \<lbrace>P and P'\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>Q and Q'\<rbrace>"
+  unfolding validI_def by auto
+
+lemmas wp_post_tautI = rg_TrueI[where P="\<top>\<top>", THEN iffD2]
+lemmas wp_post_tautEI = rgE_TrueI[where P="\<top>\<top>", THEN iffD2]
+
+lemma rg_pre_cont[simp]:
+  "prefix_closed f \<Longrightarrow> \<lbrace>\<bottom>\<bottom>\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>Q\<rbrace>"
```
lsf37, line 360: similar to the other two above -- `R` could end up as free schematic (assuming we're in a `wp` setting `G` and `Q` should be concrete).

lib/Monads/trace/Trace_RG.thy
```diff
@@ -230,6 +314,135 @@ lemma bind_twp[wp_split]:
...
+
+lemma rg_post_disjI2:
+  "\<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>\<lambda>rv. Q' rv\<rbrace> \<Longrightarrow> \<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>\<lambda>rv s0. Q rv s0 or Q' rv s0\<rbrace>"
+  unfolding validI_def by auto
+
+\<comment> \<open>FIXME MC: do we want the last assumption?  It might make unification easier in unusual cases but so
+           far is always just resolved with refl. \<close>
+lemma use_validI:
+  "\<lbrakk>(tr, Result (rv, s')) \<in> (rely f R s0 s); \<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>Q\<rbrace>; P s0 s; s0' = last_st_tr tr s0\<rbrakk>
+   \<Longrightarrow> Q rv s0' s'"
```
lsf37, line 420: Could always make a `'` version with the assumption in and then the main version where it is resolved by `refl`. Then it's available in the rarer cases where it's needed and we get the more convenient short version by default.

(nitpick: would leave out the parentheses in `(rely ..)`)

lib/Monads/trace/Trace_RG.thy
```diff
@@ -230,6 +314,135 @@ lemma bind_twp[wp_split]:
...
+  unfolding validI_def by auto
+
+lemmas post_by_rg = use_validI[rotated]
+
+lemma use_validI_guar:
+  "\<lbrakk>(tr, res) \<in> (rely f R s0 s); \<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>Q\<rbrace>; P s0 s\<rbrakk>
```
lsf37, line 426:
```suggestion
  "\<lbrakk>(tr, res) \<in> rely f R s0 s; \<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>Q\<rbrace>; P s0 s\<rbrakk>
```

lib/Monads/trace/Trace_RG.thy
```diff
@@ -230,6 +314,135 @@ lemma bind_twp[wp_split]:
...
+   \<Longrightarrow> guar_cond G s0 tr"
+  unfolding validI_def by auto
+
+lemmas guar_by_rg = use_validI_guar[rotated]
+
+lemma in_inv_by_rgD:
+  "\<lbrakk>\<And>P. f \<lbrace>R\<rbrace>,-,\<lbrace>P\<rbrace>; (tr, Result (rv, s')) \<in> (rely f R s0 s)\<rbrakk> \<Longrightarrow> s' = s"
```
lsf37, line 433:
```suggestion
  "\<lbrakk>\<And>P. f \<lbrace>R\<rbrace>,-,\<lbrace>P\<rbrace>; (tr, Result (rv, s')) \<in> rely f R s0 s\<rbrakk> \<Longrightarrow> s' = s"
```

lib/Monads/trace/Trace_RG.thy
```diff
@@ -230,6 +314,135 @@ lemma bind_twp[wp_split]:
...
+  unfolding validI_def
+  apply (drule meta_spec[where x="\<lambda>_. (=) s"])
+  apply blast
+  done
+
+lemma last_st_in_inv_by_rgD:
+  "\<lbrakk>\<And>P. f \<lbrace>R\<rbrace>,-,\<lbrace>P\<rbrace>; (tr, Result (rv, s')) \<in> (rely f R s0 s)\<rbrakk> \<Longrightarrow> last_st_tr tr s0 = s0"
```
lsf37, line 440:
```suggestion
  "\<lbrakk>\<And>P. f \<lbrace>R\<rbrace>,-,\<lbrace>P\<rbrace>; (tr, Result (rv, s')) \<in> rely f R s0 s\<rbrakk> \<Longrightarrow> last_st_tr tr s0 = s0"
```

lib/Monads/trace/Trace_RG.thy
```diff
@@ -471,38 +686,660 @@ lemma rely_cond_rtranclp:
...
+  "\<lbrakk>\<And>x. \<lbrace>P x\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,-,\<lbrace>E x\<rbrace>\<rbrakk> \<Longrightarrow> \<lbrace>\<lambda>s0 s. \<exists>x. P x s0 s\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,-,\<lbrace>\<lambda>rv s0 s. \<exists>x. E x rv s0 s\<rbrace>"
+  by (fastforce simp: validIE_def validI_def split: sum.splits)
+
+lemma rg_vcg_ex_lift_R1:
+  "\<lbrakk>\<And>x. \<lbrace>P x\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>Q\<rbrace>,-\<rbrakk> \<Longrightarrow> \<lbrace>\<lambda>s0 s. \<exists>x. P x s0 s\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>Q\<rbrace>,-"
+  by (fastforce simp: validI_def validIE_def split: sum.splits)
+
+\<comment> \<open>FIXME MC: keep/name/move?\<close>
+lemma subst2:
+  "\<lbrakk>s = t; u = v; P s u\<rbrakk> \<Longrightarrow> P t v"
```
lsf37, line 1082: Could go in `lib` -- it's for convenience, I assume.

lib/Monads/trace/Trace_RG.thy
```diff
@@ -471,38 +686,660 @@ lemma rely_cond_rtranclp:
...
+
+lemma rg_vcg_split_lift[wp]:
+  "\<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> f x y \<lbrace>G\<rbrace>,\<lbrace>Q\<rbrace> \<Longrightarrow> \<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> case (x, y) of (a, b) \<Rightarrow> f a b \<lbrace>G\<rbrace>,\<lbrace>Q\<rbrace>"
+  by simp
+
+named_theorems rg_vcg_op_lift
+lemmas [rg_vcg_op_lift] =
+  rg_vcg_const_imp_lift
+  rg_vcg_const_imp_liftE_E
+  rg_vcg_const_imp_liftE_R
```
lsf37, line 1140: Do we also need operator lifting for the guarantee part? I guess this will come out once we have more example proofs, I don't have a good intuition yet what is necessary there and what not.

lib/Monads/trace/Trace_RG.thy
```diff
@@ -531,6 +1368,332 @@ lemma Await_sync_twp:
   apply (simp add: o_def)
   done
 
+
+subsection \<open>Setting up the @{method wp} method\<close>
+
+(* Attempt to define triple_judgement to use valid_validI_wp as wp_comb rule.
+   It doesn't work. It seems that wp_comb rules cannot take more than 1 assumption *)
+lemma validI_is_triple:
```
lsf37, line 1376: I think that is correct (that they expect exactly 1 assumption).

lib/Monads/trace/Trace_More_RG.thy
```diff
@@ -0,0 +1,685 @@
...
+
+lemma rg_case_option_wpR:
+  "\<lbrakk>\<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> f None \<lbrace>G\<rbrace>,\<lbrace>Q\<rbrace>,\<lbrace>E\<rbrace>; \<And>x. \<lbrace>P' x\<rbrace>,\<lbrace>R\<rbrace> f (Some x) \<lbrace>G\<rbrace>,\<lbrace>Q' x\<rbrace>,\<lbrace>E\<rbrace>\<rbrakk>
+   \<Longrightarrow> \<lbrace>case_option P P' v\<rbrace>,\<lbrace>R\<rbrace> f v \<lbrace>G\<rbrace>,\<lbrace>\<lambda>rv. case v of None \<Rightarrow> Q rv | Some x \<Rightarrow> Q' x rv\<rbrace>,\<lbrace>E\<rbrace>"
+  by (cases v) auto
+
+\<comment> \<open>FIXME MC: delete since we won't use K for rg? If we don't delete, it needs to assume the G
+lemma K_valid[wp]:
+  "\<lbrace>K P\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>\<lambda>_. K P\<rbrace>"
+  by (simp add: valid_def)\<close>
```
lsf37, line 107: would delete, yes

lib/Monads/trace/Trace_More_RG.thy
```diff
@@ -0,0 +1,685 @@
...
+  apply (rule validI_split[rotated, simplified pred_conj_def])
+   apply assumption
+  apply (induct xs; simp)
+   apply wpsimp
+  apply (rule rg_vcg_conj_lift; simp)
+  done
+
+\<comment> \<open>FIXME MC: no longer true since it requires "prefix_closed undefined", is that a problem?
+lemma undefined_valid: "\<lbrace>\<bottom>\<bottom>\<rbrace>,\<lbrace>R\<rbrace> undefined \<lbrace>G\<rbrace>,\<lbrace>Q\<rbrace>"
+  by (rule rg_pre_cont)\<close>
```
lsf37, line 218: Interesting. It used to be convenient to get `False` for things like `haskell_error` (which maps to `undefined`) so we had to prove that the case is unreachable.

Do we have a better concept that we could map `haskell_error` to? I guess we could map `error` to produce chaos + failure, i.e. the monad that does no trace step, but produce UNIV as possible results and next states. That would actually be better on the nondet monad as well, so that we don't get these random `undefined` instances.

lsf37: (so, yes, delete -- but we need to leave a note somewhere what to do with `haskell_error`)

lib/Monads/trace/Trace_More_RG.thy
```diff
@@ -0,0 +1,685 @@
...
+
+lemma opt_return_pres_lift_rg:
+  assumes x: "\<And>v. \<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> f v \<lbrace>G\<rbrace>,\<lbrace>\<lambda>rv. P\<rbrace>"
+  shows      "\<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> case x of None \<Rightarrow> return () | Some v \<Rightarrow> f v \<lbrace>G\<rbrace>,\<lbrace>\<lambda>rv. P\<rbrace>"
+  by (wpsimp wp: x)
+
+\<comment> \<open>FIXME MC: this would be a pain to prove for validI, can it be deleted?
+lemma validI_return_unit:
+  "\<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> f >>= (\<lambda>_. return ()) \<lbrace>G\<rbrace>,\<lbrace>\<lambda>r. Q\<rbrace> \<Longrightarrow> \<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>\<lambda>r. Q\<rbrace>"
+  by (auto simp: validI_def in_bind in_return Ball_def)\<close>
```
lsf37, line 342: I think that can be removed. `wp` should be fine to deal with the bind/return pattern without a specific rule.

lib/Monads/trace/Trace_More_RG.thy
```diff
@@ -0,0 +1,685 @@
...
+  "\<lbrakk> P \<or> Q; P \<Longrightarrow> \<lbrace>R\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>S\<rbrace>; Q \<Longrightarrow> \<lbrace>T\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>S\<rbrace> \<rbrakk>
+   \<Longrightarrow> \<lbrace>\<lambda>s0 s. (P \<longrightarrow> R s0 s) \<and> (Q \<longrightarrow> T s0 s)\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>S\<rbrace>"
+  apply safe
+   apply (rule rg_weaken_pre)
+    apply simp
+   apply simp
+  apply (rule rg_weaken_pre)
+   apply simp
+  apply simp
+  done
```
lsf37, line 511: could this be a `fastforce`? If not, I feel at least the urge for `safe; rule rg_weaken_pre; simp`.

lib/Monads/trace/Trace_More_RG.thy
```diff
@@ -0,0 +1,685 @@
...
+   \<lbrace>P and P'\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>\<lambda>r. if C r then Q r else Q' r\<rbrace>"
+  by (auto simp: validI_def)
+
+lemma weak_if_twp':
+  "\<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>\<lambda>r. Q r and Q' r\<rbrace> \<Longrightarrow>
+   \<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>\<lambda>r. if C r then Q r else Q' r\<rbrace>"
+  by (auto simp: validI_def)
+
+\<comment> \<open>FIXME MC: this one also isn't obvious, and the valid equivalent doesn't seem to be used anywhere.
+lemma bindE_split_recursive_asm:
```
lsf37, line 590: this one can be removed

lib/Monads/trace/Trace_RG.thy
lsf37: Right, just saw that you had already commented on that. I agree, let's leave these for now.

lsf37 (approved):
Happy with that.

lsf37:
> It would be good to propagate the last two back to the Hoare versions, to keep things consistent and allow us to update the rest of the proofs now. 

We probably should do that, at least if the lemma rename is not too intrusive and it doesn't take too long. It's easiest to do when the cache is still hot, but priority would be the refinement rules.

> The first one might not be worth it however, as it would require too many other changes to work.

Agreed.

lib/Monads/trace/Trace_RG.thy
corlewis: > I'd recommend `(* .. *)` comments for FIXME, because then you're not constrained by LaTeX syntax (these here will crash on the underscore, I think).

Right. I have my Jedit set up to default to `\<comment> \<open> .. \<close>`, since it works inside lemma statements, but `(* .. *)` would clearly be better here. Quite a few of the FIXMEs will be removed after I incorporate all of your feedback and I'll update the remaining ones afterwards. If nothing else, that should get the tests green.

corlewis: > The `seq_ext` names are old and a bit non-standard. If we can eliminate them now in favour of variations of `bind_wp`, I'd be up for it.

I'll come up with something that makes sense to me and see what you think. That would be another thing that would be good to bring back to `nondet`.

lib/Monads/trace/Trace_RG.thy
corlewis: All of these `validI` rules given as `simp` are based on ones from nondet, in this case `hoare_TrueI`. In practice here, any proof with `P` as a schematic will almost definitely be hit with `wp` first, which will use `wp_post_tautI` to instantiate the `P` as `λ_ _. True`.

As you mentioned later on, `R` should almost never be schematic. At a guess we'll only see schematic relies in proofs dealing with `parallel` and in some refinement proofs, and in both of these cases the `R` would be instantiated pretty quickly.

corlewis: Having said that, should I take the opportunity to try and clean up some of these `[simp]` rules? That would be another thing to try in `nondet`, to see if they're actually used anywhere.

lib/Monads/trace/Trace_RG.thy
corlewis: I used it to prove a goal of `P s0 (f s) ⟹ P (last_st_tr tr s0) (f s')`, where I was able to then show that each of the arguments were equal. I couldn't think of a better way to do this than writing this new version of `subst` and then doing `apply (erule subst2[rotated 2, where P=P])`.

If we're happy to keep it then it could go in `Monads/Monad_Lib.thy` I guess. `Monads` builds directly on `HOL`, so putting it in `lib` would lead to some weird dependencies.

lib/Monads/trace/Trace_More_RG.thy
corlewis: I think this is before I realised that it was safe to leave out the `G` in a lot of these rules, so if we wanted to be complete we could do `⦃K P⦄,⦃R⦄ f -,⦃λ_. K P⦄`. Now that I think about it, I guess that might be useful for postconditions that mix conjuncts that talk about both states with others that only refer to the current state.

lib/Monads/trace/Trace_More_RG.thy
```diff
@@ -0,0 +1,685 @@
...
+  apply (wpsimp wp: x[unfolded validIE_def] split: sum.splits)
+  done
+
+lemma assertE_tsp:
+  "\<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> assertE Q \<lbrace>G\<rbrace>,\<lbrace>\<lambda>rv s0 s. Q \<and> P s0 s\<rbrace>,\<lbrace>E\<rbrace>"
+  by (wpsimp wp: assertE_wp)
+
+\<comment> \<open>FIXME MC: delete? The only reason to keep is so that the RG version can get simplified, otherwise
+           wp will lift it to a hoare triple\<close>
+lemma gets_inv_rg[simp]:
```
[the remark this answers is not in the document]
corlewis: I haven't really tested it out, but if we removed this lemma then I think the behaviour would be that `wp` would lift it to a hoare triple, and then `simp` would resolve it with `gets_inv`, the hoare equivalent to this lemma.

The only benefit I can think of for additionally having this version is so that anything adapted from existing proofs that just used `simp` would keep on working, without needing to add that `wp` step.

lib/Monads/trace/Trace_More_RG.thy
corlewis: Another to add to the list to try removing from nondet.

lib/Monads/trace/Trace_More_RG.thy
corlewis: Thanks for noticing this, `by (fastforce intro: rg_weaken_pre)` works.

lib/Monads/trace/Trace_RG.thy
corlewis: I'm currently going with renaming `seq_ext` to `bind_twp_fwd` and removing `rg_seq_ext` completely. See the most recent commit for the additional changes it leads to.

What do you think? If you're happy with that then I'll add it to the list of changes to make to `Nondet_VCG` and `Trace_VCG`.

lib/Monads/trace/Trace_RG.thy
lsf37: Yes, Monad_Lib is a good idea.

lib/Monads/trace/Trace_More_RG.thy
lsf37: Happy with that.

lib/Monads/trace/Trace_RG.thy
lsf37: `nondet` has a bunch of these from very early times, but removing then will cause breakage in strange places. I'd leave that for a little later and focus on the trace monad only for now. Maybe add a cleanup issue for it, but I'd try leaving them out of `[simp]` here so that we're at least not propagating them to the new monad.

lib/Monads/trace/Trace_More_RG.thy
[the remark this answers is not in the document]
lsf37: I'd remove it -- we're going to have to adjust proofs anyway and that should make things a bit more uniform

lib/Monads/trace/Trace_More_RG.thy
corlewis: Whoops, I wasn't thinking properly when I said that yesterday, that rule doesn't make any sense since it would be saying that `f` preserves all `P`s that only talk about the current state. I'll just delete this.

PR #722, opened 2024-02-24
Monads: refactor reader_option and nondet connection 

corlewis:
I am very open to suggestions if anyone has an idea for a nicer way to structure this or to reduce the copied content. I had a brief attempt at a locale but couldn't find a way to get it to work that didn't involve even more copying.

### DIFF
--- a/lib/Monads/reader_option/Reader_Option_Monad.thy
+++ b/lib/Monads/reader_option/Reader_Option_Monad.thy
@@ -260,7 +260,11 @@ definition
 definition
   "oskip \<equiv> oreturn ()"
 
-lemmas omonad_defs = ofail_def oreturn_def oassert_def oassert_opt_def asks_def ounless_def
+lemma ogets_def:
+  "ogets f = (\<lambda>s. Some (f s))"
+  by (clarsimp simp: asks_def obind_def oreturn_def)
+
+lemmas omonad_defs = ofail_def oreturn_def oassert_def oassert_opt_def ogets_def ounless_def
                      owhen_def
 
 
@@ -478,6 +482,8 @@ lemma obindK_is_opt_map:
   "f \<bind> (\<lambda>x. K $ g x) = f |> g"
   by (simp add: obind_def opt_map_def)
 
+lemmas in_omonad = bind_eq_Some_conv in_obind_eq in_opt_map_eq in_opt_pred Let_def
+
 
 section \<open>"While" loops over option monad.\<close>
 
@@ -583,6 +589,7 @@ proof -
       by (induct arbitrary:s) (auto intro: step final)
   qed
 qed
+
 
 section \<open>Lift @{term option_while} to the @{typ "('a,'s) lookup"} monad\<close>
 
--- a/lib/Monads/trace/Trace_Reader_Option.thy
+++ b/lib/Monads/trace/Trace_Reader_Option.thy
@@ -1,0 +1,219 @@
+(*
+ * Copyright 2024, Proofcraft Pty Ltd
+ * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ *)
+
+(* Reader option monad syntax plus the connection between the reader option monad and the trace monad *)
+
+theory Trace_Reader_Option
+imports
+  Trace_No_Fail
+  Reader_Option_VCG
+begin
+
+(* FIXME: remove this syntax, standardise on do {..} instead *)
+(* Syntax defined here so we can reuse Trace_Monad definitions *)
+syntax
+  "_doO" :: "[dobinds, 'a] => 'a"  ("(DO (_);//   (_)//OD)" 100)
+
+translations
+  "_doO (_dobinds b bs) e" == "_doO b (_doO bs e)"
+  "_doO (_nobind b) e"     == "b |>> (CONST K_bind e)"
+  "DO x <- a; e OD"        == "a |>> (\<lambda>x. e)"
+
+
+lemma ovalid_K_bind_wp[wp]:
+  "ovalid P f Q \<Longrightarrow> ovalid P (K_bind f x) Q"
+  by simp
+
+lemma ovalidNF_K_bind_wp[wp]:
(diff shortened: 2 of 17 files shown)


lib/Monads/trace/Trace_Reader_Option.thy
```diff
@@ -0,0 +1,217 @@
...
+qed
+
+lemma The_eqD:
+  "\<lbrakk>The P = x; \<exists>!x. P x\<rbrakk> \<Longrightarrow> P x"
+  by (metis theI)
+
+lemma gets_the_whileLoop:
+  fixes C :: "'a \<Rightarrow> 's \<Rightarrow> bool"
+  assumes terminates: "\<forall>s. whileLoop_terminates C (\<lambda>a. gets_the (B a)) r s"
+  shows "whileLoop C (\<lambda>a. gets_the (B a)) r = gets_the (owhile C B r)"
```
corlewis, line 165: Non-terminating loops are a bit more complicated for the trace monad compared to the nondet monad, since I don't think that we can 'just' fail. I'm pretty confident that that means we need this extra assumption here, but it's possible that with more thought there could be a change made to the definition of `whileLoop` instead.

Xaphiosis:
Could you explain what you're trying to do? I'm seeing some theory files renamed and some proofs moved, and the commits refer to "reader_option and nondet connection" without any further description. I'm guessing you're doing something you then need for integration with the trace monad, but I'm really lost as to the intent. What do you mean by connection?

corlewis:
> Could you explain what you're trying to do? I'm seeing some theory files renamed and some proofs moved, and the commits refer to "reader_option and nondet connection" without any further description. I'm guessing you're doing something you then need for integration with the trace monad, but I'm really lost as to the intent. What do you mean by connection?

My overall aim with these changes is to make it possible to use the reader_option monad with the trace monad, in the same way that it can currently be used with the nondet monad.

Without these changes, the reader_option files `Reader_Option_ND.thy` and `Reader_Option_VCG.thy` depend on `nondet` files, and prove various lemmas connecting reader_option concepts with nondet concepts. As an example,
```
lemma ovalid_gets_the:
  "ovalid P f Q ⟹ ⦃P⦄ gets_the f ⦃Q⦄"
```
shows that the nondet concept of validity of `gets_the f` can be derived from the reader_option concept of validity. Most of these lemmas were in `Reader_Option_ND.thy`, which was described by a comment as
```
(* Reader option monad syntax plus the connection between the reader option monad and the nondet monad *)
```

What I've done in the first commit is to basically reverse the dependency: the `reader_option` directory now only contains definitions and lemmas for the reader_option monad and there's a file in `nondet` that has the lemmas connecting the two monads. The third commit then adds a file in `trace` that is basically the same thing but for the trace monad.

Xaphiosis:
That's a very good explanation! Any chance some of it could make it to the commits? I had no idea until you explained it

corlewis:
> That's a very good explanation! Any chance some of it could make it to the commits? I had no idea until you explained it

Done, do those commit descriptions make sense to you?

I also rebased and updated copyright, which I had forgotten to do.

lib/Monads/trace/Trace_Reader_Option.thy
lsf37: I think this is right. Non-terminating loops make sense in the trace monad setting (producing an unbounded trace set), so mapping them to failure is no longer the right thing to do, which means the non-termination case no longer matches up with the reader monad.

There is probably a tweak possible that we can do later: since we're comparing with the reader monad, we know there is no state change and (I think) no trace produced. In that specific case, failure might actually make sense for the while loop, but I'm not clear whether that would come out of the current definitions or not.

lib/Monads/reader_option/Reader_Option_Monad.thy
```diff
@@ -614,4 +619,49 @@ proof -
   then show ?thesis by (auto intro: final split: option.split_asm)
 qed
 
+
+\<comment> \<open>FIXME: better name for this section\<close>
+section \<open>Extra definitions\<close>
+
+\<comment> \<open>FIXME: The only one of these currently used is odrop, in sep-capDL/Lookups_D.
+          As far as I can tell, the definitions in Lookups_D aren't used. Should we
+          remove them, and these?\<close>
```
lsf37, line 628: I'm very much up for deleting them, yes!

lsf37 (approved):
👍 This works for me with the updated commit messages. You might still foreshadow in the first msg why you're doing that refactor (so that you can produce the same setup later for the trace monad).

michaelmcinerney (approved):
I wish I knew more about the trace monad stuff to comment on that, but this looks like a very good thing to do!

Xaphiosis:
> > That's a very good explanation! Any chance some of it could make it to the commits? I had no idea until you explained it
> 
> Done, do those commit descriptions make sense to you?
> 
> I also rebased and updated copyrights, which I had forgotten to do.

Thanks! What Gerwin said. Also, two commits have "reader_option" but one has "reader option".

lib/Monads/reader_option/Reader_Option_Monad.thy
corlewis: I might have been a bit too hasty with what I said here. I just checked again and parts of Lookups_D are still used in `capDL-api`, but the specific definitions that use these reader_option definitions aren't.

See https://github.com/seL4/l4v/commit/e75b38713bdcefa9dea0e77dbd341433c0f33e86 for what could be removed.

lib/Monads/trace/Trace_Reader_Option.thy
lsf37: Ok, definitely something to come back to later.

lib/Monads/reader_option/Reader_Option_Monad.thy
lsf37: That's still pretty good, let's do that.

corlewis:
Commit messages updated, rebased over https://github.com/seL4/l4v/pull/725, and I realised that I had left a lemma collection in the nondet and trace files that was just about reader_option concepts and moved it.

Xaphiosis (approved):

PR #723, opened 2024-02-25
arm-hyp+aarch64 haskell: add VCPURegVMPIDR(_EL2)

lsf37:
This brings the spec in line with seL4/seL4#1181.

Test with: seL4/seL4#1181

### DIFF
--- a/proof/crefine/ARM_HYP/Retype_C.thy
+++ b/proof/crefine/ARM_HYP/Retype_C.thy
@@ -5343,7 +5343,7 @@ lemma ptr_retyp_fromzeroVCPU:
   assumes cor: "caps_overlap_reserved' {p ..+ 2 ^ vcpu_bits} \<sigma>"
   assumes ptr0: "p \<noteq> 0"
   assumes kdr: "{p ..+ 2 ^ vcpu_bits} \<inter> kernel_data_refs = {}"
-  assumes subr: "{p ..+ 456} \<subseteq> {p ..+ 2 ^ vcpu_bits}"
+  assumes subr: "{p ..+ 464} \<subseteq> {p ..+ 2 ^ vcpu_bits}" (is "{_ ..+ ?vcpusz} \<subseteq> _")
   assumes act_bytes: "region_actually_is_bytes p (2 ^ vcpu_bits) \<sigma>'"
   assumes rep0: "heap_list (hrs_mem (t_hrs_' (globals \<sigma>'))) (2 ^ vcpu_bits) p = replicate (2 ^ vcpu_bits) 0"
   assumes "\<not> snd (placeNewObject p vcpu0 0 \<sigma>)"
@@ -5360,7 +5360,8 @@ proof -
   let ?htdret = "(hrs_htd_update (ptr_retyp (vcpu_Ptr p)) (t_hrs_' (globals \<sigma>')))"
   let ?zeros = "from_bytes (replicate (size_of TYPE(vcpu_C)) 0) :: vcpu_C"
 
-  have "size_of TYPE(vcpu_C) = 456" (is "_ = ?vcpusz")
+  (* sanity check for the value of ?vcpusz *)
+  have "size_of TYPE(vcpu_C) = ?vcpusz"
     by simp
 
   have ptr_al:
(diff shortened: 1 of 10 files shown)


lsf37:
Ok, this is now ready for review. The AARCH64 proof update has to be done separately, will add a PR to the aarch64 branch fro that later.

proof/crefine/ARM_HYP/Retype_C.thy
```diff
@@ -5574,7 +5574,7 @@ proof -
     apply (clarsimp simp: ko_vcpu_def vcpu0_def)
     apply (clarsimp simp: rf_sr_def cstate_relation_def carch_state_relation_def
                           cmachine_state_relation_def Let_def h_t_valid_clift_Some_iff)
-    apply (subgoal_tac "region_is_bytes p 456 \<sigma>'")
+    apply (subgoal_tac "region_is_bytes p 464 \<sigma>'")
```
corlewis, line 5577: It's hard to tell from the diff, but can you use `?vcpusz` here as well?

corlewis (approved):

proof/crefine/ARM_HYP/Retype_C.thy
```diff
@@ -5360,7 +5360,7 @@ proof -
   let ?htdret = "(hrs_htd_update (ptr_retyp (vcpu_Ptr p)) (t_hrs_' (globals \<sigma>')))"
   let ?zeros = "from_bytes (replicate (size_of TYPE(vcpu_C)) 0) :: vcpu_C"
 
-  have "size_of TYPE(vcpu_C) = 456" (is "_ = ?vcpusz")
+  have "size_of TYPE(vcpu_C) = ?vcpusz"
```
Xaphiosis, line 5363: I'd add a comment that this is a sanity check, otherwise it's unused now and not declaring the var

proof/crefine/ARM_HYP/Retype_C.thy
Xaphiosis: Yeah, we're still in reach of the lemma statement, does it not work?

proof/crefine/ARM_HYP/Wellformed_C.thy
```diff
@@ -43,7 +43,7 @@ abbreviation
 abbreviation
   pd_Ptr :: "32 word \<Rightarrow> (pde_C[2048]) ptr" where "pd_Ptr == Ptr"
 abbreviation
-  regs_C_Ptr :: "addr \<Rightarrow> (machine_word_len word[42]) ptr" where "regs_C_Ptr \<equiv> Ptr"
+  regs_C_Ptr :: "addr \<Rightarrow> (machine_word_len word[43]) ptr" where"regs_C_Ptr \<equiv> Ptr"
```
Xaphiosis, line 46: I think I did something to improve this on AARCH64. Might be worth seeing what the lines you changed in this PR look like there in case there's improvement that we already figured out.

Xaphiosis: ah yes, the `value_type vcpuregs_count = "length enum_vcpureg"` failing; so I tried to but did not succeed. any chance you could have a stab at that now?

proof/crefine/ARM_HYP/Retype_C.thy
lsf37: Let me check, I thought this was already a different lemma. Would definitely be nicer to use here if not.

proof/crefine/ARM_HYP/Retype_C.thy
lsf37: Good point, will add.

proof/crefine/ARM_HYP/Wellformed_C.thy
lsf37: Hm, it probably needs some code generator setup and be done outside the Arch locale. I can have a look.

proof/crefine/ARM_HYP/Retype_C.thy
lsf37: You are absolutely right. Will update.

proof/crefine/ARM_HYP/Wellformed_C.thy
lsf37: It turns out there is a constant in C `seL4_VCPUReg_Num` for which we can declare `seL4_VCPUReg_Num_def[code]` and that then works fine.

lsf37:
Have addressed the review feedback. Now that AARCH64 CRefine is merged, I can add the AARCH64 part of the proofs (and deploy the `seL4_VCPUReg_Num` value type there).

Xaphiosis (approved):

PR #733, opened 2024-03-14
Proofs for SGI API

lsf37:
This PR adds verification for the SGISignalCap API of [RFC-17](https://sel4.github.io/rfcs/active/0170-multikernel-ipi-api.html) (seL4/seL4#1222).

Specs and proofs are updated for all architectures (affects mostly ARM, ARM_HYP, and AARCH64), including all affected parts of the proof stack (access control, infoflow, capDL, system initialiser etc). In the system initialiser, SGISignalCaps are currently still excluded via the definition of wellformed specifications.

~This is currently still in draft mode until the RFC recommendations have been implemented.~

Now ready for review.

Test with https://github.com/seL4/seL4/pull/1222

[RFC-17]: https://sel4.atlassian.net/browse/RFC-17?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


spec/haskell/src/SEL4/API/Invocation/AARCH64.hs
```diff
@@ -114,6 +123,11 @@ data IRQControlInvocation
         issueHandlerSlot,
         issueHandlerControllerSlot :: PPtr CTE,
         issueHandlerTrigger :: Bool }
+    | IssueSGISignal {
+        issueSGIIRQ :: Word,
+        issueSGITargets :: Word,
+        issueSGIControlSlot,
+        issueSGISlot :: PPtr CTE }
```
Xaphiosis, line 130: Do we want any more documentation here? I can guess at IRQ and Targets, but ControlSlot and Slot are completely baffling to me.

spec/haskell/src/SEL4/API/InvocationLabels/ARM.lhs
```diff
@@ -51,5 +51,5 @@ FIXME ARMHYP ARMPageMapIO is an inconsistant name (but coined by kernel team)
 >         | ARMVCPUAckVPPI
 #endif
 >         | ARMIRQIssueIRQHandler
+>         | ARMIRQIssueSGISignal
 >         deriving (Eq, Enum, Bounded, Show)
-
```
Xaphiosis, line 55: last empty line removal?

spec/haskell/src/SEL4/Kernel/VSpace/ARM.lhs
```diff
@@ -1506,13 +1503,7 @@ notion of the largest permitted object size, and checks it appropriately.
...
 >             performASIDPoolInvocation oper
 >             return []
-#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
->         InvokeVCPU _ -> fail "performARMMMUInvocation: not an MMU invocation"
-#endif
-#ifdef CONFIG_ARM_SMMU
->         InvokeIOSpace _ -> fail "performARMMMUInvocation: not an MMU invocation"
->         InvokeIOPageTable _ -> fail "performARMMMUInvocation: not an MMU invocation"
-#endif
+>         _ -> fail "performARMMMUInvocation: not an MMU invocation"
```
Xaphiosis, line 1506: Isn't removing the SMMU/IOMMU parts and inserting a default case a bit much here? Same comment as for AARCH64, this will not catch future expansion in MMU invocations, and having a few extra lines is not such a high price for it.

spec/haskell/src/SEL4/Machine/Hardware/ARM.lhs
```diff
@@ -21,7 +21,7 @@ This module defines the low-level ARM hardware interface.
 
 > import Control.Monad.Reader
 > import Data.Bits
-> import Data.Word(Word8, Word32, Word64)
+> import Data.Word(Word8, Word16, Word32, Word64)
```
Xaphiosis, line 24: I'm seeing Word16 added in a bunch of includes, but you don't appear to be using it.

spec/haskell/src/SEL4/Object/ObjectType.lhs
```diff
@@ -519,5 +520,3 @@ The following two functions returns the base and size of the object a capability
 >     = 1 -- error in haskell
 > capUntypedSize (IRQHandlerCap {})
 >     = 1 -- error in haskell
-
-
```
Xaphiosis, line 523: does this not trigger "no empty line at end of file" lint?

spec/haskell/src/SEL4/Object/ObjectType/AARCH64.hs
```diff
@@ -50,10 +51,26 @@ deriveCap _ (c@FrameCap {})
...
 deriveCap _ (c@ASIDPoolCap {}) = return $ ArchObjectCap c
 deriveCap _ (c@VCPUCap {}) = return $ ArchObjectCap c
+deriveCap _ (c@SGISignalCap {}) = return $ ArchObjectCap c
 -- TODO AARCH64 SMMU: SID/CB control caps, SID/CB caps
 
 isCapRevocable :: Capability -> Capability -> Bool
-isCapRevocable newCap srcCap = False
+isCapRevocable newCap srcCap = case newCap of
+     ArchObjectCap (SGISignalCap {}) -> srcCap == IRQControlCap
+     _ -> False
```
Xaphiosis, line 60: indentation: 4 not 5

spec/haskell/src/SEL4/Object/ObjectType/AARCH64.hs
```diff
@@ -113,12 +130,18 @@ finaliseCap (VCPUCap { capVCPUPtr = vcpu }) True = do
...
+
 -- TODO AARCH64 SMMU: C also has cap_cb_cap, cap_sid_cap (but not SID/CB control caps)
 
 finaliseCap _ _ = return (NullCap, NullCap)
 
 {- Identifying Capabilities -}
 
+isIRQControlCapDescendant :: ArchCapability -> Bool
+isIRQControlCapDescendant (SGISignalCap {}) = True
+isIRQControlCapDescendant _ = False
```
Xaphiosis, line 143: bit confused, what does it mean to be an IRQ Control Cap descendant?

spec/design/skel/ARM_HYP/ArchInterruptDecls_H.thy
```diff
@@ -8,9 +8,12 @@ theory ArchInterruptDecls_H
 imports RetypeDecls_H CNode_H
 begin
 
+requalify_types (in Arch)
+  sgisignal_invocation
+
 context Arch begin arch_global_naming (H)
 
-#INCLUDE_HASKELL SEL4/Object/Interrupt/ARM.lhs CONTEXT Arch decls_only ArchInv=
+#INCLUDE_HASKELL SEL4/Object/Interrupt/ARM.lhs CONTEXT Arch decls_only ArchInv=Arch
```
Xaphiosis, line 16: do you remember what this does again, and why you needed to add it?

spec/design/skel/CNode_H.thy
```diff
@@ -10,13 +10,19 @@ theory CNode_H
...
   TCBDecls_H
   CSpaceDecls_H
   EndpointDecls_H
   PSpaceFuns_H
 begin
 
+context begin interpretation Arch .
+requalify_consts
+  isArchMDBParentOf
+end
```
Xaphiosis, line 24: arch-split docs now say to not do this, because this is very slow and I built infrastructure to avoid it.
Does `arch_requalify_consts` with an optional `(H)` not work for you here?

tools/haskell-translator/caseconvs
```diff
@@ -2092,3 +2092,216 @@ case \x of Arch.Types.APIObjectType _ -> Arch.Types.SmallPageObject -> Arch.Type
...
+    then ->3
+    else if isASIDControlCap a \<and> isASIDControlCap b
+    then ->4
+    else if isASIDPoolCap a \<and> isASIDPoolCap b
+    then ->5
+    else if isVCPUCap a \<and> isVCPUCap b
+    then ->6
+    else if isSGISignalCap a \<and> isSGISignalCap b
+    then ->7
+    else ->8
```
Xaphiosis, line 2307: That's a lot more caseconvs than I expected! Must've been painful. They look correct though.

spec/abstract/ARM_HYP/Arch_Structs_A.thy
```diff
@@ -198,16 +203,13 @@ where
...
   arch_cap_is_device :: "arch_cap \<Rightarrow> bool"
 where
-  "arch_cap_is_device (PageCap dev x rs sz as4) = dev"
-| "arch_cap_is_device ASIDControlCap = False"
-| "arch_cap_is_device (ASIDPoolCap p as) = False"
-| "arch_cap_is_device (PageTableCap x as3) = False"
-| "arch_cap_is_device (PageDirectoryCap x as2) = False"
-| "arch_cap_is_device (VCPUCap _) = False"
+  "arch_cap_is_device (PageCap dev _ _ _ _) = dev"
+| "arch_cap_is_device _ = False"
```
Xaphiosis, line 212: again I'm not sure about using a catch-all here

proof/invariant-abstract/ARM_HYP/ArchCSpacePre_AI.thy
```diff
@@ -231,18 +232,20 @@ lemmas valid_arch_mdb_updates = valid_arch_mdb_free_index_update valid_arch_mdb_
...
 
 lemma safe_parent_arch_is_parent:
-  "\<lbrakk>safe_parent_for_arch cap pcap; caps_of_state s p = Some pcap;
-      valid_arch_mdb (is_original_cap s) (caps_of_state s)\<rbrakk> \<Longrightarrow>
+  "\<lbrakk>safe_parent_for_arch cap pcap; caps_of_state s p = Some pcap; valid_mdb s \<rbrakk> \<Longrightarrow>
     should_be_parent_of pcap (is_original_cap s p) cap f"
-  by auto
+  apply (clarsimp simp: should_be_parent_of_def is_cap_simps
+                        is_irq_control_descendant_def valid_mdb_def)
+  apply (erule (1) irq_revocableD)
```
Xaphiosis, line 243: I guess it didn't work with `elim!:`?

proof/invariant-abstract/ARM_HYP/ArchBits_AI.thy
```diff
@@ -31,7 +31,7 @@ lemma pbfs_atleast_pageBits:
   "pageBits \<le> pageBitsForSize sz"
   by (cases sz) (auto simp: pageBits_def)
 
-lemmas is_cap_simps = is_cap_simps is_arch_cap_simps
+lemmas is_cap_simps = is_cap_simps is_arch_cap_simps is_SGISignalCap_def is_ArchObjectCap_def
```
Xaphiosis, line 34: hang on, why isn't `is_SGISignalCap_def` part of `is_arch_cap_simps`?

proof/invariant-abstract/AARCH64/ArchEmptyFail_AI.thy
```diff
@@ -25,10 +25,10 @@ declare loadWord_empty_fail[EmptyFail_AI_assms]
 end
 
 global_interpretation EmptyFail_AI_load_word?: EmptyFail_AI_load_word
-  proof goal_cases
+proof goal_cases
   interpret Arch .
   case 1 show ?case by (unfold_locales; (fact EmptyFail_AI_assms)?)
-  qed
+qed
```
Xaphiosis, line 31: This requires some discussion. In this PR, you've unilaterally started re-indenting these when there are no changes in them. However:
- the people who did arch-split originally did it with the `proof` indented, and this has been enshrined in `docs/arch-split.md` which I assume everyone read and did not point this out
- doing it like this makes ad-hoc changes on only the files you've touched, leading to inconsistency between files (I've been copying the existing pattern for arch-split of Refine, for example, so as not to create noise)
- usual tradeoff where diffs will show you only `proof goal_cases` rather than what it's for, but we live with it for lemmas so we can live with it here
- these kinds of cleanups might complicate merging with `rt`, which is also why I've been torn on whether to do them or not

Anyway, if we want to do this, it might be worth doing them all in one go as a separate commit.

proof/invariant-abstract/ARM/ArchArch_AI.thy
```diff
@@ -1259,250 +1270,259 @@ lemma aligned_sum_less_kernel_base:
...
+  apply (wpsimp simp: arch_decode_invocation_def valid_arch_inv_def)
   done
 
 
 declare word_less_sub_le [simp]
 
 crunch perform_page_table_invocation, perform_page_invocation,
-           perform_asid_pool_invocation, perform_page_directory_invocation
+       perform_asid_pool_invocation, perform_page_directory_invocation,
+        perform_sgi_invocation
```
Xaphiosis, line 1507: strange indentation still

proof/invariant-abstract/ARM/ArchDetSchedSchedule_AI.thy
```diff
@@ -360,9 +360,9 @@ lemma handle_reserved_irq_valid_sched:
 end
 
 global_interpretation DetSchedSchedule_AI_handle_hypervisor_fault?: DetSchedSchedule_AI_handle_hypervisor_fault
-  proof goal_cases
+proof goal_cases
   interpret Arch .
   case 1 show ?case by (unfold_locales; (fact handle_hyp_fault_valid_sched handle_reserved_irq_valid_sched)?)
-  qed
+qed
```
Xaphiosis, line 366: e.g. here, the diff for this file is over 2/3 unrelated to this PR

proof/invariant-abstract/ARM/ArchEmptyFail_AI.thy
```diff
@@ -52,18 +52,20 @@ lemma decode_tcb_invocation_empty_fail[wp]:
...
 
-crunch find_pd_for_asid, get_master_pde, check_vp_alignment,
-                   create_mapping_entries, ensure_safe_mapping, get_asid_pool, resolve_vaddr
+crunch
+  find_pd_for_asid, get_master_pde, check_vp_alignment, create_mapping_entries, ensure_safe_mapping,
+  get_asid_pool, resolve_vaddr, decode_sgi_signal_invocation
   for (empty_fail) empty_fail[wp]
   (simp: kernel_object.splits arch_kernel_obj.splits option.splits pde.splits pte.splits)
 
+declare decode_sgi_signal_invocation_empty_fail[simp]
```
Xaphiosis, line 61: I'm never sure about this, but we sometimes declare empty_fails with `intro!` etc, do we want to be consistent with these?

proof/refine/X64/CNodeInv_R.thy
```diff
@@ -4106,6 +4106,38 @@ lemma isNTFNbadge_dest:
...
+
+lemma mdb_chunked_arch_assms_scap[simp]:
+  "mdb_chunked_arch_assms scap =  mdb_chunked_arch_assms src_cap"
+  by (simp add: mdb_chunked_arch_assms_def)
+
+lemma mdb_chunked_arch_assms_dcap[simp]:
+  "mdb_chunked_arch_assms dcap =  mdb_chunked_arch_assms dest_cap"
+  by (simp add: mdb_chunked_arch_assms_def)
+
+
```
Xaphiosis, line 4122: extra newline relevant here?

proof/refine/X64/Retype_R.thy
```diff
@@ -2637,7 +2637,7 @@ proof -
   from valid
   have "valid_badges m" ..
   thus ?thesis
-    apply (clarsimp simp: valid_badges_def)
+    apply (clarsimp simp: valid_badges_def valid_arch_badges_def) (* FIXME arch-split; take AARCH64 version *)
```
Xaphiosis, line 2640: I appreciate the arch-split annotations, thanks!

proof/refine/AARCH64/Bits_R.thy
```diff
@@ -56,8 +56,11 @@ lemma isCap_simps:
   "isASIDControlCap w = (w = ASIDControlCap)"
   "isASIDPoolCap w = (\<exists>v0 v1. w = ASIDPoolCap v0 v1)"
   "isVCPUCap w = (\<exists>v. w = VCPUCap v)"
+  "isSGISignalCap w = (\<exists>irq targets. w = SGISignalCap irq targets)"
   by (auto simp: isCap_defs split: capability.splits arch_capability.splits)
 
+lemmas isCap_simps = isCap_simps' isArchSGISignalCap_def
```
Xaphiosis, line 62: confused why this manoeuvre was needed

proof/refine/ARM_HYP/CNodeInv_R.thy
```diff
@@ -4129,6 +4147,75 @@ lemmas ep_simps =
   isEPsrc isEPbadge_src isNTFNsrc isNTFNbadge_src
   isEPdest isEPbadge_dest isNTFNdest isNTFNbadge_dest
 
+(* FIXME arch-split: arch dependent part of mdb_swap proofs, no arch_global_naming because of locale *)
```
Xaphiosis, line 4150: not sure I follow, which locale?

proof/refine/ARM_HYP/CSpace1_R.thy
```diff
@@ -133,9 +146,9 @@ lemma same_arch_region_as_relation:
   "\<lbrakk>acap_relation c d; acap_relation c' d'\<rbrakk> \<Longrightarrow>
   arch_same_region_as c c' =
   sameRegionAs (ArchObjectCap d) (ArchObjectCap d')"
-  apply (cases c)
-  apply ((cases c', auto simp: ARM_HYP_H.sameRegionAs_def sameRegionAs_def Let_def isCap_simps)[1])+
-  done
+  by (cases c; cases c')
+      (auto simp: ARM_HYP_H.sameRegionAs_def sameRegionAs_def Let_def isCap_simps mask_def
```
Xaphiosis, line 150: indent? `(auto` looks off by 1

proof/refine/ARM/CSpace1_R.thy
```diff
@@ -3179,43 +3213,48 @@ lemma n_dest:
 
 end
 
+context begin interpretation Arch .
```
Xaphiosis, line 3216: Would appreciate FIXME arch-split on these as per arch-split.md
You're also getting a time penalty when you do these. It's much faster to do:
`context Arch begin arch_global naming`
and then `arch_requalify_fact ... (* FIXME arch-split *)`

proof/crefine/AARCH64/CSpace_C.thy
```diff
@@ -2450,22 +2489,56 @@ lemma valid_cap'_PageCap_is_aligned:
   apply (simp add: valid_cap'_def capAligned_def)
 done
 
+lemma pageBitsForSize_64[simplified,simp]:
```
Xaphiosis, line 2492: these have strange names (64? not mentioning less?), they seem like generic consequences of architecture-specific constants

Xaphiosis:
This is a ton of work, I had not realised how much changes would be required!
I've done a review pass, going commit-by-commit in chronological order. I tried to only comment on any issues I found once per arch, to reduce noise, but most of my comments and questions apply to the other arches in their group (has SGI group, and doesn't have SGI group). In a PR this big, having mixed in cleanup also caused extra cognitive load, even though the cleanup is welcome in the long term.

It also looks like it maybe should go in before the arch-split of StateRelation, as that is ultimately a smaller diff where the difficulty was getting stuff right. So the current queue looks like SGI, then a race between Isabelle 2025 and FPU, and then arch-split somewhere in there (currently on top of 2025).

proof/refine/AARCH64/Bits_R.thy
lsf37: `isArchSGISignalCap` is not a "normal" isCap predicate from the data type definitions (it's defined manually), but it is used like one, that's why it's added to the standard simp set for these.

spec/haskell/src/SEL4/API/Invocation/AARCH64.hs
lsf37: They are the same as in IssueIRQHandler just above (probably not visible in the diff). The `issueSGIControlSlot` is the slot where the `SGIControl` cap sits that authorises the invocation, the `issueSGISlot` is where to put the new `SGI` cap that is being generated. We need both in the invocation, because the control cap becomes a parent of the new cap (which means we need to modify its CDT pointers).

I'm not sure this file is a good place to start explaining invocations. None of the other invocation have any explanations there. We could explain in decodeInvocation, but there they should actually become fairly clear by how they are used.

spec/haskell/src/SEL4/API/InvocationLabels/ARM.lhs
lsf37: AFAIR, `git diff --check` complained. We used to have an extra empty line at the end for literal Haskell preprocessing, but that is unused now, so we don't need to do anything special there any more.

spec/haskell/src/SEL4/Kernel/VSpace/AARCH64.hs
```diff
@@ -876,7 +876,7 @@ performARMMMUInvocation i = withoutPreemption $ do
         InvokeASIDPool oper -> do
             performASIDPoolInvocation oper
             return []
-        InvokeVCPU _ -> fail "performARMMMUInvocation: not an MMU invocation"
+        _ -> fail "performARMMMUInvocation: not an MMU invocation"
```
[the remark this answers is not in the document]
lsf37: It was, but it got pretty unwieldy, and there will be another such case with SMC caps coming next, which will make it even worse. I don't think it's worth the extra hassle. The early warning was nice when it was just one or two, but it's not like a missing case there is going to go unnoticed forever. It will come up in the proofs at the latest.

spec/haskell/src/SEL4/Kernel/VSpace/ARM.lhs
lsf37: I found the repetition annoying, and it's going to get worse. It's not just the definition here where you have the repetition, you will also get repetition in `case_convs`, which is especially annoying because there you have to enumerate the parameters of those invocations as well. Unless we're doing SMMU, we're much more likely to add non-MMU invocations, so I think this is guarding the wrong case.

spec/haskell/src/SEL4/Kernel/VSpace/AARCH64.hs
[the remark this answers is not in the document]
Xaphiosis: I can buy the argument, though the result is not best-tasting.

spec/haskell/src/SEL4/Machine/Hardware/ARM.lhs
lsf37: good catch, they need to go (the first implementation used Word16 and it looks like I didn't catch all of them when I adapted it).

spec/haskell/src/SEL4/API/Invocation/AARCH64.hs
Xaphiosis: Thanks for explaining.

We sometimes treat the haskell spec like a doc, hence I brought it up. If none of the other ones have them, then probably not the place indeed.

spec/haskell/src/SEL4/Object/Interrupt/AARCH64.hs
```diff
@@ -57,6 +69,22 @@ performIRQControl (ArchInv.IssueIRQHandler (IRQ irq) destSlot srcSlot trigger)
     setIRQState IRQSignal (IRQ irq)
     cteInsert (IRQHandlerCap (IRQ irq)) srcSlot destSlot
     return ()
+performIRQControl (ArchInv.IssueSGISignal irq targets controlSlot sgiSlot)
+    = withoutPreemption $ do
+    cteInsert (ArchObjectCap (SGISignalCap irq targets)) controlSlot sgiSlot
+    return ()
```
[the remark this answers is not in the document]
lsf37: They should. Not sure what I was doing there.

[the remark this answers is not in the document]
lsf37: I think those are just moved, it was not my sanity that left me at least. Will fix.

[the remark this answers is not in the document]
lsf37: Not moved, copied from the clause above. So some sanity should have presented itself, but no.

spec/haskell/src/SEL4/Object/ObjectType.lhs
lsf37: I'm confused. The lint wants me to not have an empty line at the end (a newline, yes, but no empty line).

spec/haskell/src/SEL4/Object/ObjectType/AARCH64.hs
lsf37: keen eye :-)

spec/haskell/src/SEL4/Object/ObjectType/AARCH64.hs
lsf37: Just what it says :-) Whether, in the CDT, the given cap is a transitive child (=descendant) of the IRQControl cap. For other caps this is more complex and you can't tell by just looking at two caps, but here there is only one IRQControl cap in the system, and it has no content, so there is nothing left to check apart from the cap type.

Happy to add more explanation. Was CDT missing, or was the shape of it unexpectedly different from other predicates on that? For other caps, those predicates don't appear in the Haskell, so I guess that is usual as well.

spec/haskell/src/SEL4/Object/ObjectType.lhs
Xaphiosis: I think I'm the one confused. In jedit, you see an empty line at the end, in vim you don't, and in git diff I'm not sure anymore

spec/abstract/ARM_HYP/Arch_Structs_A.thy
lsf37: This one I'm sure about. It just selects the `dev` part of one specific cap type, with a default for the rest. Enumerating the other cap types doesn't make sense and wouldn't guard against a new cap that has some `dev` parameter either, because you could just as easily add the wrong case there if you're just going through by type error.

spec/haskell/src/SEL4/Object/ObjectType/AARCH64.hs
Xaphiosis: For some reason I thought SGI caps were not under IRQControl cap. I don't know where this mental gap originated. Possibly hallucinating interactions with IO port caps.

proof/invariant-abstract/ARM_HYP/ArchCSpacePre_AI.thy
lsf37: Unfortunately not, it needs the second assumption (the `(1)`), which `clarsimp` on its own doesn't get. I'll try `force`, but I have the feeling I already did.

proof/invariant-abstract/ARM_HYP/ArchBits_AI.thy
lsf37: well spotted -- it should be (on AARCH64 it is). Ah, because ARM/ARM_HYP does this:

```
lemma is_arch_cap_simps:
  "is_pg_cap cap = (\<exists>dev p R sz m. cap = (ArchObjectCap (PageCap dev p R sz m)))"
  "is_pd_cap cap = (\<exists>p asid. cap = (ArchObjectCap (PageDirectoryCap p asid)))"
  "is_pt_cap cap = (\<exists>p asid. cap = (ArchObjectCap (PageTableCap p asid)))"
  by (auto simp add: is_pg_cap_def is_pd_cap_def is_pt_cap_def)
```

instead of having an actual complete `is_arch_cap_simps` with names. Back then the automatic `is_` datatype discriminators didn't exist yet, so we wrote our own.

Not entirely sure what to do there. `is_arch_cap_simps` seems to be fairly constrained (just page table things) on ARM/ARM_HYP, not the full set. 

AARCH64 has:

```
lemmas is_arch_cap_simps = is_pt_cap_eq is_ap_cap_eq is_frame_cap_eq is_SGISignalCap_def
```

proof/invariant-abstract/AARCH64/ArchEmptyFail_AI.thy
lsf37: Agreed, this should be in a separate PR. Will remove.

spec/haskell/src/SEL4/Object/ObjectType/AARCH64.hs
lsf37: It doesn't help that I called it SGIControlCap somewhere in a comment above

proof/invariant-abstract/ARM_HYP/ArchBits_AI.thy
Xaphiosis: I'm ok with leaving it for now (but yuck). One idea is to go to the `is_arch_cap_simps` on ARM/ARM_HYP, make it `is_arch_cap_simps'` and then define `is_arch_cap_simps` using the prime and the `is_SGISignalCap_def is_ArchObjectCap_def`

spec/haskell/src/SEL4/Object/ObjectType/AARCH64.hs
Xaphiosis: Ahh, that's probably the inspiration then

spec/design/skel/ARM_HYP/ArchInterruptDecls_H.thy
lsf37: Back in 2023 I did remember, but not any more. Let me see what breaks if I undo it.

spec/design/skel/CNode_H.thy
lsf37: A lot of this PR predates that tech, and there is also a lot of copy/paste from after, copying bad style. Will have a look at going through those and modernising. You're absolutely right that we should not produce more old style stuff.

proof/refine/X64/CNodeInv_R.thy
lsf37: indeed a stray

spec/design/skel/CNode_H.thy
Xaphiosis: Got it. If it got shoved into an Arch without global naming, you can do `requalify_consts Arch.isArchMDBParentOf`

proof/crefine/AARCH64/CSpace_C.thy
lsf37: Looks like I was trying to move `lemma pageBitsForSize_64 [simp]:` from SyscallArgs_C.thy, cleaning up the statement, but not the name, and then also not removing the one in SyscallArgs_C. The name is used a few times, sadly all in AARCH64 where we should have known better. This should probably go higher than both of these and get a proper name.

proof/crefine/ARM/Interrupt_C.thy
```diff
@@ -416,6 +417,104 @@ lemma checkIRQ_ret_good:
...
+  apply (simp add: unat_ucast)
+  done
+
+lemma maxIRQ_ucast_toEnum_irq_t2:
+  "x \<le> Kernel_Config.maxIRQ \<Longrightarrow>
+   UCAST(_ \<rightarrow> machine_word_len) (toEnum (unat x)::irq) \<le> ucast Kernel_C.maxIRQ"
+  for x::machine_word
+  by (simp add: Kernel_C_maxIRQ maxIRQ_ucast_toEnum_eq_irq ucast_ucast_mask word_and_le)
+
+lemma maxIRQ_ucast_toEnum_irq_t3:
```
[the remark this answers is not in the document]
lsf37: I think it was meant to be `irq_t`, and then the second and third variant of those. They are indeed extremely similar, but they do need separate statements to get through the various combinations of casts that the IRQ code throws up (signed int out of enum).

It's possible that these showing up here is a rebase failure, though. I thought we had merged the proofs that deal with IRQ size and casts a while ago (for platform work) and I remember the rebase of that being especially painful. It's strange that they interact with SGI caps. Need to have a closer look.

spec/design/skel/ARM_HYP/ArchInterruptDecls_H.thy
lsf37: Ok, I looked it up: Haskell sometimes imports arch definitions `as Arch` and sometimes `as ArchInv` (presumably when they are invocations). The line here tells the Haskell translator how to translate `ArchInv.` when it sees it.

Which means, the declaration is not necessary in this file, because this is an arch specific file and doesn't need to refer to `ArchInv.`.  Most likely a left-over from importing one of the functions in the wrong place before arch splitting. Will remove.

lsf37: Oh, but it is necessary! Subtle. It's in the types: there is a global `irqcontrol_invocation` and an Arch `irqcontrol_invocation`. `decodeIRQControlInvocation` returns an `Arch.irqcontrol_invocation` (in Haskell `ArchInv.IRQControlInvocation`), not a global one.

It looks like before this PR, the arch version of the type took precedence in Isabelle, so the qualification was not necessary, but now the global one takes precedence in the context of that file, so it picks the wrong one. With qualification we are guaranteed to get the right one, not matter what the context picks first, so I think that is an improvement.

I haven't tracked it down precisely, but `ArchInterruptDecls_H` is now imported earlier in a few places, so there might be different theory merge order going on, which can change the precedence of which name is picked first. Basically it was random before and it is determined now.

proof/invariant-abstract/ARM/ArchEmptyFail_AI.thy
lsf37: At least in this file, we declare all of them only `[wp]` (which this one is as well, from the crunch above). It turns out that `wp` is not enough to get through the proof in this case (neither is `intro`), because `wpsimp` results in multiple cases (conjunction of implications). If the rule is declared as `simp`, it will solve one of the cases and enable further progress.

I'm not too happy with these being in the simp set, though. They should be `wp` or `intro!`, but not `simp` for things to be well behaved. The use of `simp` above isn't really very well behaved. It happens to work in this proof and the one below it, but it's not a good general method to blow away cases in implications via `simp` in a `wpsimp` invocation.

Rewriting those two proof is annoying, though, so I've made the `simp` declaration local for those two only. The rest is all `wp` as standard.

proof/refine/ARM_HYP/CNodeInv_R.thy
lsf37: This whole block is inside `context mdb_swap`. I.e. in arch split later, we'll probably want an arch part of the `mdb_swap` locale. That is where lemmas would then go, with the ones that are commented below for export as the interface of this part.

proof/refine/ARM_HYP/CSpace_R.thy
```diff
@@ -4523,14 +4569,16 @@ lemma arch_update_setCTE_invs:
...
+     \<exists>parent node. m p = Some (CTE parent node) \<and> sameRegionAs parent cap \<and>
+                   (  (\<exists>irq. cap = IRQHandlerCap irq) \<and>
+                      parent = IRQControlCap \<and>
+                      (\<forall>p n'. m p \<noteq> Some (CTE cap n'))
+                    \<or>
+                      parent = IRQControlCap \<and> isArchSGISignalCap cap
+                    \<or>
+                      isUntypedCap parent \<and> descendants_of' p m = {} \<and> capRange cap \<noteq> {} \<and>
+                      capBits cap \<le> capBits parent
+                    )"
```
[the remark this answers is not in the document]
lsf37: Not sure what was going on there, possibly copied, yes. Will fix.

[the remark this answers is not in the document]
lsf37: It's strange on all architectures, including the ones I didn't touch, so at least it wasn't my invention :-)

proof/invariant-abstract/ARM_HYP/ArchCSpacePre_AI.thy
lsf37: `fastforce` does it with `elim!`

proof/crefine/ARM/Interrupt_C.thy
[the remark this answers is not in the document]
lsf37: Ok, I've figured out why these exist in this PR: the irq type length generalisation in C was only applied to AARCH64 initially, but it turned out in the SGI work that it will be needed here as well, so I adjusted the ARM and ARM_HYP decode proofs accordingly (the rest of it was already there). Will give them better names everywhere.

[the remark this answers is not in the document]
lsf37: (in defence of my naming-lemmas honour, the name `maxIRQ_ucast_toEnum_irq_t2` is from 2020 :) But I agree that it should be fixed.)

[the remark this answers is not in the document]
lsf37: Alright, the new names are: `toEnum_unat_irq_t_leq_scast`, `ucast_toEnum_unat_irq_t_leq_ucast`, and `ucast_toEnum_unat_irq_t_leq_scast`.

proof/invariant-abstract/ARM_HYP/ArchBits_AI.thy
lsf37: On further investigation, `is_arch_cap_simps` is used quite rarely on its own. Will check, but we can probably get away with adding `is_SGISignalCap_def` to that set on all architectures. It's not the situation that I feared where page table proofs might unfold unrelated predicates. Will try that out.

lsf37: Turns out, it's even better than that. The manual lemmas on ARM/ARM_HYP are completely unnecessary, they are just repeating the definitions. We can just define `lemmas is_arch_cap_simps = is_pg_cap_def is_pg_cap_def is_pt_cap_def is_SGISignalCap_def` everywhere.

proof/refine/AARCH64/Bits_R.thy
lsf37: I thought I had posted a follow-up to this. If there is a nonsensical answer on another thread, then that's probably where it went.

The block above spells out the unfolded existential for the other `isCap_simps`. For this one that's just the `_def`, so I didn't want to repeat the statement. In terms of arch split, the `isArchSGISignalCap_def` should go into the arch part, of course.

spec/haskell/src/SEL4/Object/Interrupt/AARCH64.hs
```diff
@@ -47,16 +50,41 @@ decodeIRQControlInvocation label args srcSlot extraCaps =
...
+performIRQControl (ArchInv.IssueSGISignal irq targets controlSlot sgiSlot) =
+    withoutPreemption $ do
+        cteInsert (ArchObjectCap (SGISignalCap irq targets)) controlSlot sgiSlot
+        return ()
+
+decodeSGISignalInvocation :: ArchCapability-> KernelF SyscallError ArchInv.Invocation
+decodeSGISignalInvocation cap =
+    case cap of
+        SGISignalCap irq targets -> do
+            return $ ArchInv.InvokeSGISignal $ ArchInv.SGISignalGenerate irq targets
```
corlewis: Very minor but the C uses `target` instead of `targets` and it would be nice to be consistent in whichever makes more sense.

spec/abstract/AARCH64/ArchCSpace_A.thy
```diff
@@ -50,20 +53,31 @@ fun arch_same_region_as :: "arch_cap \<Rightarrow> arch_cap \<Rightarrow> bool"
 | "arch_same_region_as ASIDControlCap c' = (c' = ASIDControlCap)"
 | "arch_same_region_as (ASIDPoolCap r _) c' = (\<exists>r' d'. c' = ASIDPoolCap r' d' \<and> r = r')"
 | "arch_same_region_as (VCPUCap r) c' = (\<exists>r'. c' = VCPUCap r' \<and> r = r')"
+| "arch_same_region_as (SGISignalCap irq targets) c' = (c' = SGISignalCap irq targets)"
 
+text \<open>Check whether giving cap is a descendant of IRQControlCap\<close>
```
corlewis:
```suggestion
text \<open>Check whether given cap is a descendant of IRQControlCap\<close>
```

spec/abstract/ARM/ArchCSpace_A.thy
```diff
@@ -55,8 +59,12 @@ where
 | "arch_same_region_as (PageDirectoryCap r x) (PageDirectoryCap r' x') = (r' = r)"
 | "arch_same_region_as ASIDControlCap ASIDControlCap = True"
 | "arch_same_region_as (ASIDPoolCap r a) (ASIDPoolCap r' a') = (r' = r)"
+| "arch_same_region_as (SGISignalCap irq targets) c' = (c' = SGISignalCap irq targets)"
 | "arch_same_region_as _ _ = False"
 
+text \<open>Check whether giving cap is a descendant of IRQControlCap\<close>
```
corlewis:
```suggestion
text \<open>Check whether given cap is a descendant of IRQControlCap\<close>
```

spec/abstract/ARM_HYP/ArchCSpace_A.thy
```diff
@@ -56,8 +60,12 @@ where
 | "arch_same_region_as ASIDControlCap ASIDControlCap = True"
 | "arch_same_region_as (ASIDPoolCap r a) (ASIDPoolCap r' a') = (r' = r)"
 | "arch_same_region_as (VCPUCap r) (VCPUCap r') = (r' = r)"
+| "arch_same_region_as (SGISignalCap irq targets) c' = (c' = SGISignalCap irq targets)"
 | "arch_same_region_as _ _ = False"
 
+text \<open>Check whether giving cap is a descendant of IRQControlCap\<close>
```
corlewis:
```suggestion
text \<open>Check whether given cap is a descendant of IRQControlCap\<close>
```

spec/abstract/RISCV64/ArchCSpace_A.thy
```diff
@@ -54,6 +54,9 @@ fun arch_same_region_as :: "arch_cap \<Rightarrow> arch_cap \<Rightarrow> bool"
 | "arch_same_region_as ASIDControlCap c' = (c' = ASIDControlCap)"
 | "arch_same_region_as (ASIDPoolCap r _) c' = (\<exists>r' d'. c' = ASIDPoolCap r' d' \<and> r = r')"
 
+text \<open>Check whether giving cap is a descendant of IRQControlCap\<close>
```
corlewis:
```suggestion
text \<open>Check whether given cap is a descendant of IRQControlCap\<close>
```

spec/abstract/X64/ArchCSpace_A.thy
```diff
@@ -71,6 +71,10 @@ where
    (\<exists>frst' lst'. c' = IOPortCap frst' lst' \<and> frst' = frst \<and> lst' = lst)"
 | "arch_same_region_as IOPortControlCap c' = (c' = IOPortControlCap \<or> (\<exists>f l. c' = IOPortCap f l))"
 
+text \<open>Check whether giving cap is a descendant of IRQControlCap\<close>
```
corlewis:
```suggestion
text \<open>Check whether given cap is a descendant of IRQControlCap\<close>
```

spec/abstract/ARM/Arch_Structs_A.thy
```diff
@@ -35,15 +35,20 @@ type_synonym vm_attributes = "vm_attribute set"
 
 section \<open>Architecture-specific capabilities\<close>
 
-text \<open>The ARM kernel supports capabilities for ASID pools and an ASID controller capability,
-along with capabilities for page directories, page tables, and page mappings.\<close>
+text \<open>
+  The ARM kernel supports capabilities for ASID pools and an ASID controller capability,
+  along with capabilities for page directories, page tables, page mappings, software-generated
+  interrupts.\<close>
```
corlewis:
```suggestion
  along with capabilities for page directories, page tables, page mappings, and software-generated
  interrupts.\<close>
```

spec/abstract/ARM_HYP/Arch_Structs_A.thy
```diff
@@ -35,8 +35,10 @@ type_synonym vm_attributes = "vm_attribute set"
 
 section \<open>Architecture-specific capabilities\<close>
 
-text \<open>The ARM kernel supports capabilities for ASID pools and an ASID controller capability,
-along with capabilities for page directories, page tables, and page mappings.\<close>
+text \<open>
+  The ARM kernel supports capabilities for ASID pools and an ASID controller capability,
+  along with capabilities for page directories, page tables, page mappings, software-generated
+  interrupts.\<close>
```
corlewis:
```suggestion
  along with capabilities for page directories, page tables, page mappings, VCPUs, and
  software-generated interrupts.\<close>
```

proof/invariant-abstract/X64/ArchInvariants_AI.thy
```diff
@@ -3371,6 +3371,10 @@ proof -
   thus ?thesis by (rule rtrancl_mono)
 qed
 
+(* unused on this architecture *)
+declare is_irq_control_descendant_def[simp]
+declare should_be_arch_parent_of_def[simp]
```
corlewis, line 3376: Should we do this in RISCV64 as well? It might make it easier to keep the proofs consistent

spec/capDL/Interrupt_D.thy
```diff
@@ -132,6 +140,10 @@ where
         od
   "
 
+(* preforms machine op only *)
```
corlewis:
```suggestion
(* performs machine op only *)
```

spec/haskell/src/SEL4/Object/Interrupt/AARCH64.hs
lsf37: It should be `target` (the first iteration of this was able to address multiple `targets`, hence some of the leftovers). Just searching for `targets` brings up too much noise, but now that I've figured out how to get the file name from searching in diffs, that might work to catch the last remaining ones.

proof/refine/AARCH64/Bits_R.thy
Xaphiosis: That might benefit from a comment. Corey was already confused by the difference between isCap_simps and isCap_defs.

lsf37: ah, yes, I didn't even think of that distinction. Happy to add.

proof/invariant-abstract/X64/ArchInvariants_AI.thy
lsf37: Will look into that, yes.

spec/haskell/src/SEL4/Kernel/VSpace/AARCH64.hs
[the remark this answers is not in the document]
lsf37: Ok, recording here that we discussed this more on MM and decided to go with removing the duplication, because we have a fall-back in the proofs.

PR #735, opened 2024-03-16
crefine: remove unused lemmas; resolve AARCH64 FIXMEs

lsf37:
First pass resolving obvious FIXMEs in AARCH64 proofs:

- questions that have resolved themselves
- lemmas that are unused (some on all architectures)
- trivial lemma moves
- some FIXME updates for easier grep later

### DIFF
--- a/proof/crefine/AARCH64/ADT_C.thy
+++ b/proof/crefine/AARCH64/ADT_C.thy
@@ -161,17 +161,6 @@ where
   "getContext_C thread \<equiv>
    \<lambda>s. from_user_context_C (tcbContext_C (the (clift (t_hrs_' (globals s)) (Ptr &(thread\<rightarrow>[''tcbArch_C''])))))"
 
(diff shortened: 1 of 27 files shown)


proof/crefine/Move_C.thy
```diff
@@ -1493,4 +1493,15 @@ lemma from_to_bool_le_1_idem:
   apply (simp add: word_le_1_and_idem)
   done
 
+(* FIXME AARCH64: this is very specific and rather ugly, is it possible to generalise? *)
```
Xaphiosis, line 1496: kind of odd to move this into generic crefine, no?

Xaphiosis (approved):

proof/crefine/Move_C.thy
lsf37: You're right, this should be in ArchMove_C, not Move_C

PR #736, opened 2024-03-16
word_lib+crefine: move lemmas to Word_Lib

lsf37:
- move AARCH64 word lemmas to the holding area in Word_Lib
- generalise word_ctz_upcast_id and provide word_ctz_upcast_id_32_64

### DIFF
--- a/lib/Word_Lib/Word_Lemmas_64_Internal.thy
+++ b/lib/Word_Lib/Word_Lemmas_64_Internal.thy
@@ -57,4 +57,13 @@ lemmas mask_64_id[simp] = mask_len_id[where 'a=64, folded word_bits_def]
 lemma neq_0_unat: "x \<noteq> 0 \<Longrightarrow> 0 < unat x" for x::machine_word
   by (simp add: unat_gt_0)
 
+(* The 32-bit version is occasionally needed on 64-bit platforms *)
+lemma word_rsplit_0_32:
+  "word_rsplit (0::32 word) = [0, 0, 0, (0::8 word)]"
+  by (simp add: word_rsplit_def bin_rsplit_def word_bits_def word_size_def Cons_replicate_eq)
+
+lemma word_ctz_upcast_id_32_64:
+  "x \<noteq> 0 \<Longrightarrow> word_ctz (UCAST(32 \<rightarrow> 64) x) = word_ctz x"
+  by (simp add: word_ctz_upcast_id is_up)
+
 end
--- a/proof/crefine/Move_C.thy
+++ b/proof/crefine/Move_C.thy
@@ -1480,17 +1480,4 @@ lemma multiple_add_less_nat:
   apply simp
   done
 
-(* FIXME AARCH64 move to wherever we keep the from_bool/to_bool lemmas *)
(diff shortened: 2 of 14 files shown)


lib/Word_Lib/Word_Lemmas_64_Internal.thy
```diff
@@ -57,4 +57,13 @@ lemmas mask_64_id[simp] = mask_len_id[where 'a=64, folded word_bits_def]
 lemma neq_0_unat: "x \<noteq> 0 \<Longrightarrow> 0 < unat x" for x::machine_word
   by (simp add: unat_gt_0)
 
+(* The 32-bit version is occasionally needed on 64-bit platforms *)
+lemma word_rsplit_0_32:
+  "word_rsplit (0::32 word) = [0, 0, 0, (0::8 word)]"
+  by (simp add: word_rsplit_def bin_rsplit_def word_bits_def word_size_def Cons_replicate_eq)
+
+lemma word_ctz_upcast_id_32_64:
+  "x \<noteq> 0 \<Longrightarrow> word_ctz (UCAST(32 \<rightarrow> 64) x) = word_ctz x"
```
Xaphiosis, line 66: there's an argument to be made that this is true of any upcast

proof/crefine/Move_C.thy
```diff
@@ -1480,19 +1480,6 @@ lemma multiple_add_less_nat:
   apply simp
   done
 
-(* FIXME AARCH64 move to wherever we keep the from_bool/to_bool lemmas *)
-lemma word_le_1_and_idem:
```
Xaphiosis, line 1484: can you confirm that Word_Lemmas_Internal are where we keep our from_bool/to_bool lemmas?

Xaphiosis (approved):
I would prefer the commit message short-line reflect that this is AARCH64-cleanup related:
"word_lib+crefine: move AARCH64 lemmas to Word_Lib" or something

lib/Word_Lib/Word_Lemmas_64_Internal.thy
lsf37: That's why I made `word_ctz_upcast_id` and use it here to prove this one :-). Just wanted to preserve the behaviour the existing proof expects and remove the additional `is_up` condition the generic one introduces.

proof/crefine/Move_C.thy
lsf37: Word_Lemmas_Internal is the area to commit Word_Lib lemmas to that should be moved into the rest of Word_Lib at the next sync with the AFP. The from_bool/to_bool lemmas are deeper within Word_Lib, but should only move there at the sync point (otherwise we risk accidentally removing things when we sync).

Xaphiosis: Oki, but then do you want to remove or keep the comment to move it near from/to_bool?

lsf37: `word_le_1_and_idem` should not actually go there, `from_to_bool_le_1_idem` should. `word_le_1_and_idem` should go where `AND` lemmas are.

PR #737, opened 2024-03-19
Yet more rules for `Lib`

michaelmcinerney:


lib/Heap_List.thy
```diff
@@ -383,6 +383,14 @@ lemma not_head_prev_not_None:
   using sym_heap_prev_None_is_start heap_path_head
   by fastforce
 
+lemma heap_ls_neighbour_in_set:
+  "\<lbrakk>heap_ls hp st xs; sym_heap hp hp'; st \<noteq> None \<longrightarrow> hp' (the st) = None; p \<in> set xs\<rbrakk>
+   \<Longrightarrow> \<forall>nbr. (hp p = Some nbr \<longrightarrow> nbr \<in> set xs) \<and> (hp' p = Some nbr \<longrightarrow> nbr \<in> set xs)"
```
lsf37, line 388: Is the object logic `∀` necessary or could `nbr` just be a free variable?

lib/Lib.thy
```diff
@@ -2533,8 +2533,18 @@ next
...
+lemma rsubst2:
+  "\<lbrakk>P s x; s = t; x = y\<rbrakk> \<Longrightarrow> P t y"
+  by simp
+
+lemma rsubst3:
+  "\<lbrakk>P a b c ; a = s; b = t; c = u\<rbrakk> \<Longrightarrow> P s t u"
+  by simp
+
+lemma rsubst4:
+  "\<lbrakk>P a b c d; a = s; b = t; c = u; d = v\<rbrakk> \<Longrightarrow> P s t u v"
```
lsf37, line 2547: It'd be good the get the naming consistent on these. I can see the reasoning for `rsubst3` and `rsubst4` with a, b for the first and s, t for the second list of arguments. We could do the same for `rsubst2`. Changing the old `rsubst` is probably a lost cause, and we do want similarity with `subst`. What do people think?

lib/HaskellLemmaBucket.thy
```diff
@@ -127,10 +127,14 @@ lemma empty_fail_stateAssert[intro!, simp]:
   "empty_fail (stateAssert P l)"
   unfolding stateAssert_def by simp
 
-lemma haskell_assert_wp:
+lemma haskell_assert_wp[wp]:
```
lsf37, line 130: It does lead to a bunch of new `hoare_drop_imps` and `crunch_wps`, but I think I agree with the overall strategy of making sure that automation doesn't lose information. Accidental `haskell_assert_inv` uses are very hard to debug.

lsf37 (approved):

lsf37:
Looks like there are still a few proof failures in the Orhpanage session. Probably missing a `crunch_wps` or `hoare_drop_imps` somewhere.

michaelmcinerney:
Sorry to force push - I didn't realise that reviews had begun. The changes I made were just to the `Orphanage.thy` files, within the commit for `haskell_assert_wp`

lib/Lib.thy
michaelmcinerney: I haven't looked much into how these are used for MCS, and I haven't given too much thought to this, but I should at least point out now that these rules (including the reformulation of `rsubst` into a `lemmas` statement) are copied straight from the `rt` branch

lib/Heap_List.thy
michaelmcinerney: This could probably be changed to have a free variable. This is really only used to prove a variant that has `list_queue_relation` in the assumptions, instead of `heap_ls` (and because it has `list_queue_relation`, which currently lives in `TcbAcc_R`, I've left it in Refine, though we could maybe move that to `Lib` too if we really wanted to). That variant is used in a fair few places with `cut_tac` and `frule`, the idea being that we then have a fairly general statement sitting around in our assumptions that `fastforce` can go to town with. I think we don't want a fixed free variable for that rule, because we want to reason about both previous pointers and next pointers

lib/HaskellLemmaBucket.thy
michaelmcinerney: I should probably have also mentioned that this is what's done in the `rt` branch, and that Corey says there in the PR that made it `[wp]` that it mirrors what's done with the other wp rules like `assert_wp`

lsf37: Yes, I'm happy with the `[wp]` for this one, was just thinking out loud.

lib/Lib.thy
lsf37: Understood -- we should take the opportunity to make these better. The `rt` branch does not historically have very high library standards in terms of thought put into naming, styles, etc, because that wasn't a priority at the time. We do want to invest those kinds of considerations on `master`, though, because they will become ingrained when they land there.

lib/Heap_List.thy
lsf37: If there's a good reason, it's fine and if  it's used mostly as `drule`/`frule` and we want to delay instantiation then that sounds good to me.

lib/Lib.thy
michaelmcinerney: I think if we went for `a, b` and `s, t` for `rsubst2`, then that would be best

lib/HaskellLemmaBucket.thy
Xaphiosis: Is this related to the long-standing annoyance of how in the abstract spec the default wp rule would keep the knowledge of what was asserted, but in the Haskell it'd just get dropped?

michaelmcinerney: Yes I think so. I think previously, `haskell_assert_inv` was generated and added to the `[wp]` set by a `crunch` somewhere and was picked up by many proofs, and my guess is that now that `haskell_assert_wp` is added to the `[wp]` set in `Lib`, that rule is now preferred and won't lose information.

lib/Lib.thy
lsf37: Yup, happy with that.

lib/Corres_UL.thy
```diff
@@ -1176,6 +1176,18 @@ lemma corres_move_asm:
 
 lemmas corres_cross_over_guard = corres_move_asm[rotated]
 
+lemma corres_cross_add_guard:
+  "\<lbrakk> \<And>s s'. \<lbrakk>(s,s') \<in> sr; P s; P' s'\<rbrakk> \<Longrightarrow> Q s';
+     corres_underlying sr nf nf' r P  (P' and Q) f g\<rbrakk>
```
Xaphiosis, line 1181: extra space before `(P'`, also wouldn't `Q'` make more sense than `Q` given it's on the concrete side?

proof/refine/AARCH64/Syscall_R.thy
```diff
@@ -2086,7 +2086,7 @@ crunches handleVMFault
 
 crunches handleHypervisorFault
   for ksit[wp]: "\<lambda>s. P (ksIdleThread s)"
-  (wp: undefined_valid simp: isFpuEnable_def)
+  (wp: undefined_valid haskell_assert_inv simp: isFpuEnable_def)
```
Xaphiosis, line 2089: didn't you already put `haskell_assert_inv` into `[wp]`?

michaelmcinerney: `haskell_assert_wp` is in the `[wp]` set

lib/Corres_UL.thy
michaelmcinerney: Yes it would make much more sense to use `Q'`. This is another rule copied straight over from the `rt` branch. Should I change it here? If I do change it for `master`, should I change it on the `rt` branch too? That would be quite annoying, because there would be many rules which do write `where Q=blah`

proof/refine/AARCH64/Syscall_R.thy
Xaphiosis: Right! my bad.

lib/Corres_UL.thy
Xaphiosis: Do we use `corres_cross_add_guard` that often? Or are you saying there are other rules which also followed the `Q` scheme for the concrete side?

michaelmcinerney: `corres_cross_add_guard` is used dozens and dozens of times on the `rt` branch. It wouldn't be horrible to change, but it would be a bit of a nuisance

Xaphiosis: If it's only `corres_cross_add_guard` I'd vote for updating it and its uses, but if other lemmas follow the same scheme, I'd say defer to another round.

michaelmcinerney: OK I will update it on both branches

lib/Corres_UL.thy
```diff
@@ -1176,6 +1176,18 @@ lemma corres_move_asm:
...
+lemma corres_cross_add_guard:
+  "\<lbrakk> \<And>s s'. \<lbrakk>(s,s') \<in> sr; P s; P' s'\<rbrakk> \<Longrightarrow> Q' s';
+     corres_underlying sr nf nf' r P (P' and Q') f g\<rbrakk>
+    \<Longrightarrow> corres_underlying sr nf nf' r P P' f g"
+  by (fastforce simp: corres_underlying_def)
+
+lemma corres_cross_add_abs_guard:
+  "\<lbrakk> \<And>s s'. \<lbrakk>(s,s') \<in> sr; P s; P' s'\<rbrakk> \<Longrightarrow> Q s;
+     corres_underlying sr nf nf' r (P and Q)  P' f g\<rbrakk>
+    \<Longrightarrow> corres_underlying sr nf nf' r P P' f g"
```
Xaphiosis, line 1188: ah, I only now noticed this, shouldn't the `⟹` align with the `⟦`?

lib/Corres_UL.thy
Xaphiosis: Cheers

Xaphiosis (approved):

lib/Corres_UL.thy
michaelmcinerney: There are quite a few lemmas in this theory file with this same problem. Should I fix up just this one, or all of them?

michaelmcinerney: Well, I can at least fix it in the lemmas I've introduced in this PR. But the one above is old and wrong, and there are other instances of this elsewhere

Xaphiosis: For this round, I'd stick to the ones you're adding, but we'd welcome a future PR that consolidates the others.

Xaphiosis:
Too late now, but one of the commit messages has an extra:
```
reindent corres_cross_add_guard and friend squash

Signed-off-by: Michael McInerney <michael.mcinerney@proofcraft.systems>
```

PR #738, opened 2024-03-19
Enlarge the prefix_refinement rule set

corlewis:
This roughly follows the structure and naming conventions of the `corres_underlying` rule set, as in `Corres.thy` and `ExtraCorres.thy`.

### DIFF
--- a/lib/Monads/trace/Trace_Monad_Equations.thy
+++ b/lib/Monads/trace/Trace_Monad_Equations.thy
@@ -10,7 +10,9 @@
 
 theory Trace_Monad_Equations
   imports
+    Trace_Empty_Fail
     Trace_No_Fail
+    Trace_No_Trace
 begin
 
 lemmas assertE_assert = assertE_liftE
@@ -214,7 +216,7 @@ lemma bind_return_unit:
   by simp
 
 lemma modify_id_return:
- "modify id = return ()"
+  "modify id = return ()"
   by (simp add: simpler_modify_def return_def)
 
 lemma liftE_bind_return_bindE_returnOk:
@@ -244,6 +246,38 @@ lemma zipWithM_x_modify:
   apply (rule ext)
   apply (simp add: simpler_modify_def bind_def split_def)
   done
+
+lemma bind_return_subst:
+  assumes r: "\<And>r. \<lbrace>\<lambda>s. P x = r\<rbrace> f x \<lbrace>\<lambda>rv s. Q rv = r\<rbrace>"
+  shows
+  "do a \<leftarrow> f x;
+      g (Q a)
+   od =
+   do _ \<leftarrow> f x;
+      g (P x)
+   od"
+proof -
+  have "do a \<leftarrow> f x;
+           return (Q a)
+        od =
+        do _ \<leftarrow> f x;
+           return (P x)
+        od"
+    using r
+    apply (subst fun_eq_iff)
+    apply (auto simp: bind_def valid_def return_def mres_def vimage_def split: tmres.splits;
+           fastforce simp: image_def intro: rev_bexI)
+    done
+  hence "do a \<leftarrow> f x;
+            return (Q a)
+         od >>= g =
+         do _ \<leftarrow> f x;
+            return (P x)
+         od >>= g"
+    by (rule bind_cong, simp)
+  thus ?thesis
+    by simp
+qed
 
 lemma assert2:
   "(do v1 \<leftarrow> assert P; v2 \<leftarrow> assert Q; c od)
@@ -306,6 +340,28 @@ lemma if_bind:
 lemma bind_liftE_distrib: "(liftE (A >>= (\<lambda>x. B x))) = (liftE A >>=E (\<lambda>x. liftE (\<lambda>s. B x s)))"
   by (clarsimp simp: liftE_def bindE_def lift_def bind_assoc)
 
+(*FIXME: the following lemmas were originally solved by monad_eq, which doesn't yet exist for the
+         trace monad due to traces making equality more complicated.*)
+lemma condition_apply_cong:
+  "\<lbrakk> c s = c' s'; s = s'; \<And>s. c' s \<Longrightarrow> l s = l' s  ; \<And>s. \<not> c' s \<Longrightarrow> r s = r' s  \<rbrakk> \<Longrightarrow>  condition c l r s = condition c' l' r' s'"
+  by (simp add: condition_def)
+
+lemma condition_cong [cong, fundef_cong]:
+  "\<lbrakk> c = c'; \<And>s. c' s \<Longrightarrow> l s = l' s; \<And>s. \<not> c' s \<Longrightarrow> r s = r' s \<rbrakk> \<Longrightarrow> condition c l r = condition c' l' r'"
+  by (simp add: condition_def fun_eq_iff)
+
+lemma lift_Inr [simp]: "(lift X (Inr r)) = (X r)"
+  by (simp add: lift_def)
+
+lemma lift_Inl [simp]: "lift C (Inl a) = throwError a"
+  by (simp add: lift_def)
+
+lemma returnOk_def2:  "returnOk a = return (Inr a)"
+  by (simp add: returnOk_def)
+
+lemma liftE_fail[simp]: "liftE fail = fail"
+  by (simp add: liftE_def)
+
 lemma if_catch_distrib:
   "((if P then f else g) <catch> h) = (if P then f <catch> h else g <catch> h)"
   by (simp split: if_split)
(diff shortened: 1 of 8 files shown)


lib/Monads/trace/Trace_Monad_Equations.thy
```diff
@@ -306,6 +340,28 @@ lemma if_bind:
 lemma bind_liftE_distrib: "(liftE (A >>= (\<lambda>x. B x))) = (liftE A >>=E (\<lambda>x. liftE (\<lambda>s. B x s)))"
   by (clarsimp simp: liftE_def bindE_def lift_def bind_assoc)
 
+(*FIXME: the following lemmas were originally solved by monad_eq, which doesn't yet exist for the
+         trace monad due to traces making equality more complicated.*)
```
lsf37, line 344: I think that's perfectly acceptable for the trace monad (not using `monad_eq`). Did you gain an impression on whether it would be feasible to update `monad_eq` in a way that would make sense for the trace monad? It's not used that much in the main refinement proofs, so it wouldn't be big loss, but if the work sparked an insight, it'd be good to know.

lib/Monads/trace/Trace_Monad_Equations.thy
```diff
@@ -328,10 +384,115 @@ lemma catch_is_if:
...
   "\<lbrakk>f = g; a s = b s\<rbrakk> \<Longrightarrow> (a >>= f) s = ((b >>= g) s)"
   by (simp add:bind_def)
 
+lemma double_gets_drop_regets:
+  "(do x \<leftarrow> gets f;
+       xa \<leftarrow> gets f;
+       m xa x
+    od) =
+   (do xa \<leftarrow> gets f;
+       m xa xa
```
lsf37, line 420: would prefer `x` and `y` instead of `x` and `xa`

lib/concurrency/Triv_Refinement.thy
```diff
@@ -1,42 +1,40 @@
...
-where
+text \<open>
+  This is a simple (almost trivial) definition of refinement, which simply resolves nondeterminism
+  to a smaller set of options.\<close>
+definition triv_refinement :: "('s,'a) tmonad \<Rightarrow> ('s,'a) tmonad \<Rightarrow> bool" where
   "triv_refinement aprog cprog = (\<forall>s. cprog s \<subseteq> aprog s)"
 
+lemmas triv_refinement_elemD = triv_refinement_def[THEN iffD1, rule_format, THEN subsetD]
+
+lemma triv_refinement_trans:
```
lsf37, line 21: Not sure if we did this before, but it'd be nice to have that `triv_refinement` is a partial order, i.e. reflexive, transitive, and and antisymmetric. We could even instantiate it into the order class which would make a bunch of consequences of that available. Not that we are likely to need it much, but it'd be a nice setup.

lsf37: I'm also almost tempted to rename this into `trace_refinement` if that name is still available. Because it's not trivial at all, it is actually the classical version of refinement on traces (it's just not data refinement).

lib/concurrency/examples/Peterson_Atomicity.thy
```diff
@@ -213,57 +182,56 @@ lemma reflp_peterson_rel[simp]:
 declare reflp_peterson_rel[THEN reflpD, simp]
 
 lemma peterson_rel_imp_assume_invs:
-  "invs x \<Longrightarrow> (peterson_rel ident x y \<and> invs x \<and> invs y \<longrightarrow> P x y)
-    \<Longrightarrow> (peterson_rel ident x y \<longrightarrow> P x y)"
+  "\<lbrakk>invs x; (peterson_rel ident x y \<and> invs x \<and> invs y \<longrightarrow> P x y)\<rbrakk>
+   \<Longrightarrow> (peterson_rel ident x y \<longrightarrow> P x y)"
```
lsf37, line 186:
```suggestion
  "\<lbrakk>invs x; peterson_rel ident x y \<and> invs x \<and> invs y \<longrightarrow> P x y\<rbrakk>
   \<Longrightarrow> peterson_rel ident x y \<longrightarrow> P x y"
```

lib/concurrency/examples/Peterson_Atomicity.thy
```diff
@@ -67,136 +65,107 @@ locale mx_locale =
...
-  enabled :: "ident \<Rightarrow> ('a, 'b) p_state \<Rightarrow> bool"
-where
-  "enabled ident s \<equiv> (critical (ab_label s ident)
-    \<or> (ab_label s ident = Awaiting \<and> t_v s = ident))"
+text \<open>
+  The key invariant. We can't both be enabled, where that means
+  either we're in the critical section or waiting to enter with priority.\<close>
+abbreviation (input) enabled :: "ident \<Rightarrow> ('a, 'b) p_state \<Rightarrow> bool" where
+  "enabled ident s \<equiv>
+     (critical (ab_label s ident) \<or> (ab_label s ident = Awaiting \<and> t_v s = ident))"
```
lsf37, line 142:
```suggestion
     critical (ab_label s ident) \<or> ab_label s ident = Awaiting \<and> t_v s = ident"
```

lib/concurrency/Prefix_Refinement.thy
```diff
@@ -1,101 +1,90 @@
...
-For a concrete program to refine an abstract program, then for every
-trace of the concrete program there must exist a well-formed fragment
-of the abstract program that matches (according to the simulation
-relation) but which leaves enough decisions available to the abstract
-environment to permit parallel composition.
-\<close>
+  This is a notion of refinement/simulation making use of prefix closure.
+  For a concrete program to refine an abstract program, then for every
+  trace of the concrete program there must exist a well-formed fragment
+  of the abstract program that matches (according to the simulation
```
lsf37, line 22:
```suggestion
  For a concrete program to refine an abstract program, for every
  trace of the concrete program there must exist a well-formed fragment
  of the abstract program that matches (according to the simulation
```

lib/concurrency/Prefix_Refinement.thy
```diff
@@ -106,74 +95,218 @@ lemmas is_matching_fragment_env_closed = is_matching_fragment_wf[THEN conjunct2,
...
-  apply (clarsimp simp add: in_fst_snd_image)
+  apply (clarsimp simp: in_fst_snd_image)
+  done
+
+\<comment> \<open>FIXME: it would be good to show this but the last assumption needs to be a lot more complicated
+          and the nondet equivalent isn't used anywhere.
+lemma prefix_refinement_propagate_no_fail:
+  "\<lbrakk>prefix_refinement sr isr osr rvr AR R P Q f f';
+    no_fail P' f; \<forall>t0 t. Q t0 t \<longrightarrow> (\<exists>s0 s. P s0 s \<and> sr s0 t0 \<and> isr s t)\<rbrakk>
+   \<Longrightarrow> no_fail Q' f'"
```
lsf37, line 172: It's interesting that this doesn't fall out easily. It's one of the fundamental properties of refinement. We don't make explicit use of it, but it's a good sanity check and it could be a sign that our definition of data refinement is still a bit too complex. No need to resolve this in the short term, but maybe there is a tweak or two yet to be discovered that might make things easier in the future. (It's not unexpected that there would be side conditions since the type for programs is a lot tighter in the nondet monad, whereas for the trace monad there are many traces that don't really make sense as programs. So maybe it's just a reflection of that).

lib/concurrency/Prefix_Refinement.thy
```diff
@@ -256,79 +389,267 @@ corollary matching_fragment_matching_tr_trivR:
...
+    F \<Longrightarrow> prefix_refinement sr isr osr rvr AR R P Q f g\<rbrakk>
+   \<Longrightarrow> prefix_refinement sr isr osr rvr AR R P Q f g"
+  by (auto simp: prefix_refinement_def)
+
+\<comment> \<open>FIXME: is it more usable to have this lemma combined like this or to split it up like Corres_UL's version\<close>
+lemma prefix_refinement_gen_asm:
+  "\<lbrakk>P \<Longrightarrow> prefix_refinement sr isr osr rvr AR R P' Q' f g\<rbrakk>
+   \<Longrightarrow> prefix_refinement sr isr osr rvr AR R (P' and (\<lambda>_ _. P)) Q' f g"
+  "\<lbrakk>P \<Longrightarrow> prefix_refinement sr isr osr rvr AR R P' Q' f g\<rbrakk>
+   \<Longrightarrow> prefix_refinement sr isr osr rvr AR R P' (Q' and (\<lambda>_ _. P)) f g"
```
lsf37, line 600: Depends on how this is used. If it is used with instantiation, then having them separate is better. If it is used by matching against a current goal to extract `P`, then a single name is better. We could get both but providing two separate names and a `lemmas` statement for the collection.

lib/concurrency/Prefix_Refinement.thy
```diff
@@ -661,22 +930,23 @@ lemma is_matching_fragment_UNION:
   apply blast
   done
 
-definition
-  mbind :: "('s, 'a) tmonad \<Rightarrow> ('s \<Rightarrow> 'a \<Rightarrow> ('s, 'b) tmonad) \<Rightarrow>
-           's \<Rightarrow> ('s, 'b) tmonad"
+definition mbind ::
+  "('s, 'a) tmonad \<Rightarrow> ('s \<Rightarrow> 'a \<Rightarrow> ('s, 'b) tmonad) \<Rightarrow> 's \<Rightarrow> ('s, 'b) tmonad"
```
lsf37, line 934: If you remember what this one does, it would be great to add a comment, because I have forgotten :-). I'm sure it'll be clear from context when you're inside Isabelle, but it'll probably speed up understanding if we have a comment here.

lib/concurrency/Prefix_Refinement.thy
```diff
@@ -938,163 +1189,614 @@ theorem prefix_refinement_bind_general[rule_format]:
...
-lemma default_prefix_refinement_ex_match_iosr_R:
-  "is_matching_fragment sr osr rvr ctr cres s0 R s
-            (rely (\<lambda>s. aprog s \<inter> ({tr'. matching_tr_pfx iosr tr' ctr} \<times> UNIV)) R s0)
-    \<Longrightarrow> \<exists>f. is_matching_fragment sr osr rvr ctr cres s0 R s f
-            \<and> triv_refinement aprog f"
-  apply (intro exI conjI, assumption)
-  apply (clarsimp simp add: triv_refinement_def rely_def)
+\<comment> \<open>FIXME: an equivalent doesn't seem to exist for corres, do we want it to or does it mean we
+          shouldn't bother with this?\<close>
+lemma prefix_refinement_handle_elseE:
```
lsf37, line 1314: I'd keep the rule here. Just because it's not used in the kernel at the moment, doesn't mean we might not need it later, so in terms of library completeness, it should stay. Might be good to add a reminder to add to corrres as well, but not high priority for right now.

lib/concurrency/Prefix_Refinement.thy
```diff
@@ -938,163 +1189,614 @@ theorem prefix_refinement_bind_general[rule_format]:
...
+  unfolding whenE_def returnOk_def
+  apply clarsimp
+  apply (rule prefix_refinement_return_imp)
+  apply simp
+  done
+
+\<comment> \<open>FIXME: an equivalent doesn't seem to exist for corres, do we want it to or does it mean we
+          shouldn't bother with this?\<close>
+lemma prefix_refinement_unless:
+  "\<lbrakk>G = G'; prefix_refinement sr isr isr rvr AR R P Q a c; rvr () ()\<rbrakk>
```
lsf37, line 1409: That would be really good to add to corres. We always unfold `unless_def` for no good reason.

lib/concurrency/Prefix_Refinement.thy
```diff
@@ -938,163 +1189,614 @@ theorem prefix_refinement_bind_general[rule_format]:
...
+   \<Longrightarrow> prefix_refinement sr isr isr rvr AR R (\<lambda>x y. \<not> G \<longrightarrow> P x y) (\<lambda>x y. \<not> G' \<longrightarrow> Q x y)
+         (unless G a) (unless G' c)"
+  by (simp add: unless_def prefix_refinement_when)
+
+\<comment> \<open>FIXME: an equivalent doesn't seem to exist for corres, do we want it to or does it mean we
+          shouldn't bother with this?\<close>
+lemma prefix_refinement_unlessE:
```
lsf37, line 1416: same as above: good to add to `corres`

lib/concurrency/Prefix_Refinement.thy
```diff
@@ -938,163 +1189,614 @@ theorem prefix_refinement_bind_general[rule_format]:
...
+    \<lbrakk>\<not>G; \<not>G'\<rbrakk> \<Longrightarrow> prefix_refinement sr isr osr rvr AR R P' Q' b d \<rbrakk>
+   \<Longrightarrow> prefix_refinement sr isr osr rvr AR R
+               (P'' and (if G then P else P')) (Q'' and (if G' then Q else Q'))
+               (if G then a else b) (if G' then c else d)"
+  by (fastforce simp: prefix_refinement_def)
+
+\<comment> \<open>FIXME: corres equivalent isn't used, keep or delete?\<close>
+lemmas prefix_refinement_if_strong' =
+  prefix_refinement_if_strong[where P=P and P'=P and P''=P for P,
+                              where Q=Q and Q'=Q and Q''=Q for Q, simplified]
```
lsf37, line 1450: I have the feeling this one can go.

lib/concurrency/Prefix_Refinement.thy
```diff
@@ -938,163 +1189,614 @@ theorem prefix_refinement_bind_general[rule_format]:
...
+
+
+\<comment> \<open>FIXME: Put more thought into whether we want this section, and if not what alternative rules
+          would we want. The comment copied from Corres_UL suggests we might not want them.
+          They would be a fair bit more complicated to prove for prefix_refinement.\<close>
+section \<open>Some equivalences about liftM and other useful simps\<close>
```
lsf37, line 1456: Good to remove this section -- even if they were easy to prove, it'd be better not to introduce this temptation for prefix refinement.

lib/concurrency/Prefix_Refinement.thy
```diff
@@ -1103,36 +1805,273 @@ lemma prefix_refinement_select:
...
+    apply (rule prefix_refinement_bind[where rvr'="rel_option rvr"])
+       apply (rule prefix_refinement_gets[where P=P and Q=Q])
+       apply simp
+      apply (erule prefix_refinement_assert_opt)
+     apply wpsimp+
+  done
+
+\<comment> \<open>FIXME: an equivalent doesn't seem to exist for corres, do we want it to or does it mean we
+          shouldn't bother with this?\<close>
+lemma prefix_refinement_gets_map:
```
lsf37, line 1836: I think we want it. Projections are even more likely to be useful in the trace monad.

lib/concurrency/Prefix_Refinement.thy
```diff
@@ -1103,36 +1805,273 @@ lemma prefix_refinement_select:
...
+  "\<lbrakk>\<And>s t. \<lbrakk>iosr s t; P s; Q t\<rbrakk> \<Longrightarrow> rel_option rvr (f s p) (g t q)\<rbrakk>
+   \<Longrightarrow> prefix_refinement sr iosr iosr rvr AR R (\<lambda>_. P) (\<lambda>_. Q) (gets_map f p) (gets_map g q)"
+  apply (subst gets_the_oapply_comp[symmetric])+
+  apply (rule prefix_refinement_gets_the)
+  apply simp
+  done
+
+\<comment> \<open>FIXME: an equivalent doesn't seem to exist for corres, do we want it to or does it mean we
+          shouldn't bother with this?\<close>
+lemma prefix_refinement_throw_opt:
```
lsf37, line 1846: would keep

lib/concurrency/Prefix_Refinement.thy
```diff
@@ -1103,36 +1805,273 @@ lemma prefix_refinement_select:
...
+                matching_tr_pfx_def matching_tr_def)
+
+lemma triv_refinement_restrict:
+  "triv_refinement f (\<lambda>s'. if s'=s then f s else {})"
+  by (clarsimp simp: triv_refinement_def)
+
+\<comment> \<open>FIXME: corres rules for condition don't exist, so maybe we don't bother with this. It feels
+          like it shouldn't be this hard to prove, but providing correct fragments is frustrating.
+          It might make it easier to instantiate if the exists was wrapped in a new definition.\<close>
+lemma prefix_refinement_condition_strong:
```
lsf37, line 1906: `condition` is mostly used in AutoCorres, so there is probably no need to add a complete rule set for it here.

lib/concurrency/Prefix_Refinement.thy
```diff
@@ -1103,36 +1805,273 @@ lemma prefix_refinement_select:
...
+  unfolding notM_def
+  apply pfx_refn_pre
+    apply (erule prefix_refinement_bind)
+      apply (rule prefix_refinement_returnTT)
+      apply wpsimp+
+  done
+
+\<comment> \<open>FIXME: an equivalent doesn't seem to exist for corres, do we want it to or does it mean we
+          shouldn't bother with this?\<close>
+lemma whenM_prefix_refinement:
```
lsf37, line 2038: should add to `corres` as well (lemma might exist on rt branch)

lsf37 (approved):
Excellent. This is what I had in mind when we started this.

lib/Monads/trace/Trace_Monad_Equations.thy
corlewis: I think it might be doable but it would be complicated.

The main issue is that for the nondet monad, equality can be nicely separated into `in_monad` style set membership and equality of the failure flag, both of which are resolved by rules we want anyway.

However, the trace monad version of `in_monad` only handles completed monad results, it doesn't say anything about the trace that leads to it or about `Failed` or `Incomplete` results. My guess is that with lemmas about those as well then we could update `monad_eq` to work for the trace monad, but there's a good chance that that would end up being more effort than it saves.

lib/concurrency/Prefix_Refinement.thy
corlewis: Unfortunately, this is something that Tom did a long time ago and that I've never had to really touch. I can say that it's a slight variant on `bind`, where the last state of the trace is also used by the `g`, and that it is used to build up the fragment required for proving `prefix_refinement_bind_general`, but beyond that I don't have a good intuition for it.

lsf37: No worries, I didn't remember at all :-) That last bit "used to build up the fragment required for proving prefix_refinement_bind_general" is probably useful for a comment.

lib/concurrency/Prefix_Refinement.thy
corlewis: This was one of the things I tried to prove pretty early on in the process of working on all this, so it might be easier now that I have a better understanding of the `prefix_refinement` definition. I was sort of thinking to leave it though, until we have a chance to put more thought into how we want to handle failure. Currently it is very different to how `corres` handles it.

corlewis: Having played around with this a bit more, it's actually a bit more complicated than I was previously thinking. The current definition allows us to show refinement between a non-failing abstract function and a failing concrete function, as long as all of the failing traces in the concrete function do not satisfy the rely.

It is almost definitely possible to resolve this, it will just require a bit of thought and possible some worked examples to find the nicest way off expressing things.

lib/concurrency/Prefix_Refinement.thy
corlewis: It looks like the `corres` version is used both ways, so I've split this up and added `lemmas prefix_refinement_gen_asms = ...` like you suggested.

lib/concurrency/Prefix_Refinement.thy
corlewis: Could we want non-simp rules for use when the return relation is already instantiated?

lib/concurrency/Prefix_Refinement.thy
```diff
@@ -938,163 +1189,614 @@ theorem prefix_refinement_bind_general[rule_format]:
...
+  apply (clarsimp simp: prefix_refinement_def)
+  apply (rule_tac x="prefix_close (\<lambda>s'. if s'=s then {(a,b)} else {})" in exI)
+  apply (clarsimp simp: triv_refinement_def)
+apply (auto simp: is_matching_fragment_def prefix_closed_def self_closed_def env_closed_def
+                matching_tr_pfx_def matching_tr_def)[1]
+oops\<close>
+
+\<comment> \<open>FIXME: not sure why this isn't using \<top>\<top> (or dc)\<close>
+abbreviation (input)
   "dc2 \<equiv> (\<lambda>_ _. True)"
```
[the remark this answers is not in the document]
corlewis: Very easy to remove, the only concern is that I'm not sure why it was added in the first place.

lib/concurrency/Prefix_Refinement.thy
corlewis: I might keep it for now, as an example of a difficult proof which might guide improvements to the definition of `prefix_refinement` and its helper lemmas.

lib/concurrency/Prefix_Refinement.thy
lsf37: No need to do this now. As long as it's flagged we can do this at any time later.

lib/Monads/trace/Trace_Monad_Equations.thy
lsf37: Ok, we should definitely defer until we know we want this then, esp since `monad_eq` is not used that often.

lib/concurrency/Prefix_Refinement.thy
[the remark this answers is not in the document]
lsf37: Esp as an input abbreviation, there is really no point introducing it -- maybe to make it look more familiar. I'd actually rather get rid of `dc` as well, it has been the source of much annoyance.

PR #739, opened 2024-03-21
AArch64 fastpath proofs

Xaphiosis:
Before I do more cleanup, I think this could use some feedback. Please ignore the horrid commit history (for now), but advice welcome on whether I should squash the actual fastpath commits into 3 parts (defs, rewrite, C refinement), or only one commit.

In any case the proofs work and look better than the ARM/ARM_HYP ones at least.

### DIFF
--- a/lib/MonadicRewrite.thy
+++ b/lib/MonadicRewrite.thy
@@ -524,13 +524,19 @@ lemma monadic_rewrite_assert2:
    \<Longrightarrow> monadic_rewrite F E ((\<lambda>s. Q \<longrightarrow> P s) and (\<lambda>_. Q)) (assert Q >>= f) g"
   by (auto simp add: assert_def monadic_rewrite_def fail_def split: if_split)
 
-lemma monadic_rewrite_state_assert_true:
-  "monadic_rewrite F E P (state_assert P) (return ())"
-  by (simp add: state_assert_def monadic_rewrite_def exec_get)
+(* under assumption of non-failure of LHS, we can assume any assertions for showing preconditions
+   of rewrite *)
+lemma monadic_rewrite_state_assert:
+  "monadic_rewrite F E P (f ()) g
+   \<Longrightarrow> monadic_rewrite F E (\<lambda>s. (F \<longrightarrow> Q s \<longrightarrow> P s) \<and> (\<not>F \<longrightarrow> P s \<and> Q s))
+                      (state_assert Q >>= f) g"
+  by (fastforce simp: monadic_rewrite_def exec_get state_assert_def bind_assoc assert_def)
 
 lemma monadic_rewrite_stateAssert:
-  "monadic_rewrite F E P (stateAssert P xs) (return ())"
-  by (simp add: stateAssert_def monadic_rewrite_def exec_get)
+  "monadic_rewrite F E P (f ()) g
+   \<Longrightarrow> monadic_rewrite F E (\<lambda>s. (F \<longrightarrow> Q s \<longrightarrow> P s) \<and> (\<not>F \<longrightarrow> P s \<and> Q s))
+                      (stateAssert Q xs >>= f) g"
+  by (fastforce simp: monadic_rewrite_def exec_get stateAssert_def bind_assoc assert_def)
 
 text \<open>Non-determinism: alternative and select\<close>
 
--- a/proof/crefine/AARCH64/Fastpath_C.thy
+++ b/proof/crefine/AARCH64/Fastpath_C.thy
@@ -1,5 +1,5 @@
 (*
- * Copyright 2023, Proofcraft Pty Ltd
+ * Copyright 2024, Proofcraft Pty Ltd
  * Copyright 2014, General Dynamics C4 Systems
  *
  * SPDX-License-Identifier: GPL-2.0-only
@@ -19,9 +19,6 @@ begin
 
 context begin interpretation Arch . (*FIXME: arch_split*)
 
-(* FIXME AARCH64 for now, comment out entire file except for stubs of main theorems;
-   this is copied from ARM_HYP and will need a fair bit of work to update
-
 lemma setCTE_obj_at'_queued:
   "\<lbrace>obj_at' (\<lambda>tcb. P (tcbQueued tcb)) t\<rbrace> setCTE p v \<lbrace>\<lambda>rv. obj_at' (\<lambda>tcb. P (tcbQueued tcb)) t\<rbrace>"
   unfolding setCTE_def
@@ -36,7 +33,7 @@ crunch obj_at'_not_queued: emptySlot "obj_at' (\<lambda>a. \<not> tcbQueued a) p
 lemma getEndpoint_obj_at':
   "\<lbrace>obj_at' P ptr\<rbrace> getEndpoint ptr \<lbrace>\<lambda>rv s. P rv\<rbrace>"
   apply (wp getEndpoint_wp)
-  apply (clarsimp simp: obj_at'_def projectKOs)
+  apply (clarsimp simp: obj_at'_def )
   done
 
 lemmas setEndpoint_obj_at_tcb' = setEndpoint_obj_at'_tcb
@@ -84,16 +81,15 @@ declare comp_apply [simp]
 crunch ksArch[wp]: asUser "\<lambda>s. P (ksArchState s)"
   (wp: crunch_wps)
 
-definition
-   tcbs_of :: "kernel_state => word32 => tcb option"
-where
-  "tcbs_of s = (%x. if tcb_at' x s then projectKO_opt (the (ksPSpace s x)) else None)"
+(* FIXME AARCH64 consider moving this, on MCS there is a tcbs_of as well *)
+definition tcbs_of :: "kernel_state \<Rightarrow> machine_word \<Rightarrow> tcb option" where
+  "tcbs_of s = (\<lambda>x. if tcb_at' x s then projectKO_opt (the (ksPSpace s x)) else None)"
 
 lemma obj_at_tcbs_of:
   "obj_at' P t s = (EX tcb. tcbs_of s t = Some tcb & P tcb)"
   apply (simp add: tcbs_of_def split: if_split)
   apply (intro conjI impI)
-   apply (clarsimp simp: obj_at'_def projectKOs)
+   apply (clarsimp simp: obj_at'_def)
   apply (clarsimp simp: obj_at'_weakenE[OF _ TrueI])
   done
 
@@ -108,13 +104,13 @@ lemma tcbs_of_ko_at':
 lemma tcbs_of_valid_tcb':
   "\<lbrakk> valid_objs' s; tcbs_of s p = Some tcb \<rbrakk> \<Longrightarrow> valid_tcb' tcb s"
   by (frule tcbs_of_ko_at')
-     (drule (1) ko_at_valid_objs', auto simp: projectKOs valid_obj'_def)
+     (drule (1) ko_at_valid_objs', auto simp:  valid_obj'_def)
 
-*)
+
 end
 
 context kernel_m begin
-(*
+
 lemma getCTE_h_val_ccorres_split:
   assumes var: "\<And>s f s'. var (var_update f s) = f (var s)
                   \<and> ((s', var_update f s) \<in> rf_sr) = ((s', s) \<in> rf_sr)"
@@ -218,7 +214,7 @@ lemma lookup_fp_ccorres':
   "ccorres (\<lambda>mcp ccp. ccap_relation (case mcp of Inl v => NullCap | Inr v => v) ccp)
                 ret__struct_cap_C_'
            (valid_cap' cap and valid_objs')
-           (UNIV \<inter> {s. ccap_relation cap (cap_' s)} \<inter> {s. cptr_' s = cptr}) []
+           ({s. ccap_relation cap (cap_' s)} \<inter> {s. cptr_' s = cptr}) []
        (cutMon ((=) s) (doE t \<leftarrow> resolveAddressBits cap cptr bits;
                              liftE (getSlotCap (fst t))
                          odE))
@@ -228,21 +224,20 @@ lemma lookup_fp_ccorres':
    apply (rule ccorres_symb_exec_r)
      apply (rule_tac xf'=ret__int_' in ccorres_abstract, ceqv)
      apply (rule_tac P="rv' = from_bool (isCNodeCap cap)" in ccorres_gen_asm2)
-     apply (simp add: from_bool_0 del: Collect_const cong: call_ignore_cong)
+     apply (simp add: from_bool_0 del: Collect_const)
      apply (rule ccorres_Cond_rhs_Seq)
-      apply (simp add: resolveAddressBits.simps split_def del: Collect_const
-                      split del: if_split)
+      apply (simp add: resolveAddressBits.simps split_def del: Collect_const)
       apply (rule ccorres_drop_cutMon)
       apply (rule ccorres_from_vcg_split_throws[where P=\<top> and P'=UNIV])
        apply vcg
       apply (rule conseqPre, vcg)
       apply (clarsimp simp: throwError_def return_def isRight_def isLeft_def
                             ccap_relation_NullCap_iff)
-     apply (clarsimp simp del: Collect_const cong: call_ignore_cong)
+     apply (clarsimp simp del: Collect_const)
      apply (rule_tac P="valid_cap' cap and valid_objs'"
-                and P'="UNIV \<inter> {s. ccap_relation cap (cap_' s) \<and> isCNodeCap cap}
-                             \<inter> {s. bits_' s = 32 - of_nat bits \<and> bits \<le> 32 \<and> bits \<noteq> 0}"
-                   in ccorres_inst)
+                 and P'="{s. ccap_relation cap (cap_' s) \<and> isCNodeCap cap}
+                         \<inter> {s. bits_' s = 64 - of_nat bits \<and> bits \<le> 64 \<and> bits \<noteq> 0}" (* FIXME AARCH64 64 *)
+                  in ccorres_inst)
      apply (thin_tac "isCNodeCap cap")
      defer
      apply vcg
@@ -254,15 +249,6 @@ lemma lookup_fp_ccorres':
             rule: resolveAddressBits.induct)
   case (1 acap acptr abits as)
 
-    have sub_mask_neq_0_eq:
-      "\<And>v :: word32. v && 0x1F \<noteq> 0 \<Longrightarrow> 0x20 - (0x20 - (v && 0x1F) && mask 5) = v && 0x1F"
-      apply (subst word_le_mask_eq)
-        apply (simp only: mask_def)
-        apply (rule word_le_minus_mono, simp_all add: word_le_sub1 word_sub_le_iff)[1]
-        apply (rule order_trans, rule word_and_le1, simp)
-       apply (simp add: word_bits_def)
-      done
-
     have valid_cnode_bits_0:
       "\<And>s acap. \<lbrakk> isCNodeCap acap; s \<turnstile>' acap \<rbrakk> \<Longrightarrow> capCNodeBits acap \<noteq> 0"
       by (clarsimp simp: isCap_simps valid_cap'_def)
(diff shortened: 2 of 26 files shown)


Xaphiosis:
rebased onto master, had a de-sync between seL4 versions

lsf37:
Nice work. Personally I'd slightly prefer 3 commits, or if you want to minimise, maybe 2 (one for defs + equiv, and one for the C part), but I don't have strong feelings about it.

lib/MonadicRewrite.thy
```diff
@@ -524,13 +524,19 @@ lemma monadic_rewrite_assert2:
...
 
-lemma monadic_rewrite_state_assert_true:
-  "monadic_rewrite F E P (state_assert P) (return ())"
-  by (simp add: state_assert_def monadic_rewrite_def exec_get)
+(* under assumption of non-failure of LHS, we can assume any assertions for showing preconditions
+   of rewrite *)
+lemma monadic_rewrite_state_assert:
+  "monadic_rewrite F E P (f ()) g
+   \<Longrightarrow> monadic_rewrite F E (\<lambda>s. (F \<longrightarrow> Q s \<longrightarrow> P s) \<and> (\<not>F \<longrightarrow> P s \<and> Q s))
+                      (state_assert Q >>= f) g"
```
lsf37, line 532: Would it add any convenience later in the proofs if we provided two versions where `F` is instantiated (and the guard therefore simplifies a bit), or does this disappear anyway before the user really sees it?

proof/crefine/AARCH64/Fastpath_Defs.thy
```diff
@@ -41,37 +41,22 @@ definition
...
-    }
-    /* vmids are the tags used instead of hw_asids in hyp mode */
-    stored_hw_asid.words[0] = asid_map_asid_map_vspace_get_stored_hw_vmid(asid_map);
-\<close>
-
-\<^cancel>\<open> this check appears to only happen on AARCH32
-    pd \<leftarrow> returnOk $ \<^cancel>\<open>capPTBasePtr\<close> XXX $ capCap $ cteCap newVTable;
-\<close>
+    vspace_root \<leftarrow> returnOk $ capPTBasePtr $ capCap $ cteCap newVTable; \<comment> \<open>cap_pd in C\<close>
+    asid \<leftarrow> returnOk $ (fst $ the $ capPTMappedAddress $ capCap $ cteCap newVTable);
```
lsf37, line 45:
```suggestion
    asid \<leftarrow> returnOk $ fst $ the $ capPTMappedAddress $ capCap $ cteCap newVTable;
```
(or leaving out the first `$` instead should also work)

proof/crefine/AARCH64/Fastpath_Defs.thy
```diff
@@ -134,44 +119,31 @@ definition
...
+    unlessE (isReplyCap callerCap) $ throwError ();
 
     caller \<leftarrow> returnOk $ capTCBPtr callerCap;
     callerFault \<leftarrow> liftE $ threadGet tcbFault caller;
     unlessE (callerFault = None) $ throwError ();
     newVTable \<leftarrow> liftE $ getThreadVSpaceRoot caller >>= getCTE;
     unlessE (isValidVTableRoot $ cteCap newVTable) $ throwError ();
 
+    vspace_root \<leftarrow> returnOk $ capPTBasePtr $ capCap $ cteCap newVTable; \<comment> \<open>cap_pd in C\<close>
+    asid \<leftarrow> returnOk $ (fst $ the $ capPTMappedAddress $ capCap $ cteCap newVTable);
```
lsf37, line 134:
```suggestion
    asid \<leftarrow> returnOk $ fst $ the $ capPTMappedAddress $ capCap $ cteCap newVTable;
```

proof/crefine/AARCH64/Fastpath_Equiv.thy
```diff
@@ -1676,9 +1758,13 @@ lemma fastpath_callKernel_SysReplyRecv_corres:
                                     in monadic_rewrite_gen_asm)
                     apply (rule monadic_rewrite_trans[OF _ monadic_rewrite_transverse])
                       apply (rule monadic_rewrite_weaken_flags[where F=False and E=True], simp)
+
                       apply (rule isolate_thread_actions_rewrite_bind
```
lsf37, line 1762: Intentional blank line?

proof/crefine/AARCH64/Fastpath_C.thy
```diff
@@ -84,16 +81,15 @@ declare comp_apply [simp]
 crunch ksArch[wp]: asUser "\<lambda>s. P (ksArchState s)"
   (wp: crunch_wps)
 
-definition
-   tcbs_of :: "kernel_state => word32 => tcb option"
-where
-  "tcbs_of s = (%x. if tcb_at' x s then projectKO_opt (the (ksPSpace s x)) else None)"
+(* FIXME AARCH64 consider moving this, on MCS there is a tcbs_of as well *)
+definition tcbs_of :: "kernel_state \<Rightarrow> machine_word \<Rightarrow> tcb option" where
+  "tcbs_of s = (\<lambda>x. if tcb_at' x s then projectKO_opt (the (ksPSpace s x)) else None)"
```
lsf37, line 86: I had completely forgotten that this was in here. Should indeed consolidate these at some point, but fine to leave here for now, esp since the FIXME makes it findable.

proof/crefine/AARCH64/Fastpath_C.thy
```diff
@@ -228,20 +224,19 @@ lemma lookup_fp_ccorres':
...
       apply (rule conseqPre, vcg)
       apply (clarsimp simp: throwError_def return_def isRight_def isLeft_def
                             ccap_relation_NullCap_iff)
-     apply (clarsimp simp del: Collect_const cong: call_ignore_cong)
+     apply (clarsimp simp del: Collect_const)
      apply (rule_tac P="valid_cap' cap and valid_objs'"
-                and P'="UNIV \<inter> {s. ccap_relation cap (cap_' s) \<and> isCNodeCap cap}
-                             \<inter> {s. bits_' s = 32 - of_nat bits \<and> bits \<le> 32 \<and> bits \<noteq> 0}"
+                and P'="{s. ccap_relation cap (cap_' s) \<and> isCNodeCap cap}
+                        \<inter> {s. bits_' s = 64 - of_nat bits \<and> bits \<le> 64 \<and> bits \<noteq> 0}" (* FIXME AARCH64 64 *)
```
lsf37, line 239: Is there any chance of making the fast path proofs generic over 32/64 bit or should we in this case just not worry about those specific kind of magic numbers? (specifically 32 and 64)

lsf37: Looks like writing things out makes them clearer: yes, I think it is worth getting rid of the magic numbers, because it might at least make some parts of the fast path proofs textually equal or more similar, which makes it easier to apply simple-minded patches like renaming etc.

proof/crefine/AARCH64/Fastpath_C.thy
```diff
@@ -272,221 +258,223 @@ lemma lookup_fp_ccorres':
...
           apply csymbr+
-          apply (rule ccorres_cond_false_seq)
-          apply (simp add: ccorres_expand_while_iff_Seq[symmetric]
-                           whileAnno_def cong: call_ignore_cong)
-          apply (rule ccorres_cond_false)
-          apply (rule ccorres_cond_true_seq)
-          apply (rule ccorres_from_vcg_split_throws[where P=\<top> and P'=UNIV])
+          apply (simp del: Collect_const)
+          apply (rule ccorres_move_array_assertion_cnode_ctes ccorres_move_c_guard_cte
+            | csymbr)+
```
lsf37, line 328:
```suggestion
          apply (rule ccorres_move_array_assertion_cnode_ctes ccorres_move_c_guard_cte
                 | csymbr)+
```

proof/crefine/AARCH64/Fastpath_C.thy
```diff
@@ -272,221 +258,223 @@ lemma lookup_fp_ccorres':
...
+          apply (rename_tac "getCTE_cap")
+          apply (csymbr | rule ccorres_Guard_Seq)+
+          apply (rule ccorres_cond_false_seq)
+          apply (simp add: ccorres_expand_while_iff_Seq[symmetric]
+                           whileAnno_def del: Collect_const)
+          apply (rule ccorres_cond_false)
+          apply (rule ccorres_cond_false_seq)
+          apply (simp del: Collect_const)
+          apply (rule_tac P'="{s. cap_' s = getCTE_cap}"
+                         in ccorres_from_vcg_throws[where P=\<top>])
```
lsf37, line 406: What's our style policy for the `in`? Should this be the following?
```suggestion
          apply (rule_tac P'="{s. cap_' s = getCTE_cap}"
                          in ccorres_from_vcg_throws[where P=\<top>])
```

proof/crefine/AARCH64/Fastpath_C.thy
```diff
@@ -1052,25 +1047,15 @@ lemma ccorres_pre_getCTE2:
...
-     lookupExtraCaps thread buffer info = returnOk []"
-  by (clarsimp simp: lookupExtraCaps_def
-                     getExtraCPtrs_def liftE_bindE
-                     upto_enum_step_def mapM_Nil
-              split: Types_H.message_info.split option.split)
-
+(* this condition is copied directly from fastpath_mi_check_spec, which comes out of guards there *)
 lemma fastpath_mi_check:
-  "((mi && mask 9) + 3) && ~~ mask 3 = 0
+  "(\<not> 4 < mi && 0x1FF)
```
lsf37, line 1058: maybe add a FIXME on this one for removing the magic numbers

proof/crefine/AARCH64/Fastpath_C.thy
```diff
@@ -1736,776 +1684,989 @@ lemma recv_ep_queued_st_tcb_at':
...
+                  split: option.splits asid_map_CL.splits)
+  apply (clarsimp simp: asid_map_lift_def Let_def ucast_ucast_mask
+                  split: if_splits)
+  done
+
+(* FIXME AARCH64 any better way? *)
+(* when we know P is true, but we can't use subst or clarsimp due to schematics *)
+lemma ccorres_If_True_drop:
+  "\<lbrakk> P; ccorres_underlying sr Gamm r xf arrel axf R R' hs a c \<rbrakk>
+   \<Longrightarrow> ccorres_underlying sr Gamm r xf arrel axf R R' hs (If P a b) c"
```
lsf37, line 1823: Looks perfectly legit to me, i.e. I don't think there's a significantly better way to do this.

proof/crefine/AARCH64/Fastpath_C.thy
```diff
@@ -1736,776 +1684,989 @@ lemma recv_ep_queued_st_tcb_at':
...
+                        apply (vcg exspec=cap_endpoint_cap_get_capCanGrant_modifies
+                                   exspec=cap_endpoint_cap_get_capCanGrantReply_modifies)
+                       apply simp
+                       (* throw away results of isHighestPrio and the fastfail shortcut *)
+                       apply (wp (once) hoare_drop_imp, wp)
+                       apply (wp (once) hoare_drop_imp, wp)
+
+(* FIXME AARCH64 this is frustrating:
+   - can't use clarsimp because it'll introduce new free variables (\<exists>_. _ = Some _) that will
+     cause schematic unification problems
```
lsf37, line 2501: We could try to remove the simp rule that turns None into Ex Some. With that, clarsimp might be fine.

lsf37 (approved):
Very cool. Modulo the small stuff above and commit squashing, I think this is ready to merge. It's still gigantic, but it looks to me like proof quality has improved a bit compared to the other fast path versions.

We should add a commit that removes the `CREFINE_QUICK_AND_DIRTY` from the Makefile.

lib/MonadicRewrite.thy
Xaphiosis: I thought about it, but didn't see any point. If you apply this rule, you end up with a wp goal later on, which will be cleared with a `simp` given any specific value of `F`. If an extreme situation, they can do `where F=..., simplified`, but again, not seeing it.

proof/crefine/AARCH64/Fastpath_Equiv.thy
Xaphiosis: oops

proof/crefine/AARCH64/Fastpath_C.thy
Xaphiosis: I think making them generic in 32/64 is not feasible, but obvious mentions of machine word lengths are something we should strive for, hence the FIXME. The issue is, if you want this merged ASAP, do I go doing these fixes now or after?

proof/crefine/AARCH64/Fastpath_C.thy
Xaphiosis: I asked about this in mattermost before I did these, and people linked me to the discussion where we had 3 acceptable versions I think (rhs `and`: the `in` is like you point out; lhs `and` with `in` indented same as the `and`; the `and/in` right-aligned with the `c` in `rule_tac` (Michael style?)). I'm a bit confused at this point. I was confused before as well, because I started the original discussion.
Worse yet, trying to align the `in` so that the `n` is under the `c`, when seen in a normal text editor (or diff), results in the `n` being under the space instead.
Advice welcome.

proof/crefine/AARCH64/Fastpath_C.thy
Xaphiosis: I put a comment right above which says this is copied by whatever shows up in `fastpath_mi_check_spec`, which I think are bitfield gen numbers? I have no clue what they are.

proof/crefine/AARCH64/Fastpath_C.thy
Xaphiosis: There's still the issue of pairs, which would need to be folded back into `fst` and `snd`. Do we have tech to handle that?
There's also what Corey wrote on the wp schematic issue, where we might have a wrong rule in the `wp_comb` set.

proof/crefine/AARCH64/Fastpath_Defs.thy
Xaphiosis: done

proof/crefine/AARCH64/Fastpath_Defs.thy
Xaphiosis: done

proof/crefine/AARCH64/Fastpath_C.thy
Xaphiosis: done

proof/crefine/AARCH64/Fastpath_C.thy
Xaphiosis: removed the FIXME

Xaphiosis:
> Very cool. Modulo the small stuff above and commit squashing, I think this is ready to merge. It's still gigantic, but it looks to me like proof quality has improved a bit compared to the other fast path versions.

I'd like your reply to my replies (maybe I catch you tonight) before resolving the rest. I also have some questions:
* do you want to merge this before `AARCH64 FIXME`s are resolved, and then we sort them out later / after the deadline? You seem to have a plan w.r.t. the timing, but I think I didn't pick up on it
* bike shed: `asid_has_vmid` name: if at some point we do RISCV64 fastpath, this name makes no sense there, and it's more than the VMID anyway. Should I do the rename as in the FIXME? A better name? `asid_has_vspace`? Leave it for now?
 
> We should add a commit that removes the `CREFINE_QUICK_AND_DIRTY` from the Makefile.

Good point, added that to the commits.

I'm not sure whether anyone else is going to look at this, but I will wait with squashing until tonight when we can hopefully resolve/discuss the remaining issues.

proof/crefine/AARCH64/Fastpath_C.thy
corlewis: To clarify, if I'm right in what I thought was going on with the `wp_comb` rule then it would just help with the problem you ran into of a schematic assumption being produced and then unified with `False`. The actual proof would still require either your careful rule application or the careful adjustment of the simp set like you two are talking about.

proof/crefine/AARCH64/Fastpath_C.thy
```diff
@@ -272,221 +258,223 @@ lemma lookup_fp_ccorres':
...
+           apply csymbr
+           apply (simp add: unat_sub word_le_nat_alt if_1_0_0 shiftl_shiftr3 word_size
+                       del: Collect_const)
+           apply (rule ccorres_Cond_rhs)
+            apply (rule ccorres_from_vcg_throws[where P=\<top> and P'=UNIV])
+            apply (rule allI, rule conseqPre, vcg)
+            apply clarsimp
+            apply (clarsimp simp: whenE_def throwError_def return_def
+                                  ccap_relation_NullCap_iff isRight_def isLeft_def split_def split: if_split)
+
```
michaelmcinerney, line 369: Intentional newline?

proof/crefine/AARCH64/Fastpath_C.thy
corlewis: Hmm, I guess our original discussion only considered the case with multiple instantiations. Sorry to add yet another version, but I thought our preference from that discussion was what I think you're describing with the right-aligned Michael style. That would make it
```suggestion
          apply (rule_tac P'="{s. cap_' s = getCTE_cap}"
                       in ccorres_from_vcg_throws[where P=⊤])
```

proof/crefine/AARCH64/Fastpath_C.thy
```diff
@@ -552,271 +541,236 @@ lemma ccorres_catch_bindE_symb_exec_l:
...
-               del: Collect_const cong: call_ignore_cong)
-   apply simp
+  "ccorres dc xfdc
+     (pspace_aligned' and pspace_distinct' and valid_objs' and no_0_obj' and valid_arch_state'
+      and tcb_at' thread
+      and cte_wp_at' (\<lambda>cte. isValidVTableRoot (cteCap cte)
+                            \<and> capPTBasePtr (capCap (cteCap cte)) = vsroot
+                            \<and> (\<exists>vaddr. capPTMappedAddress (capCap (cteCap cte))
+                                       = Some (asid, vaddr)))
+                     (thread + tcbVTableSlot * 0x20)
```
michaelmcinerney, line 719: Is this the `cteSize` or something that you were asking about on Mattermost? Is there any benefit of using that instead of the bare number here?

proof/crefine/AARCH64/Fastpath_C.thy
```diff
@@ -1736,776 +1684,988 @@ lemma recv_ep_queued_st_tcb_at':
...
            apply (rule allI, rule conseqPre, vcg)
            apply (clarsimp simp: cur_tcb'_def rf_sr_ksCurThread)
            apply (drule(1) obj_at_cslift_tcb, clarsimp)
            apply (clarsimp simp: typ_heap_simps' ctcb_relation_def cfault_rel_def)
            apply (rule rev_bexI, erule threadGet_eq)
            apply (clarsimp simp: seL4_Fault_lift_def Let_def split: if_split_asm)
           apply ceqv
+         apply (rename_tac fault fault')
+         apply (rule ccorres_alternative1) (* pick pickFastpath = True, still not clear what it's for *)
+
```
michaelmcinerney, line 1905: Another newline here, maybe unintentional

michaelmcinerney (approved):
Nice. Just a few nitpicks from me

proof/crefine/AARCH64/Fastpath_C.thy
lsf37: As long as the tag is there it's fine to do later.

Sorry for the confusing comment: initially I was thinking we could just remove the tag, because there is no point anyway, but then I discovered what the point is. So everything is fine and no action required for the PR.

proof/crefine/AARCH64/Fastpath_C.thy
lsf37: Ok, I'm glad I'm not the only one who is confused :-)

I propose we leave this one as is and figure out separately what we really want for this case.

proof/crefine/AARCH64/Fastpath_C.thy
lsf37: Pairs being split should be handled correctly by `wpfix`, so I think removing the None/Some rule has a chance in this case (it won't fix the general problem). It has a tag already, so we're fine to merge, I was just curious what people's thoughts were on this one. If I get to it in the cleanup run I might try out removing the rule and see what happens.

lsf37:
> > Very cool. Modulo the small stuff above and commit squashing, I think this is ready to merge. It's still gigantic, but it looks to me like proof quality has improved a bit compared to the other fast path versions.
> 
> I'd like your reply to my replies (maybe I catch you tonight) before resolving the rest. I also have some questions:
> 
> * do you want to merge this before `AARCH64 FIXME`s are resolved, and then we sort them out later / after the deadline? You seem to have a plan w.r.t. the timing, but I think I didn't pick up on it

I think we should merge this one first, so the proof is properly finished and `quick_and_dirty` is switched off. If we're lucky, we can rebase the existing FIXME cleanup PRs over it and they still work -- then I'd merge these too. If they need fixing, we can just see how far we get, they are less critical.

> * bike shed: `asid_has_vmid` name: if at some point we do RISCV64 fastpath, this name makes no sense there, and it's more than the VMID anyway. Should I do the rename as in the FIXME? A better name? `asid_has_vspace`? Leave it for now?

It's ok to leave it for now. The proposed `asid_has_vspace` is not bad, so I'd take that one when we get to it. The predicate will likely need changing for RISCV, since it does ASIDs differently, but the name `asid_has_vspace` would still fit. I guess it's "vspace with potentially additional data", but that is a bit long.

> > We should add a commit that removes the `CREFINE_QUICK_AND_DIRTY` from the Makefile.
> 
> Good point, added that to the commits.

👍
 
> I'm not sure whether anyone else is going to look at this, but I will wait with squashing until tonight when we can hopefully resolve/discuss the remaining issues.

We're good to merge from my side after squashing.

proof/crefine/AARCH64/Fastpath_C.thy
Xaphiosis: This looks like I'm carrying some kind of rendering issue with even placing the `in` in the right place, because the *intended* placement was where @corlewis said it should be. I'll fix those misalignments with vim for now so it's like what Corey said, and then I'll create an issue so we sort this out and can put it in the style guide.

proof/crefine/AARCH64/Fastpath_C.thy
Xaphiosis: Ok, that would be nice, but I think the wp_comb thing also should happen if we're doing cleanup at some point, because the combo of wp, simp, vcg is a disaster to work with otherwise. When you say "when I get to it in the cleanup run", does that mean you've assumed responsibility for the doing entire cleanup run?

proof/crefine/AARCH64/Fastpath_C.thy
Xaphiosis: yeah, better without it

proof/crefine/AARCH64/Fastpath_C.thy
Xaphiosis: Yeah, this is the part that refers to maybe fixing it:
```
(* FIXME AARCH64 in this file, we see ptr + 0x20 * tcbVTableSlot a lot, where the 0x20 is
   2 ^ cte_level_bits
   This used to be 0x10 on 32-bit, so unlikely to change soon, but still might be worth a cleanup
   or abbreviation. This is one way it could work (and can go outside Arch locale): *)
declare cte_level_bits_def[code]
value_abbreviation (input) cte_size "(2::machine_word) ^ cte_level_bits"
```

It's not really clear if an input abbreviation only will be helpful here. Might be better to go with a constant, but it was hard to wrangle it in the proofs. Still not sure about this one, but yeah it's what I was asking about in mattermost

proof/crefine/AARCH64/Fastpath_C.thy
Xaphiosis: OK it seems in the final ccorres proofs I wasn't making the misalignment mistakes, so maybe they didn't come from me. It's still not entirely consistent w.r.t. the `in` placement so TODO: make issue.
Fix pushed.

Xaphiosis:
Squashed and pushed. Might be worth waiting for the test results on this one as it adds the quick_and_dirty removal from the Makefile.

proof/crefine/AARCH64/Fastpath_C.thy
lsf37: I was planning to volunteer to do a first pass, but I wouldn't say I have assumed responsibility for all of the cleanup :-) There will probably be a whole bunch of things that need some insight and where we might want to discuss more, it'd be nice to do these together in some way.

lsf37:
Happy with the squash and commit messages. Will wait for the tests to pass before merging.

PR #741, opened 2024-03-25
docs: add debugging tips for vcg goals

lsf37:
Inspired by discussions on the fastpath PR and MM discussions, a minor addition to the CRefine notes in `docs/` on how to debug that `False`-in-final-goal-after-`vcg` problem.


lsf37:
This is not trying to be comprehensive, but just an improvement to the "sorry, no clue" that was there before.

docs/crefine-notes.md
```diff
@@ -243,14 +243,49 @@ However here are some tricks when it could go wrong:
   just before
 
 
-* sometimes the precondition generated by `wp` or `vcg` contains False somewhere (we figure it out while trying to prove the 'last big subgoal')
-so a simplification before the `wp`/`vcg` is needed...
+* sometimes the precondition generated by `wp` or `vcg` contains `False`
+  somewhere (we figure it out while trying to prove the 'last big subgoal') so a
+  modification before `vcg` is needed. See the next section for debugging tips
```
corlewis, line 248: Very minor, but why is it just `vcg` in this line now and not `wp`/`vcg`?

corlewis (approved):
This looks good to me and is definitely an improvement over 'no clue'!

docs/crefine-notes.md
lsf37: `wp` no longer produces just `False`, but would fail to apply if the precondition doesn't depend on the right parameters. You can still get funky goals if your assumptions have a schematic in them, but then the proof is already broken in other ways.

michaelmcinerney (approved):
Very helpful, thanks

docs/crefine-notes.md
```diff
@@ -243,14 +243,49 @@ However here are some tricks when it could go wrong:
...
+* go back to the place where the schematic `?P` is introduced and introduce the
+  parameter before the schematic, so `?P` can depend on the parameter. This could
+  happen for instance by using a better case distinction rule that manages
+  schematics, or by performing the relevant `exE`, `clarsimp` etc before the
+  schematic is introduced (not always possible).
+
+* go back to the place where the parameter `y` is introduced and prevent it from
+  being introduced. E.g. instead of using `clarsimp` to generate `x = Some y`
+  and using `y`, use `the x` instead. This is not necessarily nice, but should
+  always be available in principle.
```
Xaphiosis, line 288: another way of dealing with these situations, though rare, is using `wpc` and then declaring one of the cases is impossible

docs/crefine-notes.md
```diff
@@ -243,14 +243,49 @@ However here are some tricks when it could go wrong:
...
+Otherwise, there are two remaining strategies:
+
+* go back to the place where the schematic `?P` is introduced and introduce the
+  parameter before the schematic, so `?P` can depend on the parameter. This could
+  happen for instance by using a better case distinction rule that manages
+  schematics, or by performing the relevant `exE`, `clarsimp` etc before the
+  schematic is introduced (not always possible).
+
+* go back to the place where the parameter `y` is introduced and prevent it from
+  being introduced. E.g. instead of using `clarsimp` to generate `x = Some y`
```
Xaphiosis, line 286: I think it's better to say the complete thing: something about maintaining the `EX x. x = Some y` form in the assumptions and using `the x` for the rest of the corres/wp part of the proof. Only saying about `the x` I don't think is entirely clear

Xaphiosis: also another option is blocking a `x \noteq None` to `EX ...` rewrite, but that won't stop pairs

docs/crefine-notes.md
Xaphiosis: I've not really seen this case happen in practice. What I do see is that the schematic precondition doesn't work right, the simp outside/inside vcg resolves it to False, and we get a postcondition that has subterms like: `ALL t. some_var' t = foo --> False --> P (some_var t) /\ ...`  which then when hit by the next simp become `ALL t. some_var' t = foo`. Since `vcg` uses simp internally these are very difficult to spot. Please see my overall comment on this PR below.

Xaphiosis:
When looking at this PR, I kept having a sneaking suspicion I had written something about something related already and asked people for feedback, but couldn't find the direct text I remembered, and this PR came as a surprise to me. Turns out I did write something related, and linked it to mattermost (Michael even looked at it):
https://gist.github.com/Xaphiosis/2fa2b51b65e05c3addda1576f90a31ed
It feels like some of that would be useful here, particularly when the "proving False" bit doesn't show up as often, especially in larger goals, as the "why do I need to prove this thing I can't actually prove" case. To me that section is somewhat confusing, since as I said I rarely get to see a proper `False`.

lsf37:
> When looking at this PR, I kept having a sneaking suspicion I had written about something related already and asked people for feedback, although I couldn't find the direct text I remembered. Turns out I did write something related and linked it on mattermost pinging the whole channel (Michael even looked at it): https://gist.github.com/Xaphiosis/2fa2b51b65e05c3addda1576f90a31ed
> 
> This PR came out of nowhere, so what I wrote was probably not considered.

Indeed, I had forgotten about this, I was just triggered by the "sorry, not better clue" that came up in a `grep` for a completely different task. Your overall description of the problem is much better, so we should definitely use most of that. Maybe bits of mine are still useful to add -- if not, also Ok to just throw it out, it's not like a huge amount of consideration went into mine, I just tried to put together something short that was better than "no clue".

How about making your gist a separate file under `docs/`, adding some bits of this PR as appropriate if there is something useful, and then replace the description in here with a link to the separate file?

lsf37:
> Something to consider w.r.t. a "proving False" vs a "unprovable subterms" section, is that I rarely get to see a proper `False`. Especially in larger goals, the "why do I need to prove this thing I can't actually prove" case situation is the one I see nearly all the time. Can anyone corroborate the frequency of seeing a real `False`?

I've seen it a few times, mostly in other people's proofs (not the entire post condition being `False`, but term under a `case` or some quantifier), but the one you describe is much more common.

Xaphiosis:
Suggestion of making a separate file, putting the combined crefine/ccorres proof debugging info in there, then linking to it from `crefine-notes.md` makes sense. Let's do that. There should be a way of combining my gist with what @lsf37 wrote up, possibly favouring whichever is the most specific/comprehensible to the content they apply to.

Xaphiosis (changes requested):
Suggestion of making a separate file, putting the combined crefine/ccorres proof debugging info in there, then linking to it from `crefine-notes.md` makes sense. Let's do that. There should be a way of combining my gist with what @lsf37 wrote up, possibly favouring whichever is the most specific/comprehensible to the content they apply to.

lsf37:
Now updated with Raf's gist as a basis and extracted from the CRefine notes as a separate file. I also added all our recent new files to the README, which is attempting to give a table of contents.

docs/vcg-debugging.md
```diff
@@ -0,0 +1,124 @@
...
+  happen for instance by using a better case distinction rule that manages
+  schematics, or by performing the relevant `exE`, `clarsimp` etc before the
+  schematic is introduced (not always possible).
+
+* go back to the place where the parameter `y` is introduced and prevent it from
+  being introduced. E.g. instead of using `clarsimp` to generate `x = Some y`
+  and using `y`, be careful to keep the form `∃y. x = Some y` in your
+  assumptions (or prevent it from even occurring when possible, e.g. by blocking
+  a `x ≠ None` rewrite). Then use `the x` instead to access the content of `x`.
+  This is not necessarily nice, but should always be available in principle.
```
Xaphiosis, line 80: should we mention `not_None` in `Lib.thy`?

docs/vcg-debugging.md
```diff
@@ -0,0 +1,124 @@
...
+
+* it is sometimes possible to use `wpc` in `ccorres` goals to produce sub goals
+  that provide access to the correct set of parameters (e.g. a `None` case and
+  `Some y` case, where `y` is fine to access). This can also be used when one of
+  the cases is impossible by showing a contradiction for this branch.
```
Xaphiosis, line 85: that branch

docs/vcg-debugging.md
```diff
@@ -0,0 +1,124 @@
...
+
+The way to track down these issues is to go over the `vcg` steps to see which
+ones introduced the unexpected result. However, with large goals this is very
+difficult to work with. What else can help?
+
+First, try removing `(P ⟶ False) = (¬ P)` from the simpset, so we can look for
+`False` in the goals:
+
+```isabelle
+supply simp_thms(19)[simp del]
```
Xaphiosis, line 42: you sometimes indent and sometimes not, not sure if markdown cares

docs/vcg-debugging.md
```diff
@@ -0,0 +1,124 @@
...
+* `vcg` might be trying to inline multiple function definitions and the goal is
+  then not actually as small as it looked. When this happens the `vcg` prints
+  info messages about this unfolding. Prove specification lemmas for the
+  functions instead to avoid unfolding.
+
+If you do get a legitimate goal that is very large, you can try to at least
+speed up printing at the cost of not expanding syntax and syntax translations:
+
+* Use `supply [[goals_limit = 1]]` if you have more than one goal
+* Use the `Quickprint` option in jEdit to turn off translations
```
Xaphiosis, line 119: `Quick print` in menu

Xaphiosis (approved):
Thanks for resolving this, I think the outcome looks good now, only had trivial nitpicks. Time will tell whether anyone actually goes and looks at the debugging guide, but you've certainly left enough pointers to it.

Commit-wise, given that the "add vcg debugging guide" commit moves a whole lot of the previous "add debugging tips for vcg goals", would it make sense to merge the two?

lsf37:
> Thanks for resolving this, I think the outcome looks good now, only had trivial nitpicks. Time will tell whether anyone actually goes and looks at the debugging guide, but you've certainly left enough pointers to it.

👍 
 
> Commit-wise, given that the "add vcg debugging guide" commit moves a whole lot of the previous "add debugging tips for vcg goals", would it make sense to merge the two?

Oh, yes, absolutely. I thought I had already squashed them, but it seems I did not.

docs/vcg-debugging.md
lsf37: I should indent, markdown does show the difference

docs/vcg-debugging.md
lsf37: good idea, will add

PR #742, opened 2024-03-26
Add `deletionIsSafe2` assert to `deleteObjects`

michaelmcinerney:
Maybe the commit message could do with some elaboration. I can add something if so.

This currently includes a commit from the "yet more rules for `Lib`" PR, so should be merged after that PR is merged.

### DIFF
--- a/proof/crefine/ARM/Detype_C.thy
+++ b/proof/crefine/ARM/Detype_C.thy
@@ -1428,7 +1428,7 @@ lemma deleteObjects_ccorres':
              doMachineOp_modify modify_modify o_def ksPSpace_ksMSu_comm
              bind_assoc modify_machinestate_assert_cnodes_swap
              modify_modify_bind)
-  apply (rule ccorres_stateAssert_fwd)
+  apply (rule ccorres_stateAssert_fwd)+
   apply (rule ccorres_stateAssert_after)
   apply (rule ccorres_from_vcg)
   apply (rule allI, rule conseqPre, vcg)
--- a/proof/crefine/X64/Detype_C.thy
+++ b/proof/crefine/X64/Detype_C.thy
@@ -1542,7 +1542,7 @@ lemma deleteObjects_ccorres':
              doMachineOp_modify modify_modify o_def ksPSpace_ksMSu_comm
              bind_assoc modify_machinestate_assert_cnodes_swap
              modify_modify_bind)
-  apply (rule ccorres_stateAssert_fwd)
+  apply (rule ccorres_stateAssert_fwd)+
   apply (rule ccorres_stateAssert_after)
   apply (rule ccorres_from_vcg)
   apply (rule allI, rule conseqPre, vcg)
--- a/proof/refine/ARM/Detype_R.thy
+++ b/proof/refine/ARM/Detype_R.thy
@@ -100,6 +100,9 @@ defs deletionIsSafe_def:
        (\<forall>ko. ksPSpace s p = Some (KOArch ko) \<and> p \<in> {ptr .. ptr + 2 ^ bits - 1}
         \<longrightarrow> 6 \<le> bits)"
 
+defs deletionIsSafe_delete_locale_def:
+  "deletionIsSafe_delete_locale \<equiv> \<lambda>ptr bits s. \<forall>p. ko_wp_at' live' p s \<longrightarrow> p \<notin> {ptr .. ptr + 2 ^ bits - 1}"
+
 defs ksASIDMapSafe_def:
   "ksASIDMapSafe \<equiv> \<lambda>s. \<forall>asid hw_asid pd.
      armKSASIDMap (ksArchState s) asid = Some (hw_asid,pd) \<longrightarrow> page_directory_at' pd s"
@@ -116,6 +119,7 @@ lemma deleteObjects_def2:
   "is_aligned ptr bits \<Longrightarrow>
    deleteObjects ptr bits = do
      stateAssert (deletionIsSafe ptr bits) [];
+     stateAssert (deletionIsSafe_delete_locale ptr bits) [];
      doMachineOp (freeMemory ptr bits);
      stateAssert (\<lambda>s. \<not> cNodePartialOverlap (gsCNodes s) (\<lambda>x. x \<in> {ptr .. ptr + 2 ^ bits - 1})) [];
      modify (\<lambda>s. s \<lparr> ksPSpace := \<lambda>x. if x \<in> {ptr .. ptr + 2 ^ bits - 1}
@@ -127,6 +131,7 @@ lemma deleteObjects_def2:
      stateAssert ksASIDMapSafe []
    od"
   apply (simp add: deleteObjects_def is_aligned_mask[symmetric] unless_def deleteGhost_def)
+  apply (rule bind_eqI, rule ext)
   apply (rule bind_eqI, rule ext)
   apply (rule bind_eqI, rule ext)
   apply (simp add: bind_assoc[symmetric])
@@ -149,6 +154,7 @@ lemma deleteObjects_def3:
    do
      assert (is_aligned ptr bits);
      stateAssert (deletionIsSafe ptr bits) [];
+     stateAssert (deletionIsSafe_delete_locale ptr bits) [];
      doMachineOp (freeMemory ptr bits);
      stateAssert (\<lambda>s. \<not> cNodePartialOverlap (gsCNodes s) (\<lambda>x. x \<in> {ptr .. ptr + 2 ^ bits - 1})) [];
      modify (\<lambda>s. s \<lparr> ksPSpace := \<lambda>x. if x \<in> {ptr .. ptr + 2 ^ bits - 1}
@@ -432,6 +438,7 @@ next
 qed
 
 end
+
 locale detype_locale' = detype_locale + constrains s::"det_state"
 
 lemma (in detype_locale') deletionIsSafe:
@@ -443,9 +450,8 @@ lemma (in detype_locale') deletionIsSafe:
   shows       "deletionIsSafe base magnitude s'"
 proof -
   interpret Arch . (* FIXME: arch_split *)
-  note blah[simp del] =  atLeastatMost_subset_iff atLeastLessThan_iff
-          Int_atLeastAtMost atLeastatMost_empty_iff split_paired_Ex
-          atLeastAtMost_iff
+  note [simp del] = atLeastatMost_subset_iff atLeastLessThan_iff atLeastAtMost_iff
+                    Int_atLeastAtMost atLeastatMost_empty_iff split_paired_Ex
   have "\<And>t m r. \<exists>ptr. cte_wp_at ((=) (cap.ReplyCap t m r)) ptr s
         \<Longrightarrow> t \<notin> {base .. base + 2 ^ magnitude - 1}"
     by (fastforce dest!: valid_cap2 simp: cap obj_reply_refs_def)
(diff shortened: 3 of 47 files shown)


michaelmcinerney:
Using `schact_is_rct` in some of the top level theorems is actually a bit of a nuisance. I will attempt to redo some of this with as few changes to lemma statements as possible, to hopefully make other sessions like `InfoFlowC` pass without any further changes. I'll force push a change soon, and it would be best to hold off reviewing until then, though the change should be small

michaelmcinerney:
Tests are passing, and this is ready for review now. 

I went back and forth on where the `schact_is_rct` definition should be used. I think all the uses of it that were upsetting `InfoFlowC` were in `Syscall_R`, so I could probably use `schact_is_rct` in `Refine.thy` without upsetting anything else. But getting it all working was super annoying, so I'd rather leave it as is

lib/Corres_UL.thy
```diff
@@ -1256,6 +1256,18 @@ lemma corres_stateAssert_implied2:
   apply simp
   done
 
+lemma corres_stateAssert_add_assertion:
```
Xaphiosis, line 1259: not sure what we ended up agreeing on tagging commit messages to `lib/`, but this commit has the generic form of `lib: add something` that we discussed previously

Xaphiosis: unless this is the commit you refer to that's already in another PR, in which case please ignore

proof/crefine/ARM/Detype_C.thy
```diff
@@ -1428,7 +1428,7 @@ lemma deleteObjects_ccorres':
              doMachineOp_modify modify_modify o_def ksPSpace_ksMSu_comm
              bind_assoc modify_machinestate_assert_cnodes_swap
              modify_modify_bind)
-  apply (rule ccorres_stateAssert_fwd)
+  apply (rule ccorres_stateAssert_fwd)+
```
Xaphiosis, line 1431: not much of a save

proof/crefine/X64/Detype_C.thy
```diff
@@ -1542,7 +1542,7 @@ lemma deleteObjects_ccorres':
              doMachineOp_modify modify_modify o_def ksPSpace_ksMSu_comm
              bind_assoc modify_machinestate_assert_cnodes_swap
              modify_modify_bind)
-  apply (rule ccorres_stateAssert_fwd)
+  apply (rule ccorres_stateAssert_fwd)+
```
Xaphiosis, line 1545: does this save lines on any arch?

spec/haskell/src/SEL4/Model/PSpace.lhs
```diff
@@ -262,11 +264,14 @@ Clear the ghost state for user pages, cnodes, and arch-specific objects within t
...
 >         stateAssert ksASIDMapSafe "Object deletion would leave dangling PD pointers"
 
-In "deleteObjects" above, we assert "deletionIsSafe"; that is, that there are no pointers to these objects remaining elsewhere in the kernel state. Since we cannot easily check this in the Haskell model, we assume that it is always true; the assertion is strengthened during translation into Isabelle.
+In "deleteObjects" above, we make two assertions, which, when taken together, say that there are no pointers to these objects remaining elsewhere in the kernel state. Since we cannot easily check this in the Haskell model, we assume that it is always true; the assertion is strengthened during translation into Isabelle. We separate these properties into two assertions, since they are shown to be true by different means.
 
 > deletionIsSafe :: PPtr a -> Int -> KernelState -> Bool
 > deletionIsSafe _ _ _ = True
 
+> deletionIsSafe2 :: PPtr a -> Int -> KernelState -> Bool
+> deletionIsSafe2 _ _ _ = True
```
Xaphiosis, line 273: It's probably too much to hope for, but do these assertions have any more specificity now that there's two of them? I'm worried about the `essay_version4_final(3)_final.doc` effect. Reading it as a reviewer I have no clue what they'd be, but I guess we can live with it.

proof/refine/X64/Syscall_R.thy
```diff
@@ -400,7 +400,7 @@ lemma performInvocation_corres:
   "\<lbrakk> inv_relation i i'; call \<longrightarrow> block \<rbrakk> \<Longrightarrow>
    corres (dc \<oplus> (=))
      (einvs and valid_invocation i
-            and simple_sched_action
+            and schact_is_rct
```
Xaphiosis, line 403: not very important, but I'm curious: any thoughts on why this (and related) change? I'm trying to remember whether the `simple_sched_action` was actually thought of as the minimum needed, or copy-pasted from elsewhere

proof/refine/ARM/Untyped_R.thy
```diff
@@ -4331,7 +4331,7 @@ lemma resetUntypedCap_corres:
   apply (frule if_unsafe_then_capD'[OF ctes_of_cte_wpD], clarsimp+)
   apply (frule(1) descendants_range_ex_cte'[OF empty_descendants_range_in' _ order_refl],
     (simp add: isCap_simps)+)
-  apply (intro conjI impI; clarsimp)
+  apply (auto simp: descendants_range_in'_def valid_untyped'_def)
```
Xaphiosis, line 4334: does this imply the previous frule step did not succeed in discharging the side-conditions? feels strange this is so different now

proof/refine/X64/Detype_R.thy
```diff
@@ -727,92 +591,95 @@ lemma sym_refs_ko_wp_atD:
...
+
 locale delete_locale =
-  fixes s and base and bits and ptr and idx and d
-  assumes cap: "cte_wp_at' (\<lambda>cte. cteCap cte = UntypedCap d base bits idx) ptr s"
-  and  nodesc: "descendants_range' (UntypedCap d base bits idx) ptr (ctes_of s)"
-  and    invs: "invs' s"
-  and  ct_act: "ct_active' s"
-  and sa_simp: "sch_act_simple s"
-  and     bwb: "bits < word_bits"
+  fixes s' and base and bits and ptr and idx and d
```
Xaphiosis, line 598: is there any reason the delete_locale got a rename of `s` to `s'`? this generated a big diff

proof/refine/X64/Detype_R.thy
```diff
@@ -115,6 +118,7 @@ lemma deleteObjects_def2:
   "is_aligned ptr bits \<Longrightarrow>
    deleteObjects ptr bits = do
      stateAssert (deletionIsSafe ptr bits) [];
+     stateAssert (deletionIsSafe2 ptr bits) [];
```
Xaphiosis, line 121: I don't get it, if we are asserting one immediately after the other, why do you need two?

Xaphiosis (approved):
I'm fine with the changes, but would appreciate more information as to why stuff is happening. With nothing in the commit message, and not much in the PR description, I'm left with this vibe of reading the diff to figure out why I'm reading the diff. Probably you discussed it with someone else and they have it fresh in their memory, but in a year that won't be the case.

In particular, for expanding the commit message, as a reviewer who is a bit of an outsider on this:
* you're changing an invariant definition, to wrap up a specific term, would
  be useful to know that before having to look through a large diff
* why is this needed, when it was not needed before? it doesn't seem to have anything directly to do with your assertion that's in the commit message (does not appear in the assertion definition)
* you're adding a second assertion instead of expanding the first, why?

This would also be useful to know before starting to review the PR (usually PR description can be derived from commit message(s)).

I didn't put feedback on every instance of something, so if you do address it, don't forget it applies to the other arches too.

Oh, and it's impressive how fast you're pushing this stuff through, just don't forget to leave some breadcrumbs for those catching up. Thanks.

lib/Corres_UL.thy
michaelmcinerney: This commit is from another PR, but if there is an issue with the commit message, then I guess it needs to be fixed there. 

I thought the issue with `lib: add blah` was that we might not necessarily know "where" in `lib` we were adding the rule to. I did ask for some clarification on this in the guide we have on commit messages, but this hasn't happened yet. I thought that in this case, since one would assume (correctly, in this case) that a lemma related to `corres` would be in `Corres_UL` or `ExtraCorres`, further specificity wasn't needed here. If the current commit message isn't OK, could you suggest a better one please?

proof/crefine/ARM/Detype_C.thy
michaelmcinerney: We now have two asserts in `deleteObject` so the `+` allows us to gobble up all of them easily. 

For MCS, we would often add a bunch of asserts to the beginnings of functions (for `sym_refs (state_refs_of' s)`, `valid_idle` etc. etc.) and a `+` or the `intro` method for some assert rule worked well. It's not really intended to save lines, exactly, but hopefully make the proofs a little more resilient to further asserts being added. But I can change this if that's not a good approach

spec/haskell/src/SEL4/Model/PSpace.lhs
michaelmcinerney: The comment just before `deletionIsSafe` does talk a little bit about their intent. I could be more precise, but it's one of those situations where I would be referencing something that happens within the Isabelle theory files, and so if the definitions change there, the comment becomes outdated.

As for why there are two asserts, again the comment before these definitions is trying to say that there are two because they're proved by different means. Maybe I could slightly strengthen the comment to say "We separate these properties into two assertions only because they are shown to be true by different means."

proof/refine/X64/Detype_R.thy
michaelmcinerney: I think it was actually unnecessary, but I think I did this for MCS originally because it helped me keep track of which states are abstract and which are concrete: we have the `detype_locale'` here which has both abstract and concrete states, and then the `delete_locale` (which I modify here) which has just the concrete state. I can revert this, but I should point out that it's like this on the `rt` branch

lib/Corres_UL.thy
Xaphiosis: OK, I looked though a while of history of `Corres_UL.thy` and yeah, you have a point. Avoiding the by-the-way commits (changes to files based on some other goal than adding a rule), they oscillate between doing what you did, and adding a `to Corres_UL` at the end. I have a minor preference for the latter, because it could be one of two files, but it's definitely survivable if you don't change it.
I'm trying to steer you towards documenting stuff by default, but in my efforts some of the cases might be too trivial. Apologies on this one.

proof/refine/X64/Detype_R.thy
Xaphiosis: No no, please keep it, it's better when things make intuitive sense. It did create a bit of diff noise, so I wanted to know why.

proof/crefine/ARM/Detype_C.thy
Xaphiosis: No, I missed it. the combination of adding an assert *and* a plus looks like it doesn't save a line, but it does. All good.

michaelmcinerney:
> I'm fine with the changes, but would appreciate more information as to why stuff is happening. With nothing in the commit message, and not much in the PR description, I'm left with this vibe of reading the diff to figure out why I'm reading the diff. Probably you discussed it with someone else and they have it fresh in their memory, but in a year that won't be the case.

Yes this was discussed on Mattermost a couple of weeks back. It came out of the work for linked lists. For that work, I did want to follow along with the reasoning I used on the `rt` branch for dealing with the `tcbSchedNext` and `tcbSchedPrev` pointers that are now in the TCB object, and the separate assert for `master` was the easiest way of going about this. Gerwin mentioned that it would be best if these changes could be done separately for `master`, which is why I've put this PR up, but unfortunately, there's nothing currently in `master` that necessitates this change. Should I mention in the commit message "an assert regarding object deletion for live objects is added, in preparation for future changes to the TCB object" or something like that?

> 
> In particular, for expanding the commit message, as a reviewer who is a bit of an outsider on this:
> 
> * you're changing an invariant definition, to wrap up a specific term, would
>   be useful to know that before having to look through a large diff

What is the invariant definition that I'm changing? This should just be adding an assert and reshuffling `Detype_R` to accommodate it

> * why is this needed, when it was not needed before? it doesn't seem to have anything directly to do with your assertion that's in the commit message (does not appear in the assertion definition)

Let me know if my "in preparation for future changes" idea above is a good one

spec/haskell/src/SEL4/Model/PSpace.lhs
Xaphiosis: The thing is, when a reviewer goes over your diff and doesn't have a good idea of what's happening or why, you're going to get comments that aren't very smart from the author's perspective. This is one of those times. I just don't get what's going on with the two asserts, and how they're shown by different means :/
It's a bit of an art to figure out whether it makes sense to put in enough info to explain that, or whether it's hopeless. My instinct says that if you added a separate assertion rather than meld it into the previous one, then it deserves a better name than `2`. Especially now that you explained the reasoning behind the `+` above, your goal is to be able to add more assertions... then it wouldn't be super nice if we had the infrastructure, but they ended up named `3`, `4` etc. That's what I mean by the `essay_version4_final(3)_final.doc` effect, but it's hard to communicate over text.

proof/refine/X64/Syscall_R.thy
michaelmcinerney: All the changes for the scheduler action in later files come down to the beefed up guards in `deleteObjects_corres`. I was perhaps slightly lazy and simply went along with what's on the `rt` branch here, namely, `schact_is_rct`. It seems like I could actually weaken it to `simple_sched_action`. With nothing in Refine or CRefine arch split it's a nightmare to go and change things like this, but I can go back and weaken things

spec/haskell/src/SEL4/Model/PSpace.lhs
michaelmcinerney: The first assert is shown to be true in the lemma `deletionIsSafe` (terrible name...) which is in the `detype_locale'`. The second assert is shown to be true in the lemma `deletionIsSafe2_holds` in `delete_locale`. These locales aren't comparable, so there's no nice way of proving them in one go, hence their separation.

Yes I could come up with more specific names for these asserts, but the first assert seems like a big of a grab bag of properties that we want to hold for deletion. I would have simply added another conjunct instead of writing a new assert had it been straightforward to prove it all in one go. If you think I should change the name, I can. Maybe `live_detype` or something more longwinded and more ugly

Xaphiosis:
> > I'm fine with the changes, but would appreciate more information as to why stuff is happening. With nothing in the commit message, and not much in the PR description, I'm left with this vibe of reading the diff to figure out why I'm reading the diff. Probably you discussed it with someone else and they have it fresh in their memory, but in a year that won't be the case.
> 
> Yes this was discussed on Mattermost a couple of weeks back. It came out of the work for linked lists. For that work, I did want to follow along with the reasoning I used on the `rt` branch for dealing with the `tcbSchedNext` and `tcbSchedPrev` pointers that are now in the TCB object, and the separate assert for `master` was the easiest way of going about this. Gerwin mentioned that it would be best if these changes could be done separately for `master`, which is why I've put this PR up, but unfortunately, there's nothing currently in `master` that necessitates this change. Should I mention in the commit message "an assert regarding object deletion for live objects is added, in preparation for future changes to the TCB object" or something like that?

Yeah, but, your use of English is excellent and better than mine, can you try make it a bit... less minimal? As in convey some of the above, now that you already have the text. It's preparatory work to switch the design spec ready queues over to next/prev pointers in TCBs, but I'm missing some knowledge on how these changes relate to that. Anything interesting on why we need the live for the delete?

I like something like "Add assert regarding object deletion for live objects" but would appreciate an expansion of the second half. I realise it might be hard to think in terms of "what does someone who didn't do this work want to know", but I'm trying to convey that it was difficult to make sense of this PR until reading through the entire diff and getting some answers from you.

> > In particular, for expanding the commit message, as a reviewer who is a bit of an outsider on this:
> > 
> > * you're changing an invariant definition, to wrap up a specific term, would
> >   be useful to know that before having to look through a large diff
> 
> What is the invariant definition that I'm changing? This should just be adding an assert and reshuffling `Detype_R` to accommodate it

Uhh, I saw `theorem` about `callKernel` preconditions, did I get confused? That's why I'm saying some info is useful before the review starts. Now I'm starting to remember some work for domain ticks I did ages ago... you're right these conditions aren't invariant. See what I wrote in another comment about potentially silly sounding comments by confused reviewers.

> > * why is this needed, when it was not needed before? it doesn't seem to have anything directly to do with your assertion that's in the commit message (does not appear in the assertion definition)
> 
> Let me know if my "in preparation for future changes" idea above is a good one

Yes, good idea, I encourage you to step a bit harder on the gas pedal :)

proof/refine/X64/Syscall_R.thy
Xaphiosis: I'll buy the argument that it reduces the diff to `rt`. There are many preconditions we have that aren't strictly weakest preconditions, so it's not a significant effect. Thanks for explaining.

proof/refine/ARM/Untyped_R.thy
michaelmcinerney: I unfortunately know very little about these properties, so I'm not sure exactly what's going on here: I again simply copied what was done on the `rt` branch. None of those `frule`s can be removed, so I imagine they're doing something. I can look into this some more if it seems like there might be an issue

Xaphiosis: I didn't mean removing the frules, I was curious why we needed the switch to auto with args with what seems like an unrelated change. It doesn't make a big difference, and if `rt` has it anyway, shrugs.

spec/haskell/src/SEL4/Model/PSpace.lhs
Xaphiosis: Cool, that helps me understand what's going on, thanks.
Maybe when democracy rolls over this review, they'll have more ideas. How about renaming `deletionIsSafe2` to `deletionIsSafeLiveDetype` or something like that (`_live_detype` can also work)? The first `deletionIsSafe` is historically a grab-bag already, renaming that is probably a waste of effort.

michaelmcinerney:
> Yeah, but, your use of English is excellent and better than mine, can you try make it a bit... less minimal? As in convey some of the above, now that you already have the text. It's preparatory work to switch the design spec ready queues over to next/prev pointers in TCBs, but I'm missing some knowledge on how these changes relate to that. Anything interesting on why we need the live for the delete?

It's already the case that no live object exists in the region to be detyped, but I think I did want this to be known explicitly within the proof of `deleteObject_corres`. Not that the following explanation will be of any use in the commit messages, seeing that they won't be able to see the definitions of `live'`, but an object will (soon) be `live'` if the `tcbSchedPrev` or `tcbSchedNext` pointer is not `None`. I think I wanted the assert in order to make some arguments about what object deletion does to the heap projections `tcbSchedNexts_of/tcbSchedPrevs_of` so that we can show that `deleteObjects` establishes `ready_queues_relation`.

> 
> I like something like "Add assert regarding object deletion for live objects" but would appreciate an expansion of the second half. I realise it might be hard to think in terms of "what does someone who didn't do this work want to know", but I'm trying to convey that it was difficult to make sense of this PR until reading through the entire diff and getting some answers from you.

Yes I thought that the Mattermost discussion would have been enough background for this PR but I guess it wasn't. Sorry about that.

> Uhh, I saw `theorem` about `callKernel` preconditions, did I get confused? That's why I'm saying some info is useful before the review starts. Now I'm starting to remember some work for domain ticks I did ages ago... you're right these conditions aren't invariant. See what I wrote in another comment about potentially silly sounding comments by confused reviewers.

This does rearrange some `scheduler_action` statements into `schact_is_rct` but that's about all that's changed for the top level theorems.

spec/haskell/src/SEL4/Model/PSpace.lhs
michaelmcinerney: I'd be happy with your suggestion. Maybe we'd like to make the `2` version a grab bag for other properties that are proved in the same way, in which case we're back to having no good specific name, but I suppose we can cross that bridge when we come to it

Xaphiosis:
> > ...
> It's already the case that no live object exists in the region to be detyped, but I think I did want this to be known explicitly within the proof of `deleteObject_corres`. Not that the following explanation will be of any use in the commit messages, seeing that they won't be able to see the definitions of `live'`, but an object will (soon) be `live'` if the `tcbSchedPrev` or `tcbSchedNext` pointer is not `None`. I think I wanted the assert in order to make some arguments about what object deletion does to the heap projections `tcbSchedNexts_of/tcbSchedPrevs_of` so that we can show that `deleteObjects` establishes `ready_queues_relation`.

That's a pretty good explanation. I would not mind something like this ("a TCB will (soon) be `live'` if the [next/previous queue pointers] ...", combined with our discussion above) if it makes it into a commit message.
 
> > I like something like "Add assert regarding object deletion for live objects" but would appreciate an expansion of the second half. I realise it might be hard to think in terms of "what does someone who didn't do this work want to know", but I'm trying to convey that it was difficult to make sense of this PR until reading through the entire diff and getting some answers from you.
> 
> Yes I thought that the Mattermost discussion would have been enough background for this PR but I guess it wasn't. Sorry about that.

Could be my bad, lots of mattermost messages + deadlines.

> > Uhh, I saw `theorem` about `callKernel` preconditions, did I get confused? That's why I'm saying some info is useful before the review starts. Now I'm starting to remember some work for domain ticks I did ages ago... you're right these conditions aren't invariant. See what I wrote in another comment about potentially silly sounding comments by confused reviewers.
> 
> This does rearrange some `scheduler_action` statements into `schact_is_rct` but that's about all that's changed for the top level theorems.

Cool. Maybe drop that in the commit message with the argument that it's to match rt/mcs.

spec/haskell/src/SEL4/Model/PSpace.lhs
Xaphiosis: Cool. If it comes to that, it'll either be in the first grab bag, the same locale (thus might not need a rename), or a different one which would require a new assert.

michaelmcinerney:
Now with a new and improved commit message. I didn't want to go into too much detail about which fields I'll be adding to the TCB object, but I hope that what I have here is informative enough

spec/haskell/src/SEL4/Model/PSpace.lhs
lsf37: I do share the concern that `2` is not a helpful name. Would it be future-proof-ish to postfix with the locale the second one is proved in? (I haven't reviewed that bit yet in the proofs).

proof/refine/X64/Detype_R.thy
lsf37: I was about to ask the same question. Thanks for the explanation, also happy to leave as is.

proof/refine/ARM/Detype_R.thy
```diff
@@ -100,6 +100,9 @@ defs deletionIsSafe_def:
        (\<forall>ko. ksPSpace s p = Some (KOArch ko) \<and> p \<in> {ptr .. ptr + 2 ^ bits - 1}
         \<longrightarrow> 6 \<le> bits)"
 
+defs deletionIsSafe2_def:
+  "deletionIsSafe2 \<equiv> \<lambda>ptr bits s. \<forall>p. ko_wp_at' live' p s \<longrightarrow> p \<notin> {ptr .. ptr + 2 ^ bits - 1}"
```
lsf37, line 104: I know this is coming in pretty late in the game, but we have this abbreviation `mask_range ptr bits` (in Word_Lib) for `{ptr .. ptr + mask bits}` which is equivalent to `{ptr .. ptr + 2 ^ bits - 1}` and has a bunch of nice automation setup. Would it be extremely annoying to update and use that one instead of the expanded form?

lsf37: Actually, I see that RISCV64 does use that one. Any specific reason why it's not used here?

lsf37 (approved):
Modulo the name discussion and the `mask_range` usage for the definition of the assertion, I'm happy with this as is.

proof/refine/ARM/Detype_R.thy
michaelmcinerney: I can never keep track of which architecture uses which approach for ranges like this. I just followed along with what is used in `deletionIsSafe`. Are we supposed to have `mask_range` used everywhere on all arches?

lsf37: For any new uses, yes. The older architectures don't use it yet in their definitions, because it didn't exist yet when those proofs were written, but any new material should use `mask_range` when applicable.

michaelmcinerney: Alright. I can update these

PR #744, opened 2024-03-27
aarch64 crefine: update to match other arches

corlewis:
This was a change made in https://github.com/seL4/l4v/commit/146eaa10fc2b3a41d26617b60571eb274535234e, as part of https://github.com/seL4/l4v/pull/696.

### DIFF
--- a/proof/crefine/AARCH64/VSpace_C.thy
+++ b/proof/crefine/AARCH64/VSpace_C.thy
@@ -335,7 +335,7 @@ lemma corres_symb_exec_unknown_r:
   assumes "\<And>rv. corres_underlying sr nf nf' r P P' a (c rv)"
   shows "corres_underlying sr nf nf' r P P' a (unknown >>= c)"
   apply (simp add: unknown_def)
-  apply (rule corres_symb_exec_r[OF assms]; wp select_inv no_fail_select)
+  apply (rule corres_symb_exec_r[OF assms]; wp select_inv)
   done
 
 lemma isPageTablePTE_def2:


Xaphiosis (approved):
could use more informative commit message (the "as part of" of your initial comment); optional.

corlewis:
> could use more informative commit message (the "as part of" of your initial comment); optional.

I've updated the commit message but left off the reference to the PR, I didn't think it would be too helpful there. Happy to update again if you disagree and think it would be.

corlewis:
> Cool. I agree with leaving the PR off, but my thought was to add a hint of what this was caused by. I guess a commit reference is good enough if someone's investigating.

I could do
```
This corresponds to a change made in 146eaa1, following adjustments to some no_fail attributes.
```
That's basically pulling in the commit message of 146eaa1, although I'm not sure if it would be enough to save curious people from needing to look more closely at that commit (and its parent, which makes the `lib` change).

lsf37 (approved):

lsf37:
> > Cool. I agree with leaving the PR off, but my thought was to add a hint of what this was caused by. I guess a commit reference is good enough if someone's investigating.
> 
> I could do
> 
> ```
> This corresponds to a change made in 146eaa1, following adjustments to some no_fail attributes.
> ```
> 
> That's basically pulling in the commit message of [146eaa1](https://github.com/seL4/l4v/commit/146eaa10fc2b3a41d26617b60571eb274535234e), although I'm not sure if it would be enough to save curious people from needing to look more closely at that commit (and its parent, which makes the `lib` change).

I'd be happy with that, but also happy to follow whatever @Xaphiosis decides.

Issue #729, opened 2024-03-07
Have wp warn when resulting in a goal with a schematic assumption.


Xaphiosis:
In a ccorres or wp proof with a `∃val. x = Some val` or anything like that, an accidental clarsimp will introduce new free variable, and if the schematic only depends on `x`, but now the postcondition relies on `val`, applying `wp` will result in a schematic assumption in the goal. The very next simp (because we're trying to avoid clarsimp, right? hence also no wpsimp) will the instantiate that schematic to `False` and we're in trouble.
In order to at least get a heads-up about this, `wp` should warn when this happens.

`wpfix` does not handle these cases, `datatype_schem` doesn't help either

Discussed a bit with @corlewis so far

corlewis:
In some sense there's two issues here, there's the new free variables that the schematic precondition doesn't depend on, and there's the fact that `wp` results in a schematic assumption in the goal.

I think the first is going to be hard to deal with. It's outside of the scope of `wpfix` and the only way to recover if it happens seems to be careful application of simp rules to rewrite the variables back into their original forms. Invoking `clarsimp` throughout the proof might help: if the assumptions are always in a consistent simp-normal form then the schematics preconditions will hopefully depend on the right variables. This does feel like a fragile strategy though.

The second issue is a bit more concerning, our standard `wp` setup shouldn't be creating schematic assumptions like that. In the specific case encountered here, I'm pretty sure that it is happening due to `hoare_vcg_precond_imp` being made a `wp_comb` rule in https://github.com/seL4/l4v/blob/master/proof/invariant-abstract/Syscall_AI.thy#L1149. I'd say that our current design of `wp` means that rules like that should only be automatically applied by `wp_pre`, which has some basic built-in protection to avoid schematic assumptions.

Removing that line will hopefully avoid this issue in this specific case, but who knows what other proofs have accidentally depended on it over the years. Even if we do remove it, adding a warning to `wp` still seems like a worthwhile idea; it's not generally something that we want the tool to do.

Xaphiosis:
Thanks for tracking down the `wp_comb`! Do you think you'll have any time/willingness post-deadlines to take a look at de-`hoare_vcg_precond_imp`ing? It's not good to have an unsafe rule waiting to explode like that, and I think you're most suited to understanding why it was put in in the first place, and what's the easiest workaround (probably `supply ...[wp_comb]` in affected lemmas?)

[...]

PR #747, opened 2024-04-09
Remove hoare_vcg_precond_imp from wp_comb

corlewis:
This being a `wp_comb` rule is a leftover from before `wp_pre` was implemented and it is being removed because in some cases it can cause unintended schematics to appear in the assumptions. See https://github.com/seL4/l4v/issues/729 for more discussion of this being a problem.

This commit also updates all of the proofs that accidentally depended on it being a wp_comb rule. In many cases this involved `including classic_wp_pre`, which locally returns the wp attributes to a state similar to what it was when the proofs were first written. Other cases required small rewrites, often involving using `wpsimp` instead of `wp`, and removing some uses of `wp (once)`.

### DIFF
--- a/proof/crefine/AARCH64/Fastpath_Equiv.thy
+++ b/proof/crefine/AARCH64/Fastpath_Equiv.thy
@@ -755,8 +755,7 @@ lemma fastpath_callKernel_SysCall_corres:
                               apply ((wp assert_inv threadSet_pred_tcb_at_state cteInsert_obj_at'_not_queued | wps)+)[1]
                              apply ((wp assert_inv threadSet_pred_tcb_at_state cteInsert_obj_at'_not_queued | wps)+)[1]
                             apply ((wp assert_inv threadSet_pred_tcb_at_state cteInsert_obj_at'_not_queued | wps)+)[1]
-                            apply (wp fastpathBestSwitchCandidate_lift[where f="threadSet f t" for f t])
-                             apply simp
+                            apply (wpsimp wp: fastpathBestSwitchCandidate_lift[where f="threadSet f t" for f t])
                             apply ((wp assert_inv threadSet_pred_tcb_at_state
                                        cteInsert_obj_at'_not_queued
                                     | wps)+)[1]
@@ -1494,7 +1493,7 @@ lemma fastpath_callKernel_SysReplyRecv_corres:
      (invs' and ct_in_state' ((=) Running) and (\<lambda>s. ksSchedulerAction s = ResumeCurrentThread)
          and cnode_caps_gsCNodes')
      (callKernel (SyscallEvent SysReplyRecv)) (fastpaths SysReplyRecv)"
-  including no_pre
+  including classic_wp_pre
   supply if_cong[cong] option.case_cong[cong]
   supply if_split[split del]
   supply user_getreg_inv[wp] (* FIXME *)
--- a/proof/infoflow/Ipc_IF.thy
+++ b/proof/infoflow/Ipc_IF.thy
@@ -601,6 +601,9 @@ lemma send_signal_reads_respects:
                         set_thread_state_runnable_equiv_but_for_labels get_simple_ko_wp
                         gts_wp update_waiting_ntfn_equiv_but_for_labels
                         blocked_cancel_ipc_nosts_equiv_but_for_labels
+            \<comment> \<open>FIXME: The following line is working around the fact that wp (once) doesn't invoke
+                      wp_pre. If that is changed then it could be removed.\<close>
+            | wp_pre0
             | wpc
             | wps)+
     apply (elim conjE)
@@ -1717,7 +1720,7 @@ lemma copy_mrs_globals_equiv:
   "\<lbrace>globals_equiv s and valid_arch_state and (\<lambda>s. receiver \<noteq> idle_thread s)\<rbrace>
    copy_mrs sender sbuf receiver rbuf n
    \<lbrace>\<lambda>_. globals_equiv s\<rbrace>"
-  unfolding copy_mrs_def including no_pre
+  unfolding copy_mrs_def including classic_wp_pre
   apply (wp | wpc)+
     apply (rule_tac Q="\<lambda>_. globals_equiv s" in hoare_strengthen_post)
      apply (wp mapM_wp' | wpc)+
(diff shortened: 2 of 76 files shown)


proof/infoflow/Ipc_IF.thy
```diff
@@ -601,6 +601,7 @@ lemma send_signal_reads_respects:
                         set_thread_state_runnable_equiv_but_for_labels get_simple_ko_wp
                         gts_wp update_waiting_ntfn_equiv_but_for_labels
                         blocked_cancel_ipc_nosts_equiv_but_for_labels
+            | wp_pre0
```
corlewis, line 604: This could probably be removed if we manage to add `wp_pre` to `wp (once)`. I should maybe leave a FIXME saying that.

lsf37: yes, please

proof/refine/ARM/VSpace_R.thy
```diff
@@ -883,7 +883,13 @@ lemma deleteASID_corres:
...
       apply wp
       apply (simp add: o_def)
-      apply (wp getASID_wp)
+      \<comment> \<open>FIXME: This proof weakens the schematic precondition so that an extra variable can be used,
+                and then uses `(clarsimp, assumption)` to resolve the schematic.
+                An alternative would be to use @{thm Some_to_the} to rewrite the conclusion, so that
+                the precondition does not need to depend on that variable.
+                Both approaches feel fragile.\<close>
+      apply (rule hoare_pre)
+       apply (wp getASID_wp)
```
lsf37, line 892: At least it does this explicitly now. I'd slightly favour the `Some_to_the` approach to be honest. At least that would likely fail fast when something changes, whereas an explicit schematic and then `assumption` might do weird things.

proof/refine/ARM_HYP/VSpace_R.thy
```diff
@@ -1426,7 +1426,13 @@ lemma deleteASID_corres:
        apply (simp add: mask_asid_low_bits_ucast_ucast)
       apply wp
       apply (simp add: o_def)
-      apply (wp getASID_wp)
+      \<comment> \<open>FIXME: This proof weakens the schematic precondition so that an extra variable can be used,
+                and then uses `(clarsimp, assumption)` to resolve the schematic.
```
lsf37, line 1430: (see other instance)

proof/refine/X64/VSpace_R.thy
```diff
@@ -349,7 +349,13 @@ lemma deleteASID_corres [corres]:
        apply (simp add: mask_asid_low_bits_ucast_ucast)
       apply wp
       apply (simp add: o_def)
-      apply (wp getASID_wp)
+      \<comment> \<open>FIXME: This proof weakens the schematic precondition so that an extra variable can be used,
+                and then uses `(clarsimp, assumption)` to resolve the schematic.
```
lsf37, line 353: (see other instance)

lsf37 (approved):
Thanks for doing this. Bit scary to see so many `no_pre` still in use. So much stuff to clean up..

proof/crefine/AARCH64/Fastpath_Equiv.thy
```diff
@@ -755,8 +755,7 @@ lemma fastpath_callKernel_SysCall_corres:
                               apply ((wp assert_inv threadSet_pred_tcb_at_state cteInsert_obj_at'_not_queued | wps)+)[1]
                              apply ((wp assert_inv threadSet_pred_tcb_at_state cteInsert_obj_at'_not_queued | wps)+)[1]
                             apply ((wp assert_inv threadSet_pred_tcb_at_state cteInsert_obj_at'_not_queued | wps)+)[1]
-                            apply (wp fastpathBestSwitchCandidate_lift[where f="threadSet f t" for f t])
-                             apply simp
+                            apply (wpsimp wp: fastpathBestSwitchCandidate_lift[where f="threadSet f t" for f t])
```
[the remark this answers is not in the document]
corlewis: I have to admit that I wasn't thinking too hard, and once I realised that changing `wp` to `wpsimp` would generally fix these proofs I just went with that. The main benefit of `wpsimp` is actually the implicit `+` inside of it, which leads to additional invocations of `wp_pre`. This means that in this case, if we want to guarantee that there's no `clarsimp` steps then we could do `(wp ...)+`.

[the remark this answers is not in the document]
Xaphiosis: The main issue with stability of fastpath proofs occurs in `Fastpath_C` in the `ccorres` proofs, I don't remember suffering this hell in the monadic rewrite proofs except when I manually tried to short-cut `Some/None` cases without using `wpc + monadic_rewrite_impossible`.

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -907,23 +907,19 @@ lemma as_user_valid_sched[wp]:
...
+  unfolding tcb_sched_action_def tcb_sched_dequeue_def
+  apply wpsimp
   apply (clarsimp simp add: valid_queues_def etcb_at_def not_queued_def
                        split: option.splits)
   done
 
+lemma switch_to_thread_ct_not_queued[wp]:
+  "\<lbrace>valid_queues\<rbrace> switch_to_thread t \<lbrace>\<lambda>rv s. not_queued (cur_thread s) s\<rbrace>"
+  unfolding switch_to_thread_def
+  by wpsimp
```
Xaphiosis, line 921: haha wow

Xaphiosis (approved):
I don't see anything more than what @lsf37 already pointed out. Nicely done, thank you.

proof/invariant-abstract/DetSchedSchedule_AI.thy
corlewis: Yeah, this was one of the early cases before I found `classic_wp_pre` and when I was trying to properly fix broken proofs. The previous proof did a lot of very convoluted work to get to the `tcb_sched_action tcb_sched_dequeue` step with the appropriate precondition in place. Pulling it out to a separate lemma made things a lot simpler.

proof/crefine/AARCH64/Fastpath_Equiv.thy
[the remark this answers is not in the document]
corlewis: Hmm, sounds like it's ok to leave this as is then? Everything else being equal, I prefer `wpsimp` to `wp+`.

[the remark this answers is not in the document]
lsf37: Yes, good to leave it from my side.

PR #748, opened 2024-04-18
Make hoare_seq_ext and hoare_vcg_seqE wp rules 

corlewis:
`hoare_seq_ext` was previously made a `wp` rule in the middle of `ainvs`, which was surprising and could lead to different behaviour depending on which session a proof was inside of. This change makes it a `wp` rule in the base `Monads` setup, along with improving consistency by adding similar rules for `validE` and its variants.

### DIFF
--- a/proof/access-control/ARM/ArchInterrupt_AC.thy
+++ b/proof/access-control/ARM/ArchInterrupt_AC.thy
@@ -87,17 +87,12 @@ lemma arch_decode_irq_control_invocation_authorised[Interrupt_AC_assms]:
        (args \<noteq> [] \<longrightarrow> (pasSubject aag, Control, pasIRQAbs aag (ucast (args ! 0))) \<in> pasPolicy aag))\<rbrace>
    arch_decode_irq_control_invocation info_label args slot caps
    \<lbrace>\<lambda>x _. arch_authorised_irq_ctl_inv aag x\<rbrace>, -"
-  unfolding decode_irq_control_invocation_def arch_decode_irq_control_invocation_def
+  unfolding decode_irq_control_invocation_def arch_decode_irq_control_invocation_def Let_def
             authorised_irq_ctl_inv_def arch_authorised_irq_ctl_inv_def arch_check_irq_def
   apply (rule hoare_gen_asmE)
-  apply (rule hoare_pre)
-   apply (simp add: Let_def split del: if_split cong: if_cong)
-   apply (wp whenE_throwError_wp hoare_vcg_imp_lift hoare_drop_imps
-          | strengthen  aag_Control_owns_strg
-          | simp add: o_def del: hoare_True_E_R)+
+  apply (wpsimp wp: weak_if_wp)
   apply (cases args, simp_all)
   apply (cases caps, simp_all)
-  apply (simp add: ucast_mask_drop)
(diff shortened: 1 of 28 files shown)


lsf37 (approved):
Cool, thank you for doing that. This should make the setup more consistent overall.

Surprising that it still did need a bunch of changes, but it looks like most of them actually made the proofs shorter, which is good.

corlewis:
> Surprising that it still did need a bunch of changes, but it looks like most of them actually made the proofs shorter, which is good.

A couple of the changes were because of `wp` being more likely to apply other rules before `hoare_seq_ext`, due to it now being one of the first rules added to the `wp` set. Most of them, however, were because of me also making `hoare_vcg_seqE` a `wp` rule. It would have been easier to not do that but I do think this way is more consistent, and yes, once I worked out how to update the proofs it generally seems to have improved them.

proof/access-control/ARM/ArchInterrupt_AC.thy
```diff
@@ -87,17 +87,12 @@ lemma arch_decode_irq_control_invocation_authorised[Interrupt_AC_assms]:
...
-  unfolding decode_irq_control_invocation_def arch_decode_irq_control_invocation_def
+  unfolding decode_irq_control_invocation_def arch_decode_irq_control_invocation_def Let_def
             authorised_irq_ctl_inv_def arch_authorised_irq_ctl_inv_def arch_check_irq_def
   apply (rule hoare_gen_asmE)
-  apply (rule hoare_pre)
-   apply (simp add: Let_def split del: if_split cong: if_cong)
-   apply (wp whenE_throwError_wp hoare_vcg_imp_lift hoare_drop_imps
-          | strengthen  aag_Control_owns_strg
-          | simp add: o_def del: hoare_True_E_R)+
+  apply (wpsimp wp: weak_if_wp)
```
Xaphiosis, line 93: could you say something about what `weak_if_wp` is or when to use it? (not in the commits, but I've either not heard of it or completely forgot)

Xaphiosis (approved):
Looks like bomb disarming footage. Thank you for digging into this.

proof/access-control/ARM/ArchInterrupt_AC.thy
corlewis:
```
⟦ ⦃P⦄ f ⦃Q⦄; ⦃P'⦄ f ⦃Q'⦄ ⟧ ⟹ ⦃P and P'⦄ f ⦃λr. if C r then Q r else Q' r⦄
```
In my mind it's for when there's an `if` in the postcondition where you don't need to know the `if` condition and don't want to lift it across the function for whatever reason.

From memory, my cases in this PR were generally `if` conditions on the return value that would have been painful to lift, and where the two branches were either roughly the same or one was a form of `True`. As an alternative I probably could have used the simplifier to handle the `if`, but this approach with `weak_if_wp` was the one I found first.

Xaphiosis: That's awesome, I hope I won't forget it. It feels better than exciting combination of `if_apply_def2` and `hoare_drop_imps`, and should work in more cases. It's interesting that this isn't achieved with a `wp <some_rule>`, but I guess it does need to shuffle the order.

PR #753, opened 2024-05-07
verification for IOAPIC PR seL4/seL4#896

lsf37:
seL4/seL4#896 changes IOAPIC-related decoding slightly and requires a few new small invariants on the Haskell level.

Test with: seL4/seL4#896


proof/crefine/X64/ADT_C.thy
```diff
@@ -541,6 +541,10 @@ end
 
 context state_rel begin
 
+definition
+  "ioapic_nirqs_from_to_H cstate \<equiv>
+     \<lambda>x. if x \<le> of_nat maxNumIOAPIC then ioapic_nirqs_' cstate.[unat x] else 0"
```
corlewis, line 546: Is the 'from' needed in this name, and if so should it say what it's from?

proof/crefine/X64/Interrupt_C.thy
```diff
@@ -573,23 +602,49 @@ lemma ioapic_decode_map_pin_to_vector_ccorres:
...
+     apply (simp add: Kernel_Config.maxNumIOAPIC_def)
+     apply ccorres_rewrite
+     apply (rename_tac ioapic_nirqs)
+     apply (rule_tac Q="\<lambda>s. ioapic_nirqs = x64KSIOAPICnIRQs (ksArchState s) \<and>
+                            (0 < x64KSIOAPICnIRQs (ksArchState s) ioapic)" and Q'=\<top>
+                     in ccorres_split_when_throwError_cond)
+        apply (fastforce simp: word_le_def scast_ucast_up_eq_ucast uint_up_ucast is_up
+                               rf_sr_x64KSIOAPICnIRQs Kernel_Config.maxNumIOAPIC_def
+                               uint_less_1_eq)
+      (* Need to VCG it as the range error depends on the global state *)
```
corlewis, line 618: Thanks for these comments.

corlewis (approved):
Looks good to me (assuming that the `Word_Lib` failure is trivial)!

Did the new invariant in `valid_arch_state'` really not need any proof updates to show that it's maintained? I'm actually pretty impressed if so, even with it only depending on such specific parts of the state, I would have thought that some of the higher level functions would need at least some manual steps to show it.

proof/crefine/X64/ADT_C.thy
lsf37: That is totally a typo, thanks for spotting it!

proof/crefine/X64/Interrupt_C.thy
lsf37: I have the feeling I copied this from somewhere and shouldn't take credit for it. It sounds like something Raf might have written. But I approve as well :-)

lsf37:
> Looks good to me (assuming that the `Word_Lib` failure is trivial)!

It should be, I moved the lemma manually without checking dependencies, and that's what I get for my laziness :-)

> Did the new invariant in `valid_arch_state'` really not need any proof updates to show that it's maintained? I'm actually pretty impressed if so, even with it only depending on such specific parts of the state, I would have thought that some of the higher level functions would need at least some manual steps to show it.

I was also pretty impressed. AFAICS it works, because I managed to export the state dependency via abbreviation and therefore the lifting rule for `valid_arch_state'` still works as before and anything that unfolds `valid_arch_state'` will simplify out the new parts trivially. It's nice to see those ideas actually fully working every once in a while.

lib/Word_Lib/Word_Lemmas_Internal.thy
```diff
@@ -838,4 +838,19 @@ lemma ucast_ucast_mask_id:
    UCAST('b \<rightarrow> 'a) (UCAST('a \<rightarrow> 'b) (x && mask n)) = x && mask n" for x :: "'a::len word"
   by (simp add: ucast_ucast_len[OF eq_mask_less])
 
+lemma uint_less_1_eq:
```
Xaphiosis, line 841: the name sounds a bit strange, like uint of something less than 1, or the other meaning where "less 1 " means subtracting 1 is even more confusing... it should be in the family of `uint_less_le_eq`, so `uint_less_1_le_eq` maybe? `uint_minus_1_less_le_eq` is a bit too much.

lib/Word_Lib/Word_Lemmas_Internal.thy
```diff
@@ -838,4 +838,19 @@ lemma ucast_ucast_mask_id:
...
+lemma scast_ucast_up_minus_1_ucast:
+  assumes "LENGTH('a::len) < LENGTH('b::len)"
+  assumes "0 < w"
+  shows "SCAST('b \<rightarrow> 'c) (UCAST('a \<rightarrow> 'b) w - 1) = UCAST('a \<rightarrow> 'c::len) (w - 1)"
+  using assms
+  apply (subst scast_eq_ucast; simp)
+   apply (metis gt0_iff_gem1 msb_less_mono ucast_less_ucast_weak unsigned_0 word_upcast_neg_msb)
+  apply (smt (verit) lt1_neq0 nat_le_linear scast_ucast_simps(9) ucast_1 ucast_sub_ucast
+                     verit_comp_simplify1(3) word_le_nat_alt)
+  done
```
Xaphiosis, line 854: this proof is a bit yuck and clearly spat out of sledgehammer, but it looks like a copy-pasta

proof/crefine/X64/Interrupt_C.thy
```diff
@@ -530,32 +530,61 @@ lemma ccorres_pre_gets_x64KSNumIOAPICs_ksArchState:
...
 
+lemma ccorres_pre_gets_x64KSIOAPICnIRQs_ksArchState:
+  assumes cc: "\<And>rv. ccorres r xf (P rv) (P' rv) hs (f rv) c"
+  shows   "ccorres r xf
+                   (\<lambda>s. (\<forall>rv. x64KSIOAPICnIRQs (ksArchState s) = rv  \<longrightarrow> P rv s))
+                   {s. \<forall>rv. s \<in> P' rv } hs
+                   (gets (x64KSIOAPICnIRQs \<circ> ksArchState) >>= (\<lambda>rv. f rv)) c"
+  apply (rule ccorres_guard_imp)
+    apply (rule ccorres_symb_exec_l)
+       defer
```
Xaphiosis, line 541: this is a bit of a hot-take, but looks like the other proofs, so I guess we'll keep it

proof/crefine/X64/Interrupt_C.thy
```diff
@@ -530,32 +530,61 @@ lemma ccorres_pre_gets_x64KSNumIOAPICs_ksArchState:
...
-          whenE (uint (ioapicIRQLines - 1) < uint pin)
-                     (throwError (Inl (RangeError 0 (ioapicIRQLines - 1))));
+          whenE (uint (ucast (ioapic_nirqs ioapic - 1) :: machine_word) < uint pin)
+                     (throwError (Inl (RangeError 0 (ucast (ioapic_nirqs ioapic - 1)))));
           whenE (1 < uint level) (throwError (Inl (RangeError 0 1)));
           whenE (1 < uint polarity) (throwError (Inl (RangeError 0 1)))
        odE)
     (Call ioapic_decode_map_pin_to_vector_'proc)"
-  supply Collect_const[simp del]
+  supply Collect_const[simp del] word_less_1[simp del] (* for uniform array guard on ioapic_nirqs *)
```
Xaphiosis, line 581: very good to see

proof/crefine/X64/Interrupt_C.thy
```diff
@@ -573,23 +602,49 @@ lemma ioapic_decode_map_pin_to_vector_ccorres:
...
       apply (subst ucast_sub_ucast; fastforce simp: lt1_neq0)
+     apply (rule_tac P="numIOAPICs \<le> of_nat maxNumIOAPIC" in ccorres_gen_asm)
+     apply (clarsimp simp: not_less word_le_def[symmetric])
+     apply (prop_tac "ioapic < of_nat maxNumIOAPIC",
+            solves \<open>simp add: le_m1_iff_lt[THEN iffD1] word_neq_0_conv\<close>)
+     apply (simp add: Kernel_Config.maxNumIOAPIC_def)
+     apply ccorres_rewrite
+     apply (rename_tac ioapic_nirqs)
+     apply (rule_tac Q="\<lambda>s. ioapic_nirqs = x64KSIOAPICnIRQs (ksArchState s) \<and>
+                            (0 < x64KSIOAPICnIRQs (ksArchState s) ioapic)" and Q'=\<top>
```
Xaphiosis, line 613: style note: the `and Q'=⊤` is a bit hard to spot at the end of the `Q` instantiation like that

spec/cspec/c/gen-config-thy.py
```diff
@@ -170,6 +170,7 @@
     'CONFIG_CTZ_64': (bool, None),
     'CONFIG_CLZ_NO_BUILTIN': (bool, None),
     'CONFIG_CTZ_NO_BUILTIN': (bool, None),
+    'CONFIG_MAX_NUM_IOAPIC': (nat, 'maxNumIOAPIC'),
```
Xaphiosis, line 173: this is interesting; does this mean that if we need this for any specific architecture, we create a Kernel_Config entry for all architectures? i.e. does ARM now have a `maxNumIOAPIC`?

proof/refine/X64/Invariants_H.thy
```diff
@@ -1128,17 +1128,29 @@ definition
 where
   "valid_x64_irq_state' irqState \<equiv> \<forall>irq > maxIRQ. irqState irq = X64IRQFree"
 
-definition
-  valid_arch_state' :: "kernel_state \<Rightarrow> bool"
-where
+definition valid_ioapic_2 :: "machine_word \<Rightarrow> (machine_word \<Rightarrow> 8 word) \<Rightarrow> bool" where
+  "valid_ioapic_2 num_ioapics ioapic_nirqs \<equiv>
+    num_ioapics \<le> of_nat Kernel_Config.maxNumIOAPIC \<and>
```
Xaphiosis, line 1133: Sanity check: if you unfold `Kernel_Config.maxNumIOAPIC` in CRefine without a localised unfold in a lemma(s) that restrict the unfold to only generic properties we need, that is a pattern that prevents future parametrisation of that `Kernel_Config` entry. It's survivable for X64, but we should try avoid this pattern in principle (though this still happens in AARCH64, so it's clearly not completely escapable). Thoughts?

Xaphiosis (approved):
Cool, another long-standing seL4 PR closed.

lib/Word_Lib/Word_Lemmas_Internal.thy
lsf37: yes, this is a tweaked version of a similar lemma that also goes by sledgehammer

proof/refine/X64/Invariants_H.thy
lsf37: I was going to go the lazy way, because X64, but you're right that we should get into the habit. Will have a look.

lsf37: Ok, I managed to reduce the unfoldings by one. The other two are for array guards where `maxNumIOAPIC` corresponds to the dimension of the array, so the unfolding should be guaranteed to succeed. I've isolated them a little bit more.

spec/cspec/c/gen-config-thy.py
lsf37: yes, the tool has no way of distinguishing an option that is unset for a specific architecture from an option that is unset because it is not part of that architecture. That means we get all names for all architectures.

lsf37:
The Word_Lib failure has drawn in a whole slate of word lemma movement from CRefine, so it's probably worth looking at this again before I merge.

proof/crefine/Move_C.thy
```diff
@@ -325,19 +151,11 @@ lemma unat_ucast_prio_shiftr_simp[simp]:
   "unat (ucast (p::priority) >> n :: machine_word) = unat (p >> n)"
   by simp
 
-lemma from_bool_to_bool_and_1 [simp]:
-  assumes r_size: "1 < size r"
-  shows "from_bool (to_bool (r && 1)) = r && 1"
```
corlewis, line 330: Was this a duplicate / had already been moved at some point?

proof/crefine/Move_C.thy
```diff
@@ -325,19 +151,11 @@ lemma unat_ucast_prio_shiftr_simp[simp]:
...
-  from r_size have "r && 1 < 2"
-    by (simp add: and_mask_less_size [where n=1, unfolded mask_def, simplified])
-  thus ?thesis
-    by (fastforce simp add: from_bool_def to_bool_def dest: word_less_cases)
-qed
-
 lemma wb_gt_2:
   "2 < word_bits" by (simp add: word_bits_conv)
 
+declare from_bool_to_bool_and_1[simp]
```
corlewis, line 157: Do we want a FIXME to eventually move this as well? Or are we more likely to want to remove it.

corlewis (approved):
I'm confused by the remaining failures but assuming that it's mostly unrelated then it still looks good to me.

lsf37:
> I'm confused by the remaining failures but assuming that it's mostly unrelated then it still looks good to me.

They were because the seL4 PR was not rebased yet and didn't have some of the hyp related changes that happened in the meantime. I've pushed a rebased version of the PR and restarted the tests.

proof/crefine/Move_C.thy
lsf37: It's sort-of a duplicate in the sense that the assumption is not needed and the conclusion is an instance of the `from_bool_to_bool_and_1` that this one would shadow.

proof/crefine/Move_C.thy
lsf37: I wasn't sure either way, that's why I didn't put anything on there. At some point we should probably try to consolidate all those global declares, but for now I just wanted to preserve current behaviour.

proof/crefine/X64/Interrupt_C.thy
lsf37: It is a copy-paste of other `*_pre` lemmas, yes. I tried fiddling with it for a bit, but didn't come up with something nicer in reasonable time.

proof/crefine/lib/Ctac.thy
```diff
@@ -1949,6 +1949,18 @@ lemmas ccorres_move_const_guards
...
+  "\<lbrakk> G; ccorres_underlying rf_sr Gamm rrel xf arrel axf P P' hs m (c) \<rbrakk> \<Longrightarrow>
+   ccorres_underlying rf_sr Gamm rrel xf arrel axf P P' hs m (Guard F \<lbrace>G\<rbrace> c)"
+  by (rule ccorres_guard_imp, erule ccorres_move_const_guard; simp)
+
+lemma ccorres_prove_guard_seq:
+  "\<lbrakk> G; ccorres_underlying rf_sr Gamm rrel xf arrel axf P P' hs m (c;; d) \<rbrakk> \<Longrightarrow>
+   ccorres_underlying rf_sr Gamm rrel xf arrel axf P P' hs m (Guard F \<lbrace>G\<rbrace> c;; d)"
+  by (rule ccorres_guard_imp, erule ccorres_move_const_guard; simp)
+
+lemmas ccorres_prove_guard = ccorres_prove_guard_direct ccorres_prove_guard_seq
```
Xaphiosis, line 1962: are these moved from elsewhere, or new to this PR?

proof/crefine/X64/Interrupt_C.thy
```diff
@@ -573,23 +602,55 @@ lemma ioapic_decode_map_pin_to_vector_ccorres:
                             EXCEPTION_SYSCALL_ERROR_def EXCEPTION_NONE_def syscall_error_rel_def)
       apply (clarsimp simp: rf_sr_def cstate_relation_def carch_state_relation_def Let_def)
       apply (subst ucast_sub_ucast; fastforce simp: lt1_neq0)
+     apply (rule_tac P="numIOAPICs \<le> of_nat maxNumIOAPIC" in ccorres_gen_asm)
+     apply (clarsimp simp: not_less word_le_def[symmetric])
+     apply (prop_tac "ioapic < of_nat maxNumIOAPIC",
+            solves \<open>simp add: le_m1_iff_lt[THEN iffD1] word_neq_0_conv\<close>)
```
Xaphiosis, line 608: I like this approach. The array guard stuff is still annoying, and we should investigate the possibility of instrumenting the C parser to give these bounds names, maybe as part of <upcoming project>.

Xaphiosis (approved):
Approved v2, very nice, and I appreciate the emptying of Move_C.
Super minor nitpick:
`x64 ainvs+refine: ioapic_nirqs proof updates`
but
`x64 crefine: ioapic_nirqs updates`

proof/crefine/lib/Ctac.thy
lsf37: These are new (they are in for isolating that array guard for  `Kernel_Config.maxNumIOAPIC`)

PR #755, opened 2024-05-24
docs: document platform branch procedure

lsf37:
## Platform branches

- what they are
- how they work
- how to fix them when they break

### DIFF
--- a/docs/platform-branches.md
+++ b/docs/platform-branches.md
@@ -1,0 +1,105 @@
+<!--
+     Copyright 2024, Proofcraft Pty Ltd
+     SPDX-License-Identifier: CC-BY-SA-4.0
+-->
+
+# Automated Platform Branches
+
+The following branches for additional platforms are currently automatically
+rebased and tested on each push to `master`:
+
+- [imx8mm-fpu-ver](https://github.com/seL4/l4v/tree/imx8-fpu-ver)
+- [exynos5-ver](https://github.com/seL4/l4v/tree/exynos5-ver)
+
+## How this works
+
+The `L4V_PLAT` configuration setting provides an additional dimension to the
+configuration space for verified platforms in addition to `L4V_ARCH` and
+`L4V_FEATURES`. To test an `L4V_ARCH`, `L4V_FEATURES`, and `L4V_PLAT`
+combination, the test setup will attempt to load a corresponding kernel config
+from the seL4 repository under
+
+```sh
+configs/${L4V_ARCH}_${L4V_PLAT}_${L4V_FEATURES}_verified.cmake
+```
+
+e.g. `ARM_HYP_exynos5_verified.cmake`. Tests on the master branch will currently
+keep `L4V_PLAT` empty, and tests on the platform branches will provide the
+appropriate platform string.
+
+To keep platform branches up to date, there is [a GitHub action][rebase-action]
+that automatically attempts to rebase all platform branches over each new push
+to the master branch. If that rebase succeeds, this action will push the rebased
+platform branch to a candidate branch `<branch>-rebased`. The platform branch,
+and therefore the `-rebased` branch have a corresponding action that in turn is
+triggered by this push and runs a full proof test for this branch. If the test
+succeeds, the branch action replaces the original platform branch with the
+candidate `<branch>-rebased` branch by force push. If either step fails, CI will
+signal test failure and require manual intervention.
+
+When the test is triggered by a change in seL4 only, the rebase action will
+*not* re-trigger a new proof run for the rebase branches, because the rebase on
+the l4v repository itself will be empty and produce no change. However, if
+something substantial changed in seL4, it is very likely that there will be a
+corresponding l4v change to go with it, which then will trigger the rebase
+branches.
+
+[rebase-action]: https://github.com/seL4/l4v/blob/d869bb5ec81/.github/workflows/proof-deploy.yml#L89
+
+## How to fix things if they go wrong
+
+### Invariants
+
+- you should never have to modify the original platform branches
+
+- the original platform branches are updated by CI only
+
+- therefore, the original platform branches always are in a state that has
+  worked before with a specific seL4 version (the last one before CI signalled
+  failure)
+
+- the `-rebased` branches are ephemeral and can be force-pushed to
+
+### Rebase fails
+
+If the rebase action fails, for instance because there are conflicts with the
+master branch during `git rebase`, the manual steps to fixing the problem are
+for example:
+
+```sh
+# get up-to-date branches:
+git fetch
+# start from the old platform branch:
+git checkout verification/exynos5-ver
+# prepare a -rebased branch for it
+# The name is important, you may have to delete an old one if it already exists:
+git checkout -b exynos5-ver-rebased
+# rebase over current master:
+git rebase verification/master
+# manually fix rebase conflicts
+...
+# usually you want to check if evervthing works --
+# be sure to select correct L4V_ARCH and L4V_PLAT
+L4V_ARCH=ARM_HYP L4V_PLAT=exynos5 ./run_tests
+# push -rebased branch back to repo for automated test and update
+git push -f exynos5-ver-rebased verification
+```
+
+Note that you don't have to modify the existing `exynos5-ver` branch at all. CI
+will update it once the tests for `exynos5-ver-rebased` have passed.
+
+### Proof runs on the `-rebased` branch fails
+
+When the rebase has succeeded, but has resulted in a state where the proof check
+on the corresponding `-rebased` branch has failed, either from a CI run or from
+a manual run above, the procedure is to
+
+- check out the `-rebased` branch,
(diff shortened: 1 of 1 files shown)


docs/platform-branches.md
```diff
@@ -0,0 +1,98 @@
...
+that automatically attempts to rebase all platform branches over each new push
+to the master branch. If that rebase succeeds, this action will push the rebased
+platform branch to a candidate branch `<branch>-rebased`. The platform branch,
+and therefore the `-rebased` branch have a corresponding action that in turn is
+triggered by this push and runs a full proof test for this branch. If the test
+succeeds, the branch action replaces the original platform branch with the
+candidate `<branch>-rebased` branch by force push. If either step fails, CI will
+signal test failure and require manual intervention.
+
+[rebase-action]: https://github.com/seL4/l4v/blob/d869bb5ec81/.github/workflows/proof-deploy.yml#L89
```
Xaphiosis, line 40: this URL is refusing to load for me (blank page)

docs/platform-branches.md
```diff
@@ -0,0 +1,98 @@
...
+
+To keep platform branches up to date, there is [a GitHub action][rebase-action]
+that automatically attempts to rebase all platform branches over each new push
+to the master branch. If that rebase succeeds, this action will push the rebased
+platform branch to a candidate branch `<branch>-rebased`. The platform branch,
+and therefore the `-rebased` branch have a corresponding action that in turn is
+triggered by this push and runs a full proof test for this branch. If the test
+succeeds, the branch action replaces the original platform branch with the
+candidate `<branch>-rebased` branch by force push. If either step fails, CI will
+signal test failure and require manual intervention.
```
Xaphiosis, line 38: who gets notified, or where can we see that it broke?

Xaphiosis: Ooh, and to add to that, the above I was thinking of l4v master... but does the same thing happen with seL4 master changing (or going via manifest)?

docs/platform-branches.md
```diff
@@ -0,0 +1,98 @@
...
+# prepare a -rebased branch for it
+# The name is important, you may have delete an old one if it already exists:
+git checkout -b exynos5-ver-rebased
+# rebase over current master:
+git rebase verification/master
+# manually fix rebase conflicts
+...
+# usually you want to check if evervthing works --
+# be sure to select correct L4V_ARCH and L4V_PLAT
+L4V_ARCH=ARM_HYP L4V_PLAT=exynos5 ./run_tests
```
Xaphiosis, line 76: this is a bit off-topic for this PR, but the C kernel doesn't get regenerated if you change environment variables, right? So if someone had C code sitting around from master, and then ran the above command, would that mean everything would work up until CRefine broke? If so, it it worth to remind people at this point in the doc?

docs/platform-branches.md
```diff
@@ -0,0 +1,98 @@
...
+
+- check out the `-rebased` branch,
+- manually fix the proofs
+- usually these are very small changes -- if so, they should be squashed into
+  the original commits that make up the platform branch. If they are more
+  substantial new changes that make sense as a separate commit they can go on
+  top.
+- force-push the `-rebased` branch back (as `-rebased`)
+- CI will re-run the tests and update the main platform branch when the proofs
+  succeed.
```
Xaphiosis, line 98: if the regression test fails for infrastructure reasons, is the policy to try re-run it from github CI interface somehow?

docs/platform-branches.md
corlewis: It works for me.

docs/platform-branches.md
```diff
@@ -0,0 +1,98 @@
...
+master branch during `git rebase`, the manual steps to fixing the problem are
+for example:
+
+```sh
+# get up-to-date branches:
+git fetch
+# start from the old platform branch:
+git checkout verification/exynos5-ver
+# prepare a -rebased branch for it
+# The name is important, you may have delete an old one if it already exists:
```
corlewis, line 68:
```suggestion
# The name is important, you may have to delete an old one if it already exists:
```

docs/platform-branches.md
lsf37: For me as well, it's a standard GH permalink. Not sure if there is much I can do about that one.

Xaphiosis: It works now. Looks like some intermittent network problem on my side, because now HN doesn't load...

docs/platform-branches.md
lsf37: > Ooh, and to add to that, the above I was thinking of l4v master... but does the same thing happen with seL4 master changing (or going via manifest)?

Interesting question, I don't think we have thought through that scenario yet. The workflow triggers on seL4 changes, but there won't be anything to do for the rebase, since neither l4v master nor any of the l4v rebase branches will have changed. This means, it will not trigger a re-test of the platform branches. There could in theory be a change in seL4 that affects only those platforms and is invisible to the other ones, and those would be missed until the next l4v PR.

There is a `workflow_dispatch` trigger (= the one that is triggered by manifest/seL4 updates) in the branch workflows, but I'm not sure you can trigger anything else than the master branch with these. I should investigate.

lsf37: > Who gets notified, or where can we see that it broke? (either the rebase, or the regression test post-rebase)

Who gets notified is inscrutable GitHub magic. It does seem to figure out who pushed/merged/authored and notify that person. You can also see it in the test badges in the README -- if one of them is red and has a platform branch name, one of these tests has failed. If you then go to the Actions tab, you can see which ones.

I seem to get failure notifications on all actions that are not PRs from someone else. Not sure if that is because I'm watching the repository or if it is because the seL4 CI account re-routes to me and it does some of these pushes. 

> Further: if the rebase succeeds and the PR fails, will the next push to master result in another rebase attempt? 

Yes, it will trigger another attempt. If it failed at the rebase stage, it is likely to fail again if the last one failed.

> Will this get blocked/delayed by an existing regression test running on the rebased branch

A rebase failure won't interact with any running tests.

If two triggers made it past the rebase stage, the workflows will both keep running and they will both try to force-push their own state at the end. There is a race condition on which one wins, but only a working one will end up as the main platform branch. Unless the change makes something significantly faster, they'll proceed in the right order (i.e. later one wins if it passes).

If it's the wrong order, we can manually re-trigger the last one to run again, but I haven't seen that happening yet.

lsf37: > Right, because if we commit something to l4v master, that will trigger a rebase etc, and we will most likely do that at the same time as we advance the manifest to seL4 master... Do situations like bumping the seL4 version in cases where we expect the proof to go through trigger a normal verification test? I'd guess so, 

Yes, every seL4 change will trigger a re-run (even preprocess-equivalent ones these days)

> but then these ones won't trigger since there's nothing to rebase in that situation and they're not linked to the manifest. It wouldn't be the end of the world, but good to know/document, as I'm clearly a bit confused.

Yes, good idea. Will add.

docs/platform-branches.md
lsf37: We don't really have a policy for it as such, but I usually hit "Rerun failed jobs" when I see it.

docs/platform-branches.md
lsf37: I actually need to look into this a bit more to understand when exactly it will miss rebuilds. It does change the cmake file it is looking for, so something visible to `make` does change. Maybe what I should be doing is to also change the build directory name (i.e. encode L4V_PLAT in it, like we do L4V_ARM) -- I think then we might get the correct rebuilds. That'd be overall nicer. Will investigate.

lsf37: Ok, the wrinkle is that we do `install_C_file "../c/build/$L4V_ARCH/kernel_all.c_pp"` in `Kernel_C.thy`, and I don't think I can do any conditional magic in Isabelle there. We could set something like

```
KERNEL_BUILD_ROOT := build/${L4V_ARCH}$(if ${L4V_PLAT},_${L4V_PLAT},)
```

in the Makefile that calls the session `CKernel`, since that already needs to make sure it builds the C code first, and then use

```
install_C_file "../c/$KERNEL_BUILD_ROOT/kernel_all.c_pp
```

in `Kernel_C`. It will make everything one level more obscure, but it'd work.

lsf37: We could avoid conditionals and Makefile magic  if we accept `c/build/ARM_/` as the name of the build directory if there is no L4V_PLAT set and do

```
install_C_file "../c/build/$L4V_ARCH_$L4V_PLAT/kernel_all.c_pp"
```

in `Kernel_C`. Less pretty, but a lot less obscure. I might give that a shot for a PR.

michaelmcinerney (approved):
I'm not sure I can offer anything insightful to say, but the instructions look good to me.

docs/platform-branches.md
lsf37: This was a deeper rabbit hole than I thought. I'll adjust the description here for now and we can discuss more in #762.

PR #757, opened 2024-05-27
arm+arm-hyp spec+proofs: use UserContext datatype like other arches

Xaphiosis:
This reduces the diff to other architectures and FPU versions by having a UserContext datatype that contains the register state, instead of assuming the context is only the state of the registers.

Some of this touches old proofs and fitting in with style is wonky. Thoughts welcome.
This will break the FPU Arm platform port proofs, but once fixed the diff will be smaller.
Will squash commits to the above commit message (and same header as this PR) once people are ok with the changes.

### DIFF
--- a/proof/crefine/ARM/ADT_C.thy
+++ b/proof/crefine/ARM/ADT_C.thy
@@ -116,26 +116,24 @@ end
 definition
   "register_to_H \<equiv> inv register_from_H"
 
+context state_rel begin
+
 definition
   to_user_context_C :: "user_context \<Rightarrow> user_context_C"
 where
-  "to_user_context_C uc \<equiv> user_context_C (FCP (\<lambda>r. uc (register_to_H (of_nat r))))"
-
-context kernel_m begin
-
-lemma ccontext_rel_to_C:
-  "ccontext_relation uc (to_user_context_C uc)"
-  apply (clarsimp simp: ccontext_relation_def to_user_context_C_def)
-  apply (rule arg_cong [where f=uc])
-  apply (simp add: register_to_H_def inv_def)
-  done
-
-end
+  "to_user_context_C uc \<equiv>
+    user_context_C (ARRAY r. user_regs uc (register_to_H (of_nat r)))"
 
 definition
   from_user_context_C :: "user_context_C \<Rightarrow> user_context"
 where
-  "from_user_context_C uc \<equiv> \<lambda>r. index (registers_C uc) (unat (register_from_H r))"
+  "from_user_context_C uc \<equiv>
+     UserContext (\<lambda>r. (registers_C uc).[unat (register_from_H r)])"
+
+lemma (in kernel_m) ccontext_rel_to_C:
+  "ccontext_relation uc (to_user_context_C uc)"
+  unfolding ccontext_relation_def to_user_context_C_def cregs_relation_def
+  by (clarsimp simp: register_to_H_def inv_def)
 
 definition
   getContext_C :: "tcb_C ptr \<Rightarrow> cstate \<Rightarrow> user_context"
@@ -145,7 +143,12 @@ where
 
 lemma from_user_context_C:
   "ccontext_relation uc uc' \<Longrightarrow> from_user_context_C uc' = uc"
-  by (auto simp: ccontext_relation_def from_user_context_C_def)
+  unfolding ccontext_relation_def cregs_relation_def
+  apply (cases uc)
+  apply (auto simp: from_user_context_C_def)
+  done
+
+end
 
 context kernel_m begin
 
@@ -711,9 +714,16 @@ lemma cpspace_cte_relation_unique:
 lemma inj_tcb_ptr_to_ctcb_ptr: "inj tcb_ptr_to_ctcb_ptr"
   by (simp add: inj_on_def tcb_ptr_to_ctcb_ptr_def)
 
+lemma cregs_relation_imp_eq:
+  "cregs_relation f x \<Longrightarrow> cregs_relation g x \<Longrightarrow> f=g"
+  by (auto simp: cregs_relation_def)
+
 lemma ccontext_relation_imp_eq:
   "ccontext_relation f x \<Longrightarrow> ccontext_relation g x \<Longrightarrow> f=g"
-  by (rule ext) (simp add: ccontext_relation_def)
+  unfolding ccontext_relation_def
+  apply (cases f, cases g)
+  apply (auto dest: cregs_relation_imp_eq)
+  done
 
 lemma carch_tcb_relation_imp_eq:
   "carch_tcb_relation f x \<Longrightarrow> carch_tcb_relation g x \<Longrightarrow> f = g"
--- a/proof/crefine/ARM/IsolatedThreadAction.thy
+++ b/proof/crefine/ARM/IsolatedThreadAction.thy
@@ -9,6 +9,8 @@ theory IsolatedThreadAction
 imports ArchMove_C
 begin
 
+context begin interpretation Arch . (*FIXME: arch_split*)
+
 datatype tcb_state_regs =
   TCBStateRegs (tsrState : thread_state) (tsrContext : "MachineTypes.register \<Rightarrow> machine_word")
 
@@ -16,13 +18,16 @@ definition
   get_tcb_state_regs :: "kernel_object option \<Rightarrow> tcb_state_regs"
 where
  "get_tcb_state_regs oko \<equiv> case oko of
-    Some (KOTCB tcb) \<Rightarrow> TCBStateRegs (tcbState tcb) ((atcbContextGet o tcbArch) tcb)"
+    Some (KOTCB tcb) \<Rightarrow> TCBStateRegs (tcbState tcb) ((user_regs o atcbContextGet o tcbArch) tcb)"
 
 definition
   put_tcb_state_regs_tcb :: "tcb_state_regs \<Rightarrow> tcb \<Rightarrow> tcb"
 where
  "put_tcb_state_regs_tcb tsr tcb \<equiv> case tsr of
-     TCBStateRegs st regs \<Rightarrow> tcb \<lparr> tcbState := st, tcbArch := atcbContextSet regs (tcbArch tcb) \<rparr>"
+     TCBStateRegs st regs \<Rightarrow>
+        tcb \<lparr> tcbState := st,
+              tcbArch := atcbContextSet (UserContext regs)
+                         (tcbArch tcb) \<rparr>"
 
 definition
   put_tcb_state_regs :: "tcb_state_regs \<Rightarrow> kernel_object option \<Rightarrow> kernel_object option"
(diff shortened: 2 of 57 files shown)


proof/refine/ARM/CNodeInv_R.thy
```diff
@@ -5241,6 +5241,11 @@ lemma invalid_Thread_CNode:
   apply (clarsimp simp: obj_at'_def projectKOs)
   done
 
+(* FIXME MOVE *)
+lemma all_Not_False[simp]:
+  "All Not = False"
+  by blast
+
```
lsf37, line 5248: Have you tried if moving this to somewhere in Lib breaks stuff? It looks like a generally useful thing we should always have in scope.

proof/crefine/ARM_HYP/SyscallArgs_C.thy
```diff
@@ -1220,7 +1220,10 @@ lemma getSyscallArg_ccorres_foo:
      apply (simp add: word_less_nat_alt split: if_split)
     apply (rule ccorres_add_return2)
     apply (rule ccorres_symb_exec_l)
-       apply (rule_tac P="\<lambda>s. n < unat (scast n_msgRegisters :: word32) \<and> obj_at' (\<lambda>tcb. atcbContextGet (tcbArch tcb) (ARM_HYP_H.msgRegisters!n) = x!n) (ksCurThread s) s"
+       apply (rule_tac P="\<lambda>s. n < unat (scast n_msgRegisters :: word32)
```
lsf37, line 1223: Should this one be `word32` or `machine_word`?

proof/crefine/ARM_HYP/IsolatedThreadAction.thy
```diff
@@ -212,36 +213,49 @@ lemma isolate_thread_actions_asUser:
...
   apply (case_tac ko, simp)
+  apply (rename_tac uc)
+  apply (case_tac uc, simp add: modify_registers_def atcbContextGet_def atcbContextSet_def)
   done
 
-context begin interpretation Arch . (*FIXME: arch_split*)
-
 lemma getRegister_simple:
-  "getRegister r = (\<lambda>con. ({(con r, con)}, False))"
+  "getRegister r = (\<lambda>con. ({(user_regs con r, con)}, False))"
```
lsf37, line 221: Had forgotten that we did these monad unfoldings here. Those are going to bite when we move to the concurrency monad, but that's a problem for future us. Nothing to do with this PR in any case.

proof/crefine/ARM/SyscallArgs_C.thy
```diff
@@ -1186,7 +1187,10 @@ lemma getSyscallArg_ccorres_foo:
      apply (simp add: word_less_nat_alt split: if_split)
     apply (rule ccorres_add_return2)
     apply (rule ccorres_symb_exec_l)
-       apply (rule_tac P="\<lambda>s. n < unat (scast n_msgRegisters :: word32) \<and> obj_at' (\<lambda>tcb. atcbContextGet (tcbArch tcb) (ARM_H.msgRegisters!n) = x!n) (ksCurThread s) s"
+       apply (rule_tac P="\<lambda>s. n < unat (scast n_msgRegisters :: word32)
```
lsf37, line 1190: machine_word? (same as on ARM_HYP)

proof/crefine/ARM/ADT_C.thy
```diff
@@ -116,26 +116,24 @@ end
...
-lemma ccontext_rel_to_C:
-  "ccontext_relation uc (to_user_context_C uc)"
-  apply (clarsimp simp: ccontext_relation_def to_user_context_C_def)
-  apply (rule arg_cong [where f=uc])
-  apply (simp add: register_to_H_def inv_def)
-  done
-
-end
+  "to_user_context_C uc \<equiv>
+    user_context_C (ARRAY r. user_regs uc (register_to_H (of_nat r)))"
```
lsf37, line 125: 👍 for `ARRAY` over `FCP`

lsf37 (approved):
Nice, this is a step in the right direction for making things more generic.

proof/crefine/ARM_HYP/SyscallArgs_C.thy
Xaphiosis: I have no idea. It's casting `n_msgRegisters` which is an `int` due to being an enumeration, so that's `32 signed word` ... then it's trying to do a `unat`, so it's casting the `signed` away with `scast`, so technically that's to `32 word` or whatever we call it on other arches (`int_len word`?), which is 32 even on 64 bit arches? Then again `machine_word` is a suitably large word to carry it, so it seems arbitrary in this case. Since you pointed it out, what do you think? :)

proof/refine/ARM/CNodeInv_R.thy
Xaphiosis: I'm going to do that now. Since I can't access pcr I'll use the PR check to see if/where it fails.

proof/crefine/ARM_HYP/SyscallArgs_C.thy
lsf37: I was mostly checking if you had thought about it. If it's not really tied to machine_word it's fine to leave as is. We'll get to think properly about this bit when we attempt splitting in Refine

proof/refine/ARM/TcbAcc_R.thy
```diff
@@ -4221,34 +4255,34 @@ proof -
 
   show ?thesis using m
     unfolding setMRs_def set_mrs_def
-    apply (clarsimp simp: arch_tcb_set_registers_def arch_tcb_get_registers_def cong: option.case_cong split del: if_split)
+    apply (clarsimp  cong: option.case_cong split del: if_split)
```
corlewis, line 4258:
```suggestion
    apply (clarsimp cong: option.case_cong split del: if_split)
```

proof/refine/ARM_HYP/TcbAcc_R.thy
```diff
@@ -4272,34 +4306,34 @@ proof -
 
   show ?thesis using m
     unfolding setMRs_def set_mrs_def
-    apply (clarsimp simp: arch_tcb_set_registers_def arch_tcb_get_registers_def cong: option.case_cong split del: if_split)
+    apply (clarsimp  cong: option.case_cong split del: if_split)
```
corlewis, line 4309:
```suggestion
    apply (clarsimp cong: option.case_cong split del: if_split)
```

proof/crefine/ARM/IsolatedThreadAction.thy
```diff
@@ -233,6 +247,18 @@ lemma zipWithM_setRegister_simple:
                             simpler_modify_def fun_upd_def[symmetric])
   done
 
+(* these variant used in fastpath rewrite proof *)
```
corlewis, line 250:
```suggestion
(* this variant used in the fastpath rewrite proof *)
```

proof/crefine/ARM/IsolatedThreadAction.thy
```diff
@@ -233,6 +247,18 @@ lemma zipWithM_setRegister_simple:
...
+lemma setRegister_simple_modify_registers:
+  "setRegister r v = (\<lambda>con. ({((), modify_registers (\<lambda>f. f(r := v)) con)}, False))"
+  by (simp add: modify_registers_def setRegister_simple)
+
+(* these variant used in fastpath rewrite proof *)
```
corlewis, line 255:
```suggestion
(* this variant used in the fastpath rewrite proof *)
```

proof/crefine/ARM_HYP/IsolatedThreadAction.thy
corlewis: I don't know anything about how they're used but I don't think that it will be hard to write a similar unfolding for the trace monad. It would be a bit fiddly if we couldn't make the function atomic but even that shouldn't be too bad.

proof/drefine/Schedule_DR.thy
```diff
@@ -593,32 +593,39 @@ lemma schedule_dcorres:
...
   by (simp add: get_tcb_mrs_def Let_def arch_tcb_context_get_def)
 
 lemma transform_tcb_NextIP:
-  "transform_tcb m t (tcb_arch_update (tcb_context_update (\<lambda>ctx. ctx(NextIP:= pc))) tcb)
+  "transform_tcb m t (tcb_arch_update (tcb_context_update
+                                            (\<lambda>ctx. UserContext ((user_regs ctx)(NextIP := pc)))) tcb)
   = transform_tcb m t tcb"
+  apply (rule ext)
+  apply (simp add: transform_tcb_def transform_full_intent_def Let_def)
+
```
corlewis, line 628: Unintended newline?

corlewis (approved):
Very nice, definitely good to see more consistency!

proof/crefine/ARM/IsolatedThreadAction.thy
Xaphiosis: I was trying to say this one and `zipWithM_setRegister_simple_modify_registers` are used for the fastpath rewrite proof, so at least I tried to write "these variants" but clearly only made it halfway. Do you think the same comment above both then, or could people infer that the `_modify_registers` variants are related?

proof/refine/ARM/CNodeInv_R.thy
Xaphiosis: This causes a bunch of failure, might want to do it separately.

proof/refine/ARM/TcbAcc_R.thy
Xaphiosis: done

proof/refine/ARM_HYP/TcbAcc_R.thy
Xaphiosis: done

proof/drefine/Schedule_DR.thy
Xaphiosis: well spotted, fixed

spec/machine/ARM/MachineOps.thy
```diff
@@ -495,19 +495,20 @@ definition
...
-where
-  "getRegister r \<equiv> gets (\<lambda>uc. uc r)"
+definition getRegister :: "register \<Rightarrow> machine_word user_monad" where
+  "getRegister r \<equiv> gets (\<lambda>s. user_regs s r)"
 
 definition
-  setRegister :: "register \<Rightarrow> machine_word \<Rightarrow> unit user_monad"
-where
-  "setRegister r v \<equiv> modify (\<lambda>uc. uc (r := v))"
+  "modify_registers f uc \<equiv> UserContext (f (user_regs uc))"
```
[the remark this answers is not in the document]
Xaphiosis: will fix

proof/crefine/ARM_HYP/SyscallArgs_C.thy
Xaphiosis: ok, will leave it for now

spec/machine/ARM/MachineOps.thy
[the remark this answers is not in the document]
Xaphiosis: done

Xaphiosis:
Squashed commits down to 1 and hopefully addressed all PR feedback. Will now wait for tests to make sure I didn't mess something up.

PR #758, opened 2024-05-29
haskell-translator: avoid invalid escape sequence warnings

lsf37:
`"\<xsymbol>"` is not a valid string in python, although the compiler generally accepts it and does the right thing. More recent versions warn about it. To avoid that noise, turn these into either `r"\<xsymbol>"`, or `"\\<xsymbol>"` when the former is not possible.


tools/haskell-translator/lhs_pars.py
```diff
@@ -980,7 +980,7 @@ def type_wrapper_type(header, cons, rhs, d, decons=None):
             % (decons, header, header, header, header),
             'where',
             '  %s_update_def[simp]:' % decons,
-            ' "%s_update f y \<equiv> f y"' % decons,
+            ' "%s_update f y \\<equiv> f y"' % decons,
```
Xaphiosis, line 983: any reason behind this choice? testing on python3:
```
>>> r'test %s' % 3
'test 3'
```

seems to work. I don't really care that much, but I'm curious.

Xaphiosis (approved):
As I mentioned in comment, `r'blah %s' % whatever` seems to work so not sure what the criterion was for when to pick `\\`

tools/haskell-translator/lhs_pars.py
lsf37: I find it clearer to separate these -- the point of raw string literals is that there are no special characters. I guess the `%s` bit works because it's an implicit call to `format` on a string that is by then arbitrary (doesn't matter any more what kind of literal it was), but that's a bunch of speculation on my part. This way it's definitely clear what's happening, with `r` I would eternally be unsure when reading it.

PR #760, opened 2024-05-31
misc: remove obsolete BUILD_CACHE from isa-common.mk

lsf37:
Our `save_skip_cache` for build caching in our Isabelle branch has long been discontinued in favour of "defer proofs". Remove last traces from include makefile.


corlewis (approved):
Do we also want to remove `proof/refine/*/Cache.thy` and `proof/crefine/*/CACHE.ML`?

lsf37:
> Do we also want to remove `proof/refine/*/Cache.thy` and `proof/crefine/*/CACHE.ML`?

Awesome, yes we do. I was trying to look for them, but couldn't remember where they are. Thanks!

PR #761, opened 2024-06-02
make C build dependent on L4V_PLAT

lsf37:
The Makefile currently does not correctly track changes to L4V_PLAT, because that is an env variable and not a file.

By appending `_$L4V_PLAT` to the build directory, we make such changes visible in the file system, which means `make` will track things correctly again.

Using a plain `${L4V_ARCH}_${L4V_PLAT}` path will give us a trailing `_` when `L4V_PLAT` is empty, but this is only in generated build directories we usually don't look at, and it avoids making the `external_file` and `install_C_file` commands in Isabelle more complex.

### DIFF
--- a/spec/ROOT
+++ b/spec/ROOT
@@ -81,8 +81,8 @@ session ExecSpec in "design" = Word_Lib +
 
 session CSpec in "cspec" = CKernel +
   directories
-    "c/build/$L4V_ARCH/generated/arch/object"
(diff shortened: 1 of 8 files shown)


lsf37:
Unfortunately this doesn't work, because the path `${L4V_ARCH}_${L4V_PLAT}` is rejected by Isabelle. Can't have arbitrary substitution.

PR #764, opened 2024-06-11
Update to Isabelle2024

corlewis:
Updates all proofs and ML code to Isabelle2024.

The changes are very minor this time, and apart from one or two cases only involve ML code. The main ML differences are that:
* the simproc setup has changed slightly
* it is now an error to catch all exceptions and instead we use the new `\<^try>` antiquotation
* the underlying ML for the old-style `defs` command was downgraded, which meant that `defs` had to be correspondingly downgraded as well.

Nicely, initial testing did show some fairly significant performance improvements, particularly to the CPU time.
```
Isabelle2023:
  N = 22
  Average ARM CRefine: (00:28:45 real, 03:16:57 cpu, 22.11GB)
  Standard Deviation: (00:01:25 real, 00:16:05 cpu, 0.38GB)
Isabelle2024:
  N = 22
  Average ARM CRefine: (00:24:16 real, 02:22:21 cpu, 22.12GB)
  Standard Deviation: (00:00:29 real, 00:02:49 cpu, 0.40GB)
```

### DIFF
--- a/lib/Hoare_Sep_Tactics/Hoare_Sep_Tactics.thy
+++ b/lib/Hoare_Sep_Tactics/Hoare_Sep_Tactics.thy
@@ -344,8 +344,7 @@ done
 
 
 ML \<open>
-   fun J f x = f x
-               handle _ => x   (* FIXME! exceptions *)
+   fun J f x = \<^try>\<open>f x catch _ => x\<close>
 
    fun sep_wp thms ctxt  =
    let
--- a/lib/Defs.thy
+++ b/lib/Defs.thy
@@ -15,4 +15,3 @@ begin
 ML_file "defs.ML"
 
 end
-
--- a/lib/Eisbach_Tools/Apply_Debug.thy
+++ b/lib/Eisbach_Tools/Apply_Debug.thy
@@ -685,7 +685,7 @@ in
        let
         val (ctxt,thm) = get_state st;
 
-        val r = case Exn.interruptible_capture (fn st =>
+        val r = case Exn.result (fn st =>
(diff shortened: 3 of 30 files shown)


corlewis:
The last commit needs to be dropped before merging, it hopefully just makes the tests pick Isabelle2024 instead of Isabelle2023 for this PR.

Most of the commits labelled `update to Isabelle2024` could also probably be squashed together if we want to, the current set just reflects how they were developed.

lib/Hoare_Sep_Tactics/Hoare_Sep_Tactics.thy
```diff
@@ -344,8 +344,7 @@ done
 
 
 ML \<open>
-   fun J f x = f x
-               handle _ => x   (* FIXME! exceptions *)
+   fun J f x = \<^try>\<open>f x catch _ => x\<close>   (* FIXME! exceptions *)
```
corlewis, line 347: Hmm, this FIXME could possibly be removed now that we're using the `try` antiquotation.

lsf37: Yes, let's do that.

lsf37 (approved):
Looks good from my side. I'll prime the AWS VMs with the new Isabelle components, because it looks like some of them are failing on long downloads.

lsf37:
> I'll prime the AWS VMs with the new Isabelle components, because it looks like some of them are failing on long downloads.

Now done. New runs should not need to download components any more.

PR #759, opened 2024-05-30
Monad improvements

corlewis:
This improves the structure and naming consistency of the various monad rule sets, for both the nondet and trace monads. The main benefit is improved maintainability, in particular due to making sure that the nondet and trace monad rule sets are in sync.

There are 4 main categories of changes that were visible in the proofs:
* lemma renamings
* more consistently using `P` and `Q` for preconditions and postconditions
* the default name for bound variables produced when splitting on `bind` is now always `rv` instead of sometimes `r` or `x`. In the couple of cases where this caused confusion I added a `rename_tac` and fixed a useful name, but in general I kept this PR simple and just updated the uses of the bound variables.
* `hoare_True_E_R: : "⦃P⦄ f ⦃λ_ s. True⦄, -"` was removed from the `simp` set due to being unsafe, with `wp_post_tautE_R: ⦃λ_. True⦄ f ⦃λ_ _. True⦄` being added as a `wp` rule instead. This meant that some proof steps were changed from `simp` to `wp`, and that some proofs no longer needed to locally delete it from the `simp` set.

The lemma renamings needed in the proofs are listed in the third commit message. I have made comments in this PR showing examples of the other categories.

PR #765, opened 2024-06-11
More monad improvements

corlewis:
This makes further improvements to the consistency of the monad rule sets, following on from https://github.com/seL4/l4v/pull/759. In particular, it includes additional lemma renamings missed in that PR, and changes the variables used in the lemmas to avoid `R` and `G` being used in pre or post-conditions. These variable names are instead reserved for the rely and guarantee conditions in the trace monad's RG rule set.

### DIFF
--- a/lib/Monads/nondet/Nondet_More_VCG.thy
+++ b/lib/Monads/nondet/Nondet_More_VCG.thy
@@ -20,20 +20,20 @@ lemma hoare_take_disjunct:
   by (erule hoare_strengthen_post, simp)
 
 lemma hoare_post_add:
-  "\<lbrace>P\<rbrace> S \<lbrace>\<lambda>r s. R r s \<and> Q r s\<rbrace> \<Longrightarrow> \<lbrace>P\<rbrace> S \<lbrace>Q\<rbrace>"
+  "\<lbrace>P\<rbrace> f \<lbrace>\<lambda>r s. Q' r s \<and> Q r s\<rbrace> \<Longrightarrow> \<lbrace>P\<rbrace> f \<lbrace>Q\<rbrace>"
   by (erule hoare_strengthen_post, simp)
 
 lemma hoare_post_addE:
-  "\<lbrace>P\<rbrace> f \<lbrace>\<lambda>_ s. R s \<and> Q s\<rbrace>, \<lbrace>T\<rbrace> \<Longrightarrow> \<lbrace>P\<rbrace> f \<lbrace>\<lambda>_ s. Q s\<rbrace>, \<lbrace>T\<rbrace>"
+  "\<lbrace>P\<rbrace> f \<lbrace>\<lambda>_ s. Q' s \<and> Q s\<rbrace>,\<lbrace>E\<rbrace> \<Longrightarrow> \<lbrace>P\<rbrace> f \<lbrace>\<lambda>_ s. Q s\<rbrace>,\<lbrace>E\<rbrace>"
   by (erule hoare_strengthen_postE; simp)
 
 lemma hoare_pre_add:
-  "(\<forall>s. P s \<longrightarrow> R s) \<Longrightarrow> (\<lbrace>P\<rbrace> f \<lbrace>Q\<rbrace> \<longleftrightarrow> \<lbrace>P and R\<rbrace> f \<lbrace>Q\<rbrace>)"
+  "(\<forall>s. P s \<longrightarrow> P' s) \<Longrightarrow> \<lbrace>P\<rbrace> f \<lbrace>Q\<rbrace> \<longleftrightarrow> \<lbrace>P and P'\<rbrace> f \<lbrace>Q\<rbrace>"
   apply (subst iff_conv_conj_imp)
   by(intro conjI impI; rule hoare_weaken_pre, assumption, clarsimp)
 
 lemma hoare_pre_addE:
-  "(\<forall>s. P s \<longrightarrow> R s) \<Longrightarrow> (\<lbrace>P\<rbrace> f \<lbrace>Q\<rbrace>, \<lbrace>S\<rbrace> \<longleftrightarrow> \<lbrace>P and R\<rbrace> f \<lbrace>Q\<rbrace>, \<lbrace>S\<rbrace>)"
+  "(\<forall>s. P s \<longrightarrow> P' s) \<Longrightarrow> \<lbrace>P\<rbrace> f \<lbrace>Q\<rbrace>,\<lbrace>E\<rbrace> \<longleftrightarrow> \<lbrace>P and P'\<rbrace> f \<lbrace>Q\<rbrace>,\<lbrace>E\<rbrace>"
   apply (subst iff_conv_conj_imp)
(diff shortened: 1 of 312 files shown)


corlewis:
There are two new commits in this PR, which removes one lemma from Nondet_VCG, along with removing some FIXMEs that I had recently added and which I no longer think are worth doing.

I do still think that the lemmas the FIXMEs refer to are a bit unnecessary, but they're already used in enough proofs that they might as well be left in, and probably should even have their equivalents added to `Trace_RG`.

I removed `hoare_gets` specifically because it seemed to have the potential to cause problems, with one proof locally deleting it from the `intro` set. This did still require some proof updates but I actually think that they have ended up nicer, with a cleaner split between `wp` and `fastforce` steps.

proof/refine/X64/Retype_R.thy
```diff
@@ -4374,17 +4374,14 @@ lemma createNewCaps_idle'[wp]:
...
-                     threadSet_idle'
+             apply wpsimp
+            apply (wpsimp wp: mapM_x_wp' createObjects_idle' threadSet_idle'
                    | simp add: projectKO_opt_tcb projectKO_opt_cte
                                makeObject_cte makeObject_tcb archObjSize_def
                                tcb_cte_cases_def objBitsKO_def APIType_capBits_def
                                objBits_def createObjects_def bit_simps
                    | intro conjI impI
-                   | fastforce simp: curDomain_def)+
+                   | clarsimp simp: curDomain_def)+
```
Xaphiosis, line 4384: Is this a consequence of removing `hoare_gets`, or was fastforce more powerful than was needed in general? I've noticed a few instances of that in this commit.

lib/Monads/trace/Trace_VCG.thy
```diff
@@ -140,39 +139,39 @@ wpc_setup "\<lambda>m. \<lbrace>P\<rbrace> m -,\<lbrace>E\<rbrace>" wpc_helper_v
...
+  "\<lbrakk>\<And>rv. \<lbrace>Q' rv\<rbrace> g rv -,\<lbrace>E\<rbrace>; \<lbrace>P\<rbrace> f \<lbrace>Q'\<rbrace>,\<lbrace>E\<rbrace>\<rbrakk> \<Longrightarrow> \<lbrace>P\<rbrace> f >>=E (\<lambda>rv. g rv) -,\<lbrace>E\<rbrace>"
   apply (clarsimp simp: validE_E_def)
   by (wp | assumption)+
 
 lemmas bind_wp_fwd = bind_wp[rotated]
 lemmas bindE_wp_fwd = bindE_wp[rotated]
 
 lemma bind_wpE_R:
-  "\<lbrakk>\<And>x. \<lbrace>Q' x\<rbrace> g x \<lbrace>Q\<rbrace>,-; \<lbrace>P\<rbrace> f \<lbrace>Q'\<rbrace>\<rbrakk> \<Longrightarrow> \<lbrace>P\<rbrace> f >>= g \<lbrace>Q\<rbrace>,-"
+  "\<lbrakk>\<And>rv. \<lbrace>Q' rv\<rbrace> g rv \<lbrace>Q\<rbrace>,-; \<lbrace>P\<rbrace> f \<lbrace>Q'\<rbrace>\<rbrakk> \<Longrightarrow> \<lbrace>P\<rbrace> f >>= (\<lambda>rv. g rv) \<lbrace>Q\<rbrace>,-"
```
Xaphiosis, line 164: good, I'm glad you remember the significance of the etas

lib/Monads/nondet/Nondet_More_VCG.thy
```diff
@@ -517,36 +515,26 @@ lemma weaker_hoare_ifE:
...
-    apply simp
-   apply simp
-  apply (rule hoare_pre_imp)
-   prefer 2
-   apply simp
-  apply simp
-  done
+  "\<lbrakk> P \<or> P'; P \<Longrightarrow> \<lbrace>S\<rbrace> f \<lbrace>Q\<rbrace>; P' \<Longrightarrow> \<lbrace>T\<rbrace> f \<lbrace>Q\<rbrace> \<rbrakk>
+   \<Longrightarrow> \<lbrace>\<lambda>s. (P \<longrightarrow> S s) \<and> (P' \<longrightarrow> T s)\<rbrace> f \<lbrace>Q\<rbrace>"
+  by (fastforce intro: hoare_weaken_pre)
```
Xaphiosis, line 527: hahaha wow

Xaphiosis (approved):
I want to say a huge thanks for your recent crusade to reduce the jankiness of our proofs.

Feedback:

commits:

* is it worth having three commits for updating the proofs after the monad consistency commit? I'd squash at least the automated ones together, and maybe even the manual one.
* keeping the remove FIXMES ones separate is a good idea

commit messages:
* apart from hoare_gets -> apart from removing hoare_gets

questions:
* what the `-0777` argument to `perl` do?

proof/refine/X64/Retype_R.thy
corlewis: It's a consequence of removing `hoare_gets`. Roughly, this proof previously used `wp` until it hit `curDomain`, which the `fastforce` would unfold and then solve using `hoare_gets` as an `intro` rule. Now, the `clarsimp` is just being used to do the unfold and then it goes back to `wp` to solve the goal. I don't remember exactly why, but for some reason I thought that the unfold couldn't happen ahead of time.

corlewis: Looking at it again here, the `clarsimp` and `simp` could probably be combined though.

corlewis:
> Feedback:
>
> commits:
>
>     * is it worth having three commits for updating the proofs after the monad consistency commit? I'd squash at least the automated ones together, and maybe even the manual one.

I'm happy to squash the automated ones together, I think I mostly did it this way due to starting off with the easy lemma renames and then gradually working out what was needed to change the variables. I wouldn't mind keeping the manual one separate though, I expect it to help when merging into `rt` or any other ongoing work.

>     * keeping the remove FIXMES ones separate is a good idea

Agreed, the only one I'm not sure about is the FIXME that was added in the first commit of this PR.

> commit messages:
> 
>     * apart from hoare_gets -> apart from removing hoare_gets

👍 

> questions:
> 
>     * what the `-0777` argument to `perl` do?

I have to admit that this is a command I dug up from the last time I used `perl` to apply some complicated regexes. Looking it up, "The -0777 makes it apply the regular expression to the whole thing instead of line by line." I think I used it in this case to match multiline patterns.

lsf37:
> I'm happy to squash the automated ones together, I think I mostly did it this way due to starting off with the easy lemma renames and then gradually working out what was needed to change the variables. I wouldn't mind keeping the manual one separate though, I expect it to help when merging into `rt` or any other ongoing work.

Agreed, let's squash the automated ones and keep the manual fix-up so we can transfer it.

The gitlint check fails because the regular expressions in the commit message are too long. I think I'd be fine overriding the check for this instance. @Xaphiosis do you agree?

lsf37 (approved):

lsf37:
Thanks, one more step towards sanity! :-)

PR #766, opened 2024-06-17
Remove old-style crunch and rename crunches to crunch

corlewis:
This PR rewrites all remaining uses of `crunch` to the newer, preferred, `crunches` syntax, and then removes `crunch` and renames `crunches` to `crunch`.

It probably only makes sense to look at the second and third commits, the other two are large, automated rewrites.


Xaphiosis (approved):

corlewis:
> Did we decide in the end on renaming crunches to crunch? I'm fine either way, but I missed the decision point.

Looking back, we didn't make an explicit decision but we did all say that we were ok with the rename and I think I just went with that. My preference is to rename but I don't mind too much, and it would be easy to change it in this PR since it's a separate commit.

> As in the monads PR, I'm still not sure if it makes sense beyond the PR review to keep the separate fix-up commit after the automated replacement pass. Other than those two issues, I think it looks good.

I'm happy either way, but in my experience the fix-up commit being separate helps when merging with other work and branches. In particular, about merging this with `rt`.

lsf37 (approved):

lsf37:
> > Did we decide in the end on renaming crunches to crunch? I'm fine either way, but I missed the decision point.
> 
> Looking back, we didn't make an explicit decision but we did all say that we were ok with the rename and I think I just went with that. My preference is to rename but I don't mind too much, and it would be easy to change it in this PR since it's a separate commit.

I'm happy with renaming it to `crunch`. Slightly shorter.


> > As in the monads PR, I'm still not sure if it makes sense beyond the PR review to keep the separate fix-up commit after the automated replacement pass. Other than those two issues, I think it looks good.
> 
> I'm happy either way, but in my experience the fix-up commit being separate helps when merging with other work and branches. In particular, I'm thinking about merging this with `rt`.

Yes, I think it does make sense for these ones to keep them separate.

PR #768, opened 2024-06-19
haskell: sync hcrNative with C changes

lsf37:
seL4/seL4#1251 changes the value of HCR_NATIVE which we mirror in the Haskell spec.

Test with: seL4/seL4#1251


lsf37:
Coincidentally, the config `CONFIG_DISABLE_WFI_WFE_TRAPS` is probably something we can easily make generic in the proofs, and probably should (influences the specific value of `HCR_VCPU`).

corlewis (approved):
Making some of this generic would definitely be nice.

Xaphiosis (approved):

andybui01:
Thanks!

PR #769, opened 2024-06-26
Make 'a signed word have a different name than 'a word

simonjwinwood:
```
   "typ_name (typ_uinfo_t TYPE(8 word))  = ''word00010''"
```
and
```
   "typ_name (typ_uinfo_t TYPE(8 sword))  = ''sword00010''"
```

### DIFF
--- a/tools/autocorres/TypHeapSimple.thy
+++ b/tools/autocorres/TypHeapSimple.thy
@@ -807,7 +807,8 @@ lemmas simple_lift_simps =
   simple_lift_field_update_t
   c_guard_array_field
   nat_to_bin_string_simps
-
+  signed_or_unsigned
+  
 (* Old name for the above simpset. *)
 lemmas typ_simple_heap_simps = simple_lift_simps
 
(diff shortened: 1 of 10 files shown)


lsf37:
Can you please provide motivation for the change in the description so the we can figure out in the future why we did this?

The C proofs all seem to be fine, which is a good sign, the errors are all in `AsmRefine`. @mbrcknl might know more about these.

simonjwinwood:
So this is no longer true: https://github.com/seL4/l4v/blob/dcf271a55569616d97eecdfed078a4ec9357f64b/tools/asmrefine/ProveGraphRefine.thy#L192
```
    192 lemma typ_uinfo_t_signed_word:
    193   "typ_uinfo_t TYPE (('a :: len8) signed word) = typ_uinfo_t TYPE ('a word)"
```

lsf37:
This might actually not be needed. It's used to prove the `align_td` lemma underneath, which should still be true, and then the `ptr_inverse_safe` lemma which I have no idea what it is. The definition talks about footprints, though, which should also still be the same. So both should hold.

lsf37:
Closing because it is unfinished and needs more work. This is potentially useful, but also likely to break proofs.

If anyone would like to pick this up again, do feel free to re-open.

PR #771, opened 2024-07-04
docs: update arch-split

Xaphiosis:
Some things were confusing, some changed over time. Hopefully this is an improvement.

Particular notes:
* extend_locale was unused and removed, replaced by the `_AI_2` multiple-locale strategy
* the explicit global requalify_facts strategy was used but not documented

Reviewers:
* probably easier to read https://github.com/Xaphiosis/l4v/blob/arch_split/docs/arch-split.md than only the diff
* note the two FIXMEs, I could use some help there.
* the `_AI_asms` vs `_AI_assms` is wonderfully inconsistent (1597 instances of the former, 506 of the latter): I'd like to pick one and then standardise

### DIFF
--- a/docs/arch-split.md
+++ b/docs/arch-split.md
@@ -1,4 +1,5 @@
 <!--
+     Copyright 2024, Proofcraft Pty Ltd
      Copyright 2021, Data61, CSIRO (ABN 41 687 119 230)
 
      SPDX-License-Identifier: CC-BY-SA-4.0
@@ -6,13 +7,15 @@
 
 # Architecture Split
 
-The problem that architecture splitting ("arch_split") seeks to address is
-separating architecture-generic concepts from architecture-specific concepts in
-the proofs. Ultimately, arch_split aims to find suitable abstractions of
-architecture-specific details, so that we may prove architecture-generic lemmas
-without unfolding architecture-specific definitions, and therefore share
-architecture-generic proofs across multiple architectures. But we're not there
-yet. So far only the `ASpec` and `AInvs` sessions have really been split.
+Initially, the seL4 functional correctness proof existed for only one
+architecture and platform. When porting to other architectures, we realised
+that copying everything and fixing up the copy is not a good approach, since a
+significant proportion of the specifications and proofs are very similar. By
+splitting the specs and proofs into *generic* and *arch-specific* parts, a
+large amount of work duplication can be avoided. The Haskell spec was already
+written in a way that keeps the generic and architecture-specific definitions
+and types separated. The first arch-split was conducted during the abstract
+invariant (AInvs) proof for 32-bit ARM_HYP.
 
 We use (perhaps abuse) the Isabelle locale mechanism, together with some custom
 commands, to place architecture-specific definitions and proofs in namespaces,
@@ -20,14 +23,19 @@ to make it harder to unfold architecture-specific definitions by accident. We
 also have a collection of hacks to keep existing proofs checking while we
 continue to work towards the ultimate goal of proofs on multiple architectures.
 
-This file describes the current state of arch_split, and what still needs to be
-done.
+So far, the `ASpec` and `AInvs` sessions have been arch-split, with some work
+already done on the executable design spec (`ExecSpec`).
+
 
 ## L4V_ARCH
 
-The architecture split is controlled by an environment
-variable, `L4V_ARCH`, which is used in theory imports sections to include
-the appropriate architecture-specific theory for the currently selected
+When running a session, the `L4V_ARCH` environment variable is used to specify
+the target architecture; this affects which seL4 kernel configuration gets
+used, where Isabelle images are stored, as well as the arch-split infrastructure
+via parametrised theory imports.
+
+By using `$L4V_ARCH` in the theory imports section, we can include
+the appropriate architecture-specific theories for the currently selected
 architecture. For example:
 
 - `l4v/spec/abstract/CSpace_A.thy`
@@ -50,6 +58,7 @@ of their fully qualified names, we adopt the convention that
 architecture-specific theory names are prefixed with "Arch". We don't prefix
 with `ARM` or `X64`, because generic theories must be able to import
 architecture-specific theories without naming a particular architecture.
+
 
 ## The Arch locale
 
@@ -74,7 +83,7 @@ end
 ```
 
 All architecture-specific definitions should be placed in the Arch locale, with
-an appropriate global_naming scheme (see below).
+an appropriate `global_naming` scheme (see below).
 
 If you're not familiar with locales, you should read the [locale tutorial]. The
 `Arch` locale has no parameters and no assumptions, since we are merely using it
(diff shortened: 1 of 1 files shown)


docs/arch-split.md
```diff
@@ -244,7 +261,7 @@ available unqualified until the end of the context block. Indeed, in this case,
 the only purpose of the anonymous context block is to limit the scope of this
 `interpretation`.
 
-Note: It is critical to the success of arch_split that we *never* interpret the
+Note: It is critical to the success of arch\_split that we *never* interpret the
```
lsf37, line 264: Should not need to escape underscore in markdown

Xaphiosis: Shows up as an error in my editor. Maybe I should calm down my editor.

docs/arch-split.md
```diff
@@ -408,23 +458,28 @@ We see no reason to redo that previous work, so the above still
...
 
-In this pattern, an existing theory Foo_AI is split into two theories:
+In this pattern, an existing theory `Foo_AI` is split into two theories:
 
-- `Foo_AI` retains the architecture-generic parts, using a locale `Foo_AI`
+- `Foo_AI` retains the architecture-generic parts, using locales `Foo_AI*`
    where necessary to *assume* the existence of the appropriate
    architecture-specific parts.
+   Note: if only one locale is needed, it is named `Foo_AI` by convention,
+   otherwise naming proceeds: `Foo_AI_1`, `Foo_AI_2`, etc.
```
lsf37, line 474: I wonder if the `_1` is going to bite us at some point. It *should* be fine, but if we end up getting theorems whose names end in `_1` etc, then Isabelle's internal indexing into fact lists might get confused (shouldn't, but from time to time some code resurfaces that relies on naming).

Anyway, nothing to do right now, just wanted to point out a potential future hazard. So far, in AInvs, this seems to have worked out fine.

docs/arch-split.md
```diff
@@ -512,12 +570,84 @@ global_interpretation Retype_AI?: Retype_AI
...
+  qed
+
+(* now we get access to GD in the global context *)
+```
+
+FIXME: strictly speaking, there is no requirement that `Foo_AI_2` includes
+`Foo_AI_1`, but in practice we always do so (I guess because we want everything
+that got proved to be instantly available in case we need it in the proof
+chain). This would mean that the same fact gets exported multiple times? Are
+there downsides to this? This also affects whether
```
lsf37, line 647: End of sentence seems to be missing.

lsf37 (approved):
Cool, thanks for updating the docs.

docs/arch-split.md
Xaphiosis: Fixed

docs/arch-split.md
Xaphiosis: I don't think there will be a problem, they will be prefixed with the locale name so `whatever_AI_1.foo`. However if you want to change this, there are precisely three so far: `Deterministic_AI_1, Tcb_AI_1, Finalise_AI_1`, so it'll be easy.

docs/arch-split.md
Xaphiosis: I was going to say that when we do `interpretation Foo_AI_2?: Foo_AI_2` we don't need to prefix everything with Foo_AI_2, we can pick any other prefix, but then I lost my train of thought, because I didn't know if the "same fact exported multiple times" would cause a clobber, and forgot to delete this start of the sentence. Sorry.

docs/arch-split.md
```diff
@@ -254,12 +271,40 @@ In a generic theory, we typically only interpret the Arch locale:
...
+This immediately makes the fact available in the global context. While it is a
+violation of expected conventions and needs to be repeated in every
+arch-specific theory file, there is one important difference:
+* the `.` in `context begin interpretation Arch .` in the middle of AInvs takes 7.5s
+* `requalify_facts` in the global context is nearly instant (even for
+multiple facts).
+
+This disparity will only get worse as the Arch context grows bigger, and
+might indicate the need for some alternative functionality.
+(FIXME: check if this is true)
```
lsf37, line 299: So, it is true that it does get worse: the more lemmas there are in the locale, the longer it takes to enter. There have been some improvements there from upstream, but not dramatically so. However: we already are using the Arch locale in the current unsplit Refine, and splitting should actually reduce the number of lemmas that are inside the locale. On the other hand, we will be entering it more often. So it may go up, or go down, but I don't think it will change dramatically.

docs/arch-split.md
```diff
@@ -512,12 +573,82 @@ global_interpretation Retype_AI?: Retype_AI
...
+  case 1 show ?case by (intro_locales; (unfold_locales; fact Foo_AI_asms)?)
+  qed
+
+(* now we get access to GD in the global context *)
+```
+
+FIXME: strictly speaking, there is no requirement that `Foo_AI_2` includes
+`Foo_AI_1`, but in practice we always do so (I guess because we want everything
+that got proved to be instantly available in case we need it in the proof
+chain). This would mean that the same fact gets exported multiple times?
```
lsf37, line 649: I'm not entirely sure, but I think Isabelle is actually smart about that. We could double check, but the number of lemmas that would be instantiated twice is fairly small, so I would probably stick with a chain like this even if there was duplication.

docs/arch-split.md
lsf37: Good point. Happy to just leave it as is.

docs/arch-split.md
Xaphiosis: Well even if it stays the same, it's still ~7s per invocation vs essentially free. Would it make any sense to find some kind of way to shortcut it, e.g. something that would understand L4V_ARCH and maybe be able to add `_H` or `_A`, so it would be (just throwing out ideas): `requalify_facts_arch foo -> requalify_facts ARM.foo`? Does it seem at all worth it? Would also get rid of the duplication of the current ones that live towards the end of Arch theories.

docs/arch-split.md
Xaphiosis: Yeah, I guess I should check.

- [ ] investigate how many lemmas of same name but different prefix this chain generates

docs/arch-split.md
lsf37: That does seem like a worthwhile idea and can probably be implemented on top of the current `requalify` commands. We could provide the `_A` and `_H` postfixes as an optional parameter if we wanted to. They would be consistent across architectures, they only indicate which abstraction level of the proof the name is from. Let's turn this into an issue and investigate further.

docs/arch-split.md
Xaphiosis: Investigation results: at most one duplicate is generated, so can confirm that Isabelle is smart about it as you said. My initial thought that maybe we can make all the locales export with the same prefix has been demonstrated to be impossible as well, so we can leave the status quo as is, and I should remove this FIXME. Maybe I should resolve it into a note in the doc? Would that be of use to anyone?

docs/arch-split.md
Xaphiosis: Cool. I'll make an issue.

docs/arch-split.md
corlewis: If it can be expressed easily, it would be of use in avoiding us having the same thought and doing the same investigation some time in the future.

corlewis (approved):
Looks good to me, this reads nicer now and should avoid some confusion. I don't have any specific comments that haven't already been mentioned.

docs/arch-split.md
Xaphiosis: https://github.com/seL4/l4v/issues/777

docs/arch-split.md
lsf37: Agreed, adding a note would be nice, I'm sure we'll have the same question again in a few years otherwise.

PR #773, opened 2024-07-05
make proofs generic in maxIRQ and irq bit width

lsf37:
- clear up ancient `irqInvalid` encoding confusion
- extract `maxIRQ` and `irq_len` from kernel config
- use config `maxIRQ` and `irq_len` consistently in specs
- adapt proofs to use the config value and not unfold it. This has different impact on different architectures:
  -  on AARCH64, it was mostly avoiding unfoldings
  -  on ARM and ARM_HYP, there is a lot of cleanup of magic numbers, using the right constants, adjusting types (the irq type had the wrong size, too big -- harmless but weaker), and avoiding unfolding. Correcting the type size required a new invariant on ARM_HYP in CRefine (but not on ARM). On AARCH64 that invariant was initially avoided, because we had changed the bitfield width of VPPIEvent IRQs in CRefine, but it turns out there are AARCH64 platforms with 8 bit IRQ width, so we have it there as well now.
  -  no genericity on RISCV64 and X64, but at least reducing magic numbers a little bit
- Word_Lib + Lib additions for the above

### DIFF
--- a/proof/Makefile
+++ b/proof/Makefile
@@ -47,6 +47,8 @@ CKernel CSpec : c-kernel design-spec
 
 CBaseRefine CRefine SimplExportAndRefine : c-kernel design-spec ASpec-files
 
+InfoFlowCBase InfoFlowC: CRefine
+
 # Preprocess the kernel's source code and bitfield theory files.
 c-kernel: .FORCE
 	cd ../spec && $(ISABELLE_TOOL) env make c-kernel


Xaphiosis (approved):
This should be fine now. Cheers

spec/abstract/X64/ArchInterrupt_A.thy
```diff
@@ -30,4 +30,12 @@ definition arch_mask_irq_signal :: "irq \<Rightarrow> (unit,'z::state_ext) s_mon
 
 end
 
+context begin interpretation Arch .
+
+(* On Arm architectures, maxIRQ is defined in Kernel_Config. On RISCV64 it is defined manually. *)
```
corlewis, line 35:
```suggestion
(* On Arm architectures, maxIRQ is defined in Kernel_Config. On X64 it is defined manually. *)
```

spec/design/skel/AARCH64/ArchInterrupt_H.thy
```diff
@@ -15,6 +15,10 @@ begin
 
 context Arch begin global_naming AARCH64_H
 
+(* Kernel_Config provides a generic numeral, Haskell expects type irq *)
+abbreviation (input) maxIRQ :: irq where
+  "maxIRQ == Kernel_Config.maxIRQ"
```
corlewis, line 20: Why is this abbreviation introduced in both `ArchInterrupt_H` and `Hardware_H`? Are the imports too complicated here?

corlewis (commented):
Sorry for the late comments, but here's a few very minor things that I had found.

There's also a typo in the first commit message, in the second paragraph `were` should be `where`.

spec/design/skel/AARCH64/ArchInterrupt_H.thy
lsf37: This looks like like a left-over to be honest, the one in `ArchInterrupt_H` can probably be removed. Will look into it in a separate PR.

PR #774, opened 2024-07-05
make proofs generic in `config_ARM_PA_SIZE_BITS_40`

lsf37:
This needed only very minor tweaks, and interestingly only involved more unfolding of `Kernel_Config.config_ARM_PA_SIZE_BITS_40_def`.

Most of the unfoldings involved goals where concrete `vmlevel` numbers now overflow earlier (3 = 0 for config_ARM_PA_SIZE_BITS_40, instead of 4 = 0).

Together with the `maxIRQ` pull request #773, this enables the proof to work for the `zynqmp` (`zcu102`) and `bcm2711` (`rpi4`) platforms. See also seL4/seL4#1290.

### DIFF
--- a/proof/crefine/AARCH64/Invoke_C.thy
+++ b/proof/crefine/AARCH64/Invoke_C.thy
@@ -1463,9 +1463,9 @@ lemma getObjectSize_spec:
                          bit_simps objBits_simps' framesize_to_H_def pageBitsForSize_def
                         object_type_to_H_def Kernel_C_defs APIType_capBits_def)
   apply (simp add:nAPIObjects_def)
-  (* FIXME AARCH64 abstraction violation, looks to be not true when config_ARM_PA_SIZE_BITS_40 *)
-  apply (simp add:enum_object_type enum_apiobject_type frameSizeConstants_defs
-                  Kernel_Config.config_ARM_PA_SIZE_BITS_40_def
+  (* True for both cases of config_ARM_PA_SIZE_BITS_40 with corresponding max vm level change *)
+  apply (simp add: enum_object_type enum_apiobject_type frameSizeConstants_defs
+                   Kernel_Config.config_ARM_PA_SIZE_BITS_40_def
                split: if_split)
   apply unat_arith
   done
--- a/proof/crefine/AARCH64/Retype_C.thy
+++ b/proof/crefine/AARCH64/Retype_C.thy
@@ -2142,7 +2142,7 @@ proof (intro impI allI)
     and cover: "range_cover ptr sz (ptBits pt_t) 1"
     and al: "is_aligned ptr (ptBits pt_t)"
     and ptr0: "ptr \<noteq> 0"
-    and sz: "(ptBits pt_t) \<le> sz"
+    and sz: "ptBits pt_t \<le> sz"
     and szb: "sz < word_bits"
     and pal: "pspace_aligned' \<sigma>"
     and pdst: "pspace_distinct' \<sigma>"
@@ -2215,7 +2215,9 @@ proof (intro impI allI)
     using al[simplified bit_simps]
     apply -
     apply (rule is_aligned_c_guard[where n="ptBits pt_t" and m=3])
-    apply (simp_all add:  align_td_array align_of_def bit_simps ptr0 split: if_split)
+    apply (simp_all add:  align_td_array align_of_def bit_simps ptr0 pt_t_def
+                          Kernel_Config.config_ARM_PA_SIZE_BITS_40_def
+                    split: if_splits)
     done
 
   have guard'[unfolded array_len_def]: "\<forall>n < array_len. c_guard (pte_Ptr ptr +\<^sub>p int n)"
@@ -2266,7 +2268,7 @@ proof (intro impI allI)
                           pt_bits_def table_size_def power_add pte_bits_def word_size_bits_def
                           hrs_htd_update ht_rl foldr_upd_app_if [folded data_map_insert_def] rl
                           cvariable_array_ptr_retyps[OF szo]
-                          zero_ranges_ptr_retyps[where p="pt_Ptr ptr", simplified szo])
+                          zero_ranges_ptr_retyps[where p="vs_Ptr ptr", simplified szo])
     apply (subst h_t_valid_ptr_retyps_gen_disjoint, assumption)
      apply (simp add:szo cte_C_size cte_level_bits_def)
      apply (erule disjoint_subset)
@@ -4378,7 +4380,7 @@ lemma getObjectSize_symb:
                         APIType_capBits_def objBits_simps' bit_simps
                    split: if_split)
    apply unat_arith
-  (* FIXME AARCH64 abstraction violation *)
+  (* True in either config with corresponding max vm level change *)
   apply (simp add: Kernel_Config.config_ARM_PA_SIZE_BITS_40_def)
   done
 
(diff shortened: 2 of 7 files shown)


proof/crefine/AARCH64/Retype_C.thy
```diff
@@ -2215,7 +2215,9 @@ proof (intro impI allI)
     using al[simplified bit_simps]
     apply -
     apply (rule is_aligned_c_guard[where n="ptBits pt_t" and m=3])
```
Xaphiosis, line 2217: out of curiosity, what's this 3?

proof/crefine/AARCH64/Retype_C.thy
```diff
@@ -2215,7 +2215,9 @@ proof (intro impI allI)
...
-    apply (simp_all add:  align_td_array align_of_def bit_simps ptr0 split: if_split)
+    apply (simp_all add:  align_td_array align_of_def bit_simps ptr0 pt_t_def
+                          Kernel_Config.config_ARM_PA_SIZE_BITS_40_def
```
Xaphiosis, line 2219: I get a strange feeling when I see these unfoldings without a comment, and without a `solves`. For this one the `solves` is implicit as it's at the end of the proof, but the next one isn't. Thoughts on making these unfoldings better documented?

proof/refine/AARCH64/ArchAcc_R.thy
```diff
@@ -672,7 +674,7 @@ next
      apply (simp add: plus_one_eq_asid_pool vref_for_level_def pt_bits_left_def)
      apply (rule conjI, simp add: max_pt_level_def)
      apply (clarsimp simp: level_defs bit_simps maxPTLevel_def)
-     apply word_eqI_solve
+     apply (cases "config_ARM_PA_SIZE_BITS_40"; clarsimp?; word_eqI_solve)
```
Xaphiosis, line 677: does this need the quotes?

proof/refine/AARCH64/ArchAcc_R.thy
```diff
@@ -828,7 +832,7 @@ next
...
```
Xaphiosis, line 835: quotes?

Xaphiosis (approved):
I approve in principle, but have a strange feeling about the uncommented and sometimes-unguarded unfolds.

proof/refine/AARCH64/ArchAcc_R.thy
lsf37: A single name that also happens to be an outer syntax name doesn't really need it, but I thought that we do generally put `cases` terms in quotes. Probably one of those things where we're not consistent.

proof/crefine/AARCH64/Retype_C.thy
lsf37: It's an old proof, so not quite sure, but I suspect it's the word size of a `pte`

proof/crefine/AARCH64/Retype_C.thy
lsf37: Forcing a `solves` where that is not yet implied is probably a good idea, will look into that. It is a bit of a strange way to get genericity, but there are only two cases which we both test (just not in the same proof).

proof/refine/AARCH64/ArchAcc_R.thy
```diff
@@ -840,11 +844,13 @@ next
...
      apply (drule max_pt_level_enum)
-     apply (auto simp: level_defs split: if_split_asm)[1]
+     (* Unfolding here, because the level_enum only produces a contradiction in one of the branches
+        when PA_40 is set. *)
+     apply (auto simp: level_defs Kernel_Config.config_ARM_PA_SIZE_BITS_40_def split: if_split_asm)[1]
     apply (simp add: bit_simps)
     apply word_eqI
     apply (drule max_pt_level_enum)
-    by (auto split: if_split_asm)
+    by (cases config_ARM_PA_SIZE_BITS_40, auto simp: max_pt_level_def2)
```
lsf37, line 853: Clearly not consistent with the quotes :-)

proof/refine/AARCH64/ArchAcc_R.thy
lsf37: Have removed the quotes for consistency with the other ones -- I don't think we'll be adding any operators here in the future, so no point in having them.

lsf37:
Have added `solves` in the missing cases and added a few more comments for unfoldings.

proof/refine/AARCH64/ArchAcc_R.thy
corlewis: For what it's worth, in the past I've thought that it was more common for us to do single names without quotes. Looking back, I guess that was more for short variable names, which I would definitely want without quotes. I guess there could be an argument that longer names like here might be a different case, but consistency is nice.

proof/crefine/AARCH64/Invoke_C.thy
```diff
@@ -1463,9 +1463,9 @@ lemma getObjectSize_spec:
                          bit_simps objBits_simps' framesize_to_H_def pageBitsForSize_def
                         object_type_to_H_def Kernel_C_defs APIType_capBits_def)
   apply (simp add:nAPIObjects_def)
-  (* FIXME AARCH64 abstraction violation, looks to be not true when config_ARM_PA_SIZE_BITS_40 *)
-  apply (simp add:enum_object_type enum_apiobject_type frameSizeConstants_defs
-                  Kernel_Config.config_ARM_PA_SIZE_BITS_40_def
+  (* True for both case of config_ARM_PA_SIZE_BITS_40 with corresponding max vm level change *)
```
corlewis, line 1466:
```suggestion
  (* True for both cases of config_ARM_PA_SIZE_BITS_40 with corresponding max vm level change *)
```

proof/crefine/AARCH64/Retype_C.thy
```diff
@@ -4378,7 +4380,7 @@ lemma getObjectSize_symb:
                         APIType_capBits_def objBits_simps' bit_simps
                    split: if_split)
    apply unat_arith
-  (* FIXME AARCH64 abstraction violation *)
+  (* True in either configs with corresponding max vm level change *)
```
corlewis, line 4383:
```suggestion
  (* True in either config with corresponding max vm level change *)
```
(Or 'True in either of the configs')

proof/refine/AARCH64/ArchAcc_R.thy
```diff
@@ -684,11 +686,13 @@ next
     apply (simp add: pt_bits_def)
     apply (prop_tac "level_type (nlevel + 1) = NormalPT_T")
      apply (drule max_pt_level_enum)
-     apply (auto simp: level_defs split: if_split_asm)[1]
+     (* Unfolding here, because the level_enum only produces a contradiction in one of the branches
+        when PA_40 is set. *)
```
corlewis, line 690: Not sure about this change, but I think it would be better to say exactly what you're talking about unfolding so it's still obvious when we see this comment in the future without the context of the PR.
```suggestion
     (* Unfolding config_ARM_PA_SIZE_BITS_40_def here, because the level_enum only produces a contradiction in one of the branches
        when PA_40 is set. *)
```

proof/refine/AARCH64/ArchAcc_R.thy
```diff
@@ -840,11 +844,13 @@ next
...
```
corlewis, line 848: If you're making the other comment change.
```suggestion
     (* Unfolding config_ARM_PA_SIZE_BITS_40_def here, because the level_enum only produces a contradiction in one of the branches
        when PA_40 is set. *)
```

corlewis (approved):
Very nice!

PR #780, opened 2024-07-09
github: more authorised token for annotations

lsf37:
`GITHUB_TOKEN` only has read access when triggered from a fork on the pull_request trigger, so we use `pull_request_target` instead. This requires passing in the PR number explicitly to the thylint action.

Requires https://github.com/seL4/ci-actions/pull/354 to be merged before we can test it.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,28 @@
+# Copyright 2024, Proofcraft Pty Ltd
+#
+# SPDX-License-Identifier: BSD-2-Clause
+
+# Theory Linter action
+
+name: Lint
+
+# needs pull_request_target trigger for more authority on GITHUB_TOKEN when PR
+# originates on a fork
+on: [pull_request_target]
+
+jobs:
+  thylint:
+    name: 'Theory Linter'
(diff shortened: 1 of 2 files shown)


.github/workflows/lint.yml
```diff
@@ -0,0 +1,33 @@
...
+# needs pull_request_target trigger for more authority on GITHUB_TOKEN when PR
+# originates on a fork
+on:
+  pull_request_target:
+  push:
+    branches:
+      - 'annotations*'
+
+
+jobs:
```
corlewis, line 18: Is the double newline intentional?

corlewis (commented):
Looks good, but I guess we'll see if it works when the other PR is merged.

.github/workflows/lint.yml
lsf37: not very, let me remove that one :-)

lsf37:
Ok, getting there. #786 has come from my fork and produced an annotation now. Still need to merge https://github.com/seL4/ci-actions/pull/355 first.

corlewis (approved):

PR #782, opened 2024-07-09
Monads: resync trace monad with nondet

corlewis:
Some local changes I had that I forgot to add to https://github.com/seL4/l4v/pull/765. Now that I'm not removing these lemmas from the Nondet Monad, the Trace Monad should have them as well.

### DIFF
--- a/lib/Monads/trace/Trace_More_RG.thy
+++ b/lib/Monads/trace/Trace_More_RG.thy
@@ -138,6 +138,20 @@ lemma assertE_tsp:
   "\<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> assertE Q \<lbrace>G\<rbrace>,\<lbrace>\<lambda>rv s0 s. Q \<and> P s0 s\<rbrace>,\<lbrace>E\<rbrace>"
   by (wpsimp wp: assertE_wp)
 
+lemma select_inv_rg:
+  "\<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> select S \<lbrace>G\<rbrace>,\<lbrace>\<lambda>_. P\<rbrace>"
+  by wpsimp
+
+lemma assert_inv_rg:
+  "\<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> assert Q \<lbrace>G\<rbrace>,\<lbrace>\<lambda>_. P\<rbrace>"
+  unfolding assert_def
+  by (cases Q) wpsimp+
+
+lemma assert_opt_inv_rg:
+  "\<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> assert_opt Q \<lbrace>G\<rbrace>,\<lbrace>\<lambda>_. P\<rbrace>"
+  unfolding assert_opt_def
+  by (cases Q) (wp | simp)+
+
 lemma case_options_weak_twp:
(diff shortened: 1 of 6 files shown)


lib/Monads/trace/Trace_More_RG.thy
```diff
@@ -138,6 +138,24 @@ lemma assertE_tsp:
   "\<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> assertE Q \<lbrace>G\<rbrace>,\<lbrace>\<lambda>rv s0 s. Q \<and> P s0 s\<rbrace>,\<lbrace>E\<rbrace>"
   by (wpsimp wp: assertE_wp)
 
+lemma gets_inv_rg[simp]:
+  "\<lbrace>P\<rbrace>,\<lbrace>R\<rbrace> gets f \<lbrace>G\<rbrace>,\<lbrace>\<lambda>_. P\<rbrace>"
+  by (simp add: gets_def, wp)
```
lsf37, line 143: Do we want to replicate the `simp`/`iff` etc attributes in the trace monad? Might be our chance for getting a cleaner setup on the trace side.

lib/Monads/trace/Trace_RG.thy
```diff
@@ -362,6 +362,26 @@ lemma rg_FalseE[simp]:
   "\<lbrace>\<bottom>\<bottom>\<rbrace>,\<lbrace>R\<rbrace> f \<lbrace>G\<rbrace>,\<lbrace>Q\<rbrace>,\<lbrace>E\<rbrace> = prefix_closed f"
   by (simp add: validI_def validIE_def)
 
+lemma return_inv_rg[iff]:
+  "\<lbrace>Q\<rbrace>,\<lbrace>R\<rbrace> return x \<lbrace>G\<rbrace>,\<lbrace>\<lambda>_. Q\<rbrace>"
```
lsf37, line 366: If we decide to not do the attributes, this is another one

lsf37 (approved):
Looks good, modulo the simp/iff attribute question. I think I'd prefer to try without, but am open to arguments.

lib/Monads/trace/Trace_More_RG.thy
corlewis: Good point, although in that case I'd be tempted to remove these inv rules again. In these cases the standard `wp` rules should be fine, if not better.

Xaphiosis (approved):
Looks good. Not quite following what Gerwin's saying about the attributes, but it sounds like a good idea.

corlewis:
> Looks good. Not quite following what Gerwin's saying about the attributes, but it sounds like a good idea.

Roughly, my interpretation of it is that these lemmas being `[simp]` and `[iff]` is very old and not how we would want it if were doing it all now, since it can lead to unexpected behaviour when simplifying. It's probably not worth fixing in the existing proofs, where we have sometimes accidentally depended on that behaviour, but it probably is worth improving how the trace monad is setup. This does have trade-offs, in particular with there being even more breakages when the existing proofs are moved to the trace monad, but so much is already going to be breaking that it shouldn't make too much of a difference.

corlewis:
Changes since the reviews are that I've removed the inv lemmas pointed out by @lsf37, which I think are unnecessary, and pulled in new lemmas to `Trace_Reader_Option.thy` which were recently added in the nondet monad. There's an argument that the equivalents to those inv lemmas should also be removed from the trace monad's `valid` ruleset in `Trace_VCG.thy`, but so far we have been keeping that ruleset very close to the nondet monad's so I think it would be best to leave exploring that for a later PR.

lsf37:
> Changes since the reviews are that I've removed the inv lemmas pointed out by @lsf37, which I think are unnecessary, and pulled in new lemmas to `Trace_Reader_Option.thy` which were recently added in the nondet monad. There's an argument that the equivalents to those inv lemmas should also be removed from the trace monad's `valid` ruleset in `Trace_VCG.thy`, but so far we have been keeping that ruleset very close to the nondet monad's so I think it would be best to leave exploring that for a later PR.

:+1: Happy with that.

PR #785, opened 2024-07-09
Some rules for `Lib` and `CLib`

michaelmcinerney:
Please note that all new rules for the reader monad are currently unused in my work in progress; I have added them here just in case we would like them. 

I'd be happy to rename any of these, so suggestions are welcome.

And I hope the commit messages are sufficiently informative, and sufficiently specific about where the changes are being made. I'm still not sure what the best approach for just adding lemmas like this is in some cases.

### DIFF
--- a/lib/Corres_UL.thy
+++ b/lib/Corres_UL.thy
@@ -1298,6 +1298,16 @@ lemmas corres_split_noop_rhs2
 
 lemmas corres_split_dc = corres_split[where r'=dc, simplified]
 
+lemma corres_add_noop_rhs:
+  "corres_underlying sr nf nf' r P P' f (do _ \<leftarrow> return (); g od)
+   \<Longrightarrow> corres_underlying sr nf nf' r P P' f g"
+  by simp
+
+lemma corres_add_noop_rhs2:
+  "corres_underlying sr nf nf' r P P' f (do _ \<leftarrow> g; return () od)
+   \<Longrightarrow> corres_underlying sr nf nf' r P P' f g"
+  by simp
+
 lemma isLeft_case_sum:
   "isLeft v \<Longrightarrow> (case v of Inl v' \<Rightarrow> f v' | Inr v' \<Rightarrow> g v') = f (theLeft v)"
   by (clarsimp split: sum.splits)
--- a/lib/HaskellLib_H.thy
+++ b/lib/HaskellLib_H.thy
@@ -548,6 +548,9 @@ lemma gets_the_ohaskell_assert:
   "gets_the (ohaskell_assert P []) = assert P"
   by (clarsimp simp: ohaskell_assert_def split: if_splits)
 
-lemmas omonad_defs = omonad_defs ohaskell_assert_def ohaskell_fail_def
+definition ohaskell_state_assert :: "('s \<Rightarrow> bool) \<Rightarrow> unit list \<Rightarrow> ('s, unit) lookup" where
+  "ohaskell_state_assert P L \<equiv> ostate_assert P"
+
+lemmas omonad_defs = omonad_defs ohaskell_assert_def ohaskell_state_assert_def ohaskell_fail_def
 
 end
--- a/lib/Monads/nondet/Nondet_Reader_Option.thy
+++ b/lib/Monads/nondet/Nondet_Reader_Option.thy
@@ -98,6 +98,11 @@ lemma gets_the_assert:
   "gets_the (oassert P) = assert P"
   by (simp add: oassert_def assert_def gets_the_fail gets_the_return)
 
+lemma gets_the_ostate_assert:
+  "gets_the (ostate_assert P) = state_assert P"
+  by (clarsimp simp: ostate_assert_def state_assert_def gets_the_Some_get gets_the_obind
+                     gets_the_assert)
+
 lemma gets_the_assert_opt:
   "gets_the (oassert_opt P) = assert_opt P"
   by (simp add: oassert_opt_def assert_opt_def gets_the_return gets_the_fail split: option.splits)
--- a/lib/Monads/nondet/Nondet_Reader_Option.thy
+++ b/lib/Monads/nondet/Nondet_Reader_Option.thy
@@ -70,6 +70,10 @@ lemma gets_the_obind:
   "gets_the (f |>> g) = gets_the f >>= (\<lambda>x. gets_the (g x))"
   by (rule ext) (simp add: monad_simps obind_def split: option.splits)
 
+lemma gets_the_Some_get[simp]:
+  "gets_the Some = get"
+  by (clarsimp simp: gets_the_def gets_def assert_opt_Some)
+
 lemma gets_the_return:
   "gets_the (oreturn x) = return x"
   by (simp add: monad_simps oreturn_def)
--- a/lib/Monads/reader_option/Reader_Option_VCG.thy
+++ b/lib/Monads/reader_option/Reader_Option_VCG.thy
@@ -250,6 +250,10 @@ lemma no_ofail_ofail[wp]:
   "no_ofail \<bottom> ofail"
   by (simp add: no_ofail_def)
 
+lemma no_ofail_ask[wp]:
+  "no_ofail \<top> ask"
+  by (simp add: no_ofail_def)
+
 lemma no_ofail_asks_simp[simp]:
   "no_ofail P (asks f)"
   unfolding asks_def oreturn_def obind_def no_ofail_def
--- a/lib/Monads/reader_option/Reader_Option_VCG.thy
+++ b/lib/Monads/reader_option/Reader_Option_VCG.thy
@@ -58,6 +58,19 @@ lemmas owhile_add_inv = owhile_inv_def[symmetric]
 
 
 (* WP rules for ovalid. *)
+
+lemma ovalid_wp_comb1[wp_comb]:
+  "\<lbrakk> ovalid P' f Q; ovalid P f Q'; \<And>s. P s \<Longrightarrow> P' s \<rbrakk> \<Longrightarrow> ovalid P f (\<lambda>r s. Q r s \<and> Q' r s)"
+  by (simp add: ovalid_def)
+
+lemma ovalid_wp_comb2[wp_comb]:
+  "\<lbrakk> ovalid P f Q; \<And>s. P' s \<Longrightarrow> P s \<rbrakk> \<Longrightarrow> ovalid P' f Q"
+  by (auto simp: ovalid_def)
+
+lemma ovalid_wp_comb3[wp_comb]:
+  "\<lbrakk> ovalid P f Q; ovalid P' f Q' \<rbrakk> \<Longrightarrow> ovalid (\<lambda>s. P s \<and> P' s) f (\<lambda>r s. Q r s \<and> Q' r s)"
+  by (auto simp: ovalid_def)
+
 lemma ovalid_post_taut[wp]:
   "\<lblot>P\<rblot> f \<lblot>\<top>\<top>\<rblot>"
   by (simp add: ovalid_def)
@@ -101,6 +114,15 @@ lemma oassert_wp[wp]:
   apply (simp add: oassert_def)
   apply (intro conjI; wpsimp)
   done
+
+lemma ostate_assert_wp:
+  "\<lblot>\<lambda>s. f s \<longrightarrow> P () s\<rblot> ostate_assert f \<lblot>P\<rblot>"
+  unfolding ostate_assert_def
+  by (wpsimp wp: ask_wp)
+
+lemma ostate_assert_sp:
+  "\<lblot>P\<rblot> ostate_assert Q \<lblot>\<lambda>_ s. P s \<and> Q s\<rblot>"
+  by (wpsimp wp: ostate_assert_wp ovalid_wp_comb2)
 
 lemma ogets_wp[wp]:
   "ovalid (\<lambda>s. P (f s) s) (ogets f) P"
@@ -148,19 +170,6 @@ definition ovalid_property where "ovalid_property P x = (\<lambda>s f. (\<forall
 lemma ovalid_is_triple[wp_trip]:
   "ovalid P f Q = triple_judgement P f (ovalid_property Q (\<lambda>s f. f s))"
   by (auto simp: triple_judgement_def ovalid_def ovalid_property_def)
-
-
-lemma ovalid_wp_comb1[wp_comb]:
-  "\<lbrakk> ovalid P' f Q; ovalid P f Q'; \<And>s. P s \<Longrightarrow> P' s \<rbrakk> \<Longrightarrow> ovalid P f (\<lambda>r s. Q r s \<and> Q' r s)"
-  by (simp add: ovalid_def)
-
-lemma ovalid_wp_comb2[wp_comb]:
-  "\<lbrakk> ovalid P f Q; \<And>s. P' s \<Longrightarrow> P s \<rbrakk> \<Longrightarrow> ovalid P' f Q"
-  by (auto simp: ovalid_def)
-
-lemma ovalid_wp_comb3[wp_comb]:
-  "\<lbrakk> ovalid P f Q; ovalid P' f Q' \<rbrakk> \<Longrightarrow> ovalid (\<lambda>s. P s \<and> P' s) f (\<lambda>r s. Q r s \<and> Q' r s)"
-  by (auto simp: ovalid_def)
 
 
 (* WP rules for ovalidNF. *)
@@ -234,6 +243,11 @@ lemma ovalidNF_wp_comb3[wp_comb]:
 
 
 (* FIXME: WP rules for no_ofail, which might not be correct. *)
+
+lemma no_ofail_pre_imp[wp_pre]:
+  "\<lbrakk> no_ofail P f; \<And>s. P' s \<Longrightarrow> P s \<rbrakk> \<Longrightarrow> no_ofail P' f"
+  by (simp add: no_ofail_def)
+
 lemma no_ofailD:
   "\<lbrakk> no_ofail P m; P s \<rbrakk> \<Longrightarrow> \<exists>y. m s = Some y"
   by (simp add: no_ofail_def)
@@ -309,6 +323,11 @@ lemma no_ofail_oassert[simp, wp]:
   "no_ofail (\<lambda>_. P) (oassert P)"
   by (simp add: oassert_def no_ofail_def)
 
+lemma no_ofail_ostate_assert[wp]:
+  "no_ofail P (ostate_assert P)"
+  unfolding ostate_assert_def
+  by wpsimp
+
 lemma no_ofail_is_triple[wp_trip]:
   "no_ofail P f = triple_judgement P f (\<lambda>s f. f s \<noteq> None)"
   by (auto simp: triple_judgement_def no_ofail_def)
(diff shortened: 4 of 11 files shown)


lib/Corres_UL.thy
```diff
@@ -1298,6 +1298,16 @@ lemmas corres_split_noop_rhs2
 
 lemmas corres_split_dc = corres_split[where r'=dc, simplified]
 
+lemma corres_add_noop_rhs:
```
michaelmcinerney, line 1301: Maybe I should use an eta-expanded form here, but these are just moved for DRefine/CRefine, so I wasn't sure whether they should be modified (I did rename the schematics in `corres_add_noop_rhs`, but the rule is not currently being used with manually instantiated schematics)

lib/Monads/nondet/Nondet_Reader_Option.thy
```diff
@@ -70,6 +70,10 @@ lemma gets_the_obind:
   "gets_the (f |>> g) = gets_the f >>= (\<lambda>x. gets_the (g x))"
   by (rule ext) (simp add: monad_simps obind_def split: option.splits)
 
+lemma gets_the_Some_get:
```
michaelmcinerney, line 73: Maybe it would be a bit more "thematic" to state this with `ask`, rather than `Some`. I could then rename this to `gets_the_ask_get`, or simply `gets_the_ask` (I named this `gets_the_Some_get` only because `gets_the_Some` is already taken)

lib/Monads/reader_option/Reader_Option_VCG.thy
```diff
@@ -83,6 +83,10 @@ lemma ofail_wp[wp]:
   "ovalid (\<lambda>_. True) ofail Q"
   by (simp add: ovalid_def ofail_def)
 
+lemma ask_wp:
```
michaelmcinerney, line 86: I'm never the best judge on whether a lemma should be in one of these named theorem sets like `[wp]`, especially in `Lib` rules, where they're sometimes added in particular spots. But I can certainly make this one `[wp]` if that's best

lib/clib/CCorresLemmas.thy
```diff
@@ -586,6 +586,13 @@ lemma ccorres_assert2:
               (\<lambda>s. P \<longrightarrow> G s) {s. P \<longrightarrow> s \<in> G'} hs (assert P >>= f) c"
   by (cases P, simp_all add: ccorres_fail')
 
+(* A weaker version of ccorres_assert2 that may be easier to match with schematics, since
+   the concrete guard does not mention the asserted property *)
+lemma ccorres_assert2':
```
michaelmcinerney, line 591: Maybe this could have a more informative name. I could rename regular `ccorres_assert` to `ccorres_assert_drop_property` or something like that, `ccorres_assert2` to `ccorres_assert_add_property` or something, and then this new rule could maybe be `ccorres_assert_add_abstract_property`. I'm not married to these names, these are just suggestions to indicate what's happening

lib/clib/CCorresLemmas.thy
```diff
@@ -1040,41 +1047,17 @@ lemma ccorres_Guard_True_Seq:
...
-   apply (metis (mono_tags, lifting) HoarePartial.SeqSwap exec_stuck mem_Collect_eq)
-  apply (metis (mono_tags, lifting) HoarePartial.SeqSwap empty_iff exec_fault mem_Collect_eq)
-  done
-
 text \<open>
-  This rule is intended to be used to show correspondence between a Haskell whileLoop and a C
-  while loop, where in particular, the C while loop is parsed into a Simpl While loop that
-  updates the cstate as part of the loop condition. In such a case, the CParser will produce a Simpl
-  program in the form that is seen in the conclusion of this rule.\<close>
+  This rule is intended to be used to show correspondence between a @{term whileLoop} and a Simpl
```
michaelmcinerney, line 1051: I could add in another bullet point here talking about the requirement that the `whileLoop` condition be in the `the blah` form and that this cannot fail, and therefore, we must have the `no_ofail` assumption.

I think @lsf37 and I discussed briefly the idea that we might want an alternate `whileLoop` where the condition is always a reader monad function, and where the new `whileLoop` fails if the reader monad function fails. But that would be a lot of mucking around, especially given all the rules we already have about `whileLoop`

lib/Corres_UL.thy
lsf37: If we go to the trouble of making the PR, let's give it a shot. Doing the eta-expanded form could break some existing proofs (by renaming the variables it produces), but it'll be nicer and more stable going forward.

lsf37: Actually, now that I say this, I don't think we need it here: the rule shouldn't produce a new name, because it's not binding anything (the `_` makes sure the binding is unused).

lib/HaskellLib_H.thy
```diff
@@ -548,6 +548,30 @@ lemma gets_the_ohaskell_assert:
   "gets_the (ohaskell_assert P []) = assert P"
   by (clarsimp simp: ohaskell_assert_def split: if_splits)
 
-lemmas omonad_defs = omonad_defs ohaskell_assert_def ohaskell_fail_def
+definition ostate_assert :: "('s \<Rightarrow> bool) \<Rightarrow> ('s, unit) lookup" where
+  "ostate_assert P \<equiv> do { s \<leftarrow> ask; oassert (P s)}"
```
lsf37, line 552: This definition should be in the monad library, only `ohaskell_state_assert` should live on HaskellLib_H

lib/HaskellLib_H.thy
```diff
@@ -548,6 +548,30 @@ lemma gets_the_ohaskell_assert:
...
+
+definition ohaskell_state_assert :: "('s \<Rightarrow> bool) \<Rightarrow> unit list \<Rightarrow> ('s, unit) lookup" where
+  "ohaskell_state_assert P L \<equiv> ostate_assert P"
+
+lemma gets_the_ostate_assert:
+  "gets_the (ostate_assert P) = stateAssert P []"
+  by (clarsimp simp: ostate_assert_def stateAssert_def state_assert_def gets_the_Some_get)
```
lsf37, line 559: This should not use the Haskell `stateAssert` on the rhs, but the library `state_assert` instead and should move to the Monad library.

(We should also have a rewrite from `stateAssert` to `state_assert` somewhere so that we can get this equality automatically if we need it).

lib/HaskellLib_H.thy
```diff
@@ -548,6 +548,30 @@ lemma gets_the_ohaskell_assert:
...
+lemma ostate_assert_sp:
+  "\<lblot>P\<rblot> ostate_assert Q \<lblot>\<lambda>_ s. P s \<and> Q s\<rblot>"
+  by (wpsimp wp: ostate_assert_wp)
+
+lemma no_ofail_ostate_assert[wp]:
+  "no_ofail P (ostate_assert P)"
+  unfolding ostate_assert_def
+  by (wpsimp wp: ask_wp)
+
+lemmas omonad_defs = omonad_defs ohaskell_assert_def ohaskell_state_assert_def ohaskell_fail_def
```
lsf37, line 575: These should all move to Monad_Lib

lib/Monads/nondet/Nondet_Reader_Option.thy
lsf37: I think I'm fine with `gets_the_Some_get` -- `ask` is mainly there for Haskell compatibility and is an input abbreviation only, i.e. when you see the term in your goal, you're going to see `Some`, not `ask`, and it'll be easier to find if the lemma has `Some` in its name.

lib/Monads/reader_option/Reader_Option_VCG.thy
lsf37: This one would is basic enough to make sense as `[wp]`, yes. @corlewis any concerns there?

lib/clib/CCorresLemmas.thy
lsf37: Hm, this is a bit subtle. The default `corres_assert` doesn't actually drop the property, it does provide it in the subgoal, and it is a bit strange that we do sometimes seem to need it in the generated precondition -- ideally the sub proof should not produce anything that needs `P` any more, because `P` was available already and should have been eliminated.

That said, the way the schematics are instantiated, we often do not want to `simp` or similar in the sub proof, so I agree that we do want these variants. The problem (as you seem to have discovered if you have an abstract-only variant) is that adding implications to the preconditions can make subsequent proofs harder even though they are logically weaker/easier.

Of course the current variant names are terrible. I would propose the following scheme: `ccorres_assert` should stay the default name, and we add `ccorres_assert_both` + `ccorres_assert_abs`. If we ever need it, we could add `ccorres_assert_C` (or `_concrete` or similar).

lib/clib/CCorresLemmas.thy
lsf37: Yes, adding that bullet point would make sense, with a remark that eliminating the `no_ofail` would require a slightly different form of `whileLoop` that can deal with failure in the condition.

lsf37 (commented):
Cool. Some things to move into Monad_Lib still, but otherwise I think this pretty close. It'd be fine to do the `ccorres_assert` renames in a different PR if you want to.

lib/clib/CCorresLemmas.thy
michaelmcinerney: This is a bit tricky to figure out. After I wrote the original comment about add and dropping properties I wasn't so sure why it is that it seems like the property that's asserted disappears sometimes. Are you thinking that it's because it's `simp`ed away? And that having it in the guard explicitly preserves it until the final goal where the guard implications are resolved?

The reason I wanted this variant was that `ccorres_symb_exec_l'` won't change the schematic, and so if we use `ccorres_symb_exec_l'` and then want to use an assert rule about the value that we get, then we can't unify with `ccorres_assert2`.

lib/HaskellLib_H.thy
michaelmcinerney: The annoying thing here is that so many of our `corres/ccorres` rules are stated in terms of `stateAssert`, so if we wanted to use them with `state_assert`, we would currently have to do a whole bunch of `[simplified stateAssert_def]`ing. Would it be worth making any changes to prefer `state_assert`?

proof/crefine/autocorres-test/AutoCorresTest.thy
```diff
@@ -132,11 +132,6 @@ local_setup \<open>AutoCorresModifiesProofs.new_modifies_rules "../c/build/$L4V_
 
 text \<open>Extra corres_underlying rules.\<close>
 
-lemma corres_add_noop_rhs2:
```
Xaphiosis, line 135: you're modifying crefine but I didn't see any `crefine` tag for any of the commits

proof/drefine/Arch_DR.thy
```diff
@@ -1082,11 +1082,6 @@ lemma pde_opt_cap_eq:
   apply (clarsimp simp: valid_idle_def st_tcb_at_def obj_at_def pred_tcb_at_def)
   done
 
-lemma corres_add_noop_rhs:
```
Xaphiosis, line 1085: you're modifying drefine but I didn't see any `drefine` tag for any of the commits

lib/Monads/nondet/Nondet_Reader_Option.thy
Xaphiosis, line 73: should we mark this one as `[simp]`? I can't imagine ever wanting `gets_the Some`

lib/clib/Corres_UL_C.thy
```diff
@@ -288,6 +301,16 @@ lemma ccorres_from_vcg_nofail:
   apply (rule hoare_complete, simp add: HoarePartialDef.valid_def)
   done
 
+lemma ccorres_to_vcg_with_prop':
```
Xaphiosis, line 304: I'd put this below the non-prime version. Usually we put the `'` one above if it's used to show the non-prime one, and this one seems less intuitive than its `no_fail` cousin.

lib/clib/Corres_UL_C.thy
```diff
@@ -137,6 +141,15 @@ lemma ccorresI':
   apply simp
   done
 
+lemma ccorresI_empty_handler_stack:
+  assumes rl:
+    "\<And>s s' xstate.
+      \<lbrakk>(s, s') \<in> srel; G s; s' \<in> G'; \<Gamma> \<turnstile> \<langle>c, Normal s'\<rangle> \<Rightarrow> xstate; \<not> snd (a s)\<rbrakk>
+      \<Longrightarrow> (\<exists>t'. xstate = Normal t' \<and> (\<exists>(r, t) \<in> fst (a s). (t, t') \<in> srel \<and> rrel r (xf t')))"
+  shows "ccorresG srel \<Gamma> rrel xf G G' [] a c"
```
Xaphiosis, line 149: This is strange. I think maybe the name should be tweaked to `ccorres_empty_handler_stackI` so people don't look for a `ccorresI` like the `ccorresG`.
The other thing is, do you have a small comment on when one would use this? I'm staring at it and can't figure it out.

lib/clib/CCorresLemmas.thy
Xaphiosis: This is indeed strange, I feel like we shouldn't need this rule. We generally rewrite stuff on the C side into Haskell side as we symbolically execute forward, meaning that an assertion on the Haskell side might later provide information to clean up the final C part of the proof, so it should be available to both sides, and if it isn't relevant, it shouldn't get in the way.

It's also possible that the `ccorres_symb_exec_l'` isn't in the right form. I'm guessing none of the other `symb_exec_l` rules worked for you?

The property "disappearing" sounds like it's getting simped away, but it persists in the guards.

If you want to keep this rule, I'd suggest changing the name to `ccorres_assert2_abs` and changing the comment to something like `(* This variant only propagates the assertion into the abstract side, to aid with schematic matching on the concrete side. *)`

proof/crefine/autocorres-test/AutoCorresTest.thy
michaelmcinerney: I can make the tag for this commit `lib+drefine+crefine:`. I did have a look through the commit message guide again, and these points weren't explicitly mentioned, but my thinking here was that it's just moved up to `lib`, and to me, the tag indicates where significant changes are made, and which is the highest session that has been changed, so that we know what would need to be rebuilt.

lib/Monads/nondet/Nondet_Reader_Option.thy
michaelmcinerney: Yeah that sounds good to me, if others are happy with it.

proof/drefine/Arch_DR.thy
Xaphiosis: ah, I think I see, there's a commit that says `add ... to ...`, which is confusing. It should at minimum say `move ... to ...`, but I'd prefer adding the tags as well

proof/crefine/autocorres-test/AutoCorresTest.thy
Xaphiosis: See my comment on the other one. it shouldn't be "add" when you're moving. If it's not significant I'd punt it and say `lib+proof: move ... to ...`
@lsf37 can also weigh in, the moving to lib case is never quite clear.

lib/clib/Corres_UL_C.thy
michaelmcinerney: It is a bit confusing having `ccorresG` as an abbreviation and then having `ccorresI` and `ccorresE` (probably as well as others) being intro/elim style rules. I'm happy to change the name, but to me this name makes more sense because it is an intro rule for (ultimately a) `ccorres_underlying` statement, but with some extra restrictions, namely there being an empty handler stack.

As for when to use it, I'm no expert in C semantics or CRefine, so the handler stack is still a bit mysterious to me. Gerwin mentioned on Mattermost that if we have a schematic handler stack, then to prove `ccorres`, we would need to make assumptions about what the handler stack does, probably something like it taking any `Abrupt` state to a `Normal` state. So if we didn't want to do something like that and we still wanted to prove a `ccorres` rule, we would need to show that the `c` function in this rule always returns a `Normal` state, in which case, the handler stack may as well be empty, since it won't be used. Unfolding `ccorres_underlying_def` will leave us with terms involving `exec_handlers` which aren't as nice to deal with as regular `hoarep` statements, and the `exec_handler` statement doesn't mean much with an empty handler stack anyway.

So maybe a comment to help explain how it could be used could say "for proving low-level `ccorresG` statements where the concrete function always returns a `Normal` state, and therefore, the handler stack will not be used, and can be considered empty". But I'm not sure that's more helpful than what someone might glean from the name of the lemma

lib/clib/CCorresLemmas.thy
lsf37: > The reason I wanted this variant was that ccorres_symb_exec_l' won't change the schematic, and so if we use ccorres_symb_exec_l' and then want to use an assert rule about the value that we get, then we can't unify with ccorres_assert2.

Right, that makes sense to me. Together with Raf's proposed comment that should resolve the content at least.

> After I wrote the original comment about add and dropping properties I wasn't so sure why it is that it seems like the property that's asserted disappears sometimes. Are you thinking that it's because it's simped away? 

It's scope: the default rule makes `P` available in the assumption of all sub goals that are "inside" the scope of the assert rule applications (all new subgoals). The assert2 rule exports `P` additionally to the C vcg goal that is on the same level as the assert2 rule application -- the vcg goal already exists when the rule is being applied and the goal is being modified by partially instantiating the schematic. This in turn usually make `P` available in in the big final goal.

Usually that second part shouldn't be necessary, because you should need the `P` only for code that comes after the `assert`, which means in the subgoals of the assert rule. But sometimes that fails, because we do export `P`-dependent conditions in those vcg goals and sometimes also abstract goals.

lib/clib/Corres_UL_C.thy
Xaphiosis: Yes please to this comment. Even if it's a skill issue, I find your explanation and comment helpful.

lib/HaskellLib_H.thy
lsf37: Right, I see where this is coming from. We could add a rule that does `state_assert P = stateAssert P []` and use that with simp before we apply `corres` rules, but that does feel a bit dirty.

I'd prefer having a global `simp` that turns `stateAssert` into `state_assert` and changing the corres rules, but that could potentially be quite a few of them. Or, maybe less invasive: have variants of the stateAssert corres rules (via `lemmas`, that already are in the `state_assert` form). Then at least the application side is fine and can be adapted incrementally.

lib/Monads/nondet/Nondet_Reader_Option.thy
lsf37: yes, let's

lib/HaskellLib_H.thy
michaelmcinerney: Should I change this to `state_assert`, and make an issue for writing all of our `corres/ccorres` rules in terms of `state_assert`?

lsf37: Issue is good. I think I'd prefer having both variants for the `corres` rules, because sometimes you don't want to run `simp` first.

michaelmcinerney: OK I will write up an issue. Or seeing that we never use the string once it's in the design spec, are we really all that interested in using `stateAssert`? Could we get rid of `stateAssert` and just use `state_assert`? I think we could just change the Haskell translator in one or two spots to achieve this, maybe

lib/clib/Corres_UL_C.thy
michaelmcinerney: I was actually able to prove the non-prime one from the prime one, so I've done that, and left the prime version where it is. I'm sure I tried this before, so I'm not sure why it didn't work the first time...

michaelmcinerney:
All comments so far have now been addressed. I've forced pushed due to a rebase, and also to redo the commit message for the commit that moved some lemmas to Lib. I'll do some further renaming of the commits, due to renaming things, when I squash

lib/HaskellLib_H.thy
lsf37: I was thinking about the same thing last night. Yes, we should look at what it would take to make the Haskell translator produce `state_assert` directly. Maybe put both options into the issue. It's possible that the translator change will lead to relatively big changes in the proofs.

michaelmcinerney: Issue here: https://github.com/seL4/l4v/issues/787

proof/crefine/autocorres-test/AutoCorresTest.thy
lsf37: I don't have a strong preference either way

Xaphiosis (approved):

lsf37 (approved):

PR #789, opened 2024-07-11
Several sorries in `IpcCancel_C`

michaelmcinerney:
Please see the commit message


proof/crefine/RISCV64/Finalise_C.thy
```diff
@@ -1454,10 +1459,17 @@ lemma suspend_ccorres:
...
-   apply (wpsimp wp: hoare_vcg_all_lift)
-  by (auto simp: ThreadState_defs)[1]
+   apply (rule hoare_strengthen_post)
+    apply (rule hoare_vcg_conj_lift)
+     apply (rule hoare_vcg_conj_lift)
+      apply (rule hoare_vcg_conj_lift)
+       apply (rule cancelIPC_sch_act_simple)
+      apply (rule cancelIPC_tcb_at'_better[where P=id, simplified, where p=thread])
+     apply (rule cancelIPC_invs')
+    apply (rule cancelIPC_weak_sch_act_wf)
```
lsf37, line 1469: I usually would want to avoid explicit instantiation as was here before, but the previous version does look more stable and automatic. Would adapting the instantiation here be a very big term or otherwise ugly?

michaelmcinerney: I think it might just be adding `weak_sch_act_wf (ksSchedulerAction s) s` to what was there before, but I can't remember exactly. I can investigate

lsf37 (approved):
Cool. Happy to see more sorries melting away and somehow even Refine seems to have a lot of removed lines with this change.

proof/crefine/RISCV64/Finalise_C.thy
michaelmcinerney: Yep, I just needed to add that, and also `sch_act_simple`, and then I could follow the old proof. Updated in the last commit, which I'll squash

lsf37: Cool, thanks, I do like the new version better.

proof/crefine/RISCV64/IpcCancel_C.thy
```diff
@@ -1769,44 +1775,160 @@ lemma tcb_at_1:
...
+
 lemma isSchedulable_ccorres [corres]:
   "ccorres (\<lambda>r r'. r = to_bool r') ret__unsigned_long_'
-  (tcb_at' tcbPtr) (UNIV \<inter> {s. thread_' s = tcb_ptr_to_ctcb_ptr tcbPtr})  []
-  (isSchedulable tcbPtr) (Call isSchedulable_'proc)"
-sorry (* FIXME RT: isSchedulable_ccorres *)
+     (tcb_at' tcbPtr and no_0_obj') \<lbrace>\<acute>thread = tcb_ptr_to_ctcb_ptr tcbPtr\<rbrace> []
+     (isSchedulable tcbPtr) (Call isSchedulable_'proc)"
+  supply Collect_const[simp del]
+         if_split[split del]
```
corlewis, line 1785: Incredibly minor, but do we have a style preference between these?
```suggestion
  supply Collect_const[simp del] if_split[split del]
```

corlewis: Reviewing further, we also sometimes see this alternative
```suggestion
  supply Collect_const[simp del]
  supply if_split[split del]
```

proof/crefine/RISCV64/IpcCancel_C.thy
```diff
@@ -2815,13 +2935,14 @@ sorry (* FIXME RT: reply_remove_ccorres *)
 lemma cancelIPC_ccorres1:
   assumes cteDeleteOne_ccorres:
   "\<And>w slot. ccorres dc xfdc
-   (invs' and cte_wp_at' (\<lambda>ct. w = -1 \<or> cteCap ct = NullCap
+   (invs'
+    and cte_wp_at' (\<lambda>ct. w = -1 \<or> cteCap ct = NullCap
         \<or> (\<forall>cap'. ccap_relation (cteCap ct) cap' \<longrightarrow> cap_get_tag cap' = w)) slot)
```
corlewis, line 2940: This indentation doesn't seem quite right yet, at first glance I was still confused before realising that the final `∨` is inside the `cte_wp_at'`. This looks like a bit of a painful one to do properly without some awkward linebreaks, but at the very least I think the final line should be indented two more.

corlewis (approved):
As Gerwin said, it's very nice to see more sorries disappear, along with `Refine` being made a bit shorter. Thank you for also improving the style in a lot of places with your changes.

proof/crefine/RISCV64/IpcCancel_C.thy
michaelmcinerney: Ah I misread this bit. I can open it up and redo it

proof/crefine/RISCV64/IpcCancel_C.thy
michaelmcinerney: My thoughts here were that we might like to emphasise that we're doing different sorts of things. Sometimes I add a bunch of things to the `simp` set and then if `if_split[split del]` is right alongside, it's hard to see. So I thought maybe we'd like one line per attribute, or whatever the correct terminology is. But I'm happy to go along with whatever

proof/refine/RISCV64/IpcCancel_R.thy
```diff
@@ -3037,7 +3037,7 @@ lemma ct_not_in_ntfnQueue:
 
 lemma sch_act_wf_weak[elim!]:
   "sch_act_wf sa s \<Longrightarrow> weak_sch_act_wf sa s"
-  by (case_tac sa, (fastforce simp: weak_sch_act_wf_def st_tcb_at'_def obj_at'_def)+)
+  by (clarsimp simp: weak_sch_act_wf_def)
```
corlewis, line 3040: Haha, that's a lot nicer than what I suggested!

proof/crefine/RISCV64/IpcCancel_C.thy
corlewis: Hmm, now that you point out the different attributes I think I like your approach

proof/refine/RISCV64/IpcCancel_R.thy
michaelmcinerney: I significantly weakened `weak_sch_act_wf` as part of this PR, so it really didn't need much. I should have looked more closely when I first came across this. 

About the weakening: we've got `valid_sched_action` in AInvs that has lots of information that we can cross if we need. I really just wanted there to be a TCB at the switch to target to make CRefine easier in a few spots. We often definitely do want a TCB somewhere so we can argue about `NULL` properly, together with `no_0_obj'`

PR #791, opened 2024-07-11
[WIP] Opinionated deployment of enhanced Requalify capabilities on AInvs

Xaphiosis:
🦆🦆🦆 Temporarily only dealing with AARCH64; want feedback before applying to other arches. 
(look at the top relevant "[wip]" commits only for now! this is supposed to be on top of the requalify changes #788 that are not yet merged, but since I'm making these PRs from my fork I can't stack them, lesson learned).

This yields a performance improvement (possibly due to `context begin interpretation Arch .` not being parallalisable by Isabelle):
```
AInvs after deploying tweaks:
glow: ( 0:13:50 real,  1:15:34 cpu, 13.13GB)
      ( 0:14:07 real,  1:16:58 cpu, 17.06GB)
pcr:  ( 0:08:30 real,  0:44:48 cpu, 16.97GB)

AInvs before requalify tricks:
glow: ( 0:17:17 real,  1:24:48 cpu, 16.01GB)
      ( 0:17:15 real,  1:25:41 cpu, 17.21GB)
pcr:  ( 0:14:13 real,  1:07:29 cpu, 10.46GB)
```

Here's what I changed:
* global_naming AARCH64 -> arch_global_naming
* try get rid of `interpretation Arch .` (slow) in lieu of `(in Arch)` (faster) or proper requalifying (nearly instant)
* get rid of requalifications that were already requalified, or were global (thanks to new warnings) TODO: some of these will turn out to be broken because they're not actually global/exported on other arches
* put arch_global_naming on same line as `context Arch begin`
* annotate requalifications in Arch theories that can be moved to generic
* put FIXMEs on unusual global_naming practices

Tracking TODOs for this stage:
- [x] change all `FIXME arch_split` to `FIXME arch-split` in entire repo to match github/project labels
- [ ] add `FIXME arch_split` to PR linter to stop people reintroducing it
- [x] add arch_global_naming to the strange/missing `Arch`-only parts and fix fallout
- [x] use feedback from this PR to update `docs/arch-split.md`
- [x] deploy on other arches once approach is confirmed


proof/invariant-abstract/AARCH64/ArchDetype_AI.thy
```diff
@@ -636,8 +636,8 @@ interpretation Detype_AI_2
         Detype_AI_2.intro
         by blast
 
-context begin interpretation Arch .
-lemma delete_objects_invs[wp]:
+(* generic consequence of architecture-specific details *)
+lemma (in Arch) delete_objects_invs[wp]:
   "\<lbrace>(\<lambda>s. \<exists>slot. cte_wp_at ((=) (cap.UntypedCap dev ptr bits f)) slot s
```
lsf37, line 641: This one is less weird than I originally thought. It's in an Arch file and usually would be inside the big `context Arch` block that spans the entire file, but here it needed to be after the `interpretation` part that comes above. So it's mostly a "usual" Arch lemma that is getting exported and requalified.

Looks all good to me, and the new handling of the `wp` attribute is better.

proof/invariant-abstract/AARCH64/ArchRetype_AI.thy
```diff
@@ -274,6 +271,7 @@ locale retype_region_proofs_arch
 
 context retype_region_proofs begin
 
+(* FIXME arch_split: is there any way to optimise this interpretation out? we can't nest contexts *)
 interpretation Arch .
```
lsf37, line 275: This is a bit funky. Either `retype_region_proofs` is supposed to be generic or not. There is `retype_region_proofs_arch` just above, so you'd think `retype_region_proofs` is the generic one, in which case it shouldn't do `interpretation Arch` and the lemmas should instead be inside `retype_region_proofs_arch`. 

Changing it probably breaks something down the line, but it looks like an abstraction violation to me. Definitely good to put a FIXME here.

proof/invariant-abstract/AARCH64/ArchSchedule_AI.thy
```diff
@@ -28,6 +28,7 @@ lemma dmo_mapM_storeWord_0_invs[wp,Schedule_AI_assms]:
    apply wp
   by (simp add: upto.simps word_bits_def)
 
+(* FIXME arch_split: why not arch global naming? this doesn't make sense *)
 global_naming Arch
```
lsf37, line 32: agreed, doesn't make sense to me either. Do we access any of these under `Arch.` later?

proof/invariant-abstract/AARCH64/Machine_AI.thy
```diff
@@ -69,6 +69,7 @@ crunch_ignore (no_irq) (add:
   handleE' handleE handle_elseE forM forM_x
   zipWithM ignore_failure)
 
+(* FIXME arch_split: no global_naming? *)
 context Arch begin
```
lsf37, line 73: Looks like a potential mistake. Worth trying out what breaks if we change it to `arch_global_naming`. Would also standardise the `requalify` at the end of the file.

proof/invariant-abstract/BCorres_AI.thy
```diff
@@ -16,6 +16,16 @@ abbreviation "bcorres \<equiv> bcorres_underlying truncate_state"
 
 abbreviation "s_bcorres \<equiv> s_bcorres_underlying truncate_state"
 
+context Arch begin arch_global_naming
+
+crunch arch_post_cap_deletion
+  for (bcorres) bcorres[wp]: truncate_state
+
+end
```
lsf37, line 24: Kinda neat that we can do this now in generic files, but also a little bit dangerous. Then again, it does remove duplication.

proof/invariant-abstract/AARCH64/ArchDetype_AI.thy
Xaphiosis: The issue here is that the qualified name of `delete_objects_invs` becomes `Arch.delete_objects_invs`, instead of what we normally would do via global_prefix (`AARCH64.delete_objects_invs`). In this case that should be ok, because it's immediately requalified into the theory context (and also tagged `[wp]` for that context), but it's still a bit of a finicky policy to try canonise.

proof/invariant-abstract/BCorres_AI.thy
Xaphiosis: It is a travesty, I agree. However, what we had before (`context begin interpretation Arch .`) was just as bad, and a lot slower :/

proof/invariant-abstract/CSpaceInvPre_AI.thy
```diff
@@ -141,8 +136,9 @@ lemma empty_table_caps_of:
   "empty_table S ko \<Longrightarrow> caps_of ko = {}"
   by (cases ko, simp_all add: empty_table_def caps_of_def cap_of_def)
 
-context begin interpretation Arch .
-lemma free_index_update_test_function_stuff[simp]:
+(* FIXME arch_split: exports properties of functions that are not necessarily in global context,
+   and then they get placed in the global simpset *)
+lemma (in Arch) free_index_update_test_function_stuff[simp]:
```
lsf37, line 141: This is the slightly stranger case. We probably should avoid this kind of Arch lemma inside generic files. Then again, it has the same kind of remove-duplication benefit as the `crunch (bcorres)` we had elsewhere, so maybe we don't need to be too strict.

Thumbs up to making the name available so it can be removed from `[simp]` again if needed.

proof/invariant-abstract/CSpace_AI.thy
```diff
@@ -12,25 +12,28 @@ theory CSpace_AI
...
-context begin interpretation Arch .
-
-requalify_consts
+arch_requalify_consts
   irq_state_update
   irq_state
   final_matters_arch
   ups_of_heap
 
+(* FIXME arch_split: this should maybe have a global_naming *)
```
lsf37, line 21: agreed, it should

proof/invariant-abstract/DetSchedAux_AI.thy
```diff
@@ -8,10 +8,9 @@ theory DetSchedAux_AI
 imports DetSchedInvs_AI
 begin
 
-context begin interpretation Arch .
+(* (* FIXME arch_split: check it's global on other arches *)
 requalify_facts
-  invoke_untyped_st_tcb_at
-end
+  invoke_untyped_st_tcb_at *)
```
lsf37, line 13: This commented out now, is this getting requalified elsewhere now or was it not used?

proof/invariant-abstract/InvariantsPre_AI.thy
```diff
@@ -8,15 +8,12 @@ theory InvariantsPre_AI
...
 
-context begin interpretation Arch .
-
-requalify_types
-  aa_type
-
-requalify_consts
+(* FIXME RAF: these already appear exported to global context?!
+arch_requalify_types (A)
   aa_type
```
lsf37, line 13: Entirely possible that we have a bunch of duplication for some of these, because we didn't have warning before and would not have noticed without explicitly checking for it, e.g. after things got moved around at some point.

proof/invariant-abstract/DetSchedAux_AI.thy
Xaphiosis: it was already available in theory context on AARCH64, but I'll need to confirm that's also the case for other arches when I re-deploy there

proof/invariant-abstract/BCorres_AI.thy
lsf37: This is definitely nicer.

proof/invariant-abstract/AARCH64/ArchDetype_AI.thy
lsf37: Got it. We don't seem to have many of these, and for exceptional cases I think this is Ok.

lsf37 (commented):
This is pretty neat, and I really like the speed-up. I agree with the opinionated choices in here, and green light from my side to deploy it to the other architectures.

I don't think the missing `global_naming` parts were intentional. Most likely just forgotten and things happened to work anyway. I would add `arch_global_naming`.

Xaphiosis:
> This is pretty neat, and I really like the speed-up. I agree with the opinionated choices in here, and green light from my side to deploy it to the other architectures.
> 
> I don't think the missing `global_naming` parts were intentional. Most likely just forgotten and things happened to work anyway. I would add `arch_global_naming`.

Yay. Glad to hear it. Added the missing `global_naming` parts to TODO list.

proof/invariant-abstract/AARCH64/ArchBCorres_AI.thy
```diff
@@ -45,7 +45,7 @@ crunch prepare_thread_delete
 
 end
 
-requalify_facts AARCH64.arch_finalise_cap_bcorres AARCH64.prepare_thread_delete_bcorres
+arch_requalify_facts arch_finalise_cap_bcorres prepare_thread_delete_bcorres
 
 declare arch_finalise_cap_bcorres[wp] prepare_thread_delete_bcorres[wp]
```
corlewis, line 50: It's not clear to me what you're using to decide when requalifications in Arch files can be moved to generic ones. For instance, why not here?

proof/invariant-abstract/AARCH64/ArchDetype_AI.thy
corlewis: Would the 'canonical' approach be to export enough facts so that this can be proved in a generic file outside of `Arch`?

proof/invariant-abstract/AARCH64/ArchSchedule_AI.thy
corlewis: Agreed, very weird

proof/invariant-abstract/AARCH64/ArchVCPU_AI.thy
```diff
@@ -9,7 +9,7 @@ theory ArchVCPU_AI
 imports AInvs
 begin
 
-context Arch begin global_naming AARCH64 (*FIXME: arch_split*)
+context Arch begin arch_global_naming (*FIXME: arch_split*)
```
corlewis, line 12: Not relevant to this PR, but is the FIXME suggesting that some of this file could be generic? Because that seems unlikely, unless we come up with a new mechanism to only be generic between certain architectures.

proof/invariant-abstract/ADT_AI.thy
```diff
@@ -11,17 +11,13 @@ imports
...
 
-context begin interpretation Arch .
-
-requalify_consts
+arch_requalify_consts (A)
   empty_context
   init_A_st
+
+arch_requalify_consts
   ptable_lift
```
corlewis, line 19: Hmm, I hadn't considered that the new approach of not doing the interpretation means that when requalifying we need to be more specific about the naming. I guess this would fix itself if we were consistent, which the new `arch_global_naming` helps with. Either way, it's still a trade-off that seems worth making.

proof/invariant-abstract/BCorres_AI.thy
corlewis: I don't think it's worth doing with the BCorres files already being a bit of a mess of dependencies, but would our current preference be to assume this in a locale and move the proof to the arch file?

proof/invariant-abstract/CSpaceInv_AI.thy
```diff
@@ -12,31 +12,31 @@ theory CSpaceInv_AI
...
-requalify_consts
+arch_requalify_consts
   cap_master_arch_cap
   replaceable_final_arch_cap
   replaceable_non_final_arch_cap
   unique_table_refs
 
-requalify_facts
+(* There are multiple arch-dependent acap_rights_update_id, one for wellformed_acap,
+   one for valid_arch_cap. Prefer the latter. *)
```
corlewis, line 22: Not as part of this PR, but this sounds like something we should fix.

corlewis (commented):
I like this, the improved consistency and performance is definitely very nice! I generally agree with all of the existing FIXMEs and comments.

Currently, my remaining questions are around when requalifications need to be in Arch files, and when we're ok with Arch proofs in generic files. For the latter, it sort of seems like a sign that we've gotten the abstraction wrong, and should be exporting whatever else is needed from Arch so that the proof could be properly generic.

proof/invariant-abstract/CSpaceInv_AI.thy
Xaphiosis: I'll take a look, you're right, something is off here.

proof/invariant-abstract/ADT_AI.thy
Xaphiosis: It is a bit annoying. I thought about ways to work around it, but then it turns out that I like the aesthetic of knowing whether something came from ASpec/ExecSpec or the proofs. With a bunch of work we could make the command be able to take the `(A/H)` on a per-id level, e.g. `arch_requalify_consts empty_context (A) ptable_lift init_A_st (A)` or something, but I'm not sure about how beneficial that would be.

proof/invariant-abstract/AARCH64/ArchVCPU_AI.thy
Xaphiosis: I honestly have no idea. My guess is that it's an indication that a whole bunch of things were shoved in without looking too deeply... but I'm not seeing any generic stuff that should be reconsidered or exported either. I think removing the FIXME is the best idea here. Will do that for next iteration.

proof/invariant-abstract/BCorres_AI.thy
Xaphiosis: Yes, absolutely. I think this is an example of "no one likes BCorres, so it doesn't get the nice new toys". It got the worst approach. Then the next approach was the "Pre" thing, which also wasn't great, and then someone discovered the locale assumptions approach which I think is the one we should go with in any new proofs, barring maybe esoteric moments in `Invariants`-like files.

proof/invariant-abstract/AARCH64/ArchBCorres_AI.thy
Xaphiosis: This can definitely be moved to the theory that includes this one, I probably only missed it because it's BCorres and was super early in the run-through. If you see others, let me know. Will add FIXME.

proof/invariant-abstract/CSpaceInvPre_AI.thy
Xaphiosis: I think this is a result of the early approaches and "Pre" tech, I think we would try to avoid this in new files.

proof/invariant-abstract/AARCH64/ArchRetype_AI.thy
Xaphiosis: This one is beyond me, I think I'll leave it as a FIXME.

proof/invariant-abstract/AARCH64/ArchDetype_AI.thy
Xaphiosis: Yeah, I think exporting enough facts would be the right idea. I think someone took a shortcut here, but I'm not clear on why exactly.

proof/invariant-abstract/AARCH64/ArchBCorres_AI.thy
corlewis: I'll add comments for the others I saw, although I haven't stopped to think about whether they actually make sense in the Arch file or can be moved.

proof/invariant-abstract/AARCH64/ArchRetype_AI.thy
```diff
@@ -125,10 +125,7 @@ declare post_retype_invs_check_def[simp]
 
 end
 
-
-context begin interpretation Arch .
-requalify_consts post_retype_invs_check
-end
+arch_requalify_consts post_retype_invs_check
```
corlewis, line 128: Requalification in an Arch file

proof/invariant-abstract/AARCH64/ArchRetype_AI.thy
```diff
@@ -683,9 +681,7 @@ lemma cap_range_respects_device_region_cong[cong]:
   by (clarsimp simp: cap_range_respects_device_region_def)
 
 
-context begin interpretation Arch .
-requalify_consts region_in_kernel_window
-end
+arch_requalify_consts region_in_kernel_window
```
corlewis, line 684: Requalification in an Arch file

proof/invariant-abstract/AARCH64/ArchFinalise_AI.thy
```diff
@@ -1900,9 +1898,7 @@ lemma (* dmo_replaceable_or_arch_update *) [Finalise_AI_assms,wp]:
 
 end
 
-context begin interpretation Arch .
-requalify_consts replaceable_or_arch_update
-end
+arch_requalify_consts replaceable_or_arch_update
```
corlewis, line 1901: Requalification in an Arch file

proof/invariant-abstract/AARCH64/ArchCSpaceInv_AI.thy
```diff
@@ -210,8 +210,6 @@ lemmas cap_vptr_simps [simp] =
 
 end
 
-context begin interpretation Arch .
-requalify_facts replace_cap_invs
-end
+arch_requalify_facts replace_cap_invs
```
corlewis, line 213: Requalification in an Arch file

proof/invariant-abstract/ADT_AI.thy
corlewis: Ah right, I hadn't stopped to realise that the `A/H` is only on things from the specs. That's not as bad as I first thought then, I was thinking it was due to us not being consistent with naming.

corlewis: Still a bit annoying but I agree, probably not worth making the command more complicated to handle it.

proof/invariant-abstract/AARCH64/ArchSchedule_AI.thy
Xaphiosis: fixed

proof/invariant-abstract/AARCH64/Machine_AI.thy
Xaphiosis: fixed

proof/invariant-abstract/CSpace_AI.thy
Xaphiosis: fixed

proof/invariant-abstract/InvariantsPre_AI.thy
Xaphiosis: These were already exported in ASpec it turns out.

proof/invariant-abstract/AARCH64/ArchFinalise_AI.thy
Xaphiosis: Unavoidable: const introduced in Arch locale that needs to be referenced generically in this theory.

proof/invariant-abstract/AARCH64/ArchCSpaceInv_AI.thy
Xaphiosis: done

proof/invariant-abstract/AARCH64/ArchRetype_AI.thy
Xaphiosis: Unavoidable: const introduced in Arch locale that needs to be referenced generically in this theory.

proof/invariant-abstract/AARCH64/ArchRetype_AI.thy
Xaphiosis: Unavoidable: const introduced in Arch locale that needs to be referenced generically in this theory.

proof/invariant-abstract/CSpaceInv_AI.thy
Xaphiosis: fixed and added to PR

proof/invariant-abstract/AARCH64/ArchBCorres_AI.thy
Xaphiosis: moved

Xaphiosis:
I think I addressed everything I could for the AARCH64 test run of rolling out these strategies. Next steps will be documentation update, then deployment to other arches, possibly involving other PRs that are cross-platform.

Xaphiosis:
Closed, full deployment now at: https://github.com/seL4/l4v/pull/805

PR #794, opened 2024-07-17
Merge master into rt

corlewis:


corlewis:
This is currently done as a series of partial merges, so that it was easier to correctly handle the large crunch and monad changes. If we prefer a tidier git history I could redo it now as a single merge.

lsf37:
Happy with multiple merges, that is actually nicer, I think.

There does seem to be some fallout, though, only ARM and RISCV should come through in CI for the `rt` branch.

corlewis:
> There does seem to be some fallout, though, only ARM and RISCV should come through in CI for the `rt` branch.

That's because I accidentally targeted master instead of the rt branch. I've fixed that but it's left the tests in a confused state. Should I close this PR and create a new one?

The ARM and RISCV test failures do seem to be correct, so I'm investigating those. I'm confused by the ARM failure though, I wouldn't have expected any merge conflicts in `ArchKernelInit_AI.thy` or with the `irq` type.

lsf37:
> > There does seem to be some fallout, though, only ARM and RISCV should come through in CI for the `rt` branch.
> 
> That's because I accidentally targeted master instead of the rt branch. I've fixed that but it's left the tests in a confused state. Should I close this PR and create a new one?

That's probably easiest, yes.
 
> The ARM and RISCV test failures do seem to be correct, so I'm investigating those. I'm confused by the ARM failure though, I wouldn't have expected any merge conflicts in `ArchKernelInit_AI.thy` or with the `irq` type.

It's possible that MCS mentions the `irq` type in a few more or different places, for which we don't have a spec/proof update  yet. So I could imagine some conflict there. Happy to help with that one if it turns out to be complicated.

PR #795, opened 2024-07-17
Merge master into rt

corlewis:
This PR is a reopened copy of https://github.com/seL4/l4v/pull/794, so that the tests are in a more sensible state.


corlewis:
Relevant comments from the previous version of this PR are

> >This is currently done as a series of partial merges, so that it was easier to correctly handle the large crunch and monad changes. If we prefer a tidier git history I could redo it now as a single merge.
>
> Happy with multiple merges, that is actually nicer, I think.

> > The ARM and RISCV test failures do seem to be correct, so I'm investigating those. I'm confused by the ARM failure though, I wouldn't have expected any merge conflicts in `ArchKernelInit_AI.thy` or with the `irq` type.
> 
> It's possible that MCS mentions the `irq` type in a few more or different places, for which we don't have a spec/proof update yet. So I could imagine some conflict there. Happy to help with that one if it turns out to be complicated.

lsf37:
The gitlint compaint is about the current squash commit plus the ones with the long regexps, so fine to ignore.

lsf37 (approved):
Happy with the merge and with re-running the crunch/monad cleanup scripts.

corlewis:
> The gitlint compaint is about the current squash commit plus the ones with the long regexps, so fine to ignore.

I've changed that squash commit to be a normal one. Having had more time to think about and look at the history, it doesn't really make sense to squash it into one of the merges. It changes lines of proof that are new in the rt branch.

PR #796, opened 2024-07-17
Prove `awaken_ccorres`

michaelmcinerney:


proof/crefine/RISCV64/Finalise_C.thy
```diff
@@ -637,6 +637,14 @@ lemma merge_overlapping_head_refill_ccorres:
   done
 
 (* FIXME RT: move to Corres_UL_C? *)
+lemma ccorres_to_vcg_Normal':
+  "\<lbrakk>ccorres_underlying srel \<Gamma> rrel xf arrel axf P P' [] a c\<rbrakk>
+   \<Longrightarrow> \<Gamma> \<turnstile> {s'. P s \<and> \<not> snd (a s) \<and> s' \<in> P' \<and> (s, s') \<in> srel} c UNIV"
```
lsf37, line 642: If this is a `'` variant, can the other one be derived from it (or vice versa)?

michaelmcinerney: I think I did try, and it didn't work, but I will give it another shot

proof/crefine/RISCV64/SR_lemmas_C.thy
```diff
@@ -2237,6 +2237,12 @@ lemma rf_sr_ctcb_queue_relation:
   apply (clarsimp simp: Let_def seL4_MinPrio_def minDom_def maxDom_to_H maxPrio_to_H)
   done
 
+lemma rf_sr_ctcb_queue_relation_release_queue:
+  "(s, s') \<in> rf_sr \<Longrightarrow> ctcb_queue_relation (ksReleaseQueue s) (ksReleaseQueue_' (globals s'))"
+  unfolding rf_sr_def cstate_relation_def cready_queues_relation_def
+  apply (clarsimp simp: Let_def seL4_MinPrio_def minDom_def maxDom_to_H maxPrio_to_H)
+  done
```
Xaphiosis, line 2244: one-line proof, can use `by`

Xaphiosis:
I think the approach and proofs look ok. I'm as always a bit worried about the tendency to not have any comments in the proofs to help separate things or guide a potential reader/victim.

For when this isn't a draft: the commit tags aren't right btw (no rt, RISCV64 only, etc)

proof/crefine/RISCV64/Finalise_C.thy
michaelmcinerney: I was able to derive one from the other. Updated. Thanks

michaelmcinerney:
Now (finally) ready for review. 

I have added `rt` to the commit message, but I wasn't sure what to do about adding `riscv`. We agreed that we wouldn't add `riscv` to the `crefine` tag, since RISCV is the only arch we're doing for MCS at the moment. I have still been adding `riscv` to `refine`, since we at one point had both ARM and RISCV working for Refine, but we're now just back to RISCV, and I wasn't sure how to add `riscv refine` without then mentioning the other half a dozen sessions I'm changing here...

Xaphiosis (approved):
I'm fine with this one. You're right about the commit message, rt does seem to have different standards and in this case I have no idea how they would hang together usefully.

lsf37 (approved):

PR #797, opened 2024-07-18
bump: make mcs-devel.xml available in bump script

lsf37:
bump-ver-manifest now takes an optional parameter `--mcs` to operate on `mcs-devel.xml` instead of `devel.xml`.

Goes together with https://github.com/seL4/verification-manifest/pull/20

### DIFF
--- a/misc/bump/bump-ver-manifest
+++ b/misc/bump/bump-ver-manifest
@@ -5,6 +5,11 @@
 # SPDX-License-Identifier: BSD-2-Clause
 #
 # execute this in the root repo folder
+#
+# usage:
+#
+# bump-ver-manifest          to bump seL4 revision in devel.xml
+# bump-ver-manifest --mcs    to bump seL4 revision in mcs-devel.xml
 
 SCRIPT_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
 SEL4_DIR=seL4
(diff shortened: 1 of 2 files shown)


misc/bump/bump-ver-manifest
```diff
@@ -33,12 +53,12 @@ then
...
 echo "Generating new manifest..."
-python3 "$SCRIPT_DIR/ver-bump.py" -m "$MANIFEST_DIR/devel.xml" -r "$latest_hash"
+python3 "$SCRIPT_DIR/ver-bump.py" -m "$MANIFEST" -r "$latest_hash"
 
 echo
 echo "Committing new manifest..."
-git -C "$MANIFEST_DIR" add devel.xml
-git -C "$MANIFEST_DIR" commit -m "Bump kernel revision to '$latest'"
+git -C "$MANIFEST_DIR" add "$MANIFEST_FILE"
+git -C "$MANIFEST_DIR" commit -m "$TAG bump kernel revision to '$latest'"
```
corlewis, line 61: You've already included the space in $TAG, which I think is the right thing since you don't want it for non-mcs.
```suggestion
git -C "$MANIFEST_DIR" commit -m "$TAGbump kernel revision to '$latest'"
```

corlewis (approved):

misc/bump/bump-ver-manifest
lsf37: Good catch. I think it'll need `${TAG}bump` -- will do that.

PR #798, opened 2024-07-18
github: remove rt branch test; switch to mcs-devel.xml

lsf37:
- switch off the rt branch test -- to be added to the deployment workflow later (via merge from master)
- switch the PR test to the mcs-devel.xml manifest which is taking the role devel.xml has on the master branch

This will still use the fixed seL4 revision the mcs-devel manifest, but that revision can now be bumped with the `bump-ver-manifest` script, and in the future will also receive pre-process equivalent updates.

### DIFF
--- a/.github/workflows/proof.yml
+++ b/.github/workflows/proof.yml
@@ -9,8 +9,6 @@ on:
     paths-ignore:
       - '**.md'
       - '**.txt'
-    branches:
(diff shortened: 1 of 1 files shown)


corlewis (approved):

lsf37:
(rebased)

PR #799, opened 2024-07-19
github: add mcs manifest deployment

lsf37:
Add deployment of the mcs-devel.xml manifest to mcs.xml.

The pattern is to use the branch name in `github.ref` as a guard in the initial code freeze action, and then make any subsequent actions depend on either `code` or `mcs-code`, respectively.

All of this lives on the master branch (but does trigger on pushes to `rt`), so that it can in the future also trigger on manifest updates from `repository_dispatch` events.

Currently the latter does not happen yet, because `github.ref` will be `refs/head/master` for `repository_dispatch` events, so only the main deployment job runs for those.

### DIFF
--- a/.github/workflows/proof-deploy.yml
+++ b/.github/workflows/proof-deploy.yml
@@ -10,6 +10,7 @@ on:
   push:
     branches:
       - master
+      - rt
   repository_dispatch:
     types:
       - manifest-update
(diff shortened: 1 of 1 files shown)


lsf37:
I'm reasonably certain this won't break manifest deployment for the master branch, but the first time this will really be properly tested for the rt branch is when we merge this there.

corlewis (approved):

PR #800, opened 2024-07-23
merge master into `rt`

lsf37:
This brings in the mcs manifest deployment and mcs bump script option from master.


corlewis (approved):
Should the last commit here have a `rt` tag? It probably won't matter in this case but those tags do help in the future when merging or looking at the history for whatever reason.

lsf37:
You're right, will add one.

PR #801, opened 2024-07-24
Update arch-split doc; rename arch_split->arch-split

Xaphiosis:
There's a noisy commit for the global rename, which isn't interesting beyond the commit message.
The main thing I'd like a review of is the updates to `docs/arch-split.md` (separate commit) to see whether things make sense. The changes are based off of the test deployment work I did for AARCH64 AInvs.

### DIFF
--- a/docs/arch-split.md
+++ b/docs/arch-split.md
@@ -111,7 +111,7 @@ theory Retype_R
 imports VSpace_R
 begin
 
-context begin interpretation Arch . (*FIXME: arch_split*)
+context begin interpretation Arch . (*FIXME: arch-split*)
 
 lemma placeNewObject_def2:
  "placeNewObject ptr val gb = createObjects' ptr 1 (injectKO val) gb"
@@ -199,7 +199,7 @@ architecture. If we saw such a reference in a generic theory, we would
 immediately recognise that something was wrong.
 
 The convention is that in architecture-specific theories, we initially
-give *all* types, constants and lemmas with the architecture-specific
+give *all* types, constants and lemmas the architecture-specific
 `arch_global_naming` scheme. Then, in generic theories, we use
 *requalification* to selectively extract just those types, constants and
 facts which are expected to exist on all architectures.
@@ -339,7 +339,7 @@ available unqualified until the end of the context block. Indeed, in this case,
 the only purpose of the anonymous context block is to limit the scope of this
 `interpretation`.
 
-Note: It is critical to the success of arch_split that we *never* interpret the
+Note: It is critical to the success of arch-split that we *never* interpret the
 Arch locale, *except* inside an appropriate context block.
 
 In a generic theory, we typically only interpret the Arch locale to keep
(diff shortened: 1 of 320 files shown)


docs/arch-split.md
```diff
@@ -687,6 +746,39 @@ generates limited duplication: a fact from `Foo_AI_1` will be duplicated in
...
+requalify_facts Arch.valid_arch_cap_pspaceI
+```
+
+In this case, no matter what the architecture, the `valid_arch_cap` function
+will only ever look at the heap, so this proof will always work.
+
+There are some considerations when using this strategy:
+
+1. We use the Arch locale without a `global_naming`, as its performance better
+   than entering the Arch locale and proving the lemma there. This means its
```
lsf37, line 774:
```suggestion
1. We use the Arch locale without `global_naming`, as its performance is better
   than entering the Arch locale and proving the lemma there. This means its
```
```suggestion
1. We use the Arch locale without a `global_naming`, as its performance better
   than entering the Arch locale and proving the lemma there. This means its
```

lsf37 (approved):
:+1: Nice how the docs are getting more and more coherent.

corlewis (approved):

docs/arch-split.md
```diff
@@ -220,11 +232,73 @@ done for any type, constant or fact:
...
+(* However, its definition is not. *)
+thm empty_context_def (* ERROR *)
+```
+
+In some cases, consts/types/facts may be thrown into the `Arch` context without
+further qualification. In such cases, normal requalification may be used:
+
+```isabelle
+requalify_consts Arch.empty_context (* standard locale version, likely due to missing global_naming *)
+```
```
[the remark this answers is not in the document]
Xaphiosis: Possibly, because we do sometimes do `lemma (in Arch) foo` with an immediate `requalify_facts Arch.foo`, so it's worth knowing about at least. I think I explain that case in the docs now.

docs/arch-split.md
Xaphiosis: Did something go wrong on the second one? It says "The suggestion is invalid because no changes were made to the code"

lsf37: Nothing wrong on that line, I was just trying to provide sentence context. Did not know that this would lead to failure, sorry.

Xaphiosis: Thanks, fixed

docs/arch-split.md
```diff
@@ -192,8 +199,8 @@ architecture. If we saw such a reference in a generic theory, we would
 immediately recognise that something was wrong.
 
 The convention is that in architecture-specific theories, we initially
-give *all* types, constants and lemmas with an architecture-specific
-`global_naming` scheme. Then, in generic theories, we use
+give *all* types, constants and lemmas with the architecture-specific
+`arch_global_naming` scheme. Then, in generic theories, we use
```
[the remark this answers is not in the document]
Xaphiosis: Thanks, fixed

PR #803, opened 2024-07-26
github: factor out platform rebase into own workflow

lsf37:
Factor out the platform branch rebase job into its own workflow which is triggered explicitly by a repository-dispatch event.

We can then re-trigger the rebase job after the proofs have succeeded on the rebased branch to confirm that everything is now up to date and get a successful rebase test run after everything has completed.


Xaphiosis (approved):
seeing a merge commit in commits

lsf37:
> seeing a merge commit in commits

that's just the branch update from the GitHub UI, it'll disappear in the merge

PR #807, opened 2024-07-31
Prove `schedContext_donate_ccorres`

michaelmcinerney:


proof/crefine/RISCV64/ADT_C.thy
```diff
@@ -1143,21 +1143,21 @@ lemma map_to_cnes_eq:
   apply (clarsimp simp: projectKO_opt_cte split: kernel_object.splits)
   done
 
-lemma ksPSpace_valid_pspace_scRefs_nonzero:
-  "\<lbrakk>\<exists>s. ksPSpace s = ah \<and> valid_pspace' s; map_to_scs ah p = Some sc\<rbrakk> \<Longrightarrow>
+lemma map_to_scs_Some_scRefs_nonzero:
+  "\<lbrakk>map_to_scs (ksPSpace s) p = Some sc; valid_objs' s; no_0_obj' s\<rbrakk> \<Longrightarrow>
```
lsf37, line 1147: It's kinda sad that we have to give up on the heap abstraction that was going on in ADT_C before because we need more and more invariants (this is not only in MCS, we also needed to do some of that in AARCH64 on master). I guess we could phrase a new `no_0_objs` that takes `ksPSpace s` instead of `s` here, but that does sound out of proportion to what we want to achieve.

Long story short, it's Ok to do this here, but it seems the projection approach that started here doesn't fully carry through without more work.

proof/crefine/RISCV64/ADT_C.thy
```diff
@@ -1176,62 +1176,103 @@ lemma scRefills_unique:
   done
 
 lemma scSize_unique:
-  assumes "refill_buffer_relation ah ch gs" "refill_buffer_relation ah' ch gs"
-      and "ksPSpace s = ah \<and> ksPSpace s p = Some (KOSchedContext sc)"
-      and "ksPSpace s' = ah' \<and> ksPSpace s' p = Some (KOSchedContext sc')"
+  assumes "refill_buffer_relation (ksPSpace as) ch gs" "refill_buffer_relation (ksPSpace as') ch gs"
+      and "ksPSpace as p = Some (KOSchedContext sc)"
+      and "ksPSpace as' p = Some (KOSchedContext sc')"
```
lsf37, line 1181: It looks like `ksSpace` is not needed here. Could it be just the following?

```suggestion
  assumes "refill_buffer_relation ah ch gs" "refill_buffer_relation ah' ch gs"
      and "ah p = Some (KOSchedContext sc)"
      and "ah' p = Some (KOSchedContext sc')"
```

proof/crefine/RISCV64/ADT_C.thy
```diff
@@ -1176,62 +1176,103 @@ lemma scRefills_unique:
...
   apply (clarsimp simp: Let_def)
   apply (drule_tac x=p in spec)+
   by fastforce
 
+lemma
+  assumes "pspace_aligned' as" "pspace_distinct' as" "valid_sched_context' asc as"
+  shows tcb_at'_scTCB:
+    "bound (scTCB asc) \<longrightarrow> tcb_at' (the (scTCB asc)) as"
+  and reply_at'_scReply:
+    "bound (scReply asc) \<longrightarrow> reply_at' (the (scReply asc)) as"
```
lsf37, line 1195: `bound` is the wrong predicate here (there is nothing bound, it's just an option pointer). `is_Some` would work, but it looks to me like avoiding `the` would be even better.

lsf37 (approved):
Happy with it, modulo the tweaks in ADT_C.

proof/refine/RISCV64/IpcCancel_R.thy
```diff
@@ -1051,7 +1051,8 @@ lemma scReplies_of_scTCB_update[simp]:
 crunch schedContextDonate
   for replies_of': "\<lambda>s. P (replies_of' s)" (* this interfers with list_refs_of_replies' *)
   and scReplies_of[wp]: "\<lambda>s. P' (scReplies_of s)"
-  (simp: crunch_simps wp: crunch_wps)
+  (simp: crunch_simps wp: crunch_wps
+   rule: schedContextDonate_def[simplified updateSchedContext_def])
```
corlewis, line 1055: This use of `rule` feels a little bit strange, is it to avoid proving the equivalent lemmas for `updateSchedContext`?

corlewis (approved):

proof/refine/RISCV64/IpcCancel_R.thy
michaelmcinerney: Some of the changes in Refine are a bit of a hack job, yes. A bit of background...

I wanted to use `updateSchedContext` rather than `setSchedContext` because firstly, `updateSchedContext` does seem to be a better abstraction to use (where we get the sched context object and then immediately set it with a variant of the object we get) and second, we already have `ccorres` rules for `updateSchedContext` which are exact analogues of the preexisting rules for `threadSet`, and I wanted to just use those.

But `updateSchedContext` was only defined fairly late on in the Refine run, and it has quite different rules when compared with the rules for `setSchedContext`. For example, some of the `updateSchedContext` rules will use the `opt_pred/opt_map` style functions for heap access, rather than `obj_at'` style, and this causes the proofs to go haywire unless we just try to get it back into the original formulation. So sometimes I do properly update the proofs for this change, and other times I just try to get back what we had before. Maybe once the `opt_pred` style is the only style we have, we could make these clunky proofs nicer

proof/crefine/RISCV64/ADT_C.thy
michaelmcinerney: Yep, done, thanks

proof/crefine/RISCV64/ADT_C.thy
michaelmcinerney: I've changed this to use the `!tcbPtr. scTCB asc = Some tcbPtr --> tcb_at' tcbPtr as` format in all those statements

spec/haskell/src/SEL4/Object/SchedContext.lhs
```diff
@@ -664,10 +664,9 @@ This module uses the C preprocessor to select a target architecture.
 >         threadSet (\tcb -> tcb { tcbSchedContext = Nothing }) from
 >         cur <- getCurThread
 >         action <- getSchedulerAction
->         case action of
->             SwitchToThread candidate -> when (candidate == from || from == cur) rescheduleRequired
->             _ -> when (from == cur) rescheduleRequired
->     setSchedContext scPtr (sc { scTCB = Just tcbPtr })
+>         when (from == cur || action == SwitchToThread from)
+>             rescheduleRequired
+>     updateSchedContext scPtr (\sc -> sc { scTCB = Just tcbPtr })
```
Xaphiosis, line 669: It was kind of unexpected to see a haskell update in a commit that says "prove ..." with no further elaboration. It looks like this is a rephrase to make the proof easier, but it still left me scratching my head for a bit as to what's going on.

michaelmcinerney: Yes, this is just a rephrase to make the proof easier. This is what's happening the vast majority of the time when I update the Haskell, which was written seemingly without any thought as to the Simpl that will be produced. I can add a bit to the commit message saying that it is just to make the proof easier

proof/refine/RISCV64/IpcCancel_R.thy
Xaphiosis: This could use at least a short comment then ("prefer X because Y", or "avoid X due to Y") or something like that. This kind of hidden knowledge is what makes proofs difficult to figure out / fix / update later, e.g. in the case you cited if we ever want to go forth and change to the `opt_pred` style.

Xaphiosis (approved):

spec/haskell/src/SEL4/Object/SchedContext.lhs
Xaphiosis: Yes please. I think it's important knowledge that this is only a rephrase and not a bug fix or update to match the C.

proof/refine/RISCV64/IpcCancel_R.thy
corlewis: Right, that all makes sense, I actually just added an `updateReply` for that exact reason of it being nicer to work with.

I can definitely understand making the minimal change to keep the proofs working. A comment could be helpful if you can find a suitable spot for it.

michaelmcinerney: On second thoughts, I decided to go with writing the rule

```
lemma updateSchedContext_scReplies_of:
  "(⋀sc. scReply (f sc) = scReply sc) ⟹ updateSchedContext scPtr f ⦃λs. P' (scReplies_of s)⦄"
```

and then adding that to the `crunch` instead. Hopefully that's a nicer approach

corlewis: Yeah, that's a lot nicer, I'm happy with that. Do you definitely need to add it to the `crunch` though? I would have guessed that it picked it up automatically.

michaelmcinerney: You're right, it's not needed in the `crunch`. After all these years, I thought it had to be in the `wp` for `crunch` to pick it up. I'll remove it

michaelmcinerney:
Forced pushed in order to update the commit message

proof/refine/RISCV64/IpcCancel_R.thy
corlewis: `crunch` picks rules up in a variety of different ways, including if the rule is already in the `wp` set and solves the generated goal, or if it has the same name as what `crunch` would produce. The former is how it works  when a rule is passed in via `wp`, the latter is why you don't need to do anything in this case.

michaelmcinerney: That's great to know, thanks!

PR #805, opened 2024-07-26
Overhaul arch-split in machine, ASpec and AInvs

Xaphiosis:
In https://github.com/seL4/l4v/pull/791 I went over AInvs for AARCH64 only. I've since expanded the approach to ASpec as well, and deployed it across all the architectures. The speed benefits still apply, and now ASpec is faster as well. I'll see if I can do a better benchmark, or we can pull it out of the regression tests.

🦆🦆🦆 I have put a lot of effort into splitting up the commits into being topical, because this is a rather large PR. They should not overlap much, so if you review a line in the commit view it's unlikely to be changed in another commit.

@mbrcknl @michaelmcinerney FYI in case you are interested in the new arch-splitting tech/approach.

- [ ] add `FIXME arch_split` to PR linter to stop people reintroducing it
- [x] remove `apply_trace` I accidentally left in (`proof/invariant-abstract/X64/ArchCSpace_AI.thy` in `ainvs: disambiguate acap_rights_update_id`)

PR #809, opened 2024-08-08
Arch-split overhaul for design spec

Xaphiosis:
The adventure continues, this time deploying `arch_requalify` and `arch_global_naming` to the design spec skeleton files.
Previously @corlewis benchmarked https://github.com/seL4/l4v/pull/805 as having 13-18% wall time improvement for ASpec, and 9-14% improvement for AInvs. This time I'm expecting only a bit of improvement in ExecSpec/BaseRefine/CBaseRefine, since only 52 cases of `context begin interpretation Arch .` were eliminated from the design spec in total.

🦆🦆🦆 I have split up the commits to be topical, which should make reviewing this easier, but it's a bit too much granularity to shove into l4v. Opinions welcome on how many commits we want this squashed to.

- [ ] still need to add FIXME arch_split to PR linter to stop people reintroducing it

### DIFF
--- a/spec/design/skel/AARCH64/ArchInvocationLabels_H.thy
+++ b/spec/design/skel/AARCH64/ArchInvocationLabels_H.thy
@@ -11,7 +11,7 @@ imports
   "Word_Lib.Enumeration"
   Setup_Locale
 begin
-context Arch begin global_naming AARCH64_H
+context Arch begin arch_global_naming (H)
 
 text \<open>
   An enumeration of arch-specific system call labels.
@@ -21,11 +21,12 @@ text \<open>
 
 end
 
-context begin interpretation Arch .
-requalify_types arch_invocation_label
-end
+(* not possible to move this requalification to generic, since enum instance proofs must
+   be done outside of Arch locale *)
+arch_requalify_types (H)
+  arch_invocation_label
 
-context Arch begin global_naming AARCH64_H
+context Arch begin arch_global_naming (H)
 
 #INCLUDE_HASKELL SEL4/API/InvocationLabels/AARCH64.hs CONTEXT AARCH64_H instanceproofs ONLY ArchInvocationLabel
 
--- a/spec/design/skel/AARCH64/ArchStructures_H.thy
+++ b/spec/design/skel/AARCH64/ArchStructures_H.thy
@@ -12,7 +12,7 @@ imports
   Hardware_H
 begin
 
-context Arch begin global_naming AARCH64_H
+context Arch begin arch_global_naming (H)
 
 #INCLUDE_SETTINGS keep_constructor=asidpool
 #INCLUDE_SETTINGS keep_constructor=arch_tcb
@@ -54,10 +54,8 @@ where
 
 end
 
-context begin interpretation Arch .
-
-requalify_types
+(* not possible to move this requalification to generic, as some arches don't have vcpu *)
+arch_requalify_types (H)
   vcpu
 
-end
 end
--- a/spec/design/skel/ARM/State_H.thy
+++ b/spec/design/skel/ARM/State_H.thy
@@ -16,7 +16,7 @@ imports
   RegisterSet_H
   MachineOps
 begin
-context Arch begin global_naming ARM_H
+context Arch begin arch_global_naming (H)
 
 definition
   Word :: "machine_word \<Rightarrow> machine_word"
@@ -28,16 +28,14 @@ where
 
 end
 
-context begin interpretation Arch .
-
-requalify_consts
+(* Note: while this requalify and arch-generic Haskell import of WordLib.lhs could be moved to
+   a generic theory, no good candidate theory exists at the moment. *)
+arch_requalify_consts (H)
   wordBits
-
-end
 
 #INCLUDE_HASKELL Data/WordLib.lhs all_bits NOT wordBits
 
-context Arch begin global_naming ARM_H
+context Arch begin arch_global_naming (H)
 
 #INCLUDE_HASKELL SEL4/Machine/RegisterSet.lhs Arch=ARM CONTEXT ARM_H all_bits NOT UserContext UserMonad getRegister setRegister newContext mask Word PPtr
 
--- a/spec/design/skel/FaultHandler_H.thy
+++ b/spec/design/skel/FaultHandler_H.thy
@@ -13,15 +13,14 @@ imports
   ArchFaultHandler_H
 begin
 
-context begin interpretation Arch .
-requalify_consts
-  syscallMessage
-  fromVPtr
-  exceptionMessage
-  debugPrint
+arch_requalify_consts (H)
   makeArchFaultMessage
   handleArchFaultReply
-end
+
+(* clobbers previously requalified abstract spec constants with design spec versions *)
+arch_requalify_consts (aliasing, H)
+  syscallMessage
+  exceptionMessage
 
 #INCLUDE_HASKELL_PREPARSE SEL4/API/Failures.lhs
 
--- a/spec/design/skel/KernelInit_H.thy
+++ b/spec/design/skel/KernelInit_H.thy
@@ -15,14 +15,10 @@ imports
   Thread_H
 begin
 
-context begin interpretation Arch .
-
-requalify_consts
+arch_requalify_consts
   getMemoryRegions
   addrFromPPtr
   init_machine_state
-
-end
 
 requalify_consts (in Arch)
   newKernelState
@@ -77,11 +73,19 @@ newKernelState_def:
         ksMachineState = init_machine_state
 \<rparr>"
 
+(* disambiguate name clash between Arch and non-arch consts with same names *)
 context Arch begin
-requalify_facts
+requalify_facts (aliasing)
+   newKernelState_def
+requalify_consts (aliasing)
+   newKernelState
+
+context begin global_naming global
+requalify_facts (aliasing)
    KernelInit_H.newKernelState_def
-requalify_consts
+requalify_consts (aliasing)
    KernelInit_H.newKernelState
+end
 end
 
 end
--- a/proof/crefine/AARCH64/Invoke_C.thy
+++ b/proof/crefine/AARCH64/Invoke_C.thy
@@ -2812,10 +2812,10 @@ lemma ctes_of_ex_cte_cap_to':
 
 
 lemma Arch_isFrameType_spec:
-  "\<forall>s. \<Gamma> \<turnstile> \<lbrace>s. unat \<acute>type \<le> fromEnum (maxBound::ArchTypes_H.object_type)\<rbrace>
+  "\<forall>s. \<Gamma> \<turnstile> \<lbrace>s. unat \<acute>type \<le> fromEnum (maxBound::object_type)\<rbrace>
              Call Arch_isFrameType_'proc
   \<lbrace> \<acute>ret__unsigned_long =
-     from_bool (isFrameType ((toEnum (unat \<^bsup>s\<^esup> type))::ArchTypes_H.object_type))\<rbrace>"
+     from_bool (isFrameType ((toEnum (unat \<^bsup>s\<^esup> type))::object_type))\<rbrace>"
   apply vcg
   apply (simp add: toEnum_object_type_to_H)
   apply (frule object_type_from_to_H)
--- a/proof/crefine/AARCH64/Retype_C.thy
+++ b/proof/crefine/AARCH64/Retype_C.thy
@@ -5047,7 +5047,7 @@ lemma placeNewObject_user_data:
   done
 
 definition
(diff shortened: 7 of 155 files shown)


spec/design/skel/KernelInit_H.thy
```diff
@@ -77,11 +73,19 @@ newKernelState_def:
         ksMachineState = init_machine_state
 \<rparr>"
 
+(* disambiguate newKernelState defined here as global vs the one in Arch as Arch *)
```
corlewis, line 76: Very minor, but you seem to have a few different ways of writing this comment for what I think are equivalent cases. For instance, the one here and `(* disambiguate Arch.invocation vs global.invocation *)` in `Invocations_H.thy`. It might be good if they were consistent, to make it obvious in the future that they're handling the same thing.

spec/design/skel/AARCH64/ArchStructures_H.thy
```diff
@@ -54,10 +54,8 @@ where
 
 end
 
-context begin interpretation Arch .
-
-requalify_types
+(* not possible to move this requalification to generic, as some arches don't have vcpu *)
```
corlewis, line 57: I'm not sure if I actually like this idea, but would it be possible to have a 'semi-generic' file that is only imported by an `ARM_HYP` and `AARCH64` file and so isn't visible on other architectures?

spec/design/skel/AARCH64/ArchInvocationLabels_H.thy
```diff
@@ -21,9 +21,8 @@ text \<open>
 
 end
 
-context begin interpretation Arch .
-requalify_types arch_invocation_label
-end
+arch_requalify_types (H)
+  arch_invocation_label
```
corlewis, line 25: Out of curiosity since you didn't mention it in the commit or leave a comment, why can't this one be moved to a generic theory?

spec/design/skel/ARM/State_H.thy
```diff
@@ -28,13 +28,9 @@ where
 
 end
 
-context begin interpretation Arch .
-
-requalify_consts
+arch_requalify_consts (H)
   wordBits
```
corlewis, line 32: Same question, why can't `wordBits` be moved?

corlewis (approved):
Very nice!

For your question about the commits, I actually don't see too much of a problem with how you've currently split it up. I would maybe combine the two 'use arch_global_naming' ones, although it's then not clear whether the arch_requalify part of the first one would belong there. Splitting up the generic and arch parts of deploying the architecture makes sense to me, since you need to do the generic ones first to easily see what can be removed from the arch files.

corlewis:
Promising initial benchmark!
```
master:
  N = 10
  Average ARM ExecSpec: (00:01:34 real, 00:03:30 cpu, 4.78GB)
  Standard Deviation: (00:00:00 real, 00:00:01 cpu, 0.73GB)
arch-split_design:
  N = 10
  Average ARM ExecSpec: (00:01:20 real, 00:03:09 cpu, 4.65GB)
  Standard Deviation: (00:00:01 real, 00:00:03 cpu, 0.79GB)
Percentage change: -15.3%
```
I'll add the other architectures once the tests complete.

spec/design/skel/ARM/State_H.thy
Xaphiosis: If we want to move it, we have to move `#INCLUDE_HASKELL Data/WordLib.lhs all_bits NOT wordBits` as well... and then there's the question where to move it *to*... upwards, there's RegisterSet_H, but that's arch-specific as well. Downstream, we need this before Untyped_H (generic), Structures_H (generic), ArchVSpace_H (arch).
Since `Structures_H` (generic) pulls in `State_H` (arch), then that'd be the only place we could put the wordBits-using stuff along with the requalify. Does it make sense there?

spec/design/skel/AARCH64/ArchInvocationLabels_H.thy
Xaphiosis: Can't be moved due to enum instantiations (enum, enum_alt, enumeration_both) right after it, which can't be done inside a locale. Want me to add a comment?

spec/design/skel/AARCH64/ArchStructures_H.thy
Xaphiosis: This idea has merit, especially if we ever want to go for the golden chalice that is being generic in vcpu on/off, but I think right now that's not going to gain us much for these small blips.

corlewis: Agreed, not worth doing now, just a thought I wanted to get out of my head,

spec/design/skel/AARCH64/ArchInvocationLabels_H.thy
corlewis: Yeah, I think a comment would be helpful.

corlewis:
I'm actually a bit surprised with how good these numbers are!
```
master:
  N = 10
  Average ARM BaseRefine: (00:01:29 real, 00:02:36 cpu, 3.89GB)
  Standard Deviation: (00:00:01 real, 00:00:03 cpu, 0.37GB)
arch-split_design:
  N = 10
  Average ARM BaseRefine: (00:01:15 real, 00:02:16 cpu, 3.55GB)
  Standard Deviation: (00:00:01 real, 00:00:02 cpu, 0.51GB)
Percentage change: -15.6%
```
```
master:
  N = 10
  Average ARM_HYP BaseRefine: (00:02:07 real, 00:03:44 cpu, 4.25GB)
  Standard Deviation: (00:00:01 real, 00:00:07 cpu, 0.31GB)
arch-split_design:
  N = 10
  Average ARM_HYP BaseRefine: (00:01:41 real, 00:02:54 cpu, 4.74GB)
  Standard Deviation: (00:00:01 real, 00:00:02 cpu, 0.52GB)
Percentage change: -20.2%
```
```
master:
  N = 10
  Average X64 BaseRefine: (00:02:04 real, 00:03:34 cpu, 4.30GB)
  Standard Deviation: (00:00:01 real, 00:00:04 cpu, 0.45GB)
arch-split_design:
  N = 10
  Average X64 BaseRefine: (00:01:40 real, 00:02:56 cpu, 3.96GB)
  Standard Deviation: (00:00:02 real, 00:00:04 cpu, 0.59GB)
Percentage change: -19.5%
```
```
master:
  N = 10
  Average RISCV64 BaseRefine: (00:01:23 real, 00:02:28 cpu, 3.75GB)
  Standard Deviation: (00:00:01 real, 00:00:03 cpu, 0.67GB)
arch-split_design:
  N = 10
  Average RISCV64 BaseRefine: (00:01:07 real, 00:02:05 cpu, 3.59GB)
  Standard Deviation: (00:00:01 real, 00:00:03 cpu, 0.34GB)
Percentage change: -18.8%
```
```
master:
  N = 10
  Average AARCH64 BaseRefine: (00:01:56 real, 00:03:24 cpu, 4.44GB)
  Standard Deviation: (00:00:02 real, 00:00:05 cpu, 0.44GB)
arch-split_design:
  N = 10
  Average AARCH64 BaseRefine: (00:01:31 real, 00:02:42 cpu, 4.55GB)
  Standard Deviation: (00:00:01 real, 00:00:02 cpu, 0.89GB)
Percentage change: -22.0%
```

spec/design/skel/ARM/State_H.thy
corlewis: Hmm, my first reaction is that it makes more sense in `State_H` than in `Structures_H`, which doesn't help. It might be too hard to move now, but maybe a comment at least.

spec/design/skel/KernelInit_H.thy
corlewis: Argh, the hard question. Reading through them again, they actually are for subtly different cases, along with the third phrasing of `(* disambiguate name clash between Arch and non-arch consts with same names *)` which you seem to have used when a collection of consts is being requalified as global at the same time.

I think I am leaning towards that third phrasing as being the most applicable in the different cases, although I'm not that fussed. Consistency would be nice but this isn't worth spending too much time or brain power on optimising.

spec/design/skel/AARCH64/ArchStructures_H.thy
lsf37: Agreed, should leave that for later, for if/when we try for generic VCPU on/off.

spec/design/skel/FaultHandler_H.thy
```diff
@@ -13,15 +13,14 @@ imports
...
-  debugPrint
+arch_requalify_consts (H)
   makeArchFaultMessage
   handleArchFaultReply
-end
+
+(* FIXME arch-split: clobbers generic with arch version, no disambiguation via global. *)
+arch_requalify_consts (aliasing, H)
+  syscallMessage
+  exceptionMessage
```
lsf37, line 23: Do we really want to fix this (i.e. should it be FIXME), or should we just leave a warning comment here? It does seem like the proofs are expecting the current state, so we probably don't really want to change this.

lsf37 (approved):
Looking good, good to see that almost all of the proof impact is for nicer names. And it looks like the infrastructure has everything we need so far. Nice work!

lsf37:
In terms of commits, I wouldn't terribly mind the current state, but if you do want to squash more, I would squash the generic + arch part of the deployment, and maybe even put that together with the `design: use arch_global_naming` commit. If you want to go all-out even with the proof update for it. I'd keep the first two separate.

corlewis:
For completeness, `CBaseRefine` is already a fair bit quicker as well.
```
master:
  N = 10
  Average AARCH64 CBaseRefine: (00:19:44 real, 00:43:42 cpu, 19.60GB)
  Standard Deviation: (00:00:14 real, 00:01:02 cpu, 1.92GB)
arch-split_design:
  N = 10
  Average AARCH64 CBaseRefine: (00:16:51 real, 00:35:42 cpu, 18.39GB)
  Standard Deviation: (00:00:25 real, 00:02:12 cpu, 2.47GB)
Percentage change: -14.6%
```

spec/design/skel/AARCH64/ArchInvocationLabels_H.thy
Xaphiosis: going with:
```
(* not possible to move this requalification to generic, since enum instance proofs must
   be done outside of Arch locale *)
```
aside: it's really surprising how we set up the Arch locale, and then the Haskell translator immediately ends it to do the enum instance proofs outside the locale using `qualify/end_qualify`, after which it restarts the Arch locale, and that one's ended by the `end` you see in the skeleton file.

spec/design/skel/FaultHandler_H.thy
Xaphiosis: The issue I had for this one is that the *current*, pre-arch-split proofs expecting it doesn't mean that it's the right way around, especially minus any disambiguation. I guess the current proofs do indicate they want this as you say. Will double check.

spec/design/skel/ARM/State_H.thy
Xaphiosis: Actually, due to how theories are laid out it won't even get to Structures_H, since ArchTypes_H already needs the definitions from that import. So I'm going with this comment:
```
(* Note: while this requalify and arch-generic Haskell import of WordLib.lhs could be moved to
   a generic theory, no good candidate theory exists at the moment. *)
```

spec/design/skel/KernelInit_H.thy
Xaphiosis: Ok, going with the "third phrasing". Might as well be consistent. I think the first comments were written while I was trying to figure out what was going on.

spec/design/skel/FaultHandler_H.thy
Xaphiosis: This is very well spotted, and made me go look at the other FIXME. What's happening is that we requalified the Arch version in ASpec (e.g. from AARCH64 or AARCH64_A), and now we're requalifying the Haskell version with the same name (AARCH64_H).

When I looked at the *other* FIXME though, it turned out that this was also the case except for `sanitiseRegister` and `getSanitiseRegisterInfo` ... oh man, it turns out that we imported *only* the generic const for those two, then the Arch const+def, then requalified the Arch const to generic (clobbering the generic const), then used type-compatibility to have generic imported definitions that mention those constants end up using the Arch ones instead!
I guess someone didn't want to import a generic constant that was essentially defined to be the same as the Arch constant, but then still left the original generic constants floating around. Super confusing. So at least to alleviate confusion, I removed the imports of the unused constants, which also meant no aliasing warning.

corlewis:
The latest changes look good to me, once squashed I think this is ready to be merged.

Xaphiosis:
Thank you again for benchmarking these @corlewis!

spec/design/skel/FaultHandler_H.thy
lsf37: Cool, glad that this untangled parts of it.

lsf37:
Also happy with it, merging it now!

PR #811, opened 2024-08-16
verification for deferred cache flush in untyped reset

lsf37:
Verification update for PR seL4/seL4#1289 where we defer the cache flush from untyped reset to the actual retype operation, which means that only flushes that are actually necessary will be performed.

This reduces the time for an untyped reset from 25 sec to 1.5 sec on the `tx2` board with 8GB of memory. Smaller, but still significant benefits for boards with less memory.

This was a lot more fiddly than I thought and touches some ugly old proofs. I've tried to streamline them at least a little bit and make them more consistent, but they have not become much prettier.

Test with: seL4/seL4#1289

### DIFF
--- a/lib/Monad_Commute.thy
+++ b/lib/Monad_Commute.thy
@@ -140,6 +140,21 @@ lemma mapM_x_commute:
   apply auto
   done
 
+(* Proof needs to be different from mapM_x_commute, to eliminate "distinct" *)
+lemma mapM_x_commute_T:
+  assumes commute: "\<And>r. monad_commute \<top> (b r) a"
+  shows "monad_commute \<top> (mapM_x b xs) a"
+  apply (induct xs)
+   apply (clarsimp simp: mapM_x_Nil return_def bind_def monad_commute_def)
+  apply (clarsimp simp: mapM_x_Cons)
+  apply (rule monad_commute_guard_imp)
+   apply (rule commute_commute, rule monad_commute_split)
+     apply (rule commute_commute, assumption)
+    apply (rule commute_commute, rule commute)
+   apply wp
+  apply clarsimp
+  done
+
 lemma commute_name_pre_state:
   assumes "\<And>s. P s \<Longrightarrow> monad_commute ((=) s) f g"
   shows "monad_commute P f g"
--- a/proof/crefine/AARCH64/Recycle_C.thy
+++ b/proof/crefine/AARCH64/Recycle_C.thy
@@ -252,8 +252,6 @@ lemma range_cover_nca_neg: "\<And>x p (off :: 9 word).
   apply (simp add: pageBits_def objBits_simps)
   done
 
-lemmas unat_of_nat32' = unat_of_nat_eq[where 'a=32]
-
 lemma clearMemory_PageCap_ccorres:
   "ccorres dc xfdc (invs' and valid_cap' (ArchObjectCap (FrameCap ptr undefined sz False None))
            and (\<lambda>s. 2 ^ pageBitsForSize sz \<le> gsMaxObjectSize s)
@@ -268,30 +266,27 @@ lemma clearMemory_PageCap_ccorres:
   apply (cinit' lift: bits_' ptr___ptr_to_unsigned_long_')
    apply (rule_tac P="capAligned (ArchObjectCap (FrameCap ptr undefined sz False None))"
                 in ccorres_gen_asm)
-   apply (rule ccorres_Guard_Seq)
+   apply (rule ccorres_Guard)
    apply (simp add: clearMemory_def)
-   apply (simp add: doMachineOp_bind)
-   apply (rule ccorres_split_nothrow_novcg_dc)
-      apply (rule_tac P="?P" in ccorres_from_vcg[where P'=UNIV])
-      apply (rule allI, rule conseqPre, vcg)
-      apply (clarsimp simp: valid_cap'_def capAligned_def
-                            is_aligned_no_wrap'[OF _ word64_power_less_1])
-      apply (prop_tac "ptr \<noteq> 0")
-       subgoal
-         apply (simp add: frame_at'_def)
-         apply (drule_tac x=0 in spec)
-         apply (clarsimp simp: pageBitsForSize_def bit_simps split: vmpage_size.splits)
-         done
-      apply simp
-      apply (prop_tac "3 \<le> pageBitsForSize sz")
-       apply (simp add: pageBitsForSize_def bit_simps split: vmpage_size.split)
-      apply (rule conjI)
-       apply (erule is_aligned_weaken)
-       apply (clarsimp simp: pageBitsForSize_def split: vmpage_size.splits)
-      apply (rule conjI)
-       apply (rule is_aligned_power2)
-       apply (clarsimp simp: pageBitsForSize_def split: vmpage_size.splits)
-      apply (clarsimp simp: ghost_assertion_size_logic[unfolded o_def])
+   apply (rule_tac P="?P" in ccorres_from_vcg[where P'=UNIV])
+   apply (rule allI, rule conseqPre, vcg)
+   apply (clarsimp simp: valid_cap'_def capAligned_def
+                         is_aligned_no_wrap'[OF _ word64_power_less_1])
+   apply (prop_tac "ptr \<noteq> 0")
+    subgoal
+      apply (simp add: frame_at'_def)
+      apply (drule_tac x=0 in spec)
+      apply (clarsimp simp: pageBitsForSize_def bit_simps split: vmpage_size.splits)
+      done
+   apply simp
+   apply (prop_tac "3 \<le> pageBitsForSize sz")
+    apply (simp add: pageBitsForSize_def bit_simps split: vmpage_size.split)
+   apply (rule conjI)
+    apply (erule is_aligned_weaken)
+    apply (clarsimp simp: pageBitsForSize_def split: vmpage_size.splits)
+   apply (rule conjI)
+    apply (rule is_aligned_power2)
+    apply (clarsimp simp: pageBitsForSize_def split: vmpage_size.splits)
    apply (clarsimp simp: ghost_assertion_size_logic[unfolded o_def] frame_at'_def)
    apply (simp add: flex_user_data_at_rf_sr_dom_s bit_simps)
    apply (clarsimp simp: field_simps word_size_def mapM_x_storeWord_step)
@@ -308,7 +303,7 @@ lemma clearMemory_PageCap_ccorres:
     apply (erule allfEI[where f=of_nat])
     apply (clarsimp simp: bit_simps)
     apply (subst(asm) of_nat_power, assumption)
-    apply simp
+     apply simp
      apply (insert pageBitsForSize_64 [of sz])[1]
      apply (erule order_le_less_trans [rotated])
      apply simp
@@ -376,19 +371,8 @@ lemma clearMemory_PageCap_ccorres:
     apply (simp add: bit_simps)
     apply (simp add: of_nat_power[where 'a=64, folded word_bits_def])
     apply (simp add: pageBits_def ko_at_projectKO_opt[OF user_data_at_ko])
-   (* FIXME AARCH64 indentation *)
    apply (rule inj_Ptr)
-     apply csymbr
-     apply (ctac add: cleanCacheRange_RAM_ccorres)
-    apply wp
-   apply (simp add: guard_is_UNIV_def unat_of_nat
-                    word_bits_def capAligned_def word_of_nat_less)
-  apply (clarsimp simp: word_bits_def valid_cap'_def
-                        capAligned_def word_of_nat_less)
-  apply (frule is_aligned_addrFromPPtr_n, simp add: pageBitsForSize_def split: vmpage_size.splits)
-  apply (simp add: bit_simps pptrBaseOffset_alignment_def)+
-  apply (simp add: is_aligned_no_overflow' addrFromPPtr_mask_cacheLineBits)
-  apply (simp add: pageBitsForSize_def bit_simps split: vmpage_size.splits)
+  apply (clarsimp simp: word_bits_def valid_cap'_def capAligned_def word_of_nat_less)
   done
 
 declare replicate_numeral [simp]
--- a/proof/crefine/AARCH64/Retype_C.thy
+++ b/proof/crefine/AARCH64/Retype_C.thy
@@ -5818,6 +5818,40 @@ lemma updatePTType_ccorres:
   apply (clarsimp simp: cvariable_array_map_relation_def split: if_splits)
   done
 
+lemma placeNewDataObject_dev_ccorres:
+  "ccorresG rf_sr \<Gamma> dc xfdc
+  (createObject_hs_preconds regionBase newType us True
+      and K (APIType_capBits newType us = pageBits + us))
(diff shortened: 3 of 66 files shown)


lsf37:
Right, I forgot to update RISCV64 and X64 for the new `init_arch_objs` parameter. Will add.

lsf37:
Now done. The two additional commits should probably be squashed into the first two.

lib/Monad_Commute.thy
```diff
@@ -140,6 +140,21 @@ lemma mapM_x_commute:
   apply auto
   done
 
+(* Needs separate proof to mapM_x_commute to eliminate "distinct" *)
```
Xaphiosis, line 143: Hard to comprehend, suggest: `... separate proof from ...`

spec/machine/ARM/MachineOps.thy
```diff
@@ -454,14 +454,12 @@ where
 
 section "Memory Clearance"
 
-text \<open>Clear memory contents to recycle it as user memory\<close>
+text \<open>Clear memory contents to recycle it as user memory. Do not yet flush the cache.\<close>
```
Xaphiosis, line 457: does this comment exist on ARM_HYP? didn't change there, only ARM+AARCH64

spec/haskell/src/SEL4/Object/ObjectType/ARM.lhs
```diff
@@ -218,12 +218,20 @@ Create an architecture-specific object.
 >             modify (\ks -> ks { gsUserPages =
 >               funupd (gsUserPages ks)
 >                      (fromPPtr regionBase) (Just ARMSmallPage)})
+>             when (not isDevice) $ doMachineOp $
```
Xaphiosis, line 221: nitpick: when (not P) == unless P

proof/invariant-abstract/ARM_HYP/ArchVSpaceEntries_AI.thy
```diff
@@ -600,26 +600,25 @@ lemma init_arch_objects_valid_pdpt:
...
+                   in hoare_post_imp, simp)
    apply wp
-  apply simp
+   apply (rule mapM_x_wp')
+   apply (wp copy_global_mappings_valid_pdpt_objs)
+   apply clarsimp
+   apply (drule_tac sz=sz in retype_addrs_aligned)
+      apply (simp add:range_cover_def)
+     apply (drule range_cover.sz,simp add:word_bits_def)
+    apply (simp add:range_cover_def)
```
Xaphiosis, line 619: while we're here, spaces after `:`?

proof/drefine/Untyped_DR.thy
```diff
@@ -695,10 +695,25 @@ lemma clearMemory_unused_corres_noop:
...
+lemma dcorres_mapM_x_machine_op_noop:
+  "\<lbrakk> \<And>m r. \<lbrace>\<lambda>ms. underlying_memory ms = m\<rbrace> mop r \<lbrace>\<lambda>rv ms. underlying_memory ms = m\<rbrace> \<rbrakk>
+       \<Longrightarrow> dcorres dc \<top> \<top> (return ()) (mapM_x (\<lambda>r. do_machine_op (mop r)) xs)"
+  apply (induct xs)
+   apply (simp add: mapM_x_Nil)
+  apply (simp add: mapM_x_Cons)
+  apply (rule corres_guard_imp)
+    apply (rule corres_split_noop_rhs)
+      apply (rule dcorres_machine_op_noop, assumption)
+     apply assumption
```
Xaphiosis, line 711: can wpsimp+ from here?

proof/crefine/AARCH64/Recycle_C.thy
```diff
@@ -268,30 +266,27 @@ lemma clearMemory_PageCap_ccorres:
...
-      apply (rule conjI)
-       apply (rule is_aligned_power2)
-       apply (clarsimp simp: pageBitsForSize_def split: vmpage_size.splits)
-      apply (clarsimp simp: ghost_assertion_size_logic[unfolded o_def])
+   apply (rule_tac P="?P" in ccorres_from_vcg[where P'=UNIV])
+   apply (rule allI, rule conseqPre, vcg)
+   apply (clarsimp simp: valid_cap'_def capAligned_def
+                         is_aligned_no_wrap'[OF _ word64_power_less_1])
+   apply (prop_tac "ptr \<noteq> 0")
+  subgoal
```
Xaphiosis, line 276: subgoal indentation if you care

proof/invariant-abstract/Untyped_AI.thy
```diff
@@ -279,12 +279,15 @@ locale Untyped_AI_arch =
...
+  "\<And>x cref ty dev ptr n us y.
+     \<lbrace>\<lambda>(s::'state_ext state). descendants_range x cref s \<rbrace>
+     init_arch_objects ty dev ptr n us y
+     \<lbrace>\<lambda>rv s. descendants_range x cref s\<rbrace>"
   assumes  init_arch_objects_caps_overlap_reserved[wp]:
-  "\<And>S ty ptr n us y. \<lbrace>\<lambda>(s::'state_ext state). caps_overlap_reserved S s\<rbrace>
-   init_arch_objects ty ptr n us y
-   \<lbrace>\<lambda>rv s. caps_overlap_reserved S s\<rbrace>"
+  "\<And>S ty dev ptr n us y.
+     \<lbrace>\<lambda>(s::'state_ext state). caps_overlap_reserved S s\<rbrace>
```
Xaphiosis, line 288: in git viewer this indent is unclear, is it supposed to be two spaces under meta-forall?

proof/crefine/AARCH64/Retype_C.thy
```diff
@@ -5818,6 +5818,41 @@ lemma updatePTType_ccorres:
...
+  hs
+  (placeNewDataObject regionBase us False)
+  (global_htd_update (\<lambda>s. (ptr_retyps (2^us) (Ptr regionBase :: user_data_C ptr))))"
+  apply (simp add: placeNewDataObject_def ccorres_cond_empty_iff)
+  apply (rule ccorres_guard_imp, rule placeNewObject_user_data, simp_all)
+  apply (clarsimp simp: createObject_hs_preconds_def invs'_def
+                        valid_state'_def valid_pspace'_def)
+  apply (frule range_cover.sz(1), simp add: word_bits_def)
+  done
+
```
Xaphiosis, line 5850: extra newline

proof/crefine/AARCH64/Retype_C.thy
```diff
@@ -5866,31 +5901,62 @@ proof -
...
+         apply (rule ccorres_split_nothrow_novcg)
+             apply (rule placeNewDataObject_no_dev_ccorres[where us=18 and newType=newType, simplified])
+            apply ceqv
+           apply  (ctac (no_vcg) add: gsUserPages_update_ccorres[folded modify_gsUserPages_update])
+             apply csymbr+
+             apply (rule ccorres_Guard_Seq)
+             apply (ctac (no_vcg) add: cleanCacheRange_RAM_ccorres)
+              apply csymbr
+              apply (rule ccorres_return_C; simp)
+             apply wp
```
Xaphiosis, line 5943: out of curiosity: is there any reason why you prefer wp over wpsimp in this proof?

proof/refine/ARM/Detype_R.thy
```diff
@@ -4533,26 +4552,97 @@ lemma doMachineOp_ksArchState_commute:
...
+  apply (force simp add: select_f_def empty_fail_def)
+  done
+
+lemma doMachineOp_projectKO_pde:
+  "empty_fail f \<Longrightarrow> monad_commute \<top> (doMachineOp f) (projectKO ko :: pde kernel)"
+  apply (clarsimp simp: monad_commute_def doMachineOp_def bind_assoc projectKO_def)
+  apply monad_eq
+  by (force split: option.splits simp: fail_def return_def select_f_def empty_fail_def)
+
+lemma doMachineOp_alignCheck:
```
Xaphiosis, line 4582: I might discover this later in the review, but are all the doMachineOp lemmas in this file new, or moved?

proof/refine/ARM_HYP/Retype_R.thy
```diff
@@ -5561,6 +5553,28 @@ lemma data_page_relation_retype:
    apply (clarsimp simp: image_def)+
   done
 
+lemma init_arch_objects_APIType_map2_VCPU_noop:
+  "init_arch_objects (APIType_map2 (Inr VCPUObject)) dev ptr n m addrs = return ()"
+  apply (simp add: init_arch_objects_def APIType_map2_def)
+  done
+
+lemma reorder_createObjects_dmo_userPages:
```
Xaphiosis, line 5561: not sure this is a reorder, it's only assoc

proof/refine/AARCH64/Retype_R.thy
```diff
@@ -5385,6 +5389,45 @@ lemma data_page_relation_retype:
...
+    od) = (do
+      (addrs, faddrs) <- (do
+        addrs <- createObjects y n ko sz;
+        _ <- modify (\<lambda>ks. ks\<lparr>ksArchState := gsPTTypes_update (g ks addrs) (ksArchState ks)\<rparr>);
+        return (addrs, f addrs)
+       od);
+      _ <- mapM_x (\<lambda>addr. doMachineOp (m addr)) addrs;
+      return faddrs
+    od)"
+  by (simp add: bind_assoc)
```
Xaphiosis, line 5429: not sure how I'd improve this, but these "reorder" lemmas (that don't actually reorder) are very specific, and a lot of the specifics don't matter; would it be worth at least abstracting these a bit? The second one looks a lot like the first one with P=True, etc.

proof/crefine/ARM_HYP/Retype_C.thy
```diff
@@ -7990,6 +8184,14 @@ lemma insertNewCap_ccorres:
   apply (simp add: untypedZeroRange_def Let_def)
   done
 
+lemma Arch_createObject_not_untyped:
+  "\<forall>s. \<Gamma>\<turnstile>\<^bsub>/UNIV\<^esub>
+   {s} Call Arch_createObject_'proc {t. cap_get_tag (ret__struct_cap_C_' t) \<noteq> scast cap_untyped_cap}"
```
Xaphiosis, line 8189: I'm a bit lost as to why this spec override is needed (i.e. how you came up with it)

proof/crefine/ARM_HYP/Retype_C.thy
```diff
@@ -7808,6 +7962,43 @@ lemma cleanCacheRange_PoU_preserves_bytes:
...
+                          \<and> byte_regions_unmodified' s t}" for s])
+  apply (rule conseqPre, vcg exspec=cleanByVA_preserves_bytes)
+  by (safe intro!: byte_regions_unmodified_hrs_mem_update
+            elim!: byte_regions_unmodified_trans byte_regions_unmodified_trans[rotated],
+      (simp_all add: h_t_valid_field)+)
+
+lemma cleanCacheRange_RAM_preserves_bytes:
+  "\<forall>s. \<Gamma>\<turnstile>\<^bsub>/UNIV\<^esub> {s} Call cleanCacheRange_RAM_'proc
+      {t. hrs_htd (t_hrs_' (globals t)) = hrs_htd (t_hrs_' (globals s))
+         \<and> byte_regions_unmodified' s t}"
```
Xaphiosis, line 7991: is it the case that we can prove these on ARM and ARM_HYP but we need to assume these on AARCH64 because we made them machine ops?

Xaphiosis (approved):
I think this is not that bad, and the performance improvement on the seL4 side is definitely good.
Could I ask why x64 and RISCV are having such an easier time with this compared to the Arms?
For squashing, yes, I think the proofs update can be one commit (maybe add a note as to why it's mostly Arm-related).
Also small request: can this go in after the design spec arch-split overhaul?

proof/crefine/ARM_HYP/Retype_C.thy
lsf37: Yes, that is the case. It first made me question whether we should go to lower-level machine ops for AARCH64 as well, but I think the current level is still reasonable.

lsf37:
> I think this is not that bad, and the performance improvement on the seL4 side is definitely good. Could I ask why x64 and RISCV are having such an easier time with this compared to the Arms?

RISCV64 and X64 have different cache flushing requirements -- on X64 the architecture doesn't require it, and the RISCV boards we are currently running don't have the level of caches that would need to be flushed, so there are only interface changes for those two, but content changes for ARM.

> For squashing, yes, I think the proofs update can be one commit (maybe add a note as to why it's mostly Arm-related). 

👍 

> Also small request: can this go in after the design spec arch-split overhaul?

No problem!

proof/crefine/ARM_HYP/Retype_C.thy
lsf37: The proof that follows this one previously just ran`vcg` which happened to either unfold everything or use `modifies` rules for `Arch_createObject`, depending on which arch it was from. That unfolding now draws in cache functions which have loops and are quite big.

It turns out, that the proof doesn't actually need anything at all from `Arch_createObject`: it was only interested in the case where a new untyped cap was returned, which `Arch_createObject` never does. So I proved that instead, which can also be done without further unfolding inside `Arch_createObject` and leads to faster proofs all around.

Xaphiosis: Nice. This is a super-annoying part of CRefine, making sure to always keep one's eye out for situations like this, preventing explosions.

lsf37:
(rebased on current master)

proof/crefine/AARCH64/Retype_C.thy
lsf37: If the `simp` is not necessary, I tend to prefer just `wp` because it's the less complex method and adding clarsimp into the mix at the end can mask problems later (although it can also solve simple unexpected new side goals, so it's not a completley clear cut argument -- I think I just like to use the simpler methods when they should suffice)

proof/refine/AARCH64/Retype_R.thy
lsf37: They are more assoc/regroup lemmas than reorder lemmas, I should probably rename. I don't see a good way to abstract them, though, the second one talks about gsPTTypes_update, whereas first one is about gsUserPages with a different form of term pattern. I need to spell out the rest so that g, ks, address etc match the right parts of the rest term. We could make it slightly more abstract by replacing the mapM_x and doMachineOp with variables, but that risks those variables being matched wrong in HO unification. They are really just helpers for extremely specific situations.

lsf37: (have renamed them to `regroup_...` now)

proof/refine/ARM/Detype_R.thy
lsf37: changed versions of moved lemmas (some where not phrased with monad_commute before, but the originals where in the same theoory). I guess we could move them further up, but we also do not seem to need them anywhere else, so I tried to keep noise low.

spec/haskell/src/SEL4/Object/ObjectType/ARM.lhs
lsf37: True, but the `not` lines up more nicely with C

spec/machine/ARM/MachineOps.thy
lsf37: Well spotted, have updated.

PR #812, opened 2024-08-20
lib: Requalify: enable ctrl+click jumps for arch_requalify

Xaphiosis:
This uses a complex workaround to unify the parse location of the original name encoded in YXML with the name the user is actually aiming for. While intuitively one could think to use the position of the name we parsed, there appears to be no way to pass that location to read_const/read_type_name, as those expect to be fed the result of Parse.const and Parse.typ respectively.

🦆🦆🦆: aside: this also appears why the code does the weird `(Parse.position (Scan.ahead parse_nm -- Parse.name))` setup that I couldn't make sense of before.

### DIFF
--- a/lib/Requalify.thy
+++ b/lib/Requalify.thy
@@ -8,7 +8,6 @@
 (*
   Requalify constants, types and facts into the current naming.
   Includes command variants that support implicitly using the L4V_ARCH environment variable.
-
 *)
 
 text \<open>See theory @{text "test/Requalify_Test.thy"} for commented examples and usage scenarios.\<close>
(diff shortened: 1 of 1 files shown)


lib/Requalify.thy
```diff
@@ -141,6 +140,29 @@ val get_fact_nm = (fst oo global_fact)
...
+end
+
+(* These versions, prior to resolving the const/type name, override that name in the parsed
+   YXML-encoded string entry with a name that will actually resolve to something, while preserving
+   other parse data such as parsed document location.
+   This updates the document markup at the original parse location of entry with a reference to the
+   const/type named by "name", allowing the user to ctrl+click to jump to the const/type definition.
+
+   For example, if the user requested "arch_requalify_consts (A) some_const", then they are
+   targeting something like "ARM_H.some_const", and expect to be able to jump to its definition.
```
Xaphiosis, line 159: urgh, should be `ARM_A.`

corlewis (approved):
I can't say that I fully understand the internal details, but if it works then it looks good to me!

lsf37 (approved):
I approve of these uses of deep magic.

PR #813, opened 2024-08-27
lib: Requalify: better document annotation

Xaphiosis:
Previous attempts to update the code to allow ctrl+clicking on names resulted in markup overlap errors for arch_requalify commands. This was due to a misconception that YXML strings already containing correct position information (e.g. obtained via Parse.const) would be used directly by Proof_context.read_const/read_type_name. This is not the case. Instead, position information is adjusted for the content length, and since we add a prefix for L4V_ARCH, we got annotation that did not match the document text.

After hours of digging to uncover the above issue, we get this:
* the "check_parsed_nm" in gen_requalify was not an extra check, but a hook to do document annotation; updated comments
* use Proof_Context.check_const/check_type_name with parsed position of Parse.name, and manually report resulting markup
* since the above works for arch_requalify_const, the normal requalify_const becomes a simpler instance
* the whole Parse.const/Parse.typ was a red herring and ultimately did not add anything to the code except for read_const/read_type_name interaction; removed

The resulting code is simpler; a large cautionary comment remains for future generations.

### DIFF
--- a/lib/Requalify.thy
+++ b/lib/Requalify.thy
@@ -98,21 +98,26 @@ val arch_global_opts = Scan.optional (Args.parens arch_suffix) ""
 
 in
 
-fun gen_requalify get_proper_nm parse_nm check_parsed_nm alias_fn arch =
+fun gen_requalify get_proper_nm check_parsed_nm alias_fn arch =
 let
   val options = if arch then arch_options else generic_options
 in
-  (Parse.opt_target -- options --  Scan.repeat1 (Parse.position (Scan.ahead parse_nm -- Parse.name))
+  (Parse.opt_target -- options --  Scan.repeat1 (Parse.position Parse.name)
    >> (fn ((target, (aliasing, arch_suffix)), names) =>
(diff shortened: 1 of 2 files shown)


Xaphiosis:
Looks like this needs more work, some issues with creative requalify in CRefine got me :(

corlewis (approved):
Great job working through this horrible issue and thank you for the detailed comment!

Xaphiosis:
Updated to fix the lookup of a resolved name, because somehow it's possible to look up a short name and get a qualified name that's been hidden... which for whatever reason is fine to create aliases to, but not find to resolve to (we use this in Kernel_C for temporarily making things visible to C Parser).
Added example in Requalify_Test, theory linter is complaining about diagnostic commands there.

lsf37 (approved):
Cool. And impressive job finding all that out, this is about as arcane as it gets.

corlewis (approved):
Looks even better now with the tests passing! Once again, thank you for the thorough comments in the new test.

PR #814, opened 2024-09-02
make proofs generic in cacheLineBits 

lsf37:
- import `cacheLineBits` from config (only used on Arm, defined but unused on RISCV, fixed on x64)
- check basic properties in Arch_Kernel_Config_Lemmas
- remove all other unfoldings apart from `cacheLineBits_val` which is used in ARM and ARM_HYP to fold C values (safe, because the range of values for `cacheLineBits` is small)
- the proof updates are similar between ARM and ARM_HYP, and a bit different in AARCH64, where most of it was already generic. They are now all based on the same principles.

There is a bunch of commits. I think they make reasonable sense as logical units, but happy to be persuaded otherwise if people want more squashing.

This changes bring Cortex-A72 into the ARM/ARM_HYP range.

### DIFF
--- a/proof/crefine/AARCH64/Wellformed_C.thy
+++ b/proof/crefine/AARCH64/Wellformed_C.thy
@@ -562,6 +562,50 @@ lemma ucast_irq_array_guard[unfolded irq_array_size_val, simplified]:
   done
 
 
+text \<open>cacheLineBits interface\<close>
+
+lemmas cacheLineBits_val =
+  cacheLineBits_def[unfolded Kernel_Config.CONFIG_L1_CACHE_LINE_SIZE_BITS_def]
+
+lemma cacheLineBits_leq_pageBits:
+  "cacheLineBits \<le> pageBits"
+  using cacheLineBits_sanity
+  by (simp add: pageBits_def)
+
+lemma pageBits_leq_pptrBaseOffset_alignment:
+  "pageBits \<le> pptrBaseOffset_alignment"
+  by (simp add: pageBits_def pptrBaseOffset_alignment_def)
+
+lemma cacheLineBits_leq_pptrBaseOffset_alignment:
+  "cacheLineBits \<le> pptrBaseOffset_alignment"
+  by (rule order.trans, rule cacheLineBits_leq_pageBits, rule pageBits_leq_pptrBaseOffset_alignment)
+
+lemma cacheLineBits_leq_pbfs:
+  "cacheLineBits \<le> pageBitsForSize sz"
+  by (rule order.trans, rule cacheLineBits_leq_pageBits, rule pbfs_atleast_pageBits)
+
+lemma addrFromPPtr_mask_pptrBaseOffset_alignment:
+  "n \<le> pptrBaseOffset_alignment
+   \<Longrightarrow> addrFromPPtr ptr && mask n = ptr && mask n"
+  unfolding addrFromPPtr_def
+  by (metis is_aligned_weaken mask_add_aligned pptrBaseOffset_aligned zadd_diff_inverse)
+
+lemma addrFromPPtr_mask_cacheLineBits:
+  "addrFromPPtr ptr && mask cacheLineBits = ptr && mask cacheLineBits"
+  by (rule addrFromPPtr_mask_pptrBaseOffset_alignment,
+      rule cacheLineBits_leq_pptrBaseOffset_alignment)
+
+lemma pptrBaseOffset_cacheLineBits_aligned[simp]:
+  "pptrBaseOffset && mask cacheLineBits = 0"
+  unfolding is_aligned_mask[symmetric]
+  by (rule is_aligned_weaken[OF pptrBaseOffset_aligned cacheLineBits_leq_pptrBaseOffset_alignment])
+
+lemma ptrFromPAddr_mask_cacheLineBits[simp]:
+  "ptrFromPAddr v && mask cacheLineBits = v && mask cacheLineBits"
+  by (simp add: ptrFromPAddr_def add_mask_ignore)
+
+(* end of Kernel_Config interface section *)
+
 (* Input abbreviations for API object types *)
 (* disambiguates names *)
 
@@ -644,26 +688,6 @@ abbreviation(input)
   prioInvalid :: sword32
 where
   "prioInvalid == seL4_InvalidPrio"
-
-(* caches *)
-
-definition cacheLineSize :: nat where
-  "cacheLineSize \<equiv> 6"
-
-lemma addrFromPPtr_mask_cacheLineSize:
-  "addrFromPPtr ptr && mask cacheLineSize = ptr && mask cacheLineSize"
-  apply (simp add: addrFromPPtr_def AARCH64.pptrBase_def pptrBaseOffset_def canonical_bit_def
-                   paddrBase_def cacheLineSize_def mask_def)
-  apply word_bitwise
-  done
-
-lemma pptrBaseOffset_cacheLineSize_aligned[simp]:
-  "pptrBaseOffset && mask cacheLineSize = 0"
-  by (simp add: pptrBaseOffset_def paddrBase_def pptrBase_def cacheLineSize_def mask_def)
-
-lemma ptrFromPAddr_mask_cacheLineSize[simp]:
-  "ptrFromPAddr v && mask cacheLineSize = v && mask cacheLineSize"
-  by (simp add: ptrFromPAddr_def add_mask_ignore)
 
 (* The magic 4 comes out of the bitfield generator -- this applies to all versions of the kernel. *)
 lemma ThreadState_Restart_mask[simp]:
--- a/proof/crefine/ARM/Retype_C.thy
+++ b/proof/crefine/ARM/Retype_C.thy
@@ -4745,7 +4745,9 @@ proof -
        apply (clarsimp simp: pageBits_def pdeBits_def
                              valid_arch_state'_def page_directory_at'_def pdBits_def)
       apply (clarsimp simp: is_aligned_no_overflow'[where n=14, simplified] pdeBits_def
-                            field_simps is_aligned_mask[symmetric] mask_AND_less_0)+
+                            field_simps is_aligned_mask[symmetric] mask_AND_less_0
+                            cacheLineBits_le_PageDirectoryObject_sz[unfolded APIType_capBits_def,
+                                                                    simplified])+
     done
 qed
 
--- a/proof/crefine/ARM_HYP/Machine_C.thy
+++ b/proof/crefine/ARM_HYP/Machine_C.thy
@@ -402,39 +402,38 @@ lemma index_xf_for_sequence:
 
 (* FIXME CLEANUP on all arches: this entire cache (section) has:
    - a number of useful word lemmas that can go into WordLib
-   - a ton of hardcoded "mask 6" and "64", which on sabre is "mask 5" and "32" respectively.
    - The proofs themselves are extremely similar.
      This can be much more generic! *)
 
 lemma lineStart_le_mono:
   "x \<le> y \<Longrightarrow> lineStart x \<le> lineStart y"
-  by (clarsimp simp: lineStart_def cacheLineBits_def shiftr_shiftl1 neg_mask_mono_le)
+  by (clarsimp simp: lineStart_def shiftr_shiftl1 neg_mask_mono_le)
 
 lemma lineStart_sub:
-  "\<lbrakk> x && mask 6 = y && mask 6\<rbrakk> \<Longrightarrow> lineStart (x - y) = lineStart x - lineStart y"
-  apply (clarsimp simp: lineStart_def cacheLineBits_def shiftr_shiftl1)
+  "\<lbrakk> x && mask cacheLineBits = y && mask cacheLineBits\<rbrakk> \<Longrightarrow> lineStart (x - y) = lineStart x - lineStart y"
+  apply (clarsimp simp: lineStart_def shiftr_shiftl1)
   apply (clarsimp simp: mask_out_sub_mask)
   apply (clarsimp simp: mask_eqs(8)[symmetric])
   done
 
 
 lemma lineStart_mask:
-  "lineStart x && mask 6 = 0"
-  by (clarsimp simp: lineStart_def cacheLineBits_def shiftr_shiftl1 mask_AND_NOT_mask)
+  "lineStart x && mask cacheLineBits = 0"
+  by (clarsimp simp: lineStart_def shiftr_shiftl1 mask_AND_NOT_mask)
 
 lemma cachRangeOp_corres_helper:
-  "\<lbrakk>w1 \<le> w2; w3 \<le> w3 + (w2 - w1); w1 && mask 6 = w3 && mask 6\<rbrakk>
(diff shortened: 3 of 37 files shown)


proof/crefine/ARM_HYP/Wellformed_C.thy
```diff
@@ -529,6 +529,29 @@ lemma ucast_irq_array_guard[unfolded irq_array_size_val, simplified]:
...
+lemma ptrFromPAddr_mask_SuperSection:
+  "n \<le> pageBitsForSize ARMSuperSection
+   \<Longrightarrow> ptrFromPAddr ptr && mask n = ptr && mask n"
+  apply (simp add: ptrFromPAddr_def)
+  apply (prop_tac "pptrBaseOffset AND mask n = 0")
+   apply (rule mask_zero[OF is_aligned_weaken[OF pptrBaseOffset_aligned]], simp)
+  apply (simp flip: mask_eqs(7))
+  done
+
+(* ------------ *)
```
Xaphiosis, line 611: Not a fan of this. It's not clear what it's for. Is it the end of the cache line bits, the end of the informal abstraction interface, or what.

spec/haskell/src/SEL4/Machine/Hardware/ARM/Exynos4210.hs
```diff
@@ -164,7 +164,7 @@ cacheInvalidateL2RangeCallback _ _ _ = return ()
 cacheCleanL2RangeCallback :: Ptr CallbackData -> PAddr -> PAddr -> IO ()
 cacheCleanL2RangeCallback _ _ _ = return ()
 
--- For the ARM1136
+-- Overwritten in Isabelle
```
Xaphiosis, line 167: I don't like this phrase. Don't we have an existing one that makes sense already? I'm not sure it's "overwritten in Isabelle", but rather it's clobbered by the abstract spec, or (non-exclusive) it's not imported by the translator.

Xaphiosis: Or maybe unused by the design/exec spec ...

spec/machine/AARCH64/Arch_Kernel_Config_Lemmas.thy
```diff
@@ -117,5 +117,15 @@ lemma maxIRQ_1_plus_eq_Suc_machine[simp]:
   "unat (1 + Kernel_Config.maxIRQ :: machine_word) = Suc Kernel_Config.maxIRQ"
   by (simp add: Kernel_Config.maxIRQ_def)
 
+
+(* cacheLineBits conditions *)
+
+(* Folding cacheLineBits_val in C functions only works reliably if cacheLineBits is not 1 and
+   not too large to conflict with other values used inside cache ops.
+   12 is ptBits, which is only available after ExecSpec. Anything smaller than ptBits works. *)
```
Xaphiosis, line 125: This is a bit of a worry, since it's a purely syntactic restriction. What is the risk that the proofs will fail in the future due to the appearance of some other number that happens to be the same as cacheLineBits?

spec/machine/ARM/Arch_Kernel_Config_Lemmas.thy
```diff
@@ -119,5 +119,15 @@ lemma maxIRQ_1_plus_eq_Suc_machine[simp]:
...
   by (simp add: Kernel_Config.maxIRQ_def)
 
+
+(* cacheLineBits conditions *)
+
+(* Folding cacheLineBits_val in C functions only works reliably if cacheLineBits is not 1 and
+   not too large to conflict with other values used inside cache ops.
+   10 is ptBits, which is only available after ExecSpec. Anything smaller than ptBits works. *)
+lemma cacheLineBits_sanity:
+  "cacheLineBits \<in> {2..10}"
```
Xaphiosis, line 129: would it make sense to say ptBits for these instead of the `..10` or whatever and have a `[simplified ptBits_def]` or whatever on the lemma statement?

Xaphiosis: oh right, "which is only available after ExecSpec". Hmm. Nitpick: "anything smaller than" and "not 1" don't gel together well

proof/crefine/AARCH64/Wellformed_C.thy
```diff
@@ -562,6 +562,50 @@ lemma ucast_irq_array_guard[unfolded irq_array_size_val, simplified]:
...
+lemmas cacheLineBits_val =
+  cacheLineBits_def[unfolded Kernel_Config.CONFIG_L1_CACHE_LINE_SIZE_BITS_def]
+
+lemma cacheLineBits_leq_pageBits:
+  "cacheLineBits \<le> pageBits"
+  using cacheLineBits_sanity
+  by (simp add: pageBits_def)
+
+lemma pageBits_leq_pptrBaseOffset_alignment:
+  "pageBits \<le> pptrBaseOffset_alignment"
```
Xaphiosis, line 576: I'm a bit lost, is pptrBaseOffset_alignment related to cache line size?

proof/crefine/ARM_HYP/Wellformed_C.thy
```diff
@@ -532,6 +532,51 @@ lemma ucast_irq_array_guard[unfolded irq_array_size_val, simplified]:
...
+  using cacheLineBits_sanity
+  by (simp add: pt_bits_def pte_bits_def)
+
+lemma ptBits_leq_pageBits:
+  "ptBits \<le> pageBits"
+  by (simp add: pt_bits_def pte_bits_def pageBits_def)
+
+lemma ptBits_leq_pdBits:
+  "ptBits \<le> pdBits"
+  by (simp add: pt_bits_def pd_bits_def pde_bits_def pte_bits_def)
```
Xaphiosis, line 550: it's odd to me that the above two lemmas are in the cacheLineBits interface section

proof/crefine/ARM_HYP/Wellformed_C.thy
```diff
@@ -532,6 +532,51 @@ lemma ucast_irq_array_guard[unfolded irq_array_size_val, simplified]:
...
+  "cacheLineBits < LENGTH(machine_word_len)"
+  apply (rule le_less_trans, rule cacheLineBits_le_ptBits)
+  by (simp add: pt_bits_def pte_bits_def)
+
+lemma APIType_capBits_PageDirectoryObject_pdBits:
+  "APIType_capBits PageDirectoryObject us = pdBits"
+  by (simp add: pd_bits_def APIType_capBits_def pde_bits_def)
+
+lemma cacheLineBits_le_PageDirectoryObject_sz:
+  "cacheLineBits \<le> APIType_capBits PageDirectoryObject us"
```
Xaphiosis, line 572: what's the intuition behind this constraint?

proof/crefine/ARM_HYP/VSpace_C.thy
```diff
@@ -3129,18 +3130,13 @@ lemma ccorres_return_void_C':
   done
 
 lemma is_aligned_cache_preconds:
-  "\<lbrakk>is_aligned rva n; n \<ge> 7\<rbrakk> \<Longrightarrow> rva \<le> rva + 0x7F \<and>
-          addrFromPPtr rva \<le> addrFromPPtr rva + 0x7F \<and> rva && mask 6 = addrFromPPtr rva && mask 6"
+  "\<lbrakk>is_aligned rva n; n \<ge> 7\<rbrakk> \<Longrightarrow> rva \<le> rva + 0x7F \<and> addrFromPPtr rva \<le> addrFromPPtr rva + 0x7F"
```
Xaphiosis, line 3133: do you happen to know where the 7 and 0x7F come from here? I'm wondering if it's related to cache line size in any way, or from somewhere else

proof/crefine/ARM_HYP/VSpace_C.thy
```diff
@@ -2668,10 +2668,12 @@ definition
...
 
 lemma doFlush_ccorres:
-  "ccorres dc xfdc (\<lambda>s. vs \<le> ve \<and> ps \<le> ps + (ve - vs) \<and> vs && mask 6 = ps && mask 6
-        \<comment> \<open>ahyp version translates ps into kernel virtual before flushing\<close>
-        \<and> ptrFromPAddr ps \<le> ptrFromPAddr ps + (ve - vs)
-        \<and> unat (ve - vs) \<le> gsMaxObjectSize s)
+  "ccorres dc xfdc
+           (\<lambda>s. vs \<le> ve \<and> ps \<le> ps + (ve - vs)
+                \<and> vs && mask cacheLineBits = ps && mask cacheLineBits
+                \<comment> \<open>ahyp version translates ps into kernel virtual before flushing\<close>
```
Xaphiosis, line 2674: while we're here, could you change ahyp to arm-hyp?

proof/crefine/ARM/Retype_C.thy
```diff
@@ -4745,7 +4745,9 @@ proof -
        apply (clarsimp simp: pageBits_def pdeBits_def
                              valid_arch_state'_def page_directory_at'_def pdBits_def)
       apply (clarsimp simp: is_aligned_no_overflow'[where n=14, simplified] pdeBits_def
-                            field_simps is_aligned_mask[symmetric] mask_AND_less_0)+
+                            field_simps is_aligned_mask[symmetric] mask_AND_less_0
+                            cacheLineBits_le_PageDirectoryObject_sz[unfolded APIType_capBits_def,
+                                                                    simplified])+
```
Xaphiosis, line 4750: By the way, are you aware of `@` from Rule_By_Method? We might be able to abbreviate these kinds of things with `[@<simp add: APIType_capBits_def>]`

Xaphiosis: (not needed in this one, but more general observation)

proof/crefine/ARM_HYP/Machine_C.thy
```diff
@@ -640,12 +649,13 @@ lemma cleanCacheRange_RAM_ccorres:
...
   "ccorres dc xfdc ((\<lambda>s. unat (w2 - w1) \<le> gsMaxObjectSize s)
                     and (\<lambda>_. w1 \<le> w2 \<and> w3 \<le> w3 + (w2 - w1)
-                      \<and> w1 && mask 6 = w3 && mask 6))
+                      \<and> w1 && mask cacheLineBits = w3 && mask cacheLineBits))
                    (\<lbrace>\<acute>start = w1\<rbrace> \<inter> \<lbrace>\<acute>end = w2\<rbrace> \<inter> \<lbrace>\<acute>pstart = w3\<rbrace>) []
            (doMachineOp (cleanCacheRange_PoU w1 w2 w3))
            (Call cleanCacheRange_PoU_'proc)"
   apply (rule ccorres_gen_asm)
   apply (cinit' lift: start_' end_' pstart_')
+   apply (fold cacheLineBits_val)
```
Xaphiosis, line 658: I think some kind of comment on this line (and in other lemmas) would be good, it's the linchpin of this whole exercise.

Xaphiosis (approved):
Minor comments and questions, else good. Nice to have these abstractions, updating this was a pain all the times we went 5->6->5

proof/crefine/ARM_HYP/Wellformed_C.thy
lsf37: You're right, what we usually do is `error "..."` -- I'll use that here as well.

proof/crefine/AARCH64/Wellformed_C.thy
lsf37: yes, it's larger :-) 

The relationship is that `addrFromPPtr` alignment is preserved by anything smaller or equal to `pptrBaseOffset_alignment`, which is what we need for the `.. && mask cacheLineBits = ...` lemma

proof/crefine/ARM_HYP/Wellformed_C.thy
Xaphiosis: Er, what? Why not remove it like it was previously and have two new lines to the next "section"? I don't think I know what `error` means in this context.

lsf37: Sorry, my comment somehow went into the wrong box, it was supposed to be an answer to the next one down (the "Overwritten in Isabelle" bit).

For this one, the problem was that the entire configuration section has no end, it just keeps going into abbreviations for notifications etc, which is weird. You're right that the divider is not great either. I guess the proper way to do this would be to find a good name for whatever comes next.

spec/haskell/src/SEL4/Machine/Hardware/ARM/Exynos4210.hs
Xaphiosis: Ah, this is what the `error` comment was with regard to. Yes, might be good to error them out.

proof/crefine/ARM_HYP/Wellformed_C.thy
Xaphiosis: At that point, it makes sense to give it a section/subsection for the configuration part, and/or end it with `(* end of Kernel_Config interface section *)` or something, then the general chaos can continue afterwards without a good name.

proof/crefine/ARM_HYP/Wellformed_C.thy
lsf37: Have added a comment for it.

spec/machine/AARCH64/Arch_Kernel_Config_Lemmas.thy
lsf37: The risk is very low. This code hasn't changed in a very long time, and it would be fairly strange to mention other numbers in these functions that happen to be in that cache line size range. The only other literal number that occurs in these functions before simplification is `1` (for loop counter increment and a few other expressions).

proof/crefine/ARM_HYP/VSpace_C.thy
lsf37: It's from somewhere else, I don't remember which one it was exactly, but one of the page table related sizes. They were grouped together here because they are left over as word proof obligations in one of the big C guard implication goals at the end.

proof/crefine/ARM/Retype_C.thy
lsf37: yes, for more complicated settings that can be nicer.

spec/machine/ARM/Arch_Kernel_Config_Lemmas.thy
lsf37: Good point, have adjusted it slightly.

proof/crefine/ARM_HYP/Wellformed_C.thy
lsf37: It's a side condition in Retype, specifically for the PoU flush invocation for page directories (hence page directory size). The actual number is just `pdBits`, but the symbolic input in Retype is the `APIType_capBits` form. I was trying not to unfold that there, because that makes other instances of `APIType_capBits` also numeric which makes it harder to distinguish where things are coming from.

spec/machine/AARCH64/Arch_Kernel_Config_Lemmas.thy
lsf37: (the upper range of the interval comes from the proof constraints)

PR #815, opened 2024-09-02
arm-hyp+aarch64 spec+proof: make generic in CONFIG_DISABLE_WFI_WFE_TRAPS

lsf37:
The config option CONFIG_DISABLE_WFI_WFE_TRAPS influences the value of HCR_VCPU (hcrVCPU in Haskell).

There is not much to make generic -- it is unfolded exactly once to compare the value in the spec to the value in C. The main part is defining hcrVCPU conditionally based on the config setting.

### DIFF
--- a/spec/haskell/src/SEL4/Machine/Hardware/AARCH64.hs
+++ b/spec/haskell/src/SEL4/Machine/Hardware/AARCH64.hs
@@ -442,7 +442,21 @@ check_export_arch_timer = error "Unimplemented - machine op"
 
 {- Constants -}
 
-hcrVCPU =  (0x80086039 :: Word) -- HCR_VCPU
+hcrCommon :: Word
+--          HCR_VM  | HCR_RW   | HCR_AMO | HCR_IMO | HCR_FMO | HCR_TSC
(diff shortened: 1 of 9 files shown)


spec/haskell/src/SEL4/Machine/Hardware/AARCH64.hs
```diff
@@ -442,7 +442,21 @@ check_export_arch_timer = error "Unimplemented - machine op"
...
+hcrTWE :: Word
+hcrTWE = bit 14
+
+hcrTWI :: Word
+hcrTWI = bit 13
+
+hcrVCPU :: Word -- HCR_VCPU
+hcrVCPU = if config_DISABLE_WFI_WFE_TRAPS
+          then hcrCommon
+          else hcrCommon .|. hcrTWE .|. hcrTWI
```
Xaphiosis, line 458: I'd probably have put the hcrTWI and TWE into a `where` clause to save on unfolding, but this is fine too.

Xaphiosis (approved):

spec/haskell/src/SEL4/Machine/Hardware/AARCH64.hs
lsf37: I think I’ll leave it this way for now, the parts might be useful to name independently in lemmas at some point.

PR #816, opened 2024-09-03
Prove `cancelBadgedSends_ccorres`

michaelmcinerney:
Test with https://github.com/seL4/seL4/pull/1302

### DIFF
--- a/proof/crefine/RISCV64/Finalise_C.thy
+++ b/proof/crefine/RISCV64/Finalise_C.thy
@@ -35,102 +35,217 @@ declare if_split [split del]
 definition
   "option_map2 f m = option_map f \<circ> m"
 
-lemma head_end_ksReadyQueues_':
-  "\<lbrakk> (s, s') \<in> rf_sr; ksReadyQueues_asrt s; pspace_aligned' s; pspace_distinct' s;
-     d \<le> maxDomain; p \<le> maxPriority \<rbrakk>
-   \<Longrightarrow> head_C (index (ksReadyQueues_' (globals s')) (cready_queues_index_to_C d p)) = NULL
-       \<longleftrightarrow> end_C (index (ksReadyQueues_' (globals s')) (cready_queues_index_to_C d p)) = NULL"
-  apply (frule (2) rf_sr_ctcb_queue_relation[where d=d and p=p])
-  apply (clarsimp simp: ready_queue_relation_def ksReadyQueues_asrt_def)
-  apply (drule_tac x=d in spec)
-  apply (drule_tac x=p in spec)
-  apply clarsimp
-  apply (frule tcbQueueHead_iff_tcbQueueEnd)
-  apply (clarsimp simp: ctcb_queue_relation_def option_to_ctcb_ptr_def split: option.splits)
-  apply (frule (3) obj_at'_tcbQueueHead_ksReadyQueues)
-  apply (frule (3) obj_at'_tcbQueueEnd_ksReadyQueues)
-  apply (rename_tac head "end")
-  apply (prop_tac "tcb_at' head s")
-   apply (fastforce simp: tcbQueueEmpty_def)
-  apply (prop_tac "tcb_at' end s")
-   apply (fastforce simp: tcbQueueEmpty_def)
-  apply (fastforce dest: tcb_at_not_NULL)
+lemma thread_state_ptr_set_tcbQueued_spec_tcb_fields:
+  "\<forall>s. \<Gamma> \<turnstile>\<^bsub>/UNIV\<^esub> \<lbrace>s. s \<Turnstile>\<^sub>c (cparent \<^bsup>s\<^esup>thread_state_ptr [''tcbState_C''] :: tcb_C ptr)\<rbrace>
+       Call thread_state_ptr_set_tcbQueued_'proc
+       {s'. option_map2 tcbEPNext_C (cslift s') = option_map2 tcbEPNext_C (cslift s)
+            \<and> option_map2 tcbEPPrev_C (cslift s') = option_map2 tcbEPPrev_C (cslift s)
+            \<and> option_map2 tcbPriority_C (cslift s') = option_map2 tcbPriority_C (cslift s)
+            \<and> option_map2 tcbDomain_C (cslift s') = option_map2 tcbDomain_C (cslift s)}"
+  apply (hoare_rule HoarePartial.ProcNoRec1)
+  apply (rule allI, rule conseqPre, vcg)
+  apply clarsimp \<comment> \<open>for speed\<close>
+  apply (clarsimp simp: typ_heap_simps h_t_valid_clift_Some_iff)
+  apply (frule h_t_valid_clift)
+  apply (clarsimp simp: packed_heap_update_collapse_hrs)
+  apply (intro conjI;
+         fwd_all_new \<open>subst parent_update_child\<close>
+                     \<open>fastforce intro!: h_t_valid_c_guard
+                                  simp: option_map2_def fun_eq_iff typ_heap_simps
+                                 split: if_splits\<close>)
   done
 
-lemma tcbSchedEnqueue_cslift_spec:
-  "\<forall>s. \<Gamma>\<turnstile>\<^bsub>/UNIV\<^esub> \<lbrace>s. \<exists>d v. option_map2 tcbPriority_C (cslift s) \<acute>tcb = Some v
-                       \<and> unat v \<le> numPriorities
-                       \<and> option_map2 tcbDomain_C (cslift s) \<acute>tcb = Some d
-                       \<and> unat d < Kernel_Config.numDomains
-                       \<and> (end_C (index \<acute>ksReadyQueues (unat (d*0x100 + v))) \<noteq> NULL
-                           \<longrightarrow> option_map2 tcbPriority_C (cslift s)
-                                   (head_C (index \<acute>ksReadyQueues (unat (d*0x100 + v))))
-                                           \<noteq> None
-                             \<and> option_map2 tcbDomain_C (cslift s)
-                                   (head_C (index \<acute>ksReadyQueues (unat (d*0x100 + v))))
-                                           \<noteq> None)
-                       \<and> (head_C (index \<acute>ksReadyQueues (unat (d * 0x100 + v))) \<noteq> NULL
-                          \<longleftrightarrow> end_C (index \<acute>ksReadyQueues (unat (d * 0x100 + v))) \<noteq> NULL)\<rbrace>
+lemma thread_state_ptr_set_tsType_spec_tcb_fields:
+  "\<forall>s. \<Gamma> \<turnstile>\<^bsub>/UNIV\<^esub> \<lbrace>s. s \<Turnstile>\<^sub>c (cparent \<^bsup>s\<^esup>thread_state_ptr [''tcbState_C''] :: tcb_C ptr)\<rbrace>
+       Call thread_state_ptr_set_tsType_'proc
+       {s'. option_map2 tcbEPNext_C (cslift s') = option_map2 tcbEPNext_C (cslift s)
+            \<and> option_map2 tcbEPPrev_C (cslift s') = option_map2 tcbEPPrev_C (cslift s)
+            \<and> option_map2 tcbPriority_C (cslift s') = option_map2 tcbPriority_C (cslift s)
+            \<and> option_map2 tcbDomain_C (cslift s') = option_map2 tcbDomain_C (cslift s)}"
+  apply (hoare_rule HoarePartial.ProcNoRec1)
+  apply (rule allI, rule conseqPre, vcg)
+  apply clarsimp \<comment> \<open>for speed\<close>
+  apply (clarsimp simp: typ_heap_simps h_t_valid_clift_Some_iff)
+  apply (frule h_t_valid_clift)
+  apply (clarsimp simp: packed_heap_update_collapse_hrs)
+  apply (intro conjI;
+         fwd_all_new \<open>subst parent_update_child\<close>
+                     \<open>fastforce intro!: h_t_valid_c_guard
+                                  simp: option_map2_def fun_eq_iff typ_heap_simps
+                                 split: if_splits\<close>)
+  done
+
+lemma tcb_queue_prepend_spec_tcb_fields:
+  "\<forall>s. \<Gamma> \<turnstile>\<^bsub>/UNIV\<^esub> {s}
+       Call tcb_queue_prepend_'proc
+       {s'. option_map2 tcbEPNext_C (cslift s') = option_map2 tcbEPNext_C (cslift s)
+            \<and> option_map2 tcbEPPrev_C (cslift s') = option_map2 tcbEPPrev_C (cslift s)
+            \<and> option_map2 tcbPriority_C (cslift s') = option_map2 tcbPriority_C (cslift s)
+            \<and> option_map2 tcbDomain_C (cslift s') = option_map2 tcbDomain_C (cslift s)}"
+  apply (rule allI, rule conseqPre)
+  apply vcg
+  apply (clarsimp simp: option_map2_def fun_eq_iff typ_heap_simps h_t_valid_clift_Some_iff
+                 split: if_splits)
+  done
+
+lemma addToBitmap_spec_tcb_fields:
+  "\<forall>s. \<Gamma> \<turnstile>\<^bsub>/UNIV\<^esub> {s}
+       Call addToBitmap_'proc
+       {s'. option_map2 tcbEPNext_C (cslift s') = option_map2 tcbEPNext_C (cslift s)
+            \<and> option_map2 tcbEPPrev_C (cslift s') = option_map2 tcbEPPrev_C (cslift s)
+            \<and> option_map2 tcbPriority_C (cslift s') = option_map2 tcbPriority_C (cslift s)
+            \<and> option_map2 tcbDomain_C (cslift s') = option_map2 tcbDomain_C (cslift s)}"
+  apply (rule allI, rule conseqPre)
+  apply vcg
+  apply clarsimp
+  done
+
+lemma ready_queues_index_spec_tcb_fields:
+  "\<forall>s. \<Gamma> \<turnstile>\<^bsub>/UNIV\<^esub> {s}
+       Call ready_queues_index_'proc
+       {s'. option_map2 tcbEPNext_C (cslift s') = option_map2 tcbEPNext_C (cslift s)
+            \<and> option_map2 tcbEPPrev_C (cslift s') = option_map2 tcbEPPrev_C (cslift s)
+            \<and> option_map2 tcbPriority_C (cslift s') = option_map2 tcbPriority_C (cslift s)
+            \<and> option_map2 tcbDomain_C (cslift s') = option_map2 tcbDomain_C (cslift s)}"
+  apply (hoare_rule HoarePartial.ProcNoRec1)
+  apply (rule allI, rule conseqPre)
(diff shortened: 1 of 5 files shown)


proof/crefine/RISCV64/Recycle_C.thy
```diff
@@ -908,16 +1027,20 @@ sorry (* FIXME RT: cancelBadgedSends_ccorres *) (*
...
-                apply (simp add: tcb_queue_relation'_def EPState_Send_def mask_def)
-                apply (drule (1) tcb_and_not_mask_canonical[where n=2])
-                 apply (simp (no_asm) add: tcbBlockSizeBits_def)
-                subgoal by (simp add: mask_def canonical_bit_def)
-               subgoal by (auto split: if_split)
-              subgoal by simp
+              apply (rule conjI)
+               apply (erule(1) cpspace_relation_ep_update_ep2)
+                apply (simp add: cendpoint_relation_def Let_def)
+                apply (subgoal_tac "tcb_at' (last (a # list)) \<sigma> \<and> tcb_at' a \<sigma>")
```
Xaphiosis, line 1033: reading this proof, I have no idea what list, a, lista, and xs are ... any chance you could rename things a bit to make it easier to follow, or add some commentary?

proof/crefine/RISCV64/Recycle_C.thy
```diff
@@ -805,16 +805,123 @@ lemma ntfn_q_refs'_no_NTFNBound[simp]:
...
+      apply (rule_tac Q'="\<lambda>_. tcb_at' t and valid_objs' and no_0_obj'
+                              and pspace_aligned' and pspace_distinct' and pspace_bounded'
+                              and (\<lambda>s. ksCurDomain s \<le> maxDomain)
+                              and (\<lambda>s. weak_sch_act_wf (ksSchedulerAction s) s)"
+                   in hoare_post_imp)
+       apply (fastforce simp: obj_at'_def)
+      apply wpsimp
+     apply (vcg exspec=setThreadState_modifies)
+    apply (ctac add: setThreadState_ccorres)
+   apply (fastforce simp: mask_def[where n=4] ThreadState_defs typ_heap_simps ctcb_relation_def
```
Xaphiosis, line 901: don't know what this 4 is; should we be concerned about this magic number?

proof/crefine/RISCV64/Recycle_C.thy
```diff
@@ -805,16 +805,123 @@ lemma ntfn_q_refs'_no_NTFNBound[simp]:
...
+            apply ceqv
+           apply (ctac add: possibleSwitchTo_ccorres)
+          apply wpsimp
+         apply (vcg exspec=refill_unblock_check_modifies)
+        apply (vcg exspec=sc_sporadic_modifies)
+       apply (rule_tac xf'="ret__unsigned_long_'"
+                   and val="from_bool False"
+                   and R="obj_at' (\<lambda>tcb. tcbSchedContext tcb = None) t"
+                   and R'=UNIV
+                    in ccorres_symb_exec_r_known_rv)
```
Xaphiosis, line 883: trivial, but you can make this a bit shorter by writing `ccorres_symb_exec_r_known_rv[where R'=UNIV]` (same above), and to indicate that the R' isn't interesting

proof/crefine/RISCV64/Recycle_C.thy
```diff
@@ -805,16 +805,123 @@ lemma ntfn_q_refs'_no_NTFNBound[simp]:
...
+     apply (rule ccorres_rhs_assoc)+
+     apply (ctac add: setThreadState_ccorres)
+       apply (clarsimp simp: ifCondRefillUnblockCheck_def)
+       apply (rule ccorres_pre_threadGet)
+       apply (rule ccorres_move_c_guard_tcb)
+       apply (clarsimp simp: when_def)
+       apply (simp add: if_to_top_of_bind)
+       apply (rule ccorres_if_lhs)
+        apply (clarsimp simp: bind_assoc, rename_tac scPtr)
+        apply (wpfix add: option.sel)
```
Xaphiosis, line 843: this looks interesting, what does the `add: option.sel` do here? I might want to pick up this technique

proof/crefine/RISCV64/Finalise_C.thy
```diff
@@ -35,102 +35,226 @@ declare if_split [split del]
...
+       {s'. option_map2 tcbEPNext_C (cslift s') = option_map2 tcbEPNext_C (cslift s)
+            \<and> option_map2 tcbEPPrev_C (cslift s') = option_map2 tcbEPPrev_C (cslift s)
+            \<and> option_map2 tcbPriority_C (cslift s') = option_map2 tcbPriority_C (cslift s)
+            \<and> option_map2 tcbDomain_C (cslift s') = option_map2 tcbDomain_C (cslift s)}"
+  (is "\<forall>s. \<Gamma> \<turnstile>\<^bsub>/UNIV\<^esub> {s} ?f ({s'. ?post s s'})")
   apply (rule allI, rule conseqPre)
    apply vcg_step
    apply vcg_step
     apply vcg_step
    apply vcg_step
```
Xaphiosis, line 231: something isn't right about this, but I also don't know what you're trying to do; would something like `apply (hoare_rule HoarePartial.ProcNoRec1)` help here? I definitely don't want this going in without some comments on why it's like this or what it's trying to do

proof/crefine/RISCV64/Finalise_C.thy
```diff
@@ -35,102 +35,226 @@ declare if_split [split del]
...
-lemma head_end_ksReadyQueues_':
-  "\<lbrakk> (s, s') \<in> rf_sr; ksReadyQueues_asrt s; pspace_aligned' s; pspace_distinct' s;
-     d \<le> maxDomain; p \<le> maxPriority \<rbrakk>
-   \<Longrightarrow> head_C (index (ksReadyQueues_' (globals s')) (cready_queues_index_to_C d p)) = NULL
-       \<longleftrightarrow> end_C (index (ksReadyQueues_' (globals s')) (cready_queues_index_to_C d p)) = NULL"
-  apply (frule (2) rf_sr_ctcb_queue_relation[where d=d and p=p])
-  apply (clarsimp simp: ready_queue_relation_def ksReadyQueues_asrt_def)
-  apply (drule_tac x=d in spec)
-  apply (drule_tac x=p in spec)
+lemma thread_state_ptr_set_tcbQueued_tcb_fields:
```
Xaphiosis, line 38: This starts a chain of lemmas which are all vcg spec rules, but don't have anything in their name to differentiate them as such. I think they really should. I also have a question: why do you need so many of these?

Xaphiosis: While I'm here, do you know what the option_map2 doing in this file? It doesn't have a FIXME move on it either...

proof/crefine/RISCV64/Recycle_C.thy
michaelmcinerney: I'm pretty sure that these are related to the endpoint queue, but this is a part of the proof that I didn't change; I did try to keep as much of the old proof as possible and didn't really look into improving it. But sure, I can investigate and see whether a nicer renaming can be done

proof/crefine/RISCV64/Recycle_C.thy
michaelmcinerney: I stole this from somewhere else. The `4` comes from the guard/precondition of `setThreadState_ccorres`:

```
(⦃∀cl fl. cthread_state_relation_lifted st (cl⦇tsType_CL := ´ts && mask 4⦈, fl)⦄
      ∩ ⦃´tptr = tcb_ptr_to_ctcb_ptr thread⦄)
```

There's unfortunately no comment there but I think it stems from the body of `thread_state_ptr_set_tsType`. It's pretty gory that low down. 

I am actually able to remove the `n=4`. Would it be better to just have it as `mask_def`, or is it generally unwise to unfold `mask_def`? I regrettably have very little experience working with masking and bit operations like this, so I'm not sure of the rules and recommendations.

proof/crefine/RISCV64/Recycle_C.thy
michaelmcinerney: I'm sure Gerwin or Corey would be able to give you a better explanation, but this has the effect of moving from a schematic which has `Some scPtr` as one of its arguments, to additionally having `scPtr` as one of its arguments, so that unification can go ahead in the following line. I think providing `wpfix` with other `.sel` rules will do similar things for other datatypes, but I've only used a handful of times myself

proof/crefine/RISCV64/Finalise_C.thy
michaelmcinerney: This is the most sensitive of the `hoarep/vcg` rules that I have in this PR, because `refill_unblock_check` contains a `While` loop. I use the `vcg_step` steps to very carefully get to the `While` loop, and then perform the `HoarePartial.reannotateWhileNoGuard` step. The other uses of this rule occur within `ccorres` rules where the stepping is done by `ccorres` split rules, I think. 

Similar use of `vcg_step` was made before in `setThreadState_cslift_spec`, which has been removed.

And yes, `apply (hoare_rule HoarePartial.ProcNoRec1)` will help cut down on these steps a little bit, but I can't completely remove the uses of `vcg_step`.

Would a comment after the `apply (hoare_rule HoarePartial.ProcNoRec1)` saying `(* step through the function until you get to the While loop *)` be reasonable?

And to answer the question of why all these `vcg` rules are needed, `cancelBadgedSends` has a `for` loop that ends up as a `While` loop of some kind in the Simpl. The proof is a rather disgusting induction (the step that performs the induction uses the rule `ccorres_inst_voodoo`) and the induction requires the body of the loop to not alter the head and the end of the queue over which the `for` loop runs. The previous rule `setThreadState_cslift_spec` established this for the much smaller operations which `master` does inside these loops. However, the functions performed in the loop for MCS are much more complicated (for example, `refill_unblock_check`, with its while loop) so we wanted to be very careful with this. With lots of help from Gerwin, we figured out that the preconditions of these `hoarep` rules could in fact be `UNIV`, which meant that the rather heavy-handed rule `tcbSchedEnqueue_cslift_precond_discharge` could be removed. We did need to think quite hard to do this, though, which is why, for instance, `thread_state_ptr_set_tcbQueued_tcb_fields` has some careful reasoning involving things such as `packed_heap_update_collapse_hrs`. There are so many rules because the functions called within the `for` loop contain many smaller functions, and we did not want to let `vcg` run wild and inline a whole bunch of things, just because that gives us terrible goals, and also takes a very long time in comparison with these verbose but fast proofs.

I was secretly hoping that these terrible vcg concerns would mean that we needed to handle the endpoint and notification queues with linked lists in the Haskell, similar to what I did for the ready and release queues. Then all this would be unnecessary, though of course an awful lot of work would be needed in Refine to carry that out. I think that these vcg proofs worked out really well in the end, and I'm hopefully that they will be sufficient for all loops we do within "ipc cancel" functions.

proof/crefine/RISCV64/Finalise_C.thy
michaelmcinerney: I'm not sure what comment would be good for these "`tcb_field`" rules, but maybe I could say before the first of them "some Hoare rules to show that many functions do not alter fields of concern within functions which cancel IPC". Does that sound OK?

And I'm not sure about `option_map2` being in this file, either. Should I move it to where `option_map` is? That is in `Lib.thy`.

lsf37: I think what Raf means is that the lemmas should have something in the name that identifies them as vcg rules (like we have `_wp` for wp rules etc). They are not full spec rules, so definitely don't call them only `_spec`, but they could be `_vcg_tcb_fields`, or `_spec_tcb_fields`.

To answer Raf's question why there are so many of them: they are exactly the same as `wp` rules for a specific property -- you need one for each function. We have so far not needed to do that much in CRefine, because we always were either able to rearrange the C so that there is no need for `vcg` or we were able to write a `spec` rule without unfolding inner functions much, but neither of these works here because there is a much longer chain of inner functions here than on master. Proving one lemma per function avoids the vcg blow-up you get with unfolding.

lsf37: `option_map2`: I'm not sure either. It would probably make sense to put it into Lib, but it is also unrelated to this PR, it's just using what was already there, we only see more of it because we now need to prove more lemmas that have it in their statement.

If we do clean it up and move to Lib, it should probably happen in its own commit so we can cherry-pick that onto master.

proof/crefine/RISCV64/Finalise_C.thy
lsf37: It is perfectly possibly to annotate the while loop before we call `vcg`, let me look up the rule for that.

lsf37: I stand corrected, I can't find a rule in Simple that is analogous to `whileLoop_add_inv` in the monad world. We should definitely prove one and use that instead of `reannotateWhileNoGuard`.

The idea is to make a rule that is the same as `reannotateWhileNoGuard`, but is an equality instead of an implication. The equality can then be used with `subst` on the entire function body term before you invoke `vcg`, and when `vcg` gets to the while loop it will see a correctly annotated one. That only works nicely when there is only one while loop in the body, but I think that is the case here (if there are is more than one loop, you need to instantiate the rule a bit more to select the right one).

proof/crefine/RISCV64/Recycle_C.thy
lsf37: Yes that is right. I didn't actually know that you could give these to `wpfix` directly as parameters :-) I thought you had to declare them globally. For `option` I thought we had done that already, but it might be in a theory further down or something like that. We should probably add an issue to set up `wpfix` more properly for all of the basic datatypes we might need it for (option, sum, etc).

But separately from this PR.

proof/crefine/RISCV64/Recycle_C.thy
lsf37: If there is no other `mask` in the goal, it's safe to unfold. However: somewhere in AArch64 I made a `simp` rule that gets rid of all of these automatically for thread states. We should probably increase the use of that one instead. It's [here](https://github.com/seL4/l4v/blob/53f44dea1e7763b94a98037629a3c608a9a06c44/proof/crefine/AARCH64/Wellformed_C.thy#L668-L671), it should copy over fine to MCS and other architectures.

lsf37 (approved):
Overall I'm quite happy with this, this was one of the functions I thought would be hardest in the whole IPC set.

We should remove the `vcg_step` by proving an annotation rule for While that can be used with `subst` and use that before `vcg`, but otherwise I'm happy for this to go ahead (modulo the small cleanup bits).

proof/crefine/RISCV64/Recycle_C.thy
michaelmcinerney: I did copy that lemma across to RISCV MCS, but I was wanting the version for `Inactive`, so I added that one, too. Did you want me to just have these for RISCV? Or should I put these into the other arches for MCS?

proof/crefine/RISCV64/Finalise_C.thy
michaelmcinerney: `reannotateWhileNoGuard` is about the entire `hoarep` statement, not just about rewriting the "while" bit, as we have in `whileLoop_add_inv`, so I wasn't sure how to use that. I did find a bit of inspiration somewhere else in CRefine after searching for `whileAnno`, and have come up with this, which at least avoids the use of `vcg_step`:

```
  "∀s. Γ ⊢⇘/UNIV⇙ {s}
       Call refill_unblock_check_'proc
       {s'. option_map2 tcbEPNext_C (cslift s') = option_map2 tcbEPNext_C (cslift s)
            ∧ option_map2 tcbEPPrev_C (cslift s') = option_map2 tcbEPPrev_C (cslift s)
            ∧ option_map2 tcbPriority_C (cslift s') = option_map2 tcbPriority_C (cslift s)
            ∧ option_map2 tcbDomain_C (cslift s') = option_map2 tcbDomain_C (cslift s)}"
  (is "∀s. Γ ⊢⇘/UNIV⇙ {s} ?f ({s'. ?post s s'})")
  apply (hoare_rule HoarePartial.ProcNoRec1)
  apply (clarsimp simp: whileAnno_def)
  apply (rule_tac I1="{t. ?post t s}" in subst[OF whileAnno_def])
  apply (vcg exspec=merge_overlapping_head_refill_tcb_fields exspec=refill_head_overlapping_tcb_fields)
    subgoal by (clarsimp simp: option_map2_def typ_heap_simps split: if_splits)
   apply fastforce
  apply fastforce
  done
```

The `vcg` step does inline a few things like `refill_index` and `isRoundRobin`, so it's not very fast. And it seems like my invariant does need to mention the `s` that became a metavariable after the clarsimp did the `allI` for the HOL forall. So maybe `rule_tac` is something is the best I can do here.

Does that seem reasonable? Or should there still be something nicer that can be done?

Xaphiosis: That looks reasonable to me.

proof/crefine/RISCV64/Finalise_C.thy
Xaphiosis: Yes, that's what I meant with the names. I use these rarely, so don't know what naming convention we want, but they are alternative spec rules, so I'd go with `spec` unless there's a pressing need to do otherwise.

Yeah, for `option_map2` definitely not in this PR, I only noticed it by coincidence and it looks very silly where it is.

proof/crefine/RISCV64/Recycle_C.thy
Xaphiosis: So if we both had no idea about it and assumed that it had proper setup, does that mean that this would work without the `add: option.sel`? Do you happen to know whether that works Michael? If not, we definitely should add it to the default setup of wpfix!

michaelmcinerney: Oh it can be removed from the `wpfix`! That's good. I'll go with that

proof/crefine/RISCV64/Recycle_C.thy
Xaphiosis: I think unfolding `mask_def` in general is not that great, but sometimes it's unavoidable, and as this is a terminal step (solves the subgoal), I have no idea, but it's not too bad. The `n=4` though is bizarre to read, and since you said you can remove it, thumbs up to doing that.

Regarding where to put it... that is a good question. Ultimately this will end up in all arches, but I'm a bit lost on how we're managing MCS given the one-arch push to finish CRefine. My suggestion: leave it on RISCV only for MCS, but backport to master in another PR, and then all the arches will pick that up in the next merge.

proof/crefine/RISCV64/Recycle_C.thy
Xaphiosis: If you didn't change it, then might as well leave it. For some reason I thought it was new.

michaelmcinerney: Alrighty. Often I do need to reindent parts of the old proof, and I think that's what happened here.

proof/crefine/RISCV64/Finalise_C.thy
lsf37: Yes, agreed, that looks fine. Nice trick with using the `subst` rule there. If you want to prevent unfolding in `vcg` you could provide spec rules for those functions -- it's possible that providing the modifies rules as spec rules is enough already.

Xaphiosis (approved):
Nice!
Still not quite a fan of the `function_spec_extra_info`, but am outvoted and I guess `function_spec2` is a kind of precedent. Don't forget to squash.

PR #817, opened 2024-09-09
Lemmas for simplifying masking and thread states

michaelmcinerney:


proof/crefine/X64/Wellformed_C.thy
```diff
@@ -614,6 +614,25 @@ abbreviation(input)
 where
   "prioInvalid == seL4_InvalidPrio"
 
+(* The magic 4 comes out of the bitfield generator -- this applies to all versions of the kernel. *)
+lemma
+  shows ThreadState_Restart_mask[simp]:
+  "(scast ThreadState_Restart::machine_word) && mask 4 = scast ThreadState_Restart"
```
Xaphiosis, line 620: nitpick: I'd much prefer `scast ThreadState_Restart :: machine_word` to the version with no spaces; the no-space version visually indicates a cast of ThreadState_Restart as if it were an `'a`, but I assume that what's intended is that `machine_word` is the target of the scast

Xaphiosis: by the way: is ThreadSTate_Restart coming from the C or the Haskell? I'm trying to think if there's a nice way to generate these lemmas in one go.

michaelmcinerney: These are all from the C, in `include/object/structures.h`. The Haskell has similar things, but doesn't use the `ThreadState` prefix in the names.

And due to an `ifdef` in the C, I forgot to include `IdleThreadState`. I'll add a fixup commit for that

michaelmcinerney:
This is causing tests to fail. I'll check this out tonight and hopefully push a fixup commit tonight or tomorrow

proof/crefine/X64/Wellformed_C.thy
Xaphiosis: Right, then there's probably no nice way to generate these lemmas. There might be some hack where we go from `x : set [ThreadState_Restart, ...] ==> (scast x :: machine_word) && mask 4 = scast x` and explode it, but I'm not sure how much nicer that is.

lsf37: I'm happy with the manual enumeration for now. The result is better than what we had before and if we can think of a nice way to make this shorter, we can still deploy it later without having to update the rest of the proofs.

lsf37 (approved):
There are still proof failures in ARM_HYP and another session, but the principle is good. I'd be happy with merging this once the failures are resolved.

Xaphiosis (approved):

corlewis (approved):

PR #819, opened 2024-09-24
Update specs to use semi-lazy FPU switching

corlewis:
These are the preliminary spec changes to seL4 to implement [RFC-18](https://github.com/seL4/rfcs/blob/main/src/active/0180-fpu-switching.md). They include extending the machine interface with new ops, adding the current owner of the FPU state to the kernel state, and adding the new seL4_TCB_Set_Flags syscall and related functions.


spec/machine/AARCH64/MachineOps.thy
```diff
@@ -178,31 +179,28 @@ definition getFPUState :: "fpu_state user_monad" where
...
+
+consts' writeFpuState_impl :: "fpu_state \<Rightarrow> unit machine_rest_monad"
+definition writeFpuState :: "fpu_state \<Rightarrow> unit machine_monad" where
+  "writeFpuState val \<equiv>  do
+     state_assert fpu_enabled;
+     machine_op_lift $ writeFpuState_impl val
+   od"
+
+definition enableFpu :: "unit machine_monad" where
+  "enableFpu \<equiv> modify (\<lambda>s. s\<lparr>fpu_enabled := True \<rparr>)"
```
Xaphiosis, line 197: Left over from internal review:

@Xaphiosis 

> the (lack of) link between `enableFpu` and `enableFpuEL01` confuses me... there's no machine op for disabling the FPU? Are `enableFpu` and `enableFpuEL01` called separately by the C, or is this only the abstract spec model?

@corlewis 

> I was confused also. In the C, `enableFpu` calls either `disableTrapFpu` or `enableFpuEL01`, depending on whether it is a configuration with hypervisor support (see https://github.com/seL4/seL4/blob/85aa104eb4d527bc9f0aac28f3e9c3a84eed1d7a/include/arch/arm/arch/64/mode/machine/fpu.h#L127). I was originally going to just have `enableFpu` in the spec before realising that `enableFpuEL01` is called directly from `vcpu_disable` (https://github.com/seL4/seL4/blob/85aa104eb4d527bc9f0aac28f3e9c3a84eed1d7a/include/arch/arm/armv/armv8-a/64/armv/vcpu.h#L702)

@Xaphiosis 

> That seems odd, because it is unconditionally enabling FPU for native threads with no VCPU on hyp configurations. Doesn't that effectively ignore the native thread's configured flags?

spec/haskell/src/SEL4/Object/TCB/AARCH64.hs
```diff
@@ -51,3 +53,12 @@ getSanitiseRegisterInfo t = do
 
 postModifyRegisters :: PPtr TCB -> PPtr TCB -> UserMonad ()
 postModifyRegisters _ _ = return ()
+
+postSetFlags :: PPtr TCB -> TcbFlags -> Kernel ()
+postSetFlags t flags =
+    when (tcbFlagToWord (ArchFlag FpuDisabled) .&. flags /= 0) (fpuRelease t)
```
Xaphiosis, line 59: interesting, with one flag I guess this is the shortest way to do it, but with more flags it'll be a bunch of `tcbFlagToWord`s `.|.`'ed together?

spec/haskell/src/SEL4/Object/TCB/AARCH64.hs
```diff
@@ -51,3 +53,12 @@ getSanitiseRegisterInfo t = do
...
+
+prepareSetDomain :: PPtr TCB -> Domain -> Kernel ()
+prepareSetDomain t newDom = do
```
Xaphiosis, line 62: for documentation purposes, might be worth saying why setting domain requires FPU release

spec/haskell/src/SEL4/Object/Structures.lhs
```diff
@@ -492,3 +496,15 @@ Various operations on the free index of an Untyped cap.
 > data TcbQueue = TcbQueue {
 >     tcbQueueHead :: Maybe (PPtr TCB),
 >     tcbQueueEnd :: Maybe (PPtr TCB) }
+
+\subsubsection{TCB Flags}
+
+> type TcbFlags = Word
+
+> data TcbFlag
+>     = NoFlag
```
Xaphiosis, line 505: I'm still not sure I understand the purpose of `NoFlag`, could you enlighten me?

spec/haskell/src/SEL4/Object/Structures/AARCH64.hs
```diff
@@ -70,6 +70,11 @@ atcbContextSet uc atcb = atcb { atcbContext = uc }
 atcbContextGet :: ArchTCB -> UserContext
 atcbContextGet = atcbContext
 
+data ArchTcbFlag = FpuDisabled
+
+archTcbFlagToWord :: ArchTcbFlag -> Word
+archTcbFlagToWord (FpuDisabled) = 0x1
```
Xaphiosis, line 76: No desire for `bit`-equivalent here?

spec/haskell/src/SEL4/Model/StateData/AARCH64.hs
```diff
@@ -13,6 +13,7 @@ module SEL4.Model.StateData.AARCH64 where
 import Prelude hiding (Word)
 import SEL4.Machine
 import SEL4.Machine.Hardware.AARCH64 (PTE(..), PT_Type, config_ARM_PA_SIZE_BITS_40)
+import SEL4.Object.Structures
```
Xaphiosis, line 16: I'm noticing some import hierarchy changes. TCB has some anti-circular-import magic. Just checking: do your changes to the Haskell compile?

spec/haskell/src/SEL4/Object/TCB/AARCH64.hs
corlewis: Hmm, good question, I hadn't thought about what this will look like if/when there are more flags. Maybe it's worth adding some sort of `flagSet` helper function that does the `tcbFlagToWord ` and `.&.`. That would make this something like 
```
when (flagSet (ArchFlag FpuDisabled) flags) (fpuRelease t)
```

spec/haskell/src/SEL4/Object/Structures.lhs
corlewis: I can't speak about the original intentions in the C, but in my mind here it provides a way to abstract away the details of how the flags are implemented. For instance, we could say `flags = NoFlag` to say that no flags are set.

Having now said that, this is another argument for adding something like the `flagSet` helper function I mentioned in a different comment, since that would extend this abstraction and hide the details of the flags being implemented as a `Word`.

spec/haskell/src/SEL4/Object/Structures/AARCH64.hs
corlewis: Good point, I missed that. Thinking about it now though, the equivalent in C produces values with the final mask, so at some point we'll need to do the conversion between that and the `bit` form. Do we think that that's simple enough that it can be done wherever, or would it be better do it in `Refine` instead of `CRefine`?

spec/haskell/src/SEL4/Model/StateData/AARCH64.hs
corlewis: Yeah, that anti-circular-import magic was a pain to relearn and get right. I've tested building it, it does compile.

spec/haskell/src/SEL4/Machine/Hardware/AARCH64.hs
```diff
@@ -387,11 +387,20 @@ debugPrint str = liftIO $ putStrLn str
...
-fpuThreadDeleteOp :: Word -> MachineMonad ()
-fpuThreadDeleteOp tcbPtr = error "Unimplemented callback"
+readFpuState :: MachineMonad AARCH64.FPUState
+readFpuState = error "Unimplemented - machine op"
+
+writeFpuState :: AARCH64.FPUState -> MachineMonad ()
+writeFpuState _ = error "Unimplemented - machine op"
+
+enableFpu :: MachineMonad ()
+enableFpu = error "Unimplemented - machine op"
```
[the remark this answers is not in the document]
corlewis: It was already defined earlier in this file.
https://github.com/seL4/l4v/blob/bfd342f980e0bd46eeede5cb318de9df2e24ee57/spec/haskell/src/SEL4/Machine/Hardware/AARCH64.hs#L238-L241

spec/haskell/src/SEL4/Object/Structures/AARCH64.hs
```diff
@@ -70,6 +70,11 @@ atcbContextSet uc atcb = atcb { atcbContext = uc }
 atcbContextGet :: ArchTCB -> UserContext
 atcbContextGet = atcbContext
 
+data ArchTcbFlag = FpuDisabled
```
Xaphiosis, line 73: in some cases when we have datatypes that are boringly similar to what we have in ASpec, we don't translate them and instead introduce an alias in the skeleton files; not sure if that's useful here

spec/haskell/src/SEL4/Object/TCB/AARCH64.hs
Xaphiosis: Something like that sounds good, but I'd prefer `isFlagSet` or something, otherwise it sounds like a flag-setter.

lsf37: Maybe `inFlagSet`?

spec/abstract/AARCH64/FPU_A.thy
```diff
@@ -0,0 +1,59 @@
...
+     fpu_state \<leftarrow> as_user t getFPUState;
+     do_machine_op (writeFpuState fpu_state)
+   od"
+
+\<comment> \<open>FIXME: maybe use a case instead of the if (depends on if wpc or if_split is easier)\<close>
+definition switch_local_fpu_owner :: "obj_ref option \<Rightarrow> (unit,'z::state_ext) s_monad" where
+  "switch_local_fpu_owner new_owner \<equiv> do
+     do_machine_op enableFpu;
+     cur_fpu_owner \<leftarrow> gets (arm_current_fpu_owner \<circ> arch_state);
+     when (cur_fpu_owner \<noteq> None) $ save_fpu_state (the cur_fpu_owner);
```
lsf37, line 33: I would generally prefer `case` for option if you have both branches, because it allows you to avoid using `the`.

spec/abstract/AARCH64/FPU_A.thy
```diff
@@ -0,0 +1,59 @@
...
+     if (new_owner \<noteq> None)
```
lsf37, line 34: no need for parentheses on `if` in Isabelle (too much C :-))

spec/abstract/AARCH64/FPU_A.thy
```diff
@@ -0,0 +1,59 @@
...
+definition fpu_release :: "obj_ref \<Rightarrow> (unit,'z::state_ext) s_monad" where
+  "fpu_release thread_ptr \<equiv> do
+     cur_fpu_owner \<leftarrow> gets (arm_current_fpu_owner \<circ> arch_state);
+     when (cur_fpu_owner = Some thread_ptr) $ switch_local_fpu_owner None
+   od"
+
+definition lazy_fpu_restore :: "obj_ref \<Rightarrow> (unit,'z::state_ext) s_monad" where
+  "lazy_fpu_restore thread_ptr \<equiv> do
+     flags \<leftarrow> thread_get tcb_flags thread_ptr;
+     if (ArchFlag FpuDisabled \<in> flags)
```
lsf37, line 49: no parentheses necessary

spec/abstract/Schedule_A.thy
```diff
@@ -166,37 +167,42 @@ end
...
-  od)"
+  "schedule \<equiv> (do
+     cur \<leftarrow> gets cur_thread;
+     cur_active \<leftarrow> gets (getActiveTCB cur);
+     idl \<leftarrow> gets idle_thread;
+     if cur_active \<noteq> None \<or> idl = cur then
+       return () \<sqinter> (do pre_choose_thread_unit; choose_thread_unit od)
+     else
+       (do pre_choose_thread_unit; choose_thread_unit od)
+   od)"
```
lsf37, line 199: don't need the outer parentheses

spec/haskell/src/SEL4/Object/Structures/AARCH64.hs
lsf37: I think that would make sense, also for any future flags. Not sure of there we have a good location to put these in for arch specific data types, but it's worth looking.

spec/machine/AARCH64/MachineOps.thy
```diff
@@ -178,31 +179,28 @@ definition getFPUState :: "fpu_state user_monad" where
...
-
-(* FIXME this machine op is used to abstract the entire lazy FPU switch interrupt mechanism,
-   which can only trigger when the current thread's FPU is disabled and it performs an FPU
-   operation. We have no model for this mechanism or the state that it caches, so for
-   verification purposes we act as if the FPU is always enabled.
-   Future lazy FPU switch overhaul will involve the state that this operation reads, at which
-   point it should become a normal function. *)
 definition isFpuEnable :: "bool machine_monad" where
-  "isFpuEnable \<equiv> return True"
+  "isFpuEnable \<equiv> gets fpu_enabled"
```
lsf37, line 203: Does anyone remember why this is called `isFpuEnable` in C and not `isFpuEnabled`?

lsf37 (approved):
Nice work! I only had cosmetic comments.

We should probably target this to a new branch to use until the proofs are done and the entire thing can be merged into master.

spec/machine/AARCH64/MachineOps.thy
Xaphiosis: Probably Englishes

spec/haskell/src/SEL4/Object/TCB/AARCH64.hs
corlewis: I like `isFlagSet`, to me it's more of an abstraction where we don't care how it's implemented while `inFlagSet` implies that it's a set.

spec/abstract/AARCH64/FPU_A.thy
corlewis: That comment was actually for the `if` statement just after this `when`, which does have both branches. I'll have a dig around for a helpful combinator. Assuming one does exist, would we want something similar in the Haskell?

spec/machine/AARCH64/MachineOps.thy
corlewis: Yeah, no idea beyond that. Should we give it the better name here?

spec/abstract/AARCH64/FPU_A.thy
lsf37: Yes, having them structurally the same would be nice. (It's not that super important, it's correct as it is, just a style question, so if we don't already have a combinator, we can also leave it as is.)

spec/haskell/src/SEL4/Object/TCB/AARCH64.hs
lsf37: `hasFlag`? I keep reading `isFlagSet` as "is this a flag set", not "is this flag set/unset".

Xaphiosis: Uhh, I really think that `isFlagSet` is pretty standard for "is the flag set", I'm not sure many people would think of `isFlagASet` when reading it, and I'm not even sure that's possible in Isabelle's type system, and `isSetOfFlags` is an even more creative interpretation...

spec/abstract/AARCH64/FPU_A.thy
Xaphiosis: Context: the FIXME was originally because we weren't sure whether dealing with `if` or `case` would be easier in the `corres` proof, so the plan was to check and then remove the FIXME afterwards

spec/machine/AARCH64/MachineOps.thy
Xaphiosis: If you want to, but then it would be good to also rename it in C.

spec/abstract/AARCH64/FPU_A.thy
corlewis: I've made this change now but it's a little bit more complicated in Haskell, as to the best of my knowledge there isn't an easily accessible equivalent to `maybeM`. The most likely option that I found is https://hackage.haskell.org/package/extra-1.7.16/docs/Control-Monad-Extra.html#v:whenJust, but using that would introduce a new dependency on the `extra` package. I've currently used the underlying `for` from `Data.Traversable`, but that feels a bit unreadable and will require some care when translating to the design spec. Thoughts?
https://github.com/seL4/l4v/blob/285ecfc1ee461291cdd239b28f6b7666203b2bc3/spec/haskell/src/SEL4/Object/FPU/AARCH64.hs#L33-L41

Xaphiosis: I don't have a good solution here. There's a function `maybe` in `Data.Maybe` that we could use with `maybe (return ()) saveFpuState curFpuOwner` that we could make an alias for to remove the `return ()` part, but then the return values have to match so it would only work for unit returns. Otherwise you could `return undefined` with the correct type (whatever `'b` the function returns). How does `for` get through the Haskell translator? I don't think we have something like that in Isabelle.

corlewis: Yeah, `for` won't get through the translator without doing something manual, so on second thought probably shouldn't use it.

I think I'm ok with your `maybe` idea, particularly since now that I look at it, that's basically how `whenJust` is implemented. It lines up pretty well with `maybeM` as well, which requires a unit return.

corlewis: > Context: the FIXME was originally because we weren't sure whether dealing with `if` or `case` would be easier in the `corres` proof, so the plan was to check and then remove the FIXME afterwards

Good point, and also works as a reminder that there's two things going on here. There's switching the `when` to a `maybeM` which I think we all agree is nice to avoid `the`, and there's whether it should `if` or `case`, where `case` is again nice in that it avoids `the` but it is unknown which will be easier in the proofs. How about for now I switch it to `case` but leave the FIXME until we get to the proofs?

corlewis:
This PR has now been updated with a commit making equivalent changes to the FPU handling code in the x64 specification, and a commit making the required changes to the architectures that are not configured to have FPUs enabled so that they can still be built.

spec/abstract/AARCH64/FPU_A.thy
lsf37: yes, I'd be happy with that

lsf37:
> This PR has now been updated with a commit making equivalent changes to the FPU handling code in the x64 specification, and a commit making the required changes to the architectures that are not configured to have FPUs enabled so that they can still be built.

Good to see all tests passing up to AInvs. Nice work!

corlewis:
This has been rebased and had a few additional changes included following the earlier discussions.

spec/machine/AARCH64/MachineOps.thy
corlewis: To answer this earlier question, our confusion was due to the fact that on AArch64 with hypervisor support there are two registers that control whether the FPU is enabled. The first, `CPACR_EL1`, is generally controlled by the VM and is used to enable or disable the FPU for virtualised threads. This register is the one being enabled by `enableFpuEL01`, which seL4 only calls when disabling the VCPU. For native threads seL4 instead uses `CPTR_EL2`, which it modifies with `enableFpu` and `disableFpu`.

spec/haskell/src/SEL4/Object/TCB/AARCH64.hs
corlewis: I still prefer `isFlagSet` out of these options. @lsf37 have a look at https://github.com/seL4/l4v/pull/819/commits/52a23eb5c4298aa8a6a6295975cd62c8b64050fc and see whether seeing it in context looks better to you or not.

spec/haskell/src/SEL4/Object/Structures/AARCH64.hs
Xaphiosis: No idea. It should simplify out, just look more sensible in Haskell, especially if we get future flags.

spec/haskell/src/SEL4/Object/Structures/AARCH64.hs
Xaphiosis: What's the outcome on this @corlewis ?

Xaphiosis (approved):
I think we're mostly ok, one or two comment chains did not resolve.

On the commits, why is "x64 spec: update specifications for semi-lazy FPU changes" not merged in with the others? It's a bit confusing, there's AARCH64 work split up across spec levels, then x64 not split up, and then everything else.

Also, subjective suggestion: there's no need to start commit comments with "This <verb>", i.e. can say "Includes" instead of "This includes". Maybe we can also teach this novel idea to others :)

corlewis:
> On the commits, why is "x64 spec: update specifications for semi-lazy FPU changes" not merged in with the others? It's a bit confusing, there's AARCH64 work split up across spec levels, then x64 not split up, and then everything else.

The X64 changes were kept separate because it has FPU enabled, the other architectures don't and their changes are mostly trivial to match the modified interface.

I think the commit structure was just due to the way in which this work was developed; first for AArch64, which included changes to generic files, then copying required arch-specific changes to X64, and then doing required arch-specific changes for other architectures that don't support FPU.

I have no strong preference here and see two obvious options, would you prefer something like the current structure but with the two AArch64 commits combined, or to make it all one big commit for all architectures?

Xaphiosis:
> > On the commits, why is "x64 spec: update specifications for semi-lazy FPU changes" not merged in with the others? It's a bit confusing, there's AARCH64 work split up across spec levels, then x64 not split up, and then everything else.
> 
> The X64 changes were kept separate because it has FPU enabled, the other architectures don't and their changes are mostly trivial to match the modified interface.
> 
> I think the commit structure was just due to the way in which this work was developed; first for AArch64, which included changes to generic files, then copying required arch-specific changes to X64, and then doing required arch-specific changes for other architectures that don't support FPU.
> 
> I have no strong preference here and see two obvious options, would you prefer something like the current structure but with the two AArch64 commits combined, or to make it all one big commit for all architectures?

The question is whether you believe there is beneficial information in the extra commits. In my mind the aarch64+x64 are together because of FPU, and then the rest are not very interesting probably. Since it's all reviewed now, probably can squash them together, add a note in the commit message.

spec/haskell/src/SEL4/Object/Structures/AARCH64.hs
corlewis: I agree that it should simplify out, I've made this change now.

spec/haskell/src/SEL4/Object/Structures/AARCH64.hs
corlewis: The abstract and exec specs now share this datatype, along with `archTcbFlagToWord` and the generic equivalents.

corlewis:
> The question is whether you believe there is beneficial information in the extra commits. In my mind the aarch64+x64 are together because of FPU, and then the rest are not very interesting probably. Since it's all reviewed now, probably can squash them together, add a note in the commit message.

I think there's a little bit of use in keeping the aarch64 and x64 separate from the less interesting architectures, so I've now squashed things into two commits. Have a look and let me know what you think

spec/design/skel/Structs_B.thy
```diff
@@ -0,0 +1,24 @@
+(*
+ * Copyright 2024, Proofcraft Pty Ltd
+ *
+ * SPDX-License-Identifier: GPL-2.0-only
+ *)
+
+(* Generic data types shared by spec and abstract. *)
+
+chapter "Common, Generic Data Types"
```
Xaphiosis, line 9: I don't want to nitpick too much, but reading these now...

- "Architecture-specific data types shared by spec and abstract."
- "Generic data types shared by spec and abstract."

Um, between spec and abstract? That makes no sense. "design and abstract specs"?

The "Common" in the chapter titles also takes a moment to think through, and only clicked when I read the comments above... but it probably makes sense to keep that.

spec/design/skel/Structs_B.thy
```diff
@@ -0,0 +1,24 @@
...
+imports Arch_Structs_B
+begin
+
+arch_requalify_types (H)
+  arch_tcb_flag
+
+arch_requalify_consts (H)
+  archTcbFlagToWord
+
+#INCLUDE_SETTINGS keep_constructor = tcb_flag
```
Xaphiosis, line 21: Do you happen to know why we tell it to keep constructors for `tcb_flag` when we're importing `TcbFlag`? Does a `tcb_flag` actually exist anywhere? It doesn't seem the most easily graspable interface, I'm glad you managed to figure it out.

spec/haskell/src/SEL4/Object/Structures.lhs
```diff
@@ -501,12 +501,12 @@ Various operations on the free index of an Untyped cap.
 
 > type TcbFlags = Word
 
-> data TcbFlag
->     = NoFlag
->     | ArchFlag ArchTcbFlag
+> noFlag :: TcbFlags
+> noFlag = 0x0
```
Xaphiosis, line 505: noFlag seems unused, but I guess it has applicability in generic proofs later?

The TcbFlag/TcbFlags distiction is not immediately obvious, perhaps a comment above `TcbFlags` beyond it being a `Word`? Did I miss comments explaining any of this?

spec/haskell/src/SEL4/Object/Structures/AARCH64.hs
Xaphiosis: Very nice. I like this approach, and it made sense to make a generic Structs_B

Xaphiosis:
> > The question is whether you believe there is beneficial information in the extra commits. In my mind the aarch64+x64 are together because of FPU, and then the rest are not very interesting probably. Since it's all reviewed now, probably can squash them together, add a note in the commit message.
> 
> I think there's a little bit of use in keeping aarch64 and x64 separate from the less interesting architectures, so I've now squashed things into two commits. Have a look and let me know what you think

Commits make sense now, modulo squashes. I had one last bikeshed about comments and noFlag, but otherwise this is good to go from my side.

spec/design/skel/Structs_B.thy
corlewis: Agreed that your suggestion makes more sense. I didn't think about it much before and just mindlessly copied the existing comment from `Arch_Structs_B.thy`

spec/design/skel/Structs_B.thy
corlewis: Another case of the translator being magic; `TcbFlag` is quietly translated to `tcb_flag`.

spec/haskell/src/SEL4/Object/Structures.lhs
corlewis: Yeah, `noFlag` isn't currently used but I can imagine uses for it in the future and it better matches the C so I thought it best to include it now.

I've added a comment for `TcbFlags`, let me know if it makes sense and helps.

Xaphiosis:
All done now from my side.

PR #820, opened 2024-09-28
make ARM+ARM_HYP proofs work for smaller irq_len 

lsf37:
- Remove remaining use of direct numeral related to `irq_len`. The lemma happened to work for `irq_len >= 8`, but not for smaller `irq_len`.

- Improve sys-init example: the example assumes an irq type of at least 8 bits. Some AArch32 boards only use 7 bit. Use smaller numbers in the example so that the example happens to work on those boards as well.

### DIFF
--- a/proof/crefine/ARM/Interrupt_C.thy
+++ b/proof/crefine/ARM/Interrupt_C.thy
@@ -299,7 +299,9 @@ lemma invokeIRQControl_expanded_ccorres:
   apply (rule conjI, fastforce)+
   apply (clarsimp simp: Collect_const_mem ccap_relation_def cap_irq_handler_cap_lift
                         cap_to_H_def c_valid_cap_def cl_valid_cap_def irq_len_val ucast_and_mask)
-  apply word_eqI
+  (* if irq_len < 8, there may be an additional term in the goal *)
+  apply (rule conjI, word_eqI_solve)?
+  apply (word_eqI_solve dest: bit_imp_le_length)
   done
 
 lemma invokeIRQControl_ccorres:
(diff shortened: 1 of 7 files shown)


proof/crefine/ARM/Interrupt_C.thy
```diff
@@ -299,7 +299,9 @@ lemma invokeIRQControl_expanded_ccorres:
   apply (rule conjI, fastforce)+
   apply (clarsimp simp: Collect_const_mem ccap_relation_def cap_irq_handler_cap_lift
                         cap_to_H_def c_valid_cap_def cl_valid_cap_def irq_len_val ucast_and_mask)
-  apply word_eqI
+  (* if irq_len < 8, there may be an additional term in the goal *)
+  apply (rule conjI, word_eqI_solve)?
+  apply (word_eqI_solve dest: bit_imp_le_length)
```
corlewis, line 304: Oh, this is a clever way to handle this. Thanks for the comment, it would have been very confusing without it

proof/crefine/ARM/Refine_C.thy
```diff
@@ -36,13 +36,14 @@ lemma schedule_sch_act_wf:
...
-(* On GICv2 and GICv3, irqInvalid is picked such it is outside the range of possible IRQs
+(* On Arm, irqInvalid is picked such it is outside the range of possible IRQs
    by type alone. *)
 lemma irq_type_never_invalid_left:
   "ucast irq \<noteq> irqInvalid" for irq::irq
   unfolding irqInvalid_def
   using ucast_less[of irq, where 'a=machine_word_len]
-  by (clarsimp simp: mask_def)
+  apply (clarsimp simp: mask_def)
+  done
```
corlewis, line 46: I assume it's been too long, but any memory of why you made this change?

corlewis (approved):
Sorry for missing this PR, looks good to me!

proof/crefine/ARM/Refine_C.thy
lsf37: That probably was just experimentation while I was writing it. Will revert.

PR #821, opened 2024-10-02
`refill_budget_check_ccorres`

michaelmcinerney:


lib/Word_Lib/Word_32.thy
```diff
@@ -172,6 +172,10 @@ lemma word32_and_max_simp:
   using word_and_full_mask_simp [of x]
   by (simp add: numeral_eq_Suc mask_Suc_exp)
 
+lemma max_minus_one_word32:
```
lsf37, line 175: We should name both of them just `max_minus_one_word`, so that we can use the same name in 32 and 64 contexts (maybe with another postfix for indicating that it is for fixed lengths, not sure).

The theory name can be used to specify a specific version if that is needed.

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -11810,34 +11800,34 @@ lemma set_refills_bounded_release_time:
   apply (wpsimp wp: valid_sched_wp)
   by (clarsimp simp: vs_all_heap_simps bounded_release_time_def pred_map_simps obj_at_def
                     heap_upd_def split: if_split)
-
+thm refill_budget_check_def
```
michaelmcinerney, line 11803: Remove this `thm`

proof/invariant-abstract/RealTime_AI.thy
```diff
@@ -1138,11 +1158,27 @@ lemma refill_pop_head_no_fail[wp]:
...
-            (non_overlapping_merge_refills sc_ptr)"
-  unfolding non_overlapping_merge_refills_def refill_pop_head_def
-  by (wpsimp simp: update_refill_hd_def wp: update_sched_context_wp)
+            (set_refill_hd sc_ptr f)"
+  unfolding set_refill_hd_def refill_pop_head_def
+  by wpsimp
+
+lemma merge_nonoverlapping_head_refill_no_fail[wp]:
+  "no_fail (\<lambda>s. sc_refills_sc_at (\<lambda>refills. 1 < length refills) sc_ptr s)
+            (merge_nonoverlapping_head_refill sc_ptr)"
```
Xaphiosis, line 1176: indent looks off here, should align with `(` on prev line. same for `update_refill_hd_no_fail`

proof/invariant-abstract/RealTime_AI.thy
```diff
@@ -1138,11 +1158,27 @@ lemma refill_pop_head_no_fail[wp]:
   unfolding refill_pop_head_def
   by wpsimp
 
-lemma non_overlapping_merge_refills_no_fail[wp]:
+lemma update_refill_hd_no_fail[wp]:
+  "no_fail (\<lambda>s. sc_refills_sc_at (\<lambda>refills. refills \<noteq> []) sc_ptr s)
+            (update_refill_hd sc_ptr f)"
+  unfolding update_refill_hd_def refill_pop_head_def
+  apply wpsimp
+  by (clarsimp simp: obj_at_def sc_at_pred_n_def)
```
Xaphiosis, line 1166: you can combine two-steps into a single `by`

proof/invariant-abstract/RealTime_AI.thy
```diff
@@ -1104,6 +1114,16 @@ lemma no_ofail_read_refill_head[wp]:
...
 
+lemma no_ofail_read_refill_tail[wp]:
+  "no_ofail (\<lambda>s. sc_refills_sc_at (\<lambda>refills. refills \<noteq> []) scp s) (read_refill_tail scp)"
+  unfolding read_refill_tail_def
+  apply wpsimp
+  apply (clarsimp simp: sc_at_pred_n_def obj_at_def)
+  done
+
+lemmas no_fail_get_refill_tail[wp] =
+  no_ofail_gets_the[OF no_ofail_read_refill_tail, simplified get_refill_tail_def[symmetric]]
```
Xaphiosis, line 1125: `, folded get_refill_tail_def]` might work here

proof/invariant-abstract/RealTime_AI.thy
```diff
@@ -1104,6 +1114,16 @@ lemma no_ofail_read_refill_head[wp]:
 lemmas no_fail_get_refill_head[wp] =
   no_ofail_gets_the[OF no_ofail_read_refill_head, simplified get_refill_head_def[symmetric]]
 
+lemma no_ofail_read_refill_tail[wp]:
+  "no_ofail (\<lambda>s. sc_refills_sc_at (\<lambda>refills. refills \<noteq> []) scp s) (read_refill_tail scp)"
+  unfolding read_refill_tail_def
+  apply wpsimp
+  apply (clarsimp simp: sc_at_pred_n_def obj_at_def)
```
Xaphiosis, line 1121: two-step -> can use `by`

proof/invariant-abstract/RealTime_AI.thy
```diff
@@ -1050,6 +1050,16 @@ lemma read_refill_head_wp[wp]:
...
 
+lemma read_refill_tail_wp[wp]:
+  "\<lblot>\<lambda>s. \<forall>sc n. kheap s sc_ptr = Some (SchedContext sc n) \<longrightarrow> Q (refill_tl sc) s\<rblot>
+   read_refill_tail sc_ptr
+   \<lblot>Q\<rblot>"
+  unfolding read_refill_tail_def
+  by wpsimp
+
+lemmas get_refill_tail_wp[wp] =
+  ovalid_gets_the[OF read_refill_tail_wp, simplified get_refill_tail_def[symmetric]]
```
Xaphiosis, line 1061: also a chance `folded` will work here

proof/refine/RISCV64/Schedule_R.thy
```diff
@@ -4192,90 +4025,87 @@ lemma getRefillFull_sp:
...
+  "sc_ptr = scPtr \<Longrightarrow>
+   corres (\<lambda>rv rv'. refill_map rv' = rv)
+     (valid_objs and pspace_aligned and pspace_distinct
+      and is_active_sc sc_ptr and sc_refills_sc_at (\<lambda>refills. refills \<noteq> []) sc_ptr)
+     valid_objs'
+     (get_refill_tail sc_ptr) (getRefillTail scPtr)"
+  apply (clarsimp simp: get_refill_tail_def getRefillTail_def read_refill_tail_def
+                        readRefillTail_def getSchedContext_def[symmetric]
+                        read_sched_context_get_sched_context
+                        readSchedContext_def getObject_def[symmetric])
```
Xaphiosis, line 4047: dunno if this is from older proofs, but checking that you know you can do `clarsimp simp: ... simp flip: ...` to not need the `[symmetric]`

proof/crefine/RISCV64/Schedule_C.thy
```diff
@@ -593,35 +593,957 @@ lemma length_scRefills_bounded:
...
+         Collect_const [simp del]
+  apply (clarsimp simp: headRefillOverrun_def getRefillHead_def[symmetric])
+  apply (cinit' lift: sc_' usage_' simp: max_minus_one_word64)
+   apply (rule_tac xf'="\<lambda>s. h_val (hrs_mem (t_hrs_' (globals s))) (ret__ptr_to_struct_refill_C_' s)"
+                in ccorres_split_nothrow)
+       apply (rule ccorres_call)
+          apply (rule refill_head_ccorres)
+         apply clarsimp
+        apply clarsimp
+       apply clarsimp
```
Xaphiosis, line 616: `clarsimp+`

proof/crefine/RISCV64/Schedule_C.thy
```diff
@@ -593,35 +593,957 @@ lemma length_scRefills_bounded:
...
+                       simp: typ_heap_simps' sc_ptr_to_crefill_ptr_def crefill_relation_def)
+     apply (ctac add: refill_full_ccorres)
+       apply (rule ccorres_cond[where R=\<top>])
+         apply(clarsimp simp: to_bool_def)
+        apply (ctac add: refill_add_tail_ccorres)
+       apply (rule ccorres_rhs_assoc)
+       apply (rule ccorres_split_nothrow)
+           apply (clarsimp simp: setRefillTl_def updateRefillTl_def)
+           apply (rule updateSchedContext_ccorres_lemma3
+                        [where P="\<lambda>sc. scRefillTail sc < length (scRefills sc)" and P'="?abs"])
```
Xaphiosis, line 808: `[` should be two spaces deeper than `u`, am seeing one on the github diff (saw this above as well in same file)

proof/crefine/RISCV64/Schedule_C.thy
```diff
@@ -593,35 +593,957 @@ lemma length_scRefills_bounded:
...
+           apply (clarsimp simp: crefill_relation_def csched_context_relation_def)
+           apply (frule rf_sr_refill_buffer_relation)
+           apply (frule h_t_valid_clift_Some_iff[THEN iffD1])
+           apply (clarsimp cong: StateSpace.state.fold_congs)
+           apply (rule_tac old_sc=sc and n="scRefillTail sc"
+                           and fa="rTime_update (\<lambda>_. rTime new - rAmount rv)"
+                            in rf_sr_refill_update2; fastforce?)
+              apply (fastforce simp: sched_context.expand)
+             apply (clarsimp simp: typ_heap_simps)
+             apply (fastforce simp: typ_heap_simps' sc_ptr_to_crefill_ptr_def crefill_relation_def)
```
Xaphiosis, line 832: unusual having typ_heap_simps in two commands like this; is this proof copied?

proof/crefine/RISCV64/Schedule_C.thy
```diff
@@ -593,35 +593,957 @@ lemma length_scRefills_bounded:
...
+                    apply (intro conjI)
+                       apply (clarsimp simp: typ_heap_simps)
+                      apply (rule disjCI2)
+                      apply (rule_tac n="length (scRefills sc)" in array_assertion_shrink_right)
+                       apply (fastforce intro: sc_at_array_assertion')
+                      apply (clarsimp simp: typ_heap_simps csched_context_relation_def)
+                     apply (clarsimp simp: typ_heap_simps' sc_ptr_to_crefill_ptr_def
+                                           csched_context_relation_def)
+                    apply (clarsimp simp: typ_heap_simps csched_context_relation_def)
+                    apply normalise_obj_at'
```
Xaphiosis, line 1414: unexpected to see `normalise_obj_at'` so distant from the subgoal-solving statement, anything interesting going on herer?

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -4531,40 +4531,67 @@ lemma refill_budget_check_round_robin_cur_sc_active[wp]:
   "refill_budget_check_round_robin consumed \<lbrace>cur_sc_active\<rbrace>"
   by (rule cur_sc_active_lift; wpsimp)
 
-method handle_overrun_loop_body_simple
-  = ((clarsimp simp: handle_overrun_loop_body_def)?
-     , wpsimp wp: whileLoop_valid_inv set_refills_wp get_refills_wp
-            simp: refill_pop_head_def vs_all_heap_simps obj_at_def refill_single_def refill_size_def
-                  update_sched_context_set_refills_rewrite update_refill_hd_rewrite
-                  schedule_used_defs)
+method charge_entire_head_refill_simple
```
Xaphiosis, line 4534: are these methods used in multiple places? they seem highly specific, so if they are only used in one place, `local_method` might be better

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -4531,40 +4531,67 @@ lemma refill_budget_check_round_robin_cur_sc_active[wp]:
...
+               schedule_used_defs sc_at_ppred_def
 
-method handle_overrun_loop_simple
-  = ((clarsimp simp: handle_overrun_loop_def)?
-     , wpsimp wp: whileLoop_valid_inv set_refills_wp get_refills_wp
-            simp: handle_overrun_loop_body_def merge_overlapping_refills_def round_robin_def refill_pop_head_def
-                  sc_valid_refills_def vs_all_heap_simps obj_at_def refill_budget_check_defs
-                  update_sched_context_set_refills_rewrite schedule_used_defs)
+method handle_overrun_simple
+  = (clarsimp simp: handle_overrun_def)?,
```
Xaphiosis, line 4542: this looks nearly identical to the previous method, am I missing why we need two methods?

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -8404,98 +8431,57 @@ lemma refill_update_is_refill_sufficient[wp]:
...
-  apply (case_tac "sc_refills sc"; simp?)
-  apply (rename_tac list)
-  apply (subst unat_add_lem')
-   apply (prop_tac "unat (r_amount (hd list)) \<le> sum_list (map unat (map r_amount list))")
-    apply (fastforce intro: member_le_sum_list)
-   apply (clarsimp simp: unat_minus_one_word)
-  apply (case_tac list; simp)
-  done
+method merge_nonoverlapping_head_refill_simple
+  = clarsimp simp: merge_nonoverlapping_head_refill_def refill_pop_head_def merge_refill_def
```
Xaphiosis, line 8468: again consider `local_method`, I think this is only used once

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -10611,86 +10628,87 @@ lemma refill_budget_check_valid_refills[wp]:
...
-   apply (wpsimp wp: refill_budget_check_refills_unat_sum_helper, fastforce+)
-  apply (clarsimp simp: obj_at_def)
-  apply (wpsimp wp: set_refills_wp get_refills_wp
-              simp: vs_all_heap_simps update_sched_context_set_refills_rewrite update_refill_hd_rewrite)+
+           apply (find_goal \<open>match conclusion in "\<lbrace>_\<rbrace> head_insufficient_loop _ \<lbrace>\<lambda>_. _\<rbrace>" \<Rightarrow> \<open>-\<close>\<close>)
+           apply (rule_tac Q'="\<lambda>_ s. \<not> round_robin sc_ptr s \<and> sp_valid_refills_unbundled sc_ptr s"
+                        in hoare_strengthen_post[rotated])
+            apply (fastforce simp: valid_refills_def vs_all_heap_simps)
+           apply (intro hoare_vcg_conj_lift_pre_fix; (solves head_insufficient_loop_simple)?)
+                 apply (wpsimp wp: head_insufficient_loop_refills_sum)
```
Xaphiosis, line 10682: this looks like you can combine the rest of the proof into a `(wpsimp wp: ... | clarsimp simp: ...)+`, I don't think there's much of interest here for the proof length

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -11211,12 +11203,13 @@ lemma is_refill_ready_alt:
   by (clarsimp simp: refill_ready_def vs_all_heap_simps)
 
 method handle_overrun_loop_fail_simple
-  = (clarsimp simp: handle_overrun_loop_def
-     , rule valid_whileLoop_cond_fail
-     , (fastforce simp: head_time_buffer_def obind_def ogets_def read_sched_context_def
-                        vs_all_heap_simps refill_sufficient_def refill_capacity_def
-                        MIN_BUDGET_nonzero sc_valid_refills_def vs_all_heap_simps
-                 split: option.splits kernel_object.splits if_splits)+)
+  = clarsimp simp: handle_overrun_def,
```
Xaphiosis, line 11206: again, not sure about these / `local_method` / folding into other methods;
the methods namespace is global, i.e. `print_methods` will print all these methods along with the others

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -11364,43 +11351,49 @@ lemma refill_budget_check_is_refill_sufficient:
...
-   apply (clarsimp simp: handle_overrun_loop_def)
-   apply (wpsimp wp: valid_whileLoop[where I="\<lambda>_ s. pred_map \<top> (scs_of s) (cur_sc s)"])
-    apply (fastforce simp: head_time_buffer_true_imp_unat_buffer vs_all_heap_simps word_less_nat_alt
-                           word_le_nat_alt)
-   apply (clarsimp simp: vs_all_heap_simps)
-
-  apply (rule bind_wp[OF _ get_refills_sp])
+   apply (clarsimp simp: handle_overrun_def)
+   apply (wpsimp wp: valid_whileLoop
+                      [where I="\<lambda>_ s. sc_refills_sc_at (\<lambda>refills. refills \<noteq> []) sc_ptr s"])
```
Xaphiosis, line 11369: indent on `[` again

Xaphiosis (approved):
Nice work. That was a much bigger change than I expected. I have minor/cleanup comments mostly.
On the "prove" commit, you're changing the spec, would be nice to have some more info, because the commit diff is huge, but the message makes it sound like something rather simple.

proof/crefine/RISCV64/Schedule_C.thy
michaelmcinerney: Is there any particular reason that we would want two spaces here, rather than one? I thought that as long as it's strictly to the right of the name of the rule, it would be fine.

proof/crefine/RISCV64/Schedule_C.thy
michaelmcinerney: The `clarsimp` uses `typ_heap_simps` and then the `fastforce` uses `typ_heap_simps'`. I'm quite sure that neither can be removed, and I think I have also tried without success before to condense it into one command. It would be cool to see whether this could be improved, or to properly understand what's going on, but I don't think I could figure it out before

michaelmcinerney: And yes the proof was copied, and it appears like this in a few places, now

proof/crefine/RISCV64/Schedule_C.thy
michaelmcinerney: Maybe some of these `normalise_obj_at'` can be removed or moved elsewhere. Maybe I don't understand what you mean by it being distant from the subgoal-solving statement. Were you expecting it to be further up or further down?

proof/invariant-abstract/DetSchedSchedule_AI.thy
michaelmcinerney: Yes these `_simple` methods were quite handy, actually. I at first wanted to remove them, but there were quite a few places where they're used. Handy, but annoying, because I did have to go through quite a few iterations before I found something that would work everywhere it was being used. There might be a bit more that I could do to optimise these, but considering how long it takes to process this file, it's very painful to do

proof/invariant-abstract/DetSchedSchedule_AI.thy
michaelmcinerney: They are slightly different. It's possible that they could be combined into one method that could be used instead of the two, but it could be rather annoying to try this.

proof/crefine/RISCV64/Schedule_C.thy
Xaphiosis: as discussed offline: indent is nearly always by 2 spaces, except for subgoal-indentation where it's two spaces plus number of subgoals - 1
Also, there are other issues (the modification belongs to the lemma not to `rule`);
it might be better to have `..._lemma3[` to indicate the lemma is being modified, then wrap how it's modified onto the next line, with the `where` 2 in from `updateSched...`. Might be a tight fit still

proof/crefine/RISCV64/Schedule_C.thy
Xaphiosis: We can leave it then. If it were new, would be worth more investigation.

proof/crefine/RISCV64/Schedule_C.thy
Xaphiosis: Well, normally I see it used, and then the subgoal is discharged immediately, or on the next step. This one goes on for a bit until it's actually discharged, making me wonder if something is going the long way around.

proof/invariant-abstract/DetSchedSchedule_AI.thy
Xaphiosis: I guess take a quick look, or maybe examine it when you're next there.
Not sure if it's clear what I meant by local method, I specifically meant a tool Ed hacked up, `supply_local_method`, whose uses you can see in `Local_Method_Tests.thy` in lib, and `proof/crefine/ARM_HYP/Arch_C.thy` ... did that one ever make an appearance in MCS?

proof/invariant-abstract/DetSchedSchedule_AI.thy
michaelmcinerney: This is used 10 times, and I think the other `_simple` methods are used at least as much, so I think `supply_local_method` won't be suitable. I will see if I can consolidate them a bit.

proof/crefine/RISCV64/Schedule_C.thy
michaelmcinerney: I think in these cases, it's nicest to just have it all on the one line, even though it does go over the 100 character blue line a little bit (by 5 characters or so). If that definitely doesn't sound good, then I can go with your suggestion.

proof/refine/RISCV64/Schedule_R.thy
```diff
@@ -4192,90 +4025,87 @@ lemma getRefillFull_sp:
...
+  apply (rule stronger_corres_guard_imp)
+    apply (rule corres_split[OF get_sc_corres])
+      apply (rule corres_assert_assume_l)
+      apply clarsimp
+      apply (rule refills_tl_equal[symmetric])
+       apply simp
+      apply simp
+     apply wpsimp
+    apply wpsimp
```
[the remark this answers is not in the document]
michaelmcinerney: Couldn't just do `wpsimp+`, but can do `simp+` then `wpsimp+`

lsf37 (approved):
Yup, this is a lot larger than the title might lead one to suspect. Good work! Ready to merge from my side.

PR #822, opened 2024-10-08
Small IPC lemmas

michaelmcinerney:


proof/crefine/RISCV64/Ipc_C.thy
```diff
@@ -6082,6 +6082,18 @@ lemma schedContext_resume_ccorres:
                          to_bool_def)
   done
 
+lemma obj_at_cslift_ntfn:
+  fixes P :: "notification \<Rightarrow> bool"
+  shows "\<lbrakk>obj_at' P ntfn s; (s, s') \<in> rf_sr\<rbrakk> \<Longrightarrow>
+  \<exists>ko ko'. ko_at' ko ntfn s \<and> P ko \<and>
+        cslift s' (Ptr ntfn) = Some ko' \<and>
+        cnotification_relation (cslift s') ko ko'"
```
lsf37, line 6090: Do you actually need the type annotation/`fixes`? I would have thought the `cnotification_relation` would constrain `ko`, which then should constrain `P`.

michaelmcinerney: I can check this. This lemma started off life as a copy paste of the analogous lemma for TCBs and was then modified. That’s where the fixes and type annotations would be from

proof/crefine/RISCV64/Ipc_C.thy
```diff
@@ -6018,9 +6018,70 @@ lemma receiveIPC_dequeue_ccorres_helper:
...
+          apply (ctac add: refill_sufficient_ccorres)
+            apply csymbr
+            apply (rule ccorres_cond[where R=\<top>])
+              apply (clarsimp simp: to_bool_def)
+             apply (ctac add: postpone_ccorres)
+            apply (rule ccorres_return_Skip)
+           apply wpsimp
+          apply (vcg exspec=refill_sufficient_modifies)
+         apply (rule ccorres_cond_seq)
+         apply (rule ccorres_cond_false)
```
Xaphiosis, line 6064: when you do the `ccorres_cond_seq` + `_true` or `_false` combo, you're relying on someone having the proof state loaded in front of them. I much prefer having comments saying what is going on, because I have no idea whatsoever what branches this proof is following

proof/crefine/RISCV64/Ipc_C.thy
```diff
@@ -6097,9 +6097,90 @@ lemma obj_at_cslift_ntfn:
...
+         apply (rule conseqPre, vcg)
+         apply normalise_obj_at'
+         apply (frule (1) sc_ko_at_valid_objs_valid_sc')
+         apply (frule (1) obj_at_cslift_sc)
+         apply (force dest!: tcb_at_not_NULL
+                       simp: typ_heap_simps' option_to_ctcb_ptr_def csched_context_relation_def
+                             valid_sched_context'_def from_bool_def
+                      split: option.splits bool.splits)
+        apply ceqv
+       apply (rule ccorres_cond[where R=\<top>])
```
Xaphiosis, line 6161: I don't like how we've reverted to not having any comments in CRefine. I want to know, what are branches we're looking at...

proof/crefine/RISCV64/Ipc_C.thy
```diff
@@ -6302,9 +6302,63 @@ lemma schedContext_bindTCB_ccorres:
...
+           apply clarsimp
+           apply (intro conjI)
+              apply (rule cpspace_relation_ntfn_update_ntfn, assumption+)
+               subgoal
+                 by (fastforce intro!: obj_at'_is_canonical
+                                 simp: cnotification_relation_def Let_def
+                                       sign_extend_canonical_address
+                                split: ntfn.splits)
+               apply fastforce
+             apply (simp add: refill_buffer_relation_def image_def dom_def Let_def typ_heap_simps
```
Xaphiosis, line 6339: there's a good chance some/all of these simps can be compressed

Xaphiosis (approved):
I think generally this is good, but I'm noticing a regression to the mean of not adding comments to guide the reader and future repairer, especially in CRefine, which concerns me a bit. I have too much joyous experience with proofs that break, and then I when I have to fix them I don't know what they were trying to do since whoever wrote them depended on the implicit state to inform the reader.

proof/crefine/RISCV64/Ipc_C.thy
michaelmcinerney: You were right, the `fixes` can be removed, and I've removed the `shows` too so it's just in the usual format

michaelmcinerney:
Force pushed after rebase.

And now with some comments to guide the reader at various branching points of `ccorres` rules. I'm not sure whether these are helpful. Do let me know if they are not and how I might improve things.

lsf37 (approved):
Thanks for updating the proof with those branch comments. They are helpful. There is one occurrence left where Raf had queried that doesn't have one yet and where it is indeed unclear from just the text what is roughly going on in the proof.

Other than that, I think this is good to go after squashing in the review commits.

proof/crefine/RISCV64/Ipc_C.thy
michaelmcinerney: I thought Raf's concern was mostly with the `corres_cond_true/false` but sure, I'll add a comment at every branching point.

lsf37: Doesn't have to be every branching point, just the ones where it looks like something is going on. I.e. where the proof makes an assumption about which branch it is in, or where there is something like this `R=\top` that looks like something is going on that is not immediately obvious.

PR #824, opened 2024-10-28
Refactor det_ext to remove scheduler state

corlewis:
The scheduler is now fully part of the extensible abstract specification, because we are increasingly finding properties that are not able to be reasonably specified without it. This means that there is now no non-deterministic scheduler specification and that the only slots for non-determinism are now CDT and preemption operations.

This came up due to some properties currently being proved not being true for all execution paths of the previous non-deterministic specification. Options that were considered were:
1 - move the minimal state (the current domain and tcb's domain) required for the properties currently being considered  to the extensible abstract specification.
2 - move all of the scheduler state to the extensible abstract specification, the same as what has been done on the rt branch.
3 - move all state and remove the non-deterministic specification completely.
4 - add ghost state to sidestep the specific problem that we are currently encountering (it is unclear whether this would actually work).

Option 2 was chosen because it is the least blocking (although may be more annoying than some of the others), will result in the smallest diff to the rt branch, and we don't think it will be significantly more work than option 1. If we want to reason about threads in the future (e.g. liveness), we will require that this information about the scheduler state is accessible.

We are really not sure about whether moving specifically the scheduler action is the right decision, but we are moving it across because the correctness of the ready queues depends on it and any alternatives that we have thought of would be a lot of work that we do not want to do at this time.


corlewis:
These changes are correct enough for the AARCH64 abstract spec to still process, although of course we can't be certain that they are sufficient for our purposes until AInvs and the relevant VCPU and FPU proofs are updated. Reviews now would still be helpful to check style and to confirm that the state being moved is what we wanted. Some specific style choices made in this were to minimise the difference to the rt branch.

spec/abstract/Tcb_A.thy
```diff
@@ -247,25 +247,25 @@ where
...
+  set_domain :: "obj_ref \<Rightarrow> domain \<Rightarrow> (unit, 'z::state_ext) s_monad" where
   "set_domain tptr new_dom \<equiv> do
      cur \<leftarrow> gets cur_thread;
      tcb_sched_action tcb_sched_dequeue tptr;
      thread_set_domain tptr new_dom;
      ts \<leftarrow> get_thread_state tptr;
-     when (runnable ts) (tcb_sched_action tcb_sched_enqueue tptr);
-     when (tptr = cur) reschedule_required
+     when (runnable ts) $ tcb_sched_action tcb_sched_enqueue tptr;
+     when (tptr = cur) $ reschedule_required
```
Xaphiosis, line 262: nitpick: this `$` is redundant

spec/abstract/Structures_A.thy
```diff
@@ -506,6 +520,15 @@ datatype irq_state =
  | IRQTimer
  | IRQReserved
 
+text \<open>The current scheduler action,
+  which is part of the scheduling state.\<close>
```
Xaphiosis, line 524: can this fit on one line?

spec/abstract/Schedule_A.thy
```diff
@@ -66,36 +63,47 @@ definition
...
 
-instantiation  det_ext_ext :: (type) state_ext_sched
-begin
-
 definition
   "schedule_switch_thread_fastfail ct it ct_prio target_prio \<equiv>
-     if ct \<noteq> it
-     then return (target_prio < ct_prio)
-     else return True"
+     return $ ct \<noteq> it \<longrightarrow> target_prio < ct_prio"
```
Xaphiosis, line 106: nice

spec/abstract/Schedule_A.thy
```diff
@@ -66,36 +63,47 @@ definition
...
+     then switch_to_idle_thread
+     else guarded_switch_to (hd (max_non_empty_queue queues))
+   od"
 
 text \<open>
   Determine whether given priority is highest among queued ready threads in given domain.
   Trivially true if no threads are ready.\<close>
 definition
-  is_highest_prio :: "domain \<Rightarrow> priority \<Rightarrow> det_ext state \<Rightarrow> bool"
+  is_highest_prio :: "domain \<Rightarrow> priority \<Rightarrow> 'z::state_ext state \<Rightarrow> bool"
```
Xaphiosis, line 98: that's unexpected, any reason for this change?

spec/abstract/Deterministic_A.thy
```diff
@@ -369,9 +242,9 @@ text \<open>The CDT in the implementation is stored in prefix traversal order.
   The following functions traverse its abstract representation here to
   yield corresponding information.
 \<close>
+(* FIXME: use Lib.hd_opt instead *)
```
Xaphiosis, line 245: is this intended for this round of changes?

Xaphiosis:
Having read through it, looks good from my side. Thank you for fixing up of the style along the way!

spec/abstract/Tcb_A.thy
corlewis: Good spotting. This was copied from the rt branch, and I was originally thinking that it would be best to keep things the same even when the style isn't quite the best but I've since changed my mind and will clean up things like this. The merge back into rt is already going to be pretty bad, some extra clean up won't make it noticeably worse.

spec/abstract/Schedule_A.thy
corlewis: Now that the scheduler state is part of the base state instead of the extended state, functions like this that previously operated on `det_ext state` can instead be generic. If I didn't make this change then other functions that call `is_highest_prio` would also need to be `det_ext state`, which would unnecessarily force us to still have a deterministic and non-deterministic scheduler.

spec/abstract/Deterministic_A.thy
corlewis: Not really, this comment and the style changes to `next_child` were copied from rt.

corlewis:
I've updated AInvs for one architecture now. It's a big commit and will be a pain to review, but feedback now would be very helpful, so that I can clean things up before copying it to other architectures.

proof/invariant-abstract/AARCH64/ArchKHeap_AI.thy
```diff
@@ -984,6 +984,30 @@ lemma state_hyp_refs_of_tcb_state_update:
...
+  apply (clarsimp simp add: state_hyp_refs_of_def obj_at_def split: option.splits)
+  done
+
+lemma state_hyp_refs_of_tcb_time_slice_update:
+  "kheap s t = Some (TCB tcb) \<Longrightarrow>
+   state_hyp_refs_of (s\<lparr>kheap := (kheap s)(t \<mapsto> TCB (tcb\<lparr>tcb_time_slice := ts\<rparr>))\<rparr>)
+     = state_hyp_refs_of s"
+  apply (rule all_ext)
+  apply (clarsimp simp add: state_hyp_refs_of_def obj_at_def split: option.splits)
+  done
```
Xaphiosis, line 1009: this is a bit repetitive, but it's probably fine, especially if you pulled it from rt

proof/invariant-abstract/AARCH64/ArchDetSchedSchedule_AI.thy
```diff
@@ -508,7 +510,7 @@ lemma dmo_scheduler_act_sane[wp]:
 lemma vgic_maintenance_irq_valid_sched[wp]:
   "\<lbrace>valid_sched and invs and scheduler_act_sane and ct_not_queued\<rbrace>
    vgic_maintenance
-   \<lbrace>\<lambda>rv. valid_sched\<rbrace>"
+   \<lbrace>\<lambda>rv. valid_sched :: det_state \<Rightarrow> _\<rbrace>"
```
Xaphiosis, line 513: hang on, we can do dummy types in a type signature? how did you discover this?

proof/invariant-abstract/BCorres2_AI.thy
```diff
@@ -232,6 +222,13 @@ locale is_extended = is_extended' +
 
 context is_extended begin
 
+lemma in_f_exst:
+  "(r, s') \<in> fst (f s) \<Longrightarrow> s\<lparr>exst := exst s'\<rparr> = s'"
+  apply (drule v)
```
Xaphiosis, line 227: oof, that name, `v`, also this proof is one-lineable

proof/invariant-abstract/Invariants_AI.thy
```diff
@@ -3156,10 +3209,94 @@ lemma vms_ioc_update[iff]:
...
+
+lemma valid_ioc_domain_time_update[simp]:
+  "valid_ioc (domain_time_update f s) = valid_ioc s"
+  by (simp add: valid_ioc_def)
+
+lemma cur_tcb_cdt_update [simp]:
+  "cur_tcb (cdt_update f s) = cur_tcb s"
+  by (simp add: cur_tcb_def)
+
+lemma cur_tcb_more_update[iff]:
```
Xaphiosis, line 3228: do you know why this one is `[iff]` when all the other ones are simp?
It also looks super repetitive, they all have nearly the same proof... it almost has a crunch-like flavour, but I'm not sure we have a mechanism that can do that. Maybe scrunching the proofs into one for `cur_tcb` and another for `valid_ioc`?

proof/refine/AARCH64/Refine.thy
```diff
@@ -196,18 +196,27 @@ lemma absKState_correct:
...
-         apply (rule absIsOriginalCap_correct, clarsimp+)
+                 apply (rule absHeap_correct; clarsimp)
+                 apply (clarsimp elim!: state_relationE)
+                apply (rule absCDT_correct; clarsimp)
+               apply (rule absIsOriginalCap_correct; clarsimp)
+              apply (simp add: state_relation_def)
+             apply (simp add: state_relation_def)
+            apply (clarsimp simp: state_relation_def)
+            apply (rule absSchedulerAction_correct, simp add: state_relation_def)
+           apply (simp add: state_relation_def)
```
Xaphiosis, line 207: use `+` to fold up subsequent repeats?

proof/invariant-abstract/Syscall_AI.thy
```diff
@@ -45,37 +45,89 @@ lemmas [wp] =
...
+
+lemma guarded_switch_to_invs:
+  "\<lbrace>\<lambda>s. invs s \<and> scheduler_action s \<noteq> resume_cur_thread\<rbrace>
+   guarded_switch_to thread
+   \<lbrace>\<lambda>_. invs\<rbrace>"
+  by (wpsimp simp: guarded_switch_to_def
+               wp: switch_to_thread_invs hoare_drop_imps)
+
+(* still true without scheduler_action s \<noteq> resume_cur_thread, but the proof for schedule_invs
+   will be simpler with it *)
```
Xaphiosis, line 90: hang on, strengthening a precondition makes the proof simpler?!

proof/crefine/AARCH64/Refine_C.thy
```diff
@@ -876,18 +876,17 @@ lemma threadSet_all_invs_triv':
   unfolding all_invs'_def
   apply (rule hoare_pre)
    apply (rule wp_from_corres_unit)
-     apply (rule threadset_corresT [where f="tcb_arch_update (arch_tcb_context_set f)"])
+      apply (rule threadset_corresT [where f="tcb_arch_update (arch_tcb_context_set f)"]; simp?)
```
Xaphiosis, line 879: given all resulting subgoals get `simp` anyway, and indent for subsequent apply hasn't changed, what does the `simp?` give you?

proof/crefine/AARCH64/Retype_C.thy
```diff
@@ -3135,7 +3137,8 @@ lemma cnc_tcb_helper:
                                      [heap_update (registers_of_tcb_Ptr p)
                                                   (array_updates (h_val (hrs_mem hrs) (registers_of_tcb_Ptr p))
                                                                  initContext_registers),
-                                      heap_update (machine_word_Ptr &(p\<rightarrow>[''tcbTimeSlice_C''])) 5])
+                                      heap_update (machine_word_Ptr &(p\<rightarrow>[''tcbTimeSlice_C''])) 5,
+                                      heap_update (machine_word_Ptr &(p\<rightarrow>[''tcbDomain_C''])) (d' :: machine_word)])
```
Xaphiosis, line 3141: I'm a bit confused here... this looks like the C changed somehow, but we're only changing the det_ext definitions on the abstract level, right? So why is this change needed all the way down in CRefine?

spec/design/skel/Intermediate_H.thy
```diff
@@ -48,9 +48,8 @@ defs createNewCaps_def:
 "createNewCaps t regionBase numObjects userSize dev \<equiv>
     (case toAPIType t of
           Some TCBObject \<Rightarrow> (do
-            addrs \<leftarrow> createObjects regionBase numObjects (injectKO (makeObject ::tcb)) 0;
             curdom \<leftarrow> curDomain;
-            mapM_x (\<lambda>tptr. threadSet (tcbDomain_update (\<lambda>_. curdom)) tptr) addrs;
+            addrs \<leftarrow> createObjects regionBase numObjects (injectKO ((makeObject ::tcb)\<lparr>tcbDomain := curdom\<rparr>)) 0;
```
Xaphiosis, line 52: interesting, but it seems more consistent now to go via createObjects

proof/invariant-abstract/DetSchedAux_AI.thy
```diff
@@ -130,106 +76,147 @@ lemma cap_insert_no_pred_tcb_at:
...
   done
 
-lemmas mapM_x_defsym = mapM_x_def[symmetric]
+crunch reset_untyped_cap
+  for etcb_at[wp]: "etcb_at P t"
+  (wp: preemption_point_inv mapME_x_inv_wp crunch_wps)
+
+lemma foldr_kh_eq:
+  "foldr (\<lambda>p kh. kh(p \<mapsto> ko')) ptrs kh t = Some ko \<Longrightarrow>
+  if t \<in> set ptrs then ko = ko' else kh t = Some ko"
```
Xaphiosis, line 111: `if` indent +1?

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -467,29 +478,31 @@ lemma st_tcb_at_kh_if_split:
...
-  done
+lemma etcbs_of_update_unrelated:
+  "\<lbrakk>kh ref = Some (TCB tcb); etcb_of tcb = etcb_of tcb'\<rbrakk> \<Longrightarrow>
+    etcbs_of' (\<lambda>r. if r = ref then Some (TCB tcb') else kh r) = etcbs_of' kh"
+  by (auto simp: etcbs_of'_def)
+
+lemma etcbs_of_update_state[simp]:
+  "get_tcb ref s = Some tcb \<Longrightarrow>
+   etcbs_of' (\<lambda>r. if r = ref then Some (TCB (tcb_state_update f tcb)) else kheap s r) = etcbs_of' (kheap s)"
+  by (auto simp: etcbs_of_update_unrelated dest!: get_tcb_SomeD)
```
Xaphiosis, line 492: this doesn't look like a good candidate for `[simp]`, it both has a conditional and a *very* specific pattern to match on, that looks like an unfolded fun_upd; I remember on AARCH64 proofs we had issues like that with fun_upd somewhere, but don't remember where. I think we stuck with the fun_upd syntax though and prevented unfolding somewhere

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -467,29 +478,31 @@ lemma st_tcb_at_kh_if_split:
...
   "\<lbrace>valid_queues and K (runnable ts)\<rbrace> set_thread_state ref ts \<lbrace>\<lambda>_. valid_queues\<rbrace>"
   apply (simp add: set_thread_state_def)
-  apply (wp | wpc | simp add: set_thread_state_ext_def set_object_def get_object_def)+
+  apply (wpsimp simp: set_object_def get_object_def)
   apply (clarsimp simp: valid_queues_def st_tcb_at_kh_if_split)
   done
 
+lemma etcbs_of_update_bound_notification[simp]:
+  "get_tcb ref s = Some tcb \<Longrightarrow>
+   etcbs_of' (\<lambda>r. if r = ref then Some (TCB (tcb_bound_notification_update f tcb)) else kheap s r) = etcbs_of' (kheap s)"
```
Xaphiosis, line 503: this also looks like a strange fun_upd unfolded

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -1957,40 +1847,136 @@ lemma reschedule_required_not_cur_thread[wp]:
...
+  apply (wpsimp wp: set_object_wp)
+  apply (clarsimp simp: valid_queues_def is_etcb_at'_def etcb_at'_def etcbs_of'_def
+                        not_queued_def
+                 dest!: get_tcb_SomeD)
+  by (fastforce simp: st_tcb_at_kh_def obj_at_kh_def st_tcb_at_def obj_at_def)
+
+lemma thread_set_priority_ct_not_in_q[wp]:
+  "thread_set_priority p t \<lbrace>ct_not_in_q\<rbrace>"
+  unfolding thread_set_priority_def thread_set_def
+  by (wpsimp wp: set_object_wp)
```
Xaphiosis, line 1878: this one and `thread_set_priority_neq_st_tcb_at` look crunchable

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -2303,42 +2290,28 @@ end
...
-  apply (fastforce simp: get_etcb_def valid_queues_def is_etcb_at'_def etcb_at'_def)
-  done
+lemma thread_set_valid_blocked_except:
+  "(\<And>x. tcb_state (f x) = tcb_state x) \<Longrightarrow>
+   thread_set f tptr \<lbrace>valid_blocked_except t\<rbrace>"
+  by (wpsimp wp: hoare_drop_imps valid_blocked_except_lift thread_set_no_change_tcb_state)
+
+lemma etcbs_of_update_tcb_time_slice[simp]:
+  "get_tcb ref s = Some tcb \<Longrightarrow>
+   etcbs_of' (\<lambda>r. if r = ref then Some (TCB (tcb_time_slice_update f tcb)) else kheap s r) = etcbs_of' (kheap s)"
```
Xaphiosis, line 2300: another of the fun_upd unfolded mystery

proof/refine/AARCH64/Detype_R.thy
```diff
@@ -1084,35 +1079,32 @@ lemma deleteObjects_corres:
...
-             apply (frule_tac cref="(aa, ba)" in cte_map_untyped_range,
-                    erule cte_wp_at_weakenE[OF _ TrueI], assumption+)
-             apply (simp add: add_mask_fold)
+             apply (drule(4) cte_map_not_null_outside')
+              apply (fastforce simp add: cte_wp_at_caps_of_state)
+             apply simp
+            apply (rule ext, clarsimp simp: null_filter'_def map_to_ctes_delete)
+            apply (rule sym, rule ccontr, clarsimp)
+            apply (frule(2) pspace_relation_cte_wp_atI
+                            [OF state_relation_pspace_relation])
```
Xaphiosis, line 1092: indent, or having `[` on the prev line and indent, or try fit on same line

proof/invariant-abstract/TcbAcc_AI.thy
```diff
@@ -1824,16 +1816,367 @@ lemma get_tcb_ko_atD:
...
+  and arch_state[wp]: "\<lambda>s. P (arch_state s)"
+  and aligned[wp]: pspace_aligned
+  and distinct[wp]: pspace_distinct
+  and valid_objs[wp]: valid_objs
+  and cte_wp_at[wp]: "cte_wp_at P c"
+  and interrupt_irq_node[wp]: "\<lambda>s. P (interrupt_irq_node s)"
+  and caps_of_state[wp]: "\<lambda>s. P (caps_of_state s)"
+  and no_cdt[wp]: "\<lambda>s. P (cdt s)"
+  (simp: Let_def Let_def
+   wp: hoare_drop_imps hoare_vcg_if_lift2 mapM_wp)
```
Xaphiosis, line 2180: I have a feeling that this huge crunch and the previous huge crunch have significant overlap

proof/refine/AARCH64/StateRelation.thy
```diff
@@ -676,6 +656,22 @@ lemma ghost_relation_typ_at:
...
+         split: arch_kernel_object.split)+
+
+lemma objBits_pos_power2[simp]:
+  assumes "objBits v < word_bits"
+  shows "(1::machine_word) < (2::machine_word) ^ objBits v"
+  unfolding objBits_simps'
+  apply (insert assms)
+  apply (case_tac "injectKO v"; simp)
+  by (simp add: pageBits_def pteBits_def objBits_simps
+         split: arch_kernel_object.split)+
```
Xaphiosis, line 673: is this the right place for these? would Invariants_H be better?

proof/refine/AARCH64/Retype_R.thy
```diff
@@ -1994,87 +1705,98 @@ proof -
...
+                                    bound[simplified obj_bits_api ko])[1]
+               apply (cases ty; simp; rename_tac tp; case_tac tp;
+                      clarsimp simp: default_object_def APIType_map2_def
+                              split: arch_kernel_object.splits apiobject_type.splits)
+              apply (clarsimp simp: retype_addrs_fold[symmetric] ptr_add_def upto_enum_red' not_zero'
+                                    range_cover.unat_of_nat_n[OF cover] word_le_sub1)
+              apply (rule_tac f=g in arg_cong)
+              apply clarsimp
+             apply wpsimp+
+           apply simp+
```
Xaphiosis, line 1749: unexpected, you'd think wpsimp could clear a goal that is solved by simp?

Xaphiosis (approved):
Very solid work, saw several improvements along the way, minor comments only from me. Some of the files have diffs so big they choke my browser for over a minute, not sure what can be done about that. Probably as we discussed dump the other arches in a separate PRs to the same branch?

proof/invariant-abstract/AARCH64/ArchKHeap_AI.thy
corlewis: Agreed that this isn't nice, but it's also the same as what is done just before it for other tcb fields. There's maybe some more generic lemma that could be written to handle all of these cases but that's not happening as part of this PR.

proof/invariant-abstract/AARCH64/ArchDetSchedSchedule_AI.thy
corlewis: I think I already knew that it was possible, but this specific pattern was copied from rt.

proof/invariant-abstract/Invariants_AI.thy
corlewis: Not sure, presumably historical reasons of whoever wrote them preferring one attribute over the other. I guess `simp` would be the preference now if we wanted to make them consistent?

proof/invariant-abstract/Syscall_AI.thy
corlewis: Good point, no idea what the intention was behind this comment originally but I was able to remove that precondition and the proofs still went through.

proof/invariant-abstract/DetSchedSchedule_AI.thy
corlewis: Hmm, how against this sort of lemma are you? It was copied from rt and that specific pattern comes up in a few places in the proofs, so I don't really want to try and rework things to avoid it. I did have a quick look and the fun_upd seems to be expanded by `fun_upd_apply`, which would be a pain to have to delete from `simp` everywhere.

proof/invariant-abstract/Invariants_AI.thy
Xaphiosis: I'm not sure what the difference is between `[iff]` and `[simp]` to be honest. If it's the same, simp will be more consistent, but ok to leave it.

proof/invariant-abstract/DetSchedSchedule_AI.thy
Xaphiosis: If they're copied, I think for this PR we live with them.
I'm not sure about the fun_upd situation precisely, but one way I would handle it is writing the lemma with a `:=` so that it makes sense, and using `[simplified fun_upd_def]` (or `fun_upd_apply` or whatever, to get it into simp normal form. Generally it looks really noisy, but I guess that's fun_upd for you.

Xaphiosis: There's definitely a fair bit of `supply fun_upd_apply[simp del]` and `simp_del: fun_upd_apply` on AARCH64

proof/refine/AARCH64/Retype_R.thy
corlewis: Ah, looks like the simps are instantiating some schematics that came from an earlier `corres_guard_imp`. Not the best  but not something I really want to improve as part of this PR.

spec/design/skel/Intermediate_H.thy
corlewis: I'm not sure what you mean by this.

proof/crefine/AARCH64/Refine_C.thy
corlewis: There are new, relatively trivial subgoals that are solved by that simp.

proof/crefine/AARCH64/Retype_C.thy
corlewis: I'd have to spend some time digging through the proofs to say for certain but I think this is due to the restructuring of `createNewCaps` in the exec spec, which was simplified to do everything for tcbs in one step, instead of setting the domains in a separate step. This is then reflected into how these helper lemmas in CRefine split off commands.

spec/design/skel/Intermediate_H.thy
Xaphiosis: I mean I approve of your change, thumbs up, etc

proof/crefine/AARCH64/Retype_C.thy
Xaphiosis: That sounds fair, even if initially unexpected

proof/invariant-abstract/AARCH64/ArchDetSchedAux_AI.thy
```diff
@@ -15,67 +15,27 @@ named_theorems DetSchedAux_AI_assms
...
   and valid_sched[wp]: valid_sched
-  (wp: mapM_x_wp')
-
-(* already proved earlier *)
-declare invoke_untyped_cur_thread[DetSchedAux_AI_assms]
-
-crunch invoke_untyped
-  for ready_queues[wp, DetSchedAux_AI_assms]: "\<lambda>s. P (ready_queues s)"
-  and scheduler_action[wp, DetSchedAux_AI_assms]: "\<lambda>s. P (scheduler_action s)"
+  and etcbs_of[wp, DetSchedAux_AI_assms]: "\<lambda>s. P (etcbs_of s)"
```
lsf37, line 21: Does `etcbs_of` still exist when we removed the `ekheap`?

corlewis:
This has been rebased and includes all changes required for the AArch64 proofs. It is ready to be merged into a new branch to track this in-progress work until all of the proofs for all architectures have been updated.

proof/invariant-abstract/AARCH64/ArchDetSchedAux_AI.thy
corlewis: Yes, to minimise the changes required we still define it in `DetSchedInvs_AI`, as a projection of `kheap`. This is the same as what was done in `rt`.

proof/invariant-abstract/DetSchedAux_AI.thy
```diff
@@ -130,106 +76,147 @@ lemma cap_insert_no_pred_tcb_at:
...
   apply (simp add: delete_objects_def)
-  apply (wpsimp simp: detype_def detype_ext_def wrap_ext_det_ext_ext_def do_machine_op_def)
-  apply (simp add: valid_etcbs_def st_tcb_at_kh_def obj_at_kh_def obj_at_def is_etcb_at_def)
+  apply (wpsimp simp: detype_def)
+  apply (auto simp: detype_def etcbs_of'_def etcb_at'_def)
   done
 
-lemmas mapM_x_defsym = mapM_x_def[symmetric]
+crunch reset_untyped_cap
+  for etcb_at[wp]: "etcb_at P t"
```
lsf37, line 106: but surely `etcb_at` could now be just `tcb_at`, or am I missing something?

corlewis: Huh, yeah, that's definitely true. That does feel like a something we should do, as long as it doesn't take too long now to update everywhere. But then we would want to make the same change on `rt`, and I'm not sure if that's a priority right now.

proof/invariant-abstract/TcbAcc_AI.thy
```diff
@@ -1824,16 +1816,367 @@ lemma get_tcb_ko_atD:
...
+
+lemma thread_set_domain_valid_arch_caps[wp]:
+  "thread_set_domain t d \<lbrace>valid_arch_caps\<rbrace>"
+  unfolding thread_set_domain_def
+  by (wpsimp wp: thread_set_arch_caps_trivial simp: tcb_cap_cases_def) auto
+
+lemma thread_set_domain_valid_ioports[wp]:
+  "thread_set_domain t d \<lbrace>valid_ioports\<rbrace>"
+  unfolding thread_set_domain_def
+  by (wpsimp wp: valid_ioports_lift thread_set_caps_of_state_trivial2 simp: tcb_cap_cases_def)
```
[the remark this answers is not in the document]
lsf37: This is now rebased over `valid_ioports` I think -- should we double check what is happening with this one?

proof/invariant-abstract/DetSchedAux_AI.thy
lsf37: True. Worth noting in an issue, though, it's something we should remove at some point. Happy for that to be some other time.

proof/invariant-abstract/DetSchedSchedule_AI.thy
lsf37: The lemma is not pretty, but it should be safe in the sense that the lhs of the equation is specific enough for Isabelle not to go chasing after the side condition all the time. So it shouldn't slow things down.

And agreed that this smells like fun_upd_apply repair which should ideally prevented before it happens. Ok to live with it for now, I'd say.

proof/invariant-abstract/TcbAcc_AI.thy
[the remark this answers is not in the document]
Xaphiosis: You should not be seeing that in a non-arch-specific file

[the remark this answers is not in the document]
Xaphiosis: `l4v/proof/invariant-abstract]# ag valid_iop *.thy` should return nothing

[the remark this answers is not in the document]
corlewis: Sorry, I just discovered this independently, this a remnant of my original changes I didn't catch during the rebase. I've just pushed a new commit that fixes these proofs.

proof/invariant-abstract/DetSchedSchedule_AI.thy
lsf37: It's of course not just `fun_upd_apply` repair, it is actual content: it's showing that a `tcb_state_update` is invisible in the `etcbs_of` projection. That part is actually needed. The phrasing as an expanded `fun_upd_apply` is not so great, but if it is the actual normal form in these proofs, then it's not unjustified, just a bit ugly.

proof/invariant-abstract/TcbAcc_AI.thy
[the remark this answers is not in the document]
Xaphiosis: I looked at that commit... to the untrained eye, it looks like quite a lot got pulled into `TcbAcc_AI_valid_ipc_buffer_cap_0`. Even though it's not like we're pulling things into Arch or anything, so it's fine to leave it, I'm still curious why all that got dragged in.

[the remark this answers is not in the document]
corlewis: The `valid_ioports` changes added `thread_set_valid_arch_state` to `TcbAcc_AI_valid_ipc_buffer_cap_0`, and this section of proofs uses that (and then needs to refer to the same `'state_ext` as in that locale). Technically it's just `thread_set_domain_valid_arch_state`, `thread_set_priority_valid_arch_state`, `thread_set_domain_invs`, and `thread_set_priority_invs` that need to be in the locale, but I thought I might as well keep things simple and put the whole section into it instead of dropping in and out of the locale multiple times.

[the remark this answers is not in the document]
Xaphiosis: Yeah, that's fine then. I think I'm trying to minimise what I put in to these locales since I'm in the process of arch-split itself and don't know what will or will not end up as a circular dependency. Don't think you need to worry about that.

proof/invariant-abstract/Invariants_AI.thy
lsf37: `[iff]` also makes `[intro!]` and `[elim!]` rules out of the equation, so e.g. `clarify` would act as well on that constant. Could be that this specific one has one such instance, or it was just done by another person than the rest.

proof/invariant-abstract/DetSchedDomainTime_AI.thy
```diff
@@ -18,193 +18,192 @@ definition
...
-(* We want wp to use ethread_get_inv before ethread_get_wp *)
-declare ethread_get_inv[wp del, wp]
-
+(*
+  FIXME: cleanup
+  Many of these could be factored out into the general state_ext class instead, but they will be
+  applied to det_ext lemmas that contain e.g. preemption_point which needs the det_ext work_units,
+  i.e. those will need additional locales, because 'state_ext needs to be interpreted first
+  into ?'state_ext.
+*)
```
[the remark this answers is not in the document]
lsf37: I'm not actually convinced that this factoring out would be worth it. We are in a `DetSched*` theory after all, and it would be perfectly fine for everything in here to be in `det`, even if some facts would also be true about `state_ext`.

lsf37 (approved):
That was a lot more work than anticipated. Thanks for getting through all of this.

proof/invariant-abstract/DetSchedDomainTime_AI.thy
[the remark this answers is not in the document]
corlewis: Right, but with this change the `DetSched*` theories don't actually depend on anything from the extended state, which is a bit unfortunate with the naming of them. I do agree anyway that this might not be worth it, although I would be more tempted by similar issues in locales in earlier files. They generally don't break the proofs but mean that lemmas need to be tagged with `det_state` a fair bit more than necessary.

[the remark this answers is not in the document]
lsf37: Right, that makes complete sense. Maybe filing an issue for that is the best idea then.

PR #825, opened 2024-11-06
x64 crefine: proof update after change to C

nspin:
Updates proof after https://github.com/seL4/seL4/pull/1347

`decodeX86PortInvocation` was modified to include some explicit variable initialization to satisfy GCC.

Test with: seL4/seL4#1347


corlewis (approved):
Thanks for updating the proofs for this and sorry for the delay. The tests pass as expected, so once the seL4 PR is updated then these can both be merged.

lsf37 (approved):

PR #826, opened 2024-11-14
Discard ARM Refine proofs on rt branch

corlewis:
This restores the ARM refine proofs to match master. The rt changes to these proofs have not been maintained for a long time and have not been kept up to date with changes made to the specifications and ainvs. There also hasn't been a strong effort to keep them in a consistent state when doing merges in the past. When we want to do these proofs for ARM again in the future we will need to pull in the changes from RISCV64 anyway, and discarding them like this will make merges easier now.

### DIFF
--- a/proof/refine/ARM/ADT_H.thy
+++ b/proof/refine/ARM/ADT_H.thy
@@ -12,11 +12,6 @@ imports
   Syscall_R
 begin
 
-(* FIXME: many of the naming conventions in this file are wrong. Definitions should start with
-          lower case letters, and should be under_score, not camelCase. Some, possibly many of
-          the functions here can benefit from projections that have been developed since this
-          was originally written *)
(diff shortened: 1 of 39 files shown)


Xaphiosis (approved):
I assume this is a straight clobber, so am approving it as we discussed. We should be able to bring it back faster once arch-split makes it to MCS.

lsf37 (approved):
Agreed, let's do that.

corlewis:
I have updated this to do the correct clobbering after the last merge into the rt branch. What I'm not sure about though is if we should also discard new files that have been added to ARM Refine in the rt branch (`Reply_R.thy`, `SchedContextInv_R.thy`, `SchedContext_R.thy`) or if we don't care about them since they won't cause any merge conflicts.

Issue #746, opened 2024-03-28
Decide on style for `[def]rule_tac ... [and ...] in ...`


Xaphiosis:
We had some discussions on the Proofcraft mattermost (present were myself, @lsf37 , @corlewis , @michaelmcinerney ), which ended up with things not quite nailed down in the end. Let's nail it down and put it in the style guide.

General considerations:
* we need a left-operator (`and`) and right-operator (`and`) version
* does the `in` belong to the `rule_tac`, or the instantiation? (in previous discussions: to the `rule_tac`)
* right-alignment of the `and/in` to utilise space better (similar to `:` alignment on method args)

At the time of discussion, we settled roughly on:
```
(* left-operator space-utilising / pretty version 1 - preferred *)
apply (rule_tac x=longthing
            and y=longthing2
             in the_rule)

(* right-operator version - preferred when using right operator wrapping style *)
apply (rule_tac x="..." and
                y="..." and
                z=True
                in some_rule) (* this must not go on previous line for multi-line instantiations *)

(* left-operator naive version (something a tool could potentially handle) - permitted *)
apply (rule_tac x="..."
                and y="..."
                and z=True
                in some_rule)
```

The following were ruled out:
```
left-wrap pretty version 2:
apply (rule_tac x="..."
            and y="..."
            and z=True
                in some_rule)

left-wrap pretty version 3:
apply (rule_tac x="..."
            and y="..."
            and z=True
         in some_rule)
```

However, the above consensus caused friction in the following (one-variable long instantiation) case (pointed out by @lsf37), bringing us back to a need to go over this and pin things down:
```
(* one-long-instantiation version 1 *)
apply (rule_tac x="long instantiation"
             in some_rule)
vs
(* one-long-instantiation version 2 *)
apply (rule_tac x="long instantiation"
                in some_rule)
```

For recent proofs I've been going with version 1, because of its consistency with pretty version 1 above.
One of these days maybe we'll have a tool, but even then we'd still have to decide what the tool reformats to.

lsf37:
I 'm fine the first two (pretty left, standard right), and can live with the naive left version as well, so those are all good.

I guess variant 1 and 2 at the end just come down to being the degenerate forms of the left and right versions, so both should be permitted, and it also explains why Raf prefers variant 1 and I prefer variant 2 :-)

So we really just have the 2 preferred versions, and the naive-left version in addition to that.

Xaphiosis:
[...]

PR #831, opened 2024-11-30
Prove `cancelIPC_ccorres1`

nspin:
Requires https://github.com/seL4/seL4/pull/1358, which fixes an apparent potential bug that I discovered while working on this proof. That PR also sneaks in a few trivial changes that made verification a bit easier.

I weakened the preconditions in the still-unproven `reply_unlink_ccorres` lemma. They now correspond to those in `replyUnlinkTcb_corres`.

Edit: also requires https://github.com/seL4/seL4/pull/1297.

### DIFF
--- a/proof/crefine/RISCV64/Ctac_lemmas_C.thy
+++ b/proof/crefine/RISCV64/Ctac_lemmas_C.thy
@@ -46,6 +46,13 @@ lemma c_guard_abs_sc:
 
 lemmas ccorres_move_c_guard_sc[corres_pre] = ccorres_move_c_guards[OF c_guard_abs_sc]
 
+lemma c_guard_abs_reply:
+  fixes p :: "reply_C ptr"
+  shows "\<forall>s s'. (s, s') \<in> rf_sr \<and> reply_at' (ptr_val p) s \<and> True \<longrightarrow> s' \<Turnstile>\<^sub>c p"
+  by (fastforce dest!: reply_at_h_t_valid)
+
+lemmas ccorres_move_c_guard_reply = ccorres_move_c_guards[OF c_guard_abs_reply]
+
 lemma cte_array_relation_array_assertion:
   "gsCNodes s p = Some n \<Longrightarrow> cte_array_relation s cstate
     \<Longrightarrow> array_assertion (cte_Ptr p) (2 ^ n) (hrs_htd (t_hrs_' cstate))"
--- a/proof/crefine/RISCV64/IpcCancel_C.thy
+++ b/proof/crefine/RISCV64/IpcCancel_C.thy
@@ -3043,212 +3043,341 @@ lemma reply_remove_ccorres:
     (replyRemove replyPtr tcbPtr) (Call reply_remove_'proc)"
 sorry (* FIXME RT: reply_remove_ccorres *)
 
+lemma getBlockingObject_BlockedOnReceive_return:
+  "(getBlockingObject (Structures_H.thread_state.BlockedOnReceive oref cg ro)) = return oref"
+  unfolding getBlockingObject_def epBlocked_def by simp
+
+lemma getBlockingObject_BlockedOnSend_return:
+  "(getBlockingObject (Structures_H.thread_state.BlockedOnSend oref badge cg cgr isc)) = return oref"
+  unfolding getBlockingObject_def epBlocked_def by simp
+
+lemma BlockedOnReceive_replyObject_no_0:
+  "\<lbrakk> ts = BlockedOnReceive oref cg ro; no_0_obj' s; valid_tcb_state' ts s \<rbrakk>
+   \<Longrightarrow> ro \<noteq> Some 0"
+  by (auto simp: valid_tcb_state'_def)
+
+lemma ctcb_relation_BlockedOnReceive_blockingObject:
+  "\<lbrakk> BlockedOnReceive oref cg ro = tcbState tcb; ctcb_relation tcb ctcb \<rbrakk>
+   \<Longrightarrow> blockingObject_CL (thread_state_lift (tcbState_C ctcb)) = oref"
+  by (cases "(tcbState tcb)", simp_all add: ctcb_relation_def cthread_state_relation_def)
+
+lemma ctcb_relation_BlockedOnReceive_replyObject:
+  "\<lbrakk> BlockedOnReceive oref cg ro = tcbState tcb; ctcb_relation tcb ctcb \<rbrakk>
+   \<Longrightarrow> replyObject_CL (thread_state_lift (tcbState_C ctcb)) = option_to_0 ro"
+  by (cases "(tcbState tcb)", simp_all add: ctcb_relation_def cthread_state_relation_def)
+
+lemma ctcb_relation_BlockedOnSend_blockingObject:
+  "\<lbrakk> BlockedOnSend oref badge cg cgr isc = tcbState tcb; ctcb_relation tcb ctcb \<rbrakk>
+   \<Longrightarrow> blockingObject_CL (thread_state_lift (tcbState_C ctcb)) = oref"
+  by (cases "(tcbState tcb)", simp_all add: ctcb_relation_def cthread_state_relation_def)
+
+lemma BlockedOnReceive_ep_at':
+  "\<lbrakk> st_tcb_at' ((=) (BlockedOnReceive oref cg ro)) t s; valid_objs' s \<rbrakk>
+   \<Longrightarrow> ep_at' oref s"
+  by (fastforce dest!: tcb_in_valid_state' simp: obj_at'_def valid_tcb_state'_def)
+
+lemma BlockedOnSend_ep_at':
+  "\<lbrakk> st_tcb_at' ((=) (BlockedOnSend oref badge cg cgr isc)) t s; valid_objs' s \<rbrakk>
+   \<Longrightarrow> ep_at' oref s"
+  by (fastforce dest!: tcb_in_valid_state' simp: obj_at'_def valid_tcb_state'_def)
+
+lemma sym_ref_BlockedOnReceive_replyObject_linked:
+  assumes st: "st_tcb_at' ((=) (Structures_H.thread_state.BlockedOnReceive oref cg (Some ro))) thread s"
+  assumes sy: "sym_refs (state_refs_of' s)"
+  assumes vo: "valid_objs' s"
+  shows "obj_at' (\<lambda>reply. replyTCB reply = Some thread) ro s"
+proof -
+  from vo and st have ra:
+    "reply_at' ro s"
+    by - (drule (1) tcb_in_valid_state', fastforce simp: obj_at'_def valid_tcb_state'_def)
+  from st have ko:
+    "\<exists>tcb. ko_at' tcb thread s \<and> tcbState tcb = BlockedOnReceive oref cg (Some ro)"
+    by (clarsimp simp: st_tcb_at'_def obj_at'_def)
+  with sy and vo and st and ra show ?thesis
+    apply (clarsimp simp: st_tcb_at'_def)
+    apply (frule (1) sym_ref_Receive_or_Reply_replyTCB')
+    apply simp
+    apply (clarsimp simp: obj_at'_def)
+    done
+qed
+
+lemma thread_state_to_tsType_eq_BlockedOnReceive:
+  "BlockedOnReceive oref cg ro = ts
+   \<Longrightarrow> thread_state_to_tsType ts = scast ThreadState_BlockedOnReceive"
+  by (cases ts, simp_all add: ThreadState_defs)
+
+lemma thread_state_to_tsType_eq_BlockedOnSend:
+  "BlockedOnSend oref badge cg cgr isc = ts
+   \<Longrightarrow> thread_state_to_tsType ts = scast ThreadState_BlockedOnSend"
+  by (cases ts, simp_all add: ThreadState_defs)
+
 lemma cancelIPC_ccorres1:
   assumes cteDeleteOne_ccorres:
-  "\<And>w slot. ccorres dc xfdc
-   (invs'
-    and cte_wp_at' (\<lambda>ct. w = -1 \<or> cteCap ct = NullCap
-                         \<or> (\<forall>cap'. ccap_relation (cteCap ct) cap' \<longrightarrow> cap_get_tag cap' = w)) slot)
-   ({s. gs_get_assn cteDeleteOne_'proc (ghost'state_' (globals s)) = w}
-        \<inter> {s. slot_' s = Ptr slot}) []
-   (cteDeleteOne slot) (Call cteDeleteOne_'proc)"
+  "\<And>w slot.
+     ccorres dc xfdc
+       (invs'
+        and cte_wp_at' (\<lambda>ct. w = -1 \<or> cteCap ct = NullCap
+                             \<or> (\<forall>cap'. ccap_relation (cteCap ct) cap' \<longrightarrow> cap_get_tag cap' = w)) slot)
+       (\<lbrace>gs_get_assn cteDeleteOne_'proc \<acute>ghost'state = w\<rbrace> \<inter> \<lbrace>\<acute>slot = cte_Ptr slot\<rbrace>) []
+     (cteDeleteOne slot) (Call cteDeleteOne_'proc)"
   shows
-  "ccorres dc xfdc (tcb_at' thread and invs' and (\<lambda>s. weak_sch_act_wf (ksSchedulerAction s) s))
-                   (UNIV \<inter> {s. tptr_' s = tcb_ptr_to_ctcb_ptr thread}) []
-          (cancelIPC thread) (Call cancelIPC_'proc)"
+  "ccorres dc xfdc
+     (tcb_at' thread and invs' and (\<lambda>s. weak_sch_act_wf (ksSchedulerAction s) s))
+     \<lbrace>\<acute>tptr = tcb_ptr_to_ctcb_ptr thread\<rbrace> hs
+     (cancelIPC thread) (Call cancelIPC_'proc)"
   apply (cinit lift: tptr_' simp: Let_def cong: call_ignore_cong)
-sorry (* FIXME RT: cancelIPC_ccorres1 *) (*
    apply (rule ccorres_move_c_guard_tcb)
-   apply csymbr
+   apply (rule ccorres_stateAssert)+
    apply (rule getThreadState_ccorres_foo)
-   apply (rule ccorres_symb_exec_r)
-     apply (rule_tac xf'=ret__unsigned_longlong_' in ccorres_abstract, ceqv)
-     apply (rule_tac P="rv' = thread_state_to_tsType rv" in ccorres_gen_asm2)
-     apply wpc
-            \<comment> \<open>BlockedOnReceive\<close>
-            apply (simp add: word_sle_def ccorres_cond_iffs cong: call_ignore_cong)
-            apply (rule ccorres_rhs_assoc)+
-            apply csymbr
-            apply csymbr
-            apply (rule ccorres_pre_getEndpoint)
-            apply (rule ccorres_assert)
-            apply (rule ccorres_symb_exec_r) \<comment> \<open>ptr_get lemmas don't work so well :(\<close>
+   apply (rename_tac threadState)
+   apply csymbr
+   apply csymbr
+   apply (rule ccorres_split_nothrow)
+       apply (rule threadSet_ccorres_lemma2[where P=\<top>], rename_tac tcb)
+        apply vcg
+       apply (clarsimp simp: typ_heap_simps')
+       apply (erule (1) rf_sr_tcb_update_no_queue2, (simp add: typ_heap_simps')+)[1]
+        apply (rule ball_tcb_cte_casesI; simp)
+       apply (clarsimp simp: ctcb_relation_def seL4_Fault_lift_NullFault
+                             cfault_rel_def cthread_state_relation_def)
+       apply (case_tac "tcbState tcb"; simp add: is_cap_fault_def)
+      apply ceqv
+     apply (rule_tac xf'=ret__unsigned_longlong_'
+                 and val="thread_state_to_tsType threadState"
+                 and R="st_tcb_at' ((=) threadState) thread"
+                  in ccorres_symb_exec_r_known_rv[where R'=UNIV])
+        apply (rule conseqPre, vcg)
+        apply (clarsimp simp: st_tcb_at'_def)
+        apply (frule (1) obj_at_cslift_tcb)
+        apply (clarsimp simp: typ_heap_simps ctcb_relation_thread_state_to_tsType)
+       apply ceqv
+      apply wpc
+             \<comment> \<open>BlockedOnReceive\<close>
+             apply (rename_tac oref cg ro)
+             apply (unfold blockedCancelIPC_def)
+             apply (simp add: ccorres_cond_iffs getBlockingObject_BlockedOnReceive_return)
+             apply (rule ccorres_rhs_assoc)+
+             apply csymbr
+             apply csymbr
+             apply (rule ccorres_pre_getEndpoint)
+             apply (rule ccorres_assert2)
+             apply (rule_tac xf'=ret__unsigned_longlong_'
+                         and val="oref"
+                         and R="st_tcb_at' ((=) (BlockedOnReceive oref cg ro)) thread"
+                          in ccorres_symb_exec_r_known_rv[where R'=UNIV])
+                apply (rule conseqPre, vcg)
+                apply (clarsimp simp: st_tcb_at'_def)
+                apply (frule (1) obj_at_cslift_tcb)
+                apply (clarsimp simp: typ_heap_simps ctcb_relation_BlockedOnReceive_blockingObject)
+               apply ceqv
               apply (rule ccorres_symb_exec_r)
-                apply (simp only: fun_app_def simp_list_case_return
-                                  return_bind ccorres_seq_skip)
+                apply (simp only: list_case_If)
                 apply (rule ccorres_rhs_assoc2)
                 apply (rule ccorres_rhs_assoc2)
                 apply (rule ccorres_rhs_assoc2)
+                apply (rule ccorres_assert2)
                 apply (ctac (no_vcg) add: cancelIPC_ccorres_helper)
-                  apply (ctac add: setThreadState_ccorres_valid_queues')
-                 apply (wp hoare_vcg_all_lift set_ep_valid_objs' | simp add: valid_tcb_state'_def split del: if_split)+
-                apply (simp add: ThreadState_defs)
+                  apply (rule_tac P="tcb_at' thread" in ccorres_cross_over_guard)
+                  apply (rule_tac xf'=ret__unsigned_longlong_'
+                              and val="thread_state_to_tsType threadState"
+                              and R="st_tcb_at' ((=) threadState) thread"
+                               in ccorres_symb_exec_r_known_rv[where R'=UNIV])
+                     apply (rule conseqPre, vcg)
+                     apply (clarsimp simp: st_tcb_at'_def)
+                     apply (frule (1) obj_at_cslift_tcb)
+                     apply (fastforce simp: typ_heap_simps ctcb_relation_thread_state_to_tsType
+                                            thread_state_to_tsType_eq_BlockedOnReceive)
+                    apply ceqv
+                   apply ccorres_rewrite
+                   apply (rule ccorres_rhs_assoc)
+                   apply (rule ccorres_rhs_assoc)
+                   apply (rule_tac xf'=ret__unsigned_longlong_'
+                               and val="option_to_0 ro"
+                               and R="st_tcb_at' ((=) threadState) thread"
+                                in ccorres_symb_exec_r_known_rv[where R'=UNIV])
+                      apply (rule conseqPre, vcg)
+                      apply (clarsimp simp: st_tcb_at'_def)
+                      apply (frule (1) obj_at_cslift_tcb)
+                      apply (clarsimp simp: typ_heap_simps
+                                            ctcb_relation_BlockedOnReceive_replyObject)
+                     apply ceqv
+                    apply csymbr
+                    apply (rule_tac P="ro \<noteq> Some 0" in ccorres_gen_asm)
+                    apply wpc
+                      \<comment> \<open>case: ro is None\<close>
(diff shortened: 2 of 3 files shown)


proof/crefine/RISCV64/IpcCancel_C.thy
```diff
@@ -2739,19 +2739,20 @@ lemma epQueue_tail_sign[simp]:
 (* Clag from cancelSignal_ccorres_helper *)
 
 lemma cancelIPC_ccorres_helper:
-  "ccorres dc xfdc (invs' and (\<lambda>s. sym_refs (state_refs_of' s)) and
+  "\<lbrakk>epptr = Ptr ep\<rbrakk> \<Longrightarrow>
+   ccorres dc xfdc (invs' and (\<lambda>s. sym_refs (state_refs_of' s)) and
          st_tcb_at' (\<lambda>st. (isBlockedOnSend st \<or> isBlockedOnReceive st)
                             \<and> blockingObject st = ep) thread
         and ko_at' ep' ep)
-        {s. epptr_' s = Ptr ep}
```
Xaphiosis, line 2746: Moving this from the preconditions to the assumptions might seem convenient here, but in general this move is against the flow of automation, e.g. if you want to use `ctac`; the idea is preconditions bubble up the stack and are resolved at the end. Is there any reason you had to change this?

proof/crefine/RISCV64/IpcCancel_C.thy
```diff
@@ -2915,9 +2916,13 @@ sorry (* FIXME RT: reply_remove_tcb_corres *)
...
 lemma reply_unlink_ccorres:
   "ccorres dc xfdc
-    (invs' and tcb_at' tcbPtr and reply_at' replyPtr)
-    (\<lbrace>\<acute>tcb = tcb_ptr_to_ctcb_ptr tcbPtr\<rbrace> \<inter> \<lbrace>\<acute>reply = Ptr replyPtr\<rbrace>) []
-    (replyUnlink tcbPtr replyPtr) (Call reply_unlink_'proc)"
+    (valid_tcbs' and pspace_aligned' and pspace_distinct'
+       and st_tcb_at' (\<lambda>st. (\<exists>oref cg. st = Structures_H.BlockedOnReceive oref cg (Some replyPtr))
+                            \<or> st = Structures_H.BlockedOnReply (Some replyPtr))
+                      tcbPtr
+       and obj_at' (\<lambda>reply. replyTCB reply = Some tcbPtr) replyPtr)
```
Xaphiosis, line 2923: indentation: the advantage of putting operators on the left is we can align them with the previous line in most cases (see style guide), so `and` should align with `valid_tcbs'`.
Also, the `∨` belongs under the `∃`, i.e. under `st =`

proof/crefine/RISCV64/IpcCancel_C.thy
```diff
@@ -2934,6 +2939,105 @@ lemma reply_remove_ccorres:
...
+  "(getBlockingObject (Structures_H.thread_state.BlockedOnSend bo bib bicg bicgr biic)) = return bo"
+  unfolding getBlockingObject_def epBlocked_def by simp
+
+lemma BlockedOnReceive_replyObject_no_0:
+  assumes "no_0_obj' s"
+  assumes "valid_tcb_state' ts s"
+  assumes "ts = BlockedOnReceive bo bicg ro"
+  shows "ro \<noteq> Some 0"
+  using assms
+  by (auto simp: valid_tcb_state'_def)
```
Xaphiosis, line 2956: For these simple lemmas, declaring all the assumptions via `assumes` and then immediately using `assms` doesn't gain a lot in terms of readability. You can compress them vertically a lot, e.g.:

```
lemma BlockedOnReceive_replyObject_no_0:
  "⟦ no_0_obj' s; valid_tcb_state' ts s; ts = BlockedOnReceive bo bicg ro ⟧
   ⟹ ro ≠ Some 0"
  by (auto simp: valid_tcb_state'_def)
```

As an aside, rules tend to match better in Isabelle when their first assumption is more specific, since you usually want to make it unify with one of your goal assumptions. In this case `valid_tcb_state' ts s`  or `ts = BlockedOnReceive bo bicg ro` might be more useful to go first.

Feel free to ask us questions if it's not clear why.

proof/crefine/RISCV64/IpcCancel_C.thy
```diff
@@ -2934,6 +2939,105 @@ lemma reply_remove_ccorres:
...
+  using assms
+  by (cases "(tcbState tcb)", simp_all add: ctcb_relation_def cthread_state_relation_def)
+
+lemma BlockedOnReceive_ep_at':
+  assumes "valid_objs' s"
+  assumes "st_tcb_at' ((=) (BlockedOnReceive bo bicg ro)) t s"
+  shows "ep_at' bo s"
+  using assms
+  apply -
+  apply (drule (1) tcb_in_valid_state')
```
Xaphiosis, line 2985: if you compress these lemmas like I outlined, you won't need the `use assms`, and therefore won't need to chain in the facts with `apply -`. There's also a chance that you might be able to fold the `drule (1) tcb_in_valid_state'` into the `fastforce`, so that the proof ends up as:
`by (fastforce dest!: tcb_in_valid_state' simp: obj_at'_def valid_tcb_state'_def)`

proof/crefine/RISCV64/IpcCancel_C.thy
```diff
@@ -2934,6 +2939,105 @@ lemma reply_remove_ccorres:
...
+  shows "obj_at' (\<lambda>reply. replyTCB reply = Some thread) ro s"
+proof -
+  from vo and st have ra:
+    "reply_at' ro s"
+    by - (drule (1) tcb_in_valid_state', fastforce simp: obj_at'_def valid_tcb_state'_def)
+  from st have ko:
+    "\<exists>tcb. ko_at' tcb thread s \<and> tcbState tcb = BlockedOnReceive bo bicg (Some ro)"
+    by (clarsimp simp: st_tcb_at'_def obj_at'_def)
+  from sy and vo and st and ra and ko show ?thesis
+    apply -
```
Xaphiosis, line 3010: you don't need `apply -` for tactics like simp/clarsimp/fastforce/auto/etc, it's mostly for direct rule invocations (obvious, I know :/)
you can also shave off the `ko` by using `with` instead of `from`

proof/crefine/RISCV64/IpcCancel_C.thy
```diff
@@ -2934,6 +2939,105 @@ lemma reply_remove_ccorres:
...
+    done
+qed
+
+lemma cancelIPC_threadSet_wp:
+  shows "threadSet (tcbFault_update (\<lambda>_. None)) t
+         \<lbrace>\<lambda>s. st_tcb_at' ((=) ts) t s \<and> invs' s
+              \<and> weak_sch_act_wf (ksSchedulerAction s) s \<and> sym_refs_asrt s\<rbrace>"
+proof -
+  have sy: "threadSet (tcbFault_update (\<lambda>_. None)) t \<lbrace>sym_refs_asrt\<rbrace>"
+    by - (unfold sym_refs_asrt_def, wpsimp wp: threadSet_state_refs_of', simp,
```
Xaphiosis, line 3024: if you're doing Isar-style proofs, it makes more sense to do:

```
unfolding sym_refs_asrt_def
by wpsimp
   (force simp: tcb_bound_refs'_def)+
```

proof/crefine/RISCV64/IpcCancel_C.thy
```diff
@@ -2934,6 +2939,105 @@ lemma reply_remove_ccorres:
...
+qed
+
+lemma thread_state_to_tsType_eq_BlockedOnReceive:
+  "(thread_state_to_tsType ts = scast ThreadState_BlockedOnReceive)
+      = (\<exists>bo bicg ro. ts = BlockedOnReceive bo bicg ro)"
+  by (cases ts, simp_all add: ThreadState_defs)
+
+lemma thread_state_to_tsType_eq_BlockedOnSend:
+  "(thread_state_to_tsType ts = scast ThreadState_BlockedOnSend)
+      = (\<exists>bo bib bicg bicgr biic. ts = BlockedOnSend bo bib bicg bicgr biic)"
```
Xaphiosis, line 3038: indentation on `=`, also, do these rules really benefit you most as a rewrite rather than a destruction/elimination rule?

proof/crefine/RISCV64/IpcCancel_C.thy
```diff
@@ -2934,6 +2939,105 @@ lemma reply_remove_ccorres:
...
+  by (cases ts, simp_all add: ThreadState_defs)
+
 lemma cancelIPC_ccorres1:
```
Xaphiosis, line 3041: I have more comments for the proof of this lemma, but it would be useful to know how much of the proof is copied, as a bunch of them will be style-and-compression related.

proof/crefine/RISCV64/IpcCancel_C.thy
```diff
@@ -2948,198 +3052,261 @@ lemma cancelIPC_ccorres1:
...
+                               \<and> valid_objs' s
+                               \<and> weak_sch_act_wf (ksSchedulerAction s) s
+                               \<and> no_0_obj' s
+                               \<and> pspace_aligned' s
+                               \<and> pspace_distinct' s
+                               \<and> pspace_bounded' s
+                               \<and> valid_tcbs' s
+                               \<and> (ro = None \<longrightarrow> True)
+                               \<and> (\<forall>x. ro = Some x \<longrightarrow> obj_at' (\<lambda>reply. replyTCB reply = Some thread) x s)"
+                        in hoare_post_imp)
```
Xaphiosis, line 3158: This kind of explicit goal rephrasing is a sign that backwards reasoning is breaking down for some reason. Usually that's not a good sign, but it is at times unavoidable. Any thoughts on what happened?

Xaphiosis:
Firstly, welcome, and huge congratulations for being one of the very very few (maybe the second person ever?) to be a third-party contributor to the verification side of seL4. The proofs look ok, comments are mainly for improvements / learning / style.

proof/crefine/RISCV64/IpcCancel_C.thy
nspin: I made this change because I had a feeling that I should prefer the more structured-seeming `csymbr`  over `ccorres_symb_exec_r`, and using a value rather than a precondition on a variable in this helper lemma made it easier to use `csymbr` around it in the main proof. I've changed this lemma back to using the precondition, and I see how that results in the composition of the helper lemma and the main proof fit better into the ccorres proof flow.

proof/crefine/RISCV64/IpcCancel_C.thy
nspin: I've fixed the indentation level of the `and`s.

As for the `∨`, I put it under the `(` that precedes the `∃` because the structure of the term is `(∃ ... st = ...) ∨ st = ...`. Could you clarify where the  `∨` belongs and why?

proof/crefine/RISCV64/IpcCancel_C.thy
nspin: Done.

proof/crefine/RISCV64/IpcCancel_C.thy
nspin: Done, thanks.

proof/crefine/RISCV64/IpcCancel_C.thy
```diff
@@ -2934,6 +2939,105 @@ lemma reply_remove_ccorres:
...
+lemma ctcb_relation_BlockedOnReceive_replyObject:
+  assumes "ctcb_relation tcb ctcb"
+  assumes "BlockedOnReceive bo bicg ro = tcbState tcb"
+  shows "replyObject_CL (thread_state_lift (tcbState_C ctcb)) = option_to_0 ro"
+  using assms
+  by (cases "(tcbState tcb)", simp_all add: ctcb_relation_def cthread_state_relation_def)
+
+lemma ctcb_relation_BlockedOnSend_blockingObject:
+  assumes "ctcb_relation tcb ctcb"
+  assumes "BlockedOnSend bo bib bicg bicgr biic = tcbState tcb"
```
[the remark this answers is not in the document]
nspin: I invented them, but now I've switched to variable names found elsewhere in CRefine (`oref badge cg cgr isc`, etc.).

proof/crefine/RISCV64/IpcCancel_C.thy
nspin: I've changed these into destruction rules so that their structure more clearly aligns with how they are used.

proof/crefine/RISCV64/IpcCancel_C.thy
nspin: I've compressed these lemmas by preferring the `⟦ ... ⟧ ⟹ ...` syntax and reordered some assumptions based on the tip you gave.

proof/crefine/RISCV64/IpcCancel_C.thy
nspin: This mess was an artifact of the approach I took to figuring out how to do this proof. It's unnecessary, and I've removed it and moved the following bit to the end of the proof where it belongs.

What happened is that I was having trouble with a lhs Hoare triple goal with an option split in the postcondition. Not having discovered `hoare_case_option_wp2` yet, I tried a bunch of different things and ended up down this goal-rephrasing path in my attempt to make incremental progress. Once I finally did discover `hoare_case_option_wp2` and was able move on with the rest of the proof, and again during my subsequent synthesis+streamlining passes, I missed realizing the goal rephrasing was unnecessary.

proof/crefine/RISCV64/IpcCancel_C.thy
nspin: Some blocks, along with the overall structure, are copied from the non-MCS proof, but I've touched most of it. Sort of a ship of Theseus-like situation.

kent-mcleod:
> Firstly, welcome, and huge congratulations for being one of the very very few (maybe the second person ever?) to be a third-party contributor to the verification side of seL4.

Congratulations @nspin !!

proof/crefine/RISCV64/IpcCancel_C.thy
Xaphiosis: That looks much cleaner, thank you.

proof/crefine/RISCV64/IpcCancel_C.thy
Xaphiosis: Cool. You've now managed the C side of this, but you've introduced a similar thing on the Haskell side in `cancelIPC_ccorres_helper` via `epptr = Ptr ep`. It might make sense to do the same manoeuvre there. Traditionally that's been done by adding a `and K (epptr = Ptr ep)` to the Haskell precondition side, and then if you really need that fact immediately in the proof, using `rule ccorres_grab_asm` as the first proof step.

proof/crefine/RISCV64/IpcCancel_C.thy
Xaphiosis: You put it in the right place, for exactly the reasons you stated. If you look at the `∨`, it should be to the right of whatever it belongs to, i.e. the `λst.`, and it is on the same level as the LHS argument of the `∨`, which is the `(∃ ... st = ...)`. This means that the structure can be seen at a glance, which is important when doing textual surgery on theory files (which is diffs, git conflicts, PRs, etc).

proof/crefine/RISCV64/IpcCancel_C.thy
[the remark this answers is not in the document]
Xaphiosis: Cheers.

proof/crefine/RISCV64/IpcCancel_C.thy
Xaphiosis: There are things done during figuring out proofs that rarely make it to the final product, but are perfectly reasonable methods to go about it. Hand-holding instantiations, splitting up the proof at a "known" juncture to understand which of the facts is actually necessary, and so on. This looked like one of those cases :)

proof/crefine/RISCV64/IpcCancel_C.thy
```diff
@@ -2948,198 +3035,256 @@ lemma cancelIPC_ccorres1:
                    (UNIV \<inter> {s. tptr_' s = tcb_ptr_to_ctcb_ptr thread}) []
           (cancelIPC thread) (Call cancelIPC_'proc)"
   apply (cinit lift: tptr_' simp: Let_def cong: call_ignore_cong)
-sorry (* FIXME RT: cancelIPC_ccorres1 *) (*
    apply (rule ccorres_move_c_guard_tcb)
-   apply csymbr
+   apply (rule ccorres_stateAssert)
+   apply (rule ccorres_stateAssert)
```
Xaphiosis, line 3040: trivial: can say `apply (rule ccorres_stateAssert)+` as a message of "we're going through all the asserts here"

proof/crefine/RISCV64/IpcCancel_C.thy
```diff
@@ -2948,198 +3035,256 @@ lemma cancelIPC_ccorres1:
...
-            apply (rule ccorres_symb_exec_r) \<comment> \<open>ptr_get lemmas don't work so well :(\<close>
+   apply (rename_tac threadState)
+   apply csymbr
+   apply csymbr
+   apply (rule ccorres_split_nothrow)
+       apply (rule threadSet_ccorres_lemma2[where P=\<top>])
+        apply vcg
+       apply (clarsimp simp: typ_heap_simps')
+       apply (erule(1) rf_sr_tcb_update_no_queue2, (simp add: typ_heap_simps')+)[1]
+        apply (rule ball_tcb_cte_casesI, simp_all)[1]
```
Xaphiosis, line 3050: this might be copied from and old proof, these days we can do
`apply (rule ball_tcb_cte_casesI; simp)` instead

Also, the `erule(1)` above it should be `erule (1)`, so this might be ancient, since that line probably could also be rewritten as:
`apply (erule (1) rf_sr_tcb_update_no_queue2; simp add: typ_heap_simps')`

proof/crefine/RISCV64/IpcCancel_C.thy
```diff
@@ -2948,198 +3035,256 @@ lemma cancelIPC_ccorres1:
...
+             apply (rule ccorres_rhs_assoc)+
+             apply csymbr
+             apply csymbr
+             apply (rule ccorres_pre_getEndpoint)
+             apply (rule ccorres_assert)
+             apply (rule_tac xf'=ret__unsigned_longlong_'
+                         and val="oref"
+                         and R="st_tcb_at' ((=) (BlockedOnReceive oref cg ro)) thread"
+                         and R'=UNIV
+                         in ccorres_symb_exec_r_known_rv)
```
Xaphiosis, line 3080: trivial: on these you can save visual space on what's uninteresting by folding the UNIV into:
`in ccorres_symb_exec_r_known_rv[where R'=UNIV]`

indentation: this is sadly my fault, because there's an issue to update the style guide and I haven't done it yet (https://github.com/seL4/l4v/issues/746), but the end of the `in` should align with the end of `and` since this right-aligns the uninteresting glue bits so that the interesting bits can end up left-aligned. Applies to all the big `rule_tac`s in this proof.

proof/crefine/RISCV64/IpcCancel_C.thy
```diff
@@ -2948,198 +3035,256 @@ lemma cancelIPC_ccorres1:
...
+                      apply (clarsimp simp: typ_heap_simps
+                                            ctcb_relation_BlockedOnReceive_replyObject)
+                     apply ceqv
+                    apply csymbr
+                    apply (rule_tac P="ro \<noteq> Some 0" in ccorres_gen_asm)
+                    apply wpc
+                      \<comment> \<open>None\<close>
+                     apply simp
+                     apply (ctac add: setThreadState_ccorres)
+                    \<comment> \<open>Some\<close>
```
Xaphiosis, line 3124: for these None/Some comments (good idea), would be more useful to say what they refer to; this is obvious when you have the proof open, but not so obvious when reading the text, since it comes out of `wpc`

proof/crefine/RISCV64/IpcCancel_C.thy
```diff
@@ -2948,198 +3035,256 @@ lemma cancelIPC_ccorres1:
...
+      apply (simp add: word_sle_def ThreadState_defs ccorres_cond_iffs
+                 cong: call_ignore_cong,
+             rule ccorres_return_Skip)
+     \<comment> \<open>Post wp proofs\<close>
+     apply vcg
+    apply (rule_tac Q'="\<lambda>rv s. tcb_at' thread s
+                               \<and> st_tcb_at' ((=) threadState) thread s
+                               \<and> invs' s \<and> weak_sch_act_wf (ksSchedulerAction s) s
+                               \<and> sym_refs_asrt s"
+                in hoare_strengthen_post)
```
Xaphiosis, line 3225: hard to tell in this interface, but the `in` should right-align with the `_tac`

proof/crefine/RISCV64/IpcCancel_C.thy
```diff
@@ -2948,198 +3035,256 @@ lemma cancelIPC_ccorres1:
...
-   apply (frule (1) ko_at_valid_ep'[OF _ invs_valid_objs'])
-   subgoal by (auto simp: obj_at'_def projectKOs pred_tcb_at'_def invs'_def valid_state'_def
-                        isTS_defs cte_wp_at_ctes_of isSendEP_def
-                        cthread_state_relation_def sch_act_wf_weak valid_ep'_def
-                   split: thread_state.splits endpoint.splits)[1]
-  apply (auto simp: isTS_defs cthread_state_relation_def typ_heap_simps weak_sch_act_wf_def)
-  apply (case_tac ts,
-           auto simp: isTS_defs cthread_state_relation_def typ_heap_simps)
-  done *)
+  subgoal by (auto simp: obj_at'_def pred_tcb_at'_def valid_ep'_def isTS_defs isSendEP_def
```
Xaphiosis, line 3283: ok, and final nitpick, which is the most annoying to fix up: the current auto-indenter shoves `subgoal` was off to the left, regardless of how deep you were in the proof, which requires manually indenting the `subgoal proof`, i.e. it should be:

```
    apply (rule conjI)
     apply (clarsimp simp: sym_refs_asrt_def invs'_implies)
     subgoal by (auto simp: obj_at'_def pred_tcb_at'_def valid_ep'_def isTS_defs isSendEP_def
                     split: thread_state.splits endpoint.splits)
    apply (clarsimp simp: sym_refs_asrt_def invs'_implies valid_objs'_valid_tcbs'
                          sym_ref_BlockedOnReceive_replyObject_linked)
```
because the subgoal discharges one subgoal. This is in the style guide, but the tool works against us here (there is of course a thread on the Isabelle mailing list about this that went nowhere).

Xaphiosis (approved):
From my side, after the other nitpicks, this is good to go.

proof/crefine/RISCV64/IpcCancel_C.thy
nspin: Done.

proof/crefine/RISCV64/IpcCancel_C.thy
nspin: Done, thanks.

proof/crefine/RISCV64/IpcCancel_C.thy
nspin: Done, thanks.

proof/crefine/RISCV64/IpcCancel_C.thy
nspin: Done.

nspin:
My assumption that the rhs preconditions for `replyUnlinkTcb_corres` would suffice as lhs preconditions for `reply_unlink_ccorres` was naive. I've added a commit to this PR that proves `reply_unlink_ccorres`, eliminating the need for guesswork there.

As this corner of CRefine depends on https://github.com/seL4/seL4/pull/1297, so too does this PR. For now, I've changed the "Test with" link in this PR to point to a branch (https://github.com/coliasgroup/seL4/pull/1) which includes https://github.com/seL4/seL4/pull/1297 and the additional minor changes in https://github.com/seL4/seL4/pull/1358.

proof/crefine/RISCV64/Ctac_lemmas_C.thy
```diff
@@ -46,6 +46,16 @@ lemma c_guard_abs_sc:
 
 lemmas ccorres_move_c_guard_sc[corres_pre] = ccorres_move_c_guards[OF c_guard_abs_sc]
 
+lemma c_guard_abs_reply:
+  fixes p :: "reply_C ptr"
+  shows "\<forall>s s'. (s, s') \<in> rf_sr \<and> reply_at' (ptr_val p) s \<and> True \<longrightarrow> s' \<Turnstile>\<^sub>c p"
+  apply clarsimp
+  apply (drule (1) reply_at_h_t_valid)
+  apply simp
+  done
```
Xaphiosis, line 55: probably can one-line it with `clarsimp dest: reply_at_h_t_valid` (or if that doesn't work, `dest!`, or if that doesn't work use `dest!` and fastforce)

proof/crefine/RISCV64/IpcCancel_C.thy
```diff
@@ -2913,12 +2913,120 @@ lemma reply_remove_tcb_ccorres:
     (replyRemoveTCB tptr) (Call reply_remove_tcb_'proc)"
 sorry (* FIXME RT: reply_remove_tcb_corres *)
 
+lemma updateReply_eq:
```
Xaphiosis, line 2916: Not a great name, `_eq` lemma name would imply some kind of equality between two things... but seeing the `_lemma4` below, it looks like you copied it from somewhere else?

nspin: Yes, I should have made that more clear when I pushed the commit. All of new lemmas in the `reply_unlink_ccorres` commit, aside from `reply_unlink_ccorres` itself, are derived (in some cases very closely) from analogous lemmas found elsewhere in CRefine.

Xaphiosis: Well, then I guess you're consistent with someone's historic poor naming convention, so it's fine to leave it. Also, apart from the potential optimisation, I'm also fine with your latest commit.

nspin: `updateReply_eq` is mostly copied from `updateSchedContext_eq`, and `updateReply_ccorres_lemma4` is mostly copied from `updateSchedContext_ccorres_lemma4`. I don't totally understand the `_ccorres_lemma[234]` naming convention, but I (perhaps recklessly) just went with it.

proof/crefine/RISCV64/Ctac_lemmas_C.thy
nspin: Done.

proof/crefine/RISCV64/IpcCancel_C.thy
Xaphiosis: I think that's the right move. The whole lemma4 thing has something to do with handling updates of structs within structs at a pointer location, but not sure of the exact history of the 2/3/4. If we ever update it to better tech, having the lemma4's around to grep for will be useful.

proof/crefine/RISCV64/Ctac_lemmas_C.thy
Xaphiosis: That was fast. Thanks

proof/crefine/RISCV64/IpcCancel_C.thy
nspin: I made that first change, but the second simplification does not work.

Xaphiosis: Well, it doesn't always work out. `;` processes new goals last-to-first so there might be a unification issue. Fine to leave that as is then.

lsf37:
> As this corner of CRefine depends on [seL4/seL4#1297](https://github.com/seL4/seL4/pull/1297), so too does this PR. For now, I've changed the "Test with" link in this PR to point to a branch ([coliasgroup/seL4#1](https://github.com/coliasgroup/seL4/pull/1)) which includes [seL4/seL4#1297](https://github.com/seL4/seL4/pull/1297) and the additional minor changes in [seL4/seL4#1358](https://github.com/seL4/seL4/pull/1358).

Have now updated seL4/seL4#1358 and changed the "Test with" line to point to it.

lsf37 (approved):
Thanks, @nspin, it is nice to get a proof contribution :-)

This is ready to merge from my side once tests have passed.

proof/crefine/RISCV64/IpcCancel_C.thy
lsf37: In general I actually do prefer rewrites, and maybe and `[THEN iffD1]` or similar to get a destruction rule, but the existentials make that a bit awkward in this case. (So I'm happy with the new version -- just wanted to point out that equalities are not a bad thing).

proof/crefine/RISCV64/IpcCancel_C.thy
```diff
@@ -2935,6 +3043,88 @@ lemma reply_remove_ccorres:
...
+    by (clarsimp simp: st_tcb_at'_def obj_at'_def)
+  with sy and vo and st and ra show ?thesis
+    apply (clarsimp simp: st_tcb_at'_def)
+    apply (frule (1) sym_ref_Receive_or_Reply_replyTCB')
+    apply simp
+    apply (clarsimp simp: obj_at'_def)
+    done
+qed
+
+lemma cancelIPC_threadSet_wp:
```
michaelmcinerney, line 3104: It seems unusual to have this separate from the main proof. Is there any particular reason it is here? I think it could be put into the main proof pretty easily. 

And I thought that `sym_refs_asrt_def` was in the global `[simp]` set, so you shouldn't need to unfold it. But if not, it may be easiest to write `unfolding sym_refs_asrt_def` at the beginning of the main proof, which should unfold it everywhere for you.

michaelmcinerney (approved):
Looks great! Very impressive

lsf37:
The proof test has now failed after the update, possibly because of changes that happened in the meantime.

nspin:
> The proof test has now failed after the update, possibly because of changes that happened in the meantime.
> 
> @nspin if you have time to look at it, that would be great. If not I can probably take that on.

Yes, I will update this PR next week.

michaelmcinerney:
I have rebased this over the current `rt` branch, fixed the errors, made a few minor style adjustments, generalised the handler stack in the new proofs to be schematic, reduced warnings, and removed the separately stated Hoare triple lemma. I've also removed the "Test with" line in the PR description, since this PR should now pass with seL4 `master`.

PR #833, opened 2024-12-04
arch-split Refine up to Invariants_H for AARCH64 (Draft)

Xaphiosis:
🦆🦆🦆 This PR needs feedback before I deploy anything to the other arches or move further. In an attempt to make the diff a bit more useful for reviewing than "new files added with stuff", `Invariants_H` got duplicated into each of the new files, and then clobbered by the actual file contents.
**Do not attempt to review the whole thing at once**, or the "duplicate and move Invariants_H" commit which will be squashed into once all this works.
The Refine fixup commit is not really indicative of any policies I'm aiming for in the future, it's just there so Refine builds.
The plan is: once we are OK with the way things are done, I'll expand this PR with deployment to other arches, fix up CRefine, and then we can merge it to master before continuing the arch-split train.

Invariants_H gets split into 4 files processed in this order:
    - InvariantsPre_H: initial setup
    - ArchInvsDefs_H: arch-specific invariant property definitions
    - Invariants_H: remains to store arch-generic
      invariants/properties/lemmas
    - ArchInvsLemmas: locale interpretations and arch-specific
      invariant-related lemmas

Several lemmas/interpretations/locales got moved to InvariantUpdates_H in order to resolve dependencies without introducing more files.

AARCH64 Refine will build, but CRefine is currently broken.

### DIFF
--- a/proof/crefine/Move_C.thy
+++ b/proof/crefine/Move_C.thy
@@ -131,13 +131,11 @@ lemma length_Suc_0_conv:
 
 lemma imp_ignore: "B \<Longrightarrow> A \<longrightarrow> B" by blast
 
-lemma cteSizeBits_eq:
-  "cteSizeBits = cte_level_bits"
-  by (simp add: cte_level_bits_def cteSizeBits_def)
+lemmas cteSizeBits_eq = cteSizeBits_cte_level_bits
 
 lemma cteSizeBits_le_cte_level_bits[simp]:
   "cteSizeBits \<le> cte_level_bits"
-  by (simp add: cte_level_bits_def cteSizeBits_def)
+  by (simp add: cteSizeBits_eq)
 
 lemma unat_ucast_prio_mask_simp[simp]:
   "unat (ucast (p::priority) && mask m :: machine_word) = unat (p && mask m)"
@@ -344,6 +342,20 @@ lemma word_eq_cast_unsigned:
   "(x = y) = (UCAST ('a signed \<rightarrow> ('a :: len)) x = ucast y)"
   by (simp add: word_eq_iff nth_ucast)
 
+(* tcbIPCBufferSlot is last slot in TCB *)
+(* FIXME arch-split: Arch is needed for wordSizeCase, proof is same on all arches *)
+lemma (in Arch) cteSizeBits_2ptcbBlockSizeBits[simplified tcbIPCBufferSlot_def]:
+  "n \<le> tcbIPCBufferSlot \<Longrightarrow> n << cteSizeBits < 2 ^ tcbBlockSizeBits"
+  unfolding tcbIPCBufferSlot_def tcbBlockSizeBits_def cteSizeBits_def
+  apply (simp only: wordSizeCase_simp)
+  apply (rule shiftl_less_t2n; simp)
+  apply unat_arith
+  done
+
+requalify_facts Arch.cteSizeBits_2ptcbBlockSizeBits
+
+lemmas zero_less_2p_tcbBlockSizeBits = cteSizeBits_2ptcbBlockSizeBits[where n=0, simplified]
+
 lemma ran_tcb_cte_cases:
   "ran tcb_cte_cases =
    {(Structures_H.tcbCTable, tcbCTable_update),
@@ -351,7 +363,8 @@ lemma ran_tcb_cte_cases:
     (Structures_H.tcbReply, tcbReply_update),
     (Structures_H.tcbCaller, tcbCaller_update),
     (tcbIPCBufferFrame, tcbIPCBufferFrame_update)}"
-  by (auto simp add: tcb_cte_cases_def cteSizeBits_def split: if_split_asm)
+  unfolding tcb_cte_cases_def
+  by (auto split: if_split_asm simp: tcb_cte_cases_neqs)
 
 lemma user_data_at_ko:
   "typ_at' UserDataT p s \<Longrightarrow> ko_at' UserData p s"
@@ -568,7 +581,7 @@ lemma threadGet_again:
    (threadGet ext t >>= n) s' = n rv s'"
   apply (clarsimp simp add: threadGet_def liftM_def in_monad)
   apply (frule use_valid [OF _ getObject_obj_at'])
-     apply (simp add: objBits_simps')+
+     apply (simp add: gen_objBits_simps tcbBlockSizeBits_def)+ (* FIXME arch-split: tcbBlockSizeBits *)
   apply (frule getObject_tcb_det)
   apply (clarsimp simp: bind_def split_def)
   apply (insert no_fail_getObject_tcb)
@@ -623,15 +636,16 @@ lemma mapM_only_length:
 
 lemma tcb_aligned':
   "tcb_at' t s \<Longrightarrow> is_aligned t tcbBlockSizeBits"
-  apply (drule obj_at_aligned')
-   apply (simp add: objBits_simps)
-  apply (simp add: objBits_simps)
-  done
+  by (drule obj_at_aligned'; simp add: gen_objBits_simps)
 
-lemma cte_at_0' [dest!]:
+(* identical proof on all arches *)
+lemma (in Arch) cte_at_0'[dest!]:
   "\<lbrakk> cte_at' 0 s; no_0_obj' s \<rbrakk> \<Longrightarrow> False"
   apply (clarsimp simp: cte_wp_at_obj_cases')
   by (auto simp: tcb_cte_cases_def is_aligned_def objBits_simps' dest!:tcb_aligned' split: if_split_asm)
+
+requalify_facts Arch.cte_at_0'
+lemmas [dest!] = cte_at_0'
 
 lemma getMessageInfo_le3:
   "\<lbrace>\<top>\<rbrace> getMessageInfo sender \<lbrace>\<lambda>rv s. unat (msgExtraCaps rv) \<le> 3\<rbrace>"
@@ -1119,45 +1133,54 @@ lemma updateObject_cte_tcb:
         return (KOTCB (updF (\<lambda>_. ctea) tcb))
     od)"
   using tc unfolding tcb_cte_cases_def
-  apply -
-  apply (clarsimp simp add: updateObject_cte Let_def
-    tcb_cte_cases_def objBits_simps' tcbSlots shiftl_t2n
-    split: if_split_asm cong: if_cong)
-  done
+  by (clarsimp simp: updateObject_cte Let_def tcb_cte_cases_neqs gen_objBits_simps tcbSlots
+               split: if_split_asm
+               cong: if_cong)
+
+lemma self_elim:
+  "\<And>P Q. \<lbrakk> P \<Longrightarrow> Q; P \<rbrakk> \<Longrightarrow> Q"
+  by blast
 
 lemma tcb_cte_cases_in_range1:
   assumes tc:"tcb_cte_cases (y - x) = Some v"
   and     al: "is_aligned x tcbBlockSizeBits"
   shows   "x \<le> y"
 proof -
-  note objBits_defs [simp]
-
   from tc obtain q where yq: "y = x + q" and qv: "q < 2 ^ tcbBlockSizeBits"
     unfolding tcb_cte_cases_def
-    by (simp add: diff_eq_eq split: if_split_asm)
+    by (clarsimp simp: diff_eq_eq tcbSlot_defs
+                 simp del: shiftl_1
+                 elim!: self_elim
+                 intro!: cteSizeBits_2ptcbBlockSizeBits zero_less_2p_tcbBlockSizeBits
+                 split: if_split_asm)
 
   have "x \<le> x + 2 ^ tcbBlockSizeBits - 1" using al
     by (rule is_aligned_no_overflow)
 
   hence "x \<le> x + q" using qv
-    apply simp
-    apply unat_arith
-    apply simp
-    done
+    by unat_arith
 
   thus ?thesis using yq by simp
 qed
+
+lemma tcbBlockSizeBits_word_less_sub_le:
+  fixes x :: machine_word
+  shows "(x \<le> 2 ^ tcbBlockSizeBits - 1) = (x < 2 ^ tcbBlockSizeBits)"
+  unfolding tcbBlockSizeBits_def
+  by (subst word_less_sub_le; simp)
 
 lemma tcb_cte_cases_in_range2:
   assumes tc: "tcb_cte_cases (y - x) = Some v"
   and     al: "is_aligned x tcbBlockSizeBits"
   shows   "y \<le> x + 2 ^ tcbBlockSizeBits - 1"
 proof -
-  note objBits_defs [simp]
-
   from tc obtain q where yq: "y = x + q" and qv: "q \<le> 2 ^ tcbBlockSizeBits - 1"
     unfolding tcb_cte_cases_def
-    by (simp add: diff_eq_eq split: if_split_asm)
+    by (clarsimp simp: diff_eq_eq tcbSlot_defs tcbBlockSizeBits_word_less_sub_le
+                 simp del: shiftl_1
+                 elim!: self_elim
+                 intro!: cteSizeBits_2ptcbBlockSizeBits zero_less_2p_tcbBlockSizeBits
+                 split: if_split_asm)
 
   have "x + q \<le> x + (2 ^ tcbBlockSizeBits - 1)" using qv
     apply (rule word_plus_mono_right)
--- a/proof/refine/AARCH64/ArchInvsDefs_H.thy
+++ b/proof/refine/AARCH64/ArchInvsDefs_H.thy
@@ -1,0 +1,230 @@
+(*
+ * Copyright 2023, Proofcraft Pty Ltd
+ * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+ *
+ * SPDX-License-Identifier: GPL-2.0-only
+ *)
+
+(* This theory contains the arch-dependent part of invariant definitions.
+   Due to the large amount of definitions, we place it before Invariants_H rather than fixing
+   constants in locales. *)
+
+theory ArchInvsDefs_H
+imports
+  InvariantsPre_H
+  ArchMove_R
+begin
+
+(* disambiguate name clash between Arch and non-arch definitions with same names *)
+context Arch begin
+
+lemmas haskell_crunch_def[crunch_def] =
+  deriveCap_def finaliseCap_def
+  hasCancelSendRights_def sameRegionAs_def isPhysicalCap_def
+  sameObjectAs_def updateCapData_def maskCapRights_def
+  createObject_def capUntypedPtr_def capUntypedSize_def
+  performInvocation_def decodeInvocation_def
+
+context begin global_naming global
+requalify_facts (aliasing)
+  Retype_H.deriveCap_def Retype_H.finaliseCap_def
+  Retype_H.hasCancelSendRights_def Retype_H.sameRegionAs_def Retype_H.isPhysicalCap_def
+  Retype_H.sameObjectAs_def Retype_H.updateCapData_def Retype_H.maskCapRights_def
+  Retype_H.createObject_def Retype_H.capUntypedPtr_def Retype_H.capUntypedSize_def
+  Retype_H.performInvocation_def Retype_H.decodeInvocation_def
+end
+end
+
+context Arch begin arch_global_naming
+
+declare lookupPTSlotFromLevel.simps[simp del]
+declare lookupPTFromLevel.simps[simp del]
+
+abbreviation vcpu_at' :: "obj_ref \<Rightarrow> kernel_state \<Rightarrow> bool" where
+  "vcpu_at' \<equiv> typ_at' (ArchT VCPUT)"
+
+abbreviation pte_at' :: "obj_ref \<Rightarrow> kernel_state \<Rightarrow> bool" where
+  "pte_at' \<equiv> typ_at' (ArchT PTET)"
+
+(* hyp_refs *)
+
+definition tcb_vcpu_refs' :: "machine_word option \<Rightarrow> (obj_ref \<times> reftype) set" where
+  "tcb_vcpu_refs' t \<equiv> set_option t \<times> {TCBHypRef}"
+
+definition tcb_hyp_refs' :: "arch_tcb \<Rightarrow> (obj_ref \<times> reftype) set" where
+  "tcb_hyp_refs' t \<equiv> tcb_vcpu_refs' (AARCH64_H.atcbVCPUPtr t)"
+
+definition vcpu_tcb_refs' :: "obj_ref option \<Rightarrow> (obj_ref \<times> reftype) set" where
+  "vcpu_tcb_refs' t \<equiv> set_option t \<times> {HypTCBRef}"
+
+definition refs_of_a' :: "arch_kernel_object \<Rightarrow> (obj_ref \<times> reftype) set" where
+  "refs_of_a' x \<equiv> case x of
+     AARCH64_H.KOASIDPool asidpool \<Rightarrow> {}
+   | AARCH64_H.KOPTE pte \<Rightarrow> {}
+   | AARCH64_H.KOVCPU vcpu \<Rightarrow> vcpu_tcb_refs' (AARCH64_H.vcpuTCBPtr vcpu)"
+
+definition arch_live' :: "arch_kernel_object \<Rightarrow> bool" where
+  "arch_live' ao \<equiv> case ao of
+     AARCH64_H.KOVCPU vcpu \<Rightarrow> bound (AARCH64_H.vcpuTCBPtr vcpu)
+   | _ \<Rightarrow>  False"
+
+definition hyp_live' :: "kernel_object \<Rightarrow> bool" where
+  "hyp_live' ko \<equiv> case ko of
+     (KOTCB tcb) \<Rightarrow> bound (AARCH64_H.atcbVCPUPtr (tcbArch tcb))
+   | (KOArch ako) \<Rightarrow> arch_live' ako
+   |  _ \<Rightarrow> False"
+
+primrec azobj_refs' :: "arch_capability \<Rightarrow> obj_ref set" where
+  "azobj_refs' (ASIDPoolCap _ _) = {}"
+| "azobj_refs' ASIDControlCap = {}"
+| "azobj_refs' (FrameCap _ _ _ _ _) = {}"
+| "azobj_refs' (PageTableCap _ _ _) = {}"
+| "azobj_refs' (VCPUCap v) = {v}"
+
+lemma azobj_refs'_only_vcpu:
+  "(x \<in> azobj_refs' acap) = (acap = VCPUCap x)"
+  by (cases acap) auto
+
+
+section "Valid caps and objects (design spec)"
+
+primrec acapBits :: "arch_capability \<Rightarrow> nat" where
+  "acapBits (ASIDPoolCap _ _)       = asidLowBits + word_size_bits"
+| "acapBits ASIDControlCap          = asidHighBits + word_size_bits"
+| "acapBits (FrameCap _ _ sz _ _)   = pageBitsForSize sz"
+| "acapBits (PageTableCap _ pt_t _) = table_size pt_t"
+| "acapBits (VCPUCap v)             = vcpuBits"
+
+definition page_table_at' :: "pt_type \<Rightarrow> obj_ref \<Rightarrow> kernel_state \<Rightarrow> bool" where
+ "page_table_at' pt_t p \<equiv> \<lambda>s.
+    is_aligned p (ptBits pt_t) \<and>
+    (\<forall>i \<le> mask (ptTranslationBits pt_t). pte_at' (p + (i << pte_bits)) s)"
+
+lemmas vspace_table_at'_defs = page_table_at'_def
+
+abbreviation asid_pool_at' :: "obj_ref \<Rightarrow> kernel_state \<Rightarrow> bool" where
+  "asid_pool_at' \<equiv> typ_at' (ArchT ASIDPoolT)"
+
+definition asid_wf :: "asid \<Rightarrow> bool" where
+  "asid_wf asid \<equiv> asid \<le> mask asid_bits"
+
+definition wellformed_mapdata' :: "asid \<times> vspace_ref \<Rightarrow> bool" where
+  "wellformed_mapdata' \<equiv> \<lambda>(asid, vref). 0 < asid \<and> asid_wf asid \<and> vref \<in> user_region"
+
+definition wellformed_acap' :: "arch_capability \<Rightarrow> bool" where
+  "wellformed_acap' ac \<equiv>
+   case ac of
+     ASIDPoolCap r asid \<Rightarrow> is_aligned asid asid_low_bits \<and> asid_wf asid
+   | FrameCap r rghts sz dev  mapdata \<Rightarrow>
+       case_option True wellformed_mapdata' mapdata \<and>
+       case_option True (swp vmsz_aligned sz \<circ> snd) mapdata
+   | PageTableCap pt_t r (Some mapdata) \<Rightarrow> wellformed_mapdata' mapdata
+   | _ \<Rightarrow> True"
+
+lemmas wellformed_acap'_simps[simp] = wellformed_acap'_def[split_simps arch_capability.split]
+
+definition frame_at' :: "obj_ref \<Rightarrow> vmpage_size \<Rightarrow> bool \<Rightarrow> kernel_state \<Rightarrow> bool" where
+  "frame_at' r sz dev s \<equiv>
+     \<forall>p < 2 ^ (pageBitsForSize sz - pageBits).
+       typ_at' (if dev then UserDataDeviceT else UserDataT) (r + (p << pageBits)) s"
+
+definition valid_arch_cap_ref' :: "arch_capability \<Rightarrow> kernel_state \<Rightarrow> bool" where
+  "valid_arch_cap_ref' ac s \<equiv> case ac of
+     ASIDPoolCap r as \<Rightarrow> typ_at' (ArchT ASIDPoolT) r s
+   | ASIDControlCap \<Rightarrow> True
+   | FrameCap r rghts sz dev mapdata \<Rightarrow> frame_at' r sz dev s
+   | PageTableCap r pt_t mapdata \<Rightarrow> page_table_at' pt_t r s
+   | VCPUCap r \<Rightarrow> vcpu_at' r s"
+
+lemmas valid_arch_cap_ref'_simps[simp] =
+  valid_arch_cap_ref'_def[split_simps arch_capability.split]
+
+definition valid_arch_cap' :: "arch_capability \<Rightarrow> kernel_state \<Rightarrow> bool" where
+  "valid_arch_cap' cap \<equiv> \<lambda>s. wellformed_acap' cap \<and> valid_arch_cap_ref' cap s"
+
+lemmas valid_arch_cap'_simps[simp] =
+  valid_arch_cap'_def[unfolded wellformed_acap'_def valid_arch_cap_ref'_def,
+                      split_simps arch_capability.split, simplified]
+
+definition arch_cap'_fun_lift :: "(arch_capability \<Rightarrow> 'a) \<Rightarrow> 'a \<Rightarrow> capability \<Rightarrow> 'a" where
+  "arch_cap'_fun_lift P F c \<equiv> case c of ArchObjectCap ac \<Rightarrow> P ac | _ \<Rightarrow> F"
+
+lemmas arch_cap'_fun_lift_simps[simp] = arch_cap'_fun_lift_def[split_simps capability.split]
+
+definition valid_acap' :: "capability \<Rightarrow> kernel_state \<Rightarrow> bool" where
+  "valid_acap' \<equiv> arch_cap'_fun_lift valid_arch_cap' \<top>"
+
+definition is_device_frame_cap' :: "capability \<Rightarrow> bool" where
+  "is_device_frame_cap' cap \<equiv> case cap of ArchObjectCap (FrameCap _ _ _ dev _) \<Rightarrow> dev | _ \<Rightarrow> False"
+
+definition   valid_arch_tcb' :: "Structures_H.arch_tcb \<Rightarrow> kernel_state \<Rightarrow> bool" where
+  "valid_arch_tcb' \<equiv> \<lambda>t s. \<forall>v. atcbVCPUPtr t = Some v \<longrightarrow> vcpu_at' v s "
+
+definition valid_vcpu' :: "vcpu \<Rightarrow> bool" where
+  "valid_vcpu' v \<equiv> case vcpuTCBPtr v of None \<Rightarrow> True | Some vt \<Rightarrow> is_aligned vt tcbBlockSizeBits"
+
+(* This is a slight abuse of "canonical_address". What we really need to know for ADT_C in CRefine
+   is that the top pageBits bits of TablePTEs have a known value, because we shift left by pageBits.
+   What we actually know is that this is a physical address, so it is bound by the physical address
+   space size, which depending on config can be 40, 44, or 48. 48 happens to also be the bound for
+   the virtual address space, which canonical_address is for. This is good enough for our purposes. *)
+definition ppn_bounded :: "pte \<Rightarrow> bool" where
+  "ppn_bounded pte \<equiv> case pte of PageTablePTE ppn \<Rightarrow> canonical_address ppn | _ \<Rightarrow> True"
+
+definition valid_arch_obj' :: "arch_kernel_object \<Rightarrow> bool" where
+  "valid_arch_obj' ako \<equiv> case ako of
+     KOPTE pte \<Rightarrow> ppn_bounded pte
+   | KOVCPU vcpu \<Rightarrow> valid_vcpu' vcpu
+   | _ \<Rightarrow> True"
+
+primrec
+  acapClass :: "arch_capability \<Rightarrow> capclass"
+where
+  "acapClass (ASIDPoolCap _ _)    = PhysicalClass"
+| "acapClass ASIDControlCap       = ASIDMasterClass"
+| "acapClass (FrameCap _ _ _ _ _) = PhysicalClass"
+| "acapClass (PageTableCap _ _ _) = PhysicalClass"
+| "acapClass (VCPUCap _) = PhysicalClass"
+
+definition
+  isArchFrameCap :: "capability \<Rightarrow> bool"
+where
+ "isArchFrameCap cap \<equiv> case cap of ArchObjectCap (FrameCap _ _ _ _ _) \<Rightarrow> True | _ \<Rightarrow> False"
+
+(* Addresses of all PTEs in a VSRoot table at p *)
+definition table_refs' :: "machine_word \<Rightarrow> machine_word set" where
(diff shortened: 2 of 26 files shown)


proof/refine/InvariantsPre_H.thy
```diff
@@ -5,45 +5,76 @@
...
 
 (* global data and code of the kernel, not covered by any cap *)
 axiomatization
   kernel_data_refs :: "word64 set"
 
-context Arch begin
 
-declare lookupPTSlotFromLevel.simps[simp del]
-declare lookupPTFromLevel.simps[simp del]
+section \<open>Locale Setup for Idempotence over kernel_state Field Updates\<close>
```
corlewis, line 25: Very nice job collecting and rationalising these locales!

Just trying to think of what might help us in the future, do we want a comment here with a quick explanation of what we mean by idempotence over field updates, and maybe with an explanation that the locales are actually populated and interpreted in Invariants_H? It would be harder to explain here, but it might also help to have a comment explaining that completing the lattice is helpful to avoid future 'landmines', even if some of the locales aren't used now.

More ambitious ideas that almost definitely aren't worth the time would be to have some sort of test that confirms that the lattice is complete, and to have a new tool that constructs the lattice given `Arch` and the base locales.

proof/refine/AARCH64/ArchInvsDefs_H.thy
```diff
@@ -1,35 +1,32 @@
 (*
- * Copyright 2023, Proofcraft Pty Ltd
+ * Copyright 2024, Proofcraft Pty Ltd
```
corlewis: I thought we want to keep the earlier dates? If I'm wrong with that, you could also change the copyright in `InvariantsPre_H`.

lsf37: This is a new file, it just looks like it changed, because Raf set it up so there is a more useful diff. I guess it is debatable whether the content is really new if it's mostly moved, but there is also enough new stuff in here to actually call this new material from scratch.

proof/refine/AARCH64/ArchInvsDefs_H.thy
```diff
@@ -1,35 +1,32 @@
...
  * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
  *
  * SPDX-License-Identifier: GPL-2.0-only
  *)
 
-theory Invariants_H
+(* This theory contains the arch-dependent part of invariant definitions.
+   Due to the large amount of definitins, we opt for a Pre file rather than fixing constants in
+   locales. *)
```
lsf37, line 10: definitins -> definitions

proof/refine/AARCH64/ArchInvsDefs_H.thy
```diff
@@ -39,123 +36,20 @@ end
...
-   "tcbSchedPrevs_of s \<equiv> tcbs_of' s |> tcbSchedPrev"
-
-abbreviation tcbSchedNexts_of :: "kernel_state \<Rightarrow> obj_ref \<Rightarrow> obj_ref option" where
-  "tcbSchedNexts_of s \<equiv> tcbs_of' s |> tcbSchedNext"
-
-abbreviation sym_heap_sched_pointers :: "global.kernel_state \<Rightarrow> bool" where
-  "sym_heap_sched_pointers s \<equiv> sym_heap (tcbSchedNexts_of s) (tcbSchedPrevs_of s)"
+(* FIXME arch-split: currently global in Invariants_H
+definition max_ipc_words :: machine_word where
+  "max_ipc_words \<equiv> capTransferDataSize + msgMaxLength + msgMaxExtraCaps + 2" *)
```
lsf37, line 52: I think this definition is the same on all architectures, so it is correct for it to be global. The ingredients may be arch specific, but how the total size is computed is generic.

proof/refine/AARCH64/ArchInvsDefs_H.thy
```diff
@@ -39,123 +36,20 @@ end
...
 
 definition tcb_cte_cases :: "machine_word \<rightharpoonup> ((tcb \<Rightarrow> cte) \<times> ((cte \<Rightarrow> cte) \<Rightarrow> tcb \<Rightarrow> tcb))" where
  "tcb_cte_cases \<equiv> [ 0 << cteSizeBits \<mapsto> (tcbCTable, tcbCTable_update),
```
lsf37: I would have expected this one to be a generic definition. Do we we have architectures that add TCB cap slots?

proof/refine/InvariantsPre_H.thy
lsf37, line 25: Part of the confusion in the comment may be that "idempotence" is not the right word. Idempotence means `f (f x) = f x`, i.e. applying more than once is the same as applying once. What we are proving here is more identity, i.e. f x = x, but with side conditions. The locales are identifying which updates are the identity function on specific predicates. For instance, when a function doesn't change the `ksPSpace` component of the state, that function is the identity on all predicates that only depend on `ksPSpace`.

+1 on adding where these are populated and interpreted, I think that would be a good idea.

The setup is good, though, I'm happy with making this whole setup a bit more principled.

proof/refine/Invariants_H.thy
```diff
@@ -223,21 +174,7 @@ primrec live0' :: "Structures_H.kernel_object \<Rightarrow> bool" where
...
-  "vcpu_tcb_refs' t \<equiv> set_option t \<times> {HypTCBRef}"
-
-definition refs_of_a' :: "arch_kernel_object \<Rightarrow> (obj_ref \<times> reftype) set" where
-  "refs_of_a' x \<equiv> case x of
-     AARCH64_H.KOASIDPool asidpool \<Rightarrow> {}
-   | AARCH64_H.KOPTE pte \<Rightarrow> {}
-   | AARCH64_H.KOVCPU vcpu \<Rightarrow> vcpu_tcb_refs' (AARCH64_H.vcpuTCBPtr vcpu)"
-
+(* FIXME arch-split: do we want to expose this, or put it in Arch? *)
 definition hyp_refs_of' :: "kernel_object \<Rightarrow> (obj_ref \<times> reftype) set" where
```
lsf37, line 178: I think we're exposing it AInvs as well, so I'd keep it set up the same way as there.

proof/refine/Invariants_H.thy
```diff
@@ -536,27 +369,6 @@ definition valid_ntfn' :: "Structures_H.notification \<Rightarrow> kernel_state
   | ActiveNtfn b \<Rightarrow> True)
   \<and> valid_bound_tcb' (ntfnBoundTCB ntfn) s"
 
-(* FIXME AARCH64: remove if unused at the end *)
-definition valid_mapping' :: "machine_word \<Rightarrow> vmpage_size \<Rightarrow> kernel_state \<Rightarrow> bool" where
```
lsf37, line 540: Thanks for removing it!

proof/refine/Invariants_H.thy
```diff
@@ -1317,12 +1081,11 @@ lemma objBitsKO_Data:
   "objBitsKO (if dev then KOUserDataDevice else KOUserData) = pageBits"
   by (simp add: objBits_def objBitsKO_def word_size_def)
 
-lemmas objBits_defs = tcbBlockSizeBits_def epSizeBits_def ntfnSizeBits_def cteSizeBits_def vcpuBits_def
-lemmas untypedBits_defs = minUntypedSizeBits_def maxUntypedSizeBits_def
-lemmas objBits_simps = objBits_def objBitsKO_def word_size_def archObjSize_def
-lemmas objBits_simps' = objBits_simps objBits_defs
+(* FIXME arch-split: why are these visible outside of Arch?! *)
+thm tcbBlockSizeBits_def epSizeBits_def word_size_def
```
lsf37, line 1085: I guess they always have been, because it was too much work. If we accept `word_size_def` as generic, then `epSizeBits_def` would be as well, but `tcbBlockSizeBits_def` should probably not be. I guess one thing that is happening is that this is not properly arch split in C either: additional TCB fields are just `ifdef`ed instead of actual arch fields. Doesn't mean we can't properly split it on the formalisation side, of course.

proof/refine/Invariants_H.thy
```diff
@@ -1289,14 +1055,12 @@ lemma ntfn_splits[split]:
...
              \<not> P (f3 x3))))"
-  by (case_tac ntfn; simp)+
+  by (rule ntfn.splits)+
+
 \<comment> \<open>---------------------------------------------------------------------------\<close>
 
 section "Lemmas"
 
-schematic_goal wordBits_def': "wordBits = numeral ?n" (* arch-specific consequence *)
-  by (simp add: wordBits_def word_size)
```
lsf37, line 1298: If we leave `word_size_def` generic, we can probably leave this one generic as well.

proof/refine/Invariants_H.thy
```diff
@@ -1390,27 +1150,10 @@ lemma obj_atI' [intro?]:
...
   "cte_at' p s \<equiv> \<exists>cte::cte. fst (getObject p s) = {(cte,s)}"
   by (simp add: cte_wp_at'_def)
 
-
-lemma tcb_cte_cases_simps[simp]:
-  "tcb_cte_cases 0  = Some (tcbCTable, tcbCTable_update)"
-  "tcb_cte_cases 32 = Some (tcbVTable, tcbVTable_update)"
-  "tcb_cte_cases 64 = Some (tcbReply, tcbReply_update)"
-  "tcb_cte_cases 96 = Some (tcbCaller, tcbCaller_update)"
-  "tcb_cte_cases 128 = Some (tcbIPCBufferFrame, tcbIPCBufferFrame_update)"
```
lsf37, line 1411: With the numbers unfolded, these would be arch specific (or at least mode/machine word size specific). If we could avoid unfolding `cteLevelBits` in the proofs, then the lemma would stay generic or would not even be needed. Not sure it's worth the hassle at this point, but might be worth a FIXME.

proof/refine/Invariants_H.thy
```diff
@@ -1800,26 +1483,6 @@ lemma obj_at'_pspaceI:
   "obj_at' t ref s \<Longrightarrow> ksPSpace s = ksPSpace s' \<Longrightarrow>  obj_at' t ref s'"
   by (auto intro!: projectKO_stateI simp: obj_at'_def ps_clear_def)
 
-lemma cte_wp_at'_pspaceI:
-  "\<lbrakk>cte_wp_at' P p s; ksPSpace s = ksPSpace s'\<rbrakk> \<Longrightarrow> cte_wp_at' P p s'"
-  apply (clarsimp simp add: cte_wp_at'_def getObject_def)
```
lsf37, line 1805: I would have expected this one also to be a generic lemma. Does it need any arch specific ingredients?

proof/refine/Invariants_H.thy
```diff
@@ -2148,109 +1778,23 @@ lemma tcb_ctes_clear:
...
-  assumes "ksPSpace s ptr = Some (KOCTE cte)"
-  assumes "is_aligned ptr cte_level_bits"
-  assumes "ps_clear ptr cte_level_bits s"
-  assumes "P cte"
-  shows "cte_wp_at' P ptr s"
-  using assms by (simp add: cte_wp_at_cases' cte_level_bits_def objBits_defs)
-
-lemma cte_wp_at_tcbI':
+(* FIXME arch-split: naming: do we want to use the opaque naming style of Invariants_AI_2? *)
+locale Invariants_H_cte_ats =
```
lsf37, line 1783: I like proper names better, at least if they make sense. Sometimes there is too much stuff in there for a good name, but `cte_ats` here is good.

lsf37: In this case, if we can make `tcb_cte_cases` generic, I suspect that this locale becomes unneccessary.

proof/refine/Invariants_H.thy
```diff
@@ -2437,44 +1941,6 @@ lemma typ_at_lift_valid_untyped':
...
-
-lemma typ_at_lift_frame_at':
-  assumes "\<And>T p. f \<lbrace>typ_at' T p\<rbrace>"
-  shows "f \<lbrace>frame_at' p sz d\<rbrace>"
-  unfolding frame_at'_def
-  by (wpsimp wp: hoare_vcg_all_lift hoare_vcg_const_imp_lift assms split_del: if_split)
-
-lemma typ_at_lift_valid_cap':
-  assumes P: "\<And>P T p. \<lbrace>\<lambda>s. P (typ_at' T p s)\<rbrace> f \<lbrace>\<lambda>rv s. P (typ_at' T p s)\<rbrace>"
-  shows      "\<lbrace>\<lambda>s. valid_cap' cap s\<rbrace> f \<lbrace>\<lambda>rv s. valid_cap' cap s\<rbrace>"
```
lsf37, line 2456: Could this one be built on top of a new (similar) lemma about `valid_arch_cap'`?

lsf37: (not sure it's really worth it here, because it's year another arch thing that needs to be proved first, unless the entire `typ_at_lift_valid_cap` goes into a later theory)

proof/refine/Invariants_H.thy
```diff
@@ -2675,10 +2118,48 @@ lemma valid_mdb_ctesI [intro]:
...
+  apply (simp add: ex_cte_cap_wp_to'_def)
+  apply (rule hoare_pre)
+   apply (erule hoare_use_eq_irq_node')
+   apply (rule hoare_vcg_ex_lift)
+   apply assumption
+  apply simp
+  done
+
+
+section "Idempotence over kernel_state field updates"
```
lsf37, line 2160: Idempotence -> Identity

proof/refine/AARCH64/ArchInvsLemmas_H.thy
```diff
@@ -2191,1200 +165,104 @@ lemma cte_wp_at_cases':
...
-
-lemma invs_pspace_distinct' [elim!]:
-  "invs' s \<Longrightarrow> pspace_distinct' s"
-  by (simp add: invs'_def valid_state'_def valid_pspace'_def)
-
-lemma invs_valid_pspace' [elim!]:
-  "invs' s \<Longrightarrow> valid_pspace' s"
-  by (simp add: invs'_def valid_state'_def)
+lemma cte_wp_atE'[consumes 1, case_names CTE TCB]:
+  assumes cte: "cte_wp_at' P ptr s"
```
lsf37, line 199: There is already a comment about sorting through this. My gut feeling is that if tcb_cte_cases can be kept generic, a lot of these can be generic.

proof/refine/AARCH64/ArchInvsLemmas_H.thy
```diff
@@ -2191,1200 +165,104 @@ lemma cte_wp_at_cases':
...
-lemma invs'_gsUserPages_update[simp]:
-  "invs' (gsUserPages_update f s') = invs' s'"
-  apply (clarsimp simp: invs'_def valid_state'_def valid_bitmaps_def bitmapQ_defs
-                        valid_irq_node'_def valid_irq_handlers'_def irq_issued'_def irqs_masked'_def
-                        valid_machine_state'_def cur_tcb'_def)
-  apply (cases "ksSchedulerAction s'";
-         simp add: ct_in_state'_def tcb_in_cur_domain'_def ct_idle_or_in_cur_domain'_def
-                   ct_not_inQ_def)
-  done
+(* arch-specific consequences of idempotence over kernel_state field updates *)
```
lsf37, line 250: identity

proof/refine/AARCH64/ArchInvsLemmas_H.thy
```diff
@@ -3449,99 +327,171 @@ lemma range_cover_canonical_address:
...
-   \<Longrightarrow> obj_range' a obj \<inter> obj_range' b obj' = {}"
-  apply (frule_tac x=a in pspace_alignedD')
-   apply assumption
-  apply (frule_tac x=b in pspace_alignedD')
-   apply assumption
-  apply (frule_tac p=a and p'=b in aligned_mask_range_cases)
-   apply assumption
-  apply (fastforce dest: distinct_obj_range'_not_subset
-                   simp: obj_range'_def)
+(* FIXME arch-split: do we want a gen_typ_at_lifts version? *)
```
lsf37, line 446: Would not be useful in the proofs, but might be useful in terms of making sure that each arch has all generic objects (in case we later ever add a generic object). So in terms of organising things, probably yes, but can be a FIXME for now.

proof/refine/AARCH64/ArchInvsLemmas_H.thy
```diff
@@ -3449,99 +327,171 @@ lemma range_cover_canonical_address:
...
+lemma page_table_pte_atI':
+  "\<lbrakk> page_table_at' pt_t p s; i \<le> mask (ptTranslationBits pt_t) \<rbrakk> \<Longrightarrow>
+   pte_at' (p + (i << pte_bits)) s"
+  by (simp add: page_table_at'_def)
+
+lemmas bit_simps' = pteBits_def asidHighBits_def asidPoolBits_def asid_low_bits_def
+                    asid_high_bits_def bit_simps
+
+(* FIXME arch-split: do we want to export the gen ones? *)
+lemma objBitsT_simps:
```
lsf37, line 471: Yes, I think that would be useful in proofs about those specific generic object types.

proof/refine/AARCH64/InvariantUpdates_H.thy
```diff
@@ -5,11 +5,92 @@
...
+
+end
+
+
 (* FIXME: use locales to shorten this work *)
 
+(* FIXME arch-split: MOVE, or try make generic *)
+context Arch begin arch_global_naming
+
+lemma invs'_gsCNodes_update[simp]:
```
lsf37, line 64: This one might bite for arches where things like valid_machie_state have more ingredients, but happy to leave it like this until it does bite. Maybe we're lucky.

lsf37:
The setup overall looks good from my side. Nice job with the update locales! I have minor comments only. If we can keep the main cte cases lemmas generic (they should be if we can abstract the lemma over the word size), then I think we'll save ourselves some pain down the road in CSpace theories that should be mostly generic, but won't be if tcb_cte_case is not.

proof/refine/Invariants_H.thy
lsf37: Thinking more about it, the main reason why we tend to unfold these is because it is automatic for Isabelle to decide that e.g. `0 != 32`, but it can't automatically decide that `0 * cteLevelBits != 1 * cteLevelBits`. There is a mechanism somewhere for turning a concrete distinct list into a set of inequality `simp` lemmas. We could use that here, and prove `distinct [0 * cteLevelBits, 1 * cteLevelBits, ...]` once and use that. Then Isabelle should be able to do symbolically what it now needs the numbers for. I'll try find what that tool is called.

proof/refine/AARCH64/ArchInvsDefs_H.thy
corlewis: Hmm, I guess there's trade-offs. If we're thinking of it as new content should it keep the Data61 copyright?

proof/refine/Invariants_H.thy
lsf37: Right, it's not a tool, and it's not merged yet, it's just an application of `simplified`. Here is an example: https://github.com/seL4/l4v/blob/dc2fd26954724fca79d175528726314de834605a/proof/crefine/ARM/CSpace_C.thy#L3064-L3088

proof/refine/AARCH64/ArchInvsDefs_H.thy
```diff
@@ -39,123 +36,20 @@ end
...
-abbreviation
+abbreviation vcpu_at' :: "obj_ref \<Rightarrow> kernel_state \<Rightarrow> bool" where
   "vcpu_at' \<equiv> typ_at' (ArchT VCPUT)"
 
 abbreviation pte_at' :: "obj_ref \<Rightarrow> kernel_state \<Rightarrow> bool" where
   "pte_at' \<equiv> typ_at' (ArchT PTET)"
 
-end
-
-record itcb' =
```
corlewis: Huh, not relevant for this PR but I'm mildly surprised that we don't have a Refine equivalent to AInvs' `iarch_tcb`. In AInvs that lets us reuse the `pred_tcb_at` machinery for vcpus as well.

proof/refine/AARCH64/ArchInvsDefs_H.thy
Xaphiosis: I believe so, I'm taking other people's stuff, rearranging it in a different way, and then adding my own stuff. So that bumps the Proofcraft copyright and keeps the one that made the old stuff.

proof/refine/Invariants_H.thy
Xaphiosis: I'll go with this phrasing

proof/refine/Invariants_H.thy
Xaphiosis: ok will remove FIXME

proof/refine/AARCH64/ArchInvsDefs_H.thy
Xaphiosis: thanks

proof/refine/InvariantsPre_H.thy
Xaphiosis: > +1 on adding where these are populated and interpreted, I think that would be a good idea.

What do you mean by this?

proof/refine/AARCH64/ArchInvsDefs_H.thy
corlewis: For my clarity, in general if a change is 'just' updating existing content then we don't want to bump the copyright, right? I know that there's a spectrum between updating and new content, but that was my understanding and what led to this comment initially.

corlewis: On the topic, how about `Invariants_H` in this PR? In my mind that one at least should keep the original copyright.

proof/refine/InvariantsPre_H.thy
corlewis: I assume that that was referring to my earlier [comment](https://github.com/seL4/l4v/pull/833#discussion_r1872729019), where I suggested "maybe with an explanation that the locales are actually populated and interpreted in Invariants_H".

proof/refine/AARCH64/ArchInvsDefs_H.thy
Xaphiosis: It's hard to know what the threshold is. We're adding requalifies, rearranging stuff with the locales, renaming the locales ... @lsf37 ?

corlewis: My understanding is that how large a change to existing content is is relevant to whether we add the Proofcraft copyright or not, but that once the copyright is there we want to keep the original date. I might be thinking about this wrong of course, but I actually can't think of a situation where we want to bump a copyright. We could maybe make it a range and say something like `Copyright 2023-2024` but I don't know if that actually benefits us in anyway.

corlewis: Now that I've overthought this so much, I'm actually back to questioning the original copyright I commented on, in `ArchInvsDefs_H.thy`. We're either treating it as new content, in which case it shouldn't have the Data61 copyright, or we're considering it as based on the old content from `Invariants_H.thy` and it should keep the existing copyright.

proof/refine/InvariantsPre_H.thy
lsf37: yeah, what Corey said. I didn't realise my comment would not show up under Corey's comment (it did in the review view, but apparently not here).

Xaphiosis:
I fixed up non-proof items (including reverting the copyright changes) and added an explanation for the locale hierarchy based partly on Gerwin's explanation. Now I'll be looking at the proof changes, which will take longer.

proof/refine/InvariantsPre_H.thy
```diff
@@ -0,0 +1,129 @@
...
+  predicates. For instance, when a function doesn't change the @{term ksPSpace} component of the
+  state, that function is the identity on all predicates that only depend on @{term ksPSpace}.\<close>
+
+text \<open>
+  Since locales can dynamically gain conclusions, we want to avoid situations where someone adds
+  a conclusion somewhere in the locale graph and is surprised that it does not appear for
+  extensions of that locale. For arch-split there is an extra complication, where the conclusions
+  might use arch-specific facts, thus needing to be in Arch.
+
+  This means that for every one the update locales, we have to add an additional Arch locale.
```
corlewis, line 38:
```suggestion
  This means that for every one of the update locales, we have to add an additional Arch locale.
```

Xaphiosis: Thanks, fixed, will include in next round.

proof/refine/InvariantsPre_H.thy
Xaphiosis: comments added

proof/refine/AARCH64/ArchInvsDefs_H.thy
Xaphiosis: copyrights reverted

proof/refine/AARCH64/ArchInvsDefs_H.thy
Xaphiosis: No, but other than AARCH64 and RISCV64, they do this:
```
184: "tcb_cte_cases \<equiv> [ 0 \<mapsto> (tcbCTable, tcbCTable_update),
185-                   16 \<mapsto> (tcbVTable, tcbVTable_update),
186-                   32 \<mapsto> (tcbReply, tcbReply_update),
187-                   48 \<mapsto> (tcbCaller, tcbCaller_update),
188-                   64 \<mapsto> (tcbIPCBufferFrame, tcbIPCBufferFrame_update) ]"
```

proof/refine/InvariantsPre_H.thy
Xaphiosis: Hopefully that's done, let me know if it explains sufficiently.

proof/refine/AARCH64/InvariantUpdates_H.thy
Xaphiosis: InvariantUpdates_H is not yet arch split, so will leave for this PR and see what happens later.

proof/refine/AARCH64/ArchInvsLemmas_H.thy
Xaphiosis: I managed to create one and put it in InvariantsUpdates_H. next round

proof/refine/Invariants_H.thy
Xaphiosis: I was able to prove this with better tcb_cte_cases infrastructure. next round

proof/refine/Invariants_H.thy
Xaphiosis: I don't know, `cte_wp_at_cases'` is pretty nasty and I don't have the time available to go through it carefully... not seeing a sensible solution inside.

proof/refine/Invariants_H.thy
Xaphiosis: I've deployed something, it'll need review once I push it

proof/refine/Invariants_H.thy
Xaphiosis: From MM discussion: I will try keep it internal as an extra guard rail, even though word_size itself is exported. If things get too bad, we can abort the guard rail.

proof/refine/AARCH64/ArchInvsDefs_H.thy
Xaphiosis: removed

proof/refine/AARCH64/ArchInvsLemmas_H.thy
Xaphiosis: done, but check how I did it, it's a bit height-reduced

proof/refine/AARCH64/ArchInvsLemmas_H.thy
Xaphiosis: do not see a way of making them generic, looked through them again, nothing jumps out for migration, so removed the browsing FIXME

proof/refine/Invariants_H.thy
Xaphiosis: done

proof/refine/Invariants_H.thy
Xaphiosis: Yes, because it is a locale assumption, once it is satisfied it becomes available in general theory. I do not know whether we need a separate lifter for `valid_arch_cap`. My current guess is we shouldn't need one.

proof/refine/Invariants_H.thy
Xaphiosis: done

proof/refine/Invariants_H.thy
Xaphiosis: Please see the eventual infrastructure and the FIXMEs about overlapping lemmas. It follows on from our discussion on MM about being unable to sort the conjuncts and remove the redundant ones, and I still don't know how to do that.

proof/refine/Invariants_H.thy
Xaphiosis: I removed this FIXME but still think that we should examine hiding tcbBlockSizeBits. Maybe in a separate PR as that'll change the spec and probably break things.

Xaphiosis:
I've attempted to address all the issues raised so far, responded to the comments and fixed up Refine so it builds again. I will try to fix up CRefine next while I'm waiting for round 2 of reviewing.

Xaphiosis:
AARCH64 CRefine should be fixed now. That took longer than expected.

proof/refine/InvariantsPre_H.thy
lsf37: The section title still has "Idempotence" in it

proof/refine/Invariants_H.thy
lsf37: Happy with that.

proof/refine/Invariants_H.thy
lsf37: No worries, fine to leave for another time when we have better infrastructure for it.

proof/refine/InvariantsPre_H.thy
lsf37: Apologies, it does not, I was staring at the old version. Happy with the new version!

lsf37: I'm happy with the new comments.

proof/refine/Invariants_H.thy
```diff
@@ -1800,8 +1485,70 @@ lemma obj_at'_pspaceI:
...
+
+lemma tcb_cte_cases_distinct:
+  "distinct [
+     tcbCTableSlot << cteSizeBits :: machine_word,
+     tcbVTableSlot << cteSizeBits,
+     tcbReplySlot << cteSizeBits,
+     tcbCallerSlot << cteSizeBits,
+     tcbIPCBufferSlot << cteSizeBits]"
+  (* FIXME arch-split: warning due to tcb_cte_cases_neqs overlapping *)
+  by (simp add: tcbSlot_defs tcb_cte_cases_neqs_n)
```
lsf37, line 1533: Cool. In the distant future, this could potentially be the form that we do not unfold further in other proofs and that the tcb cases then use everywhere. With that, there would be at least some hope that it is symbolic enough to not reduce to the same terms in `tcb_cte_cases_neqs` and therefore get rid of the warning (and the additional simplification work Isabelle has to do in each proof).

proof/refine/Invariants_H.thy
lsf37: :+1: Happy with that for now. To make it nice, we probably have to adjust some of these TCB/CTE proofs quite a bit, which may even reach into spec changes. Something for some time in the future, but not for the current arch split pass.

proof/refine/AARCH64/ArchInvsLemmas_H.thy
lsf37: I think that's fine. We'll have to remember to update the number in `gen_objBitsT_simps` when we add a generic object, but I think that's reasonable.

lsf37:
I'm happy with the second round and with the proof changes.

proof/refine/Invariants_H.thy
Xaphiosis: I'm not following by what you mean here. If you mean that we don't unfold the `tcbCTableSlot` etc so we end up only needing these versions, then that would be nice, and would indeed not cause duplication in the `_neqs` lemma, which does normalise things around `0` and `1`. Don't think `Distinct_Cmd` will help much with that, it will either give a result not in simp normal form, or if we use `[simplified]`, duplicates as before.

Xaphiosis: I pushed a fix for this, which also removes the FIXMEs. Playing around with `Distinct_Cmd` didn't really improve over `distinct []` and your attribute hack, but I guess it in theory could work around the attribute hack if it wasn't so awkward to use.

lsf37: > If you mean that we don't unfold the tcbCTableSlot etc

Yup, that's what I meant. I don't know where exactly the unfoldings come in, but in old code it could already be at the spec level. In any case, that's for another time.

Nice work on removing the duplicates!

proof/crefine/Move_C.thy
```diff
@@ -1119,45 +1133,54 @@ lemma updateObject_cte_tcb:
...
-    tcb_cte_cases_def objBits_simps' tcbSlots shiftl_t2n
-    split: if_split_asm cong: if_cong)
-  done
+  by (clarsimp simp: updateObject_cte Let_def tcb_cte_cases_neqs gen_objBits_simps tcbSlots
+               split: if_split_asm
+               cong: if_cong)
+
+lemma self_elim:
+  "\<And>P Q. \<lbrakk> P \<Longrightarrow> Q; P \<rbrakk> \<Longrightarrow> Q"
+  by blast
```
corlewis, line 1142: Out of curiosity, do you remember why this is needed?

Xaphiosis:
@corlewis I have squashed the commits resulting from PR discussion, so we're back to disjoint commits for your review:

1. clone Invariants_H
2. interesting arch-split updates to split Invariants_H
3. boring Refine updates to make Refine work
4. mostly boring updates to make CRefine work (some excitement in Move_C)

proof/crefine/Move_C.thy
Xaphiosis: It's needed in the two lemmas below, where we have that Isabellism of
```
[| blah ==> thesis; asm1; asm2 |] ==> thesis
```
and this rule is what I came up with to unscrew that. If there's a nicer way, please let me know. This pattern comes up when using `obtain` which is from Isar. The HOL way, we'd get an existential.

proof/refine/InvariantsPre_H.thy
corlewis: Same

proof/refine/Invariants_H.thy
```diff
@@ -3115,48 +2642,16 @@ lemma sch_act_wf_cases:
...
-
-lemma objBitsT_koTypeOf :
-  "(objBitsT (koTypeOf ko)) = objBitsKO ko"
-  apply (cases ko; simp add: objBits_simps objBitsT_simps)
-  apply (rename_tac arch_kernel_object)
-  apply (case_tac arch_kernel_object; simp add: archObjSize_def objBitsT_simps bit_simps')
-  done
-
-lemma typ_at_aligned':
-  "\<lbrakk> typ_at' tp p s \<rbrakk> \<Longrightarrow> is_aligned p (objBitsT tp)"
```
corlewis, line 3157: Is this no longer needed? It seems to have been removed completely.

Xaphiosis: correct, it's not mentioned again in l4v/proofs

proof/refine/Invariants_H.thy
```diff
@@ -3115,48 +2642,16 @@ lemma sch_act_wf_cases:
...
-  "objBitsT KernelDataT = pageBits"
-  "objBitsT (ArchT PTET) = word_size_bits"
-  "objBitsT (ArchT ASIDPoolT) = pageBits"
-  "objBitsT (ArchT VCPUT) = vcpuBits"
-  unfolding objBitsT_def makeObjectT_def
-  by (simp add: makeObject_simps objBits_simps bit_simps')+
-
-
-lemma objBitsT_koTypeOf :
-  "(objBitsT (koTypeOf ko)) = objBitsKO ko"
```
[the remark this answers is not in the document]
Xaphiosis: Good catch. Yes, that should become an interface lemma and be requalified, it's a generic consequence partly based arch-specific results, something we expect all arch stuff to conform to

proof/refine/AARCH64/ArchInvsLemmas_H.thy
```diff
@@ -5,1835 +5,36 @@
...
-  "typ_at' T p s \<Longrightarrow> ksPSpace s = ksPSpace s' \<Longrightarrow> typ_at' T p s'"
-  by (simp add: typ_at'_def ko_wp_at'_def ps_clear_def)
-
 lemma frame_at'_pspaceI:
   "frame_at' p sz d s \<Longrightarrow> ksPSpace s = ksPSpace s' \<Longrightarrow> frame_at' p sz d s'"
   by (simp add: frame_at'_def typ_at'_def ko_wp_at'_def ps_clear_def)
 
-lemma valid_cap'_pspaceI:
+lemma valid_cap'_pspaceI[Invariants_H_pspaceI_assms]:
   "s \<turnstile>' cap \<Longrightarrow> ksPSpace s = ksPSpace s' \<Longrightarrow> s' \<turnstile>' cap"
```
corlewis: Probably not worth doing and definitely not as part of this PR, but some of these `pspaceI_assms` lemmas feel like they would have the same proof on all architectures and could be generic if the interface was a bit lower down.

proof/refine/AARCH64/InvariantUpdates_H.thy
```diff
@@ -5,11 +5,102 @@
...
+  typ_at_lift_valid_cap'
+  valid_arch_tcb_lift'
+  tcb_at_cte_at'
+  gen_objBitsT_simps
+
+(* arch-specific typ_at_lifts in Arch *)
+lemmas gen_typ_at_lifts =
+         typ_at_lift_tcb' typ_at_lift_ep' typ_at_lift_ntfn' typ_at_lift_cte' typ_at_lift_cte_at'
+         typ_at_lift_valid_untyped' typ_at_lift_valid_cap' valid_bound_tcb_lift
+         valid_arch_tcb_lift'
```
corlewis, line 25: A note for the distant future, I think that this will be a problem in `rt` with the more complicated `typ_at_lifts` setup we have there. I think for that we might want to add locales and do the arch-split dance between `Invariants_H`  and `ArchInvsLemas_H`.

Probably not worth the effort, but in a perfect world it would be nice to pull that setup back to master to test out arch-splitting it there.

proof/refine/Invariants_H.thy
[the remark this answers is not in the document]
Xaphiosis: I will probably leave the `[simp]` for now, and re-examine when I get to the `declare`

proof/refine/AARCH64/InvariantUpdates_H.thy
```diff
@@ -5,11 +5,102 @@
  *)
 
 theory InvariantUpdates_H
-imports Invariants_H
+imports ArchInvsLemmas_H
 begin
 
+(* generic consequences which require arch-specific proofs *)
```
corlewis, line 11: Hmm, I understand that this is a convenient location due to the import graph, but these requalifies being put here does feel like a bit of a negative consequence of how we're doing this arch-split. They don't really have any connection with the rest of this file and feel pretty disconnected to `Invariants_H` and `ArchInvsLemmas_H`, which are the files that actually relate to them.

I don't have any better suggestions so am fine to leave them here for now, but I can imagine negative consequences going forwards, where we make changes or add new lemmas to `ArchInvsLemmas_H` and don't think to change here as well, resulting in a new, inconsistent approach.

corlewis (commented):
This looks good to me as well and I just have minor comments or notes for the future. Thanks for also taking the opportunity to move things around to be more localised and have a better structure overall!

proof/refine/AARCH64/InvariantUpdates_H.thy
Xaphiosis: I don't quite understand. What kind of setup does rt have? If it cheers you up, `gen_typ_at_lifts` is so far unused, it's only the consequence of hiding the arch-specific ones under `typ_at_lifts` in Arch, and I left this behind as a precaution.

proof/refine/AARCH64/InvariantUpdates_H.thy
Xaphiosis: I agree it's not ideal. The issue is the second best place to put them is inside the Arch file, and then there's a risk of de-sync between arches (not that bad in general, but not fun when doing the arch-split itself).
Another way is to use locales, which is something that I'll be aiming for for the rest of the files, but for invariants the dependency graph was a bit much. Maybe I can add a comment to clarify or something?

proof/refine/AARCH64/ArchInvsLemmas_H.thy
Xaphiosis: That's very possible, but it will require effort and I'm thinking it won't save all that many lines. We'll see if any other arch comes into conflict, I'll keep your comment in mind.

proof/refine/AARCH64/InvariantUpdates_H.thy
Xaphiosis: Given your comment and how it's unused, should I remove it to prevent a derp later?

corlewis: In `rt` we have a couple of locales that contain all of the `typ_at'` lifting rules and that are then interpreted as needed for any function that preserves `typ_at'`. See the whole section building up to https://github.com/seL4/l4v/blob/rt/proof/refine/RISCV64/Invariants_H.thy#L2777 if you're particularly curious. When it comes time for it to be arch-split, I imagine we'll want to keep as much of that generic as possible to reduce duplication and risk of de-sync.

corlewis: > Given your comment and how it's unused, should I remove it to prevent a derp later?

Hmm, sorry, it wasn't actually specifically `gen_typ_at_lifts` that I was commenting on, more the general approach with `typ_at_lifts` being in `Arch`. I just put this here as the place where some of it is made generic again.

Xaphiosis: In that case, they'll end up more like the `_upd` locale hierarchy? I.e. locales get declared, we do what we can in generic, then add on the rest in Arch, and then (sigh) any stragglers that need the Arch ones in the next generic theory ... hmm. I wish there was a nicer way.

proof/refine/AARCH64/ArchInvsDefs_H.thy
Xaphiosis: Then no one did such a thing in rt either?

corlewis: Nope, rt so far has only been for architectures without hypervisor support, so there's never been a need for `pred_tcb_at` of an `arch_tcb` property.

proof/refine/AARCH64/InvariantUpdates_H.thy
corlewis: Right, comparing it to the `_upd` locales makes sense and that does give us an approach to follow when we get up to that. Hopefully it ends up being not too bad.

proof/refine/AARCH64/InvariantUpdates_H.thy
corlewis: Ah right, I hadn't thought of this being specifically an artefact of the complicated invariants files.

proof/refine/AARCH64/ArchInvsDefs_H.thy
Xaphiosis: This strikes fear into my heart

PR #834, opened 2024-12-06
Style guide: _tac instantiations and attributes

Xaphiosis:
Adds two sections:
- Fact transformers and attributes (due to some experience with PRs this year)
- Wrapping/indenting complex rule instantiations, usually as part of `_tac` methods
  - as per https://github.com/seL4/l4v/issues/746

I also added consistent double-newline before sections, and added some more emphasis on 2 being the standard indent at the start of the indentation section.

Commits are independent so can be reviewed separately. Let the bikeshedding begin.

### DIFF
--- a/docs/Style.thy
+++ b/docs/Style.thy
@@ -360,6 +360,71 @@ term "
   A" \<comment> \<open>NOT OK: implies this parses as @{text "((A \<and> B) \<longrightarrow> B) \<or> A"}\<close>
 
 
+section \<open>Wrapping/indenting complex rule instantiations, usually as part of @{text "_tac"} methods\<close>
+
+text \<open>
+  This section concerns @{text "_tac"} methods, which are generally used when the instantiations
+  contain bound variables. When bound variables do no occur, prefer instantiations via attribute
+  (@{text "fact[where ...]"}) as they are more general.\<close>
+
+text \<open>
+  For simple instantiations which can fit on one line, style them as follows.
+  Notes:
+    * for single-variable instantiations, do not use quotes
+    * no space between @{text "="} and the instantiation or start of quote\<close>
+
+lemma
+  "\<lbrakk>a; b\<rbrakk> \<Longrightarrow> a \<and> b"
+  apply (frule_tac P=a and Q="id b" in conjI) \<comment> \<open>GOOD\<close>
+  apply (frule_tac P="a" and Q = "id b" in conjI) \<comment> \<open>BAD: unnecessary quotes, unnecessary spacing\<close>
+  oops
+
+text \<open>
+  However, when the instantiation is complex, the instantiations, @{text "and"} and @{text "in"}
+  need to be distributed over multiple lines.\<close>
+
+lemma conjI3:
+  "\<lbrakk>a; b; c\<rbrakk> \<Longrightarrow> a \<and> b \<and> c"
+  by simp
+
+text \<open>
+  For left operator-wrapping style, use this version. It was chosen based on being space-optimising
+  and nice-looking (variable instantiations and rule all left-align, while operators right-align):\<close>
+
+lemma \<comment> \<open>left operator-wrapping pretty version - preferred\<close>
+  "\<lbrakk> x; y; z \<rbrakk> \<Longrightarrow> x \<and> y \<and> z"
+  apply (drule_tac a=x
+               and b=y
+               and c=z
+                in conjI3)
+  oops
+
+text \<open>
+  For right-operator wrapping style, use this version. It still provides the alignment of variable
+  instantiations, but provides less horizontal space for the instantiation bodies themselves:\<close>
+
+lemma \<comment> \<open>right operator-wrapping pretty version - preferred\<close>
+  "\<lbrakk> x; y; z \<rbrakk> \<Longrightarrow> x \<and> y \<and> z"
+  apply (drule_tac a=x and
+                   b=y and
+                   c=z
+                   in conjI3) \<comment> \<open>this must not go on previous line in multi-line instantiations\<close>
+  oops
+
+text \<open>
+  There is one more left-operator wrappings style we permit, but only due to legacy proofs and the
+  possibility of being understandable/generatable by tools. Please do not use it on new
+  hand-written lemmas:\<close>
+
+lemma \<comment> \<open>left operator-wrapping pretty version for tools/legacy - permitted\<close>
+  "\<lbrakk> x; y; z \<rbrakk> \<Longrightarrow> x \<and> y \<and> z"
+  apply (drule_tac a=x
+                   and b=y
+                   and c=z
+                   in conjI3)
+  oops
+
+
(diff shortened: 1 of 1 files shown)


lsf37 (approved):
Happy with this colour of the bike shed as it is. Thanks for adding these.

docs/Style.thy
```diff
@@ -87,11 +88,17 @@ text \<open>
...
 
 text \<open>
-  Isabelle code is much easier to maintain when indented consistently. In apply style proofs we
-  indent by 2 spaces, and add an additional space for every additional subgoal.
+  Isabelle code is much easier to maintain when indented consistently.
+  When in doubt, and not constrained by vertically aligning items or subgoal-count offsets, use
+  2 spaces when indenting something with respect to its container (see ``General layout`` in
+  ``Other`` section).
+
+  In apply style proofs we indent by 2 spaces, and add an additional space for every additional
```
michaelmcinerney, line 100: I know this is how it was phrased in the previous version, but would it be a bit more clear to say after what's here something like "for example, a command which applies when there are 3 subgoals should be indented by 4 spaces".

michaelmcinerney (approved):
Looks good

docs/Style.thy
Xaphiosis: Sure, I can add that while we're here.

Xaphiosis: Added, let me know if it's clear now.

michaelmcinerney: Yep, thanks.

docs/Style.thy
```diff
@@ -87,11 +88,18 @@ text \<open>
...
   Indent text by 2 inside the @{command text} area. This achieves visual separation.\<close>
 
+
 section \<open>Indentation\<close>
 
 text \<open>
-  Isabelle code is much easier to maintain when indented consistently. In apply style proofs we
-  indent by 2 spaces, and add an additional space for every additional subgoal.
+  Isabelle code is much easier to maintain when indented consistently.
+  When in doubt, and not constrained by vertically aligning items or subgoal-count offsets, use
```
corlewis, line 96: Extremely minor grammar nitpick
```suggestion
  When in doubt and not constrained by vertically aligning items or subgoal-count offsets, use
```

docs/Style.thy
```diff
@@ -359,6 +359,66 @@ term "
   A" \<comment> \<open>NOT OK: implies this parses as @{text "((A \<and> B) \<longrightarrow> B) \<or> A"}\<close>
 
 
+section \<open>Wrapping/indenting complex rule instantiations, usually as part of @{text "_tac"} methods\<close>
+
+text \<open>
+  For simple instantiations, they can fit it on one line, like so.
```
corlewis, line 403: This sentence is a bit unclear to me. Are you saying that simple instantiations that can fit on one line should look like the following?

docs/Style.thy
```diff
@@ -359,6 +359,66 @@ term "
...
+  Notes:
+    * for single-variable instantiations, do not use quotes
+    * no space between @{text "="} and the instantiation or start of quote\<close>
+
+lemma
+  "\<lbrakk>a; b\<rbrakk> \<Longrightarrow> a \<and> b"
+  apply (frule_tac P=a and Q="id b" in conjI) \<comment> \<open>when @{text a} and @{text b} contain bound vars\<close>
```
corlewis, line 410: It's not clear to me what the intention of the comment is here. Is it about using the `_tac` form instead of `where`?

docs/Style.thy
```diff
@@ -226,6 +226,43 @@ lemma test_lemma3:
...
+
+lemmas attr3 = conj_absorb[ \<comment> \<open>note the @{text "["} to indicate transform\<close>
+                 THEN iffD2,
+                 OF conj_absorb[where A="A \<and> B \<and> C \<and> D \<and> E \<and> F \<and> G" for A B C D E F G,
+                                simplified],
+                 simplified] \<comment> \<open>extreme wrapping case\<close>
+
+text \<open>
+  There is an important principle here: telling apart transformed/attributed facts from unaltered
+  facts at a glance. In other words avoid:\<close>
```
corlewis, line 259: This is a very useful comment, good thinking to include it. People are generally much better at following rules if they know the intention behind it, as opposed to it just being some arbitrary thing that they need to remember.

corlewis (approved):
Some very minor bikeshedding, thanks for updating the style guide for these!

docs/Style.thy
Xaphiosis: Yes, that is correct. I want to avoid guiding people towards writing out the rule_tac version when a `fact[where ...]` will suffice. The rule_tac version is less generic, as it applies only to the `_tac`s, but the attribute version has many more uses.

docs/Style.thy
Xaphiosis: Yes. I should make that more explicit.

docs/Style.thy
Xaphiosis: *shrug* ... if it helps, I'll add the comma

corlewis: I'm suggesting to remove the comma, I don't think it's right.

Xaphiosis: Oh, yes, will do then

docs/Style.thy
corlewis: I agree with that goal but then I think that the comment needs to be more explicit. As it is the mention of bound variables seems to come out of nowhere and the alternative attribute version isn't mentioned.

However, I'm then struggling to think of a way to keep it short. Would it make sense to add it as a note instead, with something vaguely like
`@{text "_tac"} methods as shown here are generally used when the instantiations contain bound vars, otherwise instantiations with the @{text "[where ...]" attribute are preferred as they are more general.`

Xaphiosis:
I fixed up non-proof items (including reverting the copyright changes) and added an explanation for the locale hierarchy based partly on Gerwin's explanation. Now I'll be looking at the proof changes, which will take longer.

docs/Style.thy
Xaphiosis: Ok, tuned, how does the latest commit look to you?

docs/Style.thy
```diff
@@ -351,6 +396,72 @@ term "
...
+
+section \<open>Wrapping/indenting complex rule instantiations, usually as part of @{text "_tac"} methods\<close>
+
+text \<open>
+  This section concerns @{text "_tac"} methods, which are generally used when the instantiations
+  contain bound variables. When bound variables do no occur, prefer instantiations via attribute
+  (@{text "fact[where ...]"}) as they are more general.\<close>
+
+text \<open>
+  For simple instantiations which can fit it on one line, style them as follows.
```
corlewis, line 408:
```suggestion
  For simple instantiations which can fit on one line, style them as follows.
```

docs/Style.thy
corlewis: One potential typo, otherwise looks good to me!

PR #835, opened 2024-12-06
github: fix concurrency group syntax

lsf37:
- fix concurrency group syntax (needs a group entry)
- cancel started runs


Xaphiosis (approved):
Cool, this should reduce accidental load on CI infrastructure.

PR #836, opened 2024-12-10
Add `sorted_ipc_queues` invariant

michaelmcinerney:
Please see the commit messages for what is here. It might be best to review the files in the order in which they would appear in jedit. So for example, AInvs before Refine, and earlier files in AInvs before later files in AInvs.

As usual, there are many lemmas that we may want to move up, or to have in `master`, and I would be happy to do that. I have moved many of the lemmas around, and I'll try to indicate them with comments. The diff is not very accurate, and some of what appears as green or red isn't anything I have actually touched.

I'll try some things out to hopefully reduce the number of times I remove `comp_apply` from the simp set.

I have weakened `valid_ep'` and `valid_ntfn'` by removing some of the properties that can be crossed from the abstract state analogues. This was done because proving distinctness was a bit of a pain with the ordered insert. So I have added some asserts in the Haskell for distinctness of these queues.


proof/invariant-abstract/DetSchedInvs_AI.thy
```diff
@@ -3009,8 +3017,58 @@ lemma active_reply_scsE:
...
+
+lemma sorted_ipc_queues_endpoint_priority_ordered:
+  "\<lbrakk>sorted_ipc_queues s; ep_queues_of s ptr = Some q\<rbrakk> \<Longrightarrow> priority_ordered q (prios_of s)"
+  by (fastforce simp: sorted_ipc_queues_def dest: spec[where x=ptr])
+
+lemma sorted_ipc_queues_notification_priority_ordered:
+  "\<lbrakk>sorted_ipc_queues s; ntfn_queues_of s ptr = Some q\<rbrakk> \<Longrightarrow> priority_ordered q (prios_of s)"
+  by (fastforce simp: sorted_ipc_queues_def dest: spec[where x=ptr])
+
+declare opt_map_Some_comp[simp del]
```
michaelmcinerney, line 3051: I did decide in AInvs to remove `opt_map_Some_comp` from the simp set, but it ended up back in the simp set during Refine (in a way I haven't tracked down yet). I think it would be best to not have it in the simp set at all.

proof/invariant-abstract/IpcDet_AI.thy
```diff
@@ -93,28 +93,304 @@ lemma sfi_tcb_at [wp]:
...
+lemma mapM_gets_the_sp:
+  "\<lbrace>P\<rbrace> mapM (gets_the \<circ> f) ts \<lbrace>\<lambda>rv s. P s \<and> map (\<lambda>t. f t s) ts = map Some rv\<rbrace>"
+  by (wpsimp wp: mapM_gets_the_wp simp: comp_def)
+
+lemma mapM_gets_the_sp':
+  "\<lbrace>P\<rbrace> mapM (gets_the \<circ> f) ts \<lbrace>\<lambda>rv s. P s \<and> (\<forall>i < length rv. f (ts ! i) s = Some (rv ! i))\<rbrace>"
+  apply (wpsimp wp: mapM_gets_the_wp simp: comp_def)
+  by (simp add: map_equality_iff)
+
+lemma distinct_map_fst_filter_zip:
```
michaelmcinerney, line 124: Everything between here and `opt_ord_R` is old and was moved from `DetSchedSchedule`

proof/invariant-abstract/IpcDet_AI.thy
```diff
@@ -93,28 +93,304 @@ lemma sfi_tcb_at [wp]:
...
+lemma sorted_insort_filter:
+  assumes tot: "total {(x,y). cmp x y}"
+  assumes tr: "transp cmp"
+  assumes re: "reflp cmp"
+  assumes sorted: "sorted_wrt cmp xs"
+  shows "sorted_wrt cmp (insort_filter (\<lambda>x. cmp x z) z xs)"
+  apply (subst sorted_insort_filter_eq_insort_partition[OF tr sorted])
+  by (rule sorted_insort_partition[OF tot tr re sorted])
+
+fun opt_ord_R where
```
michaelmcinerney, line 266: This is a generalisation of the older `opt_ord`. I added the `_R` to the name and allowed it to take an arbitrary relation. Then I define `opt_ord` below as it was before

proof/invariant-abstract/IpcDet_AI.thy
```diff
@@ -93,28 +93,304 @@ lemma sfi_tcb_at [wp]:
...
+                            simp: obj_at_def img_ord_def)
+          apply (wpsimp wp: mapM_wp_inv no_fail_mapM_wp)+
+    apply (wpsimp wp: mapM_gets_the_wp thread_get_wp' simp: gets_the_thread_read')+
+  apply (clarsimp simp: obj_at_def)
+  apply (fastforce split: kernel_object.splits
+                    simp: is_tcb_def thread_read_def oliftM_def obind_def get_tcb_def
+                          map_equality_iff nth_equalityI)
+  done
+
+lemma map_fst_filter_zip_map_reduce:
```
michaelmcinerney, line 307: Lemmas from here until `total_opt_ord` are old and were just moved

lib/Lib.thy
```diff
@@ -2269,6 +2269,10 @@ lemma in_set_zip2:
   "(x, y) \<in> set (zip xs ys) \<Longrightarrow> y \<in> set ys"
   by (erule in_set_zipE)
 
+lemma set_zip_helper:
```
Xaphiosis, line 2272: This is a pretty generic name to go into lib; this looks something like a `in_set_zipD` based on the names of lemmas above

spec/haskell/src/SEL4/Object/TCB.lhs
```diff
@@ -1113,26 +1113,14 @@ On some architectures, the thread context may include registers that may be modi
...
->         then return [tptr]
->         else do
->             index <- tcbEPFindIndex tptr queue (length queue - 1)
->             return $ take index queue ++ [tptr] ++ drop index queue
+> tcbEPAppend tptr queue = do
+>     stateAssert (priority_ordered'_asrt queue) "queue must be ordered by priority"
+>     prio <- threadGet tcbPriority tptr
+>     prios <- mapM (threadGet tcbPriority) queue
+>     zprios <- return $ zip queue prios
+>     zprios' <- return $ filter (\(t, p) -> p >= prio) zprios ++ [(tptr, prio)] ++ filter (\(t, p) -> p < prio) zprios
```
Xaphiosis, line 1122: this looks much nicer than the tcbEPFindIndex way

spec/abstract/IpcCancel_A.thy
```diff
@@ -300,37 +300,20 @@ where
...
+definition tcb_ep_append :: "obj_ref \<Rightarrow> obj_ref list \<Rightarrow> (obj_ref list, 'z::state_ext) s_monad" where
+  "tcb_ep_append tptr qs \<equiv> do
+     prio \<leftarrow> thread_get tcb_priority tptr;
+     prios \<leftarrow> mapM (thread_get tcb_priority) qs;
+     zprios \<leftarrow> return $ zip qs prios;
+     zprios' \<leftarrow> return $ filter (\<lambda>(_, p). p \<ge> prio) zprios
+                         @ [(tptr, prio)]
+                         @ filter (\<lambda>(_, p). p < prio) zprios;
+     return (map fst zprios')
    od"
```
Xaphiosis, line 315: comprehension nitpick: if it's placing the tptr into a list such that it's sorted, does it still make sense to call it "append" without comment?

proof/invariant-abstract/ARM/ArchDetSchedAux_AI.thy
```diff
@@ -16,27 +16,91 @@ lemmas arch_machine_ops_valid_sched_pred[wp] =
...
+  "set_asid_pool ptr pool \<lbrace>\<lambda>s. P (prios_of s)\<rbrace>"
+  by (set_object_easy_cases def: set_asid_pool_def)
 
 lemma set_asid_pool_bound_sc_obj_tcb_at[wp]:
   "set_asid_pool ptr pool \<lbrace>valid_sched_pred_strong P\<rbrace>"
+  apply (rule hoare_lift_Pf[where f=ntfn_queues_of, rotated], wpsimp)
+  apply (rule hoare_lift_Pf[where f=ep_queues_of, rotated], wpsimp)
+  apply (rule hoare_lift_Pf[where f=prios_of, rotated], wpsimp)
+  apply (rule hoare_lift_Pf[where f=eps_of, rotated], wpsimp)
+  apply (rule hoare_lift_Pf[where f=ntfns_of, rotated], wpsimp)
```
Xaphiosis, line 85: if you had a lifting lemma for valid_sched_pred_strong, you could apply that here and then wp most of the conditions away; the hoare_lift_Pf way looks a bit uphill and happens 3 times in this file, so presumably also on RISCV64

proof/invariant-abstract/DetSchedInvs_AI.thy
Xaphiosis: That looks painful. I'm not really an expert on the construction and deployment of these, so I'll leave it to @lsf37. I can say that it should be possible to track down where it gets reintroduced in Refine and kick it out early. Maybe in Invariants_H?

spec/abstract/IpcCancel_A.thy
lsf37: Good point, it'd be more like an `insert`. I suspect the name hasn't changed in C either, but we could still do better in the specs.

lib/Lib.thy
lsf37: +1 for `in_set_zipD`

proof/crefine/RISCV64/IpcCancel_C.thy
```diff
@@ -2552,11 +2553,11 @@ lemma ep_ptr_set_queue_spec:
   oops
 
 lemma valid_ep_blockedD:
-  "\<lbrakk> valid_ep' ep s; (isSendEP ep \<or> isRecvEP ep) \<rbrakk> \<Longrightarrow> (epQueue ep) \<noteq> [] \<and> (\<forall>t\<in>set (epQueue ep). tcb_at' t s) \<and> distinct (epQueue ep)"
+  "\<lbrakk> valid_ep' ep s; (isSendEP ep \<or> isRecvEP ep) \<rbrakk>
+   \<Longrightarrow> (epQueue ep) \<noteq> [] \<and> (\<forall>t\<in>set (epQueue ep). tcb_at' t s)"
```
lsf37, line 2557: These were all there before, but if we're touching it, let's remove the extra parentheses:

```suggestion
  "\<lbrakk> valid_ep' ep s; isSendEP ep \<or> isRecvEP ep \<rbrakk>
   \<Longrightarrow> epQueue ep \<noteq> [] \<and> (\<forall>t\<in>set (epQueue ep). tcb_at' t s)"
```

proof/crefine/RISCV64/SR_lemmas_C.thy
```diff
@@ -1212,8 +1212,7 @@ lemma ntfn_blocked_in_queueD:
 (* MOVE *)
 lemma valid_ntfn_isWaitingNtfnD:
   "\<lbrakk> valid_ntfn' ntfn s; isWaitingNtfn (ntfnObj ntfn) \<rbrakk>
-  \<Longrightarrow> (ntfnQueue (ntfnObj ntfn)) \<noteq> [] \<and> (\<forall>t\<in>set (ntfnQueue (ntfnObj ntfn)). tcb_at' t s)
-    \<and> distinct (ntfnQueue (ntfnObj ntfn))"
+  \<Longrightarrow> (ntfnQueue (ntfnObj ntfn)) \<noteq> [] \<and> (\<forall>t\<in>set (ntfnQueue (ntfnObj ntfn)). tcb_at' t s)"
```
lsf37, line 1215: (also old, but nicer this way)

```suggestion
  \<Longrightarrow> ntfnQueue (ntfnObj ntfn) \<noteq> [] \<and> (\<forall>t\<in>set (ntfnQueue (ntfnObj ntfn)). tcb_at' t s)"
```

proof/invariant-abstract/Tcb_AI.thy
```diff
@@ -1737,31 +1737,29 @@ lemma inj_on_snd_set_zip_map:
   "distinct xs \<Longrightarrow> inj_on snd (set (zip (map f xs) xs))"
   using distinct_map by fastforce
 
+lemma monadic_rewrite_is_valid':
+  "\<lbrakk> \<And>x. monadic_rewrite False True (P' x) (g x) (g' x); \<lbrace>P\<rbrace> f \<lbrace>P'\<rbrace>;
+    \<lbrace>P\<rbrace> do x <- f; g' x od \<lbrace>Q\<rbrace> \<rbrakk>
+   \<Longrightarrow> \<lbrace>P\<rbrace> do x <- f; g x od \<lbrace>Q\<rbrace>"
+  by (fastforce simp: monadic_rewrite_def valid_def bind_def)
```
Xaphiosis, line 1744: This looks a bit like monadic_rewrite_bind immediately applies and would yield precisely the kind of goals which you wrote as assumptions. Bit confused how this is working.

proof/refine/RISCV64/CNodeInv_R.thy
```diff
@@ -6944,6 +6944,7 @@ proof (induct rule: rec_del.induct,
                       red_zombie_will_fail.simps ball_simps(5))
   case (1 slot exposed)
   show ?case
+    supply opt_map_Some_comp[simp del] comp_apply[simp del]
```
Xaphiosis, line 6947: since you're in a `proof` block, you can use `note` to mark these as `[simp del]` for the entire lemma, which will save the repetition

proof/invariant-abstract/ARM/ArchDetSchedAux_AI.thy
```diff
@@ -16,27 +16,91 @@ lemmas arch_machine_ops_valid_sched_pred[wp] =
...
+  "set_pd ptr pd \<lbrace>\<lambda>s. P (ntfns_of s)\<rbrace>"
+  by (set_object_easy_cases def: set_pd_def)
+
+lemma set_pt_prios_of[wp]:
+  "set_pt ptr pt \<lbrace>\<lambda>s. P (prios_of s)\<rbrace>"
+  by (set_object_easy_cases def: set_pt_def)
+
+lemma set_pd_prios_of[wp]:
+  "set_pd ptr pd \<lbrace>\<lambda>s. P (prios_of s)\<rbrace>"
+  by (set_object_easy_cases def: set_pd_def)
```
lsf37, line 41: We should be proving `P (tcbs_of s)` here (and wherever we can), because that's the more general case that includes `prios_of` and all other tcb projections.

proof/invariant-abstract/ARM/ArchDetSchedAux_AI.thy
```diff
@@ -16,27 +16,91 @@ lemmas arch_machine_ops_valid_sched_pred[wp] =
...
-  by (auto simp: obj_at_kh_kheap_simps vs_all_heap_simps a_type_def fun_upd_def
-           split:  kernel_object.splits if_splits)
+  apply (fastforce simp: obj_at_kh_kheap_simps vs_all_heap_simps a_type_def
+                  split: kernel_object.splits if_splits)
+  done
+
+crunch copy_global_mappings
+  for eps_of[wp]: "\<lambda>s. P (eps_of s)"
+  and ntfns_of[wp]: "\<lambda>s. P (ntfns_of s)"
+  and prios_of[wp]: "\<lambda>s. P (prios_of s)"
```
lsf37, line 94: Since these 3 always occur together: I wonder if there is a combinator that we could build so that we only have to prove one lemma and then apply a rule or something else that is automatic to get all three together. Not for this PR, but in general for projections.

lsf37: (No action necessary for this, but if someone has an idea, do let me know)

proof/invariant-abstract/DetSchedAux_AI.thy
```diff
@@ -19,31 +19,93 @@ lemmas [wp] =
...
+  "set_cap c p \<lbrace>\<lambda>s. P (prios_of s)\<rbrace>"
+  by (set_object_easy_cases def: set_cap_def)
+
 lemma set_cap_valid_sched_pred[wp]:
   "set_cap cap slot \<lbrace>valid_sched_pred_strong P\<rbrace>"
+  apply (rule hoare_lift_Pf[where f=ntfn_queues_of, rotated], wpsimp)
+  apply (rule hoare_lift_Pf[where f=ep_queues_of, rotated], wpsimp)
+  apply (rule hoare_lift_Pf[where f=prios_of, rotated], wpsimp)
+  apply (rule hoare_lift_Pf[where f=eps_of, rotated], wpsimp)
+  apply (rule hoare_lift_Pf[where f=ntfns_of, rotated], wpsimp)
```
Xaphiosis, line 47: oo, another candidate for that lifting rule

proof/invariant-abstract/DetSchedAux_AI.thy
```diff
@@ -19,31 +19,93 @@ lemmas [wp] =
...
+ "set_mrs thread buf msgs \<lbrace>\<lambda>s. P (prios_of s)\<rbrace>"
+  by (set_object_easy_cases def: set_mrs_def store_word_offs_def final: get_tcb_def)
 
 lemma set_mrs_valid_sched_pred[wp]:
   "set_mrs thread buf msgs \<lbrace>valid_sched_pred_strong P\<rbrace>"
+  apply (rule hoare_lift_Pf[where f=ntfn_queues_of, rotated], wpsimp)
+  apply (rule hoare_lift_Pf[where f=ep_queues_of, rotated], wpsimp)
+  apply (rule hoare_lift_Pf[where f=prios_of, rotated], wpsimp)
+  apply (rule hoare_lift_Pf[where f=eps_of, rotated], wpsimp)
+  apply (rule hoare_lift_Pf[where f=ntfns_of, rotated], wpsimp)
```
Xaphiosis, line 105: yes, lifting rule

proof/invariant-abstract/DetSchedAux_AI.thy
```diff
@@ -622,13 +684,88 @@ lemma released_sc_cur_time_increasing:
...
+lemma in_ntfn_queue_st_tcb_at:
+  "\<lbrakk>ntfn_queues_of s ntfn_ptr = Some q; t \<in> set q; valid_objs s; sym_refs (state_refs_of s)\<rbrakk>
+   \<Longrightarrow> st_tcb_at (\<lambda>st. st = BlockedOnNotification ntfn_ptr) t s"
+  apply (clarsimp simp: opt_map_def split: option.splits)
+  apply (frule st_in_waitingntfn[where q=q])
+     apply (rename_tac notification, case_tac notification; clarsimp)
+     apply (case_tac ntfn_obj; clarsimp)
+    apply fastforce
+   apply fastforce
+  apply fastforce
```
Xaphiosis, line 737: vertical compression: `fastforce+` or auto or something

proof/invariant-abstract/DetSchedAux_AI.thy
```diff
@@ -830,7 +968,96 @@ lemma valid_sched_tcb_state_preservation_gen:
    apply (frule (1) non_empty_sc_replies_nonz_cap)
    apply (rule use_valid, assumption, rule sc_refill_cfg)
    by (intro conjI; assumption)
-  by simp
+  apply (prop_tac "sorted_ipc_queues s'")
+   subgoal for s r s'
```
Xaphiosis, line 972: nice that you isolated this, thanks

proof/invariant-abstract/IpcDet_AI.thy
```diff
@@ -93,28 +93,304 @@ lemma sfi_tcb_at [wp]:
...
+   mapM (gets_the \<circ> f) ts
+   \<lbrace>P\<rbrace>"
+  apply (wpsimp wp: mapM_gets_the_wp)
+  by (simp add: map_equality_iff)
+
+lemma mapM_gets_the_sp:
+  "\<lbrace>P\<rbrace> mapM (gets_the \<circ> f) ts \<lbrace>\<lambda>rv s. P s \<and> map (\<lambda>t. f t s) ts = map Some rv\<rbrace>"
+  by (wpsimp wp: mapM_gets_the_wp simp: comp_def)
+
+lemma mapM_gets_the_sp':
```
Xaphiosis, line 119: yes, these could go into `lib`, suggest `_nth` instead of `'`

proof/invariant-abstract/IpcDet_AI.thy
Xaphiosis: bikeshed: opt_ord_R vs opt_ord_rel? we don't seem to name defs with stuff like `R` as far as I can tell

proof/invariant-abstract/IpcDet_AI.thy
```diff
@@ -93,28 +93,304 @@ lemma sfi_tcb_at [wp]:
...
+  apply (subst sorted_insort_filter_eq_insort_partition[OF tr sorted])
+  by (rule sorted_insort_partition[OF tot tr re sorted])
+
+fun opt_ord_R where
+  "opt_ord_R R (Some x) (Some y) = (R x y)"
+| "opt_ord_R R x y = (y = None \<longrightarrow> x = None)"
+
+definition opt_ord where "opt_ord \<equiv> opt_ord_R (\<le>)"
+
+declare opt_ord_def[simp]
```
Xaphiosis, line 272: this doesn't make sense ... if you have a definition where you don't even bother with a type signature, and then you immediately add that definition to `[simp]`, how is it any better than an abbreviation?

proof/invariant-abstract/IpcDet_AI.thy
```diff
@@ -93,28 +93,304 @@ lemma sfi_tcb_at [wp]:
...
+  "transp R \<Longrightarrow> transp (opt_ord_R R)"
+  apply (clarsimp simp: transp_def)
+  by (case_tac x; case_tac y; case_tac z; clarsimp elim!: order_trans)
+
+lemma reflp_img_ord:
+  "reflp cmp \<Longrightarrow> reflp (img_ord f cmp)"
+  unfolding reflp_def img_ord_def by blast
+
+lemma reflp_opt_ord:
+  "reflp R \<Longrightarrow> reflp (opt_ord_R R :: ('a::preorder) option \<Rightarrow> 'a option \<Rightarrow> bool)"
```
Xaphiosis, line 358: hmm, yeah my brain would prefer `opt_ord_rel R` to `opt_ord_R R`

proof/invariant-abstract/IpcDet_AI.thy
```diff
@@ -123,11 +399,22 @@ lemma tcb_ep_append_valid_ep:
    \<lbrace>\<lambda>q'. valid_ep (update_ep_queue ep q')\<rbrace>"
   by (cases ep) (wpsimp wp: tcb_ep_append_valid_SendEP tcb_ep_append_valid_RecvEP)+
 
+lemma set_map_fst_filter_zip:
+  "set (map fst (filter P (zip xs ys))) \<subseteq> set xs"
```
Xaphiosis, line 403: another lib candidate; if you're ever in the mood for janitorial work, there's a fair few things I've seen in this review we could move to lib

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -1033,6 +1083,11 @@ lemma set_thread_state_valid_sched_pred_strong':
   apply (rule hoare_lift_Pf2[where f=etcbs_of, rotated], wpsimp)
   apply (rule hoare_lift_Pf2[where f=last_machine_time_of, rotated], wpsimp)
   apply (rule hoare_lift_Pf2[where f=time_state_of, rotated], wpsimp)
+  apply (rule hoare_lift_Pf2[where f=ep_queues_of, rotated], wpsimp)
+  apply (rule hoare_lift_Pf2[where f=ntfn_queues_of, rotated], wpsimp)
+  apply (rule hoare_lift_Pf2[where f=prios_of, rotated], wpsimp)
+  apply (rule hoare_lift_Pf2[where f=eps_of, rotated], wpsimp)
+  apply (rule hoare_lift_Pf2[where f=ntfns_of, rotated], wpsimp)
```
Xaphiosis, line 1090: anothere one for team lift

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -26550,6 +27301,7 @@ lemma call_kernel_valid_sched:
         \<and> cur_sc_offset_sufficient (consumed_time s) s\<rbrace>
    call_kernel e
    \<lbrace>\<lambda>_. valid_sched :: det_state \<Rightarrow> _\<rbrace>"
+  supply comp_apply[simp del]
```
Xaphiosis, line 27304: I would like to take this opportunity to express concern at the size of DetSchedSchedule_AI on rt. My browser took longer than my tea does to brew in order to open this file's diff for review.

Xaphiosis:
I like how you structured the proof of this invariant, and the way the asserts are lining up and not polluting Refine works out well. I didn't spot anything style-wise, and also appreciated where you pointed out a bunch of lemmas were moved.

I think the outstanding non-minor issues from my side are:
- the over-use of `hoare_lift_Pf` rules indicates missing lifting rules; `hoare_lift_Pf` rules are generally avoided if possible because, amongst other issues, they nearly always apply and so can loop, and are the antithesis of automation
- you are correct about the `opt_map_Some_comp` situation. Apparently, this might be able to be addressed with making `||>` a definition, and we might want to give this a go before we merge this PR

lib/Lib.thy
michaelmcinerney: This is an old lemma that I moved to `Lib` because it appears in a few places and if I hadn't been moved to `Lib`, I would need to state it twice, due to the import structure not being linear. So I could rename it and then do one separate commit to use the new name in the proofs, and then fix up the name in the main commit

proof/invariant-abstract/IpcDet_AI.thy
michaelmcinerney: I think I'll go with `opt_ord_rel`

proof/invariant-abstract/Tcb_AI.thy
michaelmcinerney: I've removed this `monadic_rewrite` lemma and instead used `monadic_rewrite_refine_valid` and `monadic_rewrite_bind_tail` within the proof where the previous lemma was used.

proof/invariant-abstract/IpcDet_AI.thy
michaelmcinerney: Good point. I've made this an abbreviation

lib/Lib.thy
michaelmcinerney: And this lemma with the name `set_zip_helper` occurs on `master`, too. Considering that we want `Lib` to be as similar as possible between `master` and `rt`, I suppose I'll have to make a PR for `master` moving this lemma and renaming it...

proof/invariant-abstract/IpcDet_AI.thy
michaelmcinerney: Done

michaelmcinerney:
I think I have addressed most of what we would like to change for this PR, with the exception of renaming `tcb_ep_append/tcbEPAppend` to `tcb_ep_insert/tcbEPInsert`. There are also further lemmas that I could move to `Lib`. If we would very much like those tasks done as part of this PR, then I could do that, but I would prefer doing those in separate PRs, if that is OK.

lib/Lib.thy
Xaphiosis: Good idea. It doesn't look big, so it'll go quickly.

proof/invariant-abstract/Tcb_AI.thy
Xaphiosis: Cool, makes sense now

Xaphiosis:
> I think I have addressed most of what we would like to change for this PR, with the exception of renaming `tcb_ep_append/tcbEPAppend` to `tcb_ep_insert/tcbEPInsert`. There are also further lemmas that I could move to `Lib`. If we would very much like those tasks done as part of this PR, then I could do that, but I would prefer doing those in separate PRs, if that is OK.

Fine to do separately. Don't forget to make the issue for the misc/lifting/hoare_liftPf situation if you haven't already. Someone should double-check the squashed commits once they're squashed, but from my side this looks good.

Xaphiosis (approved):

michaelmcinerney:
Here's my issue for the heap projections: https://github.com/seL4/l4v/issues/841

spec/haskell/src/SEL4/Object/TCB.lhs
lsf37: Agreed.

lsf37 (approved):
Good work! I have nothing further to add over the minor queries from before and Raf's thorough review.

lsf37:
And nice job on resolving the `comp_apply` issue, that went even better than I expected.

michaelmcinerney:
I've force pushed to update this PR so that it now consists of the commits from https://github.com/seL4/l4v/pull/845 ("Some changes for `Lib`") cherry-picked onto the `rt` branch, plus squashed commits for the new work for `sorted_ipc_queues`

PR #837, opened 2024-12-11
AArch64 Access Control: initial setup

ryybrr:
Partial access control proofs for AArch64:
- Introduce a minimal set of assumption such that the remaining theories build without sorries. These assumptions are defined at the end of `AARCH64/ArchAccess.thy` and represent new integrity constraints
- Add `FIXME AARCH64` comments for additional proof obligations not captured by the current spec
- Update `ASpec` and `AInvs` to enforce a stricter partitioning of VCPU saves/restores

This PR is structured into 4 commits:
1. Relax various preconditions in generic and arch-specific access control theories
2. Port existing arch-specific theories verbatim over to AArch64
3. Update these theories such that `Access` builds with `quick_and_dirty` enabled
4. Update `run_tests` to enable `Access` for AArch64

To review the fine-grained AArch64 changes, refer only to the 3rd commit

### DIFF
--- a/proof/access-control/AARCH64/ArchADT_AC.thy
+++ b/proof/access-control/AARCH64/ArchADT_AC.thy
@@ -1,0 +1,124 @@
+(*
+ * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+ *
+ * SPDX-License-Identifier: GPL-2.0-only
+ *)
+
+theory ArchADT_AC
+imports ADT_AC
+begin
+
+context Arch begin global_naming RISCV64
+
+named_theorems ADT_AC_assms
+
+lemma mask_ptTranslationBits_ucast_ucast:
+  "(asid && mask ptTranslationBits) = ucast (ucast asid :: 9 word)"
+  by (word_eqI_solve simp: ptTranslationBits_def)
+
+lemma ptr_offset_in_ptr_range:
+  "\<lbrakk> invs s; x \<notin> kernel_mappings;
+     get_vspace_of_thread (kheap s) (arch_state s) tcb \<noteq> global_pt s;
+     get_page_info (aobjs_of s)
+                   (get_vspace_of_thread (kheap s) (arch_state s) tcb) x = Some (base, sz, attr, r) \<rbrakk>
+     \<Longrightarrow> ptrFromPAddr base + (x && mask sz) \<in> ptr_range (ptrFromPAddr base) sz"
+  apply (simp add: ptr_range_def mask_def)
+  apply (rule conjI)
+   apply (rule_tac b="2 ^ sz - 1" in word_plus_mono_right2)
+    apply (frule some_get_page_info_umapsD)
+           apply (fastforce dest: get_vspace_of_thread_reachable
+                            simp: canonical_not_kernel_is_user get_page_info_def)+
+    apply clarsimp
+    apply (drule is_aligned_ptrFromPAddr_n)
+     apply (simp add: pageBitsForSize_def pageBits_def canonical_bit_def ptTranslationBits_def
+               split: vmpage_size.splits)
+    apply (clarsimp simp: is_aligned_no_overflow' word_and_le1)+
+  apply (subst p_assoc_help)
+  apply (rule word_plus_mono_right)
+   apply (rule word_and_le1)
+  apply (frule some_get_page_info_umapsD)
+         apply (fastforce dest: get_vspace_of_thread_reachable
+                          simp: canonical_not_kernel_is_user get_page_info_def)+
+  apply clarsimp
+  apply (drule is_aligned_ptrFromPAddr_n)
+   apply (simp add: pageBitsForSize_def pageBits_def canonical_bit_def ptTranslationBits_def
+             split: vmpage_size.splits)
+  apply (clarsimp simp: is_aligned_no_overflow')
+  done
+
+lemma user_op_access[ADT_AC_assms]:
+  "\<lbrakk> invs s; pas_refined aag s; is_subject aag tcb; ptable_lift tcb s x = Some ptr;
+     auth \<in> vspace_cap_rights_to_auth (ptable_rights tcb s x) \<rbrakk>
+     \<Longrightarrow> abs_has_auth_to aag auth tcb (ptrFromPAddr ptr)"
+  apply (case_tac "x \<in> kernel_mappings")
+  using get_vspace_of_thread_asid_or_global_pt
+   apply (fastforce simp: ptable_rights_def vspace_cap_rights_to_auth_def invs_def
+                          valid_state_def valid_arch_state_def kernel_mappings_canonical
+                    dest: some_get_page_info_kmapsD split: option.splits)
+  apply (clarsimp simp: ptable_lift_def split: option.splits)
+  apply (insert get_vspace_of_thread_asid_or_global_pt)
+  apply (erule_tac x=s in meta_allE)
+  apply (erule_tac x=tcb in meta_allE)
+  apply (cases "get_vspace_of_thread (kheap s) (arch_state s) tcb = global_pt s"; clarsimp)
+   apply (frule get_page_info_gpd_kmaps[rotated 2]; fastforce simp: get_page_info_def)
+  apply (frule (3) ptr_offset_in_ptr_range)
+  apply (frule get_vspace_of_thread_reachable; clarsimp)
+  apply (frule vs_lookup_table_vspace)
+     apply fastforce+
+  apply (clarsimp simp: get_vspace_of_thread_def get_page_info_def ptable_rights_def pt_lookup_slot_def
+                 split: if_splits option.splits kernel_object.splits cap.splits arch_cap.splits)
+  apply (frule (1) canonical_not_kernel_is_user)
+  apply (frule pt_lookup_slot_from_level_is_subject)
+          apply (fastforce elim: vs_lookup_table_vref_independent)+
+   apply (rule aag_Control_into_owns)
+    apply (clarsimp simp: pas_refined_def auth_graph_map_def state_objs_to_policy_def)
+    apply (erule subsetD)
+    apply (drule_tac addr="tcb_cnode_index 1" in caps_of_state_tcb)
+    apply (clarsimp simp: tcb_cnode_map_def)
+    apply (drule sbta_caps)
+      apply (fastforce simp: obj_refs_def)
+     apply (fastforce simp: cap_auth_conferred_def arch_cap_auth_conferred_def)
+    apply (rule_tac x=tcb in exI, fastforce)
+   apply simp
+  apply (clarsimp simp: pt_lookup_slot_from_level_def)
+  apply (drule_tac vref'=x in vs_lookup_table_vref_independent, rule order_refl)
+  apply (drule pt_walk_level)
+  apply (drule (1) vs_lookup_table_extend, rule order_refl)
+  apply (rename_tac level vref)
+  apply (case_tac "level = asid_pool_level", simp add: pt_walk_top)
+  apply (frule vs_lookup_table_is_aligned; clarsimp simp: canonical_not_kernel_is_user)
+  apply (clarsimp simp: pas_refined_def pte_info_def split: pte.splits)
+  apply (erule subsetD)
+  apply (clarsimp simp: auth_graph_map_def state_objs_to_policy_def)
+  apply (intro exI conjI sbta_vref | erule sym | rule refl)+
+  apply (clarsimp simp: state_vrefs_def ptes_of_Some pts_of_Some)
+  apply (intro exI conjI)
+      apply (simp add: canonical_not_kernel_is_user)+
+  apply (clarsimp simp: vs_refs_aux_def)
+  apply (rule conjI; clarsimp)
+   apply (clarsimp simp: graph_of_def pte_ref2_def Bex_def ptes_of_Some pts_of_Some aobjs_of_Some)
+   apply (rule_tac x="table_index (pt_slot_offset max_pt_level vref x)" in exI)
+   apply (fastforce simp: table_index_max_level_slots canonical_not_kernel_is_user
+                          image_iff ptrFromPAddr_def mult_is_add.mult_ac)
+  apply (clarsimp simp: graph_of_def pte_ref2_def ptes_of_Some pts_of_Some aobjs_of_Some)
+  apply (rule_tac x="table_index (pt_slot_offset level vref x)" in exI)
+  apply (fastforce simp: image_iff table_index_offset_pt_bits_left
+                         ptrFromPAddr_def mult_is_add.mult_ac)
+  done
+
+lemma write_in_vspace_cap_rights[ADT_AC_assms]:
+  "AllowWrite \<in> ptable_rights (cur_thread s) s va
+   \<Longrightarrow> Write \<in> vspace_cap_rights_to_auth (ptable_rights (cur_thread s) s va)"
+  by (clarsimp simp: vspace_cap_rights_to_auth_def)
+
+end
+
+
+global_interpretation ADT_AC_1?: ADT_AC_1
+proof goal_cases
+  interpret Arch .
+  case 1 show ?case
+    by (unfold_locales; fact ADT_AC_assms)
+qed
+
+end
--- a/proof/access-control/AARCH64/ArchADT_AC.thy
+++ b/proof/access-control/AARCH64/ArchADT_AC.thy
@@ -8,40 +8,37 @@ theory ArchADT_AC
 imports ADT_AC
 begin
 
-context Arch begin global_naming RISCV64
+context Arch begin global_naming AARCH64
 
 named_theorems ADT_AC_assms
 
-lemma mask_ptTranslationBits_ucast_ucast:
-  "(asid && mask ptTranslationBits) = ucast (ucast asid :: 9 word)"
-  by (word_eqI_solve simp: ptTranslationBits_def)
-
 lemma ptr_offset_in_ptr_range:
-  "\<lbrakk> invs s; x \<notin> kernel_mappings;
-     get_vspace_of_thread (kheap s) (arch_state s) tcb \<noteq> global_pt s;
+  "\<lbrakk> invs s; get_vspace_of_thread (kheap s) (arch_state s) tcb \<noteq> global_pt s;
      get_page_info (aobjs_of s)
-                   (get_vspace_of_thread (kheap s) (arch_state s) tcb) x = Some (base, sz, attr, r) \<rbrakk>
+       (get_vspace_of_thread (kheap s) (arch_state s) tcb) x = Some (base, sz, attr, r) \<rbrakk>
      \<Longrightarrow> ptrFromPAddr base + (x && mask sz) \<in> ptr_range (ptrFromPAddr base) sz"
   apply (simp add: ptr_range_def mask_def)
   apply (rule conjI)
    apply (rule_tac b="2 ^ sz - 1" in word_plus_mono_right2)
     apply (frule some_get_page_info_umapsD)
-           apply (fastforce dest: get_vspace_of_thread_reachable
-                            simp: canonical_not_kernel_is_user get_page_info_def)+
+          apply (fastforce dest: get_vspace_of_thread_reachable
+                           simp:  get_page_info_def)+
     apply clarsimp
     apply (drule is_aligned_ptrFromPAddr_n)
-     apply (simp add: pageBitsForSize_def pageBits_def canonical_bit_def ptTranslationBits_def
+     apply (simp add: pageBitsForSize_def pageBits_def canonical_bit_def
+                      ptTranslationBits_def pptrBaseOffset_alignment_def
                split: vmpage_size.splits)
     apply (clarsimp simp: is_aligned_no_overflow' word_and_le1)+
   apply (subst p_assoc_help)
   apply (rule word_plus_mono_right)
    apply (rule word_and_le1)
   apply (frule some_get_page_info_umapsD)
-         apply (fastforce dest: get_vspace_of_thread_reachable
-                          simp: canonical_not_kernel_is_user get_page_info_def)+
+        apply (fastforce dest: get_vspace_of_thread_reachable
+                         simp: get_page_info_def)+
   apply clarsimp
   apply (drule is_aligned_ptrFromPAddr_n)
-   apply (simp add: pageBitsForSize_def pageBits_def canonical_bit_def ptTranslationBits_def
+   apply (simp add: pageBitsForSize_def pageBits_def canonical_bit_def
+                    ptTranslationBits_def pptrBaseOffset_alignment_def
              split: vmpage_size.splits)
   apply (clarsimp simp: is_aligned_no_overflow')
   done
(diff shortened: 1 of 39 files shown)


proof/invariant-abstract/AARCH64/ArchVCPU_AI.thy
```diff
@@ -116,12 +116,14 @@ crunch vcpu_restore, do_machine_op, vcpu_save_reg, vgic_update, save_virt_timer,
   (wp: valid_cur_vcpu_lift crunch_wps simp: active_cur_vcpu_of_def)
 
 lemma switch_vcpu_valid_cur_vcpu_cur_thread_update[wp]:
-  "\<lbrace>arch_tcb_at (\<lambda>itcb. itcb_vcpu itcb = v) t\<rbrace>
+  "\<lbrace>\<lambda>s. arch_tcb_at (\<lambda>itcb. itcb_vcpu itcb = v) t s\<rbrace>
```
lsf37, line 119: Was this eta expansion needed? The two terms should be equivalent. It's possible that the rule `hoare_lift_Pf` only matches agains the eta-expanded form (even though unification should be modulo eta). In that case, we could try to use the `wps` method instead of the explicit rule application.

proof/access-control/AARCH64/ArchFinalise_AC.thy
```diff
@@ -0,0 +1,424 @@
...
+   apply (rule thread_st_auth_fun_upd)
+   apply (clarsimp simp: obj_at_def asid_pools_of_ko_at get_tcb_def)
+  apply (rule thread_bound_ntfns_fun_upd)
+  apply (clarsimp simp: obj_at_def asid_pools_of_ko_at get_tcb_def)
+  done
+
+lemma set_vcpu_pas_refined:
+  "\<lbrace>\<lambda>s. pas_refined aag s \<and> vcpu_at ptr s\<rbrace>
+   set_vcpu ptr vcpu
+   \<lbrace>\<lambda>_ s. pas_refined aag s\<rbrace>"
```
lsf37, line 104: This is modelled on older lemmas I assume, because we did this a lot in the beginning -- the stronger invariant form `set_vcpu ptr vcpu ⦃pas_refined aag⦄` should also be true, because `set_vcpu` should fail if there is no vcpu, and therefore partial correctness triples shouldn't need the `vcpu_at`. The stronger form is easier to use.

proof/access-control/AARCH64/ArchAccess.thy
```diff
@@ -0,0 +1,363 @@
...
+subsection \<open>Generating a policy from the current ASID distribution\<close>
+
+definition pte_ref2 where
+  "pte_ref2 level pte \<equiv> case pte of
+     PagePTE paddr _ _ rights
+       \<Rightarrow> Some (ptrFromPAddr paddr,
+                pt_bits_left level,
+                vspace_cap_rights_to_auth rights)
+   | PageTablePTE ppn
+       \<Rightarrow> Some (ptrFromPAddr (paddr_from_ppn ppn), 0, {Control})
```
lsf37, line 33: style nitpick: the `=>` should go on the same line as the case expression (e.g. PagePTE).

proof/access-control/AARCH64/ArchAccess.thy
```diff
@@ -0,0 +1,363 @@
...
+lemma arch_troa_vcpu_save_reg:
+    "\<lbrakk> aobjs_of s vptr = Some (VCPU vcpu); ao' = VCPU vcpu';
+       option_map fst (arm_current_vcpu (arch_state s)) = Some vptr;
+       vcpu' = vcpu\<lparr>vcpu_regs := (vcpu_regs vcpu)(reg := vcpuHardwareReg_val reg (machine_state s))\<rparr> \<rbrakk>
+       \<Longrightarrow> arch_integrity_obj_atomic aag subjects l ao ao'"
+  sorry
+
+(* FIXME AARCH64: assert a connection to the current (or soon-to-be-switched-to) thread? *)
+\<comment> \<open>Anyone can update the virtual count offset in the current VCPU\<close>
+lemma arch_troa_vcpu_restore_vtimer:
```
lsf37, line 302: The offset update happens only after we have switched the current VCPU to be active, which implies that it is the VCPU of the current thread. This means we might be able to drop this rule, because the change is happening in the current label, which is allowed to change anything that belongs to it already anyway.

lsf37: Thinking more about it, the `integrity` predicate is going to refer to the current VCPU in the initial state, though, not in the final state. So maybe we do need this clause. Still something to keep an eye on for an opportunity to tighten up the possible state changes more.

lsf37 (approved):
Great work! That goes a good deal beyond just setup, it already contains the bulk of the main proof and will allow us to concentrate on the tricky bits in the next part.

I have only some very minor style nitpicks. Otherwise very nice!

proof/access-control/Tcb_AC.thy
```diff
@@ -177,7 +176,7 @@ lemma (in is_extended') cte_wp_at[wp]: "I (cte_wp_at P a)"
   by (rule lift_inv, simp)
 
 lemma checked_insert_pas_refined:
-  "\<lbrace>pas_refined aag and pspace_aligned and valid_vspace_objs and valid_arch_state and valid_mdb and
+  "\<lbrace>pas_refined aag and valid_mdb and
```
Xaphiosis, line 179: this is very cheerful to see, I'm glad you're spotting these overly-strong preconditions

proof/access-control/RISCV64/ArchFinalise_AC.thy
```diff
@@ -241,7 +240,7 @@ global_interpretation Finalise_AC_1?: Finalise_AC_1
 proof goal_cases
   interpret Arch .
   case 1 show ?case
-    by (unfold_locales; (fact Finalise_AC_assms | wp finalise_cap_replaceable))
+    by (unfold_locales; wpsimp wp: Finalise_AC_assms)
```
Xaphiosis, line 243: This is a strange formulation, it implies that `Finalise_AC_assms` is only `wp` rules. The previous formulation said: "if it isn't a fact then it's a wp rule".

proof/access-control/RISCV64/ArchFinalise_AC.thy
```diff
@@ -229,10 +227,11 @@ lemma arch_finalise_cap_respects[wp]:
...
 
-crunch arch_post_cap_deletion
-  for pspace_aligned[Finalise_AC_assms, wp]: "\<lambda>s :: det_ext state. pspace_aligned s"
-  and valid_vspace_objs[Finalise_AC_assms, wp]: "\<lambda>s :: det_ext state. valid_vspace_objs s"
-  and valid_arch_state[Finalise_AC_assms, wp]: "\<lambda>s :: det_ext state. valid_arch_state s"
+declare prepare_thread_delete_st_tcb_at_halted[Finalise_AC_assms]
+declare finalise_cap_valid_list[Finalise_AC_assms]
+declare arch_finalise_cap_pas_refined[Finalise_AC_assms]
+declare prepare_thread_delete_pas_refined[Finalise_AC_assms]
+declare finalise_cap_replaceable[Finalise_AC_assms]
```
Xaphiosis, line 234: what happened here? previously there was a crunch, and then we picked out the `Finalise_AC_assms`... did the crunch move elsewhere?

proof/access-control/RISCV64/ArchSyscall_AC.thy
```diff
@@ -168,7 +169,7 @@ global_interpretation Syscall_AC_1?: Syscall_AC_1
 proof goal_cases
   interpret Arch .
   case 1 show ?case
-    by (unfold_locales; (fact Syscall_AC_assms | wp init_arch_objects_inv))
+    by (unfold_locales; wpsimp wp: Syscall_AC_assms)
```
Xaphiosis, line 172: same as my other comment, is it only wp rules in the assms? surprised it doesn't work with `fact`

proof/access-control/RISCV64/ArchCNode_AC.thy
```diff
@@ -56,29 +56,28 @@ lemma sata_update2[CNode_AC_assms]:
...
                 split: if_split_asm)
 
+lemma vs_lookup_table_eqI':
+    "\<lbrakk> asid_table s' (asid_high_bits_of asid) = asid_table s (asid_high_bits_of asid);
+       \<forall>pool_ptr. asid_table s' (asid_high_bits_of asid) = Some pool_ptr
+                  \<longrightarrow> bot_level \<le> max_pt_level
+                  \<longrightarrow> vspace_for_pool pool_ptr asid (asid_pools_of s') =
+                      vspace_for_pool pool_ptr asid (asid_pools_of s);
+       bot_level < max_pt_level \<longrightarrow> pts_of s' = pts_of s \<rbrakk>
+   \<Longrightarrow> vs_lookup_table bot_level asid vref s' = vs_lookup_table bot_level asid vref s"
```
Xaphiosis, line 66: indentation:
- `⟹ vs_lookup_table` should be under `⟦`
- `vspace_for_pool` indent should be by 2 instead of 1

proof/access-control/DomainSepInv.thy
```diff
@@ -1060,6 +1000,70 @@ lemma invoke_tcb_domain_sep_inv:
...
+    apply (wp cap_insert_domain_sep_inv' | simp)+
+   apply (rename_tac irq cap cslot_ptr s)
+   apply (case_tac cap, simp_all add: domain_sep_inv_cap_def)[1]
+  apply (wp | clarsimp)+
+  done
+
+(* similarly, the preconditions here tend to contradict one another *)
+lemma invoke_control_domain_sep_inv:
+  "\<lbrace>domain_sep_inv irqs st and irq_control_inv_valid i\<rbrace>
+    invoke_irq_control i
```
Xaphiosis, line 1044: indentation

proof/access-control/Arch_AC.thy
```diff
@@ -196,9 +190,7 @@ lemma store_word_offs_integrity_autarch:
   done
 
 lemma set_mrs_pas_refined[wp]:
-  "\<lbrace>pspace_aligned and valid_vspace_objs and valid_arch_state and pas_refined aag\<rbrace>
-   set_mrs thread buf msgs
-   \<lbrace>\<lambda>_. pas_refined aag\<rbrace>"
+  "set_mrs thread buf msgs \<lbrace>pas_refined aag\<rbrace>"
```
Xaphiosis, line 193: I really appreciate the style overhaul you've included in this PR

proof/access-control/ARM/ArchInterrupt_AC.thy
```diff
@@ -75,7 +74,7 @@ global_interpretation Interrupt_AC_1?: Interrupt_AC_1 "arch_authorised_irq_ctl_i
 proof goal_cases
   interpret Arch .
   case 1 show ?case
-    by (unfold_locales; (fact Interrupt_AC_assms)?)
+    by (unfold_locales; fact Interrupt_AC_assms)
```
Xaphiosis, line 77: you can keep this change, but should know the reason for `(fact Interrupt_AC_assms)?`. The reason that's useful is while developing, when this instantiation errors, you can see which fact you're missing from your assumption; with the non-`?` one, you get "proof command failed".

proof/access-control/ARM/ArchSyscall_AC.thy
```diff
@@ -179,7 +179,7 @@ global_interpretation Syscall_AC_1?: Syscall_AC_1
 proof goal_cases
   interpret Arch .
   case 1 show ?case
-    by (unfold_locales; (fact Syscall_AC_assms)?)
+    by (unfold_locales; wpsimp wp: Syscall_AC_assms)
```
Xaphiosis, line 182: I don't really see why the change to instantiating locale assumptions using `wpsimp` is an improvement. You're very consistent with it, so I'd like to hear your thoughts.

proof/access-control/ARM/ArchDomainSepInv.thy
```diff
@@ -25,6 +25,21 @@ lemma arch_finalise_cap_rv[DomainSepInv_assms]:
...
+  arch_switch_to_thread, arch_switch_to_idle_thread, arch_activate_idle_thread
+  for domain_sep_inv[DomainSepInv_assms, wp]: "domain_sep_inv irqs st"
+
+lemma arch_derive_cap_domain_sep_inv[DomainSepInv_assms, wp]:
+  "\<lbrace>\<top>\<rbrace> arch_derive_cap acap \<lbrace>\<lambda>rv _. domain_sep_inv_cap irqs rv\<rbrace>,-"
+  unfolding arch_derive_cap_def
+  by wpsimp
+
+lemma arch_post_modify_registers_domain_sep_inv[DomainSepInv_assms, wp]:
+  "arch_post_modify_registers cur x31 \<lbrace>domain_sep_inv irqs st\<rbrace>"
```
Xaphiosis, line 40: x31, haha

proof/access-control/AARCH64/ArchInterrupt_AC.thy
```diff
@@ -0,0 +1,111 @@
...
+  by (wpsimp wp: dmo_wp mol_respects simp: maskInterrupt_def plic_complete_claim_def)
+
+crunch arch_check_irq for inv[Interrupt_AC_assms, wp]: P
+
+end
+
+
+context begin interpretation Arch .
+
+requalify_consts arch_authorised_irq_ctl_inv
```
Xaphiosis, line 68: Not for this PR, but there's tech to do this more efficiently, documented in `docs/arch-split.md` (arch_requalify_facts)

proof/access-control/AARCH64/ArchAccess_AC.thy
```diff
@@ -0,0 +1,261 @@
...
+     \<Longrightarrow> (\<forall>x a. integrity_asids aag subjects x a s s'')"
+  by clarsimp metis
+
+lemma integrity_asids_refl[Access_AC_assms, simp]:
+  "integrity_asids aag subjects x a s s"
+  by simp
+
+lemma integrity_asids_update_autarch[Access_AC_assms]:
+  "\<lbrakk> \<forall>x a. integrity_asids aag {pasSubject aag} x a st s; is_subject aag ptr \<rbrakk>
+     \<Longrightarrow> \<forall>x a. integrity_asids aag {pasSubject aag} x a st (s\<lparr>kheap := (kheap s)(ptr \<mapsto> obj)\<rparr>)"
```
Xaphiosis, line 85: I'm guessing all of these lemmas with old-style `==>` indents are only moved from elsewhere?

proof/access-control/AARCH64/ArchAccess.thy
```diff
@@ -0,0 +1,363 @@
...
+   | _ \<Rightarrow> None"
+
+definition vs_refs_aux :: "vm_level \<Rightarrow> arch_kernel_obj \<Rightarrow> (obj_ref \<times> obj_ref \<times> aa_type \<times> auth) set"
+  where
+  "vs_refs_aux level \<equiv> \<lambda>ko. case ko of
+     ASIDPool pool \<Rightarrow> (\<lambda>(r,p). (p, ucast r, AASIDPool, Control)) ` graph_of (option_map ap_vspace o pool)
+   | PageTable pt \<Rightarrow> (case pt of
+       VSRootPT pt \<Rightarrow> \<Union>(r,(p, sz, auth)) \<in> graph_of (pte_ref2 level o pt).
+                      (\<lambda>(p, a). (p, ucast r, APageTable VSRootPT_T, a)) ` (ptr_range p sz \<times> auth)
+     | NormalPT pt \<Rightarrow> \<Union>(r,(p, sz, auth)) \<in> graph_of (pte_ref2 level o pt).
```
Xaphiosis, line 43: the `|` should line up

proof/access-control/AARCH64/ArchAccess.thy
```diff
@@ -0,0 +1,363 @@
...
+context Arch begin
+
+primrec aobj_ref' where
+  "aobj_ref' (ASIDPoolCap ref _) = {ref}"
+| "aobj_ref' ASIDControlCap = {}"
+| "aobj_ref' (FrameCap ref _ sz _ _) = ptr_range ref (pageBitsForSize sz)"
+| "aobj_ref' (PageTableCap ref _ _) = {ref}"
+| "aobj_ref' (VCPUCap ref) = {ref}"
+
+fun acap_asid' :: "arch_cap \<Rightarrow> asid set" where
```
Xaphiosis, line 91: I'm not sure I understand the role of `aobj_ref'` and `acap_asid'`. Are there non-`'` versions? Why are they not what we want in access?

proof/access-control/AARCH64/ArchAccess.thy
```diff
@@ -0,0 +1,363 @@
...
+      \<longrightarrow> (\<forall>pool. pools x = Some pool
+                  \<longrightarrow> pools' x \<noteq> None \<and>
+                      (\<forall>pool'. pools' x = Some pool' \<longrightarrow>
+                               (pool \<noteq> pool' \<longrightarrow> (\<exists>asid. atab (asid_high_bits_of asid) = Some x)))))"
+
+definition integrity_asids ::
+  "'a PAS \<Rightarrow> 'a set \<Rightarrow> obj_ref \<Rightarrow> asid \<Rightarrow> 'y::state_ext state \<Rightarrow> 'z::state_ext state  \<Rightarrow> bool" where
+  "integrity_asids aag subjects x asid s s' \<equiv>
+   integrity_asids_aux aag subjects x asid (asid_table s) (asid_table s')
+                                           (asid_pools_of s) (asid_pools_of s')"
```
Xaphiosis, line 141: this is interestingly the other way round to the `_2` setup we use in AInvs/Refine. There we want the abbreviation to expose the state projections, and the definition to hide the complicated bits and pieces. Here you get no benefit from the definition (need to unfold to benefit from the fact that the state projections don't change), and you don't get a real benefit from the abbreviation, because it doesn't hide the complexity ... thoughts?

proof/access-control/AARCH64/ArchADT_AC.thy
```diff
@@ -8,40 +8,37 @@ theory ArchADT_AC
...
-  by (word_eqI_solve simp: ptTranslationBits_def)
-
 lemma ptr_offset_in_ptr_range:
-  "\<lbrakk> invs s; x \<notin> kernel_mappings;
-     get_vspace_of_thread (kheap s) (arch_state s) tcb \<noteq> global_pt s;
+  "\<lbrakk> invs s; get_vspace_of_thread (kheap s) (arch_state s) tcb \<noteq> global_pt s;
      get_page_info (aobjs_of s)
-                   (get_vspace_of_thread (kheap s) (arch_state s) tcb) x = Some (base, sz, attr, r) \<rbrakk>
+       (get_vspace_of_thread (kheap s) (arch_state s) tcb) x = Some (base, sz, attr, r) \<rbrakk>
      \<Longrightarrow> ptrFromPAddr base + (x && mask sz) \<in> ptr_range (ptrFromPAddr base) sz"
```
Xaphiosis, line 19: I'm a bit surprised this lemma is in Access and not in AInvs, but I guess it's only needed up here.

Xaphiosis (approved):
I think this is very good work. Sometimes it's hard to tell what got moved/copied, and my comments may be related to something that's just being consistent with other arches, in which case ignore those. Mostly I have nitpicks/style and some questions to try bridge gaps in my knowledge.

proof/access-control/RISCV64/ArchFinalise_AC.thy
lsf37: This formulation lets `wp` use the ones that are rules and will apply them as rule otherwise. I don't really mind, but it is true that the old version showed slightly more intention. Then again, it would also use `fact` with `Finalise_AC_assms` rules that are `wp` rules. So it didn't *really* make that distinction.

proof/access-control/AARCH64/ArchAccess_AC.thy
lsf37: They are Tom-style indents, not old style indents. I don't think anybody else has used them :-)

proof/access-control/AARCH64/ArchAccess.thy
lsf37: The ASpec versions don't return sets, which we do want here.

proof/access-control/AARCH64/ArchAccess.thy
lsf37: It might be worth trying out what happens if we do it the "right" way around. It might conflict with old style in the Access session, but it'd be good to see if that is true.

proof/access-control/AARCH64/ArchAccess.thy
lsf37: After more discussion offline, we probably do need this, so I'll mark it as resolved.

proof/access-control/AARCH64/ArchAccess.thy
Xaphiosis: maybe `_set` instead of `'` or a comment then?

proof/access-control/AARCH64/ArchAccess_AC.thy
Xaphiosis: Right, those are not new then, fine to leave as is

proof/access-control/RISCV64/ArchFinalise_AC.thy
Xaphiosis: So what this is doing is making use of rules declared `[wp]` in Arch in order to discharge locale assumptions without having to mark the relevant rules `[whatever_assms]`? I guess that makes sense; in theory if you put all the assms into the wp set, then `wpsimp` would solve all subgoals without further arguments?

Xaphiosis: It definitely was not obvious at first glance though, and even now I seek confirmation. If we want to go down this route, we might want to say something in the arch-split docs.

ryybrr: Agree that this isn't the ideal formulation. The issue is that some arch-specific `wp` rules have weaker preconditions than the locales, so `fact` isn't good enough. I'm still trying to think of a better solution.

proof/access-control/RISCV64/ArchSyscall_AC.thy
ryybrr: See my response in the other thread

proof/access-control/ARM/ArchInterrupt_AC.thy
ryybrr: This is also how I use them, can re-add if we want this artifact to exist outside of development

proof/access-control/AARCH64/ArchAccess.thy
ryybrr: Here I'm aligning `VSRootPT` and `NormalPT`. The `of` should be moved to make this clearer. I also have an alternative definition for `vs_refs_aux` in `ArchArch_AC` that avoids pt splitting. It's currently tagged with a `FIXME AARCH64` since it needs more work to be integrated, but I've proved its equivalence to this definition.

proof/access-control/AARCH64/ArchAccess.thy
ryybrr: This is a quirk of access control. I can address this in a later refactor, but it will touch all architectures

proof/access-control/AARCH64/ArchAccess.thy
ryybrr: I'll take a look, but refactoring would involve exposing the abbreviations. For example, `asid_pools_of s` would become `(λs. kh |> aobj_of |> asid_pool_of)` for some generic `kh` instantiated with `kheap s`

lsf37:
(just noting that the linter failures are Ok to ignore, they are picking up the sorried assumptions mentioned in the PR description)

ryybrr:
Superseded by https://github.com/seL4/l4v/pull/889

PR #840, opened 2024-12-12
Update ARM AInvs and Access for det_ext changes

corlewis:
Following on from https://github.com/seL4/l4v/pull/824, this PR copies the AInvs update across to ARM and then additionally updates the Access Control proofs for the changes to det_ext and the scheduler state.

### DIFF
--- a/proof/access-control/ARM/ArchAccess.thy
+++ b/proof/access-control/ARM/ArchAccess.thy
@@ -88,6 +88,11 @@ lemmas state_vrefs_upd =
   revokable_update.state_vrefs
   machine_state_update.state_vrefs
   more_update.state_vrefs
+  scheduler_action_update.state_vrefs
+  domain_index_update.state_vrefs
+  cur_domain_update.state_vrefs
+  domain_time_update.state_vrefs
+  ready_queues_update.state_vrefs
 
 end
 
@@ -138,6 +143,7 @@ abbreviation integrity_asids_aux ::
      pp_opt = pp_opt' \<or> (\<forall>asid'. asid' \<noteq> 0 \<and> asid_high_bits_of asid' = asid_high_bits_of asid
                                  \<longrightarrow> pasASIDAbs aag asid' \<in> subjects)"
 
+\<comment> \<open>FIXME: could we define integrity_asids to operate on arch_state only?\<close>
 definition integrity_asids ::
   "'a PAS \<Rightarrow> 'a set \<Rightarrow> obj_ref \<Rightarrow> asid \<Rightarrow> 'y::state_ext state \<Rightarrow> 'z:: state_ext state  \<Rightarrow> bool" where
   "integrity_asids aag subjects x a s s' \<equiv>
@@ -159,6 +165,11 @@ lemmas integrity_asids_updates =
   interrupt_update.integrity_asids_update
   cur_thread_update.integrity_asids_update
   machine_state_update.integrity_asids_update
+  scheduler_action_update.integrity_asids_update
+  domain_index_update.integrity_asids_update
+  cur_domain_update.integrity_asids_update
+  domain_time_update.integrity_asids_update
+  ready_queues_update.integrity_asids_update
 
 (* The kheap isn't used in ARM's integrity_asids definition,
    but we need the following lemmas in some generic contexts *)
--- a/proof/access-control/ARM/ArchRetype_AC.thy
+++ b/proof/access-control/ARM/ArchRetype_AC.thy
@@ -17,7 +17,7 @@ lemma invs_mdb_cte':
 context retype_region_proofs begin interpretation Arch .
 
 lemma vs_refs_no_global_pts_default[simp]:
-  "vs_refs_no_global_pts (default_object ty dev us) = {}"
+  "vs_refs_no_global_pts (default_object ty dev us d) = {}"
   by (simp add: default_object_def default_arch_object_def tyunt
                 vs_refs_no_global_pts_def pde_ref2_def pte_ref_def
                 o_def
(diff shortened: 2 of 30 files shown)


corlewis:
Ignore the first commit for reviewing purposes, it contains the limited changes that could be copied directly across from AArch64 AInvs.

proof/access-control/Syscall_AC.thy
```diff
@@ -397,20 +400,13 @@ lemma handle_reply_respects:
...
-  apply (erule notE)
-  apply (erule domains_of_state_aux.cases, simp add: get_etcb_def split: if_split_asm)
-   apply (force intro: domtcbs)+
-  done
+
 
 lemma thread_set_time_slice_pas_refined[wp]:
-  "thread_set_time_slice tptr time \<lbrace>pas_refined aag\<rbrace>"
-  by (simp add: thread_set_time_slice_def | wp)+
+  "\<lbrace>pas_refined aag and pspace_aligned and valid_vspace_objs and valid_arch_state\<rbrace>
```
corlewis, line 406: This precondition (and those of several similar lemmas) could be weakened using the changes from https://github.com/seL4/l4v/pull/837/commits/357ccc987dee33b7b9780145e17f3205d462ceda. I'll do so once that commit is either in master or is able to be cherry-picked into this branch.

lsf37:
Is the plan to rebase this over #844?

Xaphiosis:
> Is the plan to rebase this over #844?

I think it already is. This PR has 3 commits, but only one is new, and the second one is the same as in 844 (currently)

proof/access-control/Syscall_AC.thy
Xaphiosis: does this also apply to `lemma pas_refined`?

proof/access-control/Retype_AC.thy
```diff
@@ -418,8 +419,17 @@ end
...
+lemma tcb_domain_map_wellformed:
+  "\<lbrakk>tcb_domain_map_wellformed aag s; pas_cur_domain aag s; \<forall>x\<in> set (retype_addrs ptr ty n us). is_subject aag x\<rbrakk>
+   \<Longrightarrow> tcb_domain_map_wellformed aag s'"
+  unfolding s'_def ps_def
+  apply (clarsimp simp: tcb_domain_map_wellformed_aux_def)
+  apply (erule domains_of_state_aux.cases)
+  apply (clarsimp simp: etcbs_of'_def split: if_split_asm kernel_object.splits)
+   apply (clarsimp simp: default_object_def default_tcb_def tyunt split: apiobject_type.splits)
+   defer
+  apply (force intro: domtcbs)
```
Xaphiosis, line 431: is this copied from elsewhere? post-`defer` indent is wrong, and probably there's no need to split obj and aobj separately anyway, and on top of that force can probably finish it after the erule, or maybe even do the whole lemma if we're lucky

Xaphiosis (approved):
One minor comment, else good. Nice to see more red than expected.

corlewis:
> > Is the plan to rebase this over #844?
> 
> I think it already is. This PR has 3 commits, but only one is new, and the second one is the same as in 844 (currently)

The exact history is a bit more roundabout but yes, I've rebased this PR over the squashed #844 so that this one is just the Access commit.

proof/access-control/Retype_AC.thy
corlewis: I have to admit that I don't remember but it doesn't look like my style so I assume that I copied it from somewhere.

It's actually even weirder than you thought, the post-`defer` indent is correct, because the defer is done when there is only one subgoal. I've changed it like you suggested and moved everything into the force.

proof/access-control/Syscall_AC.thy
corlewis: > This precondition (and those of several similar lemmas) could be weakened using the changes from [357ccc9](https://github.com/seL4/l4v/commit/357ccc987dee33b7b9780145e17f3205d462ceda). I'll do so once that commit is either in master or is able to be cherry-picked into this branch.

This has now been done, see https://github.com/seL4/l4v/pull/840/commits/e3797a8b3895833d08b108e4fa26f87fd0687d9e for the preconditions that were relaxed.

corlewis: > does this also apply to `lemma pas_refined`?

If I'm looking at the right lemma then no. This was specifically for preconditions where I'd had to add `pspace_aligned and valid_vspace_objs and valid_arch_state`

proof/access-control/ARM/ArchAccess.thy
```diff
@@ -138,6 +143,7 @@ abbreviation integrity_asids_aux ::
      pp_opt = pp_opt' \<or> (\<forall>asid'. asid' \<noteq> 0 \<and> asid_high_bits_of asid' = asid_high_bits_of asid
                                  \<longrightarrow> pasASIDAbs aag asid' \<in> subjects)"
 
+\<comment> \<open>FIXME: could we define integrity_asids to operate on arch_state only?\<close>
```
ryybrr, line 146: Unfortunately not, we use `asid_pools_of` in RISCV64

proof/access-control/ARM/ArchRetype_AC.thy
```diff
@@ -33,12 +33,25 @@ end
 
 context retype_region_proofs' begin interpretation Arch .
 
+\<comment> \<open>FIXME: move to pas_refined_state_objs_to_policy_subset?\<close>
```
ryybrr, line 36: If the proof impact is minimal, that would be ideal. Curious what case you ran into where the subset relation was too strong.

ryybrr (approved):
All looks good to me, hope my commit helped!

corlewis:
> All looks good to me, hope my commit helped!

It definitely did, thank you! I had started doing the same thing before realising that the conditions were coming from another arch, which was way out of scope for me at the time.

proof/access-control/ARM/ArchAccess.thy
corlewis: Ah, that makes sense, I was only looking at ARM. We could probably make this a two layer definition with projections so that we have a nicer internal version within arch files while maintaining the interface, but I'm not sure that that is worthwhile.

proof/access-control/ARM/ArchRetype_AC.thy
corlewis: The subset relation is on `domains_of_state`, which retyping now could change. I needed to use the subjects from `tcb_domain_map_wellformed` to show that in this case it still doesn't affect `pas_refined`.

PR #842, opened 2024-12-18
arch-split Refine up to Invariants_H for X64

Xaphiosis:
You'll have to review this commit-by-commit. I tried to keep them as separate as possible, and some should probably get squashed into commits that are already on the `arch-split` branch (from https://github.com/seL4/l4v/pull/833).

The approach I used was:
1. I put the kernel_data_refs junk first, along with the fixup of the locale graph (I missed two edges)
2. All the tweaks to Invariants_H that are needed to make it work with X64 (see commit message)
3. Update AARCH64 Refine to deal with (2).
4. Use result of (3) to copy ArchInvsDefs/Lemmas_H from AARCH64 to X64, and delete X64's Invariants_H
5. Put the X64 definitions and invariants into ArchInvsDefs/Lemmas_H manually, to generate at least a comprehensible diff
6. Fix up X64 Refine, and then CRefine

Here are some observations I had along the way:
- projectKOs are not in simp on X64 or earlier arches, but we all agree they should be
- `tcb_cte_cases_def tcb_cte_cases_neqs` can often replace using `tcb_cte_cases_def cteSizeBits_def` and is much faster; maybe I should add the neqs to some kind of bundle so tcb_cte_cases_def includes it automatically? what's a good way of doing that?
- `tcbBlockSizeBits_def` is still exported into generic context for now; doesn't seem too dangerous, but someday it might be nice to fix that


lsf37:
> - projectKOs are not in simp on X64 or earlier arches, but we all agree they should be

Agreed. Putting them into the simset does lead to some failures, but they should be few. The main annoyance is a lot of warnings, but they could be removed separately.

> - tcb_cte_cases_def tcb_cte_cases_neqs can often replace using tcb_cte_cases_def cteSizeBits_def and is much faster; maybe I should add the neqs to some kind of bundle so tcb_cte_cases_def includes it automatically? what's a good way of doing that?

Even just `lemmas tcb_cte_cases = tcb_cte_cases_def tcb_cte_cases_neqs` would probably be quite nice already.

- tcbBlockSizeBits_def is still exported into generic context for now; doesn't seem too dangerous, but someday it might be nice to fix that

It might be worth starting an issue where we collect some of these smaller TODO items so that they don't all become FIXMEs.

proof/refine/AARCH64/ArchInvsDefs_H.thy
```diff
@@ -40,6 +40,17 @@ context Arch begin arch_global_naming
 declare lookupPTSlotFromLevel.simps[simp del]
 declare lookupPTFromLevel.simps[simp del]
 
+(* this architecture does not use kernel_mappings / pspace_in_kernel_mappings *)
+definition kernel_mappings :: "machine_word set" where
+  "kernel_mappings \<equiv> UNIV"
+
+definition pspace_in_kernel_mappings' :: "kernel_state \<Rightarrow> bool" where
+ "pspace_in_kernel_mappings' \<equiv> \<top>"
```
lsf37, line 48: Not for now, but maybe to note in a FIXME or issue: we should really rename these two, because pspace and cap references are of course not in the kernel mappings, but in the area mapped by the kernel mappings (maybe `pspace_in_kernel_window'`, which I think we did add in RISCV64 AInvs).

For splitting, I'd leave them, though, the rename can probably be automated by script at some later point.

lsf37 (approved):
Massive amount of work! And I cannot find anything to complain about either :-). Really nice to see an effective > 2k lines removed even at this stage already.

There are FIXMEs, of course, but I we will get to these points later and I had expected much worse from how far apart X64 and AARCH64 in terms of concepts.

Xaphiosis:
> > * projectKOs are not in simp on X64 or earlier arches, but we all agree they should be
> 
> Agreed. Putting them into the simset does lead to some failures, but they should be few. The main annoyance is a lot of warnings, but they could be removed separately.
> 
> > * tcb_cte_cases_def tcb_cte_cases_neqs can often replace using tcb_cte_cases_def cteSizeBits_def and is much faster; maybe I should add the neqs to some kind of bundle so tcb_cte_cases_def includes it automatically? what's a good way of doing that?
> 
> Even just `lemmas tcb_cte_cases = tcb_cte_cases_def tcb_cte_cases_neqs` would probably be quite nice already.
> 
> * tcbBlockSizeBits_def is still exported into generic context for now; doesn't seem too dangerous, but someday it might be nice to fix that
> 
> It might be worth starting an issue where we collect some of these smaller TODO items so that they don't all become FIXMEs.

Let me confirm: you agree that handling some of these would be good, but not as part of this PR, and so I should make an issue? Or should I add some of these to this PR? I think I can try do some of these on the next arch, or if you are keen while I'm away, though I'll have a week in Jan before I leave to maybe get another arch or two through, leaving one for mop-up (RISCV64?) and some of these fixes. Then again if they don't get fixed on next arch, they'll get duplicated. Hmm. Thoughts?

lsf37:
> Let me confirm: you agree that handling some of these would be good, but not as part of this PR, and so I should make an issue? 

yes

> Or should I add some of these to this PR? I think I can try do some of these on the next arch, or if you are keen while I'm away, though I'll have a week in Jan before I leave to maybe get another arch or two through, leaving one for mop-up (RISCV64?) and some of these fixes. Then again if they don't get fixed on next arch, they'll get duplicated. Hmm. Thoughts?

It's ok for them to get duplicated in the arch split and then later cleaned up. The only one we could deploy now already is `lemmas tcb_cte_cases = tcb_cte_cases_def tcb_cte_cases_neqs`, because that might make life easier during the splitting already.

Xaphiosis:
> It's ok for them to get duplicated in the arch split and then later cleaned up. The only one we could deploy now already is `lemmas tcb_cte_cases = tcb_cte_cases_def tcb_cte_cases_neqs`, because that might make life easier during the splitting already.

This has the disadvantage of being difficult to grep for later vs the constant. For the next PR, how about I add a `tcb_cte_cases_def'` with a FIXME to add to simpset? Then grepping for the `def'` version will show places that can trivially be replaced by the def version once we add it to the simpset.

PR #843, opened 2024-12-19
Update explicit FPU specifications for det_ext changes

corlewis:
Now that domains are part of the extensible state, arch_prepare_set_domain does not need to be in the det_ext monad.


lsf37 (approved):
Nice.

Xaphiosis (approved):

PR #844, opened 2024-12-19
Update ARM AInvs for det_ext changes

corlewis:
Following on from https://github.com/seL4/l4v/pull/824, this PR copies the AInvs update across to ARM for the changes to det_ext and the scheduler state.


corlewis:
Ignore the first commit for reviewing purposes, it contains the limited changes that could be copied directly across from AArch64 AInvs.

Xaphiosis (approved):
Looks good. Is this incremental work, since it's ARM only?

corlewis:
> Looks good. Is this incremental work, since it's ARM only?

Yes, very incremental, there's still a fair bit more to do before the det_ext_state branch can be merged into master but there's other work that depends on this so I wanted it reviewed early.

PR #845, opened 2024-12-19
Some changes for `Lib`, December '24 edition

michaelmcinerney:
These are some changes that cropped up during my work for `sorted_ipc_queues` for MCS.

I would recommend reviewing commit-by-commit, so that unrelated changes appearing in the same file can be considered individually.

### DIFF
--- a/lib/Monad_Lists.thy
+++ b/lib/Monad_Lists.thy
@@ -512,6 +512,33 @@ lemma mapM_wp':
   apply simp
   done
 
+lemma mapM_gets_the_wp:
+  "\<lbrace>\<lambda>s. \<forall>vs. map (\<lambda>t. f t s) ts = map Some vs \<longrightarrow> P vs s\<rbrace>
+   mapM (gets_the \<circ> f) ts
+   \<lbrace>P\<rbrace>"
+  unfolding comp_def
+proof (induct ts arbitrary: P)
+  case Nil thus ?case by (wpsimp simp: mapM_Nil)
+next
+  case (Cons x xs) show ?case by (wpsimp simp: mapM_Cons wp: Cons)
+qed
+
+lemma mapM_gets_the_wp_nth:
+  "\<lbrace>\<lambda>s. \<forall>vs. (\<forall>i < length vs. f (ts ! i) s = Some (vs ! i)) \<longrightarrow> P vs s\<rbrace>
+   mapM (gets_the \<circ> f) ts
+   \<lbrace>P\<rbrace>"
+  apply (wpsimp wp: mapM_gets_the_wp)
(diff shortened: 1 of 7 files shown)


lib/Monad_Lists.thy
```diff
@@ -512,6 +512,33 @@ lemma mapM_wp':
   apply simp
   done
 
+lemma mapM_gets_the_wp:
+  "\<lbrace>\<lambda>s. \<forall>vs. map (\<lambda>t. f t s) ts = map Some vs \<longrightarrow> P vs s\<rbrace>
+   mapM (gets_the \<circ> f) ts
+   \<lbrace>P\<rbrace>"
+unfolding comp_def
```
Xaphiosis, line 519: @lsf37 edge case for indentation, where does this one go? it's a new one for me

Xaphiosis (approved):
Cheers!
(although do fix the AARCH64 CRefine, not sure why only it broke, probably because it uses latest tech)

lib/Monad_Lists.thy
lsf37: I was going to say this is correct as is, but it is true that we usually put `unfolding` on the same indent as `apply`. That would probably look weird here. What does the Isabelle auto indenter do?

lsf37 (approved):
Cool. For the indent comment above, I would say in this case, go with whatever Isabelle's indenter wants to do. No point in fighting it there.

lib/Monad_Lists.thy
michaelmcinerney: The auto indenter indents this by 2 spaces. Sigh. I'll cancel the tests that are running and push an update

corlewis: Feel free to ignore if you don't want to keep on fiddling with this, but I imagine that the unfold isn't needed in the `Nil` case, and if that's right you could just add `comp_def` to the `simp` in the `Cons` case and possibly avoid fighting with the indentation at all.

corlewis (approved):

lib/Monad_Lists.thy
michaelmcinerney: It does seem necessary in both cases, and even adding it to the `simp` in `wpsimp` in both cases doesn't work. So I just went with the `unfolding`. If this is suspicious, I can look into it further

corlewis: Huh, that seems unexpected to me but not enough to be worth further investigation.

Xaphiosis: Yeah. If you want to sweep it under the rug, you can put a `unfold ...,` before the `induct` in `proof`, otherwise yeah just do what auto-indenter does it this case, it's a really rare event it seems.

PR #846, opened 2025-01-05
Add GIC v3 support for AArch64 platforms

lsf37:
- make the option `config_ARM_GIC_V3` visible to the proofs (cherry-pick from SGI PR #733)
- adjust specs to configurable `virq` bit positions and widths
- update proofs, including:
  - abstract over the value of `max_armKSGICVCPUNumListRegs` (changes with GIC version)
  - adjust `virq` access proofs
  - abstract over `vcpu_t` type size in C (changes with GIC version, because number of LR registers changes)
 
This PR enables the functional correctness proof on all remaining AArch64 platforms.

Test with: seL4/seL4#1337

### DIFF
--- a/proof/crefine/AARCH64/Retype_C.thy
+++ b/proof/crefine/AARCH64/Retype_C.thy
@@ -5447,7 +5447,7 @@ lemma ptr_retyp_fromzeroVCPU:
   assumes cor: "caps_overlap_reserved' {p ..+ 2 ^ vcpuBits} \<sigma>"
   assumes ptr0: "p \<noteq> 0"
   assumes kdr: "{p ..+ 2 ^ vcpuBits} \<inter> kernel_data_refs = {}"
-  assumes subr: "{p ..+ 752} \<subseteq> {p ..+ 2 ^ vcpuBits}" (is "{_ ..+ ?vcpusz} \<subseteq> _")
+  assumes subr: "{p ..+ size_of TYPE(vcpu_C)} \<subseteq> {p ..+ 2 ^ vcpuBits}" (is "{_ ..+ ?vcpusz} \<subseteq> _")
   assumes act_bytes: "region_actually_is_bytes p (2 ^ vcpuBits) \<sigma>'"
   assumes rep0: "heap_list (hrs_mem (t_hrs_' (globals \<sigma>'))) (2 ^ vcpuBits) p = replicate (2 ^ vcpuBits) 0"
   assumes "\<not> snd (placeNewObject p vcpu0 0 \<sigma>)"
@@ -5463,10 +5463,6 @@ proof -
   let ?s' = "\<sigma>\<lparr>ksPSpace := ?ks\<rparr>"
   let ?htdret = "(hrs_htd_update (ptr_retyp (vcpu_Ptr p)) (t_hrs_' (globals \<sigma>')))"
   let ?zeros = "from_bytes (replicate (size_of TYPE(vcpu_C)) 0) :: vcpu_C"
-
-  (* sanity check for the value of ?vcpusz *)
-  have "size_of TYPE(vcpu_C) = ?vcpusz"
-    by simp
 
   have ptr_al:
     "ptr_aligned (vcpu_Ptr p)" using al
@@ -5496,7 +5492,7 @@ proof -
     (is "?hl = ?rep0")
     using retyp_p' rep0
     apply (simp add: lift_t_if h_val_def)
-    apply (subst take_heap_list_le[where k="?vcpusz" and n="2^vcpuBits", symmetric])
+    apply (subst take_heap_list_le[where k="?vcpusz" and n="2^vcpuBits", simplified, symmetric])
      apply (simp add: vcpuBits_def)+
     done
 
@@ -5550,7 +5546,7 @@ proof -
      apply (case_tac r; clarsimp simp: index_foldr_update)
     apply (rule conjI, clarsimp)
      (* vgic_C array initialisation *)
-     apply (subst index_fold_update; clarsimp)
+     apply (subst index_fold_update; clarsimp simp: max_armKSGICVCPUNumListRegs_val)
     (* vppi array initialisation *)
     apply clarsimp
     apply (case_tac vppi; clarsimp)
@@ -5692,7 +5688,7 @@ proof -
     apply (subgoal_tac "region_is_bytes p ?vcpusz \<sigma>'")
      prefer 2
      apply (fastforce simp: region_actually_is_bytes[OF act_bytes]
-                            region_is_bytes_subset[OF _ subr])
+                            region_is_bytes_subset[OF _ subr, simplified])
     apply (simp add: projectKO_opt_retyp_other' objBitsKO_vcpu cl_vcpu
                      htd_safe[simplified] kernel_data_refs_domain_eq_rotate)
     apply (subst ptr_retyp_gen_one[symmetric])+
@@ -5719,7 +5715,7 @@ lemma placeNewObject_vcpu_fromzero_ccorres:
   apply (rule ccorres_from_vcg_nofail, clarsimp)
   apply (rule conseqPre, vcg)
   apply (clarsimp simp: rf_sr_htd_safe)
-  apply (subgoal_tac "{regionBase..+752} \<subseteq> {regionBase..+2^vcpuBits}")
+  apply (subgoal_tac "{regionBase..+size_of TYPE(vcpu_C)} \<subseteq> {regionBase..+2^vcpuBits}")
    prefer 2
    apply clarsimp
    apply (drule intvlD, clarsimp)
--- a/proof/crefine/AARCH64/Syscall_C.thy
+++ b/proof/crefine/AARCH64/Syscall_C.thy
@@ -1564,7 +1564,10 @@ lemma virq_virq_active_set_virqEOIIRQEN_spec':
   apply (case_tac virq)
   apply clarsimp
   apply (rule array_ext)
-  apply (clarsimp simp: virq_get_tag_def virq_tag_defs mask_def virq_type_def split: if_split)
+  (* unfold config to match up with bitfield gen *)
+  apply (clarsimp simp: virq_get_tag_def virq_tag_defs mask_def virq_type_def
+                        virq_type_shift_def eoiirqen_shift_def Kernel_Config.config_ARM_GIC_V3_def
+                  split: if_split)
   done
 
 lemma virq_virq_invalid_set_virqEOIIRQEN_spec':
@@ -1577,7 +1580,10 @@ lemma virq_virq_invalid_set_virqEOIIRQEN_spec':
   apply (case_tac virq)
   apply clarsimp
   apply (rule array_ext)
-  apply (clarsimp simp: virq_get_tag_def virq_tag_defs mask_def virq_type_def split: if_split)
+  (* unfold config to match up with bitfield gen *)
+  apply (clarsimp simp: virq_get_tag_def virq_tag_defs mask_def virq_type_def
+                        virq_type_shift_def eoiirqen_shift_def Kernel_Config.config_ARM_GIC_V3_def
+                  split: if_split)
   done
 
 lemma virq_virq_pending_set_virqEOIIRQEN_spec':
@@ -1590,16 +1596,10 @@ lemma virq_virq_pending_set_virqEOIIRQEN_spec':
   apply (case_tac virq)
   apply clarsimp
   apply (rule array_ext)
-  apply (clarsimp simp: virq_get_tag_def virq_tag_defs mask_def virq_type_def split: if_split)
-  done
-
-lemma gic_vcpu_num_list_regs_cross_over:
-  "\<lbrakk> of_nat (armKSGICVCPUNumListRegs (ksArchState s)) = gic_vcpu_num_list_regs_' t;
-     valid_arch_state' s \<rbrakk>
-   \<Longrightarrow> gic_vcpu_num_list_regs_' t \<le> 0x3F"
-  apply (drule sym, simp)
-  apply (clarsimp simp: valid_arch_state'_def max_armKSGICVCPUNumListRegs_def)
-  apply (clarsimp simp: word_le_nat_alt unat_of_nat)
+  (* unfold config to match up with bitfield gen *)
+  apply (clarsimp simp: virq_get_tag_def virq_tag_defs mask_def virq_type_def
+                        virq_type_shift_def eoiirqen_shift_def Kernel_Config.config_ARM_GIC_V3_def
+                  split: if_split)
   done
 
 lemma virqSetEOIIRQEN_id:
@@ -1608,13 +1608,15 @@ lemma virqSetEOIIRQEN_id:
      virq_get_tag (virq_C (ARRAY _. idx)) \<noteq> scast virq_virq_invalid \<rbrakk>
     \<Longrightarrow> virqSetEOIIRQEN idx 0 = idx"
   apply (clarsimp simp: AARCH64_A.virqSetEOIIRQEN_def virq_get_tag_def virq_tag_defs mask_def
-                        virq_type_def
+                        virq_type_def eoiirqen_shift_def
                   split: if_split)
-  apply (rule_tac x="idx >> 28" in two_bits_cases; simp)
+  (* unfold config to match up with bitfield gen *)
+  apply (rule_tac x="idx >> virq_type_shift" in two_bits_cases;
+         simp add: virq_type_shift_def Kernel_Config.config_ARM_GIC_V3_def)
   done
 
 lemma vgicUpdateLR_ccorres_armHSCurVCPU:
-  "\<lbrakk> v' = v ; n' = n ; n \<le> 63 \<rbrakk> \<Longrightarrow>
+  "\<lbrakk> v' = v ; n' = n ; n < max_armKSGICVCPUNumListRegs \<rbrakk> \<Longrightarrow>
    ccorres dc xfdc
        (\<lambda>s. (\<exists>active. armHSCurVCPU (ksArchState s) = Some (vcpuptr,active) \<and>
                       vcpu_at' vcpuptr s))
@@ -1653,8 +1655,12 @@ lemma vgicUpdateLR_ccorres_armHSCurVCPU:
                             typ_heap_simps' cpspace_relation_def update_vcpu_map_tos)
       apply (erule (1) cmap_relation_updI
              ; clarsimp simp: cvcpu_relation_regs_def cvgic_relation_def cvcpu_vppi_masked_relation_def
+                        simp del: fun_upd_apply
              ; (rule refl)?)
-      apply (fastforce simp: virq_to_H_def split: if_split)
+      (* needs max_armKSGICVCPUNumListRegs_val so it can prove CARD side condition in
+         simp rule Arrays.index_update2 *)
+      apply (rule conjI, clarsimp simp: max_armKSGICVCPUNumListRegs_val split: if_split)
+      apply (fastforce split: if_split)
      apply (clarsimp simp add: objBits_simps bit_simps)+
   apply (clarsimp dest!: rf_sr_ksArchState_armHSCurVCPU simp: cur_vcpu_relation_def split: option.splits)
   done
@@ -1664,6 +1670,12 @@ definition
   eisr_calc :: "32 word \<Rightarrow> 32 word \<Rightarrow> nat"
 where
   "eisr_calc eisr0 eisr1 \<equiv> if eisr0 \<noteq> 0 then word_ctz eisr0 else word_ctz eisr1 + 32"
+
+(* This lives here in the woods, because we don't want it to be widely used.
+   It is only for meant for matching against the corrsponding numeral in C. *)
+schematic_goal max_armKSGICVCPUNumListRegs_int_val:
+  "int max_armKSGICVCPUNumListRegs = numeral ?n"
+  by (simp add: max_armKSGICVCPUNumListRegs_val)
 
 lemma ccorres_vgicMaintenance:
   notes Collect_const[simp del]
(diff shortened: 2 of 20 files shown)


proof/crefine/AARCH64/Syscall_C.thy
```diff
@@ -1785,20 +1786,32 @@ proof -
...
+  have maxListRegs_sint_ge_0:
+    "\<And>n. n < max_armKSGICVCPUNumListRegs \<Longrightarrow> 0 \<le> sint (word_of_nat n :: int_word)"
+    by (simp add: max_armKSGICVCPUNumListRegs_val int_eq_sint)
+
+  have maxListRegs_sint_less:
+    "\<And>n. n < max_armKSGICVCPUNumListRegs \<Longrightarrow>
+          sint (word_of_nat n :: int_word) < int max_armKSGICVCPUNumListRegs"
+    by (simp add: max_armKSGICVCPUNumListRegs_val int_eq_sint)
+
+  (* needs "simp del: of_nat_add", because of rhs *)
```
corlewis, line 1809: This should maybe say `(* needs "simp del: of_nat_add" when used, because of rhs *)`. I was confused when I first read it and there wasn't a `simp del` immediately visible.

corlewis (approved):
Great work, its very nice to see the proofs work for even more platforms!

spec/abstract/AARCH64/VCPU_A.thy
```diff
@@ -106,20 +106,30 @@ definition
...
-      prioShift = 23;
-      irqPending = 1 << 28;
-      eoiirqen = 1 << 19
-    in ((grp && 1) << groupShift) || ((prio && 0x1F) << prioShift) || (irq && 0x3FF)
+      groupShift = if config_ARM_GIC_V3 then 60 else 30;
+      prioShift = if config_ARM_GIC_V3 then 48 else 23;
+      irqPending = 1 << virq_type_shift;
+      eoiirqen = 1 << eoiirqen_shift;
+      irqMask = mask (if config_ARM_GIC_V3 then 32 else 10);
+      prioMask = mask (if config_ARM_GIC_V3 then 8 else 5)
```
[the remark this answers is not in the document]
lsf37: Yes, that's right

proof/crefine/AARCH64/Syscall_C.thy
```diff
@@ -1668,6 +1674,10 @@ definition
 where
   "eisr_calc eisr0 eisr1 \<equiv> if eisr0 \<noteq> 0 then word_ctz eisr0 else word_ctz eisr1 + 32"
 
+schematic_goal max_armKSGICVCPUNumListRegs_int_val:
+  "int max_armKSGICVCPUNumListRegs = numeral ?n"
+  by (simp add: max_armKSGICVCPUNumListRegs_val)
```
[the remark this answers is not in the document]
lsf37: I was trying to make it this one not too available, because I want to discourage its use. I'd usually have made it a `have` or something like that, but that doesn't work with schematics in the goal.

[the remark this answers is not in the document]
corlewis: Ah, that makes sense. Probably worth a comment then otherwise we'll move it when someone stumbles across it in the future and we've forgotten your intent.

spec/abstract/AARCH64/VCPU_A.thy
```diff
@@ -106,20 +106,30 @@ definition
...
     let
-      groupShift = 30;
-      prioShift = 23;
-      irqPending = 1 << 28;
-      eoiirqen = 1 << 19
-    in ((grp && 1) << groupShift) || ((prio && 0x1F) << prioShift) || (irq && 0x3FF)
+      groupShift = if config_ARM_GIC_V3 then 60 else 30;
+      prioShift = if config_ARM_GIC_V3 then 48 else 23;
+      irqPending = 1 << virq_type_shift;
+      eoiirqen = 1 << eoiirqen_shift;
```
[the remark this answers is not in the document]
lsf37: Since it's only a `1 <<` I think factoring out a definition and having a separate name for it is not really worth it.

proof/refine/AARCH64/InterruptAcc_R.thy
```diff
@@ -166,5 +166,25 @@ lemma preemptionPoint_invs [wp]:
   "\<lbrace>invs'\<rbrace> preemptionPoint \<lbrace>\<lambda>_. invs'\<rbrace>"
   by (wp preemptionPoint_inv | clarsimp)+
 
+lemma virqType_eq[simp]:
+  "virqType = virq_type"
```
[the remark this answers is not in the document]
lsf37: That is a good idea, yes.

lsf37:
(rebased and implemented review feedback)

spec/haskell/src/SEL4/Object/VCPU/AARCH64.hs
```diff
@@ -206,18 +206,26 @@ invokeVCPUWriteReg vcpuPtr reg val = do
...
-          irqPending = bit 28
-          eoiirqen = bit 19
+    ((grp .&. 1) `shiftL` groupShift) .|.((prio .&. prioMask) `shiftL` prioShift) .|.
+    (irq .&. irqMask) .|. irqPending .|. eoiirqen
+    where groupShift = if config_ARM_GIC_V3 then 60 else 30
+          prioShift = if config_ARM_GIC_V3 then 48 else 23
+          irqPending = bit virqTypeShift
+          eoiirqen = bit eoiirqenShift
+          irqMask = mask (if config_ARM_GIC_V3 then 32 else 10)
+          prioMask = mask (if config_ARM_GIC_V3 then 8 else 5)
```
Xaphiosis, line 225: I like how you managed to wrap all the horrors up into these 3 functions

spec/abstract/AARCH64/VCPU_A.thy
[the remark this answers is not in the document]
Xaphiosis: stupid question: do we have any preference for or against `bit n` over `1 << n`? (not suggesting changing really, mostly curious)

proof/refine/AARCH64/InterruptAcc_R.thy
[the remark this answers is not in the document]
Xaphiosis: Yeah, this is my feeling as well, these equivalences are similar to what Kernel_Config_Lemmas does for us, and I don't feel very comfortable having them sitting out in InterruptAcc_R in terms of maintenance and stability.

proof/refine/AARCH64/ArchInvsDefs_H.thy
```diff
@@ -216,7 +216,7 @@ definition valid_asid_table' :: "(asid \<rightharpoonup> machine_word) \<Rightar
 definition "is_vcpu' \<equiv> \<lambda>ko. \<exists>vcpu. ko = (KOArch (KOVCPU vcpu))"
 
 definition max_armKSGICVCPUNumListRegs :: nat where
-  "max_armKSGICVCPUNumListRegs \<equiv> 63"
+  "max_armKSGICVCPUNumListRegs \<equiv> if config_ARM_GIC_V3 then 16 else 64"
```
Xaphiosis, line 219: Interesting, I don't recall exactly how we ended up with 63 before, but now we're upping to a power of two. Can you comment on what's going on?

proof/crefine/AARCH64/Wellformed_C.thy
```diff
@@ -86,8 +86,10 @@ abbreviation
 abbreviation
   vs_Ptr :: "machine_word \<Rightarrow> vs_ptr" where "vs_Ptr == Ptr"
 
+value_type vgic_lr_len = "max_armKSGICVCPUNumListRegs"
+
 abbreviation
-  vgic_lr_C_Ptr :: "addr \<Rightarrow> (virq_C[64]) ptr" where "vgic_lr_C_Ptr \<equiv> Ptr"
+  vgic_lr_C_Ptr :: "addr \<Rightarrow> (virq_C[vgic_lr_len]) ptr" where "vgic_lr_C_Ptr \<equiv> Ptr"
```
Xaphiosis, line 92: yay value_type! so glad we made it, and a fun memory

proof/crefine/AARCH64/Wellformed_C.thy
```diff
@@ -616,6 +618,25 @@ schematic_goal hcrVCPU_val:
   by (simp add: hcrVCPU_def hcrCommon_def hcrTWE_def hcrTWI_def
                 Kernel_Config.config_DISABLE_WFI_WFE_TRAPS_def)
 
+
+text \<open>@{text max_armKSGICVCPUNumListRegs} interface\<close>
+
+(* The definition is in Invariants_H, but the properties are only needed in
+   CRefine. *)
```
Xaphiosis, line 625: Um. Would this interface combine with the comment about the "maybe we want a separate file near the start of Refine" in ArchInterruptAcc_R?
What is your thought on putting this interface there (that doesn't mention any C constants) with the comment that these won't be needed until CRefine?

proof/crefine/AARCH64/VSpace_C.thy
```diff
@@ -2871,8 +2848,7 @@ lemma vgicUpdateLR_ccorres:
   apply (rule ccorres_grab_asm)
   apply (simp add: vgicUpdateLR_def vgicUpdate_def)
   apply vcpuUpdate_ccorres
-  supply from_bool_eq_if[simp] from_bool_eq_if'[simp] from_bool_0[simp]
-  apply (fastforce simp: virq_to_H_def cvcpu_vppi_masked_relation_def split: if_split)
+  apply (auto simp: max_armKSGICVCPUNumListRegs_val split: if_split)
```
Xaphiosis, line 2851: I'm never quite comfortable with these. I know why they're needed, but a part of me wants to yell "abstraction violation" ... not clear how we can go about cleaning this up except via static include functions or enums in C

proof/crefine/AARCH64/VSpace_C.thy
```diff
@@ -2779,32 +2779,32 @@ lemma vcpu_restore_ccorres:
...
                  apply (rule_tac val="of_nat n" in ccorres_abstract_known[where xf'=i_'], ceqv)
-                 apply (rule_tac P="n \<le> 63" in ccorres_gen_asm)
+                 apply (rule_tac P="n < max_armKSGICVCPUNumListRegs" in ccorres_gen_asm)
                  apply (rule ccorres_move_c_guard_vcpu)
                  apply (ctac (no_vcg) add: set_gic_vcpu_ctrl_lr_ccorres)
                 apply (clarsimp simp: virq_to_H_def ko_at_vcpu_at'D upt_Suc)
                 apply (rule conjI)
                  apply (subst scast_eq_ucast; (rule refl)?)
-                 apply (fastforce intro!: not_msb_from_less simp: word_less_nat_alt unat_of_nat)
+                 apply (solves \<open>simp add: max_armKSGICVCPUNumListRegs_msb\<close>)
```
Xaphiosis, line 2796: very good idea with the msb

proof/crefine/AARCH64/Syscall_C.thy
```diff
@@ -2014,13 +2030,14 @@ proof -
...
-      apply (clarsimp simp: invs'_HScurVCPU_vcpu_at' valid_arch_state'_def max_armKSGICVCPUNumListRegs_def dest!: invs_arch_state')
-      using sint_ctz[where x=eisr0, simplified]
-      apply (clarsimp simp: word_sless_alt word_sle_eq sint_int_ctz_ge_0 split: if_split)
-      using sint_ctz[where x=eisr1, simplified]
-      apply (subst signed_arith_sint; clarsimp simp: word_size; simp add: sint_int_ctz_ge_0)
-      using sint_ctz[where x=eisr1, simplified]
-      apply (subst signed_arith_sint; clarsimp simp: word_size; simp add: sint_int_ctz_less_32)
+      apply (clarsimp simp: invs'_HScurVCPU_vcpu_at' valid_arch_state'_def dest!: invs_arch_state')
+      apply (clarsimp simp: eisr_calc_def not_le word_sless_alt word_sle_eq)
+      (* fold max_armKSGICVCPUNumListRegs_int_val to prevent arithmetic with the "+ 32" term *)
```
Xaphiosis, line 2035: here's a thought: is there any benefit to changing `GIC_VCPU_MAX_NUM_LR` in C to use an enum? it's a number small enough to fit into an int, would prevent this situation, but would probably break a casting or two. then you could have a lemma that `Kernel_C.appropriately_renamed_GIC_VCPU_MAX_NUM_LR = max_armKSGICVCPUNumListRegs_int_val` somewhere?

proof/crefine/AARCH64/Retype_C.thy
```diff
@@ -5447,7 +5447,7 @@ lemma ptr_retyp_fromzeroVCPU:
   assumes cor: "caps_overlap_reserved' {p ..+ 2 ^ vcpuBits} \<sigma>"
   assumes ptr0: "p \<noteq> 0"
   assumes kdr: "{p ..+ 2 ^ vcpuBits} \<inter> kernel_data_refs = {}"
-  assumes subr: "{p ..+ 752} \<subseteq> {p ..+ 2 ^ vcpuBits}" (is "{_ ..+ ?vcpusz} \<subseteq> _")
+  assumes subr: "{p ..+ size_of TYPE(vcpu_C)} \<subseteq> {p ..+ 2 ^ vcpuBits}" (is "{_ ..+ ?vcpusz} \<subseteq> _")
```
Xaphiosis, line 5450: my fault! glad you fixed it

Xaphiosis (approved):
Generally good to go, with two ideas identified:
- making an explicit interface file for Refine and moving lemmas there
- making the C "constant" a real constant via an enumeration, see `domainConstants` and `numDomains = CONFIG_NUM_DOMAINS`

proof/crefine/AARCH64/Wellformed_C.thy
lsf37: > Um. Would this interface combine with the comment about the "maybe we want a separate file near the start of Refine" in ArchInterruptAcc_R? 

This here is different, this is just CRefine. Corey's comment was about ASpec/Haskell constants that are equal and where it would make sense to decide on one direction for use in simp rules etc. 

> What is your thought on putting this interface there (that doesn't mention any C constants) with the comment that these won't be needed until CRefine?

I'm not keen -- I want them to be constrained to CRefine so that people don't start using them in other places, esp. the numeral stuff. `max_armKSGICVCPUNumListRegs_word_bits` is the only one that would make sense higher up, and on its own it's a bit strange as an interface section, given that it is not actually needed.

proof/crefine/AARCH64/Syscall_C.thy
lsf37: In general yes, but I don't think it's needed for this one. The number is not freely configurable and it's not used in many places. It can only have exactly one of two values and they occur in controlled circumstances in the C. I.e. the space for testing this is small and we already cover it just by running the proofs for all platforms. If it ever goes out of sync, we'll get failures immediately. That's different to NUM_DOMAINS or others where the user can provide a number and we can't test the entire space.

lsf37:
Rebased proofs are failing, because `max_armKSGICVCPUNumListRegs` is now in the Arch locale and I need to export the code equation for it. Will have a look.

proof/crefine/AARCH64/Syscall_C.thy
lsf37: You're right that it would get rid of the `_val` numeral theorems. I still have the urge to keep those C changes to a minimum, and reserve annoyance for the kernel side for those where we really do need it.

proof/crefine/AARCH64/Wellformed_C.thy
lsf37: and it keeps giving :-)

proof/refine/AARCH64/ArchInvsDefs_H.thy
lsf37: I could swear I had a commit message about it, but it is gone. Must have lost it in a squash.

Which is a shame, because now I don't remember the exact details. I think the old value might actually have been just wrong. The confusion was the `max` and `num` in the same name. The maximum list register index you can use is 63 and 23, but the max number of registers is 64 and 24. But let me try to track down why this didn't come up before, because we do at some point match that number to C and the 63/64 did not change there.

spec/abstract/AARCH64/VCPU_A.thy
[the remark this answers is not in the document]
lsf37: Is `bit n` actually available in ASpec? I thought it was a Haskell_Lib thing, but now that you mention it, it's probably in Word_Lib these days. Generally it does sound like a good idea to prefer `bit n`. I somehow find the explicit shift clearer, but I'm not sure I could really explain why. Maybe because it's a more widely used operator and not very long. I would not be opposed to using `bit n`.

proof/crefine/AARCH64/Wellformed_C.thy
Xaphiosis: Hmm. Mixed feelings, but it does achieve the intent of not being usable before CRefine.

proof/refine/AARCH64/InterruptAcc_R.thy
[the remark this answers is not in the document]
lsf37: I'll add an issue, because this will be a bit of a cleanup task. These kinds of lemmas are currently sitting all over the place in Refine, not just here(the one I added is just one more such occasion). I'm not actually sure how to identify them easily.

proof/crefine/AARCH64/Wellformed_C.thy
lsf37: We do have all the other interfaces for CRefine kernel config in this theory, so I think this is actually the right place.

proof/crefine/AARCH64/Syscall_C.thy
Xaphiosis: It's a strategic question; do we push constants to be constants in C over time, given there is a problem with defining constants in C. I guess you're right in this case, and it's only one of two values, but you understand that the general principle of folding a `#define` that appears in C might fail based on it getting mutated by simp, which is part of cinit, etc. In each specific case we can argue whether it's passable or not (e.g. in this case it is), but then if we have to test them in aggregate, I wonder where that leads. Fine to leave for this one, but do keep a vision of the future in mind.

proof/crefine/AARCH64/Wellformed_C.thy
Xaphiosis: Sure, but the reason the other interfaces ended up here was because they referred to C constants. I guess the concept has spread out a bit. Wellformed_C is definitely the place for them if they go into CRefine, agreed.

proof/refine/AARCH64/ArchInvsDefs_H.thy
lsf37: Yes, you are right, that's mostly it. The reason it worked before is that we side-stepped the issue by not using `max_armKSGICVCPUNumListRegs` where it was compared in C, we just used the number instead. It's this change in StateRelation_C that required me doing something about it:

```diff
-     ∧ (∀i≤63. vgicLR vgic i = virq_to_H ((gicVCpuIface_C.lr_C cvgic).[i]))
-     ∧ (∀i≥64. vgicLR vgic i = 0)"
+     ∧ (∀i<max_armKSGICVCPUNumListRegs. vgicLR vgic i = virq_to_H ((gicVCpuIface_C.lr_C cvgic).[i]))
+     ∧ (∀i≥max_armKSGICVCPUNumListRegs. vgicLR vgic i = 0)"
```

That's where the `<=` changed to `<` now.

lsf37: So it wasn't really wrong before, just a bit misleading.

proof/refine/AARCH64/InterruptAcc_R.thy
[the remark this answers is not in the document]
lsf37: https://github.com/seL4/l4v/issues/858

proof/crefine/AARCH64/Wellformed_C.thy
lsf37: I think I will leave them here. Putting them into some random place in Refine does not feel right, even if the theorem dependencies would allow it.

spec/abstract/AARCH64/VCPU_A.thy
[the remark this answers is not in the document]
lsf37: I tried it out: in the spec it is nicer as expected, but the overall result is that I think we should generally avoid the type class version of `bit` -- it automatically gets expanded into `2^n` form by the simplifier, which painfully messes with word proofs and leads the simplifier to do more arithmetic which usually goes down the wrong track. The `1 << n` form is comparatively stable.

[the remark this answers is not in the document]
lsf37: I guess we could try to take the corresponding rules out of the simpset, but that leads to the general problem that theory merges will bring it back and lead to annoyance.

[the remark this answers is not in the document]
Xaphiosis: ugh, that's terrible for dealing with bits of words then... there's never a good heuristic to be able to get back from 2^n to a shift. Ok, we stick with the shifts and try remember why we did that.
Last gasp: is there any way to make a rule that is more specific to words that overrides the generic simp rule? I'm guessing not, but worth checking.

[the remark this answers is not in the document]
lsf37: > Last gasp: is there any way to make a rule that is more specific to words that overrides the generic simp rule? I'm guessing not, but worth checking.

I don't think so, because we mostly want it do to nothing, i.e. there isn't a rule the simplifier should be applying. I guess we could try to rewrite `bit` to `<<` instead, but then we can just as well use that. The real way to fix it would be to remove the `simp` rule in the Isabelle distribution, which I don't really want to do.

PR #847, opened 2025-01-08
arch-split Refine up to Invariants_H for ARM+ARM_HYP+X64+RISCV64

Xaphiosis:
As seen with X64 (#842), you'll have to review this commit-by-commit. Fortunately there didn't need to be any interface changes this time, so we only have three commits. Copying from AARCH64 (nothing interesting there), updating the definitions/lemmas for Refine, and then fixing up anything else. This seems like a good approach, and keeping the copy commit separate is turning out to be a good idea for large changes like these.
Then I did the same thing again for ARM_HYP, adapting changes from ARM. Ran out of time for RISCV64 though.

I'm not entirely clear whether I should be saying "Up to Invariants_H" or "Up to InvariantUpdates_H". Currently going with the latter.

observations from this pass:
- lemmas tcb_cte_cases = tcb_cte_cases_def tcb_cte_cases_neqs
  OR adding tcb_cte_cases_neqs to simpset (for future)
- valid_arch_state_lift' can't be an interface lemma as it stands now due to
  PDEs on ARM; to do a proper lifting lemma we have to say preservation of all
  arch objects instead of PDEs (or as Gerwin suggested enforce a specific class of vspace objects, which might be a bit painful to retrofit to older arches)
- some of the X64.objBits_simps/ARM.objBits_simps/etc can be replaced with
  gen_objBits_simps because I wasn't consistent/observant enough to notice early
- are things like valid_pde_mappings' crossable? (ARM) -> resolved, don't bother

### DIFF
--- a/proof/refine/ARM_HYP/ArchInvsDefs_H.thy
+++ b/proof/refine/ARM_HYP/ArchInvsDefs_H.thy
@@ -151,6 +151,12 @@ definition valid_arch_cap' :: "arch_capability \<Rightarrow> kernel_state \<Righ
 lemmas valid_arch_cap'_simps[simp] =
   valid_arch_cap'_def[split_simps arch_capability.split, simplified]
 
+definition arch_valid_irq :: "irq \<Rightarrow> bool" where
+  "arch_valid_irq irq \<equiv> irq \<le> maxIRQ"
+
+(* within the Arch locale, we want automatic expansion of the valid irq conditions *)
+lemmas [simp] = arch_valid_irq_def
+
 definition valid_arch_tcb' :: "Structures_H.arch_tcb \<Rightarrow> kernel_state \<Rightarrow> bool" where
   "valid_arch_tcb' \<equiv> \<lambda>t s. \<forall>v. atcbVCPUPtr t = Some v \<longrightarrow> vcpu_at' v s "
 
--- a/proof/refine/AARCH64/ArchInvsDefs_H.thy
+++ b/proof/refine/AARCH64/ArchInvsDefs_H.thy
@@ -157,6 +157,12 @@ lemmas valid_arch_cap'_simps[simp] =
   valid_arch_cap'_def[unfolded wellformed_acap'_def valid_arch_cap_ref'_def,
                       split_simps arch_capability.split, simplified]
 
+definition arch_valid_irq :: "irq \<Rightarrow> bool" where
+  "arch_valid_irq irq \<equiv> irq \<le> maxIRQ"
+
+(* within the Arch locale, we want automatic expansion of the valid irq conditions *)
+lemmas [simp] = arch_valid_irq_def
+
 definition is_device_frame_cap' :: "capability \<Rightarrow> bool" where
   "is_device_frame_cap' cap \<equiv> case cap of ArchObjectCap (FrameCap _ _ _ dev _) \<Rightarrow> dev | _ \<Rightarrow> False"
 
--- a/proof/refine/ARM/ArchInvsDefs_H.thy
+++ b/proof/refine/ARM/ArchInvsDefs_H.thy
@@ -124,6 +124,12 @@ definition valid_arch_cap' :: "arch_capability \<Rightarrow> kernel_state \<Righ
 lemmas valid_arch_cap'_simps[simp] =
   valid_arch_cap'_def[split_simps arch_capability.split, simplified]
 
+definition arch_valid_irq :: "irq \<Rightarrow> bool" where
+  "arch_valid_irq irq \<equiv> irq \<le> maxIRQ"
+
+(* within the Arch locale, we want automatic expansion of the valid irq conditions *)
+lemmas [simp] = arch_valid_irq_def
+
 definition valid_arch_tcb' :: "Structures_H.arch_tcb \<Rightarrow> kernel_state \<Rightarrow> bool" where
   "valid_arch_tcb' \<equiv> \<lambda>t. \<top>"
 
--- a/proof/refine/Invariants_H.thy
+++ b/proof/refine/Invariants_H.thy
@@ -20,6 +20,7 @@ arch_requalify_consts
   canonical_address
   valid_arch_tcb'
   valid_arch_obj'
+  arch_valid_irq
   isArchFrameCap
   acapClass
   global_refs'
@@ -311,7 +312,7 @@ where valid_cap'_def:
   | ThreadCap r \<Rightarrow> tcb_at' r s
   | ReplyCap r m x \<Rightarrow> tcb_at' r s
   | IRQControlCap \<Rightarrow> True
-  | IRQHandlerCap irq \<Rightarrow> irq \<le> maxIRQ
+  | IRQHandlerCap irq \<Rightarrow> arch_valid_irq irq \<comment> \<open>arch-dependent maxIRQ bound and invalidIRQ constraint\<close>
   | Zombie r b n \<Rightarrow> n \<le> zombieCTEs b \<and> zBits b < word_bits
                     \<and> (case b of ZombieTCB \<Rightarrow> tcb_at' r s | ZombieCNode n \<Rightarrow> n \<noteq> 0
                     \<and> (\<forall>addr. real_cte_at' (r + 2^cteSizeBits * (addr && mask n)) s))
--- a/proof/refine/RISCV64/ArchInvsDefs_H.thy
+++ b/proof/refine/RISCV64/ArchInvsDefs_H.thy
@@ -129,6 +129,13 @@ definition arch_cap'_fun_lift :: "(arch_capability \<Rightarrow> 'a) \<Rightarro
 
 lemmas arch_cap'_fun_lift_simps[simp] = arch_cap'_fun_lift_def[split_simps capability.split]
 
+(* on RISCV64, there is an invalid IRQ within the bound of maxIRQ *)
+definition arch_valid_irq :: "irq \<Rightarrow> bool" where
+  "arch_valid_irq irq \<equiv> irq \<le> maxIRQ \<and> irq \<noteq> irqInvalid"
+
+(* within the Arch locale, we want automatic expansion of the valid irq conditions *)
+lemmas [simp] = arch_valid_irq_def
+
 definition is_device_frame_cap' :: "capability \<Rightarrow> bool" where
   "is_device_frame_cap' cap \<equiv> case cap of ArchObjectCap (FrameCap _ _ _ dev _) \<Rightarrow> dev | _ \<Rightarrow> False"
 
--- a/proof/refine/RISCV64/Interrupt_R.thy
+++ b/proof/refine/RISCV64/Interrupt_R.thy
@@ -465,11 +465,8 @@ lemma invoke_irq_handler_invs'[wp]:
   done
 
 lemma IRQHandler_valid':
-  "(s' \<turnstile>' IRQHandlerCap irq) = (irq \<le> maxIRQ)"
-  (* "(s' \<turnstile>' IRQHandlerCap irq) = (irq \<le> maxIRQ \<and> irq \<noteq> irqInvalid)" *)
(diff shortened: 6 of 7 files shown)


proof/crefine/ARM/Retype_C.thy
```diff
@@ -5346,6 +5346,10 @@ lemma pspace_no_overlap_induce_notification:
             split: arch_kernel_object.split_asm)
   done
 
+lemma add_abc_bac:
+  "a + b + c = b + a + (c::'a::semiring)"
+  by (simp add: field_simps)
+
```
lsf37, line 5352: We should probably leave a FIXME here and clean up the word proof that follows below where this is used. Surely we have better tech and/or insight since the original crimes on that were committed.

(But Ok to leave in for arch split purposes, it's not substantially worse than before)

proof/crefine/ARM_HYP/ArchMove_C.thy
```diff
@@ -531,21 +531,23 @@ lemma valid_untyped':
...
+  apply (simp only: add_mask_fold)
   apply (fold obj_range'_def)
   apply (rule iffI)
    apply auto[1]
   apply (rule conjI)
    apply (rule ccontr, simp)
    apply (simp add: Set.psubset_eq)
    apply (erule conjE)
-   apply (case_tac "obj_range' ptr' ko \<inter> {ptr..ptr + 2 ^ bits - 1} \<noteq> {}", simp)
+   apply (case_tac "obj_range' ptr' ko \<inter> mask_range ptr bits \<noteq> {}", simp)
```
lsf37, line 542: Very happy to see all these manual sets being replaced by `mask_range`. Thank you for doing that.

proof/refine/ARM/CSpace_R.thy
```diff
@@ -2805,6 +2805,22 @@ lemma setCTE_arch [wp]:
   apply (wp|wpc|simp del: hoare_fail_any)+
   done
 
+(* FIXME arch-split: cleaner way of doing this? can't use obj_at_setObject2 on CTEs *)
+lemma setCTE_ko_at'_pde[wp]:
+  "setCTE p c \<lbrace> \<lambda>s. P (ko_at' (pde::ARM_H.pde) p' s) \<rbrace>"
```
lsf37, line 2810: I don't see any immediate cleaner way. CTEs on Haskell have always been annoying, this proof still looks pretty Ok.

proof/refine/ARM/KHeap_R.thy
```diff
@@ -1818,16 +1856,44 @@ lemma valid_global_refs_lift':
...
+  assumes pds: "\<And>(pd::ARM_H.pde) p P. f \<lbrace>\<lambda>s. P (ko_at' pd p s) \<rbrace>"
+  shows "\<lbrace>\<lambda>s. valid_pde_mappings' s\<rbrace> f \<lbrace>\<lambda>rv s. valid_pde_mappings' s\<rbrace>"
+  unfolding valid_pde_mappings'_def valid_pde_mapping'_def
+  by (wpsimp wp: hoare_vcg_all_lift hoare_vcg_imp_lift' pds)
+
+(* FIXME arch-split: no interface lemma prospects due to valid_pde_mappings', meaning interface
+   lemmas will be limited to preserving valid_arch_state' in most cases *)
 lemma valid_arch_state_lift':
   assumes typs: "\<And>T p P. \<lbrace>\<lambda>s. P (typ_at' T p s)\<rbrace> f \<lbrace>\<lambda>_ s. P (typ_at' T p s)\<rbrace>"
+  assumes pds: "\<And>(pd::ARM_H.pde) p P. f \<lbrace>\<lambda>s. P (ko_at' pd p s) \<rbrace>"
```
lsf37, line 1869: We should have a look at whether it would make sense to designate a vspace object level on all architectures and if that would be sufficient interface for valid_arch_state. They all do have some object that happens to be at the top, for RISCV64 it would come down to all PTEs, but that might still be useful.

Anyway, for later. Happy with leaving the FIXME for this pass.

proof/refine/ARM_HYP/Interrupt_R.thy
```diff
@@ -787,6 +787,11 @@ lemma virqSetEOIIRQEN_eq[simp]:
   unfolding virqSetEOIIRQEN_def ARM_HYP_A.virqSetEOIIRQEN_def
   by (rule ext) auto
 
+(* FIXME arch-split: RAF MOVE *)
+lemma not_pred_tcb':
+  "(\<not>pred_tcb_at' proj P t s) = (\<not>tcb_at' t s \<or> pred_tcb_at' proj (\<lambda>a. \<not>P a) t s)"
+  by (auto simp: pred_tcb_at'_def obj_at'_def)
```
lsf37, line 793: That one is probably worth moving before we merge.

proof/refine/ARM_HYP/Retype_R.thy
```diff
@@ -4759,6 +4766,18 @@ lemma createNewCaps_vms:
...
+lemma new_range_subset':
+  assumes
+        cover: "range_cover ptr sz (objBitsKO ko) n"
+    and addr: "x \<in> set (new_cap_addrs n ptr ko)"
+  shows       "mask_range x (objBitsKO ko) \<subseteq> {ptr .. (ptr && ~~ mask sz) + 2 ^ sz - 1}"
+  using assms
+  apply -
+  apply (drule (1) new_range_subset)
+  apply (simp add: mask_def add_diff_eq)
+  done
```
lsf37, line 4779: Is this moved from somewhere? The indentation is a bit strange and the `assumes`/`shows` is unnecessary, because not used in the proof. Removing that would also get rid of the `apply -`.

proof/refine/ARM/InvariantUpdates_H.thy
```diff
@@ -90,15 +201,43 @@ lemma inQ_context[simp]:
...
+
+lemma valid_arch_state'_interrupt[simp]:
+  "valid_arch_state' (ksInterruptState_update f s) = valid_arch_state' s"
+  by (simp add: valid_arch_state'_def)
+
+end
+
+(* FIXME arch-split: locales? *)
+arch_requalify_facts
+  valid_arch_state'_interrupt (* FIXME arch-split: safe to export? any arch where this wouldn't be true *)
```
lsf37, line 233: If it's true between AARCH64/ARM_HYP and ARM then it should be true everywhere. The only difference in interrupt state handling would be in the hyp code.

proof/refine/ARM/ArchInvsLemmas_H.thy
```diff
@@ -481,40 +420,55 @@ lemma objBitsT_koTypeOf:
...
+   = valid_pde_mapping_offset' offset"
+  "valid_pde_mapping' offset (PageTablePDE ptr x z'')
+   = valid_pde_mapping_offset' offset"
+  by (clarsimp simp: valid_pde_mapping'_def)+
+
+(* FIXME arch-split: possibly can go into Invariants_H *)
+lemma valid_pspaceI'[intro]:
+  "\<lbrakk>valid_objs' s; pspace_aligned' s; pspace_distinct' s; valid_mdb' s; no_0_obj' s\<rbrakk>
+   \<Longrightarrow> valid_pspace' s"
+  unfolding valid_pspace'_def by simp
```
lsf37, line 470: I think that might work, yes. Would at least be nice. Ok to leave for the next pass, though.

proof/refine/ARM_HYP/ArchInvsDefs_H.thy
```diff
@@ -63,33 +65,35 @@ definition tcb_vcpu_refs' :: "machine_word option \<Rightarrow> (obj_ref \<times
...
 definition vcpu_tcb_refs' :: "obj_ref option \<Rightarrow> (obj_ref \<times> reftype) set" where
   "vcpu_tcb_refs' t \<equiv> set_option t \<times> {HypTCBRef}"
 
 definition refs_of_a' :: "arch_kernel_object \<Rightarrow> (obj_ref \<times> reftype) set" where
   "refs_of_a' x \<equiv> case x of
-     AARCH64_H.KOASIDPool asidpool \<Rightarrow> {}
-   | AARCH64_H.KOPTE pte \<Rightarrow> {}
-   | AARCH64_H.KOVCPU vcpu \<Rightarrow> vcpu_tcb_refs' (AARCH64_H.vcpuTCBPtr vcpu)"
+     ARM_HYP_H.KOASIDPool asidpool \<Rightarrow> {}
+   | ARM_HYP_H.KOPDE pte \<Rightarrow> {}
```
lsf37, line 76: Should be `ARM_HYP_H.KOPDE pde`

lsf37:
> I'm not entirely clear whether I should be saying "Up to Invariants_H" or "Up to InvariantUpdates_H". Currently going with the latter.

No problem with that.
 
> * lemmas tcb_cte_cases = tcb_cte_cases_def tcb_cte_cases_neqs
>   OR adding tcb_cte_cases_neqs to simpset (for future)
> * valid_arch_state_lift' can't be an interface lemma as it stands now due to
>   PDEs on ARM; to do a proper lifting lemma we have to say preservation of all
>   arch objects instead of PDEs

See code comment above -- maybe preservation of vspace objects could be a thing if we introduce them, but it would be something for later.

> * some of the X64.objBits_simps/ARM.objBits_simps/etc can be replaced with
>   gen_objBits_simps because I wasn't consistent/observant enough to notice early

Would be good to record in an issue, but no need to fix now.

> * are things like valid_pde_mappings' crossable? (ARM)

Theoretically yes, but it looks messy. The formulations on abstract and Haskell level are different, and both are a bit strangely written. Probably to make some proof easier somewhere deep down in the guts of it.

The differences are that the abstract level has no duplicated PTEs and has one object per table with multiple entries, whereas the Haskell level has the hardware required duplications and has single objects. And the predicate is formulated via address masking. Crossing that probably requires a whole bunch of new predicates to appear in corres lemma preconditions. Unlikely to be worth it.

There might be a more straightforward formulation that is also easy to cross (we really only want that there are a bunch of specific PTEs set to InvalidPTE), but it would be a little project to rip it all out and put it back together again.

lsf37 (approved):
Looking good, no real issues from my side, only the small inline remarks above.

proof/refine/ARM_HYP/Retype_R.thy
Xaphiosis: This is a quick copy-pasta of `new_range_subset` with the definition changed to use mask, to smooth over old phrasings and new. I'll tweak it.

proof/crefine/ARM/Retype_C.thy
Xaphiosis: The proof uses extremely specific rewrites, where any simplification or mis-subst stops it from working, hence the extremely specific rewrite to fix it up. It would be interesting to see how much it can be cleaned up. I'll add a normal FIXME for this one then, as it's not very arch-split-specific.

Xaphiosis: The other "solution" is to dump this very specific lemma into Lib, in case anyone ever needs that reordering again.

proof/refine/ARM/InvariantUpdates_H.thy
Xaphiosis: Cheers. Will remove the safety question comments

proof/refine/ARM/ArchInvsLemmas_H.thy
Xaphiosis: Will leave for next pass because the `[intro]` might trigger needing fixups on other arches.

proof/refine/ARM_HYP/ArchInvsDefs_H.thy
Xaphiosis: Good idea with `_`, we should use that when we don't care what the value is.

proof/refine/ARM_HYP/Interrupt_R.thy
Xaphiosis: Will do.

PR #848, opened 2025-01-29
Improve reasoning for linked lists

michaelmcinerney:
Please see the commit message

### DIFF
--- a/proof/refine/RISCV64/Ipc_R.thy
+++ b/proof/refine/RISCV64/Ipc_R.thy
@@ -3597,9 +3597,10 @@ lemma tcbReleaseEnqueue_corres:
     apply (corres corres: setReprogramTimer_corres)
 
    \<comment> \<open>prepend t\<close>
+   apply (simp add: bind_assoc)
    apply (rule corres_from_valid_det)
       apply (fastforce intro: det_wp_modify det_wp_pre)
-     apply (wpsimp simp: tcbQueuePrepend_def wp: hoare_vcg_if_lift2)
+     apply (wpsimp simp: tcbQueuePrepend_def wp: hoare_vcg_imp_lift' hoare_vcg_if_lift2)
     apply (clarsimp simp: ex_abs_def)
     apply (frule state_relation_release_queue_relation)
     apply (clarsimp simp: release_queue_relation_def list_queue_relation_def)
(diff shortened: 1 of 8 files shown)


michaelmcinerney:
There's a typo in the commit message: it should say "the various functions which manipulate the linked list pointers"

proof/refine/RISCV64/Ipc_R.thy
```diff
@@ -3677,35 +3666,23 @@ lemma tcbReleaseEnqueue_corres:
...
-   apply (clarsimp simp: opt_map_def opt_pred_def split: if_splits option.splits)
+   apply (clarsimp simp: release_queue_relation_def)
+   apply (intro hoare_vcg_conj_lift_pre_fix)
+    apply ((wpsimp wp: tcbQueueAppend_list_queue_relation threadSet_sched_pointers | wps)+)[1]
+    apply (frule (1) valid_sched_pointersD[where t=t];
+           clarsimp simp: not_in_release_q_2_def in_opt_pred opt_map_red obj_at'_def)
+   apply (rule hoare_allI, rename_tac t')
+   apply (case_tac "t' = t"; clarsimp)
+    apply (wpsimp wp: threadSet_wp hoare_vcg_all_lift)
+   apply (wpsimp wp: threadSet_opt_pred_other)
```
lsf37, line 3685: This looks like a dangerous thing to do in general -- is the precondition in the `wp` goal schematic here? If it is fixed, it is fine, but if it is schematic then doing `case_tac` means the second branch needs to work with the precondition the first branch instantiates. In that case we should leave a comment here that we are expecting that to work (that it's not an accident).

spec/haskell/src/SEL4/Kernel/Thread.lhs
```diff
@@ -854,5 +854,3 @@ Kernel init will created a initial thread whose tcbPriority is max priority.
 >         new_queue <- tcbQueueRemove releaseQueue tcbPtr
 >         setReleaseQueue new_queue
 >         threadSet (\t -> t { tcbInReleaseQueue = False }) tcbPtr
-
-%
```
lsf37, line 858: Looks like accidentally committed, should be removed from this commit.

proof/refine/RISCV64/TcbAcc_R.thy
```diff
@@ -2364,15 +2364,387 @@ lemma in_correct_ready_qD:
...
+  "\<lbrace>\<lambda>s. heap_ls (tcbSchedNexts_of s) st ls \<and> t \<notin> set ls\<rbrace>
+   threadSet F t
+   \<lbrace>\<lambda>_ s. heap_ls (tcbSchedNexts_of s) st ls\<rbrace>"
+  apply (wpsimp wp: threadSet_wp)
+  by (fastforce dest: heap_path_heap_upd_not_in)
+
+lemma heap_path_not_Nil:
+  "heap_ls hp (Some st) ls \<Longrightarrow> ls \<noteq> []"
+  by fastforce
+
```
lsf37, line 2377: This one should go into the `heap_ls` theory.

lsf37 (approved):
Nice improvement! And I like the `find_goal` uses. It's more to type, but should be less brittle and no longer dependent on goal ordering.

proof/refine/RISCV64/Ipc_R.thy
michaelmcinerney: The precondition is fixed here. This is one of those “lopsided” `corres` rules where we have one small thing happening in the abstract (adding or removing an item from a ready or release queue) and a bunch of stuff happening in the Haskell (performing a linked list operation, setting a head/end queue, and then setting a flag). So these rules go via `corres_from_valid`, which I found easiest to perform with fixed guards and preconditions.

proof/refine/RISCV64/TcbAcc_R.thy
michaelmcinerney: I’ll put up a small PR for this rule for `master` and then once that’s merged, I can cherry pick it onto this branch

proof/refine/RISCV64/Ipc_R.thy
lsf37: 👍 Ok, in that case it's all fine.

spec/haskell/src/SEL4/Kernel/Thread.lhs
michaelmcinerney: I tried to revert this change, but I'm not sure it worked. I always get confused about a blank line at the end and what the red sign thing means

michaelmcinerney: OK I think I've got it now

PR #849, opened 2025-01-31
remove valid_arch_mdb_ctes

lsf37:
Cross over the contents of `valid_arch_mdb_ctes` from the abstract side after proving the missing bits as invariants.

Introduces the assertion `archMDBAssertions` in Haskell, which is defined generically as `arch_mdb_assert (ctes_of s)`. `arch_mdb_assert` in turn is architecture dependent and `True` anywhere that is not X64.

This took quite a bit longer than I had anticipated, and doesn't actually remove that many lines of proof, but it hopefully removes many annoying lines that we would have had to write in the future for arch split to continue. So overall still worth it, I think.

This is on top of PR #847 -- when we start RISCV64 that should probably go on top of this work to avoid duplication.

### DIFF
--- a/proof/invariant-abstract/X64/ArchCSpace_AI.thy
+++ b/proof/invariant-abstract/X64/ArchCSpace_AI.thy
@@ -419,17 +419,32 @@ lemma cap_insert_derived_ioports:
   apply (drule_tac cap=cap in valid_ioports_issuedD, simp+)
   done
 
+lemma is_derived_not_ioport_control:
+  "is_derived (cdt s) src cap cap' \<Longrightarrow> cap \<noteq> ArchObjectCap IOPortControlCap"
+  by (clarsimp simp add: is_derived_def is_derived_arch_def split: if_splits)
+
+lemma cap_insert_derived_ioport_control:
+  "\<lbrace>ioport_control_unique and (\<lambda>s. cte_wp_at (is_derived (cdt s) src cap) src s)\<rbrace>
+     cap_insert cap src dest
+   \<lbrace>\<lambda>rv. ioport_control_unique\<rbrace>"
+  unfolding cap_insert_def set_untyped_cap_as_full_def
+  apply (wpsimp wp: get_cap_wp)
+  apply (auto dest!: is_derived_not_ioport_control
+              simp: cte_wp_at_caps_of_state ioport_control_unique_def is_cap_simps)
+  done
+
 lemma cap_insert_derived_valid_arch_state[CSpace_AI_assms]:
   "\<lbrace>valid_arch_state and (\<lambda>s. cte_wp_at (is_derived (cdt s) src cap) src s)\<rbrace>
    cap_insert cap src dest
    \<lbrace>\<lambda>rv. valid_arch_state \<rbrace>"
-  by (wp valid_arch_state_lift_ioports_aobj_at cap_insert_aobj_at cap_insert_derived_ioports)+
+  by (wp valid_arch_state_lift_ioports_aobj_at cap_insert_aobj_at cap_insert_derived_ioports
+         cap_insert_derived_ioport_control)+
      (simp add: cap_insert_aobj_at valid_arch_state_def)
 
 lemma cap_insert_simple_ioports:
   "\<lbrace>valid_ioports and (\<lambda>s. cte_wp_at (\<lambda>cap'. safe_ioport_insert cap cap' s) dest s) and
         K (is_simple_cap cap \<and> \<not>is_ap_cap cap)\<rbrace>
-     cap_insert cap src dest
+   cap_insert cap src dest
    \<lbrace>\<lambda>rv. valid_ioports\<rbrace>"
   apply (simp add: cap_insert_def)
   apply (wp get_cap_wp set_cap_ioports_safe set_untyped_cap_as_full_ioports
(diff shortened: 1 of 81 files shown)


lsf37:
> This is on top of PR #847 -- when we start RISCV64 that should probably go on top of this work to avoid duplication.

Actually, it is better to do it the other way around. I haven't applied the changes to RISCV64 here, because that one doesn't work yet, but they are probably much easier to apply after RISCV64 is split than trying to get the split to work on things that are broken because of the changes here. So RISCV64 split without this PR first, then I can adjust this PR, and then everything should be fine.

proof/invariant-abstract/X64/ArchUntyped_AI.thy
```diff
@@ -386,6 +386,17 @@ lemma create_cap_ioports[wp]:
   by (wpsimp wp: set_cap_ioports_safe set_cdt_cte_wp_at
               simp: safe_ioport_insert_not_ioport[OF default_cap_not_ioport] create_cap_def)
 
+lemma default_cap_neq_IOPortControlCap[simp]:
+  "default_cap tp oref sz dev \<noteq> ArchObjectCap IOPortControlCap"
+  apply (cases tp; simp)
+  apply (simp add: arch_default_cap_def split: aobject_type.split)
+  done
```
Xaphiosis, line 393: can `by`

proof/refine/X64/Interrupt_R.thy
```diff
@@ -500,6 +500,13 @@ crunch X64_H.updateIRQState
   and ex_cte_cap_wp_to'[wp]: "ex_cte_cap_wp_to' a b"
   (simp: ex_cte_cap_wp_to'_def valid_mdb'_def)
 
+crunch set_irq_state
+  for valid_arch_state[wp]: valid_arch_state
+
+lemma maxUserIRQ_le_maxIRQ:
+  "X64.maxUserIRQ \<le> maxIRQ"
+  by (simp add: X64.maxUserIRQ_def maxIRQ_def)
```
Xaphiosis, line 508: worth moving this somewhere more interface-related in CRefine?

proof/refine/X64/CSpace_R.thy
```diff
@@ -4876,6 +4786,15 @@ lemma maskedAsFull_revokable_safe_parent:
 
 context begin interpretation Arch . (*FIXME: arch-split*)
 
+(* FIXME arch-split: generic statement, arch specific proof *)
```
Xaphiosis, line 4789: ok to leave here, but I should ask whether there's a better place...

proof/invariant-abstract/X64/ArchInvariants_AI.thy
```diff
@@ -1051,6 +1051,17 @@ definition
 where
   "valid_x64_irq_state irqState \<equiv> \<forall>irq > maxIRQ. irqState irq = IRQFree"
 
+definition ioport_control_unique_2 :: "(cslot_ptr \<Rightarrow> cap option) \<Rightarrow> bool" where
```
Xaphiosis, line 1054: might be good to have a comment here about what this means and its intent/use for crossing to Refine

proof/invariant-abstract/X64/ArchCSpace_AI.thy
```diff
@@ -437,12 +452,21 @@ lemma cap_insert_simple_ioports:
          | wpc | simp split del: if_splits)+
   done
 
+lemma cap_insert_simple_ioport_control:
+  "\<lbrace>ioport_control_unique and K (is_simple_cap cap)\<rbrace>
+     cap_insert cap src dest
```
Xaphiosis, line 457: are these copy-pasted somehow? we don't indent the function relative to Hoare triple like that anymore?

proof/invariant-abstract/X64/ArchInvariants_AI.thy
Xaphiosis: Also, for my own curiosity since I wasn't a part of this: how did you come up with this as the necessary invariant?

Xaphiosis:
Looks good. Thank you for doing this!
The upstream conflicts are my fault due to the arch-split merge situation. Apologies.

These parts from non-x64 InvariantUpdates_H should be able to go after your changes:
```
(* FIXME arch-split: valid_arch_mdb_ctes only exists to contain ioport_control on x64, and it is not
   yet clear what the best way to arch-split it is, or whether it can be crossed from AInvs.
   Therefore, for now, export the truth that it doesn't do anything on this arch beyond this point *)
arch_requalify_facts valid_arch_mdb_ctes_def
lemmas [simp] = valid_arch_mdb_ctes_def
```

and from Invariants_H, the requalify of `valid_arch_mdb_ctes` as well as dropping it from the definition of `valid_mdb_ctes` if that already hasn't happened.

Also, upstream got RISCV64 as well now.

proof/refine/X64/Interrupt_R.thy
lsf37: maxUserIRQ is an X64 concept, I don't think the other architectures have that

proof/refine/X64/CSpace_R.thy
lsf37: I was going to leave that question for you :-). There probably is eventually, but for now I think leaving it in this theory is good.

proof/invariant-abstract/X64/ArchCSpace_AI.thy
lsf37: yes, they are based on copy/paste, will fix up the indent

proof/invariant-abstract/X64/ArchInvariants_AI.thy
lsf37: it's the same invariant as what used to be in Refine minus the bits that are already proved elsewhere in AInvs

lsf37:
The conflicts disappear after rebase, so no problem at all. I'll have a look at fixing up the review feedback and RISCV.

proof/refine/X64/CSpace_R.thy
lsf37: (Will leave this in for later when we get there in the arch split process)

proof/invariant-abstract/X64/ArchInvariants_AI.thy
lsf37: (comment added)

Xaphiosis (approved):
Looks good!

PR #850, opened 2025-02-10
Update ARM and AARCH64 AInvs for explicit FPU changes

corlewis:
This updates the AInvs proofs for ARM and AARCH64 following the specification changes made in https://github.com/seL4/l4v/pull/819 that made FPU state more explicit and changed the kernel to use semi-lazy FPU switching. It additionally proves a new invariant about the FPU that we expect to be necessary for later stages of the functional correctness proofs and also updates Access for ARM.

PR #852, opened 2025-02-14
Refactor arch_thread_get/set lemmas

corlewis:
Move earlier and make generic where possible.

### DIFF
--- a/proof/invariant-abstract/AARCH64/ArchArch_AI.thy
+++ b/proof/invariant-abstract/AARCH64/ArchArch_AI.thy
@@ -942,13 +942,9 @@ lemma associate_vcpu_tcb_sym_refs_hyp[wp]:
                              sym_refs (state_hyp_refs_of s)"
                           in hoare_post_imp)
     apply (clarsimp simp: obj_at_def)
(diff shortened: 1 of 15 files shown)


lsf37 (approved):
Looks good from my side, nice to have them all in one place.

Xaphiosis:
Nice cleanup. Out of curiosity, what triggered this?
Only minor comment I have is that there's a bunch of `apply/apply/done` that can be made into `by`, but since they're being moved, entirely optional.

Xaphiosis (approved):

corlewis:
> Nice cleanup. Out of curiosity, what triggered this?

In https://github.com/seL4/l4v/pull/850 I needed a bunch of the `arch_thread_set` lemmas in an earlier file than they were previously in, and once I was moving them I thought that I might as well try and do things properly.

> Only minor comment I have is that there's a bunch of `apply/apply/done` that can be made into `by`, but since they're being moved, entirely optional.

I've done some of that sort of cleanup where I saw the opportunity and was already modifying a lemma to make it generic, but yeah, at this stage I'd be keen to get this merged so I can go back to the branches that use it.

PR #853, opened 2025-02-16
arch-split Refine up to InvariantUpdates_H

Xaphiosis:
This is the combinations of all arches now, intended for merging to master.
It's been mostly reviewed, but is missing RISCV64 review.
Would also appreciate a read-through of the final commits and commit messages.

This includes:
* https://github.com/seL4/l4v/pull/833
* https://github.com/seL4/l4v/pull/842
* https://github.com/seL4/l4v/pull/847 (unfortunately clobbered in a novel way during rebase repair, but comments are there, apologies for the mess)

Final stats for this part are `7737 insertions(+), 17216 deletions(-)`, so 9479 lines of proof removed.


proof/refine/Invariants_H.thy
```diff
@@ -374,12 +312,18 @@ where valid_cap'_def:
   | ThreadCap r \<Rightarrow> tcb_at' r s
   | ReplyCap r m x \<Rightarrow> tcb_at' r s
   | IRQControlCap \<Rightarrow> True
-  | IRQHandlerCap irq \<Rightarrow> irq \<le> maxIRQ \<and> irq \<noteq> irqInvalid
+  | IRQHandlerCap irq \<Rightarrow> arch_valid_irq irq \<comment> \<open>arch-dependent maxIRQ bound and invalidIRQ constraint\<close>
```
lsf37, line 315: This looks like the `maxIRQ` condition is now in arch_valid_irq as well -- that one is generic, we'll want to use it for generic invariant proofs where the value of `maxIRQ` doesn't matter (the value is arch specific, the concept is generic). The `irqInvalid` reasoning is only needed once, but `maxIRQ` happens more often.

Xaphiosis: Oh. That is a good point. Do you want me to change it now, or add a FIXME?

lsf37 (approved):
Apart from the issue that I think the <= maxIRQ condition should stay generic, this looks good from my side. Commit messages and structure also looks good.

proof/refine/Invariants_H.thy
lsf37: I'm fine to merge it now if we fix it up as one of the next steps -- if we leave it much later it's likely to become ingrained.

lsf37: (Unblocking the PR queue would be good, though, so maybe separate PR is good)

lsf37:
I should say: excellent work, it's hard to get this thing going and it looks like we're now well underway with an impressive about of lines deleted already.

proof/refine/Invariants_H.thy
Xaphiosis: Ok. I agree. If you want you can also do it in a next PR, it's going to be pretty easy: move the maxIRQ back, update the arch-specific definitions (to \<top> for non-RISCV64), and then I think at this stage of arch-split the proofs should go through with minimal or no fixes.

PR #855, opened 2025-02-24
Use projections in `schedulable`

michaelmcinerney:
Please also see the commit message

### DIFF
--- a/proof/invariant-abstract/AInvs.thy
+++ b/proof/invariant-abstract/AInvs.thy
@@ -29,13 +29,12 @@ lemma pred_tcb_at_upd_apply:
 
 lemma thread_set_tcb_arch_ct_schedulable[wp]:
   "thread_set (\<lambda>tcb. tcb\<lparr>tcb_arch := arch_tcb_context_set us (tcb_arch tcb)\<rparr>) t \<lbrace>ct_schedulable\<rbrace>"
+  apply (clarsimp simp: schedulable_def)
   apply (simp add: thread_set_def)
   apply (rule bind_wp[OF _ assert_get_tcb_ko'])
   apply (wpsimp wp: set_object_wp)
-  apply (fastforce simp: schedulable_def is_sc_active_def get_tcb_def ko_atD
-                         in_release_queue_def
-                  split: option.splits )
-  done
+  by (fastforce simp: obj_at_def opt_pred_def opt_map_def tcbs_of_kh_def scs_of_kh_def
+               split: option.splits)
 
 lemma thread_set_tcb_arch_ct_not_running[wp]:
   "thread_set (\<lambda>tcb. tcb\<lparr>tcb_arch := arch_tcb_context_set us (tcb_arch tcb)\<rparr>) t \<lbrace>\<lambda>s. \<not> ct_running s\<rbrace>"
@@ -53,7 +52,7 @@ lemma akernel_invs:
    (call_kernel e) :: (unit, unit) s_monad
    \<lbrace>\<lambda>_ s. invs s \<and> (ct_running s \<or> ct_idle s)\<rbrace>"
   unfolding call_kernel_def preemption_path_def
-  apply (wpsimp wp: activate_invs check_budget_invs charge_budget_invs is_schedulable_wp
+  apply (wpsimp wp: activate_invs check_budget_invs charge_budget_invs
                     update_time_stamp_invs hoare_drop_imps hoare_vcg_all_lift hoare_vcg_if_lift2)
   apply (fastforce intro: schact_is_rct_ct_active_sc
                     simp: schedulable_def2 ct_in_state_def pred_tcb_at_def obj_at_def)
@@ -65,7 +64,7 @@ lemma akernel_invs_det_ext:
    (call_kernel e) :: (unit, det_ext) s_monad
    \<lbrace>\<lambda>_ s. invs s \<and> (ct_running s \<or> ct_idle s)\<rbrace>"
   unfolding call_kernel_def preemption_path_def
-  apply (wpsimp wp: activate_invs check_budget_invs charge_budget_invs is_schedulable_wp
+  apply (wpsimp wp: activate_invs check_budget_invs charge_budget_invs
                     update_time_stamp_invs hoare_drop_imps hoare_vcg_all_lift hoare_vcg_if_lift2)
   apply (fastforce intro: schact_is_rct_ct_active_sc
                     simp: schedulable_def2 ct_in_state_def pred_tcb_at_def obj_at_def)
@@ -151,7 +150,7 @@ lemma schedule_ct_activateable:
         apply (wpsimp simp: schedule_switch_thread_fastfail_def tcb_sched_action_def
                             set_tcb_queue_def get_tcb_queue_def
                         wp: thread_get_wp' stt_activatable)
-       apply (wp add: is_schedulable_wp)+
+       apply wp+
     apply (rule hoare_strengthen_post[where Q'="\<lambda>_. invs and valid_sched"])
      apply wp
     apply (subgoal_tac "\<forall>x. scheduler_action s = switch_thread x \<longrightarrow> st_tcb_at activatable x s")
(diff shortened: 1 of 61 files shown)


michaelmcinerney:
ARM tests will fail, but I can apply the same changes as in RISCV to get it working. I'll update when I make other changes, if that's alright.

spec/abstract/KHeap_A.thy
```diff
@@ -304,50 +347,11 @@ definition
...
-     | Some tcb \<Rightarrow>
-       (case tcb_sched_context tcb of None => False
-        | Some sc_ptr =>
-            (runnable (tcb_state tcb) \<and> (is_sc_active sc_ptr s)
-              \<and> \<not>(in_release_queue tcb_ptr s)))"
+definition schedulable :: "obj_ref \<Rightarrow> 'z::state_ext state \<Rightarrow> bool" where
+  "schedulable t s \<equiv>
+   (runnable |< (tcbs_of s ||> tcb_state)) t
+   \<and> (sc_active |< (tcbs_of s |> tcb_sched_context |> scs_of s)) t
+   \<and> t \<notin> set (release_queue s)"
```
lsf37, line 354: This is so much nicer than the triplicate terrible manual case distinctions from before. Thanks for taking on that cleanup task.

spec/abstract/KHeap_A.thy
```diff
@@ -11,12 +11,55 @@ Functions to access kernel memory.
...
+
+lemma tcb_of_None:
+  "tcb_of ko = None \<longleftrightarrow> (\<forall>tcb. ko \<noteq> TCB tcb)"
+  by (cases ko; simp)
+
+definition tcbs_of_kh :: "('obj_ref \<rightharpoonup> kernel_object) \<Rightarrow> 'obj_ref \<rightharpoonup> tcb" where
+  "tcbs_of_kh kh \<equiv> kh |> tcb_of"
+
+abbreviation tcbs_of :: "'z state \<Rightarrow> obj_ref \<rightharpoonup> tcb" where
+  "tcbs_of s \<equiv> tcbs_of_kh (kheap s)"
```
lsf37, line 41: We don't usually do that additional wrapping in a definition for `kh`. It'll interfere somewhat with the projection stack, because the simplifier can't see the `|>` in the final term and can't apply the generic projection rules. Is there any specific reason for it or is it just what was there before? It looks a bit like the locales version of projections.

We could make it an abbreviation if we just want to have the name.

spec/abstract/KHeap_A.thy
```diff
@@ -11,12 +11,55 @@ Functions to access kernel memory.
...
+lemma sc_of_Some[simp]:
+  "sc_of ko = Some sc \<longleftrightarrow> (\<exists>n. ko = SchedContext sc n)"
+ by (cases ko; simp)
+
+lemma sc_of_None:
+  "sc_of ko = None \<longleftrightarrow> (\<forall>sc n. ko \<noteq> SchedContext sc n)"
+ by (cases ko; simp)
+
+definition scs_of_kh :: "('obj_ref \<rightharpoonup> kernel_object) \<Rightarrow> 'obj_ref \<rightharpoonup> sched_context" where
+  "scs_of_kh kh \<equiv> kh |> sc_of"
```
lsf37, line 57: (same as above, in case we do want to change it)

spec/abstract/KHeap_A.thy
michaelmcinerney: This stuff was just copied, and I’m quite sure it was written by Matt, so maybe there are some difference in intention and approach here. I don’t think having the name would be helpful. Could you maybe write out exactly what you’d like? Or I can take a stab at it. 

I also wasn’t sure whether to commit to having all the projections in this file. I can move the others here if we’d like to do that

proof/invariant-abstract/AInvs.thy
```diff
@@ -29,13 +29,12 @@ lemma pred_tcb_at_upd_apply:
...
+  apply (clarsimp simp: schedulable_def)
   apply (simp add: thread_set_def)
   apply (rule bind_wp[OF _ assert_get_tcb_ko'])
   apply (wpsimp wp: set_object_wp)
-  apply (fastforce simp: schedulable_def is_sc_active_def get_tcb_def ko_atD
-                         in_release_queue_def
-                  split: option.splits )
-  done
+  by (fastforce simp: obj_at_def opt_pred_def opt_map_def tcbs_of_kh_def scs_of_kh_def
+               split: option.splits)
```
lsf37, line 37: Those `cbs_of_kh_def scs_of_kh_def` are ones I meant with interfering with the simplifier.

proof/refine/RISCV64/Refine.thy
```diff
@@ -1268,37 +1235,27 @@ lemma ckernel_invariant:
      apply (rule hoare_weaken_pre)
       apply (rule kernelEntry_invs')
      apply clarsimp
+     apply (frule_tac a=sa and b=s in curthread_relation)
```
lsf37, line 1238: Does this one have rename_tac for `sa` and `s`  before? (I didn't spot any, but it's hard to see in the GitHub interface). I've just had a bunch of these fail for Isabelle2025..

proof/refine/RISCV64/TcbAcc_R.thy
```diff
@@ -3035,57 +3039,126 @@ lemma ko_at'_valid_tcbs'_valid_tcb':
...
+  "(\<And>s s'. \<lbrakk>(s, s') \<in> sr; P s; P' s'\<rbrakk> \<Longrightarrow> f s = v)
+   \<Longrightarrow> corres_underlying sr nf nf' (=) P P' (gets f) (return v)"
+  unfolding corres_underlying_def gets_def get_def return_def bind_def
+  by clarsimp
+
+definition schedulable' :: "machine_word \<Rightarrow> kernel_state \<Rightarrow> bool" where
+  "schedulable' tcbPtr s' \<equiv>
+   (runnable' |< (tcbs_of' s' ||> tcbState)) tcbPtr
+   \<and> ((\<lambda>sc. 0 < scRefillMax sc) |< (tcbs_of' s' |> tcbSchedContext |> scs_of' s')) tcbPtr
+   \<and> (Not \<circ> tcbInReleaseQueue |< tcbs_of' s') tcbPtr"
```
lsf37, line 3055: I think I have seen these terms occur somewhere in the proofs (CRefine?) -- I would still vote for giving them a name, at least the ones that are referenced again. It's not only that it makes it easier to understand, you can also click on the name and directly get to the entity that the proof makes a connection to.

lsf37 (approved):
This is cool, I'm all up for it. It makes things nicer.

Now that I've seen the entire PR, I can see that the `kh` definition is from the locale setup where it was moved from. I'm still tempted to make it an abbreviation, but it's possible that this has a bunch of fallout that would only pollute this PR. Maybe try it out and if things fail all over the place, wipe it again and keep what you have now.

You don't have to, but I suspect your future self might be grateful if you do the `FIXME RT: move to Lib` now instead of later :-)

spec/abstract/KHeap_A.thy
lsf37: Hm. That might turn into a greater rabbit hole than we like. If it's Matt's locales then we should move the locale setup as well, and then there is the question if we really want to keep that if we're doing the other projection style elsewhere. In terms of types and spec and expressivity they are the same, so for ASpec it's not really a problem, but I think I would only do the ones we need for now.

michaelmcinerney: Yes, thinking about it some more, it probably is a rabbit hole we want to avoid for now. I think if we were to get rid of anything here, then it's very likely to break some of the very sensitive simplification rules that are part of the locales, and it would be a monstrous task to fix that up. 

Should we just leave it here, maybe adding a comment before the definitions of `tcbs_of_kh` and `scs_of_kh` saying "Consider eliminating in favour of a direct definition below in terms of `kheap`"?

proof/refine/RISCV64/Refine.thy
michaelmcinerney: Oops, added now.

spec/abstract/KHeap_A.thy
lsf37: Yes, that sounds good. Maybe also a hint that there is locale setup for this later, so that we don't forget that this exists.

proof/refine/RISCV64/TcbAcc_R.thy
lsf37: I guess that one is short enough to just leave. It's short enough , and a name for this one would require some acrobatics around `tcbInReleaseQueue`, because that is close to the name I would probably expect for it.

michaelmcinerney:
I've got all the `FIXME RT: move` bits now in a PR for `master`, and I'd like to merge that PR and then cherry-pick it onto the `rt` branch before I update this PR, if that suits.

michaelmcinerney:
I've addressed the comments I've received so far, and updated the PR, but I've cancelled the tests, seeing that I'll want to cherry-pick some commits from `master` before merging this PR.

PR #861, opened 2025-03-03
More platform configs for CI

lsf37:
The first commit adds a broader spread of platforms to CI for every merge to cover GICv3 on AARCH64 and different IRQ and other platform values for ARM.

The second commit adds all supported platforms to the weekly test. This might be too many -- I'm not sure and we should watch cost as well. Ideas welcome.

### DIFF
--- a/.github/workflows/weekly-clean.yml
+++ b/.github/workflows/weekly-clean.yml
@@ -19,6 +19,53 @@ jobs:
       matrix:
         arch: [ARM, ARM_HYP, AARCH64, RISCV64, X64]
         num_domains: ['1', '7', '']
+        plat: ['']
+        include:
+          - arch: ARM
+            plat: exynos4
+          - arch: ARM
+            plat: exynos5410
+          - arch: ARM
+            plat: exynos5422
+          - arch: ARM
+            plat: hikey
+          - arch: ARM
+            plat: tk1
+          - arch: ARM
+            plat: zynq7000
+          - arch: ARM
+            plat: zynqmp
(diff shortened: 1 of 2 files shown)


.github/workflows/weekly-clean.yml
```diff
@@ -19,7 +19,53 @@ jobs:
       matrix:
         arch: [ARM, ARM_HYP, AARCH64, RISCV64, X64]
         num_domains: ['1', '7', '']
-    steps:
+        include:
```
Xaphiosis, line 22: The commit says "tested with the (empty) default for num_domains" but the above says `num_domains: ['1', '7', '']`. I don't understand

Xaphiosis (approved):
Looks ok to me, confused about num_domains

.github/workflows/weekly-clean.yml
lsf37: `include:` is separate from the matrix cross product, if it wanted to include `num_domains`, it would have to list it explicitly e.g.
```
          - arch: ARM_HYP
            plat: exynos5
            num_domains: 7
```
All parameters that are left off are empty, so this will test `num_domains` 1, 7, and 16 (default) for the default platform of ARM, ARM_HYP, AARCH64, RISCV64, X64, and then separately each explicitly listed arch/platform combination, with `num_domains` unset.

lsf37:
I'm still not sure if this will bring the test over the VM instance limit, but we'll see what happens on the weekend..

PR #862, opened 2025-03-03
aarch64: updates for GICv3 split EOI mode

lsf37:
Brings changes for split EOI (End Of Interrupt) mode from seL4/seL4#1183 into the verification.

Most of what changes in the seL4 pull request is underneath the machine interface and implements a different mechanism for masking/unmasking interrupts (using priority drop instead of actual masking/unmasking). For the purposes of the proof, what we want from the hardware remains the same: when deactivated, the interrupt will no longer fire.

Affects AARCH64 only.

Test with: seL4/seL4#1183

### DIFF
--- a/proof/crefine/AARCH64/Interrupt_C.thy
+++ b/proof/crefine/AARCH64/Interrupt_C.thy
@@ -15,9 +15,17 @@ lemma invokeIRQHandler_AckIRQ_ccorres:
   "ccorres dc xfdc
        invs' (UNIV \<inter> {s. irq_' s = ucast irq}) []
      (InterruptDecls_H.invokeIRQHandler (AckIRQ irq)) (Call invokeIRQHandler_AckIRQ_'proc)"
-  apply (cinit lift: irq_' simp: Interrupt_H.invokeIRQHandler_def invokeIRQHandler_def)
+  (* C has "if (config_set(CONFIG_ARM_GIC_V3_SUPPORT))", which will render as IF 0 = 1 or IF 1 = 1.
+     Pretend we don't know what the result of that is and make a case distinction that matches
+     the condition up with config_ARM_GIC_V3. *)
+  apply (cinit (no_ccorres_rewrite)
+               lift: irq_'
+               simp: Interrupt_H.invokeIRQHandler_def invokeIRQHandler_def dmo_if)
+   apply (rule ccorres_cond_both'[where Q=\<top> and Q'=\<top>])
+     apply (simp add: Kernel_Config.config_ARM_GIC_V3_def) (* match C condition *)
+    apply (ctac add: deactivateInterrupt_ccorres)
    apply (ctac add: maskInterrupt_ccorres)
-  apply simp
+  apply (simp add: theIRQ_def)
   done
 
 lemma getIRQSlot_ccorres:
--- a/proof/crefine/AARCH64/Machine_C.thy
+++ b/proof/crefine/AARCH64/Machine_C.thy
@@ -62,6 +62,12 @@ assumes maskInterrupt_ccorres:
            (doMachineOp (maskInterrupt m irq))
            (Call maskInterrupt_'proc)"
 
+(* This function is only implemented on GICv3 configs, hence the precondition *)
+assumes deactivateInterrupt_ccorres:
+  "ccorres dc xfdc (\<lambda>_. config_ARM_GIC_V3) (\<lbrace>\<acute>irq = ucast irq\<rbrace>) []
+           (doMachineOp (deactivateInterrupt irq))
+           (Call deactivateInterrupt_'proc)"
+
 (* This is a simplification until complete FPU handling is added at a future date *)
 assumes fpuThreadDelete_ccorres:
   "ccorres dc xfdc (tcb_at' thread) (\<lbrace>\<acute>thread = tcb_ptr_to_ctcb_ptr thread\<rbrace>) hs
--- a/proof/crefine/AARCH64/Syscall_C.thy
+++ b/proof/crefine/AARCH64/Syscall_C.thy
@@ -2229,6 +2229,24 @@ lemma ccorres_handleReservedIRQ:
                   simp flip: word_unat.Rep_inject split: if_splits)
   done
 
+(* Because of a case distinction, this proof occurs multiple times in handleInterrupt_ccorres below.
+   The proof is carefully manual to match up the C expression
+   "if (!config_set(CONFIG_ARM_GIC_V3_SUPPORT))", which renders as IF False or IF True in Simpl,
+   with config_ARM_GIC_V3 from the spec side. *)
+method maybe_maskInterrupt_before_ack =
+    simp only: maskIrqSignal_def when_def,
+    rule ccorres_cond_seq,
+    rule ccorres_cond_both[where P="\<lambda>_. \<not>config_ARM_GIC_V3" and R=\<top>],
+       simp add: Kernel_Config.config_ARM_GIC_V3_def, (* match up !config_set(..) condition *)
+      rule ccorres_gen_asm[where P="\<not>config_ARM_GIC_V3"],
+      simp,
+      ctac (no_vcg) add: maskInterrupt_ccorres,
+       ctac add: ackInterrupt_ccorres,
+      wp,
+     rule ccorres_gen_asm[where P="config_ARM_GIC_V3"],
+     simp,
+     ctac add: ackInterrupt_ccorres
+
 lemma handleInterrupt_ccorres:
   "ccorres dc xfdc
      (invs' and (\<lambda>s. irq \<in> non_kernel_IRQs \<longrightarrow> sch_act_not (ksCurThread s) s))
@@ -2236,19 +2254,20 @@ lemma handleInterrupt_ccorres:
      hs
      (handleInterrupt irq)
      (Call handleInterrupt_'proc)"
-  apply (cinit lift: irq_' cong: call_ignore_cong)
+  (* Use (no_ccorres_rewrite) to avoid rewriting the static condition coming out of
+     "if (!config_set(CONFIG_ARM_GIC_V3_SUPPORT))" in C *)
+  apply (cinit (no_ccorres_rewrite) lift: irq_' cong: call_ignore_cong)
    apply (rule ccorres_Cond_rhs_Seq)
     apply (simp add:  Kernel_C_maxIRQ del: Collect_const)
     apply (rule ccorres_rhs_assoc)+
     apply (subst doMachineOp_bind)
       apply (rule maskInterrupt_empty_fail)
      apply (rule ackInterrupt_empty_fail)
+    apply ccorres_rewrite
     apply (ctac add: maskInterrupt_ccorres)
       apply (subst bind_return_unit[where f="doMachineOp (ackInterrupt irq)"])
       apply (ctac add: ackInterrupt_ccorres)
-        apply (rule ccorres_split_throws)
-         apply (rule ccorres_return_void_C)
-        apply vcg
+        apply (rule ccorres_return_void_C)
        apply wp
       apply (vcg exspec=ackInterrupt_modifies)
      apply wp
@@ -2285,31 +2304,26 @@ lemma handleInterrupt_ccorres:
          apply csymbr
          apply csymbr
          apply (ctac (no_vcg) add: sendSignal_ccorres)
-          apply (simp add: maskIrqSignal_def)
-          apply (ctac (no_vcg) add: maskInterrupt_ccorres)
-           apply (ctac add: ackInterrupt_ccorres)
-          apply wp+
+          apply maybe_maskInterrupt_before_ack
+         apply clarsimp
+         apply wp
         apply (simp del: Collect_const)
         apply (rule ccorres_cond_true_seq)
         apply (rule ccorres_rhs_assoc)+
         apply csymbr+
         apply (rule ccorres_cond_false_seq)
-        apply (simp add: maskIrqSignal_def)
-        apply (ctac (no_vcg) add: maskInterrupt_ccorres)
-         apply (ctac add: ackInterrupt_ccorres)
-        apply wp
+        apply (rule ccorres_seq_skip[THEN iffD2])
+        apply maybe_maskInterrupt_before_ack
        apply (rule_tac P=\<top> and P'="{s. ret__int_' s = 0 \<and> cap_get_tag cap \<noteq> scast cap_notification_cap}" in ccorres_inst)
        apply (clarsimp simp: isCap_simps simp del: Collect_const)
        apply (case_tac rva, simp_all del: Collect_const)[1]
                   prefer 3
                   apply metis
-                 apply ((simp add: maskIrqSignal_def,
-                        rule ccorres_guard_imp2,
+                 apply ((rule ccorres_guard_imp2,
                         rule ccorres_cond_false_seq, simp,
-                        rule ccorres_cond_false_seq, simp,
-                        ctac (no_vcg) add: maskInterrupt_ccorres,
-                        ctac (no_vcg) add: ackInterrupt_ccorres,
-                        wp, simp)+)
+                        rule ccorres_cond_false_seq, rule ccorres_seq_skip[THEN iffD2],
+                        maybe_maskInterrupt_before_ack,
+                        simp)+)[11]
       apply (wpsimp wp: getSlotCap_wp)
      apply simp
      apply vcg
--- a/proof/invariant-abstract/AARCH64/ArchInterrupt_AI.thy
+++ b/proof/invariant-abstract/AARCH64/ArchInterrupt_AI.thy
@@ -82,10 +82,10 @@ lemma maskInterrupt_invs_ARCH[Interrupt_AI_assms]:
                          cur_tcb_def valid_irq_states_def valid_irq_masks_def)
   done
 
-crunch plic_complete_claim
+crunch plic_complete_claim (* FIXME AARCH64: remove plic_complete_claim *)
   for device_state_inv[wp]: "\<lambda>ms. P (device_state ms)"
 
-lemma dmo_plic_complete_claim[wp]:
+lemma dmo_plic_complete_claim[wp]: (* FIXME AARCH64: remove plic_complete_claim *)
   "do_machine_op (plic_complete_claim irq) \<lbrace>invs\<rbrace>"
   apply (wp dmo_invs)
   apply (auto simp: plic_complete_claim_def machine_op_lift_def machine_rest_lift_def in_monad select_f_def)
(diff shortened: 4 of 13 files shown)


lsf37:
I should add that this is trialling a new technique to support more configs: usually, when code is conditionally removed/added via `#ifdef MACRO`, we can only deal with that when it is its own architecture. I.e. we can deal with `#ifdef CONFIG_ARCH_RISCV`, because that will go into RISCV64 files and not other files. If the same happens within one architecture, such as within AARCH64 here, we have a problem, because CRefine sees two different programs depending on the value of the macro.

The solution here is to not use `#idef` in C, but an actual `if` statement together with the expression `defined(MACRO)`. The latter will come out as a static 0 or 1 after preprocessing, so CRefine will see both branches of the `if`, but with something like `if (1)`, depending on the value of the macro.

If we are careful to not rewrite the other branch away (as the compiler will do), but to match it up with the value of the config constant that is generated into the spec, we can do one proof that considers both branches and will work for both values of the config setting. This is similar to what Raf did before for NUM_DOMAINS, but used for config settings directly.

The downside is that C can no longer have functions that only exists for one config option, because the compiler will always see both branches (e.g. `deactivateInterrupt` has to exist for GICv2 and be empty, even though it is only ever used in GICv3).

proof/invariant-abstract/AARCH64/ArchInterrupt_AI.thy
```diff
@@ -109,6 +109,14 @@ lemma (* set_irq_state_valid_cap *)[Interrupt_AI_assms]:
 crunch set_irq_state
   for valid_global_refs[Interrupt_AI_assms]: "valid_global_refs"
 
+lemma deactivateInterrupt_invs:
+  "\<lbrace>invs and (\<lambda>s. interrupt_states s irq \<noteq> IRQInactive) and K config_ARM_GIC_V3\<rbrace>
+   do_machine_op (deactivateInterrupt irq)
+   \<lbrace>\<lambda>rv. invs\<rbrace>"
```
corlewis, line 115: Hmm, would `config_ARM_GIC_V3` make more sense as an assumption instead of in the preconditions? It's unlikely to make a difference in practice but having it as an assumption would make a proof fail immediately if a rule with it is used in the wrong configuration, while it being in in the precondition feels to me that it's more likely that it would make it to the final big condition and not be clear where it came from.

proof/crefine/AARCH64/Syscall_C.thy
```diff
@@ -2229,26 +2229,42 @@ lemma ccorres_handleReservedIRQ:
...
+    rule ccorres_cond_both[where P="\<lambda>_. \<not>config_ARM_GIC_V3" and R=\<top>],
+       simp add: Kernel_Config.config_ARM_GIC_V3_def, (* match up !defined(..) condition *)
+      rule ccorres_gen_asm[where P="\<not>config_ARM_GIC_V3"],
+      simp,
+      ctac (no_vcg) add: maskInterrupt_ccorres,
+       ctac add: ackInterrupt_ccorres,
+      wp,
+     rule ccorres_gen_asm[where P="config_ARM_GIC_V3"],
+     simp,
+     ctac add: ackInterrupt_ccorres
```
corlewis, line 2247: Wow, that can't have been a fun proof to work on and end up with!

I have to admit that seeing this method doesn't fill me with joy but if there's that sort of repetition then I guess this could be the least bad option. Hopefully this never needs any changes.

proof/invariant-abstract/AARCH64/ArchInterrupt_AI.thy
```diff
@@ -139,8 +147,7 @@ lemma invoke_irq_handler_invs'[Interrupt_AI_assms]:
    done
   show ?thesis
   apply (cases i, simp_all)
-    apply (wp dmo_plic_complete_claim maskInterrupt_invs)
-    apply simp+
+    apply (rule conjI; wpsimp wp: deactivateInterrupt_invs maskInterrupt_invs)
```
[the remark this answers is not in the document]
lsf37: Yes, that's from the if. It's not really different to making a case distinction, so I'm not really worried about it in a one-liner.

proof/invariant-abstract/AARCH64/ArchInterrupt_AI.thy
lsf37: I initially had it as an explicit assumption, but that does require it to be in the goal immediately, which is not always the case.

proof/crefine/AARCH64/Syscall_C.thy
```diff
@@ -2229,26 +2229,42 @@ lemma ccorres_handleReservedIRQ:
...
+
 lemma handleInterrupt_ccorres:
   "ccorres dc xfdc
      (invs' and (\<lambda>s. irq \<in> non_kernel_IRQs \<longrightarrow> sch_act_not (ksCurThread s) s))
      \<lbrace>\<acute>irq = ucast irq\<rbrace>
      hs
      (handleInterrupt irq)
      (Call handleInterrupt_'proc)"
-  apply (cinit lift: irq_' cong: call_ignore_cong)
+  apply (cinit (no_ccorres_rewrite) lift: irq_' cong: call_ignore_cong)
```
corlewis, line 2256: It might be worth a bit of a comment here about the `no_ccorres_rewrite`, like there is in `Interrupt_C.thy`. I don't think we've used it in our proofs before.

proof/invariant-abstract/AARCH64/ArchInterrupt_AI.thy
corlewis: Ah right, I was thinking that it would come from the config and always be available but I guess that it's more from the guarding ifs.

proof/invariant-abstract/AARCH64/ArchInterrupt_AI.thy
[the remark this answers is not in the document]
corlewis: We could hide the `if` behind a definition and give it a lifting rule that we make `[wp]`. I think that could end up being invisible in the proofs while protecting the `if` from being split.

Might be more overhead than it's worth though, like you said, it only appears as a one-liner in a few places.

proof/invariant-abstract/AARCH64/ArchInterrupt_AI.thy
lsf37: Yes, that’s right. In fact we try to keep the proof agnostic of which config is actually selected, so the only way it should know whether it is on or not should be via explicit case distinction or ‘if’.

proof/crefine/AARCH64/Syscall_C.thy
lsf37: Fair, there should be the same kind of comment here, yes. Raf had an idea for maybe wrapping the ‘defined’ such that it is not being rewritten, but I’d like to try that after we merge.

proof/crefine/AARCH64/Syscall_C.thy
lsf37: It was actually surprisingly easy — this is just the proof script from the first occurrence and with one small tweak it worked for all of the 13 cases.

corlewis (approved):
Looks good to me!

How do you feel about the new if technique now that you've put it into practice? I guess it only makes sense for relatively simple config options like this but is it a solution to part of our config matrix problem?

proof/crefine/AARCH64/Interrupt_C.thy
```diff
@@ -15,9 +15,17 @@ lemma invokeIRQHandler_AckIRQ_ccorres:
   "ccorres dc xfdc
        invs' (UNIV \<inter> {s. irq_' s = ucast irq}) []
      (InterruptDecls_H.invokeIRQHandler (AckIRQ irq)) (Call invokeIRQHandler_AckIRQ_'proc)"
-  apply (cinit lift: irq_' simp: Interrupt_H.invokeIRQHandler_def invokeIRQHandler_def)
+  (* C has "if (defined(CONFIG_ARM_GIC_V3))", which will render as IF 0 = 1 or IF 1 = 1.
```
Xaphiosis, line 18: You've mentioned this `defined()` before to me, but I don't see it anywhere... all I see is the macro `config_set`.
As we also discussed regarding stability, this is not stable w.r.t. ccorres_rewrite. What my current idea looks like is we have a C static inline function, say `bool_wrap` that takes a bool-like (int?) and returns that same bool-like (also int?). Then in util.h we'd have:

```
#define config_set(macro) bool_wrap(_is_set_(macro))
```

And when you get to it in crefine, the old traditional way you'd handle that would be csymbr to get the value back and do a ccorres_rewrite to take a particular branch... but this is precisely what we're trying to avoid here.
So with the new approach/idea:
* rewrite the return value of the bool_wrap to be the same as a value on the Haskell side with a split_rv_known (where you'd have to unwrap the value of the config on the abstract side)
* now you have both branches on the C side predicated on the same value that the abstract side takes without the exact value being available to the simplifier, and you can go down each branch independently
* ultimately somewhere the equivalence between 1/0 and the value of the config variable must be shown, but we want is as contained as possible
* if we do end up using this trick a lot, we can even make a tactic to sort it out

As you also said below, this can be done after the current PR because of time pressure, but I'd like to have stability available for the future.

proof/crefine/AARCH64/Machine_C.thy
```diff
@@ -62,6 +62,12 @@ assumes maskInterrupt_ccorres:
            (doMachineOp (maskInterrupt m irq))
            (Call maskInterrupt_'proc)"
 
+(* This function is only implemented on GICv3 configs, hence the precondition *)
+assumes deactivateInterrupt_ccorres:
+  "ccorres dc xfdc (\<lambda>_. config_ARM_GIC_V3) (\<lbrace>\<acute>irq = ucast irq\<rbrace>) []
```
Xaphiosis, line 67: very nice idea

proof/crefine/AARCH64/Syscall_C.thy
```diff
@@ -2229,26 +2229,42 @@ lemma ccorres_handleReservedIRQ:
                   simp flip: word_unat.Rep_inject split: if_splits)
   done
 
+(* Because of a case distinction, this proof occurs multiple times in handleInterrupt_ccorres below.
+   The proof is carefully manual to match up the C expression "if (!defined(CONFIG_ARM_GIC_V3))",
```
Xaphiosis, line 2233: again, I'm not sure what `defined()` is, did you mean `config_set()` or am I blind

proof/crefine/AARCH64/Syscall_C.thy
Xaphiosis: As with a certain someone's "apply hammer" philosophy showed, these are not that hard to write, but maintaining them can be a slog.

proof/crefine/AARCH64/Syscall_C.thy
Xaphiosis: Yes, please see my writeup in the comment in Interrupt_C ... I'm not a fan of disabling ccorres_rewrite like this, and believe there are other ways towards stability on the C side.

proof/invariant-abstract/AARCH64/ArchInterrupt_AI.thy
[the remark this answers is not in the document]
Xaphiosis: By the way, why not `split del: if_split`?

spec/haskell/src/SEL4/Machine/Hardware.lhs
```diff
@@ -66,7 +66,7 @@ Depending on the architecture, real physical addresses may be the same as the ad
 
 An interrupt request from an external device, or from the CPU's timer, is represented by the type "IRQ".
 
-> newtype IRQ = IRQ Arch.IRQ
+> newtype IRQ = IRQ { theIRQ :: Arch.IRQ }
```
Xaphiosis, line 69: oh, this ended up a much smaller fix than our discussion about caseconvs. nice.

spec/machine/AARCH64/MachineOps.thy
```diff
@@ -131,6 +136,14 @@ definition ackInterrupt :: "irq \<Rightarrow> unit machine_monad" where
 definition setInterruptMode :: "irq \<Rightarrow> bool \<Rightarrow> bool \<Rightarrow> unit machine_monad" where
   "setInterruptMode \<equiv> \<lambda>irq levelTrigger polarityLow. return ()"
 
+text \<open>Only exists on GICv3 platforms. We model interrupt deactivation as unmasking
+  for the purposes of the interrupt oracle.\<close>
```
Xaphiosis, line 140: out of curiosity, what's the difference between masking and activation on GICv3? on v2 I think there was only masking.

Xaphiosis (approved):
Some discussion for future directions. Also please check which one of us two is confused about `defined()` existing, as it's mentioned a few times.

proof/invariant-abstract/AARCH64/ArchInterrupt_AI.thy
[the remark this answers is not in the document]
lsf37: Because I actually want it to make the case distinction -- I don't think there is a problem that needs to be fixed, this kind of situation is what splitting `if` is for. We just tend to avoid it because it interferes with how `wp` wants to run (because of schematics), but in this case it doesn't interfere and works as intended.

proof/crefine/AARCH64/Syscall_C.thy
lsf37: I don't agree: this is not a hammer, it's a specific manual single-stepped proof that is being repeated. I have had to debug it and it's not hard.

spec/machine/AARCH64/MachineOps.thy
lsf37: There is a whole bunch of discussion on the seL4 PR about it. Basically there is an additional state the IRQ can be in where the interrupt is not being delivered, but not masked. I think Indan [describes it](https://github.com/seL4/seL4/pull/1183#discussion_r1979226825) well.

proof/crefine/AARCH64/Syscall_C.thy
Xaphiosis: You're right, it's missing that critical `(... | ... | ... | ... | ...)+` of a hammer. So it'll be hard to step through, but probably not that hard to debug in practice.

proof/crefine/AARCH64/Interrupt_C.thy
lsf37: Sorry about the `defined`/`config_set` confusion everywhere. `defined` is what you use in a preprocessor `#if` to get a 0/1. `config_set` is what you use in `if`, so you do have the correct macro above. Will update the comments.

There is a tradeoff to consider between wrapping each single `config_set` or only wrapping once around a larger expression, but I agree with the general idea. We could also have a `config_set` that wraps and one that does not so that we don't disturb code that we don't see. There are a bunch of experiments to run first, in particular if the compiler will actually produce the same binary (I think it will, but we need to check). 

It'd be nice to make a tactic, but I think we need to see some more pattern instances first, because the structure of the condition can vary, i.e. the SIMPL you get between `config_set(X)` and `!config_set(X)` esp when wrapped will be different enough to trip things up and if there is a `&&` in there, you get multiple statements etc. Maybe having a tactic for just the simple case will be good enough already. Anyway, to be seen, and agreed that it would be nice to make this stable and clean.

lsf37:
> Some discussion for future directions. Also please check which one of us two is confused about `defined()` existing, as it's mentioned a few times.

Definitely me, unfortunately :-) Will fix.

proof/crefine/AARCH64/Syscall_C.thy
lsf37: Yes, the main annoyance was to copy it and put `apply` in front of them. Eisbach does have breakpoints, but I keep forgetting how they work and you mainly need them for the actual hammers with `+`. The one above either succeeds quickly or fails quickly, the main time is spent in the `ctac` for `maskInterupt` and `ackInterrupt`, but even they fail immediately if they don't see the right function.

proof/crefine/AARCH64/Interrupt_C.thy
Xaphiosis: That's a good point about automatically wrapping vs manually wrapping were verification actually cares about both branches simultaneously. Either explicit wrap, or a separate `config_set` sound like a less intrusive idea (though it might confuse the kernel coders).

Oh yes, the tactic was just part of the sales pitch, definitely only needed if this starts happening more and more, but so far your platform ports work seems both very effective and popular.

spec/machine/AARCH64/MachineOps.thy
Xaphiosis: So, uh... the CPU can receive the interrupt (it's not masked), but not get woken up (because the priority is low) ... so if you then raise the priority, you'll get an interrupt immediately from the irq you already received? ... It's so clear, it truly brings tears to my eyes

lsf37: That's why the GICv3 spec only has a 1000 pages :-)

lsf37:
(updated comments)

spec/machine/AARCH64/MachineOps.thy
Indanz: So my explanation was slightly wrong in the sense that the IRQ's priority doesn't change, it's the current CPU's active priority that changes. In other words, the whole priority drop thing is misleading and can be ignored. What is important is that when dropping the priority, the IRQ is also marked as handled by the interrupt controller, and that is what stops the IRQ from waking up the CPU again until the "deactivation".

GIC already queues one additional interrupt per IRQ, that's why it has both active and active & pending states, so it's never been a simple on/off thing. I think it's to avoid losing interrupts that happen during handling of the current interrupt (practically only a thing for edge-triggered IRQs). So you could already immediately get a new interrupt for the IRQ you just handled, both in the old code as in the new one, that didn't change.

In the old code you would get it when unmasking, in the new code you'll get it when deactivating the IRQ. The only difference is that in the old code the IRQ would be pending, while in the split EOI mode the IRQ will be active & pending until the `seL4_IRQHandler_Ack` call. Then in both cases it will go to pending until we call `getActiveIRQ`, when it will become active.

The IRQ enable/disable bit is totally separate state from all the above.

The GIC spec is written badly, from the parts I've read. In the end it should just define what states there are, but it doesn't give a clear overview anywhere. It keeps being mixing terminology and being fuzzy about CPU interrupt state and per IRQ state.

proof/crefine/AARCH64/Interrupt_C.thy
lsf37: Logging here that the wrapping function does still let the compiler optimise: it produces equal code in the obj dump. The only visible change is the name of a compiler-internal name (CSWTCH) in the symbol table that is not referenced anywhere else.

PR #863, opened 2025-03-04
Prove `schedule_ccorres` and friends

michaelmcinerney:
Michael recommends reviewing commit-by-commit, though this shouldn’t make too much of a difference. 

Test with https://github.com/seL4/seL4/pull/1418


michaelmcinerney:
The proof of `setNextInterrupt_ccorres` has been updated after the latest change to the C which introduced a local variable for the head of the release queue.

proof/invariant-abstract/SchedContext_AI.thy
```diff
@@ -814,13 +814,22 @@ lemma commit_times_invs_helper:
                   wp: valid_irq_node_typ hoare_vcg_imp_lift')
   done
 
+lemma get_sc_active_sp:
+  "\<lbrace>P\<rbrace>
+   get_sc_active sc_ptr
+   \<lbrace>\<lambda>rv s. P s
+           \<and> (\<exists>sc n. ko_at (kernel_object.SchedContext sc n) sc_ptr s \<and> rv = (0 < sc_refill_max sc))\<rbrace>"
```
lsf37, line 821: Do we need the `kernel_object_.` qualifier here? I would have thought that at this point in the theories there is only one `SchedContext` constructor defined.

lsf37: Just saw that it was moved from Refine, where the qualifier would have been needed. But in AInvs we should be able to drop it.

proof/refine/RISCV64/Schedule_R.thy
```diff
@@ -2914,82 +2884,97 @@ crunch restart, suspend
...
+  apply (clarsimp simp: setNextInterrupt_def set_next_interrupt_def)
+  apply (rule corres_stateAssert_add_assertion[rotated])
+   apply (fastforce dest: curthread_relation intro!: tcb_at_cross
+                    simp: cur_tcb'_def vs_all_heap_simps obj_at_kh_kheap_simps is_tcb_def)
+  apply (rule corres_stateAssert_ignore)
+   apply (fastforce intro: ksReleaseQueue_asrt_cross)
+  apply (rule corres_split_forwards'[OF _ gets_sp getCurTime_sp])
+   apply corres
+  apply (rule corres_split_forwards'[OF _ gets_sp getCurThread_sp])
+   apply corres
```
lsf37, line 2911: It looks like for small steps like this it might be worth making a forward mode for `corres`, or a `fwd_corres` method maybe.

lsf37: Actually, now that I think about it, `corres` is already forwards. What does `corres_split_forwards'` do differently to `corres_split` or to just using `corres` for that step?

michaelmcinerney: I probably should have said something about this proof and its style. Seeing that `setNextInterrupt` is, for the most part, right up until the end, just grabbing a whole bunch of stuff from the state and returning the time to set for the next interrupt, I wrote this in a forwards style. I really do think this works nicely when we're just reading from the state and we have good sp rules. Given the rather sorry state of the previous proof and its cumbersome helper lemma, I think it's an improvement.

So `corres_split_forwards` just rearranges the assumptions of `corres_split`, and `corres_split_fowards` is just `corres_split_forwards` where the guards have just the one schematic, which is helpful when we're using sp rules. 

I haven't tried using `corres` for forwards stepping like this, actually. It's quite a long function, and it does have some asserts and things, which, from a little experiment I've just done, don't work immediately when just giving some rules to the `corres` method. It's definitely worth investigating, though.

proof/crefine/RISCV64/Schedule_C.thy
```diff
@@ -1543,12 +1535,336 @@ lemma refill_budget_check_ccorres:
...
+       apply (rule allI, rule conseqPre, vcg)
+       apply (clarsimp simp: setConsumedTime_def rf_sr_def cstate_relation_def Let_def
+                             modify_def get_def put_def bind_def carch_state_relation_def
+                             cmachine_state_relation_def)
+      apply wpsimp
+     apply (clarsimp simp: guard_is_UNIV_def)
+    apply (wpsimp wp: scActive_wp)
+   apply (vcg exspec=sc_active_modifies)
+  apply clarsimp
+  apply safe
```
lsf37, line 1745: Despite its name `safe` is not such a safe tactic :-). Would it make sense to say `apply (rule conjI; clarsimp)` instead here? Or maybe even `apply (intro conjI impI)`. That would provide some check against the goal form that is expected here and also for the subgoals that are being produced. With `safe` after a change you might get more or fewer goals and it'll be hard to match up the following proofs to what was expected.

lsf37 (approved):
Cool stuff. And nice use of `corres` tactic :-), even though I'm not entirely sure about the forward bits. They are not particularly getting in the way here, so I don't think they need changing, but was there a particular reason for them? Or just nicer structure in the proof in this case?

proof/refine/RISCV64/Schedule_R.thy
lsf37: Cool, thanks for that explanation. I think it makes sense to use here, so no problem with leaving it in. I think it would be entirely possible to make a `corres_fwd` that splits off a step and uses sp rules when there is still a concrete precondition. Not sure if/when I'll have the time to actually do it, though.

proof/crefine/RISCV64/Schedule_C.thy
```diff
@@ -1745,250 +2061,563 @@ lemma awaken_ccorres:
...
+           apply (clarsimp simp: cscheduler_action_relation_def)
+          (* toplevel case: action is choose new thread *)
+          apply (rule ccorres_cond_true_seq)
+          apply (rule ccorres_rhs_assoc)+
+          apply csymbr
+          (* ct runnable? *)
+          apply (ctac add: isSchedulable_ccorres, rename_tac schedulable)
+            apply (clarsimp simp: to_bool_def)
+            apply (rule ccorres_split_nothrow_dc)
+               (* enqueue if runnable *)
```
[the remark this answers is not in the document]
michaelmcinerney: These were in the old proof, but I'll take some praise for not deleting them :-)

proof/crefine/RISCV64/Schedule_C.thy
michaelmcinerney: Just `apply (rule conjI; clarsimp)` won't quite do it, but I can just replace the `safe` with `apply (intro conjI impI allI; clarsimp)` and it'll work with the rest of the proof unchanged.

I sometimes use `safe` when I'm exploring and not sure whether I've got provable subgoals. Is it a reasonable thing to use then? Or is it better to just go with the good old `intro conjI impI allI`?

lsf37: For exploration it's fine, because you will see if it misbehaves, but I would leave something more controlled in the final proof, so I usually just do the `intro conjI impI allI`. Come to think of it, it is so common that we might want to give it a name so that it is as little typing as `safe`, and actually safe in the sense we want.

PR #864, opened 2025-03-10
Prove `invokeSchedControl_ConfigureFlags_ccorres`

michaelmcinerney:
Please see the commit messages.

Test with https://github.com/seL4/seL4/pull/1423

### DIFF
--- a/proof/crefine/RISCV64/SchedContext_C.thy
+++ b/proof/crefine/RISCV64/SchedContext_C.thy
@@ -584,11 +584,249 @@ lemma schedContext_updateConsumed_ccorres:
 sorry (* FIXME RT: schedContext_updateConsumed_ccorres *)
 
 lemma invokeSchedControl_ConfigureFlags_ccorres:
-  "ccorres dc xfdc
-     invs' UNIV []
-     (invokeSchedControlConfigureFlags iv)
+  "ccorres (K (K \<bottom>) \<currency> dc) (liftxf errstate id (K ()) ret__unsigned_long_')
+     (invs' and (\<lambda>s. weak_sch_act_wf (ksSchedulerAction s) s)
+      and valid_sc_ctrl_inv'
+            (InvokeSchedControlConfigureFlags scPtr budget period mRefills badge flags))
+     (\<lbrace>\<acute>target___ptr_to_struct_sched_context_C = Ptr scPtr\<rbrace>
+      \<inter> \<lbrace>\<acute>budget = budget\<rbrace> \<inter> \<lbrace>\<acute>period = period\<rbrace> \<inter> \<lbrace>\<acute>max_refills = of_nat mRefills\<rbrace>
+      \<inter> \<lbrace>\<acute>badge___unsigned_long = badge\<rbrace> \<inter> \<lbrace>\<acute>flags = flags\<rbrace>) hs
+     (liftE (invokeSchedControlConfigureFlags
+              (InvokeSchedControlConfigureFlags scPtr budget period mRefills badge flags)))
      (Call invokeSchedControl_ConfigureFlags_'proc)"
-sorry (* FIXME RT: invokeSchedControl_ConfigureFlags_ccorres *)
+  supply Collect_const [simp del]
+  apply (cinit' lift: target___ptr_to_struct_sched_context_C_' budget_' period_' max_refills_'
+                      badge___unsigned_long_' flags_'
+                simp: invokeSchedControlConfigureFlags_def gets_bind_ign liftE_liftM)
+   apply (rule ccorres_pre_getObject_sc, rename_tac sc)
+   apply (rule ccorres_move_c_guard_sc)
+   apply (rule_tac xf'=thread_'
+               and val="option_to_ctcb_ptr (scTCB sc)"
+               and R="obj_at' (\<lambda>sc'. scTCB sc' = scTCB sc) scPtr and valid_objs'"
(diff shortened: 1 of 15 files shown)


proof/refine/RISCV64/KHeap_R.thy
```diff
@@ -4398,6 +4398,22 @@ lemma no_fail_updateSchedContext[wp]:
...
+  apply (simp add: updateSchedContext_def)
+  apply (wpsimp wp: set_sc'.obj_at')
+  apply (clarsimp split: if_splits simp: obj_at'_real_def ko_wp_at'_def)
+  done
+
+lemma updateSchedContext_sc_obj_at'_inv:
+  "(\<And>sc. P (f sc) = P sc) \<Longrightarrow> updateSchedContext scPtr f \<lbrace>\<lambda>s. Q (obj_at' P scPtr' s)\<rbrace>"
+  unfolding updateSchedContext_def
+  apply (wpsimp wp: set_sc'.obj_at')
+  by (clarsimp split: if_splits simp: obj_at'_real_def ko_wp_at'_def)
```
Xaphiosis, line 4415: if you're using by on a two-liner, make it a single `by ... ...`

lsf37 (approved):
Looks good. I agree with Raf's suggestion for the small `by` tweak.

proof/crefine/RISCV64/SchedContext_C.thy
```diff
@@ -584,11 +584,251 @@ lemma schedContext_updateConsumed_ccorres:
...
+                  apply clarsimp
+                  apply (rule ccorres_cond_false)
+                  apply ccorres_rewrite
+                  apply (rule ccorres_cond_true)
+                  apply (ctac add: rescheduleRequired_ccorres)
+                 apply (clarsimp simp: when_def to_bool_def)
+                 apply (rule_tac Q="\<lambda>s. ctPtr = ksCurThread s" in ccorres_cond_both'[where Q'=\<top>])
+                   apply (clarsimp dest!: rf_sr_ksCurThread simp: option_to_ctcb_ptr_def)
+                  (* the target sc is not associated with the current thread, but is associated
+                     with a runnable thread, so we perform possibleSwitchTo *)
```
Xaphiosis, line 724: comment on commentary: this "so we perform" is both novel and kind of long-winded, could I suggest that we go back to calling or invoking functions? i.e. "the target is [...], call possibleSwitchTo"
(same for other occurrences)

proof/crefine/RISCV64/SchedContext_C.thy
```diff
@@ -584,11 +584,251 @@ lemma schedContext_updateConsumed_ccorres:
...
+           apply clarsimp
+           apply ccorres_rewrite
+           apply (rule ccorres_cond_seq)
+           apply (rule ccorres_cond_false)
+           apply ccorres_rewrite
+           apply (rule ccorres_cond_seq)
+           apply (rule ccorres_cond_false)
+           apply ccorres_rewrite
+           apply (rule ccorres_cond_false)
+           apply (ctac (no_vcg) add: refill_new_ccorres)
```
Xaphiosis, line 690: little bit confused here; I'm only seeing one ccorres_cases above, we're in the not active case, but you're clearing out a bunch of branches (and they're not clearing up by themselves either, it's quite forceful) - can you say what's happening here?

Xaphiosis (approved):
Proofs look good. For the history, I'd prefer the "do everything" commit that changes spec+proofs to say something about what's going on: there's a fix for a spec bug in there, and an actual spec change, but no info on what happened or why.

proof/crefine/RISCV64/SchedContext_C.thy
Xaphiosis: to be fair, in the context of seL4, we do "perform" invocations, but we shouldn't extend that to everything

proof/crefine/RISCV64/SchedContext_C.thy
michaelmcinerney: Yes, in the case that the sched context is not active, we will perform `refill_new`. But the Simpl will just use the variable `ret_int_'` to keep track of the truth value of the guards that it uses. So when the sc is not active, it assigns `0` to `ret_int_'`, and then it will take the false branch of every conditional until we eventually get to `refill_new`

Xaphiosis: right, but that's why I'm surprised... it sounds like simp, ccorres_rewrite, simp, ccorres_rewrite, etc, since it never takes the branches, but you're doing explicit ccorres_rewrite_cond_false which defers showing that it's false until later

michaelmcinerney: Sorry, is `ccorres_rewrite_cond_false` a rule? If so, I can't find it. I'm not sure how else this can be handled. I really do not like the way that the CParser handles these sorts of `if` statements and uses local variables to determine the guards. If there is a nicer way, please let me know.

Xaphiosis: Apologies, that's a typo by me, I meant the `ccorres_cond_false` rule you're using. What I'm trying to get at is that if we know the return for the first ret_int, we can csymbr that, and we end up with something in the next IF branch condition. If we simp that, it should boil down to the empty set, which we can discard with ccorres_rewrite. So my mind is seeing some kind of a `csymbr, simp, ccorres_rewrite` that will clear up the branches. It's ok to do it the way you did, but it is surprising to be deferring these IF conditions 'til later when they seem obvious in the current context.

PR #866, opened 2025-03-11
arm-hyp+aarch64 haskell+crefine: fix saved-when-disabled registers

ryybrr:
Mark additional registers as being saved when a VCPU is disabled

Test with seL4/seL4#1384


lsf37 (approved):
Pleasingly painless.

PR #867, opened 2025-03-12
aarch64: avoid saving CPACR to a disabled VCPU

ryybrr:
Avoid saving CPACR to a VCPU that has already been disabled

Test with seL4/seL4#1404


lsf37 (approved):
Looking good!

PR #868, opened 2025-03-14
github: update to central workflows

lsf37:
Use GitHub workflow_call feature to reduce workflow duplication.

~~The central workflows in https://github.com/seL4/ci-actions/blob/master/.github/workflows/ don't support the `READ_TOKEN` bit yet, so we should hold off on merging until that is done.~~

~~I also still need to implement the ignore file format for the link checker.~~

~~Will set to non-draft when this is done.~~

Both features are implemented (https://github.com/seL4/ci-actions/pull/393 and https://github.com/seL4/ci-actions/pull/390), now ready for review.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -1,4 +1,4 @@
-# Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+# Copyright 2025, Proofcraft Pty Ltd
 #
 # SPDX-License-Identifier: BSD-2-Clause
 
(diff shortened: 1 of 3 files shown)


.github/workflows/pr.yml
```diff
@@ -1,42 +1,16 @@
-# Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+# Copyright 2025, Proofcraft Pty Ltd
```
corlewis, line 1: Don't we need to keep the old copyright, or does this remove enough that it doesn't make sense now?

Either way, presumably we want to change the copyright for push.yml as well.

corlewis (approved):

.github/workflows/pr.yml
lsf37: There wasn't anything left of the old content, really, that's why I removed it. You are right that I should also do something in push.yml.

PR #869, opened 2025-03-17
arm-hyp: save and restore CNTKCTL

ryybrr:
Save CNTKCTL alongside other virtual timer registers when switching VCPUs

Test with seL4/seL4#1425


lsf37 (approved):
Looks good.

PR #870, opened 2025-03-20
Some sorries in `Syscall_C`

michaelmcinerney:
Please see the commit messages for a rundown on what's here.

Michael recommends reviewing commit-by-commit.

Test with https://github.com/seL4/seL4/pull/1429


proof/refine/RISCV64/Ipc_R.thy
```diff
@@ -2176,13 +2176,66 @@ lemma refillSufficient_sp:
    \<lbrace>\<lambda>rv s. P s \<and> (\<exists>sc. scs_of' s scp = Some sc \<and> rv = (refillSufficient usage (refillHd sc)))\<rbrace>"
   by wpsimp
 
+lemma cap_in_tcbFaultHandlerSlot:
+  "\<lbrakk>ko_at' (tcb :: tcb) t s; valid_tcbs' s\<rbrakk>
+   \<Longrightarrow> \<exists>cap. cte_wp_at' (\<lambda>c. cteCap c = cap) (t + 2 ^ cte_level_bits * tcbFaultHandlerSlot) s
+             \<and> valid_cap' cap s
+             \<and> (\<exists>n\<in>dom tcb_cte_cases. cte_wp_at' (\<lambda>cte. cteCap cte = cap) (t + n) s)"
```
lsf37, line 2183: I have the feeling that `t + 2 ^ cte_level_bits * tcbFaultHandlerSlot` has a name somewhere, or at least it should. It may be written in a slightly different form (e.g. `1 << _`), or in the worst case even expanded. Don't quite remember where, though. I think @Xaphiosis encountered these recently in arch-splitting Refine.

lsf37 (approved):
Cool. That is a lot of solved sorries!

I'm not so sure about the large number of small commits, at least some of them should be combined. I don't mind the one commit per solved function for this, because it makes it easier to see progress in the history, but I don't think we need single commits for each `declared [simp]`. Speaking of these, we have often regretted making things globally `simp`. Are you sure about them? (They do seem benign, just double-checking).

michaelmcinerney:
> I'm not so sure about the large number of small commits, at least some of them should be combined. I don't mind the one commit per solved function for this, because it makes it easier to see progress in the history, but I don't think we need single commits for each `declared [simp]`. Speaking of these, we have often regretted making things globally `simp`. Are you sure about them? (They do seem benign, just double-checking).

I do like having the Haskell assert `defs` as `[simp]`, and I don't think it would be something we will end up regretting. In a `corres` rule you'll have to prove that the asserted property is true and often they'll be used in conjunction with one of the `add_` lemmas like `add_cur_tcb'`. So if the assert def is in the simp set, they often go away immediately. If they're not in the simp set, then I'll often stare directly at `_asrt` and wonder why it's not going away, until I realise that I just need to unfold the definition. In `ccorres` rules, we often get to the last two subgoals and, if they asserts defs aren't in the simp set, we'll need to unfold sometimes several of them before we can start using them. So I do find it very convenient.

As far as squashing some of the commits, I'd be happy to, but I'm not sure which should be squashed and into which commits. Adding those assert defs to the simp set often ends up with lines becoming unnecessary, and those lines are often unrelated to the other sorries I'm working on, so I put them into their own commit. I guess I could squash the CSlot pointer functions into another commit where they're used, but they're used in two commits, actually. `getCurSc` is used all over the place, so I didn't think it should be squashed into any of the commits for a sorry, here. The lib commit is from `master`, too.

lsf37:
> I do like having the Haskell assert `defs` as `[simp]`, and I don't think it would be something we will end up regretting. In a `corres` rule you'll have to prove that the asserted property is true and often they'll be used in conjunction with one of the `add_` lemmas like `add_cur_tcb'`. So if the assert def is in the simp set, they often go away immediately. If they're not in the simp set, then I'll often stare directly at `_asrt` and wonder why it's not going away, until I realise that I just need to unfold the definition. In `ccorres` rules, we often get to the last two subgoals and, if the asserts defs aren't in the simp set, we'll need to unfold sometimes several of them before we can start using them. So I do find it very convenient.
 
Alright, happy with that. I'll let @Xaphiosis comment on the commits, he will likely have opinions about them anyway :-)

spec/haskell/src/SEL4/Kernel/Thread.lhs
```diff
@@ -834,18 +834,22 @@ Kernel init will created a initial thread whose tcbPriority is max priority.
 
 > endTimeslice :: Bool -> Kernel ()
 > endTimeslice canTimeoutFault = do
->     ct <- getCurThread
+>     stateAssert cur_tcb'_asrt
```
Xaphiosis, line 837: very minor: `stateAssert cur_tcb'_asrt` fits on one line with the string, so you can save a few vertical lines if you want

spec/haskell/src/SEL4/Kernel/FaultHandler.lhs
```diff
@@ -33,11 +33,22 @@ This module contains functions that determine how recoverable faults encountered
...
->     case cteCap (tcbTimeoutHandler tcb) of
+>     stateAssert (tcb_at'_asrt tptr)
+>         "Assert that there is a TCB at tptr"
+>     timeoutHandlerSlot <- getThreadTimeoutHandlerSlot tptr
+>     timeoutHandlerCap <- getSlotCap timeoutHandlerSlot
+>     case timeoutHandlerCap of
 >         EndpointCap {} -> return True
 >         _ -> return False
 
+These functions will return a physical pointer to a thread's fault or timeout handler slot.
```
Xaphiosis, line 44: why "will return"? you use that phrasing for assert statements whose definitions will be defined later... but the functions do return a pointer? I'd suggest:

`These functions return a physical pointer to a thread's fault and timeout handler slot respectively.`

spec/abstract/Ipc_A.thy
```diff
@@ -537,33 +537,40 @@ where
 
 section \<open>Sending Fault Messages\<close>
 
-text \<open>When a thread encounters a fault, retreive its fault handler capability
-and send a fault message.\<close>
-definition
-  send_fault_ipc :: "obj_ref \<Rightarrow> cap \<Rightarrow> fault \<Rightarrow> bool \<Rightarrow> (bool, 'z::state_ext) f_monad"
-where
+text \<open>
+  When a thread encounters a fault, retreive its fault handler capability
```
Xaphiosis, line 541: since we're here: retrieve

proof/refine/RISCV64/TcbAcc_R.thy
```diff
@@ -3029,9 +3029,16 @@ lemmas no_fail_scActive[wp] =
...
 
+defs tcb_at'_asrt_def:
+  "tcb_at'_asrt \<equiv> \<lambda>tcbPtr s. tcb_at' tcbPtr s"
+
+declare tcb_at'_asrt_def[simp]
+
 defs sc_at'_asrt_def:
   "sc_at'_asrt \<equiv> \<lambda>scPtr s. sc_at' scPtr s"
 
+declare sc_at'_asrt_def[simp]
```
Xaphiosis, line 3040: It is a shame that the defs command doesn't take [simp] as a modifier, because the declare is a bit of an eyesore.

proof/refine/RISCV64/Syscall_R.thy
```diff
@@ -1970,8 +1902,14 @@ lemma chargeBudget_corres:
...
    apply (fastforce intro: cur_sc_tcb_sc_at_cur_sc)
   apply add_cur_tcb'
+  apply (rule corres_stateAssert_ignore, simp)
+  apply clarsimp
+  apply (rule corres_stateAssert_ignore, simp)
+   apply (fastforce intro: sch_act_wf_cross)
   apply (rule corres_underlying_split[rotated 2, OF gets_sp getCurSc_sp])
    apply (corresKsimp corres: getCurSc_corres)
+  apply (rule corres_stateAssert_ignore, simp)
+   apply (fastforce intro: active_sc_at'_cross)
```
Xaphiosis, line 1912: this looks a bit odd... the first `corres_stateAssert_ignore` looks like it could be combined with the subsequent clarsimp:
`apply (rule corres_stateAssert_ignore; clarsimp)`
the second and third look like that simp isn't doing much, and they could be:
`apply (rule corres_stateAssert_ignore, fastforce intro: sch_act_wf_cross)`
and
`apply (rule corres_stateAssert_ignore, fastforce intro: active_sc_at'_cross)`

proof/refine/RISCV64/Refine.thy
```diff
@@ -756,10 +754,10 @@ lemma kernel_preemption_corres:
...
       apply clarsimp
      apply wpsimp
-    apply (rule_tac Q'="\<lambda>irq s. invs' s \<and> sc_at' (ksCurSc s) s \<and>
-                               (\<forall>irq'. irq = Some irq' \<longrightarrow>
-                                         intStateIRQTable (ksInterruptState s ) irq' \<noteq> IRQInactive)"
-           in hoare_post_imp)
+    apply (rule_tac Q'="\<lambda>irq s. invs' s \<and> sc_at' (ksCurSc s) s \<and> cur_tcb' s \<and>
+                                (\<forall>irq'. irq = Some irq' \<longrightarrow>
+                                        intStateIRQTable (ksInterruptState s ) irq' \<noteq> IRQInactive)"
+             in hoare_post_imp)
```
Xaphiosis, line 760: if you're fiddling with indent, I think you want the `in` further right

lib/Corres_UL.thy
```diff
@@ -947,6 +947,11 @@ lemma corres_assert_assume_l:
   \<Longrightarrow> corres_underlying sr nf nf' rrel (P and (\<lambda>s. P')) Q (assert P' >>= f) g"
   by (force simp: corres_underlying_def assert_def return_def bind_def fail_def)
 
+lemma corres_assert_assume_l_forward:
+  "\<lbrakk>\<And>s. P s \<Longrightarrow> P'; corres_underlying sr nf nf' rrel (P and K P') Q (f ()) g\<rbrakk>
+  \<Longrightarrow> corres_underlying sr nf nf' rrel P Q (assert P' >>= f) g"
+  by (force simp: corres_underlying_def assert_def return_def bind_def fail_def)
```
Xaphiosis, line 953: I'm puzzled by this. It appears if you use the lemma, you immediately get that `P s ==> P'` as your first goal, which means you'd have to specify P', or defer/rotate it and work on the corres_underlying first. What am I missing?

proof/crefine/RISCV64/Syscall_C.thy
```diff
@@ -1234,55 +1246,500 @@ sorry (* FIXME RT: handleRecv_ccorres
...
-sorry (* FIXME RT: checkBudgetRestart_ccorres *)
+  supply Collect_const[simp del]
+  apply cinit
+   apply (rule ccorres_stateAssert)
+   apply (ctac (no_vcg) add: checkBudget_ccorres, rename_tac result result')
+    apply (rule ccorres_pre_getCurThread)
+    apply csymbr
+    apply clarsimp
+    apply (rule_tac P="to_bool result'" in ccorres_cases)
+     (* checkBudget returned True, so we simply return True *)
```
Xaphiosis, line 1601: um... there is no need to say "so we" for everything that occurs; "checkBudget returned True -> return True" is sufficient

below: "so if the current thread is runnable, we set its thread state to Restart, and then return False" -> "; when the current thread is runnable, set its thread state to Restart and return False"

It's good language for papers, but the comments are describing imperative code. Pretty please consider, since it is a pattern appearing in this PR in many places.

proof/refine/RISCV64/Ipc_R.thy
Xaphiosis: In `tcb_cte_cases_simps`, we have something like this on master (as part of the new arch-split and `tcb_cte_cases_neqs` infrastructure):
`tcb_cte_cases (3 << cteSizeBits) = Some (tcbCaller, tcbCaller_update)`
so I expect that the compressed version for this case might be:
`t + (tcbFaultHandlerSlot << cteSizeBits)`? in fact, something like this shows up below

Xaphiosis (approved):
The spec improvements on this are a big relief, clearly the spec was wonky and I'm glad you sorted it out.
I agree with the [simp] for the Haskell asserts; we have to wrap them in a name in order to point at them in Haskell, but were we to write them by hand in the design spec, we'd probably write the actual state assertion verbatim, unless it spilled out.
Will read through the commits and see what squash suggestions come up.

Xaphiosis:
Commit review time.

I think the `prove ..._corres` commits can stay as they are. Good indication of progress, and progress is the topic of the day of MCS, making them topical commits.

The `rt riscv refine: make ..._asrt_def [simp]` belong with the commits that introduced the assertions, but those are not in this PR. I propose fusing them together (by the way we should say "declare" not "make"):

```
rt riscv refine: sch_act_wf_asrt_def/declare sc_at'_asrt_def [simp]

We only need the names for Haskell to use, for proofs we want their contents.
```

The rest, even though there's a fair few them, are actually topical commits, so no real benefit from squashing.

lib/Corres_UL.thy
michaelmcinerney: This is a forwards style rule; I put `forward` in the name to indicate this, though as Gerwin pointed out recently, we do have many forwards style rules but not a proper naming convention.

So this rule would be used when the guards are concrete, not schematic. It will generate the first subgoal for you, which you immediately solve, and then you are left with a new `corres_underlying` goal, with concrete guards, and now with the property that was just asserted as a new conjunction in the Haskell guard.

Xaphiosis: Yes, that's what I was missing. Thank you! All good.

spec/haskell/src/SEL4/Kernel/FaultHandler.lhs
lsf37: It's the same text as for the existing getSlot functions. If we are changing it, I would prefer not to call them physical pointers, because that whole PPtr terminology in Haskell has always been terminally confusing once things hit virtual memory. I'd just say "a pointer".

proof/refine/RISCV64/Syscall_R.thy
michaelmcinerney: I was able to tidy up this area of the proof. One of the `clarsimp`s was needed because we just did an `unfolding` that left a `K_bind`, which I don't like, I combined the `fastforce`s as you wrote, and then changed a `corresKsimp` to a `corres`.

proof/refine/RISCV64/TcbAcc_R.thy
lsf37: The syntax of the command is now under our control, actually. We could probably make it take attributes if we wanted to.

proof/refine/RISCV64/Syscall_R.thy
Xaphiosis: Cool. Thank you, and the switch to `corres` gives a chance that this will be understandable if it breaks later.

proof/refine/RISCV64/TcbAcc_R.thy
Xaphiosis: Do you think we should? Or make an issue? It is kinda rare.

Xaphiosis: cheers

PR #875, opened 2025-03-23
config_set wrap and GICv3 support for ARM and ARM_HYP

lsf37:
This PR applies the GICv3/split EOI changes also to ARM and ARM_HYP, because 32-bit boards can have a GICv3 after all -- when they are actually Armv8 boards that run in 32 bit mode.

This now requires small changes in DRefine/Access/Infoflow that were not present on AARCH64.

In the previous GICv3/split EOI PR (#862) we discussed making the proofs more stable for `ccorres_rewrite` and the `config_set` macro. The PR here does this now, and is necessary for the corresponding seL4 PR https://github.com/seL4/seL4/pull/1430. The two parts are in separate commits, but probably best reviewed together, hence in the same PR.


proof/drefine/Corres_D.thy
```diff
@@ -11,7 +11,7 @@
  *)
 
 theory Corres_D
-imports StateTranslationProofs_DR
+imports StateTranslationProofs_DR Lib.Corres_Method
```
corlewis, line 14: Nice to see us starting to gradually improve `drefine` as well with the `corres` method!

corlewis (approved):
Looks good to me, and the improved stability from the `config_set` change is very nice!

lsf37:
Forgot the `Test with` bit, re-running the proof checks..

Indanz:
I don't think ARM_HYP has GICv3 support yet, at least not in the C code. Probably easy to add if anyone wants it.

lsf37:
> I don't think ARM_HYP has GICv3 support yet, at least not in the C code. Probably easy to add if anyone wants it.

It does -- the imx8mm board has hyp support in 32 bit mode and runs the GICv3 driver in that mode.

Indanz:
> > I don't think ARM_HYP has GICv3 support yet, at least not in the C code. Probably easy to add if anyone wants it.
> 
> It does -- the imx8mm board has hyp support in 32 bit mode and runs the GICv3 driver in that mode.

No, it doesn't (note the missing `arm_hyp`):

https://github.com/seL4/seL4/blob/master/src/plat/imx8m-evk/config.cmake#L13

It would be very surprising if it did though:

https://github.com/seL4/seL4/blob/master/include/arch/arm/arch/machine/gic_v3.h#L106

lsf37:
> > It does -- the imx8mm board has hyp support in 32 bit mode and runs the GICv3 driver in that mode.
> 
> No, it doesn't (note the missing `arm_hyp`):
> 
> https://github.com/seL4/seL4/blob/master/src/plat/imx8m-evk/config.cmake#L13
> 
> It would be very surprising if it did though:
> 
> https://github.com/seL4/seL4/blob/master/include/arch/arm/arch/machine/gic_v3.h#L106

Right, so that was actually premature for ARM_HYP and would fail if configured. Well, at least the infrastructure is going to be there when we do support it at some point.

PR #876, opened 2025-03-25
access+infoflow: relax conditions

ryybrr:
* Remove various instances of `pspace_aligned`, `valid_vspace_objs`, and `valid_asid_table`/`valid_arch_state` from preconditions
* Relax the `is_subject` predicate in various Syscall_AC lemmas
* Enforce explicit `wp` rules in locale interpretations

### DIFF
--- a/proof/access-control/ARM/ArchDomainSepInv.thy
+++ b/proof/access-control/ARM/ArchDomainSepInv.thy
@@ -25,6 +25,21 @@ lemma arch_finalise_cap_rv[DomainSepInv_assms]:
   "\<lbrace>\<lambda>_. P (NullCap,NullCap)\<rbrace> arch_finalise_cap c x \<lbrace>\<lambda>rv _. P rv\<rbrace>"
   unfolding arch_finalise_cap_def by wpsimp
 
+crunch
+  invalidate_tlb_by_asid, handle_reserved_irq, handle_vm_fault,
+  handle_hypervisor_fault, handle_arch_fault_reply, arch_mask_irq_signal,
+  arch_switch_to_thread, arch_switch_to_idle_thread, arch_activate_idle_thread
+  for domain_sep_inv[DomainSepInv_assms, wp]: "domain_sep_inv irqs st"
+
+lemma arch_derive_cap_domain_sep_inv[DomainSepInv_assms, wp]:
+  "\<lbrace>\<top>\<rbrace> arch_derive_cap acap \<lbrace>\<lambda>rv _. domain_sep_inv_cap irqs rv\<rbrace>,-"
+  unfolding arch_derive_cap_def
+  by wpsimp
+
+lemma arch_post_modify_registers_domain_sep_inv[DomainSepInv_assms, wp]:
+  "arch_post_modify_registers cur x31 \<lbrace>domain_sep_inv irqs st\<rbrace>"
+  unfolding arch_post_modify_registers_def by wpsimp
+
 end
 
 
@@ -32,17 +47,11 @@ global_interpretation DomainSepInv_1?: DomainSepInv_1
 proof goal_cases
   interpret Arch .
   case 1 show ?case
-    by (unfold_locales; fact DomainSepInv_assms)
+    by (unfold_locales; (fact DomainSepInv_assms)?)
 qed
 
 
 context Arch begin global_naming ARM_A
-
-crunch
-  invalidate_tlb_by_asid, handle_reserved_irq, handle_vm_fault,
-  handle_hypervisor_fault, handle_arch_fault_reply, arch_mask_irq_signal,
-  arch_switch_to_thread, arch_switch_to_idle_thread, arch_activate_idle_thread
-  for domain_sep_inv[DomainSepInv_assms, wp]: "domain_sep_inv irqs st"
 
 lemma perform_page_invocation_domain_sep_inv:
   "\<lbrace>domain_sep_inv irqs st and valid_page_inv pgi\<rbrace>
@@ -120,15 +129,6 @@ lemma arch_invoke_irq_control_domain_sep_inv[DomainSepInv_assms]:
    apply (wpsimp wp: do_machine_op_domain_sep_inv simp: arch_irq_control_inv_valid_def)+
   done
 
-lemma arch_derive_cap_domain_sep_inv[DomainSepInv_assms, wp]:
-  "\<lbrace>\<top>\<rbrace> arch_derive_cap acap \<lbrace>\<lambda>rv _. domain_sep_inv_cap irqs rv\<rbrace>,-"
-  unfolding arch_derive_cap_def
-  by wpsimp
-
-lemma arch_post_modify_registers_domain_sep_inv[DomainSepInv_assms, wp]:
-  "arch_post_modify_registers cur x31 \<lbrace>domain_sep_inv irqs st\<rbrace>"
-  unfolding arch_post_modify_registers_def by wpsimp
-
 end
 
 
@@ -136,7 +136,7 @@ global_interpretation DomainSepInv_2?: DomainSepInv_2
 proof goal_cases
   interpret Arch .
   case 1 show ?case
-    by (unfold_locales; fact DomainSepInv_assms)
+    by (unfold_locales; (fact DomainSepInv_assms | solves \<open>wp only: DomainSepInv_assms; simp\<close>)?)
 qed
 
 end
--- a/proof/access-control/ARM/ArchInterrupt_AC.thy
+++ b/proof/access-control/ARM/ArchInterrupt_AC.thy
@@ -21,8 +21,7 @@ definition arch_authorised_irq_ctl_inv ::
        (pasSubject aag, Control, pasIRQAbs aag irq) \<in> pasPolicy aag"
 
 lemma arch_invoke_irq_control_pas_refined[Interrupt_AC_assms]:
-  "\<lbrace>pas_refined aag and pspace_aligned and valid_vspace_objs and valid_arch_state
-                    and valid_mdb and arch_irq_control_inv_valid irq_ctl_inv
+  "\<lbrace>pas_refined aag and valid_mdb and arch_irq_control_inv_valid irq_ctl_inv
                     and K (arch_authorised_irq_ctl_inv aag irq_ctl_inv)\<rbrace>
    arch_invoke_irq_control irq_ctl_inv
    \<lbrace>\<lambda>_. pas_refined aag\<rbrace>"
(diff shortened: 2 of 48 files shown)


proof/access-control/ARM/ArchInterrupt_AC.thy
```diff
@@ -75,7 +74,7 @@ global_interpretation Interrupt_AC_1?: Interrupt_AC_1 "arch_authorised_irq_ctl_i
 proof goal_cases
   interpret Arch .
   case 1 show ?case
-    by (unfold_locales; (fact Interrupt_AC_assms)?)
+    by (unfold_locales; fact Interrupt_AC_assms)
```
lsf37, line 77: We should leave these in -- there are useful for debugging when the statement fails, because it will print a goal with the ones that could not be solved, whereas the non-? variant fails and only prints the original goal

proof/access-control/ARM/ArchSyscall_AC.thy
```diff
@@ -179,7 +179,7 @@ global_interpretation Syscall_AC_1?: Syscall_AC_1
 proof goal_cases
   interpret Arch .
   case 1 show ?case
-    by (unfold_locales; (fact Syscall_AC_assms)?)
+    by (unfold_locales; wpsimp wp: Syscall_AC_assms)
```
lsf37, line 182: Do we need wpsimp for this one? I think we discussed this before, and I'm starting to lean towards keeping interfaces more strictly. The point of the `Syscall_AC_assms` set is that it should collect the interface statements -- so if there is still proof work necessary at this point, then something is not quite right.

proof/access-control/RISCV64/ArchFinalise_AC.thy
```diff
@@ -229,10 +227,11 @@ lemma arch_finalise_cap_respects[wp]:
...
 
-crunch arch_post_cap_deletion
-  for pspace_aligned[Finalise_AC_assms, wp]: "\<lambda>s :: det_ext state. pspace_aligned s"
-  and valid_vspace_objs[Finalise_AC_assms, wp]: "\<lambda>s :: det_ext state. valid_vspace_objs s"
-  and valid_arch_state[Finalise_AC_assms, wp]: "\<lambda>s :: det_ext state. valid_arch_state s"
+declare prepare_thread_delete_st_tcb_at_halted[Finalise_AC_assms]
+declare finalise_cap_valid_list[Finalise_AC_assms]
+declare arch_finalise_cap_pas_refined[Finalise_AC_assms]
+declare prepare_thread_delete_pas_refined[Finalise_AC_assms]
+declare finalise_cap_replaceable[Finalise_AC_assms]
```
lsf37, line 234: This is a bit awkward (not that the previous crunch was great either). Do we have statements of these somewhere? Because if yes, the declaration should be there so that we know at the statement/proof that they are interface lemmas. If they come out of different crunches now, we should move the corresponding declarations to those crunches.

proof/access-control/RISCV64/ArchSyscall_AC.thy
```diff
@@ -168,7 +169,7 @@ global_interpretation Syscall_AC_1?: Syscall_AC_1
 proof goal_cases
   interpret Arch .
   case 1 show ?case
-    by (unfold_locales; (fact Syscall_AC_assms | wp init_arch_objects_inv))
+    by (unfold_locales; wpsimp wp: Syscall_AC_assms)
```
lsf37, line 172: `wpsimp` for interface check

proof/infoflow/FinalCaps.thy
```diff
@@ -317,7 +317,7 @@ lemma caps_ref_either_an_object_or_irq':
 
 locale FinalCaps_1 =
   fixes aag :: "'a subject_label PAS"
-  (* FIXME IF: precludes X64 *)
+  (* FIXME IF: precludes x64 *)
```
lsf37, line 320: X64 was correct

lsf37 (approved):
Nice! I do have some nitpicks in the comments, but the substance is great. Less cruft.

proof/infoflow/FinalCaps.thy
ryybrr: Was basing these off Raf's comments, should I flip those too?

lsf37: not sure what those referred to -- X64 is the L4V_ARCH which this comment was referring to, x64 is the processor architecture in normal text.

proof/access-control/RISCV64/ArchFinalise_AC.thy
corlewis: There's already these sorts of declares in quite a few arch files. My guess is that it is generally from when one architecture proves the lemma earlier, before the relevant locale and named theorem exists, which means that the crunch in this file doesn't do anything.

Alternatively, it could be a sign that the interface is too strong and that some of the lemmas could be removed from the locale, although we'd have to be careful to check all of the architectures to try that. In this case I'm a bit curious about whether the lemmas about generic functions could be removed, although that might not be worth exploring in this PR.

proof/access-control/ARM/ArchDomainSepInv.thy
```diff
@@ -120,23 +129,14 @@ lemma arch_invoke_irq_control_domain_sep_inv[DomainSepInv_assms]:
...
-
 end
 
 
 global_interpretation DomainSepInv_2?: DomainSepInv_2
 proof goal_cases
   interpret Arch .
   case 1 show ?case
-    by (unfold_locales; fact DomainSepInv_assms)
+    by (unfold_locales; wpsimp wp: DomainSepInv_assms)
```
corlewis, line 139: Another `wpsimp` here that should probably stay as `fact`.

proof/access-control/CNode_AC.thy
```diff
@@ -1178,7 +1165,7 @@ lemma cap_swap_pas_refined[wp]:
      by (erule subsetD;
          force simp: is_transferable_weak_derived intro!: sbta_cdt_transferable auth_graph_map_memI)+
          apply (blast intro: state_bits_to_policy.intros auth_graph_map_memI)
-   by fastforce+
+   done
```
corlewis, line 1168: Not directly related to this PR but I think that the indentation for the blast is off, probably due to the auto-indenter getting confused by the subgoal.

proof/access-control/RISCV64/ArchCNode_AC.thy
```diff
@@ -56,29 +56,28 @@ lemma sata_update2[CNode_AC_assms]:
...
                 split: if_split_asm)
 
+lemma vs_lookup_table_eqI':
+    "\<lbrakk> asid_table s' (asid_high_bits_of asid) = asid_table s (asid_high_bits_of asid);
+       \<forall>pool_ptr. asid_table s' (asid_high_bits_of asid) = Some pool_ptr
+                  \<longrightarrow> bot_level \<le> max_pt_level
+                  \<longrightarrow> vspace_for_pool pool_ptr asid (asid_pools_of s') =
+                      vspace_for_pool pool_ptr asid (asid_pools_of s);
+       bot_level < max_pt_level \<longrightarrow> pts_of s' = pts_of s \<rbrakk>
+   \<Longrightarrow> vs_lookup_table bot_level asid vref s' = vs_lookup_table bot_level asid vref s"
```
corlewis, line 66: Indentation doesn't seem quite right here, I think the assumptions are indented too far.

proof/access-control/RISCV64/ArchFinalise_AC.thy
```diff
@@ -241,7 +240,7 @@ global_interpretation Finalise_AC_1?: Finalise_AC_1
 proof goal_cases
   interpret Arch .
   case 1 show ?case
-    by (unfold_locales; (fact Finalise_AC_assms | wp finalise_cap_replaceable))
+    by (unfold_locales; wpsimp wp: Finalise_AC_assms)
```
corlewis, line 243: `wpsimp`

proof/infoflow/ADT_IF.thy
```diff
@@ -1091,23 +1086,54 @@ lemma handle_preemption_if_invs:
...
+  apply (wpsimp simp: handle_preemption_if_def wp: handle_kernel_interrupt_domain_sep_inv)
+   apply (rule_tac Q'="\<lambda>rv s. domain_sep_inv  irqs st s \<and> (rv \<noteq> None \<longrightarrow> the rv \<notin> non_kernel_IRQs)"
+                in hoare_strengthen_post)
+    apply (wpsimp wp: getActiveIRQ_rv_None)+
+  done
+
+lemma handle_kernel_interrupt_pas_refined:
+  "\<lbrace>pas_refined aag and K (irq \<notin> non_kernel_IRQs)\<rbrace>
+    handle_interrupt irq
+   \<lbrace>\<lambda>_. pas_refined aag\<rbrace>"
```
corlewis, line 1112:
```suggestion
  "\<lbrace>pas_refined aag and K (irq \<notin> non_kernel_IRQs)\<rbrace>
   handle_interrupt irq
   \<lbrace>\<lambda>_. pas_refined aag\<rbrace>"
```

proof/infoflow/ADT_IF.thy
```diff
@@ -1379,11 +1405,17 @@ lemma check_active_irq_if_wp:
...
+context ADT_IF_1 begin
+
 lemma handle_preemption_if_only_timer_irq_inv[wp]:
   "handle_preemption_if tc \<lbrace>only_timer_irq_inv irq st\<rbrace>"
   by (wp only_timer_irq_inv_pres handle_preemption_if_irq_masks handle_preemption_if_domain_sep_inv
       | simp | blast)+
 
+end
+
+
```
corlewis, line 1418: Very minor but you added these extra newlines a couple of times around context blocks like this. Was that intentional? I don't think we usually do it, unless the block is its own section in some way.

corlewis (approved):
Thanks for doing this, there's some very nice cleanup here!

proof/access-control/RISCV64/ArchFinalise_AC.thy
lsf37: It's alright, let's leave it.

proof/access-control/ARM/ArchDomainSepInv.thy
lsf37: Now replaced. I think the new `(solves ‹wp only: ..; simp›)?` are a good compromise between preventing too much magic and still allowing straightforward weakening.

lsf37 (approved):
If you fix up the white space bit Corey commented on (unless it is an actual section), then I think this is ready to merge.

proof/infoflow/ADT_IF.thy
ryybrr: The current security proof style is to treat context blocks as sections, so will keep as is for now

PR #878, opened 2025-03-31
remove obsolete .licenseignore file

lsf37:
No longer used by any of the license checking tools.


lsf37:
(have cancelled the proof checks, no need to heat the planet)

lsf37:
It looks like there is still a bug with the thylint check when there are only removed files in a PR. Unrelated to the content here, though.

lsf37:
> check if merge commit is real or github phantom

GitHub phantom

PR #879, opened 2025-03-31
docs: update python setup instructions

lsf37:
Update instructions for new python versions.


Xaphiosis (approved):
Documentation! Thank you

corlewis (approved):
That's some fancy markdown! Is it worth mentioning that there's alternative approaches to using a venv or should we just keep it simple?

lsf37:
> Is it worth mentioning that there's alternative approaches to using a venv or should we just keep it simple?

I went with the approach on the main seL4 doc site. We could mention `--break-system-packages`, but I don't really want people come complaining if things then actually do break :-)

corlewis:
> > Is it worth mentioning that there's alternative approaches to using a venv or should we just keep it simple?
> 
> I went with the approach on the main seL4 doc site. We could mention `--break-system-packages`, but I don't really want people come complaining if things then actually do break :-)

That's reasonable.

PR #880, opened 2025-04-01
Update other arches for det_ext changes

corlewis:
This extends the changes made in https://github.com/seL4/l4v/pull/824 for the functional correctness proofs for all architectures.

### DIFF
--- a/proof/invariant-abstract/RISCV64/ArchKHeap_AI.thy
+++ b/proof/invariant-abstract/RISCV64/ArchKHeap_AI.thy
@@ -718,7 +718,7 @@ lemma pspace_respects_region_cong[cong]:
 
 definition "obj_is_device tp dev \<equiv>
   case tp of Untyped \<Rightarrow> dev
-    | _ \<Rightarrow>(case (default_object tp dev 0) of (ArchObj (DataPage dev _)) \<Rightarrow> dev
+    | _ \<Rightarrow>(case default_object tp dev 0 0 of (ArchObj (DataPage dev _)) \<Rightarrow> dev
           | _ \<Rightarrow> False)"
 
 lemma cap_is_device_obj_is_device[simp]:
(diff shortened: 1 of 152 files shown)


proof/refine/ARM_HYP/Retype_R.thy
```diff
@@ -4154,21 +4078,21 @@ lemma createNewCaps_idle'[wp]:
...
-                               vspace_bits_defs objBits_def
-                               createObjects_def tcb_cte_cases_neqs
-                   | intro conjI impI
-                   | clarsimp simp: curDomain_def)+
+        apply (rename_tac apiobject_type)
+        apply (case_tac apiobject_type, simp_all split del: if_split)[1]
+            apply wpsimp
+           (* The following step does not use wpsimp to avoid clarsimp_no_cond, which for some reason
+              leads to a failed proof state. If this could be fixed then the inclusion of
+              classic_wp_pre could also be removed. *)
```
Xaphiosis, line 4086: uh, do we need an issue for this?

proof/invariant-abstract/RISCV64/ArchKHeap_AI.thy
```diff
@@ -761,6 +761,30 @@ lemma state_hyp_refs_of_tcb_state_update:
   apply (clarsimp simp add: state_hyp_refs_of_def obj_at_def split: option.splits)
   done
 
+lemma state_hyp_refs_of_tcb_domain_update:
+  "kheap s t = Some (TCB tcb) \<Longrightarrow>
+   state_hyp_refs_of (s\<lparr>kheap := (kheap s)(t \<mapsto> TCB (tcb\<lparr>tcb_domain := d\<rparr>))\<rparr>)
+     = state_hyp_refs_of s"
```
Xaphiosis, line 767: where is this style coming from? the point of right-hand operator wrapping is so we don't need to add extra indentation, otherwise we'd just stick it on the previous line; for this and the next two lemmas the `=` should go under `state`

Xaphiosis (approved):
I read through the non-squash commits. Didn't find much, and this is a massive diff, but the stuff I did find I'm not sure whether it came from manual reapplication of failed automation, or what. In any case I've also seen that strange left-operator wrap indent style on your other PR, so I suggest we don't keep propagating that.

Xaphiosis:
Incidentally, any important reason why the tests are all failed?

corlewis:
> Incidentally, any important reason why the tests are all failed?

A bunch of non-functional correctness proofs need to be updated as well, that's coming in the next PR.

proof/refine/ARM_HYP/Retype_R.thy
corlewis: Maybe but it's not new and I'm not sure what I would put in that issue beyond what's in this comment. Happy to do that if you want.

proof/invariant-abstract/RISCV64/ArchKHeap_AI.thy
corlewis: It looks like I was copying the existing lemmas that are just before these ones. I agree that this is obviously the wrong indentation but I don't want to fix those other lemmas in all architectures in this PR, so do we prefer correct indentation or local consistency?

Xaphiosis: Maybe note the files down somewhere and then handle in another PR then.

lsf37 (approved):
Looking good from my side.

proof/invariant-abstract/RISCV64/ArchKHeap_AI.thy
corlewis: I ended up just fixing up the style for both my new lemmas and the ones I was copying. There's plenty of other horribly styled lemmas around here but I'll leave fixing them for next time they're touched.

PR #881, opened 2025-04-01
`ccorres` rules for the reply functions

michaelmcinerney:
Test with https://github.com/seL4/seL4/pull/1440


lsf37:
To get rid of the link check failure, cherry-picking commit 9a28f5007ee6b6a4d01b8854c1ff1e82951843b8 should work.

proof/crefine/RISCV64/Machine_C.thy
```diff
@@ -457,6 +457,9 @@ lemma getTimerPrecision_spec:
   "\<forall>s. \<Gamma>\<turnstile> {s} Call getTimerPrecision_'proc \<lbrace>\<acute>ret__unsigned_longlong = timerPrecision\<rbrace>"
   apply vcg
   by (simp add: timerPrecision_def)
+lemma wrap_config_set_spec:
```
lsf37, line 460: Empty line missing between those two.

proof/refine/RISCV64/Finalise_R.thy
```diff
@@ -2430,16 +2436,14 @@ lemma replyPop_valid_objs'[wp]:
...
-                apply (wpsimp wp: hoare_vcg_all_lift hoare_vcg_imp_lift hoare_vcg_if_lift2 )+
-  apply (simp add: isHead_to_head)
-  apply (drule_tac k=x in ko_at_valid_objs'; clarsimp simp: valid_obj'_def
-                 valid_sched_context'_def valid_sched_context_size'_def objBits_simps refillSize_def)
-  apply (drule_tac k=ko in ko_at_valid_objs'; clarsimp simp: valid_obj'_def
-                 valid_sched_context'_def valid_sched_context_size'_def objBits_simps refillSize_def)
-  apply (clarsimp simp: valid_reply'_def)
+                 apply (wpsimp wp: replyNext_update_valid_objs' hoare_vcg_if_lift2
+                                   hoare_vcg_all_lift
+                        | wp (once) hoare_drop_imp)+
```
lsf37, line 2443: This is much nicer!

proof/refine/RISCV64/Finalise_R.thy
```diff
@@ -2453,18 +2457,22 @@ lemma replyRemove_valid_objs'[wp]:
...
   unfolding replyPop_def
   supply if_split[split del]
-  apply (wpsimp wp: hoare_vcg_imp_lift)
-                 apply (wpsimp wp: updateReply_valid_replies'_bound hoare_vcg_imp_lift
-                                   hoare_vcg_all_lift hoare_vcg_ex_lift hoare_vcg_if_lift)+
-  apply (rename_tac prevReplyPtr)
-  apply (drule_tac rptr=prevReplyPtr in valid_replies'D)
+  apply (wpsimp wp: hoare_vcg_imp_lift'
+         | wpsimp wp: updateReply_valid_replies'_bound hoare_vcg_all_lift
+                      hoare_vcg_ex_lift hoare_vcg_if_lift2 gts_wp')+
```
lsf37, line 2472: I'm not so convinced about the `(wpsimp .. | wpsimp ..)+` pattern. The old one didn't look pleasant, but it has clearer failure modes when things break. In the big `(.. | ..)+` patterns you get more automation, but also more pain on breakage.

Is the `wpsimp wp: hoare_vcg_imp_lift` needed a second time now or was the change mostly cosmetic?

proof/crefine/RISCV64/IpcCancel_C.thy
```diff
@@ -3029,19 +3021,758 @@ lemma reply_unlink_ccorres:
...
+                                      \<and> weak_sch_act_wf (ksSchedulerAction s) s"
+                         in hoare_post_imp)
+             apply (clarsimp simp: valid_reply'_def)
+            apply (wpsimp wp: schedContextDonate_valid_objs')
+           apply (clarsimp simp: guard_is_UNIV_def)
+          apply (rule_tac Q'="\<lambda>_ s. reply_at' replyPtr s \<and> sc_at' (theHeadScPtr (replyNext reply)) s
+                                    \<and> valid_objs' s \<and> no_0_obj' s
+                                    \<and> pspace_aligned' s \<and> pspace_distinct' s \<and> pspace_bounded' s
+                                    \<and> sym_heap_sched_pointers s \<and> valid_sched_pointers s
+                                    \<and> weak_sch_act_wf (ksSchedulerAction s) s"
```
lsf37, line 3595: I'm guessing the instantiations are necessary -- would there be any potential benefit in directly assuming `invs'` at this point already? Mostly depends on what lemmas are available about the functions and how ugly they are, but sometimes it's surprisingly nice and short.

proof/crefine/RISCV64/Ipc_C.thy
```diff
@@ -5366,110 +5366,238 @@ lemma ctcb_relation_blockingIPCCanGrantD:
...
+                    apply ceqv
+                   apply (ctac add: schedContext_donate_ccorres)
+                  apply (wpsimp wp: updateReply_valid_objs')
+                 apply (clarsimp simp: guard_is_UNIV_def)
+                apply (rule_tac Q'="\<lambda>_ s. sc_at' scPtr s \<and> tcb_at' calleePtr s
+                                          \<and> reply_at' replyPtr s \<and> valid_objs' s \<and> no_0_obj' s
+                                          \<and> weak_sch_act_wf (ksSchedulerAction s) s
+                                          \<and> pspace_aligned' s \<and> pspace_distinct' s
+                                          \<and> pspace_canonical' s"
+                             in hoare_post_imp)
```
lsf37, line 5555: This pattern of instantiation in the `wp` chain at the end is happening a lot. We should only need that in exceptional circumstances, because all of these are stopping points in maintenance when things break. Is something going on in the goals here that is different? Maybe we can find another solution.

lsf37 (commented):
I realise this was still in draft, apologies for jumping in early with the comments. If we can resolve those instantiation chains, I think this is pretty much ready to go.

proof/crefine/RISCV64/IpcCancel_C.thy
michaelmcinerney: `invs'` isn't so nice here because we'll want to start arguing about non-zombie caps and other annoying things, unfortunately.

proof/crefine/RISCV64/Ipc_C.thy
michaelmcinerney: This is mostly due to the ridiculous repeated use of `updateReply` and `updateSchedContext`. We need `valid_objs'` right at the end of this function, and so to show that `valid_objs'` is preserved, we need to show that the updates preserve validity of objects.

One thing I could investigate is making further use of asserts in some Haskell definitions, which might reduce the need for `valid_objs'` in `ccorres` guards, so that these Hoare triples just do not appear.

If these functions were more nicely abstracted and used separate functions to handle the linked list of replies, then we could state nicer separate rules for those functions preserving `valid_objs`'...

proof/refine/RISCV64/Finalise_R.thy
michaelmcinerney: This was a very annoying proof. `hoare_vcg_imp_lift` is definitely needed more than once. I can get something more similar to the previous proof working. Here is another option:

```
  apply (wpsimp wp: hoare_vcg_imp_lift')
                apply (wpsimp wp: updateReply_valid_replies'_bound hoare_vcg_all_lift
                                  hoare_vcg_ex_lift hoare_vcg_if_lift2 gts_wp'
                                  hoare_drop_imp[where Q'="λ_ s. replyTCBs_of s replyPtr = Some tcbPtr"]
                                  hoare_vcg_imp_lift')+
```

The instantiation in `hoare_drop_imp` is needed there, sadly. 

Implications seem to be the worst thing in Hoare triple proofs. I wish I knew how to handle them better.

Xaphiosis:
Minor question on commit (message are generally good and helpful, thank you):
In `rt haskell+riscv refine+crefine: ccorres rules for reply`, the tag is kind of confusing. @lsf37 ?

proof/refine/RISCV64/Finalise_R.thy
Xaphiosis: I think the instantiated hoare_drop_imp is safer; you might be able to have a shorter instantiation by picking the function, but what you have is good.
Are you able to compress these two into a single wpsimp by picking the order of the instantiated hoare_drop_imp and hoare_vcg_imp_lift'? If you instantiated it precisely, it should apply first and only where you need it, and for other implications `hoare_vcg_imp_lift'` should trigger. In theory.

lsf37:
> Minor question on commit (message are generally good and helpful, thank you): In `rt haskell+riscv refine+crefine: ccorres rules for reply`, the tag is kind of confusing. @lsf37 ?

It is trying to say generic Haskell and RISCV refine+crefine. Maybe in this case we should just shorten in to `rt haskell+refine+crefine`

proof/crefine/RISCV64/Ipc_C.thy
Xaphiosis: Potenatially you could also explore using strengthen to lift predicates into valid_objs', depending on how many or how complex they are.

proof/crefine/RISCV64/IpcCancel_C.thy
```diff
@@ -3026,19 +3021,758 @@ lemma reply_unlink_ccorres:
...
+                                   \<and> pspace_aligned' s \<and> pspace_distinct' s
+                                   \<and> weak_sch_act_wf (ksSchedulerAction s) s"
+                      in hoare_post_imp)
+          apply (clarsimp simp: valid_reply'_def)
+         apply (wpsimp wp: updateReply_valid_objs')
+        apply (clarsimp simp: guard_is_UNIV_def)
+       apply (rule_tac Q'="\<lambda>_ s. valid_reply' reply s \<and> reply_at' replyPtr s \<and> valid_objs' s
+                                 \<and> no_0_obj' s \<and> pspace_aligned' s \<and> pspace_distinct' s
+                                 \<and> weak_sch_act_wf (ksSchedulerAction s) s"
+                    in hoare_post_imp)
```
Xaphiosis, line 3761: If this pattern of instantiating can't be avoided in a nice way, one proposal is to have it bound to something (would require an Isar wrapper on the lemma) so we see ?something_post three times instead of the whole block.

proof/crefine/RISCV64/IpcCancel_C.thy
```diff
@@ -3026,19 +3021,758 @@ lemma reply_unlink_ccorres:
...
+         (obj_at' (P :: sched_context \<Rightarrow> bool) scPtr and P') R hs
+         (updateSchedContext scPtr g) c"
+  apply (rule ccorres_from_vcg)
+  apply (rule allI)
+  apply (case_tac "obj_at' P scPtr \<sigma>")
+   apply (drule obj_at_ko_at', clarsimp)
+   apply (rule conseqPre, rule conseqPost)
+      apply assumption
+     apply clarsimp
+     apply (rule rev_bexI, rule updateSchedContext_eq)
```
Xaphiosis, line 3102: clarsimp should discharge the assumption, then we have clarsimp, three simps, and a clarsimp... I think you can safely put a `; clarsimp simp: obj_at_simps` at the end here and save some lines

Xaphiosis: (or clarsimp with a +)

michaelmcinerney: These were just moved. I meant to write a nicer PR description before I marked this as ready for review. But these `updateSchedContext_ccorres_*` lemmas were just moved, along with `schedContext_donate_ccorres` and `tcbReleaseRemove_ccorres`

proof/refine/RISCV64/Finalise_R.thy
michaelmcinerney: I really can't figure out a way to compress this into a single `wpsimp`. Every time I try something, I end up having to prove that `updateReply` preserves `~ ko_at' reply replyPtr s`, which isn't true.

proof/crefine/RISCV64/Ipc_C.thy
michaelmcinerney: I think I will try asserting `valid_objs'` in the Haskell, and see if I can improve this.

proof/crefine/RISCV64/IpcCancel_C.thy
Xaphiosis: It's still a pretty easy compression if you're interested while we're here.

Xaphiosis:
> > Minor question on commit (message are generally good and helpful, thank you): In `rt haskell+riscv refine+crefine: ccorres rules for reply`, the tag is kind of confusing. @lsf37 ?
> 
> It is trying to say generic Haskell and RISCV refine+crefine. Maybe in this case we should just shorten in to `rt haskell+refine+crefine`

Yeah I think we'd normally have two commits, one for spec changes and one for proof update. In this case, I'd probably go with your suggestion given the nature of the `rt` tag.

proof/refine/RISCV64/Finalise_R.thy
Xaphiosis: Hmm. That likely means that a lift is occurring before the drop you want happens. In that case, if you can't get around it, I'd go with the option you listed above.

Out of curiosity, if you try the single wpsimp without the hoare_drop_imp at all, do you end up in the same `~ ko_at' reply replyPtr s` state?

michaelmcinerney: I don't think that `wpsimp` without anything for implications does much at all here. I'm just checking the proofs again after adding in the assert, but I will check soon and report back if I was wrong about that.

Xaphiosis: I wanted to check whether your instantiated hoare_drop_imp is firing, or whether it's being ignored. No rush.

michaelmcinerney: A plain `wpsimp` doesn't do much at all. It will get me to the point where I would like to drop the antecedent that I instantiated in the `hoare_drop_imp`. If I do drop that, then I next come to an `updateReply` preserving `valid_replies'`. We have a rule for this, but it's a bit ugly:

```
lemma updateReply_valid_replies'_bound:
  "⦃λs. valid_replies' s
        ∧ (∀reply. ko_at' reply rptr s ⟶
                     (∃tptr. replyTCB (f reply) = Some tptr
                             ∧ st_tcb_at' ((=) (BlockedOnReply (Some rptr))) tptr s))⦄
   updateReply rptr f
   ⦃λ_. valid_replies'⦄"
```

So this then produces a fairly tricky postcondition involving the `ko_at'`. Just a little earlier in the function, we have another `updateReply`, which is why I was coming up against the `~ ko_at' reply replyPtr s` bit.

The first line of just `wpsimp wp: hoare_vcg_imp_lift'` seems to be doing a fair bit of work. Given that the `hoare_drop_imp` is instantiated, I'm not sure why removing that first line and having the `hoare_vcg_imp_lift'` in the larger `wpsimp` block doesn't work

lsf37: The two separate `wpsimp` (one single, one `+`) is fine. It's not the prettiest, but when it fails, it is at least somewhat clear what is going on.

michaelmcinerney:
I've pushed an update with the fix for the broken links, a commit that reduces the use of `valid_objs'` in CRefine, and another commit that removes the nasty `hoare_post_imp` chains that I had in the `ccorres` rules for the reply functions. The changes to get rid of `valid_objs'` were a bit larger than I was hoping for, but it should really make life easier in CRefine (apart from potentially having to add the assert to more new functions)

michaelmcinerney:
I think I missed a few updates for the new asserts, so I'll run the proofs again locally and push an updated version soon.

Xaphiosis (approved):
Looks like you were on the right track with the valid_objs' crossing, and the proofs look neater now.

lsf37:
```
 Installing seL4 python deps
    error: subprocess-exited-with-error
```

I'm coming to really dislike Python. Ok to ignore this here, it's some setup problem of the style checker.

lsf37 (approved):
Agreed, the valid_objs' idea was good. We have a bit of the post condition instantiation moving into Refine, but now concentrated to just once and also in a nicer form. I like it.

lsf37:
> ```
>  Installing seL4 python deps
>     error: subprocess-exited-with-error
> ```
> 
> I'm coming to really dislike Python. Ok to ignore this here, it's some setup problem of the style checker.

And it now works on the same OS and python version without anything having changed. Maybe it was a networking problem or something like that.

PR #882, opened 2025-04-02
update to Isabelle 2025

lsf37:
- pull in Word_Lib and Simpl from AFP, fix style issues
- update the rest

Mostly small changes in ML code, almost no proof impact this time.

One thing to note is that the name space for constants and free/bound variable names is completely separate now. That means, Isabelle will generate some names that clash with constant names. If people followed `rename_tac` discipline that would not have led to any changes, but it turns out we don't quite follow it everywhere, so there are some additional `rename_tac`s.

Test with: https://github.com/seL4/isabelle/pull/5

### DIFF
--- a/lib/Eisbach_Tools/Trace_Schematic_Insts.thy
+++ b/lib/Eisbach_Tools/Trace_Schematic_Insts.thy
@@ -325,7 +325,8 @@ fun detach_rule_result_annotations ctxt st =
 \<close>
 fun instantiate_terms ctxt bounds var_insts term =
   let
-    val vars = Variable.variant_frees ctxt (term :: map #2 var_insts) bounds
+    val ctxt' = fold Variable.declare_names  (term :: map #2 var_insts) ctxt
+    val vars = Variable.variant_names ctxt' bounds
     fun var_inst_beta term term' =
       (term, Term.betapplys (term', map Free vars))
     val var_insts' = map (uncurry var_inst_beta) var_insts
--- a/lib/Insulin.thy
+++ b/lib/Insulin.thy
@@ -55,7 +55,7 @@
 
 theory Insulin
 imports
-  Pure
+  Main
 keywords
   "desugar_term" "desugar_thm" "desugar_goal" :: diag
 begin
@@ -134,7 +134,8 @@ fun desugar_reconst ctxt (tr as XML.Elem ((tag, attrs), children))
           (* try to look up the const's info *)
           case Syntax.read_term ctxt name
                |> Thm.cterm_of ctxt
-               |> Proof_Display.pp_cterm (fn _ => Proof_Context.theory_of ctxt)
+               |> Thm.term_of
+               |> Syntax.pretty_term ctxt
                |> Pretty.string_of
                |> dropQuotes
                |> YXML.parse
--- a/lib/Monads/reader_option/Reader_Option_Monad.thy
+++ b/lib/Monads/reader_option/Reader_Option_Monad.thy
@@ -198,7 +198,7 @@ definition obind :: "('s,'a) lookup \<Rightarrow> ('a \<Rightarrow> ('s,'b) look
 
 (* Enable "do { .. }" syntax *)
 adhoc_overloading
-  Monad_Syntax.bind obind
+  Monad_Syntax.bind \<rightleftharpoons> obind
 
 definition ofail :: "('s, 'a) lookup" where
   "ofail = K None"
--- a/lib/Word_Lib/Aligned.thy
+++ b/lib/Word_Lib/Aligned.thy
@@ -2,6 +2,7 @@
  * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
  *
  * SPDX-License-Identifier: BSD-2-Clause
+Proofs tidied by LCP, 2024-09
  *)
 
 section "Word Alignment"
@@ -11,6 +12,7 @@ theory Aligned
     "HOL-Library.Word"
     More_Word
     Bit_Shifts_Infix_Syntax
+
 begin
 
 context
@@ -54,11 +56,7 @@ lemma is_aligned_at_0 [simp]:
 
 lemma is_aligned_beyond_length:
   \<open>is_aligned w n \<longleftrightarrow> w = 0\<close> if \<open>LENGTH('a) \<le> n\<close> for w :: \<open>'a::len word\<close>
-  using that
-  apply (simp add: is_aligned_iff_udvd)
-  apply transfer
-  apply auto
-  done
+  using that by (simp add: is_aligned_iff_take_bit_eq_0 take_bit_word_beyond_length_eq)
 
 lemma is_alignedI [intro?]:
   \<open>is_aligned x n\<close> if \<open>x = 2 ^ n * k\<close> for x :: \<open>'a::len word\<close>
@@ -235,12 +233,7 @@ qed
 corollary aligned_sub_aligned:
   "\<lbrakk>is_aligned (x::'a::len word) n; is_aligned y m; m \<le> n\<rbrakk>
    \<Longrightarrow> is_aligned (x - y) m"
-  apply (simp del: add_uminus_conv_diff add:diff_conv_add_uminus)
-  apply (erule aligned_add_aligned, simp_all)
-  apply (erule is_alignedE)
-  apply (rule_tac k="- of_nat q" in is_alignedI)
-  apply simp
-  done
+  by (metis (no_types, lifting) diff_zero is_aligned_mask is_aligned_weaken mask_eqs(4))
 
 lemma is_aligned_shift:
   fixes k::"'a::len word"
@@ -320,34 +313,14 @@ proof -
   next
     case False
     then have sne: "0 < sz" ..
-
     show ?thesis
     proof -
       have uq: "unat (of_nat q ::'a::len word) = q"
-        apply (subst unat_of_nat)
-        apply (rule mod_less)
-        apply (rule order_less_trans [OF qv])
-        apply (rule power_strict_increasing [OF diff_less [OF sne]])
-         apply (simp_all)
-        done
-
+        by (metis diff_less le_unat_uoi len_gt_0 order_less_imp_le qv sne unat_power_lower)
       have uptr: "unat ptr = 2 ^ sz * q"
-        apply (subst ptrq)
-        apply (subst iffD1 [OF unat_mult_lem])
-         apply (subst unat_power_lower [OF szv])
-         apply (subst uq)
-         apply (rule nat_less_power_trans [OF qv order_less_imp_le [OF szv]])
-        apply (subst uq)
-        apply (subst unat_power_lower [OF szv])
-        apply simp
-        done
-
+        by (simp add: ptrq qv unat_mult_power_lem)
       show "unat ptr + unat off < 2 ^ LENGTH('a)" using szv
-        apply (subst uptr)
-        apply (subst mult.commute, rule nat_add_offset_less [OF _ qv])
-        apply (rule order_less_le_trans [OF unat_mono [OF off] order_eq_refl])
-         apply simp_all
-        done
+        by (metis le_add_diff_inverse2 less_imp_le mult.commute nat_add_offset_less off qv unat_less_power uptr)
     qed
   qed
 qed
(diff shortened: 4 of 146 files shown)


lib/Word_Lib/Enumeration_Word.thy
```diff
@@ -2,18 +2,15 @@
  * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
  *
  * SPDX-License-Identifier: BSD-2-Clause
+Proofs tidied by LCP, 2024-09
```
lsf37, line 5: I thought I had moved these to the AFP changelog. Looks like I still need to apply that patch.

proof/crefine/RISCV64/IpcCancel_C.thy
```diff
@@ -7,7 +7,7 @@
  *)
 
 theory IpcCancel_C
-imports SyscallArgs_C
+imports SyscallArgs_C "HOL-Library.Divides"
```
lsf37, line 10: I forgot what this was for -- there was some lemma that was not available any more, but I think we might be able to eliminate the import again.

proof/crefine/ARM/Detype_C.thy
```diff
@@ -1669,8 +1669,7 @@ proof -
     apply (case_tac "pageBits \<le> bits")
      apply (simp add: objBitsKO_def projectKOs  split: kernel_object.splits)
      apply clarsimp
-     apply (rule aligned_range_offset_mem
-       [where 'a=32, folded word_bits_def, simplified, OF _ _ al _ wb])
+     apply (rule aligned_range_offset_mem [simplified, OF _ _ al])
```
Xaphiosis, line 1672: space before `[` inconsistent with change below, suggest remove

proof/crefine/ARM_HYP/ArchMove_C.thy
```diff
@@ -326,17 +326,9 @@ lemma ucast_ucast_mask_pageBits_shift:
...
   "unat (ucast (p && mask pageBits >> 2) :: 10 word) = unat ((p::word32) && mask pageBits >> 2)"
-  apply (simp only: unat_ucast)
-  apply (rule Divides.mod_less, simp)
-  apply (rule unat_less_power)
-   apply (simp add: word_bits_def)
-  apply (rule shiftr_less_t2n)
-  apply (rule order_le_less_trans [OF word_and_le1])
-  apply (simp add: pageBits_def mask_def)
-  done
+  by (metis ucast_ucast_mask_pageBits_shift unat_ucast_10_32)
```
Xaphiosis, line 331: in AARCH64 this goes via:
```
  apply (rule unat_ucast_mask_shift)
  apply (simp add: pageBits_def)
```
how come the difference?

lsf37: AARCH64 doesn't have the helper lemma, so it has a different proof. Could probably unify towards AARCH64, but the statements are different (different types), so no point, really.

tools/c-parser/standalone-parser/Makefile
```diff
@@ -53,8 +53,11 @@ include $(STP_PFX)/../Makefile
 STP_CLEAN_TARGETS := $(STPARSERS) $(TOKENIZERS) $(STP_PFX)/c-parser.o $(STP_PFX)/table.ML
 
 $(STP_PFX)/table.ML: $(ISABELLE_HOME)/src/Pure/General/table.ML $(STP_PFX)/Makefile
-	sed -e '/\(\* cache \*\)/,/final.declarations/d' < $< | \
-	sed -e "s/^  val unsynchronized_cache:.*a/  (* removed unsynchronized_cache *)/" > $@
+	sed -e '/\(\* cache \*\)/,/final declarations of this structure/d' < $< | \
+  sed -e "s/^  type 'a cache_ops.*int./  (* removed type 'a cache_ops *)/" | \
+  sed -e "s/^  val unsynchronized_cache:.*cache_ops/  (* removed unsynchronized_cache *)/" | \
+  sed -e "s/^  val apply_unsynchronized_cache.*a/  (* removed apply_unsynchronized_cache *)/" > $@
+
```
Xaphiosis, line 60: I'm trying to understand this and can't... where is the first sed getting its inputs?
Also, do you want the extra newline after?

lib/Monads/reader_option/Reader_Option_Monad.thy
```diff
@@ -198,7 +198,7 @@ definition obind :: "('s,'a) lookup \<Rightarrow> ('a \<Rightarrow> ('s,'b) look
 
 (* Enable "do { .. }" syntax *)
 adhoc_overloading
-  Monad_Syntax.bind obind
+  Monad_Syntax.bind \<rightleftharpoons> obind
```
Xaphiosis, line 201: This change seems to indicate that you need to spell out the notation both here and at the definition point of obind, right? I'm confused by this.

lib/Eisbach_Tools/Trace_Schematic_Insts.thy
```diff
@@ -325,7 +325,8 @@ fun detach_rule_result_annotations ctxt st =
 \<close>
 fun instantiate_terms ctxt bounds var_insts term =
   let
-    val vars = Variable.variant_frees ctxt (term :: map #2 var_insts) bounds
+    val ctxt' = fold Variable.declare_names  (term :: map #2 var_insts) ctxt
+    val vars = Variable.variant_names ctxt' bounds
```
Xaphiosis, line 329: so much more convenient now :/

lib/Insulin.thy
```diff
@@ -55,7 +55,7 @@
 
 theory Insulin
 imports
-  Pure
+  Main
```
Xaphiosis, line 58: Japheth probably put the Pure in with some kind of intent. Naturally, we don't do anything on top of Pure anyway, but any reason why it has to be Main now?

lib/Word_Lib/Bitwise.thy
```diff
@@ -148,7 +148,7 @@ proof (cases "n < size w")
   case True
   then show ?thesis
     by (simp add: bl_sshiftr takefill_last_def word_size takefill_alt
-           rev_take last_rev drop_nonempty_def)    
+           rev_take last_rev drop_nonempty_def)
```
Xaphiosis, line 151: yuck, yuck, yuck... is there any way you can force upstream to get their trailing whitespace sorted out? next we're going to see the return of tabs

lib/Word_Lib/Word_Syntax.thy
```diff
@@ -18,22 +18,22 @@ context
 begin
 
 abbreviation
-  wordNOT  :: "'a::len word \<Rightarrow> 'a word"      ("~~ _" [70] 71)
+  wordNOT  :: "'a::len word \<Rightarrow> 'a word"      (\<open>(\<open>open_block notation=\<open>prefix ~~\<close>\<close>~~ _)\<close> [70] 71)
```
Xaphiosis, line 21: what are these strange complicated notation blocks doing? is this even documented?

lib/Word_Lib/Strict_part_mono.thy
```diff
@@ -35,22 +35,27 @@ lemma strict_part_mono_reverseE:
...
+      by simp
+    moreover have \<open>2 ^ n < (2::nat) ^ Suc n\<close>
+      by simp
+    with Suc.prems have \<open>word_of_nat (2 ^ n) < (word_of_nat (2 ^ Suc n) :: 'a word)\<close>
+      by (simp only: of_nat_word_less_iff take_bit_of_exp) simp
+    ultimately show ?case
+      by (subst strict_part_mono_by_steps) simp
+  qed
+  then show ?thesis
+    by simp
```
Xaphiosis, line 58: I am not seeing how a proof like this is an improvement over the one it replaces. Am I missing something? (I know it's upstream, I just don't understand what they're doing and wonder if you do)

lib/Word_Lib/Signed_Division_Word.thy
```diff
@@ -170,28 +177,19 @@ lemmas word_sdiv_0 = word_sdiv_div0
...
-   apply simp_all
-  apply transfer
-  apply simp
-  apply (rule order_trans)
-   defer
-   apply assumption
-  apply simp
-  apply (metis abs_le_iff add.inverse_inverse le_cases le_minus_iff not_le signed_take_bit_int_eq_self_iff signed_take_bit_minus)
-  done
+  by (smt (verit, ccfv_SIG) atLeastAtMost_iff sdiv_int_range sint_ge sint_lt wsst_TYs(3))
```
Xaphiosis, line 180: do you have any feelings on non-smt proofs being replaced by smt?

lib/Word_Lib/Aligned.thy
```diff
@@ -2,7 +2,6 @@
  * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
  *
  * SPDX-License-Identifier: BSD-2-Clause
-Proofs tidied by LCP, 2024-09
```
Xaphiosis: this is added in other files as well, but you only reverted this one, any reason?

lib/Word_Lib/Many_More.thy
```diff
@@ -10,13 +10,11 @@ theory Many_More
     "HOL-Library.Word"
     More_Word
     Even_More_List
+
```
Xaphiosis, line 13: why'd they put this line here? I really don't get what upstream is doing

lib/Word_Lib/Bitwise.thy
```diff
@@ -10,6 +10,7 @@ theory Bitwise
     More_Arithmetic
     Reversed_Bit_Lists
     Bit_Shifts_Infix_Syntax
+
```
Xaphiosis, line 13: again, who is adding these lines? what for? is there some new standard?

lib/Word_Lib/Rsplit.thy
```diff
@@ -2,16 +2,262 @@
...
 
 (* Author: Jeremy Dawson and Gerwin Klein, NICTA *)
 
 section \<open>Splitting words into lists\<close>
 
 theory Rsplit
-  imports "HOL-Library.Word" More_Word Bits_Int
+  imports More_Word Bit_Shifts_Infix_Syntax
+
+
```
Xaphiosis, line 15: oh this is even better, two lines for no reason whatsoever

lib/Word_Lib/More_Word_Operations.thy
```diff
@@ -923,47 +845,57 @@ lemma aligned_mask_ranges_disjoint:
...
+  shows "- sint (of_nat (word_clz (w :: 'a :: len word)) :: 'a signed word) \<le> int (LENGTH('a))"
+proof -
+  have "\<forall>w. \<not> 2 ^ (LENGTH('a) - 1) < - sint (w::'a signed word)"
+    by (smt (verit, ccfv_SIG) len_signed sint_ge)
+  have "word_clz w < 2 ^ LENGTH('a)"
+    by (metis pow_mono_leq_imp_lt word_clz_max wsst_TYs(3))
+  then show ?thesis
+    using assms small_powers_of_2 len_signed negative_zle order_le_less_trans
+    by (smt (verit, best) int_eq_sint word_clz_max wsst_TYs(3))
+qed
```
Xaphiosis, line 898: oh yes, this is so much clearer and nicer :/

Xaphiosis (approved):
I'm happy with your changes, but am rather annoyed that people upstream arbitrarily change the AFP entry without permission or review, add random lines without consideration for standards or the way existing files look, leave trailing whitespace, add themselves in random headers in ways that again completely do not fit any kind of existing scheme, and even leave inline attribution markers in the middle of proofs. Open source does not mean you can just go into my repo and merge stuff. I am not happy this happened, even though there are visible improvements (and a few pointless "we don't like apply style" conversions). If you can push back on the pointless stuff, I'll support it, even though I probably don't have much ammo.

lib/Word_Lib/Enumeration_Word.thy
lsf37: now applied.

lib/Monads/reader_option/Reader_Option_Monad.thy
lsf37: it's only the syntax of `adhoc_overloading` that changed. It wants a `⇌` now, no other changes than that really.

lib/Insulin.thy
lsf37: Importing Pure is generally a bad idea, everything should be on HOL.Main, otherwise you can get weird context merge effects. Japheth wanted to express more generality, but we're not ever going to apply this to ZF or the lambda cube..

lib/Word_Lib/Strict_part_mono.thy
lsf37: In my view this change was entirely pointless. Even more so since this file has now been dropped upstream.

lib/Word_Lib/Aligned.thy
lsf37: I think I started with this file and then later noticed that it's everywhere.

tools/c-parser/standalone-parser/Makefile
lsf37: The input is `< $< ` (Makefile `$< `). This part is mostly black sed magic with Makefile fairy dust sprinkled on. It removes some bits from the Isabelle implementation of `table.ML` that would draw in more dependencies. The really gross one is the `sed` in the `mk_release` which operates on this Makefile and removes the `sed` here..

This time I did actually want the additional line, the next set of targets are a different kind, so I wanted a new section.

lib/Word_Lib/Bitwise.thy
lsf37: I can't force it, of course, but it is reverted upstream (and complained about)

lib/Word_Lib/Word_Syntax.thy
lsf37: Yes, it's document in NEWS (and the reference manual, I presume, but I have not checked). I don't actually fully understand what this is doing yet, but I haven't really looked into it. Something with having blocks that do not affect pretty printing.

lib/Word_Lib/Signed_Division_Word.thy
lsf37: `smt (verit)` is fairly stable (compared to `smt (z3)`, which may break more easily) and allowed in the core Isabelle distribution, so I don't think it's a problem. It is proof-checked by Isabelle, so there is no soundness difference to a normal Isabelle proof, and the one it replaces was equally unreadable, so this instance really is fine.

PR #883, opened 2025-04-03
Update other top level proofs for det_ext changes

corlewis:
This follows on from https://github.com/seL4/l4v/pull/824 and https://github.com/seL4/l4v/pull/880. All proofs except for Infoflow should now be working.

### DIFF
--- a/proof/access-control/RISCV64/ArchAccess.thy
+++ b/proof/access-control/RISCV64/ArchAccess.thy
@@ -74,6 +74,11 @@ lemmas state_vrefs_upd =
   revokable_update.state_vrefs
   machine_state_update.state_vrefs
   more_update.state_vrefs
+  scheduler_action_update.state_vrefs
+  domain_index_update.state_vrefs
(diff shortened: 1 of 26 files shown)


lsf37 (approved):
Looking nice and shiny from here. I hadn't realised we had so much `ekheap` stuff in DRefine. Good to see that it's gone now.

corlewis:
> (The Bisim session on X64 is still complaining)

Ah, I had assumed from it not being arch split that it was generic. At least it was an easy fix.

Now that I'm looking more closely, why do we build it for ARM, X64 and RISCV64 but not ARM_HYP or AARCH64?

lsf37:
> Now that I'm looking more closely, why do we build it for ARM, X64 and RISCV64 but not ARM_HYP or AARCH64?

For AARCH64 we just haven't gotten to it yet, but it is likely that HYP has some wrinkles and is not entirely straightforward.

Xaphiosis (approved):
Thumbs up from my side

PR #884, opened 2025-04-07
Update InfoFlow for det_ext changes

corlewis:
This updates InfoFlow for the changes made in https://github.com/seL4/l4v/pull/824. All proofs should now be working.


lsf37 (approved):
Looking good, no complaints. We could move the FIXME move while we are here, but also fine to leave it for later.

Xaphiosis (approved):
Looks ok. Bit surprised ARM needed so many manual changes, but then again it is old.

PR #885, opened 2025-04-09
Refactor det_ext to remove scheduler state

corlewis:
The scheduler is now fully part of the extensible abstract specification, because we are increasingly finding properties that are not able to be reasonably specified without it. This means that there is now no non-deterministic scheduler specification and that the only slots for non-determinism are now CDT and preemption operations.

This came up due to some properties currently being proved not being true for all execution paths of the previous non-deterministic specification. Options that were considered were:
1 - move the minimal state (the current domain and tcb's domain) required for the properties currently being considered to the extensible abstract specification.
2 - move all of the scheduler state to the extensible abstract specification, the same as what has been done on the rt branch.
3 - move all state and remove the non-deterministic specification completely.
4 - add ghost state to sidestep the specific problem that we are currently encountering (it is unclear whether this would actually work).

Option 2 was chosen because it is the least blocking (although may be more annoying than some of the others), will result in the smallest diff to the rt branch, and we don't think it will be significantly more work than option 1. If we want to reason about threads in the future (e.g. liveness), we will require that this information about the scheduler state is accessible.

For most of the scheduler state it is clear that we do want to move it but we were not sure about the scheduler action in particular. For now we have decided to move it across because the correctness of the ready queues depends on it and any alternatives that we have thought of would be a lot of work that we do not want to do at this time.

All content in this has been previously reviewed in the following PRs:
https://github.com/seL4/l4v/pull/824
https://github.com/seL4/l4v/pull/844
https://github.com/seL4/l4v/pull/840
https://github.com/seL4/l4v/pull/880
https://github.com/seL4/l4v/pull/883
https://github.com/seL4/l4v/pull/884


corlewis:
While all content has been previously reviewed, we might still want to consider the commit structure of these changes. The current set of commits reflects how this was developed, but if we would prefer a shorter history I would be happy to squash some of them. The two options that come to mind would be to squash all of the changes for a particular architecture together, or to squash the different commits modifying the same proof session for different architectures.

Xaphiosis:
> While all content has been previously reviewed, we might still want to consider the commit structure of these changes. The current set of commits reflects how this was developed, but if we would prefer a shorter history I would be happy to squash some of them. The two options that come to mind would be to squash all of the changes for a particular architecture together, or to squash the different commits modifying the same proof session for different architectures.

The extreme case would be: spec updates (one for aspec, one for haskell+design), and then proof update in one go. The changes to generic theories were done during aarch64, right?

Thinking of our future selves, when doing a git show it's easier to slice a commit by invariants-abstract/refine/crefine/etc than it is to slice by architecture. So if I were to squash in a way that still leaves a bunch of commits, I'd split by arches, so that we can answer "how bad was the change for X64". The generic lemma changes being part of aarch64 might complicate that assessment. Not converging on a clear answer when arguing with myself like this. @lsf37 any thoughts?

lsf37:
> The extreme case would be: spec updates (one for aspec, one for haskell+design), and then proof update in one go. The changes to generic theories were done during aarch64, right?
> 
> Thinking of our future selves, when doing a git show it's easier to slice a commit by invariants-abstract/refine/crefine/etc than it is to slice by architecture. So if I were to squash in a way that still leaves a bunch of commits, I'd split by arches, so that we can answer "how bad was the change for X64". The generic lemma changes being part of aarch64 might complicate that assessment. Not converging on a clear answer when arguing with myself like this. @lsf37 any thoughts?

I still like the idea of keeping the other architectures in their own commits, even if AARCH64 has the generic part. Squashing too much would make this one gigantic and not really provide more value give we can't really bisect for proof state on this repo anyway, so there is no point in trying too hard to have working versions between commits.

lsf37 (approved):
Happy to approve since this was already reviewed.

Xaphiosis:
> > The extreme case would be: spec updates (one for aspec, one for haskell+design), and then proof update in one go. The changes to generic theories were done during aarch64, right?
> > Thinking of our future selves, when doing a git show it's easier to slice a commit by invariants-abstract/refine/crefine/etc than it is to slice by architecture. So if I were to squash in a way that still leaves a bunch of commits, I'd split by arches, so that we can answer "how bad was the change for X64". The generic lemma changes being part of aarch64 might complicate that assessment. Not converging on a clear answer when arguing with myself like this. @lsf37 any thoughts?
> 
> I still like the idea of keeping the other architectures in their own commits, even if AARCH64 has the generic part. Squashing too much would make this one gigantic and not really provide more value give we can't really bisect for proof state on this repo anyway, so there is no point in trying too hard to have working versions between commits.

Can you be more explicit whether you want anything squashed at all in this PR then?

Xaphiosis:
In the spec commit message, you say "We are really not sure about whether moving specifically the scheduler action is the right decision" ... can you clarify what "moving specifically the scheduler action" means here?

lsf37:
> In the spec commit message, you say "We are really not sure about whether moving specifically the scheduler action is the right decision" ... can you clarify what "moving specifically the scheduler action" means here?

I read it as "the rest is clear, but for the scheduler action we were not sure". Maybe spelling that out is all that is needed.

lsf37:
> > I still like the idea of keeping the other architectures in their own commits, even if AARCH64 has the generic part. Squashing too much would make this one gigantic and not really provide more value give we can't really bisect for proof state on this repo anyway, so there is no point in trying too hard to have working versions between commits.
> 
> Can you be more explicit whether you want anything squashed at all in this PR then?

I would squash all of the proof commits for each architecture separately + have separate commits for the specs. This will include generic updates in one of the architectures (AArch64, I think). I think that's fine.

lsf37:
> Yep, that was the intended meaning. How about "For most of the scheduler state it is clear that we do want to move it but we were not sure about the scheduler action in particular. For now we have decided to move it across because ..."

That works for me.

corlewis:
Squashed and commit messages updated as discussed.

Xaphiosis (approved):

PR #886, opened 2025-04-10
Update AInvs and Access for other arches for explicit FPU changes

corlewis:
This extends the changes made in https://github.com/seL4/l4v/pull/850 for the remaining architectures. Changes to X64 were  more extensive due to its proofs needing to be made consistent with the more recent AARCH64 proofs that were being copied.

### DIFF
--- a/proof/invariant-abstract/ARM_HYP/ArchInvariants_AI.thy
+++ b/proof/invariant-abstract/ARM_HYP/ArchInvariants_AI.thy
@@ -743,6 +743,14 @@ where
 |  _ \<Rightarrow> False"
 
 
+(* arch_tcb_live stub; there is no state in the arch specific part of TCBs on this platform that
+   marks a TCB as live *)
+definition arch_tcb_live :: "arch_tcb \<Rightarrow> bool" where
+  "arch_tcb_live arch_tcb \<equiv> False"
+
+definition valid_cur_fpu :: "'z::state_ext state \<Rightarrow> bool" where
+  "valid_cur_fpu \<equiv> \<top>"
+
 definition
   is_vcpu :: "kernel_object \<Rightarrow> bool"
 where
@@ -1260,7 +1268,6 @@ lemma valid_pd_kernel_mappings [iff]:
       = valid_pd_kernel_mappings uses s"
   by (rule ext, simp add: valid_pd_kernel_mappings_def)
 
-
 (* FIXME: Clagged *)
 lemma get_cap_update [iff]:
   "(fst (get_cap p (f s)) = {(cap, f s)}) = (fst (get_cap p s) = {(cap, s)})"
@@ -1331,6 +1338,10 @@ lemma valid_vspace_objs_update' [iff]:
   "valid_vspace_objs (f s) = valid_vspace_objs s"
   by (simp add: valid_vspace_objs_def)
 
+lemma valid_cur_fpu_update [iff]:
+  "valid_cur_fpu (f s) = valid_cur_fpu s"
+  by (auto simp: valid_cur_fpu_def)
+
 end
 
 context Arch begin arch_global_naming
@@ -1364,6 +1375,10 @@ lemma valid_arch_state_lift:
    apply (case_tac "arm_current_vcpu x"; simp add: split_def)
     apply (wp arch vcpus typs hoare_vcg_conj_lift hoare_vcg_const_Ball_lift)+
   done
+
+lemma valid_cur_fpu_updates[simp]:
+  "\<And>f. valid_cur_fpu (arch_state_update f s) = valid_cur_fpu s"
+  by (auto simp: valid_cur_fpu_def)
 
 lemma aobj_at_default_arch_cap_valid:
   assumes "ty \<noteq> ASIDPoolObj"
@@ -2107,9 +2122,9 @@ lemma aa_type_vcpuD:
   by (clarsimp simp: aa_type_def
                split: arch_kernel_obj.splits if_split_asm)
 
-lemma global_refs_asid_table_update [iff]:
-  "global_refs (s\<lparr>arch_state := arm_asid_table_update f (arch_state s)\<rparr>) = global_refs s"
-  by (simp add: global_refs_def)
+lemma global_refs_updates[simp]:
+  "\<And>f. global_refs (s\<lparr>arch_state := arm_asid_table_update f (arch_state s)\<rparr>) = global_refs s"
+  by (auto simp: global_refs_def)
 
 lemma pspace_in_kernel_window_arch_update[simp]:
   "arm_kernel_vspace (f (arch_state s)) = arm_kernel_vspace (arch_state s)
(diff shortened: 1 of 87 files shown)


proof/invariant-abstract/ARM_HYP/ArchInvariants_AI.thy
```diff
@@ -2418,17 +2418,14 @@ lemma valid_tcb_arch_ref_lift:
   "tcb_arch_ref t = tcb_arch_ref t' \<Longrightarrow> valid_arch_tcb (tcb_arch t) = valid_arch_tcb (tcb_arch t')"
   by (simp add: valid_arch_tcb_def tcb_arch_ref_def)
 
-lemma valid_arch_tcb_context_update[simp]:
-  "valid_arch_tcb (tcb_context_update f t) = valid_arch_tcb t"
-  unfolding valid_arch_tcb_def obj_at_def by simp
+lemma valid_arch_tcb_simps[simp]:
+  "\<And>f. valid_arch_tcb (tcb_context_update f t) = valid_arch_tcb t"
```
Xaphiosis, line 2422: do these meta-foralls do anything? (same for `valid_arch_tcb_context_simps` below)

proof/invariant-abstract/X64/ArchInvariants_AI.thy
```diff
@@ -3048,9 +3109,25 @@ lemma vs_lookup_2ConsD:
...
+
+(* FIXME: pspace_update_eq is interpreted for arch_state_update later, in Invariants_AI *)
+lemma valid_arch_state_updates[simp]:
+  "\<And>f. valid_arch_state (s\<lparr>arch_state := x64_current_fpu_owner_update f (arch_state s)\<rparr>) = valid_arch_state s"
+  by (auto simp: valid_arch_state_def valid_asid_table_def valid_global_pts_def valid_global_pds_def
+                 valid_global_pdpts_def valid_ioports_def all_ioports_issued_def issued_ioports_def
+                 Arch_pspace_update_eq.caps_of_state_update Arch_pspace_update_eq_def pspace_update_eq_def
+                 obj_at_def a_type_simps)
+
+(* FIXME: pspace_update_eq is interpreted for arch_state_update later, in Invariants_AI *)
```
Xaphiosis, line 3125: how do you propose we fix these?

proof/invariant-abstract/X64/ArchInvariants_AI.thy
```diff
@@ -1734,8 +1753,51 @@ lemma global_refs_update [iff]:
...
+  "valid_arch_caps (f s) = valid_arch_caps s"
+  by (simp add: valid_arch_caps_def asid_table)
+
+lemma valid_kernel_mappings_update [iff]:
+  "valid_kernel_mappings (f s) = valid_kernel_mappings s"
+  by (simp add: valid_kernel_mappings_def pspace second_level_tables)
+
+lemma valid_asid_map_update [iff]:
+  "valid_asid_map (f s) = valid_asid_map s"
+  by (simp add: valid_asid_map_def)
```
Xaphiosis, line 1792: Something feels off about seeing these here... are these properties that we needed on other arches, but somehow got away with not needing on x64 before? Are they moved from somewhere else? Duplicated? It's part of `Arch_asid_table_update_eq` here.

proof/invariant-abstract/X64/ArchInvariants_AI.thy
```diff
@@ -1734,8 +1753,51 @@ lemma global_refs_update [iff]:
...
+
+end
+
 context Arch_p_arch_update_eq begin
 
+sublocale Arch_p_asid_table_update_eq
+  by unfold_locales (simp add: arch second_level_tables_def)+
```
Xaphiosis, line 1799: still not following what's happening here... for some reason I'm under the impression that lemmas in this locale that do not depend on the inclusion of Arch_p_asid_table_update_eq should go above the sublocale command, but here we have it at the top of the context block

proof/invariant-abstract/X64/ArchInvariants_AI.thy
```diff
@@ -24,30 +24,31 @@ end_qualify
...
+definition arch_tcb_to_iarch_tcb :: "arch_tcb \<Rightarrow> iarch_tcb" where
+  "arch_tcb_to_iarch_tcb arch_tcb \<equiv> \<lparr> itcb_cur_fpu = tcb_cur_fpu arch_tcb \<rparr>"
+
+(* Need one of these simp rules for each field in 'iarch_tcb' *)
+lemma arch_tcb_to_iarch_tcb_simps[simp]:
+  "itcb_cur_fpu (arch_tcb_to_iarch_tcb arch_tcb) = tcb_cur_fpu arch_tcb"
+  by (auto simp: arch_tcb_to_iarch_tcb_def)
+
+lemma iarch_tcb_simps[simp]:
+  "\<And>f. arch_tcb_to_iarch_tcb (tcb_context_update f arch_tcb) = arch_tcb_to_iarch_tcb arch_tcb"
```
Xaphiosis, line 42: more of the strange meta-forall

proof/invariant-abstract/X64/ArchBits_AI.thy
```diff
@@ -10,6 +10,14 @@ begin
 
 context Arch begin arch_global_naming
 
+(* arch-specific interpretations of update locales: *)
+
+(* FIXME: do this for other x64 arch_state fields *)
```
Xaphiosis, line 15: is this an arch-split, or do you think it can be left for arch-local improvements if they ever happen for x64?

Xaphiosis (approved):
I think this is ok. I noticed the potentially superfluous meta-foralls, which might have been in previous PRs and I didn't notice maybe. The x64 locale situation is probably fine, but I'd like to understand what happened as I recently was overhauling these and am now confused.

Something to consider for final merge: this will be a lot of commits, and riscv seems really uninteresting, will we want to ultimately squash more (thought occurred when looking at #885) before this hits master?

proof/invariant-abstract/ARM_HYP/ArchInvariants_AI.thy
corlewis: It's a pattern I've been using for when I collect a whole bunch of simp rules into one lemma. For example, see the AARCH64 version of this lemma
https://github.com/seL4/l4v/blob/78ad44e4a5594df14a4e50fe17680df2f1eeb4e0/proof/invariant-abstract/AARCH64/ArchInvariants_AI.thy#L1567-L1569
There it's needed because `f` has a different type for each update.
In this case where it's just a single simp rule then it's not needed and I probably did it due to copy-pasting from AARCH64. I could remove it but I sort of like leaving it in, both for more consistency between architectures and to make it easier to modify if this architecture has more fields added in the future.

proof/invariant-abstract/X64/ArchInvariants_AI.thy
corlewis: The high-level summary is that these locale changes are me backporting improvements from AARCH64, made a bit complicated due to not doing a complete job of it and focusing on the parts that I need now.

In this case, it's the addition of a locale that is specifically for functions that do not touch the kheap and asid_table. Some of the properties were then moved from a more general locale (where all of arch_state is preserved), while a couple of new properties were copied from AARCH64. The main benefit for me is that this new locale can be instantiated for `x64_current_fpu_owner_update` later on, giving me some useful lemmas for free.

proof/invariant-abstract/X64/ArchInvariants_AI.thy
corlewis: I have to admit that I was mostly just copying AARCH64 and not thinking too hard about what it was doing, but my intuition is that `Arch_p_arch_update_eq` is a strict extension of `Arch_p_asid_table_update_eq` and so it's fine to do the sublocale as the first thing.

proof/invariant-abstract/X64/ArchBits_AI.thy
corlewis: I think the latter. See https://github.com/seL4/l4v/blob/78ad44e4a5594df14a4e50fe17680df2f1eeb4e0/proof/invariant-abstract/AARCH64/ArchBits_AI.thy#L14 for what I was copying from AARCH64. It would probably be fine to do the same for all of the arch_state fields on X64 but there's quite a lot of them and at this stage I just wanted to get my current proofs through, so I just did the one I need now.

proof/invariant-abstract/X64/ArchInvariants_AI.thy
corlewis: That is a good question. The specific issue is that I'm using a lemma from `pspace_update_eq` and manually unfolding its axioms to show that they're satisfied. If we're fine with that then I could remove the FIXMEs.

I also just realised that this could probably be resolved by moving these lemmas into `pspace_update_eq`. That would have the problem though that it's then not consistent with AARCH64, which doesn't need to do that due to these properties there not being dependent on `caps_of_state`.

corlewis:
> Something to consider for final merge: this will be a lot of commits, and riscv seems really uninteresting, will we want to ultimately squash more (thought occurred when looking at #885) before this hits master?

Yes, this work will be the same as with #885 about whether it makes sense to squash more. I think I would prefer to keep this branch as is with a lot of commits while it's being finished and then depending on what we decide in #885 we can do a big squash as part of the final merge into master.

proof/invariant-abstract/ARM_HYP/ArchInvariants_AI.thy
lsf37: It is consistent with the lemmas just below where it is required, so I'd be fine with leaving it for this one as well.

lsf37 (approved):
Cool. I have no new observations to add from what Raf already commented. Thanks for back-porting the machine op setup in Machine_AI, that is much nicer now.

I'll leave the locale discussion mostly to Raf, because he is the expert on that one.

Xaphiosis:
I'll think about some of the FIXMEs and other replies, but regardless that's not for this PR, so clear to merge from my side.

PR #888, opened 2025-04-16
Rename fpu_state to user_fpu_state in user contexts

ryybrr:
On architectures with FPU support, user contexts have a `user_regs` field and an `fpu_state` field. This commit renames `fpu_state` to `user_fpu_state`.

Test with https://github.com/seL4/seL4/pull/1325

PR #889, opened 2025-04-21
AARCH64 Access Control

ryybrr:
More details to come


spec/design/m-skel/AARCH64/MachineTypes.thy
```diff
@@ -52,16 +52,39 @@ text \<open>
...
+  vcpu_vgic :: "gic_vcpu_interface"
+  vcpu_regs :: "vcpureg \<Rightarrow> machine_word"
+
+record machine_state =
   irq_masks :: "AARCH64.irq \<Rightarrow> bool"
   irq_state :: nat
   underlying_memory :: "machine_word \<Rightarrow> word8"
   device_state :: "machine_word \<Rightarrow> word8 option"
+  vcpu_state :: vcpu_state
+  fpu_state :: AARCH64.fpu_state
```
corlewis, line 87: Now that https://github.com/seL4/l4v/pull/888 has been merged, does this still need to be qualified?

spec/machine/AARCH64/MachineOps.thy
```diff
@@ -194,21 +184,25 @@ definition setFPUState :: "fpu_state \<Rightarrow> unit user_monad" where
 \<comment> \<open>FIXME FPU: The underlying register is saved and restored when switching between VCPUs, so this
      will probably need to be moved out of machine\_rest\_monad.\<close>
 consts' enableFpuEL01_impl :: "unit machine_rest_monad"
+consts' enableFpuEL01_val :: "machine_word \<Rightarrow> machine_word"
 definition enableFpuEL01 :: "unit machine_monad" where
-  "enableFpuEL01 \<equiv> machine_op_lift enableFpuEL01_impl"
+  "enableFpuEL01 \<equiv> do
+     modify (vcpu_state_update (vcpu_regs_update (\<lambda>r. r(VCPURegCPACR := enableFpuEL01_val (r VCPURegCPACR)))));
+     machine_op_lift enableFpuEL01_impl
+   od"
```
corlewis, line 192: Can you explain what's going on here? It's not clear to me how the _val is being used, and whether this is enough to remove the FIXME.

spec/machine/AARCH64/MachineOps.thy
```diff
@@ -194,21 +184,25 @@ definition setFPUState :: "fpu_state \<Rightarrow> unit user_monad" where
...
    od"
 
 consts' writeFpuState_impl :: "fpu_state \<Rightarrow> unit machine_rest_monad"
 definition writeFpuState :: "fpu_state \<Rightarrow> unit machine_monad" where
-  "writeFpuState val \<equiv>  do
+  "writeFpuState fpu \<equiv>  do
      state_assert fpu_enabled;
-     machine_op_lift $ writeFpuState_impl val
+     modify (fpu_state_update (\<lambda>_. fpu));
+     machine_op_lift $ writeFpuState_impl fpu
```
corlewis, line 205: Do we still need `writeFpuState_impl`? It's not clear to me what else `writeFpuState` would be doing.

spec/machine/AARCH64/MachineOps.thy
```diff
@@ -280,29 +278,35 @@ definition addressTranslateS1 :: "machine_word \<Rightarrow> machine_word machin
...
-consts' gic_vcpu_ctrl_vmcr_val :: "machine_state \<Rightarrow> 32 word"
 definition get_gic_vcpu_ctrl_vmcr :: "32 word machine_monad" where
-  "get_gic_vcpu_ctrl_vmcr \<equiv> gets gic_vcpu_ctrl_vmcr_val"
+  "get_gic_vcpu_ctrl_vmcr \<equiv> gets (\<lambda>ms. vgic_vmcr (vcpu_vgic (vcpu_state ms)))"
 
 consts' set_gic_vcpu_ctrl_vmcr_impl :: "32 word \<Rightarrow> unit machine_rest_monad"
 definition set_gic_vcpu_ctrl_vmcr :: "32 word \<Rightarrow> unit machine_monad" where
-  "set_gic_vcpu_ctrl_vmcr w \<equiv> machine_op_lift (set_gic_vcpu_ctrl_vmcr_impl w)"
+  "set_gic_vcpu_ctrl_vmcr w \<equiv> do
+     modify (vcpu_state_update (vcpu_vgic_update (vgic_vmcr_update (\<lambda>_. w))));
```
corlewis, line 297: Probably not needed for this PR but it could be nice to have some wrappers to make it easier to get and modify specific vcpu registers.

proof/access-control/RISCV64/ArchAccess_AC.thy
```diff
@@ -71,19 +78,73 @@ lemma auth_ipc_buffers_tro[Access_AC_assms]:
...
+  "integrity_fpu_2 a b c d e f g = integrity_fpu_2 t u v w x y z"
+  by simp
+
+lemmas integrity_arch_triv =
+  integrity_hyp_wp_triv
+  integrity_fpu_wp_triv
+  integrity_hyp_triv
+  assumption2[OF integrity_hyp_triv]
+  assumption2[OF assumption2[OF integrity_hyp_triv]]
+  assumption2[OF assumption2[OF assumption2[OF integrity_hyp_triv]]]
```
lsf37, line 129: Can you explain what is going on with these instances? What are we trying to achieve?

I have the feeling that we can get away without assumption2, but I'm not actually clear on what final lemmas we are trying to build.

spec/machine/AARCH64/MachineOps.thy
ryybrr: `enableFpuEL01_val` abstracts from how the current value is modified. FIXME can be removed, thanks for spotting

spec/machine/AARCH64/MachineOps.thy
ryybrr: Not sure whether the write to machine state has side effects, so have left it in. Happy to remove if we know it won't trigger anything

proof/access-control/RISCV64/ArchAccess_AC.thy
ryybrr: Couldn't get `fact` to solve unless the assumption count matched. Could use `rule` or such instead

lsf37: Ah, interesting. Too used to  `assumption`. I think it would make sense to use the `assumption` method instead of `fact` in this case. We do want the interface to match up, but just having fewer assumptions shouldn't receive that much punishment.

ryybrr: The `assumption` method wouldn't actually work. Using OF means the rules end up looking like:
`P ==> integrity_hyp`
`P ==> Q ==> integrity_hyp`
etc.

spec/machine/AARCH64/MachineOps.thy
ryybrr: I did think about this. There was a lemma I wrote (but may have refactored out) for these generic `modify` VCPU updates.

spec/design/m-skel/AARCH64/MachineTypes.thy
ryybrr: Still necessary, this builds off that commit. Note that `fpu_state` (the type) has moved. It's similar to how `AARCH64.irq` is qualified

proof/access-control/RISCV64/ArchAccess_AC.thy
ryybrr: I do agree that this is brittle and feels redundant. If we let ourselves stray away from `fact`, we could have `solves (rule integrity_hyp_triv)`

spec/machine/AARCH64/MachineOps.thy
lsf37: It's the weaker assumption, i.e. overall better if we don't need to assume no side effect. If the proofs work, I'd leave it as is.

spec/machine/AARCH64/MachineOps.thy
lsf37: It seems mostly the vcpu+vgic combination that could use a wrapper. It's not used *that* often, though, so I'm happy with it either way.

proof/access-control/CNode_AC.thy
```diff
@@ -1176,13 +1177,19 @@ lemma sts_respects_restart_ep:
...
   apply (rule_tac tro_tcb_restart [OF refl refl])
     apply (fastforce dest!: get_tcb_SomeD)
    apply (fastforce dest!: get_tcb_SomeD)
   apply simp
   done
 
+(* FIXME AC: move *)
+lemma opt_map_upd_Some_unfolded[simp]:
+  "(\<lambda>x. if x = p then Some v else f x) |> g =
+   (\<lambda>x. if x = p then g v else (f |> g) x)"
```
lsf37, line 1190: This is probably moved from elsewhere -- it'd make sense to put it into the option reader monad theory if we're touching it already anyway. I don't think it should be globally `[simp]`, though, at least it's not clear to me that this direction is always the one you'd want. We could do a `declare` simp here and move just the lemma.

proof/access-control/CNode_AC.thy
```diff
@@ -411,6 +392,29 @@ lemma dmo_wp:
   shows "\<lbrace>\<lambda>s. P s (machine_state s)\<rbrace> do_machine_op mop \<lbrace>R\<rbrace>"
   by (wpsimp simp: do_machine_op_def use_valid[OF _ mopv])
 
+(* FIXME: MOVE *)
+(* do_machine_op distributing over binds/basic operations*)
+lemma dmo_distr:
```
lsf37, line 397: Also move internally here, but would make sense to put somewhere high up in AInvs if we're touching it already.

lsf37 (approved):
Otherwise no further nitpicks from my side. All good to go when those are done.

proof/invariant-abstract/AARCH64/ArchInvariants_AI.thy
```diff
@@ -686,10 +686,19 @@ locale_abbrev valid_global_tables :: "'z::state_ext state \<Rightarrow> bool" wh
 
 lemmas valid_global_tables_def = valid_global_tables_2_def
 
+(* The number of list registers should not exceed the maximum word value *)
```
Xaphiosis, line 689: This doesn't really help, and the commit it comes in says "add numlistregs invariant" which also doesn't match the name of this invariant... I'm not really sure why it's fine to have a bound of `2^word_bits` anyhow.

Xaphiosis (approved):
I'm not an access control expert, but things look ok to me.

None of the commits have any extra details in the message though, and I'm not a fan of "misc updates". I'd have thought that you'd have some commentary to commit to the history books, especially given all the discussion that happened about how to go about these.

ryybrr:
> None of the commits have any extra details in the message though, and I'm not a fan of "misc updates".

Fixed @Xaphiosis. Will keep `misc updates` as the changes are minor but varied (now described in the commit message).

proof/invariant-abstract/AARCH64/ArchInvariants_AI.thy
lsf37: Why not? (they are `nat`)

ryybrr: Getting and setting list registers involves `word` to `nat` conversions (`vgic_lr` expects `nat`). This puts a limit on the number of "real" list registers, regardless of whichever `arm_gicvcpu_numlistregs` value you chose. The invariant ensures that these "real" list registers map directly onto `[0, arm_gicvcpu_numlistregs)` (i.e. no truncation).

Xaphiosis: Yeah, so something about this would make a better comment than only repeating in words what the function already does. Roughly speaking we're ensuring that the state relation is a function by clamping the abstract side to the storage size of the concrete side. As you can see this wasn't obvious to either of us, and would make a better comment.

Xaphiosis:
> > None of the commits have any extra details in the message though, and I'm not a fan of "misc updates".
> 
> Fixed @Xaphiosis. Will keep `misc updates` as the changes are minor but varied (now described in the commit message).

Would still prefer to see `arm_gicvcpu_numlistregs` instead of `numlistregs` in the first commit line if there's room, and the commit message has the same issue as that comment about the invariant we discussed above.

proof/invariant-abstract/AARCH64/ArchInvariants_AI.thy
ryybrr: @Xaphiosis updated

PR #873, opened 2025-03-21
Update AARCH64 Refine and CRefine for explicit FPU changes

corlewis:

PR #892, opened 2025-04-23
Update Refine and CRefine for other arches for explicit FPU changes

corlewis:
This extends the changes made in https://github.com/seL4/l4v/pull/873 for the remaining architectures. Changes to X64 were generally more extensive due to its proofs needing to be made consistent with the more recent AARCH64 proofs that were being copied.

### DIFF
--- a/proof/crefine/ARM/Schedule_C.thy
+++ b/proof/crefine/ARM/Schedule_C.thy
@@ -357,11 +357,20 @@ lemma nextDomain_ccorres:
   apply simp
   done
 
+lemma prepareNextDomain_ccorres:
+  "ccorres dc xfdc invs' UNIV [] prepareNextDomain (Call prepareNextDomain_'proc)"
+  apply cinit
+  apply (rule ccorres_return_Skip)
+  by clarsimp
+
 lemma scheduleChooseNewThread_ccorres:
   "ccorres dc xfdc
      (\<lambda>s. invs' s \<and> ksSchedulerAction s = ChooseNewThread) UNIV hs
      (do domainTime \<leftarrow> getDomainTime;
-         y \<leftarrow> when (domainTime = 0) nextDomain;
+         y \<leftarrow> when (domainTime = 0) (do
+             y <- prepareNextDomain;
+             nextDomain
+         od);
          chooseThread
       od)
      (Call scheduleChooseNewThread_'proc)"
@@ -370,12 +379,14 @@ lemma scheduleChooseNewThread_ccorres:
    apply (rule ccorres_split_nothrow)
        apply (rule_tac R="\<lambda>s. ksDomainTime s = domainTime" in ccorres_when)
         apply (fastforce simp: rf_sr_ksDomainTime)
-       apply (rule_tac xf'=xfdc in ccorres_call[OF nextDomain_ccorres] ; simp)
+       apply (ctac (no_vcg) add: prepareNextDomain_ccorres)
+        apply (rule ccorres_call[OF nextDomain_ccorres, where xf'=xfdc] ; simp)
+       apply wpsimp
       apply ceqv
      apply (ctac (no_vcg) add: chooseThread_ccorres)
     apply (wp nextDomain_invs_no_cicd')
    apply clarsimp
-   apply (vcg exspec=nextDomain_modifies)
+   apply (vcg exspec=nextDomain_modifies exspec=prepareNextDomain_modifies)
   apply (clarsimp simp: if_apply_def2 invs'_invs_no_cicd')
   done
 
--- a/proof/crefine/ARM/Tcb_C.thy
+++ b/proof/crefine/ARM/Tcb_C.thy
@@ -4418,6 +4418,95 @@ lemma decodeSetTLSBase_ccorres:
   apply (auto simp: unat_eq_0 le_max_word_ucast_id)+
   done
 
+lemma invokeTCB_SetFlags_ccorres:
+  notes hoare_weak_lift_imp [wp]
+  shows
+  "ccorres (cintr \<currency> (\<lambda>rv rv'. rv = [])) (liftxf errstate id (K ()) ret__unsigned_long_')
+   (invs' and tcb_at' tcb)
+   (\<lbrace>\<acute>thread = tcb_ptr_to_ctcb_ptr tcb\<rbrace>
+    \<inter> \<lbrace>\<acute>clear = clears\<rbrace>
+    \<inter> \<lbrace>\<acute>set = sets\<rbrace>) []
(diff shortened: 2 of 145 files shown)


proof/crefine/X64/Machine_C.thy
```diff
@@ -95,6 +78,26 @@ assumes invalidateLocalPageStructureCacheASID_ccorres:
            (doMachineOp (invalidateLocalPageStructureCacheASID vspace asid))
            (Call invalidateLocalPageStructureCacheASID_'proc)"
 
+
+(* FPU-related machine ops *)
+
+assumes saveFpuState_ccorres:
```
lsf37, line 84: Much nicer machine interface than what we had before. Glad it worked out so well.

spec/machine/X64/MachineOps.thy
```diff
@@ -360,6 +360,7 @@ where
 "numIODomainIDBits \<equiv> undefined"
 *)
 
+(* FIXME: unused and clashes with X64_H.hwASIDInvalidate, remove? *)
 definition
   hwASIDInvalidate :: "word64 \<Rightarrow> machine_word \<Rightarrow> unit machine_monad"
 where
```
lsf37, line 366: If it is unused, yes should definitely remove.

corlewis:
> We should assemble all the parts into one gigantic PR for everything to check that all is consistent and all tests are passing when everything is merged.

Yep, that's the plan!

proof/refine/X64/VSpace_R.thy
```diff
@@ -44,6 +44,71 @@ crunch findVSpaceForASID, haskell_fail
   for inv[wp]: "P"
   (simp: const_def crunch_simps wp: loadObject_default_inv crunch_wps ignore_del: getObject)
 
+lemma dmo_invs_no_cicd_lift': (* FIXME AARCH64: move up *)
```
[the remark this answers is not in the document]
corlewis: Hmm, that does make sense but there's also a part of me that wants to keep a link to AARCH64, so that if it gets fixed there then someone might update these as well. Thoughts on if that makes sense and if so what would be a nice way to do it?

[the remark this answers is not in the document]
lsf37: If I'm in the AARCH64 session resolving FIXMEs I'd be fairly unlikely to switch over to X64 and resolve some there even if they are similar. It'd be way too much rebuild time for a single FIXME.

proof/crefine/ARM/Schedule_C.thy
```diff
@@ -357,11 +357,20 @@ lemma nextDomain_ccorres:
...
+  by clarsimp
+
 lemma scheduleChooseNewThread_ccorres:
   "ccorres dc xfdc
      (\<lambda>s. invs' s \<and> ksSchedulerAction s = ChooseNewThread) UNIV hs
      (do domainTime \<leftarrow> getDomainTime;
-         y \<leftarrow> when (domainTime = 0) nextDomain;
+         y \<leftarrow> when (domainTime = 0) (do
+             y <- prepareNextDomain;
+             nextDomain
```
Xaphiosis, line 372: indent question: shouldn't the terms under the `when` be indented more than the `when`, hence also the following `od`?

proof/crefine/ARM/Tcb_C.thy
```diff
@@ -4418,6 +4418,95 @@ lemma decodeSetTLSBase_ccorres:
...
+lemma invokeTCB_SetFlags_ccorres:
+  notes hoare_weak_lift_imp [wp]
+  shows
+  "ccorres (cintr \<currency> (\<lambda>rv rv'. rv = [])) (liftxf errstate id (K ()) ret__unsigned_long_')
+   (invs' and tcb_at' tcb)
+   (\<lbrace>\<acute>thread = tcb_ptr_to_ctcb_ptr tcb\<rbrace>
+    \<inter> \<lbrace>\<acute>clear = clears\<rbrace>
+    \<inter> \<lbrace>\<acute>set = sets\<rbrace>) []
+   (invokeTCB (SetFlags tcb clears sets))
+   (Call invokeSetFlags_'proc)"
```
Xaphiosis, line 4430: indentation above

proof/refine/X64/Finalise_R.thy
```diff
@@ -3581,6 +3574,15 @@ lemma cap_delete_one_corres:
...
   apply fastforce
   done
+
+context
+notes option.case_cong_weak[cong]
+begin
+crunch ThreadDecls_H.suspend, unbindNotification
+  for no_0_obj'[wp]: no_0_obj'
+  (simp: crunch_simps wp: crunch_wps getCTE_no_0_obj'_helper)
+end
```
Xaphiosis, line 3584: doesn't crunch take cong these days?

proof/refine/X64/IpcCancel_R.thy
```diff
@@ -1276,40 +1276,29 @@ lemma (in delete_one) suspend_corres:
...
-      apply (rule corres_Id, rule refl, simp)
-      apply wpsimp
-     apply clarsimp
-     apply wpsimp
-    apply wpsimp
-   apply auto
+context begin interpretation Arch .
+
+lemma tcb_ko_at':
+  "tcb_at' t s \<Longrightarrow> \<exists>ta::tcb. ko_at' ta t s"
```
Xaphiosis, line 1282: t' nicer than ta

Xaphiosis (approved):
Quick skim only found one or two very minor things.

proof/refine/X64/Finalise_R.thy
corlewis: Not that I'm aware of.

proof/crefine/ARM/Schedule_C.thy
corlewis: Hmm, this is a confusing case with the variable binding but I think it's right. If we ignore the spurious variables then the `od` is the same indentation as the `when`, and the terms inside it are indented further like you say. This also matches what we were happy with in the abstract:

https://github.com/seL4/l4v/blob/eeafbc3e6fa018186c38e2544a85fc00a0ddb480/spec/abstract/Schedule_A.thy#L94-L97

proof/refine/X64/VSpace_R.thy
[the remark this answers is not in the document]
corlewis: Fair enough. I guess what I'm then not sure about is what benefit we get from mentioning an arch at all, and now that they occurs on multiple architectures whether they should just be plain FIXMEs. That's out of scope of this PR though so for now I'll just mark them X64.

PR #894, opened 2025-04-29
Prove cur_fpu_in_cur_domain

corlewis:
Show that the current FPU owner is always in the current domain.


corlewis:
Hmm, I didn't think about where these new files should fit into the sessions so that they get tested. The obvious choice would be to have them be imported by `ArchDetSchedSchedule_AI` so that they're built as part of `AInvs`, but that would require moving some lemmas around to keep the proofs working. The only alternative that I can think of is that that they could be part of a new session like with `RefineOrphanage`, but that feels pretty heavyweight for this.

corlewis:
> Hmm, I didn't think about where these new files should fit into the sessions so that they get tested. The obvious choice would be to have them be imported by `ArchDetSchedSchedule_AI` so that they're built as part of `AInvs`, but that would require moving some lemmas around to keep the proofs working. The only alternative that I can think of is that that they could be part of a new session like with `RefineOrphanage`, but that feels pretty heavyweight for this.

Another alternative is that we could restructure `AInvs` in the same way that has been done in the `rt` branch, where `AInvs.thy` is the top-level theory file. This would make it possible for the import structure to be `AInvs_AI` imports `ArchFPU_AI` which imports `ArchDetSchedSchedule_AI`. Now that I think about it, this is actually something that I would like to do anyway, although it would by far be the most work out of these options.

proof/invariant-abstract/AARCH64/ArchFPU_AI.thy
```diff
@@ -0,0 +1,381 @@
...
+
+section \<open>cur_fpu_in_cur_domain\<close>
+
+definition cur_fpu_in_cur_domain :: "det_ext state \<Rightarrow> bool" where
+  "cur_fpu_in_cur_domain s \<equiv> none_top (\<lambda>t. in_cur_domain t s) (arm_current_fpu_owner (arch_state s))"
+
+\<comment> \<open>FIXME: move\<close>
+lemma none_top_post_wp:
+  "\<lbrakk>\<And>x. \<lbrace>P x\<rbrace> f \<lbrace>\<lambda>_. Q x\<rbrace>\<rbrakk>
+   \<Longrightarrow> \<lbrace>\<lambda>s. none_top (\<lambda>x. P x s) y\<rbrace> f \<lbrace>\<lambda>_ s. none_top (\<lambda>x. Q x s) y\<rbrace>"
```
lsf37, line 21: Nice. Is there are good theory to put this one in? The wp theories probably don't include none_top yet and vice versa. Maybe somewhere else in Lib then.

proof/invariant-abstract/X64/ArchFPU_AI.thy
```diff
@@ -0,0 +1,390 @@
...
+
+context Arch begin arch_global_naming
+
+section \<open>cur_fpu_in_cur_domain\<close>
+
+definition cur_fpu_in_cur_domain :: "det_ext state \<Rightarrow> bool" where
+  "cur_fpu_in_cur_domain s \<equiv> none_top (\<lambda>t. in_cur_domain t s) (x64_current_fpu_owner (arch_state s))"
+
+\<comment> \<open>FIXME: move\<close>
+lemma none_top_post_wp:
```
lsf37, line 19: Would definitely be good to move before we merge, since it is duplicated.

lsf37 (approved):
Short and sweet compared to many of the other things we have been doing recently :-)

lsf37:
> Another alternative is that we could restructure `AInvs` in the same way that has been done in the `rt` branch, where `AInvs.thy` is the top-level theory file. This would make it possible for the import structure to be `AInvs_AI` imports `ArchFPU_AI` which imports `ArchDetSchedSchedule_AI`. Now that I think about it, this is actually something that I would like to do anyway, although it would by far be the most work out of these options.

I'm not actually a fan of that specific restructure. It is very nice to know that `ArchDetSchedSchedule_AI` is not needed for the rest of `AInvs`, which is what the current structure gives us. What we could do is introduce a new top-level theory that includes both. That might also be easier to implement without moving lemmas etc.

In MCS `ArchDetSchedSchedule_AI` should really mostly disappear and be spread out either into existing theories or at the very least into multiple logical units in their own theory files, preferably in a way that allows some concurrency. The current state of that theory in MCS is abysmal. The reason the structure above works in master, but not MCS is that pretty much the entire MCS AInvs proof lives in that one theory and of course AInvs does need those MCS invariants. Whereas in master, the theory really is about the deterministic scheduler and therefore AInvs does not need it, because the invariants need to hold on the generic version.

So, my proposal would be to make a new top-level theory for the AInvs session, which is not the current AInvs_AI. Happy for AInvs_AI to be renamed and for the new one to take that name if we want to keep the session name and the top-level theory name the same. That'd be fine.

proof/invariant-abstract/AARCH64/ArchFPU_AI.thy
corlewis: I guess `lib/NonDetMonadLemmaBucket`? I'm not sure if that's a file we actually want to keep around but for now it seems like the most appropriate place.

lsf37: Yes, that one would work. I don't think that file will go away completely, but maybe it can be renamed at some point to serve as a point for things like these that need a larger context.

Xaphiosis (approved):
Looks ok, Gerwin pointed out the FIXME moves already.

PR #895, opened 2025-05-05
use mlton from Isabelle distribution

lsf37:
Simplify installation and use the `mlton` included in the Isabelle distribution.

Needs to be merged after Isabelle2025 (#882), because before that the included `mlton` is too old to support Arm platforms.

### DIFF
--- a/tools/c-parser/globalmakevars
+++ b/tools/c-parser/globalmakevars
@@ -16,9 +16,10 @@ GLOB_PFX := $(realpath $(dir $(lastword $(MAKEFILE_LIST))))
 CC ?= gcc
 
 SML_COMPILER ?= mlton
-#ifndef SML_COMPILER
-#SML_COMPILER := $(if $(shell which mlton),mlton,poly)
-#endif
+
+ifndef MLTON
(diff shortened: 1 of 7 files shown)


Xaphiosis (approved):
Where is `$MLTON` defined? Does it really work for standalone parser?

tools/c-parser/globalmakevars
```diff
@@ -16,9 +16,10 @@ GLOB_PFX := $(realpath $(dir $(lastword $(MAKEFILE_LIST))))
 CC ?= gcc
 
 SML_COMPILER ?= mlton
-#ifndef SML_COMPILER
-#SML_COMPILER := $(if $(shell which mlton),mlton,poly)
-#endif
+
+ifndef MLTON
+MLTON := $(shell $(GLOB_PFX)/../../isabelle/bin/isabelle getenv -b ISABELLE_MLTON)
+endif
```
lsf37, line 22: This is where the definition of `$MLTON` happens

lsf37:
Reading isabelle-users, the mlton version from the repo doesn't actually work properly on MacOS, so instead of trying to fix this PR, I'll just close it. We can revisit when it works universally.

PR #896, opened 2025-06-03
FPU: return new flag state after invokeSetFlags

corlewis:
This modifies the specifications and updates the proofs to match a change made to the C code. When setting a TCB's flags the invocation now returns the new flag state. As part of this the type of the flags in the specifications had to be modified to be the same on all architectures, to match the C.

### DIFF
--- a/proof/crefine/ARM/VSpace_C.thy
+++ b/proof/crefine/ARM/VSpace_C.thy
@@ -1713,6 +1713,41 @@ lemma setRegister_ccorres:
   apply (auto intro: typ_heap_simps elim: obj_at'_weakenE)
   done
 
+lemma msgRegisters_ccorres:
+  "n < unat n_msgRegisters \<Longrightarrow>
+  register_from_H (ARM_H.msgRegisters ! n) = (index kernel_all_substitute.msgRegisters n)"
+  apply (simp add: kernel_all_substitute.msgRegisters_def msgRegisters_unfold fupdate_def)
+  apply (simp add: Arrays.update_def n_msgRegisters_def nth_Cons' split: if_split)
+  done
+
+(* usually when we call setMR directly, we mean to only set a single message register
+   which will fit in actual registers *)
+lemma setMR_as_setRegister_ccorres:
+  "ccorres (\<lambda>rv rv'. rv' = of_nat offset + 1) ret__unsigned_'
+      (tcb_at' thread and K (TCB_H.msgRegisters ! offset = reg \<and> offset < length msgRegisters))
+      (\<lbrace>\<acute>reg___unsigned_long = val\<rbrace>
+       \<inter> \<lbrace>\<acute>offset = of_nat offset\<rbrace>
+       \<inter> \<lbrace>\<acute>receiver = tcb_ptr_to_ctcb_ptr thread\<rbrace>) hs
+    (asUser thread (setRegister reg val))
+    (Call setMR_'proc)"
+  apply (rule ccorres_grab_asm)
+  apply (cinit' lift: reg___unsigned_long_' offset_' receiver_')
+   apply (clarsimp simp: length_msgRegisters)
+   apply (rule ccorres_cond_false)
+   apply (rule ccorres_move_const_guards)
+   apply (rule ccorres_add_return2)
+   apply (ctac add: setRegister_ccorres)
+     apply (rule ccorres_from_vcg_throws[where P'=UNIV and P=\<top>])
+     apply (rule allI, rule conseqPre, vcg)
+     apply (clarsimp simp: return_def)
+    apply (rule hoare_TrueI[of \<top>])
+   apply (vcg exspec=setRegister_modifies)
+  apply (clarsimp simp: length_msgRegisters n_msgRegisters_def not_le conj_commute)
+  apply (subst msgRegisters_ccorres[symmetric])
+   apply (clarsimp simp: n_msgRegisters_def unat_of_nat_eq)
+  apply (clarsimp simp: word_less_nat_alt word_le_nat_alt unat_of_nat_eq not_le[symmetric])
+  done
+
 lemma wordFromMessageInfo_spec:
   defines "mil s \<equiv> seL4_MessageInfo_lift \<^bsup>s\<^esup>mi"
   shows "\<forall>s. \<Gamma> \<turnstile> {s} Call wordFromMessageInfo_'proc
--- a/proof/crefine/ARM_HYP/Tcb_C.thy
+++ b/proof/crefine/ARM_HYP/Tcb_C.thy
@@ -4512,32 +4512,82 @@ lemma decodeSetTLSBase_ccorres:
 lemma invokeTCB_SetFlags_ccorres:
   notes hoare_weak_lift_imp [wp]
   shows
-  "ccorres (cintr \<currency> (\<lambda>rv rv'. rv = [])) (liftxf errstate id (K ()) ret__unsigned_long_')
-   (invs' and tcb_at' tcb)
-   (\<lbrace>\<acute>thread = tcb_ptr_to_ctcb_ptr tcb\<rbrace>
-    \<inter> \<lbrace>\<acute>clear = clears\<rbrace>
-    \<inter> \<lbrace>\<acute>set = sets\<rbrace>) []
-   (invokeTCB (SetFlags tcb clears sets))
-   (Call invokeSetFlags_'proc)"
-  apply (cinit lift: thread_' clear_' set_' simp: setFlags_def postSetFlags_def)
-   apply (simp add: liftE_def bind_assoc)
-   apply (rule ccorres_pre_threadGet, rename_tac flags)
-   \<comment> \<open>split off flags_' updates and threadSet\<close>
-   apply (rule ccorres_rhs_assoc2, rule ccorres_rhs_assoc2, rule ccorres_rhs_assoc2)
-   apply (rule ccorres_split_nothrow)
-       apply (rule_tac P="\<lambda>tcb. tcbFlags tcb = flags" in threadSet_ccorres_lemma2)
-        apply vcg
-       apply (clarsimp simp: tcb_at_h_t_valid[OF obj_tcb_at'])
-       apply (erule rf_sr_tcb_update_no_queue2[rotated],
-              (simp add: typ_heap_simps')+, simp_all?)[1]
-        apply (rule ball_tcb_cte_casesI, simp+)
-       apply (clarsimp simp: ctcb_relation_def)
-      apply ceqv
-     apply (rule ccorres_return_CE[folded return_returnOk]; simp)
-    apply (wpsimp simp: guard_is_UNIV_def)+
-   apply vcg
-  apply (clarsimp simp: obj_at'_def typ_heap_simps' ctcb_relation_def)
-  done
+  "ccorres dc xfdc
+     (invs' and (\<lambda>s. ksCurThread s = thread) and ct_in_state' ((=) Restart))
+     (\<lbrace>\<acute>thread = tcb_ptr_to_ctcb_ptr tcb\<rbrace> \<inter> \<lbrace>\<acute>clear = clears\<rbrace> \<inter> \<lbrace>\<acute>set = sets\<rbrace> \<inter> \<lbrace>\<acute>call = from_bool isCall\<rbrace>) []
+     (do reply \<leftarrow> invokeSetFlags tcb clears sets;
+         replyOnRestart thread [reply] isCall od)
+     (Call invokeSetFlags_'proc)"
+  apply (cinit' lift: thread_' clear_' set_' call_' simp: invokeSetFlags_def setFlags_def postSetFlags_def)
+   apply (clarsimp simp: liftE_liftE bind_bindE_assoc bindE_assoc bind_assoc simp del: Collect_const
+                 intro!: ccorres_liftE')
+   apply (rule ccorres_symb_exec_r)
+     apply (rule ccorres_pre_threadGet, rename_tac flags)
+     \<comment> \<open>split off flags_' updates and threadSet\<close>
+     apply (rule ccorres_rhs_assoc2, rule ccorres_rhs_assoc2, rule ccorres_rhs_assoc2)
+     apply (rule ccorres_split_nothrow)
+         apply (rule_tac P="\<lambda>tcb. tcbFlags tcb = flags" in threadSet_ccorres_lemma2)
+          apply vcg
+         apply (clarsimp simp: tcb_at_h_t_valid[OF obj_tcb_at'])
+         apply (erule rf_sr_tcb_update_no_queue2[rotated],
+                (simp add: typ_heap_simps')+, simp_all?)[1]
+          apply (rule ball_tcb_cte_casesI, simp+)
+         apply (clarsimp simp: ctcb_relation_def)
+        apply ceqv
+
+       apply (rule ccorres_Cond_rhs_Seq[rotated]; clarsimp)
+        apply (simp add: replyOnRestart_def liftE_def bind_assoc)
+        apply (rule getThreadState_ccorres_foo, rename_tac tstate)
+        apply (rule_tac P="tstate = Restart" in ccorres_gen_asm)
+        apply clarsimp
+        apply (ctac (no_vcg) add: setThreadState_ccorres)
+       apply (rule ccorres_rhs_assoc)+
+       apply (clarsimp simp: replyOnRestart_def liftE_def bind_assoc)
+       apply (rule getThreadState_ccorres_foo, rename_tac tstate)
+       apply (rule_tac P="tstate = Restart" in ccorres_gen_asm)
+       apply (clarsimp simp: bind_assoc)
+       apply (simp add: replyFromKernel_def bind_assoc)
+       apply (ctac add: lookupIPCBuffer_ccorres)
+         apply (ctac add: setRegister_ccorres)
+           apply (simp add: setMRs_single)
+           apply (ctac add: setMR_as_setRegister_ccorres[where offset=0])
+             apply clarsimp
+             apply csymbr
+             apply (simp only: setMessageInfo_def bind_assoc)
+             apply ctac
+               apply simp
+               apply (ctac add: setRegister_ccorres)
+                 apply (ctac add: setThreadState_ccorres)
+                apply wpsimp
+               apply (vcg exspec=setRegister_modifies)
+              apply wpsimp
+             apply clarsimp
+             apply vcg
+            apply wpsimp
+           apply (clarsimp simp: msgInfoRegister_def ARM_HYP.msgInfoRegister_def
+                                 Kernel_C.msgInfoRegister_def)
+           apply (vcg exspec=setMR_modifies)
+          apply wpsimp
+         apply clarsimp
+         apply (vcg exspec=setRegister_modifies)
+        apply wpsimp
+       apply clarsimp
+       apply (vcg exspec=lookupIPCBuffer_modifies)
+      apply (wpsimp wp: hoare_vcg_all_lift hoare_vcg_imp_lift' threadSet_pred_tcb_no_state
+                        threadSet_sch_act weak_if_wp' | strengthen invs_valid_objs')+
+     apply vcg
+    apply clarsimp
+    apply vcg
+   apply (rule conseqPre, vcg, clarsimp)
+  apply (clarsimp simp: invs_no_0_obj' tcb_at_invs' invs_valid_objs' invs_sch_act_wf'
+                        rf_sr_ksCurThread msgRegisters_unfold
+                        seL4_MessageInfo_lift_def message_info_to_H_def
+                        obj_at'_def typ_heap_simps' ctcb_relation_def)
+  by (auto dest: invs_valid_objs' elim!: valid_objsE'
+           simp: ARM_HYP.badgeRegister_def badgeRegister_def Kernel_C.badgeRegister_def
+                 fromPAddr_def ThreadState_defs C_register_defs projectKOs
+                 pred_tcb_at'_def obj_at'_def ct_in_state'_def typ_heap_simps' mask_2pm1
+                 valid_obj'_def valid_tcb'_def word_ao_dist word_bw_assocs word_bw_lcs)
 
 lemma decodeSetFlags_ccorres:
   "ccorres (intr_and_se_rel \<currency> dc) (liftxf errstate id (K ()) ret__unsigned_long_')
@@ -4547,11 +4597,12 @@ lemma decodeSetFlags_ccorres:
               and sysargs_rel args buffer)
        (\<lbrace>ccap_relation cp \<acute>cap\<rbrace>
         \<inter> \<lbrace>unat \<acute>length___unsigned_long = length args\<rbrace>
-        \<inter> \<lbrace>\<acute>buffer = option_to_ptr buffer\<rbrace>) []
+        \<inter> \<lbrace>\<acute>buffer = option_to_ptr buffer\<rbrace>
+        \<inter> \<lbrace>\<acute>call = from_bool isCall\<rbrace>) []
      (decodeSetFlags args cp
         >>= invocationCatch thread isBlocking isCall InvokeTCB)
      (Call decodeSetFlags_'proc)"
-  apply (cinit' lift: cap_' length___unsigned_long_' buffer_'
+  apply (cinit' lift: cap_' length___unsigned_long_' buffer_' call_'
                 simp: decodeSetFlags_def )
    apply csymbr+
    apply (rule ccorres_Cond_rhs_Seq)
@@ -4575,12 +4626,18 @@ lemma decodeSetFlags_ccorres:
                         bindE_assoc injection_handler_returnOk
                         ccorres_invocationCatch_Inr performInvocation_def)
        apply (ctac add: setThreadState_ccorres)
+         apply (rule ccorres_nondet_refinement)
+          apply (rule is_nondet_refinement_bindE)
+           apply (rule is_nondet_refinement_refl)
+          apply (simp split: if_split, rule conjI[rotated])
+           apply (rule impI, rule is_nondet_refinement_refl)
+          apply (rule impI, rule is_nondet_refinement_alternative1)
+         apply (clarsimp simp: invokeTCB_def liftE_bindE bind_assoc liftE_left)
+         apply (rule ccorres_add_returnOk)
+         apply (rule ccorres_liftE_Seq)
          apply (ctac (no_vcg) add: invokeTCB_SetFlags_ccorres)
-           apply simp
-           apply (rule ccorres_alternative2)
-           apply (rule ccorres_return_CE, simp+)[1]
-          apply (rule ccorres_return_C_errorE, simp+)[1]
-         apply (wpsimp wp: sts_invs_minor')+
+          apply (rule ccorres_return_CE, simp+)[1]
+         apply (wpsimp wp: sts_invs_minor' ct_in_state'_set)+
        apply (vcg exspec=setThreadState_modifies)
       apply wpsimp
      apply (vcg exspec=getSyscallArg_modifies)
--- a/proof/crefine/RISCV64/Wellformed_C.thy
+++ b/proof/crefine/RISCV64/Wellformed_C.thy
@@ -369,6 +369,7 @@ lemma ptr_val_tcb_ptr_mask2:
   apply (simp add: is_aligned_add_helper ctcb_offset_defs objBits_simps')
   done
 
+
 section \<open>Domains\<close>
 
 text \<open>
@@ -476,12 +477,29 @@ lemma num_tcb_queues_calculation:
   "num_tcb_queues = numDomains * numPriorities"
   unfolding num_tcb_queues_val by eval
 
+
 text \<open>maxIRQ interface\<close>
 
 declare Kernel_C.maxIRQ_def[code]
 
 (* FIXME: compute maxIRQ for kernel config instead *)
 value_type irq_array_size = "Suc (unat Kernel_C.maxIRQ)"
+
+
+text \<open>TCBFlags interface\<close>
+
+lemma scast_seL4_TCBFlag_simps[simp]:
+  "scast seL4_TCBFlag_NoFlag = 0"
+  "scast seL4_TCBFlag_fpuDisabled = tcbFlagToWord FpuDisabled"
+  by (clarsimp simp: seL4_TCBFlag_fpuDisabled_def seL4_TCBFlag_NoFlag_def tcbFlagToWord_def)+
+
+(* config_HAVE_FPU has to be unfolded here to match the implicit enum value from the preprocessor. *)
+lemma scast_seL4_TCBFlag_MASK_tcbFlagMask[simp]:
+  "scast seL4_TCBFlag_MASK = tcbFlagMask"
+  by (clarsimp simp: seL4_TCBFlag_MASK_def tcbFlagMask_def Kernel_Config.config_HAVE_FPU_def tcbFlagToWord_def)
+
+ (* end of Kernel_Config interface section *)
+
 
 (* Input abbreviations for API object types *)
 (* disambiguates names *)
--- a/proof/crefine/X64/Tcb_C.thy
+++ b/proof/crefine/X64/Tcb_C.thy
@@ -4509,8 +4509,8 @@ lemma decodeSetTLSBase_ccorres:
   done
 
 lemma scast_seL4_TCBFlag_fpuDisabled[simp]:
-  "scast seL4_TCBFlag_fpuDisabled = tcbFlagToWord (ArchFlag FpuDisabled)"
-  by (clarsimp simp: seL4_TCBFlag_fpuDisabled_def tcbFlagToWord_def archTcbFlagToWord_def)
+  "scast seL4_TCBFlag_fpuDisabled = tcbFlagToWord FpuDisabled"
+  by (clarsimp simp: seL4_TCBFlag_fpuDisabled_def tcbFlagToWord_def)
 
 lemma invokeTCB_SetFlags_ccorres:
   notes hoare_weak_lift_imp [wp]
--- a/proof/crefine/X64/Tcb_C.thy
+++ b/proof/crefine/X64/Tcb_C.thy
@@ -4508,43 +4508,93 @@ lemma decodeSetTLSBase_ccorres:
   apply (auto simp: unat_eq_0 le_max_word_ucast_id)+
   done
 
-lemma scast_seL4_TCBFlag_fpuDisabled[simp]:
-  "scast seL4_TCBFlag_fpuDisabled = tcbFlagToWord FpuDisabled"
-  by (clarsimp simp: seL4_TCBFlag_fpuDisabled_def tcbFlagToWord_def)
-
 lemma invokeTCB_SetFlags_ccorres:
   notes hoare_weak_lift_imp [wp]
   shows
-  "ccorres (cintr \<currency> (\<lambda>rv rv'. rv = [])) (liftxf errstate id (K ()) ret__unsigned_long_')
-           (invs' and tcb_at' tcb)
-           (\<lbrace>\<acute>thread = tcb_ptr_to_ctcb_ptr tcb\<rbrace>
-            \<inter> \<lbrace>\<acute>clear = clears\<rbrace>
-            \<inter> \<lbrace>\<acute>set = sets\<rbrace>) []
-           (invokeTCB (SetFlags tcb clears sets))
-           (Call invokeSetFlags_'proc)"
-  apply (cinit lift: thread_' clear_' set_' simp: setFlags_def postSetFlags_def)
-   apply (simp add: liftE_def bind_assoc)
-   apply (rule ccorres_pre_threadGet, rename_tac flags)
-   \<comment> \<open>split off flags_' updates and threadSet\<close>
-   apply (rule ccorres_rhs_assoc2, rule ccorres_rhs_assoc2, rule ccorres_rhs_assoc2)
-   apply (rule ccorres_split_nothrow)
-       apply (rule_tac P="\<lambda>tcb. tcbFlags tcb = flags" in threadSet_ccorres_lemma2)
-        apply vcg
-       apply (clarsimp simp: tcb_at_h_t_valid[OF obj_tcb_at'])
-       apply (erule rf_sr_tcb_update_no_queue2[rotated],
-              (simp add: typ_heap_simps')+, simp_all?)[1]
-        apply (rule ball_tcb_cte_casesI, simp+)
-       apply (clarsimp simp: ctcb_relation_def)
-      apply ceqv
-     apply (rule ccorres_split_nothrow_novcg_dc)
-        apply (rule_tac R=\<top> and R'="{s. flags_' s = flags && ~~ clears || sets}" in ccorres_when_strong)
-         apply (clarsimp simp: isFlagSet_def word_bw_comms)
-        apply (ctac add: fpuRelease_ccorres)
-       apply (rule ccorres_return_CE[folded return_returnOk]; simp)
-      apply (wpsimp simp: guard_is_UNIV_def)+
-   apply vcg
-  apply (clarsimp simp: obj_at'_def typ_heap_simps' ctcb_relation_def)
-  done
+  "ccorres dc xfdc
+     (invs' and (\<lambda>s. ksCurThread s = thread) and ct_in_state' ((=) Restart))
+     (\<lbrace>\<acute>thread = tcb_ptr_to_ctcb_ptr tcb\<rbrace> \<inter> \<lbrace>\<acute>clear = clears\<rbrace> \<inter> \<lbrace>\<acute>set = sets\<rbrace> \<inter> \<lbrace>\<acute>call = from_bool isCall\<rbrace>) []
+     (do reply \<leftarrow> invokeSetFlags tcb clears sets;
+         replyOnRestart thread [reply] isCall od)
+     (Call invokeSetFlags_'proc)"
+  apply (cinit' lift: thread_' clear_' set_' call_' simp: invokeSetFlags_def setFlags_def postSetFlags_def)
+   apply (clarsimp simp: liftE_liftE bind_bindE_assoc bindE_assoc bind_assoc simp del: Collect_const
+                 intro!: ccorres_liftE')
+   apply (rule ccorres_symb_exec_r)
+     apply (rule ccorres_pre_threadGet, rename_tac flags)
+     \<comment> \<open>split off flags_' updates and threadSet\<close>
+     apply (rule ccorres_rhs_assoc2, rule ccorres_rhs_assoc2, rule ccorres_rhs_assoc2)
+     apply (rule ccorres_split_nothrow)
+         apply (rule_tac P="\<lambda>tcb. tcbFlags tcb = flags" in threadSet_ccorres_lemma2)
+          apply vcg
+         apply (clarsimp simp: tcb_at_h_t_valid[OF obj_tcb_at'])
+         apply (erule rf_sr_tcb_update_no_queue2[rotated],
+                (simp add: typ_heap_simps')+, simp_all?)[1]
+          apply (rule ball_tcb_cte_casesI, simp+)
+         apply (clarsimp simp: ctcb_relation_def)
+        apply ceqv
+       apply (rule ccorres_split_nothrow_dc)
+          apply (rule_tac R=\<top> and R'="\<lbrace>\<acute>flags = flags && ~~ clears || sets && tcbFlagMask\<rbrace>"
+                       in ccorres_when_strong)
+           apply (clarsimp simp: isFlagSet_def word_bw_comms)
+          apply (ctac add: fpuRelease_ccorres)
+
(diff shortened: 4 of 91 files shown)


proof/refine/AARCH64/Tcb_R.thy
```diff
@@ -1695,14 +1709,39 @@ lemma word_to_tcb_flags_union:
...
+  apply (frule ex_tcbFlagToWord)
+  apply clarsimp
+  apply (rule_tac x=flag in exI)
+  apply (clarsimp simp: nth_is_and_neq_0 word_bw_comms)
+  apply fastforce
+  done
+
+lemma word_or_le_mask:
+  "\<lbrakk>a \<le> mask n; b \<le> mask n\<rbrakk> \<Longrightarrow> (a :: 'a :: len word) || b \<le> mask n"
+  by (metis (no_types, lifting) le_or_mask order_trans_rules(24) word_bool_le_funs(21) word_bw_assocs(2) word_bw_comms(2) word_bw_same(2))
```
lsf37: Please move to Word_Lib (one of the _Internals theories as staging area)

proof/refine/ARM/Tcb_R.thy
```diff
@@ -1706,13 +1719,39 @@ lemma word_to_tcb_flags_union:
...
+  apply (frule ex_tcbFlagToWord)
+  apply clarsimp
+  apply (rule_tac x=flag in exI)
+  apply (clarsimp simp: nth_is_and_neq_0 word_bw_comms)
+  apply fastforce
+  done
+
+lemma word_or_le_mask:
+  "\<lbrakk>a \<le> mask n; b \<le> mask n\<rbrakk> \<Longrightarrow> (a :: 'a :: len word) || b \<le> mask n"
+  by (metis (no_types, lifting) le_or_mask order_trans_rules(24) word_bool_le_funs(21) word_bw_assocs(2) word_bw_comms(2) word_bw_same(2))
```
lsf37: (obsolete when the other instance is moved)

proof/refine/ARM_HYP/Tcb_R.thy
```diff
@@ -1711,13 +1724,39 @@ lemma word_to_tcb_flags_union:
...
+  apply (frule ex_tcbFlagToWord)
+  apply clarsimp
+  apply (rule_tac x=flag in exI)
+  apply (clarsimp simp: nth_is_and_neq_0 word_bw_comms)
+  apply fastforce
+  done
+
+lemma word_or_le_mask:
+  "\<lbrakk>a \<le> mask n; b \<le> mask n\<rbrakk> \<Longrightarrow> (a :: 'a :: len word) || b \<le> mask n"
+  by (metis (no_types, lifting) le_or_mask order_trans_rules(24) word_bool_le_funs(21) word_bw_assocs(2) word_bw_comms(2) word_bw_same(2))
```
lsf37: (for removal)

proof/refine/RISCV64/Tcb_R.thy
```diff
@@ -1687,13 +1700,39 @@ lemma word_to_tcb_flags_union:
...
+  apply (frule ex_tcbFlagToWord)
+  apply clarsimp
+  apply (rule_tac x=flag in exI)
+  apply (clarsimp simp: nth_is_and_neq_0 word_bw_comms)
+  apply fastforce
+  done
+
+lemma word_or_le_mask:
+  "\<lbrakk>a \<le> mask n; b \<le> mask n\<rbrakk> \<Longrightarrow> (a :: 'a :: len word) || b \<le> mask n"
+  by (metis (no_types, lifting) le_or_mask order_trans_rules(24) word_bool_le_funs(21) word_bw_assocs(2) word_bw_comms(2) word_bw_same(2))
```
lsf37: (for removal)

proof/refine/X64/Tcb_R.thy
```diff
@@ -1719,14 +1733,39 @@ lemma word_to_tcb_flags_union:
...
+  apply (frule ex_tcbFlagToWord)
+  apply clarsimp
+  apply (rule_tac x=flag in exI)
+  apply (clarsimp simp: nth_is_and_neq_0 word_bw_comms)
+  apply fastforce
+  done
+
+lemma word_or_le_mask:
+  "\<lbrakk>a \<le> mask n; b \<le> mask n\<rbrakk> \<Longrightarrow> (a :: 'a :: len word) || b \<le> mask n"
+  by (metis (no_types, lifting) le_or_mask order_trans_rules(24) word_bool_le_funs(21) word_bw_assocs(2) word_bw_comms(2) word_bw_same(2))
```
lsf37: (and one more)

spec/haskell/src/SEL4/Object/Structures.lhs
```diff
@@ -501,6 +501,9 @@ Various operations on the free index of an Untyped cap.
 
 > data TcbFlag = FpuDisabled
 
+> numFlags :: Int
+> numFlags = 1
```
Xaphiosis: this is a confusing name; given that it's called TcbFlag above it should be `numTcbFlags`

proof/crefine/X64/Tcb_C.thy
```diff
@@ -4512,39 +4512,104 @@ lemma scast_seL4_TCBFlag_fpuDisabled[simp]:
   "scast seL4_TCBFlag_fpuDisabled = tcbFlagToWord FpuDisabled"
   by (clarsimp simp: seL4_TCBFlag_fpuDisabled_def tcbFlagToWord_def)
 
+lemma scast_seL4_TCBFlag_MASK_numFlags[simp]:
+  "scast seL4_TCBFlag_MASK = mask numFlags"
+  by (clarsimp simp: seL4_TCBFlag_MASK_def numFlags_def)
+
+\<comment> \<open>FIXME: move\<close>
```
Xaphiosis: Please prefer `(* *)` comments for FIXMEs, for consistency.
Also, is this forgotten in this PR or to be done in a future PR?

proof/refine/X64/Tcb_R.thy
```diff
@@ -1928,19 +1984,31 @@ lemma setTLSBase_invs'[wp]:
...
+crunch postSetFlags
   for invs'[wp]: invs'
   (ignore: threadSet)
 
 lemma invokeSetFlags_invs'[wp]:
   "\<lbrace>invs' and tcb_inv_wf' (tcbinvocation.SetFlags tcb clears' sets')\<rbrace>
    invokeTCB (tcbinvocation.SetFlags tcb clears' sets')
    \<lbrace>\<lambda>rv. invs'\<rbrace>"
-  by (wpsimp simp: invokeTCB_def)
+  apply (clarsimp simp: invokeTCB_def invokeSetFlags_def)
```
Xaphiosis: suggest unfolding

proof/refine/X64/Tcb_R.thy
```diff
@@ -1800,6 +1844,15 @@ where
 | "tcb_inv_wf' (tcbinvocation.SetFlags ref clears sets)
              = (tcb_at' ref and ex_nonz_cap_to' ref)"
 
+lemma invokeSetFlags_helper:
+  "flags = word_to_tcb_flags flags' \<Longrightarrow>
+     corres (=) \<top> (K (flags' \<le> mask numFlags))
+       (return [tcb_flags_to_word (flags - word_to_tcb_flags clears' \<union> word_to_tcb_flags sets')])
+       (return [flags' && ~~ clears' || sets' && mask numFlags])"
```
Xaphiosis: incorrect indentation on `corres`

proof/refine/X64/Tcb_R.thy
```diff
@@ -1707,6 +1708,15 @@ lemma tcbFlagToWord_bit:
   "\<exists>n. tcbFlagToWord flag = bit n"
   by (auto simp: tcbFlagToWord_def split: tcb_flag.splits simp del: bit_def)
 
+lemma tcbFlagToWord_le_mask_numFlags:
+  "tcbFlagToWord flag \<le> mask numFlags"
+  by (auto simp: tcbFlagToWord_def numFlags_def split: tcb_flag.splits)
+
+\<comment> \<open>This will probably be much harder if there are more flags\<close>
```
Xaphiosis: um, not sure what this comment helps with

proof/refine/ARM_HYP/Tcb_R.thy
```diff
@@ -1920,19 +1976,31 @@ lemma setTLSBase_invs'[wp]:
...
+  "\<lbrace>invs' and K (\<forall>flags. flags \<le> mask numFlags \<longrightarrow> f flags \<le> mask numFlags)\<rbrace>
+   threadSet (tcbFlags_update f) t
+   \<lbrace>\<lambda>_. invs'\<rbrace>"
+  by (rule hoare_gen_asm) (wpsimp wp: threadSet_invs_trivial)
+
+lemma setFlags_invs'[wp]:
+  "\<lbrace>invs' and K (flags \<le> mask numFlags)\<rbrace>
+   setFlags t flags
+   \<lbrace>\<lambda>_. invs'\<rbrace>"
+  by (wpsimp simp: setFlags_def wp: threadSet_flags_invs')
```
Xaphiosis, line 1927: I'm curious why this requires the precondition... from discussions, it seemed that extra flags get ignored. Is this a constraint resulting from an update to TCB validity? I understand needing the precondition for [c]corres, but confused why for valid.

proof/refine/ARM_HYP/Tcb_R.thy
```diff
@@ -1791,6 +1835,15 @@ where
 | "tcb_inv_wf' (tcbinvocation.SetFlags ref clears sets)
              = (tcb_at' ref and ex_nonz_cap_to' ref)"
 
+lemma invokeSetFlags_helper:
+  "flags = word_to_tcb_flags flags' \<Longrightarrow>
+     corres (=) \<top> (K (flags' \<le> mask numFlags))
+       (return [tcb_flags_to_word (flags - word_to_tcb_flags clears' \<union> word_to_tcb_flags sets')])
+       (return [flags' && ~~ clears' || sets' && mask numFlags])"
+  apply clarsimp
```
Xaphiosis: is the clarsimp here necessary?

proof/crefine/X64/Tcb_C.thy
```diff
@@ -4512,39 +4512,104 @@ lemma scast_seL4_TCBFlag_fpuDisabled[simp]:
...
+               apply (simp only: setMessageInfo_def bind_assoc)
+               apply ctac
+                 apply simp
+                 apply (ctac add: setRegister_ccorres)
+                   apply (ctac add: setThreadState_ccorres)
+                  apply wpsimp
+                 apply (vcg exspec=setRegister_modifies)
+                apply wpsimp
+               apply clarsimp
+               apply (vcg)
```
Xaphiosis: parentheses

proof/crefine/ARM_HYP/Tcb_C.thy
```diff
@@ -4509,35 +4509,90 @@ lemma decodeSetTLSBase_ccorres:
...
+  apply (clarsimp simp: invs_no_0_obj' tcb_at_invs' invs_valid_objs' invs_sch_act_wf'
+                        rf_sr_ksCurThread msgRegisters_unfold
+                        seL4_MessageInfo_lift_def message_info_to_H_def
+                        obj_at'_def typ_heap_simps' ctcb_relation_def)
+  by (auto dest: invs_valid_objs' elim!: valid_objsE'
+           simp: ARM_HYP.badgeRegister_def badgeRegister_def Kernel_C.badgeRegister_def
+                 fromPAddr_def ThreadState_defs Kernel_C.R0_def Kernel_C.R1_def projectKOs
+                 pred_tcb_at'_def obj_at'_def ct_in_state'_def typ_heap_simps' mask_2pm1
+                 valid_obj'_def valid_tcb'_def word_or_le_mask[simplified mask_2pm1] word_and_le
+                 word_bool_le_funs)
```
Xaphiosis: there is a group of lemmas that represent all of the C register definitions, which might make more sense than adding R0_def, R1_def, etc, as those are also arch-dependant.

proof/crefine/ARM_HYP/Tcb_C.thy
```diff
@@ -4509,35 +4509,90 @@ lemma decodeSetTLSBase_ccorres:
   apply (auto simp: unat_eq_0 le_max_word_ucast_id)+
   done
 
+lemma scast_seL4_TCBFlag_MASK_numFlags[simp]:
+  "scast seL4_TCBFlag_MASK = mask numFlags"
+  by (clarsimp simp: seL4_TCBFlag_MASK_def numFlags_def)
```
Xaphiosis: this is probably fine for now, but is skirting close to being an interface lemma for future arch/config links to the C, which might make Wellformed_C a better place for it... but maybe we should leave that until we have more flags

proof/crefine/ARM/VSpace_C.thy
```diff
@@ -1713,6 +1713,41 @@ lemma setRegister_ccorres:
...
   done
 
+lemma msgRegisters_ccorres:
+  "n < unat n_msgRegisters \<Longrightarrow>
+  register_from_H (ARM_H.msgRegisters ! n) = (index kernel_all_substitute.msgRegisters n)"
+  apply (simp add: kernel_all_substitute.msgRegisters_def msgRegisters_unfold fupdate_def)
+  apply (simp add: Arrays.update_def n_msgRegisters_def nth_Cons' split: if_split)
+  done
+
+(* usually when we call setMR directly, we mean to only set a registers, which will
```
Xaphiosis: only set a registers? not sure what this means... only set a single register?

Xaphiosis (approved):
Looks like solid progress. Minor nitpicks only (one annoying one with `numFlags`).

For commit messages:
* `by the types used in the abstract` : `abstract spec`?
* `In particular, this makes` -> `Make`

proof/refine/AARCH64/Tcb_R.thy
corlewis: This ended up not being needed and has been removed.

spec/haskell/src/SEL4/Object/Structures.lhs
corlewis: This ended up not being needed and has been removed.

proof/refine/X64/Tcb_R.thy
corlewis: I think I originally put this here because I wasn't sure if it was the right approach or if it only makes sense when we have just one flag. I'm a bit more confident in it now and agree that the comment doesn't help much, so I've removed it.

proof/refine/ARM_HYP/Tcb_R.thy
corlewis: `setFlags` is just a wrapper around `threadSet`, the flags are masked earlier in `invokeSetFlags` and then the knowledge that they have been masked is passed via that precondition. It's needed because of the update to TCB validity like you said, we now need to know that the flags in each TCB are within the appropriate subset for the specific configuration.

proof/refine/ARM_HYP/Tcb_R.thy
corlewis: I think it might have been necessary at one point because the word proof was being annoyingly fiddly but it isn't anymore, so I've folded it into one fastforce.

proof/crefine/ARM_HYP/Tcb_C.thy
corlewis: Oh nice, I don't think I knew about `C_register_defs`. It's definitely a lot nicer to use that, it was a pain to find the right register names when copying this proof to each architecture and that would have been trivial with this.

There's still more improvements we could make here, it looks like `RISCV64` and `AARCH64` have made some improvements by adding additional register definitions that would be good to backport to the other architectures and there's lots more places we could be using `C_register_defs` throughout the proofs, but I think both of those are for other PRs.

proof/crefine/ARM/VSpace_C.thy
corlewis: I actually copied this from another architecture so wasn't sure of the original intention either but while looking at it now I found another version of this lemma on `ARM` in a later file, and that one had a better comment. No idea about what led to this situation in the past but I've updated all architectures to be consistent on this.

spec/haskell/src/SEL4/Object/Structures/AARCH64.hs
```diff
@@ -70,7 +70,6 @@ atcbContextSet uc atcb = atcb { atcbContext = uc }
 atcbContextGet :: ArchTCB -> UserContext
 atcbContextGet = atcbContext
 
-
```
Xaphiosis, line 78: change seems not related to commit

spec/haskell/src/SEL4/Object/Structures.lhs
```diff
@@ -516,3 +513,6 @@ Sets of TCB flags are bitwise OR'd and represented as a word.
 
 > isFlagSet :: TcbFlag -> TcbFlags -> Bool
 > isFlagSet flag flags = tcbFlagToWord flag .&. flags /= 0
+
+> tcbFlagMask :: Word
+> tcbFlagMask = if config_HAVE_FPU then tcbFlagToWord FpuDisabled else 0
```
Xaphiosis, line 518: if we were writing code only, I'd suggest noFlag here instead of 0. Incidentally, the 0x0 above can be 0, but I'm fine to leave both.

spec/haskell/src/SEL4/Machine/Hardware.lhs
```diff
@@ -216,3 +216,9 @@ The constant "nullPointer" is a physical pointer guaranteed to be invalid.
 > nullPointer :: PPtr a
 > nullPointer = PPtr 0
 
+\subsection{Config parameters}
+
+Whether FPUs are enabled or not.
```
Xaphiosis: "FPUs are" or "the FPU is"?

spec/haskell/src/SEL4/Object/Structures/AARCH64.hs
corlewis: I think this was to make the architectures consistent, and should be a bit less noisy once squashed into the earlier commit.

spec/haskell/src/SEL4/Machine/Hardware.lhs
corlewis: I think I was conflating "FPUs are supported" and "the FPU is enabled". I'll change it to the latter.

spec/abstract/Structures_A.thy
```diff
@@ -373,10 +373,10 @@ type_synonym domain = word8
 type_synonym tcb_flags = "tcb_flag set"
 
 definition word_to_tcb_flags :: "machine_word \<Rightarrow> tcb_flags" where
-  "word_to_tcb_flags w \<equiv> {flag. tcbFlagToWord flag && w \<noteq> 0}"
+  "word_to_tcb_flags w \<equiv> {flag. tcbFlagToWord flag && w && tcbFlagMask \<noteq> 0}"
 
 definition tcb_flags_to_word :: "tcb_flags \<Rightarrow> machine_word" where
-  "tcb_flags_to_word flags \<equiv> THE w. word_to_tcb_flags w = flags \<and> w \<le> mask numFlags"
+  "tcb_flags_to_word flags \<equiv> THE w. word_to_tcb_flags w = flags \<and> w && ~~ tcbFlagMask = 0"
```
Xaphiosis, line 382: These feel slightly unusual. Let me think through them:

Converting a word to flags is the set of all flags that when ANDed with w and the flag mask, we get a non-zero result. I think this was a bit hard to parse because I'd put `w` first. I'm also a bit confused why we need the mask, because `tcbFlagToWord` should not return anything outside of `tcbFlagMask`, no?

Converting flags to a word... again, given that you've defined it via the inverse of `word_to_tcb_flags`, doesn't this mean that checking the word against the flag mask here is unnecessary for the conversion? I understand if you need it somehow in general to ensure the validity of words which represent flags in the TCB, but it's definitely hard to see what's going on from just reading this. At minimum, it would require a comment to explain.

proof/crefine/RISCV64/Wellformed_C.thy
```diff
@@ -486,6 +486,22 @@ value_type irq_array_size = "Suc (unat Kernel_C.maxIRQ)"
...
 
+text \<open>TCBFlags interface\<close>
+
+lemma scast_seL4_TCBFlag_simps[simp]:
+  "scast seL4_TCBFlag_NoFlag = 0"
+  "scast seL4_TCBFlag_fpuDisabled = tcbFlagToWord FpuDisabled"
+  by (clarsimp simp: seL4_TCBFlag_fpuDisabled_def seL4_TCBFlag_NoFlag_def tcbFlagToWord_def)+
+
+(* config_HAVE_FPU is unfolded because the C is defined with an ifdef, and so we just see the final
+   value instead of how it is constructed. *)
```
Xaphiosis: can we just basically have this be a bit more terse+precise? I'm not sure what this comment helps really. Are you trying to say that config_HAVE_FPU matches up with the implicit enum value from the preprocessor?

proof/crefine/RISCV64/Wellformed_C.thy
```diff
@@ -486,6 +486,22 @@ value_type irq_array_size = "Suc (unat Kernel_C.maxIRQ)"
 (* Input abbreviations for API object types *)
 (* disambiguates names *)
 
+text \<open>TCBFlags interface\<close>
```
Xaphiosis, line 489: something seems wrong here. not only should there be a new line above here like we did with the other ones (well except maxIRQ, sigh), the comment above is talking about input abbreviations for API object types, which this is not. Pasted into wrong spot?

proof/crefine/RISCV64/Wellformed_C.thy
```diff
@@ -486,6 +486,22 @@ value_type irq_array_size = "Suc (unat Kernel_C.maxIRQ)"
...
+  "scast seL4_TCBFlag_fpuDisabled = tcbFlagToWord FpuDisabled"
+  by (clarsimp simp: seL4_TCBFlag_fpuDisabled_def seL4_TCBFlag_NoFlag_def tcbFlagToWord_def)+
+
+(* config_HAVE_FPU is unfolded because the C is defined with an ifdef, and so we just see the final
+   value instead of how it is constructed. *)
+lemma scast_seL4_TCBFlag_MASK_tcbFlagMask[simp]:
+  "scast seL4_TCBFlag_MASK = tcbFlagMask"
+  by (clarsimp simp: seL4_TCBFlag_MASK_def tcbFlagMask_def Kernel_Config.config_HAVE_FPU_def tcbFlagToWord_def)
+
+ (* end of Kernel_Config interface section *)
```
Xaphiosis, line 501: I think previously seeing "Input abbreviations for API object types" indicated the end of the Kernel_Config section. The problem is that there isn't a *start* of the Kernel_Config interface section per se, it starts with "Domains", so not sure what I think of this comment now

spec/abstract/Structures_A.thy
corlewis: The `tcb_flags` type can be larger than the flags that we accept on a specific architecture. The `tcbFlagMask`  in `word_to_tcb_flags` is so that we ignore any such flags (for example, `FpuDisabled` when the FPU is not enabled).

The mask is then also needed in `tcb_flags_to_word` to make it so that there is a unique word for it to return, since otherwise spurious bits could be set.

I'm happy to reorder `word_to_tcb_flags` so that `w` is first but I've been looking and making changes to these definitions for too long and can't really think of a helpful comment, that isn't just restating the definition. Would it help to add something saying that we don't care about actually computing `tcb_flags_to_word` and that it is defined as an inverse of `word_to_tcb_flags` to try and make the proofs easier?

proof/crefine/X64/Wellformed_C.thy
```diff
@@ -526,6 +526,22 @@ declare Kernel_C.maxIRQ_def[code]
...
+  "scast seL4_TCBFlag_fpuDisabled = tcbFlagToWord FpuDisabled"
+  by (clarsimp simp: seL4_TCBFlag_fpuDisabled_def seL4_TCBFlag_NoFlag_def tcbFlagToWord_def)+
+
+(* config_HAVE_FPU is unfolded because the C is defined with an ifdef, and so we just see the final
+   value instead of how it is constructed. *)
+lemma scast_seL4_TCBFlag_MASK_tcbFlagMask[simp]:
+  "scast seL4_TCBFlag_MASK = tcbFlagMask"
+  by (clarsimp simp: seL4_TCBFlag_MASK_def tcbFlagMask_def Kernel_Config.config_HAVE_FPU_def tcbFlagToWord_def)
+
+(* end of Kernel_Config interface section *)
```
Xaphiosis, line 544: see comments on other arches, but in this case the input abbreviations comment is in the right place

proof/refine/AARCH64/Schedule_R.thy
```diff
@@ -249,9 +249,23 @@ lemma switchLocalFpuOwner_corres[corres]:
    apply (auto simp: current_fpu_owner_Some_tcb_at fpuOwner_asrt_cross)
   done
 
+(*FIXME FPU: move tcbFlag lemmas*)
```
Xaphiosis: FIXME

proof/refine/AARCH64/Schedule_R.thy
```diff
@@ -249,9 +249,23 @@ lemma switchLocalFpuOwner_corres[corres]:
...
+lemma tcbFlagToWord_bit:
+  "\<exists>n<64. tcbFlagToWord flag = bit n"
```
Xaphiosis: is this machine_word_len, or specifically 64?

proof/crefine/RISCV64/Wellformed_C.thy
corlewis: You're right, pasted into the wrong spot, it should be before that comment. I'll add the extra line, I think I might have even had it originally before removing it to be consistent with `maxIRQ`.

proof/refine/AARCH64/Schedule_R.thy
```diff
@@ -249,9 +249,23 @@ lemma switchLocalFpuOwner_corres[corres]:
...
+  by (auto simp: tcbFlagToWord_def ex_nat_less_eq split: tcb_flag.splits simp del: bit_def)
+
+lemma FpuDisabled_in_tcbFlagMask[simp]:
```
Xaphiosis: not sure what to do in such naming cases (capital starting letter): @lsf37 ?

proof/refine/AARCH64/Schedule_R.thy
```diff
@@ -249,9 +249,23 @@ lemma switchLocalFpuOwner_corres[corres]:
...
+  "FpuDisabled \<in> word_to_tcb_flags tcbFlagMask"
+  by (clarsimp simp: word_to_tcb_flags_def tcbFlagToWord_def tcbFlagMask_def Kernel_Config.config_HAVE_FPU_def)
+
+lemma tcbFlagToWord_tcbFlagMask:
+  "flag \<in> word_to_tcb_flags tcbFlagMask \<Longrightarrow> tcbFlagToWord flag && tcbFlagMask = tcbFlagToWord flag"
+  by (cut_tac flag=flag in tcbFlagToWord_bit) (fastforce simp: word_to_tcb_flags_def word_eq_iff)
+
 lemma isFlagSet_set:
```
Xaphiosis: not the most illuminating name, but I guess we're not going to use flags for a lot more

proof/crefine/RISCV64/Wellformed_C.thy
corlewis: The `end of Kernel_Config interface section` comment came from `AARCH64`, which is where I made this change first. I'll do another pass and try to make the architectures a bit more consistent.

proof/refine/AARCH64/Schedule_R.thy
corlewis: I'll check but I think you're right and that it should be `machine_word_len`.

proof/refine/X64/Schedule_R.thy
```diff
@@ -150,9 +150,23 @@ lemma switchLocalFpuOwner_corres[corres]:
    apply (auto simp: current_fpu_owner_Some_tcb_at fpuOwner_asrt_cross)
   done
 
+(*FIXME FPU: move tcbFlag lemmas*)
```
Xaphiosis: FIXME (all arches?)

proof/refine/AARCH64/Schedule_R.thy
corlewis: This isn't the right place for these but there's not really anywhere obvious to me to move them to. Any objections to leaving them here for now until there's more flags and we actually have an idea of where these lemmas might be needed?

corlewis: I guess if we do leave them here then I should remove the FPU tag from the FIXME.

I'd also be happy with a suggestion for a location of course.

proof/refine/ARM_HYP/Tcb_R.thy
Xaphiosis: Does adding a comment make sense?

proof/refine/X64/Tcb_R.thy
```diff
@@ -1704,66 +1704,60 @@ crunch set_flags
...
+  "word_to_tcb_flags (flags && flags') = word_to_tcb_flags flags \<inter> word_to_tcb_flags flags'"
   apply (clarsimp simp: word_to_tcb_flags_def intro!: set_eqI)
   apply (cut_tac flag=x in tcbFlagToWord_bit)
-  apply (fastforce simp: word_eq_iff tcbFlagToWord_bit)
+  apply (fastforce simp: word_eq_iff)
   done
 
-lemma word_to_tcb_flags_union:
-  "word_to_tcb_flags (flags || flags') = word_to_tcb_flags flags \<union> word_to_tcb_flags flags'"
+lemmas word_to_tcb_flags_simps = word_to_tcb_flags_not word_to_tcb_flags_or word_to_tcb_flags_and
```
Xaphiosis: reading through these a few times (once per arch), I keep thinking: are these properties arch-agnostic? maybe it'll have to be absorbed in arch-split if so, but it really looks like some kind of library that could even be spun out into its own little theory

proof/crefine/RISCV64/Wellformed_C.thy
corlewis: The main intention of the comment was to say that `config_HAVE_FPU` being unfolded is necessary and that we wont be able to match up the C and spec and split an `if` like we do in other cases. How about this?

```
(* config_HAVE_FPU has to be unfolded here to match the implicit enum value from he preprocessor. *)
```

proof/refine/ARM_HYP/Tcb_R.thy
corlewis: I don't think so, I think it makes more sense in context when you can see that `setFlags` is just a wrapper around `threadSet`, and that this lemma is just a simplification of `threadSet_flags_invs'`. Most other `threadSet` lemmas have a similar precondition constraining the value being set so that TCB validity is maintained.

spec/haskell/src/SEL4/Object/Structures/AARCH64.hs
Xaphiosis: oki

proof/refine/AARCH64/Schedule_R.thy
corlewis: I guess it could be `isFlagSet_in_word_to_tcb_flags` if we want more illumination?

proof/refine/X64/Tcb_R.thy
corlewis: You're right, these properties are arch-agnostic now. The proofs for them weren't originally, and them being so now is one of the things giving me some confidence that they'll be able to be extended for additional flags in the future.

I'd been thinking along the lines of them being absorbed in arch-split. The exact form of some of these lemmas is pretty specific to how they're used and it can be annoyingly fiddly to get the simplifier to apply some of them due to fighting with associativity and commutativity, so spinning them out on their own would take a bit of work to do nicely. That might be something to come back to but I'd prefer not as part of this PR.

spec/abstract/Structures_A.thy
Xaphiosis: Thanks, that's what I was missing here: the set of flags can be larger than the set of configured flags. `tcbFlagToWord` deals with flags, `tcbFlagMask` with configured flags. I think I would appreciate commentary to that end that above the first one, e.g.

```
text <
  The set of TCB flags may be larger than the set of configured flags (e.g. FpuDisabled is only
  used when FPU is configured), hence we only convert flags in `tcbFlagMask`. >
```

Cheers for `w` first.

For the `tcb_flags_to_word`, I think it's evident you only needed an inverse here, and the above commentary on flags vs configured flags is sufficient to clear things up.

proof/crefine/RISCV64/Wellformed_C.thy
Xaphiosis: Ah crap, fine to leave it in then. Happy for you to make it more consistent between arches though.

proof/crefine/RISCV64/Wellformed_C.thy
Xaphiosis: That's more helpful (typo "he" -> "the"), that'll work.

Question is, how is that different from the other ones in this section though? Next time we have a Kernel_Config it's probably going to be doing the same thing, as with numDomains which uses a similar trick (an enum value set from preprocessor). Hmm. Maybe @lsf37 can weigh in, but I think your comment settles this specific instance.

proof/refine/ARM_HYP/Tcb_R.thy
Xaphiosis: I'm convinced, fine to leave as is.

proof/refine/AARCH64/Schedule_R.thy
Xaphiosis: I guess it does encode the relationship between them, so I'll buy that.

proof/crefine/RISCV64/Wellformed_C.thy
Xaphiosis: I think I'd prefer it you added a line before maxIRQ part for consistency. We must've missed that before during reviewing.

proof/refine/AARCH64/Schedule_R.thy
Xaphiosis: If we leave it, then yes to removing FPU flag from FIXME.
Generally, I'm not opposed to having the flag lemmas in Tcb_R for the moment. Here we are in Schedule_R, and you felt like these don't belong here. Maybe if you can expand your thought on why they don't belong, a place will pop up?

In CRefine, we often put these kinds of things into SR_Lemmas, I guess Invariants_H is one possible place, StateRelation another. Or if it's nearly all generic (or with identical derivations in Arch), then having a `TcbFlags_R.thy` theory might be warranted. I'm not sure myself, having not worked with these lemmas closely.

proof/refine/X64/Tcb_R.thy
Xaphiosis: While arch-split isn't magic, if you put these outside of the Arch locale (or are they already?) I can pick them up when I get there, and at least the redundancy between arches should disappear.

proof/crefine/RISCV64/Wellformed_C.thy
corlewis: While I'm here, should I add an explicit `start of Kernel_Config interface section` comment instead of the implicit start with "Domains"?

Xaphiosis: Maybe not in this PR. I think having an official section would warrant a proper comment to set it up. As is stands, people can look for the end of it and add their stuff above

proof/refine/AARCH64/Schedule_R.thy
corlewis: I ended up moving these lemmas to a new arch-generic theory. Have a look and let me know if what I've done makes sense from an arch-split perspective.

proof/crefine/RISCV64/Wellformed_C.thy
lsf37: No issue with the revised comment from my side. Likely good to have it when I ever get to doing a more generic FPU config thing.

proof/refine/AARCH64/Schedule_R.thy
lsf37: I was about to write that my gut feeling tends towards lower case, but looking at it for longer, if it was `None` we would not turn it into `none` if we wanted to refer to the constructor. So I think we should leave it.

proof/refine/X64/Schedule_R.thy
```diff
@@ -150,30 +150,12 @@ lemma switchLocalFpuOwner_corres[corres]:
...
-  apply (drule tcbFlagToWord_tcbFlagMask)
-  by (clarsimp simp: isFlagSet_def word_to_tcb_flags_def word_bw_lcs intro!: eq_eqI word_bw_comms)
-
 lemma lazyFpuRestore_corres[corres]:
   "corres dc (pspace_aligned and pspace_distinct and valid_cur_fpu and tcb_at t) \<top>
              (lazy_fpu_restore t) (lazyFpuRestore t)"
   unfolding lazy_fpu_restore_def lazyFpuRestore_def
   by (corres corres: threadGet_corres[where r="\<lambda>flags flags'. flags = word_to_tcb_flags flags'"]
-          term_simp: tcb_relation_def isFlagSet_set
+          term_simp: tcb_relation_def Kernel_Config.config_HAVE_FPU_def
```
Xaphiosis, line 160: is this future-proof? we're unfolding a kernel config; I guess the function only exists when there is an FPU?

proof/refine/TcbFlags_R.thy
```diff
@@ -0,0 +1,96 @@
...
+  apply (frule ex_tcbFlagToWord)
+  apply clarsimp
+  apply (rule_tac x=flag in exI)
+  apply (clarsimp simp: not_nth_is_and_neq_0[symmetric] word_bw_assocs[symmetric] word_bw_comms)
+  done
+
+lemma isFlagSet_in_word_to_tcb_flags[simp]:
+  "flag \<in> word_to_tcb_flags tcbFlagMask \<Longrightarrow> isFlagSet flag flags = (flag \<in> word_to_tcb_flags flags)"
+  apply (drule tcbFlagToWord_tcbFlagMask)
+  by (clarsimp simp: isFlagSet_def word_to_tcb_flags_def word_bw_lcs intro!: eq_eqI word_bw_comms)
```
Xaphiosis: two line proof -> by

proof/refine/TcbFlags_R.thy
```diff
@@ -0,0 +1,96 @@
...
+
+lemma word_to_tcb_flags_tcbFlagMask[simp]:
+  "word_to_tcb_flags flags \<inter> word_to_tcb_flags tcbFlagMask = word_to_tcb_flags flags"
+  apply (clarsimp simp: word_to_tcb_flags_def intro!: set_eqI)
+  apply (cut_tac flag=x in tcbFlagToWord_bit)
+  apply (fastforce simp: word_eq_iff)
+  done
+
+lemma not_nth_is_and_neq_0:
+  "(\<not> (x::'a::len word) !! n) = (x AND 2 ^ n = 0)"
```
Xaphiosis: why `neq_0`? surely it's `eq_0`?
also: WordLib candidate?

proof/refine/TcbFlags_R.thy
```diff
@@ -0,0 +1,96 @@
...
+
+lemma ex_tcbFlagToWord:
+  "tcbFlagMask !! n \<Longrightarrow> \<exists>flag. tcbFlagToWord flag = bit n"
+  by (auto simp: tcbFlagToWord_def tcbFlagMask_def split: tcb_flag.splits if_splits)
+
+lemma tcbFlagToWord_tcbFlagMask:
+  "flag \<in> word_to_tcb_flags tcbFlagMask \<Longrightarrow> tcbFlagToWord flag && tcbFlagMask = tcbFlagToWord flag"
+  by (cut_tac flag=flag in tcbFlagToWord_bit) (fastforce simp: word_to_tcb_flags_def word_eq_iff)
+
+
```
Xaphiosis, line 25: intentional extra space?

proof/refine/TcbFlags_R.thy
```diff
@@ -0,0 +1,96 @@
...
+(* Lemmas connecting how the design and abstract specifications represent TCB flags.
+   That is, bitwise with a word in the design spec and as a set in the abstract spec.
+*)
+
+theory TcbFlags_R
+imports BaseRefine.Include Move_R
+begin
+
+lemma tcbFlagToWord_bit:
+  "\<exists>n<word_bits. tcbFlagToWord flag = bit n"
```
Xaphiosis, line 15: ex_tcbFlagToWord_bit?

proof/refine/TcbFlags_R.thy
```diff
@@ -0,0 +1,96 @@
+(*
+ * Copyright 2025, Proofcraft Pty Ltd
+ *
+ * SPDX-License-Identifier: GPL-2.0-only
+ *)
+
+(* Lemmas connecting how the design and abstract specifications represent TCB flags.
+   That is, bitwise with a word in the design spec and as a set in the abstract spec.
+*)
```
Xaphiosis: suggest migrate end of comment to prev line

proof/refine/TcbFlags_R.thy
```diff
@@ -0,0 +1,96 @@
...
+
+lemma tcbFlagToWord_bit:
+  "\<exists>n<word_bits. tcbFlagToWord flag = bit n"
+  by (auto simp: tcbFlagToWord_def ex_nat_less_eq word_bits_conv split: tcb_flag.splits simp del: bit_def)
+
+lemma ex_tcbFlagToWord:
+  "tcbFlagMask !! n \<Longrightarrow> \<exists>flag. tcbFlagToWord flag = bit n"
+  by (auto simp: tcbFlagToWord_def tcbFlagMask_def split: tcb_flag.splits if_splits)
+
+lemma tcbFlagToWord_tcbFlagMask:
```
Xaphiosis: tcbFlagToWord_and_tcbFlagMask ? I again forget what the naming policy is for idempotency rules... eq? id? idem? @lsf37 ?

proof/refine/InvariantsPre_H.thy
```diff
@@ -15,6 +15,7 @@ imports
   "AInvs.ArchDetSchedSchedule_AI"
   "Lib.Heap_List"
   Move_R
+  TcbFlags_R
```
Xaphiosis, line 17: Your TcbFlags_R already pulls in Move_R, making the Move_R above unnecessary.

proof/refine/TcbFlags_R.thy
```diff
@@ -0,0 +1,96 @@
...
+ * Copyright 2025, Proofcraft Pty Ltd
+ *
+ * SPDX-License-Identifier: GPL-2.0-only
+ *)
+
+(* Lemmas connecting how the design and abstract specifications represent TCB flags.
+   That is, bitwise with a word in the design spec and as a set in the abstract spec.
+*)
+
+theory TcbFlags_R
```
Xaphiosis, line 10: This extraction into a separate generic theory looks good to me.

proof/refine/X64/Schedule_R.thy
corlewis: Yeah, currently this only exists when there is an FPU. When we make the FPU properly configurable for the proofs then I think it would be a precondition for this lemma and we wouldn't be unfolding it. Should I add a comment saying that?

proof/refine/TcbFlags_R.thy
corlewis: Oh yeah, this lemma. It's a tweaked version of an existing lemma from `Word_Lib/More_Word.thy`
```
lemma nth_is_and_neq_0:
  "bit (x::'a::len word) n = (x AND 2 ^ n ≠ 0)"
  ```
Looking at it again, you're right that my version should be `eq_0`. What I'm less confident in is if it is generally useful or if I just needed it because I was giving up on fighting with these word proofs. Happy to move it if we do think that it's useful.

proof/refine/TcbFlags_R.thy
corlewis: It's a remnant of when I was thinking that I would give this file sections. That first part was going to be for lemmas specifically about `tcbFlagToWord` while the next was for `word_to_tcb_flags`. I don't think that there's enough going on here for that to be needed so I'll remove it.

proof/refine/X64/Schedule_R.thy
Xaphiosis: Yeah, something brief to that effect would be good.

proof/refine/TcbFlags_R.thy
Xaphiosis: It can probably go next to `nth_is_and_neq_0` then? Doesn't have to be in this PR, I think at this point I'm fine with a `(* FIXME move after nth_is_and_neq_0 *)` unless you feel like a lib update

proof/refine/TcbFlags_R.thy
corlewis: Renamed to `tcbFlagToWord_and_tcbFlagMask_eq` as separately discussed.

proof/refine/TcbFlags_R.thy
corlewis: I'll just move it now, I'm already making changes and it's not like adding it there will break anything.

Xaphiosis:
All my comments appear to be addressed. Thank you!

PR #897, opened 2025-06-05
Prove `activateThread_ccorres` et al.

michaelmcinerney:


lsf37 (approved):
👍  No complaints :-)

spec/haskell/src/SEL4/API/Syscall.lhs
```diff
@@ -121,9 +121,9 @@ System call events are dispatched here to the appropriate system call handlers,
 Interrupt handling is performed by "handleInterrupt", defined in \autoref{sec:object.interrupt.kernel.handling}.
 
 > handleEvent Interrupt = withoutPreemption $ do
->     active <- doMachineOp (getActiveIRQ False)
 >     updateTimeStamp
 >     checkBudget
+>     active <- doMachineOp (getActiveIRQ False)
```
Xaphiosis, line 126: what changed? was there an error in the spec, or was something else involved in order to require this change?

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -21173,27 +21173,26 @@ lemma handle_event_valid_sched:
...
+                           and current_time_bounded" in hoare_strengthen_post[rotated])
+    apply (clarsimp simp: if_split)
+    apply (intro conjI allI impI)
+          apply fastforce
+         apply (fastforce intro: invs_cur_sc_chargeableE)
+        apply (fastforce elim: ct_in_state_weaken)
+       apply fastforce
+      apply (fastforce intro: invs_cur_sc_chargeableE)
+     apply (fastforce elim: ct_in_state_weaken)
+    apply fastforce
```
Xaphiosis, line 21194: this looks highly compressible, but it also looks like it's not originally yours, so fine to leave

spec/haskell/src/SEL4/API/Syscall.lhs
michaelmcinerney: The Haskell was just written very poorly. It should have been in the order it is now.

spec/haskell/src/SEL4/Kernel/Thread.lhs
```diff
@@ -80,6 +80,8 @@ The "activateThread" function is used to prepare a thread to run. If the thread
 
 > activateThread :: Kernel ()
 > activateThread = do
+>         stateAssert ct_activatable'_asrt
+>             "Assert that `ct_in_state' activatable' s` holds"
```
Xaphiosis, line 84: same redundancy comment as in other PR: `stateAssert X` means "assert that X holds"

michaelmcinerney:
> Not a fan of commits like that say `rt spec+proof: prove some top-level ccorres rules for handling events` changing specs without any commentary as to why, especially since it's a reorder as opposed to adding an assert like the other commit (which also doesn't say what's changed in the haskell spec). At some point in the future, looking at these it'll be very hard to differentiate spec changes from assert insertions etc.

Minor reorderings like this are quite common. I could say in the commit message the usual "lines have been rearranged in the Haskell to align with the C", if that's preferable. And I think I add asserts in almost every PR for CRefine, always for the same reason: to provide information in a `ccorres` rule without having to prove further Hoare triples in CRefine. Should I mention that too?

michaelmcinerney:
Force pushed with new commit messages; the proofs are the same.

PR #898, opened 2025-06-05
Prove `doReplyTransfer_ccorres`

michaelmcinerney:
Please see the commit message for a rundown on this. Due to some rules being moved and some rules being adjacent, it would be quite difficult to separate the work into more fine-grained commits, but I can try this if this is strongly preferred.

Test with https://github.com/seL4/seL4/pull/1475


proof/crefine/RISCV64/Ipc_C.thy
```diff
@@ -266,6 +266,14 @@ lemma length_syscallMessage:
   apply (simp add: fromEnum_def enum_register)
   done
 
+lemma length_timeoutMessage:
+  "length RISCV64_H.timeoutMessage = unat n_timeoutMessage"
+  apply (simp add: timeoutMessage_def RISCV64.timeoutMessage_def
+                msgRegisters_unfold n_timeoutMessage_def)
```
lsf37: indent looks off in the 2nd line

proof/crefine/RISCV64/Ipc_C.thy
```diff
@@ -312,6 +322,18 @@ lemma syscallMessage_ccorres:
...
+lemma timeoutMessage_ccorres:
+  "n < unat n_timeoutMessage
+   \<Longrightarrow> register_from_H (RISCV64_H.timeoutMessage ! n) = index timeoutMessageC n"
+  apply (simp add: RISCV64_H.timeoutMessage_def timeoutMessageC_def RISCV64.timeoutMessage_def
+                   MessageID_Exception_def MessageID_TimeoutReply_def
+                   n_timeoutMessage_def msgRegisters_unfold)
+  apply (simp add: upto_enum_def fromEnum_def enum_register)
+  apply (simp add: toEnum_def enum_register)
+  apply (clarsimp simp: fupdate_def
+              | drule nat_less_cases' | erule disjE)+
```
lsf37: Indentation is off here. Probably all fits on one line.

proof/crefine/RISCV64/Ipc_C.thy
```diff
@@ -4014,6 +4150,209 @@ lemma copyMRsFaultReply_ccorres_syscall:
...
+  shows "ccorres dc xfdc
+           (valid_pspace' and tcb_at' s
+                          and tcb_at' r
+                          and obj_at' (\<lambda>t. tcbFault t = Some f) r
+                          and K (r \<noteq> s)
+                          and K (len \<le> unat n_timeoutMessage))
+           (UNIV \<inter> \<lbrace>\<acute>sender = tcb_ptr_to_ctcb_ptr s\<rbrace>
+                 \<inter> \<lbrace>\<acute>receiver = tcb_ptr_to_ctcb_ptr r\<rbrace>
+                 \<inter> \<lbrace>\<acute>id___anonymous_enum = MessageID_TimeoutReply \<rbrace>
+                 \<inter> \<lbrace>\<acute>length___unsigned_long = of_nat len \<rbrace>) hs
```
lsf37: Possibly moved, but the indentation doesn't fit any current style. If we're touching it anyway, we should fix it.

proof/crefine/RISCV64/Ipc_C.thy
```diff
@@ -4014,6 +4150,209 @@ lemma copyMRsFaultReply_ccorres_syscall:
...
+                                      (drop (unat (scast n_msgRegisters :: machine_word))
+                                            (take len RISCV64_H.timeoutMessage))
+                                 od)
+           (Call copyMRsFaultReply_'proc)"
+  proof -
+  let ?obj_at_ft = "obj_at' (\<lambda>tcb. tcbFault tcb = Some f) s"
+  note symb_exec_r_fault = ccorres_symb_exec_r_known_rv_UNIV
+          [where xf'=ret__unsigned_' and R="?obj_at_ft" and R'=UNIV]
+  note empty_fail_cond[simp]
+  show ?thesis
```
lsf37: Can't tell from the GitHub view, but either the `proof -` lines is indented too much, or the rest is not indented by 2 when it should be.

proof/crefine/RISCV64/Syscall_C.thy
```diff
@@ -94,9 +94,592 @@ lemma performInvocation_Notification_ccorres:
...
+  done
+
+lemma validTimeoutHandler_ccorres:
+  "ccorres (\<lambda>rv rv'. rv = to_bool rv') ret__unsigned_long_'
+     (tcb_at' tptr) \<lbrace>\<acute>tptr = tcb_ptr_to_ctcb_ptr tptr\<rbrace> hs
+     (isValidTimeoutHandler tptr) (Call validTimeoutHandler_'proc)"
+  apply (cinit lift: tptr_')
+   apply (rule ccorres_stateAssert)
+   apply (simp add: getThreadTimeoutHandlerSlot_def locateSlot_conv cte_C_size)
+   apply ccorres_remove_UNIV_guard
```
lsf37, line 282: I had forgotten this existed..

proof/crefine/RISCV64/Ipc_C.thy
lsf37: (We can also leave out the `UNIV` these days.)

michaelmcinerney: I forgot to mention that a lot of the work here for the timeout faults is almost perfectly ripped off from the corresponding lemmas for "syscall", for example `copyMRsFaultReply_ccorres_timeout` is very similar to `copyMRsFaultReply_ccorres_syscall`. I couldn't find a nice way to generalise all this work, but maybe it is possible.

lsf37: I don't think there is much we can generalise on the C proof level there, even if the shape of the proof is very similar.

lsf37 (approved):
Happy with the fixes.

spec/haskell/src/SEL4/Object/Endpoint.lhs
```diff
@@ -47,10 +47,12 @@ This function performs an IPC send operation, given a pointer to the sending thr
 The normal (blocking) version of the send operation will remove a recipient from the endpoint's queue if one is available, or otherwise add the sender to the queue.
 
 > sendIPC blocking call badge canGrant canGrantReply canDonate thread epptr = do
+>         stateAssert invs'_asrt "assert that `invs'` holds"
```
Xaphiosis, line 50: still not a fan of saying `stateAssert ... "assert that ... holds"` due to the redundancy, but I guess it fits with the other ones

spec/haskell/src/SEL4/Kernel/Thread.lhs
```diff
@@ -201,6 +201,11 @@ Replies sent by the "Reply" and "ReplyRecv" system calls can either be normal IP
 
 > doReplyTransfer :: PPtr TCB -> PPtr Reply -> Bool -> Kernel ()
 > doReplyTransfer sender reply grant = do
+>     stateAssert sym_refs_asrt
+>         "Assert that `sym_refs (state_refs_of' s)` holds"
+>     stateAssert cur_tcb'_asrt
+>         "Assert that `cur_tcb' s` holds"
+>     stateAssert weak_sch_act_wf_asrt "assert that `weak_sch_act_wf` holds"
```
Xaphiosis, line 208: can't we save space, fit each of these on one line by saying, for example:
```
stateAssert sym_refs_asrt "`sym_refs (state_refs_of' s)`"
```
which says the same thing? `stateAssert X` is literally "assert that X holds"

(same for other files/functions)

proof/crefine/RISCV64/Ipc_C.thy
```diff
@@ -312,6 +319,17 @@ lemma syscallMessage_ccorres:
               | drule nat_less_cases' | erule disjE)+
   done
 
+lemma timeoutMessage_ccorres:
```
Xaphiosis, line 322: the `syscallMessage_ccorres` already exists on master with a silly name, given it's not a ccorres lemma but related to from_H only. given that we're not likely to rename it, I'm still unsure if consistency with it for a new lemma is better than giving it a proper name. @lsf37 ?

proof/crefine/RISCV64/Ipc_C.thy
```diff
@@ -1069,7 +1087,10 @@ lemma setMR_ccorres:
   notes if_cong[cong]
   notes unat_of_nat32 = unat_of_nat_eq[where 'a=32, unfolded word_bits_len_of]
   shows
-  "ccorres (\<lambda>r r'. r = unat (r' && mask msgLengthBits)) ret__unsigned_'
+  "ccorres
+     (\<lambda>r r'. r = unat (r' && mask msgLengthBits)
+                 \<and> (offset < length msgRegisters \<longrightarrow> r' = word_of_nat offset + 1))
```
Xaphiosis: looks like incorrect indentation, the second conjunct goes under the `r =`, otherwise it looks like `r = (X /\ Y)`

proof/crefine/RISCV64/Ipc_C.thy
```diff
@@ -4142,7 +4475,17 @@ lemma handleFaultReply_ccorres [corres]:
             apply (vcg exspec=copyMRsFaultReply_modifies)
            apply (wpsimp wp: threadGet_wp)+
          (* Timeout *)
-         subgoal sorry (* FIXME RT: Timeout case for handleFaultReply_ccorres *)
+         apply (rename_tac number)
```
Xaphiosis: number?

proof/crefine/RISCV64/Syscall_C.thy
```diff
@@ -94,9 +94,592 @@ lemma performInvocation_Notification_ccorres:
...
+  and pspace_distinct'[wp]: pspace_distinct'
+  and pspace_bounded'[wp]: pspace_bounded'
+  and no_0_obj'[wp]: no_0_obj'
+  (wp: crunch_wps)
+
+lemma tcbFault_update_valid_objs'[wp]:
+  "threadSet (tcbFault_update f) tcbPtr \<lbrace>valid_objs'\<rbrace>"
+  by (wpsimp wp: threadSet_valid_objs')
+
+lemma doReplyTransfer_ccorres:
```
Xaphiosis, line 307: is this proof largely derived from an existing one, or mostly new?

proof/crefine/RISCV64/Syscall_C.thy
```diff
@@ -94,9 +94,592 @@ lemma performInvocation_Notification_ccorres:
...
+                apply (frule (1) obj_at_cslift_tcb)
+                apply (frule (1) tcb_ko_at_valid_objs_valid_tcb')
+                apply (clarsimp simp: typ_heap_simps ctcb_relation_def option_to_ctcb_ptr_def
+                                      valid_tcb'_def
+                               split: if_splits)
+                apply (case_tac "tcbSchedContext tcb"; clarsimp)
+               apply ceqv
+              apply (rule_tac P="scOpt' = None" in ccorres_cases)
+               apply ccorres_rewrite
+               apply (rule ccorres_cond_false)
```
Xaphiosis, line 542: can't really follow what's going on here

Xaphiosis (approved):
Minor nitpicks only. Commit message is fine

spec/haskell/src/SEL4/Kernel/Thread.lhs
michaelmcinerney: I could do that. Would you like me to do that for all of the `stateAssert`s?

proof/crefine/RISCV64/Syscall_C.thy
michaelmcinerney: I can't remember all that well, but I'm pretty sure it's almost all new. Maybe a few bits copy pasted from the old proof.

proof/crefine/RISCV64/Ipc_C.thy
michaelmcinerney: I think this is copy paste from the previous case. I guess it should be `badge`, but I can check this.

Xaphiosis: If it's copy/paste then fine to leave as is

proof/crefine/RISCV64/Syscall_C.thy
Xaphiosis: Ah. I was having much difficulty following it, but parts seemed familiar. Maybe a few comments marking the branches would help?

spec/haskell/src/SEL4/Kernel/Thread.lhs
Xaphiosis: I would very much appreciate it, yes. However, a caveat, I don't want to generate too much noise in this and the other PR, so if a function already has a bunch of these in the old style and you're adding one more, leave it in the old style. But for new functions, functions that don't have anything in the old style, of blocks like this, definitely.

michaelmcinerney: These lines were copied and pasted from other functions, since these asserts are old ones. So if I changed these but left the others, it becomes a bit out of sync. I could always merge these PRs with the current version of these `stateAssert`s and then make a separate PR that changes all of them, if people would like that.

proof/crefine/RISCV64/Ipc_C.thy
lsf37: I think the name is correct. It doesn’t need to be a corres predicate, it just needs to be a correspondence between Haskell and C, which it is.

lsf37: I do agree that with ccorres it does look more out of place than it does in Refine, but I think the principle is still there.

proof/crefine/RISCV64/Ipc_C.thy
michaelmcinerney: Turns out the green variable wasn't even used, so I've removed this `rename_tac`

michaelmcinerney:
Force pushed, since the previous target branch was merged

PR #899, opened 2025-06-11
arch-split: InvariantUpdates_H and StateRelation

Xaphiosis:
Please review on a commit-by-commit basis. The general scheme is I move the arch specific files in one commit, then do the arch split in the next commit, and if updates are more complex/interesting, they end up with more commits.

Also note: some of the `FIXME arch-split` comments are interesting for this review, but should probably be chucked out or converted to normal comments when we actually want to merge this.

Interesting things, mostly for StateRelation:

Similarly to invariants, the state relation contains a large number of
definitions. To reduce load on locales and number of theory files, the
architecture-specific definitions go first (ArchStateRelation), then the
generic definitions and lemmas (StateRelation), and finally the
arch-specific lemmas (ArchStateRelationLemmas).

- due to ghost_relation being arch-specific on AARCH64, ghost_relation
  becomes arch-specific, with a generic `ghost_relation_wrapper`
  exported into generic context
- ghost_relation is required to not depend on machine state
- the [a]obj_relation_cuts infrastructure conflated arch and generic
  object cases in other_obj_relation and those had to be split up
- obj_relation_cutsE unfortunately still spills all the cases, but
  defining that properly can only happen when dealing with KHeap_R


Xaphiosis:
Currently broken due to conflicts with master, need to rebase and fix.

proof/refine/ARM/ArchInvariantUpdates_H.thy
```diff
@@ -5,482 +5,23 @@
...
-  "ct_not_inQ (s\<lparr>ksSchedulerAction := SwitchToThread t\<rparr>)"
-   by (simp add: ct_not_inQ_def)
-
-lemma invs'_update_cnt[elim!]:
-  "invs' s \<Longrightarrow> invs' (s\<lparr>ksSchedulerAction := ChooseNewThread\<rparr>)"
-   by (clarsimp simp: invs'_def valid_state'_def valid_irq_node'_def cur_tcb'_def
-                      ct_idle_or_in_cur_domain'_def tcb_in_cur_domain'_def bitmapQ_defs)
+global_interpretation InvariantUpdates_H?: InvariantUpdates_H
+  proof goal_cases
+  interpret Arch .
```
lsf37: We should not perpetuate that style, the indentation was never correct. No need to update existing ones (I'll do a global one at some point), but new interpretations shouldn't use it.

proof/refine/InvariantUpdates_H.thy
```diff
@@ -0,0 +1,457 @@
...
+(* generic consequences which require arch-specific proofs from ArchInvsLemmas_H *)
+arch_requalify_facts
+  cte_wp_atE'
+  cte_wp_at_cteI'
+  tcb_at_cte_at'
+  typ_at_lift_valid_cap'
+  valid_arch_tcb_lift'
+  tcb_at_cte_at'
+  gen_objBitsT_simps
+  objBitsT_koTypeOf (* FIXME arch-split: consider declaring [simp] here rather than Schedule_R *)
```
lsf37, line 20: That might be nice to do, but I'd leave it for later, so that the fallout from `[simp]` isn't mixed in with the rest. I.e. leave in the FIXME for now.

proof/refine/InvariantUpdates_H.thy
```diff
@@ -0,0 +1,457 @@
...
+lemma ct_not_inQ_ksMachineState_update[simp]:
+  "ct_not_inQ (ksMachineState_update f s) = ct_not_inQ s"
+  by (simp add: ct_not_inQ_def)
+
+lemma ct_in_current_domain_ksMachineState[simp]:
+  "ct_idle_or_in_cur_domain' (ksMachineState_update p s) = ct_idle_or_in_cur_domain' s"
+  by (simp add: ct_idle_or_in_cur_domain'_def tcb_in_cur_domain'_def)
+
+(* FIXME arch-split: export? *)
+lemma invs'_machine:
```
lsf37: I don't think I understand the comment. We're in a generic file at the top level, so where would we export to?

proof/refine/AARCH64/ArchStateRelation.thy
```diff
@@ -6,53 +6,30 @@
...
-context begin interpretation Arch .
-
-definition cte_map :: "cslot_ptr \<Rightarrow> machine_word" where
-  "cte_map \<equiv> \<lambda>(oref, cref). oref + (of_bl cref << cte_level_bits)"
-
-lemmas cte_map_def' = cte_map_def[simplified cte_level_bits_def shiftl_t2n mult_ac, simplified]
+(* FIXME arch-split: these are generic, but due to many definitions, the Arch theory goes before
+   the generic one, so there's nowhere sensible to move to *)
+type_synonym obj_relation_cut = "Structures_A.kernel_object \<Rightarrow> Structures_H.kernel_object \<Rightarrow> bool"
+type_synonym obj_relation_cuts = "(machine_word \<times> obj_relation_cut) set"
```
lsf37: A small `_pre` theory?

proof/refine/AARCH64/ArchStateRelation.thy
```diff
@@ -289,133 +122,22 @@ definition is_other_obj_relation_type :: "a_type \<Rightarrow> bool" where
...
-
-definition cdt_list_relation :: "cdt_list \<Rightarrow> cdt \<Rightarrow> cte_heap \<Rightarrow> bool" where
- "cdt_list_relation \<equiv> \<lambda>t m m'.
-    \<forall>c cap node. m' (cte_map c) = Some (CTE cap node)
-        \<longrightarrow> (case next_slot c t m of None \<Rightarrow> True
-                                   | Some next \<Rightarrow> mdbNext node = cte_map next)"
+(* FIXME arch-split: provided only so that ghost_relation can be used within generic definition
+   of state_relation (since arch_state_relation doesn't have access to kheap, and
+   gsPTTypes on AARCH64 isn't generic) *)
+definition ghost_relation_wrapper :: "det_state \<Rightarrow> kernel_state \<Rightarrow> bool" where
```
lsf37, line 129: do we want it to be `det_state`? it looks like it would be fine the generic state

proof/refine/AARCH64/ArchStateRelation.thy
```diff
@@ -427,78 +149,9 @@ definition arch_state_relation :: "(arch_state \<times> AARCH64_H.kernel_state)
...
-| "syscall_error_map (ExceptionTypes_A.FailedLookup b lf)   = Fault_H.FailedLookup b (lookup_failure_map lf)"
-| "syscall_error_map ExceptionTypes_A.TruncatedMessage      = Fault_H.TruncatedMessage"
-| "syscall_error_map ExceptionTypes_A.DeleteFirst           = Fault_H.DeleteFirst"
-| "syscall_error_map ExceptionTypes_A.RevokeFirst           = Fault_H.RevokeFirst"
-| "syscall_error_map (ExceptionTypes_A.NotEnoughMemory n)   = Fault_H.syscall_error.NotEnoughMemory n"
-
+(* FIXME arch-split: object_type is arch-specific with one generic field, inverse of usual setup,
+   but if we split off the generic part, it would have to go before arch definitions, which would
+   require more locales or an extra generic theory file *)
 definition APIType_map :: "Structures_A.apiobject_type \<Rightarrow> AARCH64_H.object_type" where
```
lsf37: I wouldn't be opposed to a small extra theory if it makes the definitions nicer. We don't want that everywhere in the proofs, but for invariants and state relation I'd say it's worth it.

proof/refine/AARCH64/ArchStateRelationLemmas.thy
```diff
@@ -0,0 +1,176 @@
...
+    | AArch (ADeviceData sz) \<Rightarrow> {(x + (n << pageBits), \<lambda>_ obj. obj = KOUserDataDevice )
+                                 | n . n < 2 ^ (pageBitsForSize sz - pageBits) }
+    | AArch _ \<Rightarrow> {(x, other_aobj_relation)}
+    | AGarbage _ \<Rightarrow> {(x, \<bottom>\<bottom>)}
+    | _ \<Rightarrow> {(x, other_obj_relation)})"
+  by (simp add: obj_relation_cuts_def2 a_type_def well_formed_cnode_n_def length_set_helper
+           split: Structures_A.kernel_object.split AARCH64_A.arch_kernel_obj.split)
+
+(* FIXME arch-split: split off generic? *)
+lemma obj_relation_cutsE:
```
lsf37, line 54: This is one of the more fundamental questions. If we have case distinction lemmas like this in here, that means that all lemmas that use them must be arch specific. That's not necessarily a problem, but it is a potential point for optimisation later.

I'd vote for leaving the FIXME in for now and see how it goes with just the arch-specific version of it. If it draws in a lot of later lemmas that we really would like to be generic, then we should come back here and split it up into generic cases and arch cases. Doing it immediately feels like premature optimisation.

proof/refine/AARCH64/ArchStateRelationLemmas.thy
```diff
@@ -0,0 +1,176 @@
...
+                     AARCH64_A.arch_kernel_obj.splits)
+  done
+
+(* FIXME arch-split: export? *)
+lemma is_other_obj_relation_type_CapTable:
+  "\<not> is_other_obj_relation_type (ACapTable n)"
+  by (simp add: is_other_obj_relation_type_def)
+
+(* FIXME arch-split: export? *)
+lemma is_other_obj_relation_type_TCB:
```
lsf37: Yes, those look both worthy of export

proof/refine/StateRelation.thy
```diff
@@ -0,0 +1,444 @@
...
+  used to model the ready queues, and so an update to such a field would correspond to an update
+  to a ready queue (see ready_queues_relation below).\<close>
+definition
+  other_obj_relation :: "Structures_A.kernel_object \<Rightarrow> Structures_H.kernel_object \<Rightarrow> bool"
+where
+  "other_obj_relation obj obj' \<equiv>
+   (case (obj, obj') of
+      (Endpoint ep, KOEndpoint ep') \<Rightarrow> ep_relation ep ep'
+    | (Notification ntfn, KONotification ntfn') \<Rightarrow> ntfn_relation ntfn ntfn'
+    | _ \<Rightarrow> False)"
```
lsf37: Was probably there before, but the outer parentheses around `case` are not needed (only for `=`, not for `==`)

proof/refine/ARM/ArchInvariantUpdates_H.thy
Xaphiosis: Will do. When you do that, don't forget to update arch-split.md

- [x] reindent global_interpretation

proof/refine/InvariantUpdates_H.thy
Xaphiosis: Agreed, I think it'll also be easier to fix the fallout when things are more arch-split. Currently having to go do a fix 5 times is annoying.

proof/refine/InvariantUpdates_H.thy
Xaphiosis: I think it's a leftover from when it was in the Arch locale, forgot to remove comment when I moved it.

- [x] remove `FIXME arch-split: export?` comments here and for `invs_no_cicd'_machine`

proof/refine/AARCH64/ArchStateRelation.thy
Xaphiosis: I tried to avoid creating yet another pre theory given it only needed a few lines. I'm mindful that pretty much every theory is going to become two, and Invariants_H already became 4 theories, so to have this one be 4 as well... if you think that's an improvement, I'll do it.

proof/refine/AARCH64/ArchStateRelation.thy
Xaphiosis: I think I blindly copied it from ghost_relation, which might've gotten updated in the meantime. Will check.

proof/refine/AARCH64/ArchStateRelationLemmas.thy
Xaphiosis: Yes, as you can tell from the updates, this one was a huge pain and initial attempts to make it nicer did not work. I have ideas, but they can only be properly tested when we get to how these are used. It's a very strange formulation of something I already don't quite understand (the obj_cuts framework).

proof/refine/AARCH64/ArchStateRelation.thy
lsf37: I think so. Not in general, of course, but for StateRelation it's worth it.

proof/refine/AARCH64/ArchStateRelation.thy
Xaphiosis: Turns out on master all `state_relation` takes `det_state` already, and that's where I got it from. After discussion, we both decided that indicates no need for generalisation here.

proof/refine/AARCH64/ArchStateRelationLemmas.thy
Xaphiosis: - [x] export these two lemmas

proof/refine/StateRelation.thy
Xaphiosis: done

proof/refine/AARCH64/ArchStateRelation.thy
Xaphiosis: After successfully arch-splitting this, the hilarious result was my discovering that this is unused, so I removed it. It turns out in Retype_R someone introduced `APIType_map2` that uses a sum type, and that's the only one that sees use. Sigh.

proof/refine/AARCH64/ArchStateRelation.thy
Xaphiosis: this is done now

Xaphiosis:
Hopefully addressed all comments as separate commits. Please take another look while the regression tests run and make sure I didn't accidentally mess something up.

proof/refine/ARM/ArchInvariantUpdates_H.thy
```diff
@@ -5,482 +5,23 @@
...
-  done
-
-(* FIXME arch-split: MOVE *)
-context Arch begin arch_global_naming
-
-lemma valid_arch_state'_interrupt[simp]:
+lemma valid_arch_state'_interrupt[simp, InvariantUpdates_H_assms]:
   "valid_arch_state' (ksInterruptState_update f s) = valid_arch_state' s"
-  by (simp add: valid_arch_state'_def)
+  by (simp add: valid_arch_state'_def cong: option.case_cong)
```
corlewis: Out of curiosity, any idea why this now needs that cong?

corlewis: Oh, is it to make it consistent with the other architectures?

proof/refine/InvariantUpdates_H.thy
```diff
@@ -0,0 +1,457 @@
...
+                ct_not_inQ_def ct_idle_or_in_cur_domain'_def tcb_in_cur_domain'_def
+                bitmapQ_defs)
+
+lemma if_unsafe_then_cap_arch'[simp]:
+  "if_unsafe_then_cap' (ksArchState_update f s) = if_unsafe_then_cap' s"
+  by (simp add: if_unsafe_then_cap'_def ex_cte_cap_to'_def)
+
+lemma valid_bitmaps_ksSchedulerAction_update[simp]:
+   "valid_bitmaps (ksSchedulerAction_update f s) = valid_bitmaps s"
+   by (simp add: valid_bitmaps_def bitmapQ_defs)
```
corlewis: Do we care about fixing style while moving these lemmas? If we do, this is overindented.

proof/refine/AARCH64/ArchStateRelation.thy
corlewis: Do we want to rename `APIType_map2` and move it here (or to a generic file if it can be split)? That could probably be done in a separate PR, except that we would lose track of this location and the connection to the definition that you're removing.

proof/refine/AARCH64/ArchStateRelation.thy
```diff
@@ -0,0 +1,149 @@
...
+definition
+  other_aobj_relation :: "Structures_A.kernel_object \<Rightarrow> Structures_H.kernel_object \<Rightarrow> bool"
+where
+  "other_aobj_relation obj obj' \<equiv>
+   (case (obj, obj') of
+      (ArchObj (AARCH64_A.ASIDPool ap), KOArch (KOASIDPool ap')) \<Rightarrow> asid_pool_relation ap ap'
+    | (ArchObj (AARCH64_A.VCPU vcpu), KOArch (KOVCPU vcpu')) \<Rightarrow> vcpu_relation vcpu vcpu'
+    | _ \<Rightarrow> False)"
+
+(* FIXME arch-split: previously, aobj_relation_cuts used other_obj_relation *)
```
corlewis: I don't understand the intent of this comment. What do you want to do here and how is `aobj_relation_cuts` having changed relevant?

proof/refine/AARCH64/ArchStateRelation.thy
lsf37: Agree, I would do that when we hit Retype_R.

proof/refine/RISCV64/ArchStateRelation.thy
```diff
@@ -0,0 +1,115 @@
...
+
+theory ArchStateRelation
+imports StateRelationPre
+begin
+
+(* FIXME arch-split: these are generic, but due to many definitions, the Arch theory goes before
+   the generic one, so there's nowhere sensible to move to *)
+type_synonym obj_relation_cut = "Structures_A.kernel_object \<Rightarrow> Structures_H.kernel_object \<Rightarrow> bool"
+type_synonym obj_relation_cuts = "(machine_word \<times> obj_relation_cut) set"
+
```
lsf37: These look still left over

proof/refine/X64/ArchStateRelation.thy
```diff
@@ -0,0 +1,202 @@
...
+
+(* FIXME arch-split: these are generic, but due to many definitions, the Arch theory goes before
+   the generic one, so there's nowhere sensible to move to *)
+text \<open>
+  A pspace and a tree are related if every object in the pspace
+  corresponds to an object in the tree. Some abstract objects
+  like CapTables correspond to multiple concrete ones, thus we
+  have to make cuts.\<close>
+type_synonym obj_relation_cut = "Structures_A.kernel_object \<Rightarrow> Structures_H.kernel_object \<Rightarrow> bool"
+type_synonym obj_relation_cuts = "(machine_word \<times> obj_relation_cut) set"
```
lsf37: also here

proof/refine/ARM_HYP/ArchStateRelation.thy
```diff
@@ -0,0 +1,198 @@
...
+
+(* FIXME arch-split: these are generic, but due to many definitions, the Arch theory goes before
+   the generic one, so there's nowhere sensible to move to *)
+text \<open>
+  A pspace and a tree are related if every object in the pspace
+  corresponds to an object in the tree. Some abstract objects
+  like CapTables correspond to multiple concrete ones, thus we
+  have to make cuts.\<close>
+type_synonym obj_relation_cut = "Structures_A.kernel_object \<Rightarrow> Structures_H.kernel_object \<Rightarrow> bool"
+type_synonym obj_relation_cuts = "(machine_word \<times> obj_relation_cut) set"
```
lsf37: and here :-)

proof/refine/ARM/ArchStateRelation.thy
```diff
@@ -0,0 +1,174 @@
...
+
+(* FIXME arch-split: these are generic, but due to many definitions, the Arch theory goes before
+   the generic one, so there's nowhere sensible to move to *)
+text \<open>
+  A pspace and a tree are related if every object in the pspace
+  corresponds to an object in the tree. Some abstract objects
+  like CapTables correspond to multiple concrete ones, thus we
+  have to make cuts.\<close>
+type_synonym obj_relation_cut = "Structures_A.kernel_object \<Rightarrow> Structures_H.kernel_object \<Rightarrow> bool"
+type_synonym obj_relation_cuts = "(machine_word \<times> obj_relation_cut) set"
```
lsf37: one more

lsf37 (approved):
Looks good, but you still need to remove the StateRelationPre definitions from the other architectures. StateRelationPre is now smaller than we had planned, but I think it's still Ok to have it. We might find more things in the future that we want to put there.

proof/refine/StateRelation.thy
```diff
@@ -0,0 +1,444 @@
...
+lemma pspace_dom_relatedE:
+  "\<lbrakk> s' x = Some ko'; pspace_relation s s';
+     \<And>y ko P. \<lbrakk> s y = Some ko; (x, P) \<in> obj_relation_cuts ko y; P ko ko' \<rbrakk> \<Longrightarrow> R \<rbrakk> \<Longrightarrow> R"
+  apply (rule pspace_dom_revE [OF in_related_pspace_dom]; assumption?)
+  apply (fastforce dest: pspace_relation_absD)
+  done
+
+lemma eq_trans_helper:
+  "\<lbrakk> x = y; P y = Q \<rbrakk> \<Longrightarrow> P x = Q"
+  by simp
```
corlewis, line 434: I guess this is a generic lemma that is used in the various `ArchStateRelationLemmas` files so it makes sense here, but it does feel a bit unexpected and out of place. Maybe a comment about how it's used?

proof/refine/AARCH64/CSpace1_R.thy
```diff
@@ -6659,6 +6659,7 @@ lemma cteSwap_corres:
          (cap_swap scap src dcap dest) (cteSwap scap' src' dcap' dest')"
   (is "corres _ ?P ?P' _ _") using assms including no_pre
   supply None_upd_eq[simp del]
+  supply AARCH64.ghost_relation_wrapper_def[simp] (* FIXME arch-split *)
```
corlewis, line 6662: Is this and all the other uses of `ghost_relation_wrapper_def` still needed? I thought you declared it `simp` in each of the `ArchStateRelation` files.

proof/refine/ARM/ArchInvariantUpdates_H.thy
Xaphiosis: Yeah, I think this one was for consistency between arches.

proof/refine/AARCH64/ArchStateRelation.thy
Xaphiosis: I think I should remove this comment. It was more a pointer for the review that I changed the way this worked: previously all the cases were shoved into other_obj_relation, so it was used in both the generic and the arch relation_cuts, so I had to restructure that for arch-split.

Xaphiosis: Right, and I did remove it on all arches but AARCH64. Sigh.

proof/refine/RISCV64/ArchStateRelation.thy
Xaphiosis: Well spotted, thanks. Removed it for AARCH64, but forgot the rest somehow.

proof/refine/StateRelation.thy
Xaphiosis: Would a comment help much, or should we dump it into lib somewhere do you think? (I can add a FIXME move for this PR too)

proof/refine/AARCH64/CSpace1_R.thy
Xaphiosis: Will review again. Some of them are definitely needed because someone already did some of these proofs outside of Arch locale. Thanks for the reminder.

Xaphiosis: Reviewed. There was a spurious one in ArchAcc on AARCH64, but otherwise:
* CSpace1_R.thy - needed in cteSwap_corres (not in Arch)
* Retype_R.thy - currently needed for state_relationE expansion
* Detype_R.thy - currently needed for state_relationE expansion
* Untyped_R.thy - needed for updateFreeIndex_corres (not in Arch)

proof/refine/RISCV64/ArchStateRelation.thy
Xaphiosis: These should all be done now.

proof/refine/AARCH64/ArchStateRelation.thy
Xaphiosis: Removed the comment

proof/refine/InvariantUpdates_H.thy
Xaphiosis: Fixed, thanks.

proof/refine/AARCH64/ArchStateRelation.thy
Xaphiosis: When we get to Retype_R, we'll be forced to arch-split APIType_map2, which would be a good point to rename it. Given the grief it gave me today, I'm not likely to forget it.

proof/refine/StateRelation.thy
Xaphiosis: added comment: `(* used for generating architecture-specific cap_relation split lemmas *)` for now, let me know what you think

Xaphiosis:
Addressed @corlewis's review comments and the derp @lsf37 pointed out. Hopefully that's all of them, let me know if there's anything else.

proof/refine/StateRelation.thy
corlewis: I think that's helpful enough, thanks. It's a bit of a weird lemma and for such a specific use-case that I don't think it should go into lib.

lsf37:
Good to go from my side!

PR #900, opened 2025-06-13
tests: make haskell-translator depend on kernel-config

corlewis:
This protects against both tests being run simultaneously and trying to write to the same file.


lsf37 (approved):
Thanks for finding that.

Xaphiosis (approved):
Nicely spotted!

PR #901, opened 2025-06-16
Improve messages for `stateAssert`s

michaelmcinerney:
Please see the commit message.


spec/haskell/src/SEL4/Object/TCB.lhs
```diff
@@ -1110,23 +1103,20 @@ On some architectures, the thread context may include registers that may be modi
...
->     stateAssert tcbQueueHead_ksReleaseQueue_active_sc_tcb_at'_asrt
->         "if the release queue is not empty, then the head of the release queue is associated with an active scheduling context"
+>     stateAssert tcbQueueHead_ksReleaseQueue_active_sc_tcb_at'_asrt "if the release queue is not empty, then the head of the release queue is associated with an active scheduling context"
 
 > awaken :: Kernel ()
 > awaken = do
 >     stateAssert ksReleaseQueue_asrt ""
->     stateAssert tcbInReleaseQueue_imp_active_sc_tcb_at'_asrt
->         "every thread in the release queue is associated with an active scheduling context"
+>     stateAssert tcbInReleaseQueue_imp_active_sc_tcb_at'_asrt "every thread in the release queue is associated with an active scheduling context"
```
Xaphiosis: I don't remember the exact limit we agreed on for haskell files, but are you sure this line isn't too long? (same goes for related ones that ended up as a very long line)

spec/haskell/src/SEL4/Object/TCB.lhs
```diff
@@ -1014,12 +1011,10 @@ On some architectures, the thread context may include registers that may be modi
 
 > chargeBudget :: Ticks -> Bool -> Kernel ()
 > chargeBudget consumed canTimeoutFault = do
->     stateAssert cur_tcb'_asrt
->         "Assert that `cur_tcb' s` holds"
->     stateAssert sch_act_wf_asrt
->         "Assert that `sch_act_wf (ksSchedulerAction s) s` holds"
+>     stateAssert cur_tcb'_asrt "`cur_tcb'`"
+>     stateAssert sch_act_wf_asrt "`sch_act_wf (ksSchedulerAction s) s`"
```
Xaphiosis, line 1015: I'm wondering if we get much benefit from the backticks, considering they're a concept for markdown. I'm fine leaving them in, but thought I'd raise it. @lsf37 ?

Xaphiosis (approved):
Some concern about line length on a few of these, but overall much better, thank you.

spec/haskell/src/SEL4/Object/TCB.lhs
michaelmcinerney: Yeah I wasn’t sure if we make some sort of document out of the Haskell. I’m used to using them on Mattermost and I kind of like how it separates terms we use within Isabelle from regular words

spec/haskell/src/SEL4/Object/TCB.lhs
michaelmcinerney: I thought Haskell had no limits on the number of characters. We have lots of comments in the literate Haskell files that are comically long.

spec/haskell/src/SEL4/Object/TCB.lhs
Xaphiosis: I guess we can adapt the practice. We already use it a bit in Isabelle `(* *)` comments for example, so that words don't get confused with terms (since `@{term }` is so long-winded and more for text<> and \<comment> comments), but we're very irregular about it. It's more when there's nothing to differentiate, like here, that it looks odd.

spec/haskell/src/SEL4/Object/TCB.lhs
Xaphiosis: The language doesn't impose a limit, but usually there's standards of some kind to stop runaway (by default it's 80 chars or so). I guess they weren't enforced often enough, so I'm sure people have been very lax about them.
I looked at the multi-line string situation in Haskell, and while there looks to be a decent way for the normal language (see: [How can I write multiline strings in Haskell? - Stack Overflow](https://stackoverflow.com/questions/22918837/how-can-i-write-multiline-strings-in-haskell) ), I don't know if it'll work well in literate haskell.

So my proposal would be: for these really long ones (>90 chars) do it like you did before, i.e. put the text on the subsequent line and then we hope that's not too long. Vertical compression only yields benefits when stuff doesn't line wrap anyway.

spec/haskell/src/SEL4/Object/TCB.lhs
lsf37: I'm easy. We used to generate LaTeX from literate Haskell, but that is long broken and not scheduled to be resurrected. I do use the backtick from time to time in non-markdown contexts, but I do agree that they are not a Haskell thing in general. But it's inside a string, I guess we can do whatever we like there.

PR #902, opened 2025-06-19
Prove `decodeSchedControl_configureFlags_ccorres` and `decodeInvocation_ccorres`

michaelmcinerney:


proof/invariant-abstract/Invariants_AI.thy
```diff
@@ -25,6 +25,7 @@ arch_requalify_facts (A)
   reply_bits_def
   parse_time_arg_def
   words_from_time_def
+  TIME_ARG_SIZE_def
```
Xaphiosis: Is this a good idea, given we want to do arch-split of MCS in the future? I guess it's consistent with the other ones, but when doing arch-split I'd rather keep the definition in Arch and then break it with RISCV64.TIME_ARG_SIZE_def in the places it's needed (and with a `(* FIXME arch-split *)` at the places where it's egregious.

I guess given how the rest are already leaking, now might not be the time, but the direction this goes in will cause friction later in a way that's annoying to detect. You'd have to notice that arch stuff is being leaked through the Arch interface after being unable to reconcile proofs on two arches, then go back to this point, investigate what this is, remove the requalify and do more rounds.

proof/crefine/RISCV64/SchedContext_C.thy
```diff
@@ -1819,6 +1819,400 @@ lemma invokeSchedControl_ConfigureFlags_ccorres:
...
+  apply (clarsimp simp: ccorres_underlying_def)
+  apply (erule (1) my_BallE)
+  apply clarsimp
+  apply (erule allE, erule allE, erule (1) impE)
+  apply (clarsimp simp: return_def unif_rrel_def split: xstate.splits)
+  done
+
+lemma refillAbsoluteMax_size_helper:
+  "\<lbrakk>valid_cap' cap s; isSchedContextCap cap\<rbrakk>
+   \<Longrightarrow> MIN_REFILLS \<le> refillAbsoluteMax cap \<and> refillAbsoluteMax cap < 2 ^ LENGTH(64)"
```
Xaphiosis: word_size_len?

proof/crefine/RISCV64/SchedContext_C.thy
```diff
@@ -1819,6 +1819,400 @@ lemma invokeSchedControl_ConfigureFlags_ccorres:
...
+lemma decodeSchedControl_ConfigureFlags_ccorres:
+  "interpret_excaps extraCaps' = excaps_map extraCaps \<Longrightarrow>
+   ccorres (intr_and_se_rel \<currency> dc)  (liftxf errstate id (K ()) ret__unsigned_long_')
+     (\<lambda>s. (invs' s \<and> sch_act_wf (ksSchedulerAction s) s
+           \<and> ksCurThread s = thread \<and> ct_active' s
+           \<and> (excaps_in_mem extraCaps (ctes_of s))
+           \<and> (\<forall>v \<in> set extraCaps. s \<turnstile>' fst v)
+           \<and> (\<forall>v \<in> set extraCaps. \<forall>y \<in> zobj_refs' (fst v). ex_nonz_cap_to' y s)
+           \<and> sysargs_rel args buffer s)
+          \<and> isSchedControlCap cp)
```
Xaphiosis, line 1879: indent looks off (on github at least)

proof/crefine/RISCV64/SchedContext_C.thy
```diff
@@ -1819,6 +1819,400 @@ lemma invokeSchedControl_ConfigureFlags_ccorres:
...
+     (\<lbrace>\<acute>current_extra_caps = extraCaps'\<rbrace>
+      \<inter> \<lbrace>unat \<acute>length___unsigned_long = length args\<rbrace>
+      \<inter> \<lbrace>\<acute>buffer = option_to_ptr buffer\<rbrace>
+      \<inter> \<lbrace>ccap_relation cp \<acute>cap\<rbrace>) hs
+     (decodeSchedControl_ConfigureFlags (map fst extraCaps) args
+      >>= invocationCatch thread isBlocking isCall canDonate InvokeSchedControl)
+     (Call decodeSchedControl_ConfigureFlags_'proc)"
+  apply (intro ccorres_gen_asm[simplified pred_conj_def])
+  supply Collect_const[simp del] if_cong[cong] option.case_cong[cong]
+         sched_context_C_size[simp del] refill_C_size[simp del]
```
Xaphiosis: how come you need to hide the two `_size`s?

proof/crefine/RISCV64/SchedContext_C.thy
```diff
@@ -1819,6 +1819,400 @@ lemma invokeSchedControl_ConfigureFlags_ccorres:
...
+    apply (rule us_to_ticks_mono)
+     apply (clarsimp simp: maxPeriodUs_def word_le_nat_alt word_less_nat_alt)
+    apply (force intro: mult_le_mono3[where a="5 * unat MAX_PERIOD_US"])
+   apply (clarsimp simp: minBudgetUs_def MIN_BUDGET_def2)
+   apply (rule us_to_ticks_mono)
+    apply fastforce
+   apply (clarsimp simp: maxPeriodUs_def word_less_nat_alt)
+   apply (force intro: mult_le_mono3[where a="5 * unat MAX_PERIOD_US"])
+  apply (clarsimp simp: excaps_map_def neq_Nil_conv cte_wp_at_ctes_of dest!: interpret_excaps_eq)
+  done
```
Xaphiosis, line 2182: usual comment: this proof is very hard to review without any guiding comments and not seeing the intermittent proof state myself

proof/crefine/RISCV64/Syscall_C.thy
```diff
@@ -707,11 +707,20 @@ lemma performInvocation_Reply_ccorres:
   apply (simp add: rf_sr_ksCurThread)
   done
 
+lemma ct_in_state'_cur_tcb':
+  "ct_in_state' st s \<Longrightarrow> cur_tcb' s"
+  by (clarsimp simp: ct_in_state'_def cur_tcb'_def)
+
+lemma ct_in_state'_rewrite:
+  "ct_in_state' st s = st_tcb_at' st (ksCurThread s) s"
+  by (clarsimp simp: ct_in_state'_def cur_tcb'_def)
```
Xaphiosis: do we want to move such things to SR_lemmas? @lsf37

proof/crefine/RISCV64/Syscall_C.thy
```diff
@@ -899,41 +921,80 @@ sorry (* FIXME RT: decodeInvocation_ccorres
...
+      apply (simp add: throwError_bind invocationCatch_def)
+      apply (rule syscall_error_throwError_ccorres_n)
+      apply (simp add: syscall_error_to_H_cases)
+     apply (simp only: liftME_invocationCatch comp_def id_def)
+     apply (rule ccorres_call, rule decodeSchedControlInvocation_ccorres[where cp=cp])
+        apply assumption
+       apply (simp+)[3]
+    apply vcg
+   apply (rule ccorres_Cond_rhs)
+    apply (simp add: if_to_top_of_bind)
```
Xaphiosis, line 933: seeing a lot of `if_to_top_of_bind` ... are there multiple ifs in here, or is something dodgy with automation?

Xaphiosis (approved):
Good progress. Minor comments on content.

Have an issue with the commit "rt spec+crefine: prove decodeSchedControl_ConfigureFlags_ccorres": it's doing too much. Doing a multi-arch restructure like that and making something arch-specific should be in its own commit. This will be particularly important when dealing with arch-split in future.

The `decode_sched_control_configure_flags` update can indeed go in with the proof commit.

proof/crefine/RISCV64/Syscall_C.thy
lsf37: In fact, isn't this just `ct_in_state'_def`?

In Invariants_H:
```
 "ct_in_state' test \<equiv> \<lambda>s. st_tcb_at' test (ksCurThread s) s"
```

`ct_in_state'_cur_tcb'` should go into `Invariants_H` I think.

proof/crefine/RISCV64/SchedContext_C.thy
lsf37: yes, please

proof/crefine/RISCV64/Syscall_C.thy
lsf37: It's a pattern in some of the decode proofs. I don't remember exactly where anymore (maybe here), but I remember doing the same for a few of these, mostly adjusting existing ones, to avoid blowup of `if` cases.

proof/invariant-abstract/Invariants_AI.thy
lsf37: Agreed, if we can we should not export the `_def`. Ideally we'd only really need it in generic theories in CRefine, and only some conditions exported in the rest. Not sure if that ideal case is easily achievable, but Raf's suggestion of using the fully qualified name with a FIXME should at least help us to find them later.

spec/abstract/ARM/Arch_Structs_A.thy
```diff
@@ -330,6 +330,9 @@ definition parse_time_arg :: "nat \<Rightarrow> data list \<Rightarrow> time" wh
 definition words_from_time :: "time \<Rightarrow> machine_word list" where
   "words_from_time t = [ucast t, ucast (t >> 32)]"
 
+definition TIME_ARG_SIZE :: nat where
+  "TIME_ARG_SIZE \<equiv> 2" (* sizeof(ticks_t) / sizeof(word_t) *)
```
lsf37: That comment is exciting, because it would actually make that definition generic. If you define it as what the comment says, only the C version of this will need to be arch specific. Not sure it does that much since `words_from_time` still will need to be arch specific, but you could add a comment with `FIXME arch-split` that this could be a generic definition so we remember when we get there.

lsf37 (approved):
Nice. Only some arch split and lemma moving comments, content is good.

proof/crefine/RISCV64/SchedContext_C.thy
michaelmcinerney: This grouping is done intentionally so that I can do a `ccorres_gen_asm` at the beginning of the proof

proof/crefine/RISCV64/SchedContext_C.thy
michaelmcinerney: I was actually able to remove these from this proof, but normally I do hide these two so that I don't get a whole bunch of magic numbers popping up in the proof, making it hard to see what's going on and making it hard to apply some rules that are stated in terms of `size_of sched_context_C/size_of refill_C`

proof/crefine/RISCV64/SchedContext_C.thy
michaelmcinerney: I have added some comments to point out which `whenE` statement we are at during the proof. These follow the Haskell function very closely, so I'm not sure they will be of any use to someone who has to work on the proof. The biggest difficulty in this proof is working with the size of the scheduling context, and this is all done within `ccorres_symb_exec_r_known_rv` rules, so that's hopefully clear already.

proof/crefine/RISCV64/Syscall_C.thy
michaelmcinerney: I have modified the proof a little bit so that I don't need to use this rewrite rule, and I've removed it, but it's a little interesting, so I'll still ask. 

So my rewrite rule is almost the same as `ct_in_state'_def`, except that `ct_in_state'_def` is eta-contracted. I have had a look at a bunch of attributes, but I can't see one that will transform the definition into the form I was wanting. Is there a way to do that? I could just write the definition of `ct_in_state'` without the eta contraction (this is also done plenty in `Invariants_H`), but that seems a bit over the top, and we'd probably want to do the same on `master`, which is too time-consuming.

proof/refine/RISCV64/SchedContextInv_R.thy
```diff
@@ -344,28 +348,62 @@ lemma decode_sc_inv_corres:
   apply (clarsimp simp: sc_relation_def)
   done
 
+lemma parseTimeArg_parse_time_arg:
+  "parseTimeArg arg n = parse_time_arg arg n"
+  by (clarsimp simp: parseTimeArg_def parse_time_arg_def)
+
+lemma wordBits_word_bits:
+  "wordBits = word_bits"
+  by (simp add: wordBits_def' word_bits_def)
```
lsf37: Wow, I can't believe we have not got this lemma yet (and I checked we indeed do not). It should go high up somewhere where both are in scope, though, at least `Bits_R`.

spec/abstract/Structures_A.thy
```diff
@@ -527,6 +527,9 @@ record sched_context =
   sc_replies    :: "obj_ref list"
   sc_sporadic   :: "bool"
 
+definition TIME_ARG_SIZE :: nat where
+  "TIME_ARG_SIZE \<equiv> LENGTH(ticks_len) div word_bits"
```
lsf37, line 543: Cool, this is nice.

proof/crefine/RISCV64/SchedContext_C.thy
lsf37: The new comments are helpful, it gives you some structure without having to read the entire proof (or function) when something breaks.

proof/crefine/RISCV64/Syscall_C.thy
lsf37: I think there is a way, let me check. The question is, though: why do you want it expanded? Usually the contracted form works everywhere the expanded form does (unless you want to explicitly instantiate `s`).

lsf37: `ct_in_state'_def[unfolded atomize_eq, THEN fun_cong]` gets you `ct_in_state'_rewrite`. Not the prettiest, though.

lsf37: If you do need to keep it, the right name would be something with extensionality, e.g. `ct_in_state'_ext` or `ct_in_state'_arg`.

michaelmcinerney: I don't need this rule any more, but before, I had some kind of long Boolean expression with the state written in as usual (so `st_tcb_at' P t s /\ ...`) and I wanted for some reason to get a conjunct like that into a `ct_in_state' P s` form, so I had this rewrite rule used in a `simp flip: `. If this crops up again I can try the attributes you used or just write out a new lemma with a name like that. Thanks.

proof/refine/RISCV64/SchedContextInv_R.thy
michaelmcinerney: Strangely enough, I couldn't see any other lemmas about `wordBis` in `Bits_R`, but that does sound like a good place for it, and I've put it at the very top of that file.

proof/crefine/RISCV64/Syscall_C.thy
lsf37: Right, the reverse direction might not always work. Good to keep in mind for the future. So far I thought shifting the lambda to the right is strictly more applicable, but that's only true in the left-to-right right direction.

michaelmcinerney:
Force pushed after rebasing over `rt` (including the merges from `master`)

PR #903, opened 2025-06-20
Add `det_wp_whileLoop`

michaelmcinerney:


lib/DetWPLib.thy
```diff
@@ -124,4 +124,67 @@ lemma det_wp_assert_opt :
...
+  "det_wp P f = (\<forall>s. P s \<longrightarrow> (\<not> snd (f s)) \<and> (\<exists>r. r \<in> fst (f s) \<and> (\<forall>r'. r' \<in> fst (f s) \<longrightarrow> r' = r)))"
+  apply (clarsimp simp: det_wp_def)
+  apply (subst singleton_iff[symmetric])
+  apply (intro iffI conjI impI allI)
+    apply fastforce
+   apply fastforce
+  apply clarsimp
+  apply (drule_tac x=s in spec)
+  apply clarsimp
+  apply (rule_tac x=r in exI)
```
lsf37: these two `rule_tac` need a `rename_tac` before to make the names `s` and `r` stable

lib/DetWPLib.thy
```diff
@@ -124,4 +124,67 @@ lemma det_wp_assert_opt :
...
+     apply (simp add: whileLoop_cond_fail return_def)
+    apply clarsimp
+    apply (elim impE)
+      apply (fastforce elim: use_valid[where f="B _ "] intro: inv)
+     apply (clarsimp simp: whileLoop_def)
+     apply (rule whileLoop_result_det_wp_intermediate)
+         apply (clarsimp simp: det_wp_def2)
+         apply force+
+       using det[THEN det_wp_no_fail] det
+       apply (fastforce simp: no_fail_def det_wp_def2)
```
lsf37: This `using` adds `det` twice. Do we need the one without instantiation?

lsf37 (approved):
Cool. Shorter than I feared :)

lib/DetWPLib.thy
michaelmcinerney: Yeah I wasn't sure about this sort of case: here Isabelle will generate green variables with the same names that I used in the statement of the lemma, right? Do we still need to `rename_tac` this sort of thing?

corlewis: While it would probably always be fine in this case, the name of those green variables depends on the context and can change unexpectedly, while using `rename_tac` makes it (almost) completely stable. We're not particularly consistent with this in the main proofs but it's more important in `lib` and is a requirement for anything that might be released to the AFP.

corlewis (approved):

lib/DetWPLib.thy
lsf37: Yes, they are not guaranteed. It would be fairly stable here, but the Isabelle2025 update for instance merged the name space for constants and free variables, changing a lot of names that previously were thought stable. So the policy is to always explicitly name green variables somewhere.

lib/DetWPLib.thy
michaelmcinerney: Actually, the instantiated assumption can be removed, so I've done that

lsf37 (approved):
Good to squash and merge!

Issue #687, opened 2023-10-23
Investigate adding `projectKOs` to the simp set


michaelmcinerney:
`projectKOs` is in the simp set on RISCV64 and AARCH64, but not on the other architectures. We should investigate standardising this and adding it to the simp set on all architectures.

lsf37:
In particular also between `rt` and `master`

Xaphiosis:
For master, this will come into effect with https://github.com/seL4/l4v/pull/904 (though cleanup will still be needed due to all the warnings from `simp: projectKOs`)

PR #904, opened 2025-06-25
Arch split Corres.thy and Bits_R.thy

Xaphiosis:
Once again, please review commit-by-commit. I will want to ultimately squash the refine and crefine updates together, and also squash the hierarchy update into the Bits_R arch-split (smaller chunks make review easier).

For Corres.thy, there's nothing interesting, I picked the one that had the most stuff in it (ARM) and removed the rest.

For Bits_R, there were some adventures needed:
* `projectKOs` has a generic and arch-specific part; since RISCV64 and AARCH64 expect both to be [simp], and the other arches expect neither to be [simp], there was no good way to reconcile. After discussion, projectKOs exports both arch and generic lemmas into the global context, and is  now [simp] on all architectures
* `isCap_simps` shows up a lot, but it had to be arch-split into `gen_` and `arch_ `; there are some places where I had to use the arch version in a "generic" proof, which will have to be untangled further down the arch-split line

Addresses https://github.com/seL4/l4v/issues/687 but will need a projectKOs warning cleanup pass on arches beside RISCV64 and AARCH64.

Removes `1658` lines of proof.


proof/refine/Bits_R.thy
```diff
@@ -9,4 +9,430 @@ theory Bits_R
 imports Corres ArchStateRelationLemmas
 begin
 
+crunch_ignore (add:
+  bind withoutFailure throw catchFailure rethrowFailure capFaultOnFailure lookupErrorOnFailure
+  nullCapOnFailure nothingOnFailure withoutPreemption preemptionPoint maskInterrupt unifyFailure
+  ignoreFailure emptyOnFailure clearMemoryVM  assertDerived
+  setObject getObject updateObject loadObject)
+
+(* FIXME arch-split: RISCV64 has bigger set than AARCH64 *)
```
corlewis: Is this a FIXME that we want to think about before merging this PR, or one to come back to later?

proof/refine/AARCH64/ArchBits_R.thy
```diff
@@ -108,357 +51,17 @@ lemma projectKO_VCPU:
...
-  apply simp
-  done
+end
 
-lemma set_ep_arch':  "\<lbrace>\<lambda>s. P (ksArchState s)\<rbrace> setEndpoint ntfn p \<lbrace>\<lambda>_ s. P (ksArchState s)\<rbrace>"
-  apply (simp add: setEndpoint_def setObject_def split_def)
-  apply (wp updateObject_default_inv|simp)+
-  done
+(* for projectKO_opt, we want to export the arch-specific instantiation lemmas *)
+arch_requalify_facts arch_projectKOs
```
corlewis, line 62: Should this be in a generic file instead, so that it isn't duplicated in each ArchBits_R and can't get out of sync?

proof/refine/ARM/PageTableDuplicates.thy
```diff
@@ -40,28 +41,12 @@ lemma set_ntfn_valid_duplicate' [wp]:
   apply (clarsimp simp:updateObject_default_def assert_def bind_def
     alignCheck_def in_monad when_def alignError_def magnitudeCheck_def
     assert_opt_def return_def fail_def split:if_splits option.splits)
-   apply (rule_tac ko = ba in valid_duplicates'_non_pd_pt_I)
+   apply (rule_tac ko ="KONotification a" in valid_duplicates'_non_pd_pt_I)
        apply simp+
-  apply (rule_tac ko = ba in valid_duplicates'_non_pd_pt_I)
+  apply (rule_tac ko ="KONotification a" in valid_duplicates'_non_pd_pt_I)
```
corlewis: While you're touching these, the spaces before the `=` should be removed as well.

proof/refine/ARM_HYP/CSpace1_R.thy
```diff
@@ -4723,12 +4725,14 @@ lemma mdb_inv_preserve_updateCap:
...
-  apply (clarsimp simp:mdb_inv_preserve_def modify_map_dom isCap_simps modify_map_def split:if_splits)
+  apply (clarsimp simp: mdb_inv_preserve_def modify_map_dom gen_isCap_simps modify_map_def
+                  split: if_splits)
   apply (intro conjI impI allI)
-    apply fastforce
-    apply (simp add:sameRegionAs_update_untyped)
-    apply (rule ext,simp add:sameRegionAs_update_untyped')
-    apply (simp add:mdb_next_def split:if_splits)
+      apply fastforce
+     apply (simp add: ARM_HYP.isCap_defs)
```
corlewis: Out of curiosity, any reason why this is `isCap_defs` instead of `isCap_simps` like all the other cases? And also, should it be marked `FIXME arch-split` like those other cases?

corlewis (approved):
Very nice progress and it's great to see `projectKOs` standardised!

proof/refine/Bits_R.thy
Xaphiosis: Ah, I missed that. Yes, I should consolidate these, thanks!

proof/refine/ARM/PageTableDuplicates.thy
Xaphiosis: done, will appear when I push

proof/refine/AARCH64/ArchBits_R.thy
Xaphiosis: I can't put it earlier since I can't phrase an interface lemma due to terms like `isFrameCap` not being available outside of Arch. So it has to go after. We can indeed put this in the next generic file in the tree, but ArchBits_R is imported by both Invocations_R and EmptyFail. I actually thought of you when I put this here, because you pointed out in the past that my putting a requalify in a semi-related unrelated file was a bit confusing... so a completely unrelated one would be even less maintainable.

I think the current method is a reasonable compromise, but I agree that it's duplication and has minor de-sync potential.

proof/refine/ARM_HYP/CSpace1_R.thy
Xaphiosis: I have not a clue. To give context, defs look like this:
```
isZombie ?v ≡ case ?v of capability.Zombie v0 v1 v2 ⇒ True | _ ⇒ False
```
while simps look like this:
```
isZombie ?v = (∃v0 v1 v2. ?v = capability.Zombie v0 v1 v2)
```

And the FIXME... yeah, for a moment I thought this file was in CRefine. Since it's in Refine, it should get one. Will do.

proof/refine/AARCH64/ArchBits_R.thy
corlewis: Ah, yeah, it's a tricky balance between reducing duplication and having things in locations that make sense. I agree with what you've done being a good compromise, it feels unlikely that this will change at some point and de-sync.

proof/refine/Bits_R.thy
Xaphiosis: Ok, I think I've consolidated these. I'm getting two warnings left: one is for clearMemory, which indeed is currently only crunch_ignored on some arches, but the other one was for `throw` whose `crunch_ignore` I can't seem to locate. So I've dropped `throw` and I'll let regression test decide if it's indeed ignored on all arches.

proof/refine/AARCH64/ArchBits_R.thy
lsf37: I'm also happy with it.

lsf37 (approved):
This is good. Nice to finally get projectKOs consolidated. We can clean up warnings later at our leisure. I'm pleasantly surprised that `gen_isCap_simps` is sufficient in so many places. I had expected the generic cap proofs to be much more accidentally dependent on arch cap case distinctions. Nice that they are not.

proof/refine/ARM_HYP/CSpace1_R.thy
lsf37: If it gets the proof done, either is fine. Agreed with the FIXME.

PR #906, opened 2025-06-26
arch-split: EmptyFail, Invocations_R, Machine_R

Xaphiosis:
Again, review commit-by-commit. This one should be merged after https://github.com/seL4/l4v/pull/904

EmptyFail had some noise resulting from the platform wp sets not being consistent, hence the `empty_fail_getObject` appearance.

Invocations_R turned out to be small and while it did have an architecture-specific proof, it was uninteresting and identical on all platforms. Hence the `(in Arch)` plus requalify, we don't gain much from another file here. One could even argue `Invocations_R` could be entirely absorbed by `StateRelation`.

Machine_R is largely machine and arch-specific stuff, so I cleaned it up and factored out the common parts. Nothing interesting there, so only one commit. Hopefully the diff isn't too bad.


proof/refine/ARM/EmptyFail_H.thy
```diff
@@ -170,7 +170,7 @@ crunch
   cancelIPC, setThreadState, tcbSchedDequeue, setupReplyMaster, isStopped,
   possibleSwitchTo, tcbSchedAppend
   for (empty_fail) empty_fail[intro!, wp, simp]
-  (simp: crunch_simps)
+  (simp: crunch_simps wp: empty_fail_getObject)
```
corlewis: It's not clear to me what's going on with `empty_fail_getObject`, from the diff it doesn't look like anything was different on ARM and ARM_HYP with regards to it being in the `wp` set or not.

proof/refine/AARCH64/ArchEmptyFail.thy
```diff
@@ -4,134 +4,27 @@
...
-  "empty_fail (lookupIPCBuffer a b)"
+lemma empty_fail_lookupIPCBuffer[EmptyFail_R_assms]:
+  "empty_fail (lookupIPCBuffer r t)"
   by (clarsimp simp: lookupIPCBuffer_def Let_def getThreadBufferSlot_def locateSlot_conv
               split: capability.splits arch_capability.splits | wp | wpc | safe)+
 
-lemma empty_fail_updateObject_default [intro!, wp, simp]:
-  "empty_fail (updateObject_default v ko a b c)"
-  by (fastforce simp: updateObject_default_def typeError_def unless_def split: kernel_object.splits )
+declare setRegister_empty_fail[intro!, simp] (* originally from a crunch *)
```
corlewis: What is your intention with this comment? I'm guessing, but would it be more helpful as `(* FIXME: tag original instead *)`, which I've copied from a different comment you added?

I'm also not sure if this `declare` is needed on all architectures, on AARCH64 at least `setRegister_empty_fail` should already be in those sets.

proof/refine/EmptyFail.thy
```diff
@@ -0,0 +1,140 @@
...
+ *)
+
+theory EmptyFail
+imports ArchBits_R
+begin
+
+(* Collect empty_fail lemmas here. naming convention is empty_fail_NAME.
+   Unless there is a good reason, they should all be [intro!, wp, simp] *)
+
+lemma empty_fail_projectKO [simp, intro!]:
```
corlewis: This isn't new in this PR but since the comment is right there, is there any reason that `empty_fail_projectKO` shouldn't be `wp`?

proof/refine/EmptyFail.thy
```diff
@@ -0,0 +1,140 @@
...
+
+lemma empty_fail_threadSet [intro!, wp, simp]:
+  "empty_fail (threadSet f p)"
+  by (fastforce simp: threadSet_def getObject_def setObject_def split_def)
+
+lemma empty_fail_getThreadState[iff]:
+  "empty_fail (getThreadState t)"
+  by (simp add: getThreadState_def)
+
+declare empty_fail_stateAssert [wp]
```
corlewis: This declare is repeated on line 134.

proof/refine/EmptyFail.thy
lsf37: Hm, in a different EmptyFail fail theory somewhere (came up on the SGI PR), we only made them `wp` there, which is probably the better policy. `intro!` is harmless, but having them as `simp` can be dangerous because it can solve parts in the middle of a formula, leaving schematics uninstantiated. It's safe in the logical sense (i.e. it's true), but it can make the automation confused later.

Problem is of course that by now some proofs will be relying on the `simp`, and fixing those shouldn't be part of arch split.

So I guess we should add `wp` here, and either make an issue or add a FIXME to properly clean up attributes for `empty_fail` everywhere.

proof/refine/ARM/EmptyFail_H.thy
lsf37: Would it make sense to try to make `empty_fail_getObject` `[wp]` on all architectures to increase consistency in that direction instead? Or is there too much potential for breakage?

proof/refine/AARCH64/ArchEmptyFail.thy
lsf37: Hm, in a different EmptyFail fail theory somewhere (came up on the SGI PR), we only made them wp there, which is probably the better policy. intro! is harmless, but having them as simp can be dangerous because it can solve parts in the middle of a formula, leaving schematics uninstantiated. It's safe in the logical sense (i.e. it's true), but it can make the automation confused later.

Problem is of course that by now some proofs will be relying on the simp, and fixing those shouldn't be part of arch split.

So I guess we should add wp here, and either make an issue or add a FIXME to properly clean up attributes for empty_fail everywhere.

proof/refine/EmptyFail.thy
lsf37: Hm, in a different EmptyFail fail theory somewhere (came up on the SGI PR), we only made them wp there, which is probably the better policy. intro! is harmless, but having them as simp can be dangerous because it can solve parts in the middle of a formula, leaving schematics uninstantiated. It's safe in the logical sense (i.e. it's true), but it can make the automation confused later.

Problem is of course that by now some proofs will be relying on the simp, and fixing those shouldn't be part of arch split.

So I guess we should add wp here, and either make an issue or add a FIXME to properly clean up attributes for empty_fail everywhere.

proof/refine/ARM/EmptyFail_H.thy
corlewis: My understanding of the current state is that it is already consistent, with `empty_fail_getObject` not being `wp` and variants for specific types like `empty_fail_getObject_tcb` and `empty_fail_getObject_ntfn` being `wp` instead. That's why I'm confused about it being needed to be added later on now.

Xaphiosis: Your interpretation is correct: I think the intended stable version is to have the specific typed versions in wp, but not the generic `empty_fail_getObject`. I think as I was doing the update, I noticed some stuff not working, which was possibly due to `empty_fail_getObject_tcb` missing on some arches. I added that later, but in the meantime fixed proofs that failed due to `empty_fail_getObject_tcb` being removed from the `wp` set on arches that used to have it in the `wp` set (ARM+ARM_HYP). 

I can double-check whether having added the tcb version makes any of these redundant, but I'm also willing to leave that to when the arch-split front line gets to EmptyFail_H. Thoughts?

proof/refine/AARCH64/ArchEmptyFail.thy
Xaphiosis: Yeah, I'm confused on this one. The crunch crunches a bunch of other stuff, and seemingly we only want this one part of it to be `[intro!, simp]` ... but now that you mention it, I could try go back and tag the original crunch and see what breaks. Hmm. But that might cause more rebuilding. Maybe adding a FIXME like you suggested would be the best idea for now.

proof/refine/EmptyFail.thy
Xaphiosis: Should I change the comment to no longer include `simp` in the advisory? Dropping the ones already declared simp sounds like a headache, but updating the policy to a sane one is free.

lsf37: Yes, that sounds good. Follow the policy for now (this specific one is going to be harmless), but add a FIXME comment that we should rethink the policy.

Xaphiosis: I meant changing the comment to `Unless there is a good reason, they should all be [intro!, wp]` instead of `[intro!, wp, simp]` and then not changing any of the lemma tags for now...

proof/refine/ARM/EmptyFail_H.thy
corlewis: If you do double-check, use ARM_HYP as it has more of these. Otherwise, do you think that they will be easy to notice when you reach this point? If not, maybe an "arch-split" comment?

proof/refine/EmptyFail.thy
lsf37: Fine with not changing the tags. I'd make visible that there is/should be a policy change that has not been executed yet. Otherwise we might be confused in the future with the mismatch to reality.

proof/refine/EmptyFail.thy
Xaphiosis: cheers! removed duplication

proof/refine/ARM/EmptyFail_H.thy
Xaphiosis: You are indeed correct, I was able to eliminate these changes (ARM+ARM_HYP).

proof/refine/AARCH64/ArchEmptyFail.thy
Xaphiosis: Changed to `(* FIXME: tag original instead *)`

proof/refine/EmptyFail.thy
Xaphiosis: Removed `[simp]` from policy, added FIXME on policy deployment needing review

Xaphiosis:
Rebased, fixed merge issues, retargeted to master. Hopefully all tests will pass now.

PR #907, opened 2025-07-07
arm+arm-hyp+aarch64 crefine: fix proof for GICv3

lsf37:
The value of gicNumTargets_bits was wrong for GICv3 platforms.


Xaphiosis (approved):
Commit message should be rephrased. This does not fix the proof, it fixes/updates the spec. It also doesn't change crefine, it changes machine/Platform.

lsf37:
> Commit message should be rephrased. This does not fix the proof, it fixes/updates the spec. It also doesn't change crefine, it changes machine/Platform.

You are very correct, my mind was all on the proofs that were breaking, on on the file I was modifying.

PR #908, opened 2025-07-23
arch-split: KHeap_R

Xaphiosis:
Please review commit-by-commit.

X64 has extra stuff in ArchKHeap_R related to cross lemmas, and I've segregated it to show that it's not relevant for the interface needed by KHeap_R, but figuring out where they're supposed to go will happen later.

I tried to carve out a generic section from the obj_cuts situation, but did not manage it. As a result `setObject_other_corres` needs to be requalified.

The amount of lines removed is pretty nice:
`60 files changed, 3561 insertions(+), 11093 deletions(-)` i.e. 7532 lines removed.


Xaphiosis:
Rebased after FPU merge.

proof/refine/AARCH64/ArchKHeap_R.thy
```diff
@@ -0,0 +1,413 @@
...
+lemma setEndpoint_pspace_in_kernel_mappings'[KHeap_R_assms]:
+  "setEndpoint p ko \<lbrace>pspace_in_kernel_mappings'\<rbrace>"
+  by wp
+
+lemma setNotification_pspace_in_kernel_mappings'[KHeap_R_assms]:
+  "setNotification p ko \<lbrace>pspace_in_kernel_mappings'\<rbrace>"
+  by wp
+
+(* FIXME arch-split: possible interface lemma, but can't be done via locale *)
+lemma valid_global_refs_lift':
```
lsf37: This can probably turn into a normal comment -- I don't think there is a way around the free type variable limitation in locales in general.

proof/refine/ARM/ArchKHeap_R.thy
```diff
@@ -0,0 +1,438 @@
...
+lemma setEndpoint_pspace_in_kernel_mappings'[KHeap_R_assms]:
+  "setEndpoint p ko \<lbrace>pspace_in_kernel_mappings'\<rbrace>"
+  by wp
+
+lemma setNotification_pspace_in_kernel_mappings'[KHeap_R_assms]:
+  "setNotification p ko \<lbrace>pspace_in_kernel_mappings'\<rbrace>"
+  by wp
+
+(* FIXME arch-split: possible interface lemma, but can't be done via locale *)
+lemma valid_global_refs_lift':
```
lsf37: (convert from FIXME to normal comment)

proof/refine/KHeap_R.thy
```diff
@@ -9,6 +11,59 @@ imports
...
+arch_requalify_facts
+  valid_arch_obj'_valid
+
+lemmas [wp] = valid_arch_obj'_valid
+
+(* FIXME: this should go somewhere in spec *)
+translations
+  (type) "'a kernel" <=(type) "kernel_state \<Rightarrow> ('a \<times> kernel_state) set \<times> bool"
+
+declare fromPPtr_def[simp] (* FIXME arch-split: move this somewhere high up *)
```
lsf37: KHeap is probably high enough, tbh. I think we can remove the FIXME from this one.

proof/refine/KHeap_R.thy
```diff
@@ -9,6 +11,59 @@ imports
...
+  "(1::machine_word) < 2 ^ ntfnSizeBits"
+  by (auto simp: objBits_simps')
+
+requalify_facts Arch.zero_one_less_2p_SizeBits
+declare zero_one_less_2p_SizeBits[simp]
+
+(* FIXME arch-split: true on all arches, move *)
+lemma zero_less_tcbBlockSizeBits[simp]:
+  "0 < tcbBlockSizeBits"
+  unfolding tcbBlockSizeBits_def by simp
```
lsf37: Is this tagged "move" because it's in between (in Arch) lemmas? if yes, we should probably do the move now, the theory KHeap does otherwise look to be the right place for it (maybe Bits_R, but if it's not needed higher up, KHeap is high enough).

proof/refine/KHeap_R.thy
```diff
@@ -230,26 +242,25 @@ lemma obj_at_setObject1:
   apply (rule Q)
   done
 
+(* variant which can handle \<not> (obj_at' ...) *)
 lemma obj_at_setObject2:
```
lsf37, line 205: Nice that some of these can stay generic! I had not dared to hope :-)

proof/refine/AARCH64/ArchKHeap_R.thy
```diff
@@ -0,0 +1,413 @@
...
+  by wp
+
+(* FIXME arch-split: check with other arches what is actually in common, standardise *)
+lemma setEndpoint_pspace_in_kernel_mappings'[KHeap_R_assms]:
+  "setEndpoint p ko \<lbrace>pspace_in_kernel_mappings'\<rbrace>"
+  by wp
+
+lemma setNotification_pspace_in_kernel_mappings'[KHeap_R_assms]:
+  "setNotification p ko \<lbrace>pspace_in_kernel_mappings'\<rbrace>"
+  by wp
```
lsf37: It looks like it's these two lemmas and the FIXME can now be removed (in the other ArchKHeap_R files as well)

lsf37 (approved):
Nice! Lots of generic proof and pleasantly little fallout down the road. Only some FIXME cleanup comments, content is great.

I had actually expected the `setObject` type variable situation to be worse, it's nice that we can keep a whole bunch of them generic.

proof/refine/AARCH64/ArchKHeap_R.thy
```diff
@@ -0,0 +1,413 @@
...
+imports
+  KHeap_R
+begin
+
+context Arch begin arch_global_naming
+
+named_theorems KHeap_R_assms
+
+lemma getObject_inv_vcpu[wp]:
+  "\<lbrace>P\<rbrace> getObject l \<lbrace>\<lambda>(rv :: ArchStructures_H.vcpu). P\<rbrace>"
```
corlewis: Very nitpick but I think the more standard style for this would be
```suggestion
  "\<lbrace>P\<rbrace> getObject l \<lbrace>\<lambda>_::ArchStructures_H.vcpu. P\<rbrace>"
```

proof/refine/AARCH64/ArchKHeap_R.thy
```diff
@@ -1165,835 +301,30 @@ lemma setObject_other_arch_corres:
...
-  for ksInterruptState[wp]: "\<lambda>s. P (ksInterruptState s)"
-  (wp: setObject_ksInterrupt updateObject_default_inv)
-
-lemmas setEndpoint_irq_handlers[wp]
-    = valid_irq_handlers_lift'' [OF set_ep_ctes_of setEndpoint_ksInterruptState]
-
-declare set_ep_arch' [wp]
+(* FIXME arch-split: check with other arches what is actually in common, standardise *)
+lemma setEndpoint_pspace_in_kernel_mappings'[KHeap_R_assms]:
+  "setEndpoint p ko \<lbrace>pspace_in_kernel_mappings'\<rbrace>"
```
corlewis: Should this happen now or is it intended for when more progress is made into Refine?

proof/refine/AARCH64/ArchKHeap_R.thy
```diff
@@ -0,0 +1,413 @@
...
+
+lemma set_ntfn_global_refs'[KHeap_R_assms_2, wp]:
+  "\<lbrace>valid_global_refs'\<rbrace> setNotification ptr val \<lbrace>\<lambda>_. valid_global_refs'\<rbrace>"
+  by (rule valid_global_refs_lift'; wp)
+
+crunch setEndpoint, setNotification
+  for valid_arch'[wp]: valid_arch_state'
+  (wp: valid_arch_state_lift')
+
+lemmas [KHeap_R_assms_2] = setEndpoint_valid_arch' setNotification_valid_arch'
```
corlewis, line 400: Out of curiosity, do you do the `KHeap_R_assms_2` attribute this way to avoid it being applied to other lemmas that crunch produces?

proof/refine/AARCH64/ArchKHeap_R.thy
```diff
@@ -0,0 +1,413 @@
...
+
+lemmas setObject_typ_ats[wp] = typ_at_lifts[OF setObject_typ_at']
+
+lemmas doMachineOp_typ_ats[wp] = typ_at_lifts[OF doMachineOp_typ_at']
+
+lemmas setEndpoint_typ_ats[wp] = typ_at_lifts[OF setEndpoint_typ_at']
```
corlewis, line 406: I just had the thought that a lot of what `typ_at_lifts` is doing is generic. Do we want to split it out so those parts can be done in generic files or would that be making it more complicated for not much gain?

proof/refine/AARCH64/ArchKHeap_R.thy
```diff
@@ -0,0 +1,413 @@
...
+
+end
+
+interpretation KHeap_R_2?: KHeap_R_2
+proof goal_cases
+  interpret Arch .
+  case 1 show ?case by (intro_locales; (unfold_locales; fact KHeap_R_assms_2)?)
+qed
+
+arch_requalify_facts setObject_other_corres (* FIXME arch-split: move *)
```
corlewis: Should this move be done now or is it being left until we've arch-split up to a file where it makes sense to be moved to?

proof/refine/AARCH64/ArchStateRelationLemmas.thy
```diff
@@ -164,6 +164,14 @@ lemma ghost_relation_wrapper_machine_state_upd_id[StateRelation_R_assms,simp]:
    = ghost_relation_wrapper s s'"
   by simp
 
+(* interface lemma, can't be used in locale assumptions due to free type variable *)
+lemma valid_arch_obj'_valid[wp]:
+  assumes P: "\<And>P T p. \<lbrace>\<lambda>s. P (typ_at' T p s)\<rbrace> f \<lbrace>\<lambda>rv s. P (typ_at' T p s)\<rbrace>"
+  notes [wp] = hoare_vcg_all_lift hoare_vcg_imp_lift hoare_vcg_const_Ball_lift typ_at_lifts [OF P]
+  shows "\<lbrace>valid_arch_obj' ako\<rbrace> f \<lbrace>\<lambda>rv. valid_arch_obj' ako\<rbrace>"
```
corlewis: I'm not sure if this comment will still make sense when we've forgotten the context and it's not clear what it's an interface lemma for and what locale it isn't able to be used in.

proof/refine/KHeap_R.thy
```diff
@@ -11,4 +11,2094 @@ imports
...
+          | wp hoare_vcg_split_case_option
+               hoare_drop_imps hoare_vcg_all_lift)+
+  done
+
+lemma getObject_inv:
+  assumes x: "\<And>p q n ko. \<lbrace>P\<rbrace> loadObject p q n ko \<lbrace>\<lambda>(rv :: 'a :: pspace_storable). P\<rbrace>"
+  shows      "\<lbrace>P\<rbrace> getObject p \<lbrace>\<lambda>(rv :: 'a :: pspace_storable). P\<rbrace>"
+  by (simp add: getObject_def split_def | wp x)+
+
+lemma getObject_inv_tcb [wp]: "\<lbrace>P\<rbrace> getObject l \<lbrace>\<lambda>(rv :: Structures_H.tcb). P\<rbrace>"
```
corlewis: Might as well fix up this style while we're touching this.

proof/refine/KHeap_R.thy
```diff
@@ -11,4 +11,2094 @@ imports
...
+
+lemma updateObject_cte_is_tcb_or_cte:
+  fixes cte :: cte and ptr :: machine_word
+  shows "\<lbrakk> fst (lookupAround2 p (ksPSpace s)) = Some (q, ko);
+           snd (lookupAround2 p (ksPSpace s)) = n;
+           (ko', s') \<in> fst (updateObject cte ko p q n s) \<rbrakk> \<Longrightarrow>
+  (\<exists>tcb getF setF. ko = KOTCB tcb \<and> s' = s \<and> tcb_cte_cases (p - q) = Some (getF, setF)
+    \<and> ko' = KOTCB (setF (\<lambda>x. cte) tcb) \<and> is_aligned q tcbBlockSizeBits \<and> ps_clear q tcbBlockSizeBits s) \<or>
+  (\<exists>cte'. ko = KOCTE cte' \<and> ko' = KOCTE cte \<and> s' = s
+        \<and> p = q \<and> is_aligned p cte_level_bits \<and> ps_clear p cte_level_bits s)"
```
corlewis: More style/indentation that it would be nice to fix while we're doing this.

proof/refine/KHeap_R.thy
```diff
@@ -11,4 +11,2094 @@ imports
...
+  "\<lbrace>\<top>\<rbrace> getNotification ep \<lbrace>\<lambda>rv. ko_at' rv ep\<rbrace>"
+  apply (simp add: getNotification_def)
+  apply (rule getObject_ko_at)
+   apply simp
+  apply (simp add: gen_objBits_simps)
+  done
+
+lemmas ps_clear_def3 = ps_clear_def2[OF order_less_imp_le[OF aligned_less_plus_1]]
+
+declare diff_neg_mask[simp del] (* FIXME arch-split: do we want to keep this? *)
```
corlewis: Is this something that we can decide now that KHeap_R is finished?

proof/refine/AARCH64/ArchKHeap_R.thy
corlewis: Oh right, we do have and are using `gen_typ_at_lifts`. It's a bit hard to tell from the diffs though, does `gen_typ_at_lifts` and `typ_at_lifts` overlapping cause any annoyances?

proof/refine/KHeap_R.thy
```diff
@@ -11,4 +11,2094 @@ imports
...
+  apply (drule updateObject_objBitsKO)
+  apply (clarsimp simp: lookupAround2_char1)
+  done
+
+crunch setNotification, setEndpoint
+  for pspace_domain_valid[wp]: "pspace_domain_valid"
+
+(* There are two wp rules for preserving valid_ioc over set_object.
+   First, the more involved rule for CNodes and TCBs *)
+(* Second, the simpler rule suitable for all objects except CNodes and TCBs. *)
```
corlewis: Unless I'm blind, these comments haven't been connected to anything for a very long time. They should probably be removed.

proof/refine/AARCH64/CNodeInv_R.thy
```diff
@@ -4962,6 +4962,26 @@ lemmas tcbSlots =
...
+lemma updateCap_pspace_in_kernel_mappings'[wp]:
+  "updateCap next prev \<lbrace>pspace_in_kernel_mappings'\<rbrace>"
+  by wpsimp
+
+end
+
+(* FIXME arch-split: interface (these are only interesting on ARM/ARM_HYP *)
+arch_requalify_facts
+  updateMDB_pspace_in_kernel_mappings'
+  updateCap_pspace_in_kernel_mappings'
```
corlewis, line 4980: I'm a bit confused by this comment with these requalifies only appearing on `AARCH64`.

corlewis (approved):
Very nice to see so much being made generic!

I pretty quickly stopped making cleanup suggestions on things that were being moved. Feel free to ignore the couple I did make, although now that KHeap is all nice and generic it is going to be pretty tempting to go through it and at least fixup the worst of the style.

proof/refine/AARCH64/CNodeInv_R.thy
Xaphiosis: It's trying to say that this is interesting on ARM/ARM_HYP, but trivial here, hence we crunch+requalify. I guess judging by your reaction this is not very useful...

proof/refine/KHeap_R.thy
Xaphiosis: Sure, the problem is I have no idea what the decision should be. My instinct would be to remove this kind of thing (because simpsets resurrect via theory merges) and also the lemma says:
```
"ptr - (ptr && ~~ mask sz) = (ptr && mask sz)"
```
which seems fair to keep in the simpset. Unfortunately removing this will potentially require another update cycle.

proof/refine/AARCH64/ArchStateRelationLemmas.thy
Xaphiosis: I'd appreciate your input here then. The comment is trying to anticipate the future where someone goes "oh this is requalified, why isn't it part of the Arch interface" and then rediscovering that it's not really possible (e.g. when we have to migrate to MCS).

The term "interface lemma" appears in quite a few places, including in AInvs as shorthand for "Arch interface lemma".

proof/refine/AARCH64/ArchKHeap_R.thy
Xaphiosis: The latter, but looking at it now I'm not sure such a place will appear, as the next one that pulls it in is SubMonad_R (and also Init_R). Hmm.

proof/refine/AARCH64/ArchKHeap_R.thy
Xaphiosis: Haven't noticed any issues with that yet. Outside of Arch, I use `gen_typ_at_lifts` and inside Arch `typ_at_lifts`, and stick to that same naming scheme with other lemma groupings (e.g. `gen_objBits_simps`). I think there is a bit of a quibble between `objBits_simps` and `objBits_defs`, but that's from a different PR.

proof/refine/AARCH64/ArchKHeap_R.thy
Xaphiosis: Yes. Although tagging the crunch with `[KHeap_R_assms_2]` won't export anything more than what the interface requires anyway, it would be confusing/misleading to look at.

proof/refine/AARCH64/ArchKHeap_R.thy
Xaphiosis: Ah, yes indeed. It seems that on all arches we can say that `setObject` preserves `pspace_in_kernel_mappings'`, so we can try to export that instead and maybe even do the Endpoint/Notification instantiations outside of Arch if RISCV64/X64 permit. Will take another look.

proof/refine/AARCH64/ArchKHeap_R.thy
Xaphiosis: Removed? They're currently part of the interface... we can show that setObject preserves `pspace_in_kernel_mappings'` on all arches, and so can change the interface to require that... or did you mean something else here?

proof/refine/KHeap_R.thy
Xaphiosis: I think it was tagged when it was still in the Arch theory, but then after I moved it, it still didn't feel like it was in a good spot. I'll consider moving it to Bits_R. Same for `not_is_other_obj_relation_type` which should go where the other_obj mess lives (and oh man, that name is so bad, I'll take a look too)

proof/refine/KHeap_R.thy
Xaphiosis: Will do

proof/refine/ARM/ArchKHeap_R.thy
Xaphiosis: as in `interface lemma, but can't be done via locale`?

proof/refine/AARCH64/ArchKHeap_R.thy
Xaphiosis: done, will show up in next push

proof/refine/KHeap_R.thy
Xaphiosis: done, will show up in next push

proof/refine/KHeap_R.thy
Xaphiosis: done in next commit

proof/refine/KHeap_R.thy
Xaphiosis: Agreed, will remove

proof/refine/AARCH64/ArchKHeap_R.thy
corlewis: You might have already done it, but you would want to do that for all architectures

proof/refine/AARCH64/ArchKHeap_R.thy
corlewis: Sounds good and the consistent naming scheme will definitely be a help.

Just a note for the future while this is on my mind, `typ_at_lifts` might be one of the pain points for when it's time to arch-split MCS. We made some pretty significant changes there that ended up being a big improvement but which will mean that what's done on master to it won't map across nicely, if at all. I'm not sure what the best path forwards for this specifically will be, it's possible that backporting the `typ_at_lifts` changes to `master` will help but that also might be more work than it's worth.

proof/refine/AARCH64/ArchKHeap_R.thy
Xaphiosis: Well no, trying to add a `setObject` assumption to the locale also results in a free type variable, so looks like we work around it same as for `setObject_other_corres_ep` et al

But the comments can still go.

proof/refine/AARCH64/ArchKHeap_R.thy
Xaphiosis: Don't think so. They're needed by, for instance `set_ep_valid_pspace'` which is in the locale.

proof/refine/AARCH64/ArchStateRelationLemmas.thy
corlewis: Fair point with "interface lemma" being used consistently for this. I guess I'm ok with it as is, something about how it's written still feels a bit off to me but I can't think of anything better.

corlewis: And oh, just because I've started at this for too long now, could we remove the space from `typ_at_lifts [OF P]` while we're here?

proof/refine/AARCH64/ArchKHeap_R.thy
lsf37: I meant just removing the FIXME -- these two are the interface, basically.

proof/refine/KHeap_R.thy
corlewis: Yeah, with that being the lemma my preference would be to remove this but then I'm not the one who would have to do the update cycle. Maybe if you remember you could try removing it when you're having to update things anyway for something else?

Xaphiosis: removed on my side, cleared immediate breakage, waiting on tests

proof/refine/AARCH64/ArchKHeap_R.thy
Xaphiosis: cool. done in next push

proof/refine/AARCH64/CNodeInv_R.thy
corlewis: I guess it doesn't really matter, since presumably this comment will go away once arch-split reaches CNodeInv_R. I'm still a bit confused though about them being interface lemmas when they don't appear on `X64` or `RISCV64`, but again that will be resolved later once we reach here.

Xaphiosis:
I've pushed PR update commits with a bunch of cleanup (see commit message on the "address PR issues" one).

proof/refine/AARCH64/ArchKHeap_R.thy
Xaphiosis: done

proof/refine/ARM/ArchKHeap_R.thy
Xaphiosis: done

proof/refine/KHeap_R.thy
Xaphiosis: done

proof/refine/KHeap_R.thy
Xaphiosis: moved to Bits_R

proof/refine/AARCH64/ArchKHeap_R.thy
Xaphiosis: done

proof/refine/AARCH64/ArchKHeap_R.thy
Xaphiosis: done

proof/refine/AARCH64/ArchKHeap_R.thy
Xaphiosis: this should be resolved now, but see the setObject turbulence in latest commits

proof/refine/AARCH64/ArchKHeap_R.thy
Xaphiosis: My knowledge of MCS is very poor, not sure what we can do about it. How different is it over there?

proof/refine/AARCH64/ArchKHeap_R.thy
Xaphiosis: Changed the comment on this, and added the setObject pspace_in_kernel_mappings' lemma that has the same curse.

proof/refine/AARCH64/ArchStateRelationLemmas.thy
Xaphiosis: ah, missed that one due to a race. will tweak after tests pass

proof/refine/KHeap_R.thy
Xaphiosis: done

proof/refine/KHeap_R.thy
Xaphiosis: done

proof/refine/KHeap_R.thy
Xaphiosis: done, still waiting on CRefine

proof/refine/KHeap_R.thy
Xaphiosis: done

proof/refine/AARCH64/CNodeInv_R.thy
Xaphiosis: Yeah, I think that's the right move for now. The FIXMEs that appear in downstream  theories usually indicate there's some work to be done that we don't want to forget.

proof/refine/AARCH64/ArchKHeap_R.thy
corlewis: Very different, see https://github.com/seL4/l4v/blob/rt/proof/refine/ARM/Invariants_H.thy#L3035. There's a couple of locales that are built up to contain a larger collection of lifting lemmas than are in `typ_at_lifts` in master, and that are then interpreted as needed instead of being instantiated with `OF`.

Some of this is more complicated than it would be on master due to MCS specifics around `sc_at_n`, but I guess at the very least there will need to be an arch version of the locales split out.

proof/refine/AARCH64/ArchKHeap_R.thy
```diff
@@ -15,7 +15,7 @@ context Arch begin arch_global_naming
 named_theorems KHeap_R_assms
 
 lemma getObject_inv_vcpu[wp]:
-  "\<lbrace>P\<rbrace> getObject l \<lbrace>\<lambda>(rv :: ArchStructures_H.vcpu). P\<rbrace>"
+  "\<lbrace>P\<rbrace> getObject l \<lbrace>\<lambda>(_ :: ArchStructures_H.vcpu). P\<rbrace>"
```
corlewis: Incredibly minor but just to clarify, I was originally more commenting on the brackets not being needed around the type annotation here and that we normally don't have spaces around `::` when it's inside a lambda like that.

Xaphiosis: oh, got it, will add to list

proof/refine/AARCH64/ArchStateRelationLemmas.thy
Xaphiosis: done now

proof/refine/AARCH64/ArchKHeap_R.thy
Xaphiosis: done. have a look

corlewis: Looks good, thanks!

proof/refine/AARCH64/ArchKHeap_R.thy
lsf37: Works for me.

proof/refine/KHeap_R.thy
lsf37: Cool, nice to reduce those global declarations.

lsf37 (approved):
Good to merge from my side!

PR #911, opened 2025-07-30
unify SubMonad_R and arch-split ArchAcc_R

Xaphiosis:
This one is a bit less exciting numbers-wise:
`58 files changed, 922 insertions(+), 1651 deletions(-)`

The reason for it is that nearly everything in ArchAcc_R seems arch-specific. In theory there are commonalities, such as every arch having ASIDs and ASID tables, but I'm not seeing a way to extract such genius, since the objects are under the "arch obj" umbrella, and the ASID tables themselves are different.

Let me know if you see something I've missed.
Also, I realise this PR is immediately after the last arch-split PR, but with @lsf37 travelling next week, it seems a good idea to put it up now.


proof/refine/ARM/Retype_R.thy
```diff
@@ -212,7 +212,7 @@ lemma null_filter_parent_of'':
      \<Longrightarrow> ys \<turnstile> x \<leadsto> c"
   apply (clarsimp simp add: mdb_next_unfold)
   apply (drule arg_cong[where f="\<lambda>xs. xs x"])
-  apply (simp add: null_filter'_def nullPointer_def split: if_split_asm)
+  apply (simp add: null_filter'_def nullPointer_def split: if_split_asm cong: if_weak_cong)
```
lsf37: is `if_cong` global again at that point the theory chain? Or is this a left-over?

lsf37 (approved):
I like it. The fallout from `if_cong` is much smaller than I feared and it's nice to see things converging more and more.

proof/refine/AARCH64/ArchArchAcc_R.thy
```diff
@@ -966,10 +944,10 @@ lemma arch_deriveCap_valid:
   "\<lbrace>valid_cap' (ArchObjectCap arch_cap)\<rbrace>
      Arch.deriveCap u arch_cap
    \<lbrace>\<lambda>rv. valid_cap' rv\<rbrace>,-"
-  apply (simp add: AARCH64_H.deriveCap_def split del: if_split)
+  apply (simp add: AARCH64_H.deriveCap_def split del: if_split cong: if_cong)
   apply (wp undefined_validE_R)
   apply (cases arch_cap; simp add: isCap_defs)
-  apply (simp add: valid_cap'_def capAligned_def capUntypedPtr_def AARCH64_H.capUntypedPtr_def)
+  apply (simp add: valid_cap'_def capAligned_def global.capUntypedPtr_def capUntypedPtr_def)
```
corlewis, line 950: Don't we want the unqualified name to be the generic version, not the arch-specific one? This being this way around suggests that something is wrong with the setup for `capUntypedPtr`.

proof/refine/X64/ArchArchAcc_R.thy
```diff
@@ -35,6 +30,181 @@ lemma asid_low_bits [simp]:
   "asidLowBits = asid_low_bits"
   by (simp add: asid_low_bits_def asidLowBits_def)
 
+lemma pspace_aligned_cross[ArchAcc_R_assms]:
+  "\<lbrakk> pspace_aligned s; pspace_relation (kheap s) (ksPSpace s') \<rbrakk> \<Longrightarrow> pspace_aligned' s'"
```
corlewis, line 34: I was a bit surprised to see this being moved in this commit. I guess it was here in the other arches but X64 had it in ArchKHeap? Very minor but could be worth mentioning in the commit message.

proof/refine/AARCH64/ArchArchAcc_R.thy
```diff
@@ -5,12 +5,10 @@
...
  *)
 
-(*
-  Lemmas on arch get/set object etc
-*)
+(* Arch-specific lemmas on arch get/set object etc *)
 
 theory ArchAcc_R
-imports SubMonad_R
+imports ArchArchAcc_R
```
corlewis: If we want to clean up the history, this should probably be changed to
```
theory ArchArchAcc_R
imports ArchAcc_R
```
here in the rename commit instead of in the following arch-split commit.

corlewis (commented):
Very nice progress! Same as Gerwin, much less pain from `if_cong` than expected.

However, I'm not a big fan of having a file called `ArchArchAcc_R`, it feels a bit silly as a name and it also feels a little bit strange to have a generic version of a file described with `Arch-specific lemmas on arch get/set object etc`.

It might be more changes than we want to do in this PR, but I think in my perfect world the generic `ArchAcc_R` would be removed and its contents moved to an earlier file. The lemmas that are currently in the interface would also need to move earlier, possibly to `ArchKHeap_R` like they were on X64. That would make it so that the arch-specific files can go back to being called `ArchAcc_R`, and they would stay focused on the arch get/set object lemmas like the comment suggests.

If that's too much, then maybe the generic file could be `ArchAccPre_R`? That's still not great but at least avoids the need for `ArchArchAcc_R`. It could maybe do with a comment also, describing the purpose of the file.

proof/refine/AARCH64/ArchArchAcc_R.thy
Xaphiosis: This is completely screwed up, I've commented on it before and I don't know why it is this way. Essentially, when you go into Arch, it's the way you see before part of the diff, but then when you say `named_theorems blah`, it becomes what you see in the after part.

For some reason this made me get completely confused, and I started thinking that the arch-specific version should be the one in the Arch locale (or I was lazy and realised I could have the same text on all arches)... but something is broken here, and I've not been able to figure out why.

@lsf37 : thoughts? I might need to spend a bit of time debugging this issue

proof/refine/X64/ArchArchAcc_R.thy
Xaphiosis: Exactly, it was migrated by someone for X64, but then it turned out here that it didn't have to be migrated.

- [x] add to arch-split commit message

proof/refine/AARCH64/ArchArchAcc_R.thy
Xaphiosis: good point, will try to untangle

proof/refine/AARCH64/ArchArchAcc_R.thy
lsf37: Not sure I'm seeing enough context in the diff, but these would be my expectations:

If we are inside the Arch locale, then I it would not be unexpected to me to have the arch version be the unqualified name. It is the last name that was made locally available, and the generic one would have been shadowed.

If we are outside the Arch locale, I would expect the generic one to be unqualified and the arch name to be qualified.

lsf37:
In general, I'm not too bothered by having `ArchArchAcc` and `ArchAcc`, esp given the precedent in `Arch_AI`. That said, if the comment for the generic `ArchAcc` theory doesn't fit any more we should either update the comment or make it true again, whichever is easier.

Moving some of the helper lemmas further up would work, but updating the comment would also be fine.

proof/refine/AARCH64/ArchArchAcc_R.thy
Xaphiosis: That would normally be the case if we didn't go to the trouble of the whole global/arch dance as documented in the arch-split docs and in `lib/test/Requalify.thy`. It really looks like the intention was that for these cases you always get the global one, and need an `Arch` prefix for the Arch one. *However*, the arch-split docs I inherited only talk about *consts*, and this still works correctly:
```
term deriveCap (* global.deriveCap *)

named_theorems Moo
term deriveCap (* still global.deriveCap *)
end
```

One problem is that somewhere in my arch-split overhaul I must've hit something that made me believe that fact naming could be handled the same way. And indeed, they normally could, as seen in `lib/test/Requalify.thy` examples I wrote up. However once `named_theorems` are added, things break for fact names:
```
(* when interpreted with named_theorems *)
context begin interpretation Arch .
named_theorems Moo
thm requalify_collision (* BROKEN: should be False = False *)
thm global.requalify_collision (* False = False *)
thm Arch.requalify_collision (* True = True *)
end

(* and in the locale context with named_theorems *)
context Arch begin
named_theorems Moo
thm requalify_collision (* BROKEN: should be False = False *)
thm global.requalify_collision (* False = False *)
thm Arch.requalify_collision (* True = True *)
end

(* BUT if we try again without named_theorems *)
context Arch begin
thm requalify_collision (* WORKS AGAIN: False = False *)
thm global.requalify_collision (* False = False *)
thm Arch.requalify_collision (* True = True *)
end
```

In `ArchInvsDefs_H` I do indeed try to do this global/Arch disambiguation with `capUntypedPtr_def`, and hence we encounter this problem where the naming shifts once `named_thms` is introduced.

The current situation can be summarised as:
* `named_theorems` breaks the default case of the global/Arch disambiguation system for facts, but `global.name` and `Arch.name` still work fine
* the arch-split manual only talks about disambiguating constants, not facts, and so isn't broken (but could be amended with this new info)
* `lib/test/Requalify_Test.thy` does not test consts, but only facts, and acts like they will always work; this is misleading given the new information, and so needs adding consts and demonstrating the named_theorems discrepancy
* the global/Arch disambiguation is still needed for definitions; using the qualifiers `Arch.` and `global.` when both need to be unfolded is something we can do, but there's no system to enforce it (and the fact the default can flip based on the presence of `named_theorems` makes it horrid for anyone reading it)
* there's no way I know of to get hide the unqualified name only inside the Arch locale
* this isn't an issue for ASpec/AInvs because we use an `arch_` prefix there

One possible solution that stabilises things is to force `named_theorems` to be introduced separately:
```
named_theorems Moo
(* NOTHING ELSE HERE *)
end

(* actual Arch proofs ... *)
```

This is not free however. Even though `context Arch` (2.3s) is much faster than the `context begin interpretation Arch .` (6.3s), it's still over 2 seconds on a recently upgraded machine.

I am looking into named_theorems to see why it screws us up the way it does, but so far am only confused.

@lsf37 @corlewis what are your thoughts? Also, might be that for this PR we leave it, and create an issue to deal with it properly :/

Xaphiosis:
> For the second part, what led to my opinion is that the lemmas in the interface don't seem to have anything to do with the commented intent of `ArchAcc_R`, in that instead of being lemmas to do with arch get/set object they are unrelated cross lemmas and `obj_relation_cuts_range_mask_range`. That's why my suggestion was to follow X64's approach and move specifically those lemmas and their interface to `ArchKHeap_R` and `KHeap_R`.

I think we should create an issue for this, mostly because I don't really know what is supposed to be "ArchAcc" and not. It would mean those poor X64 lemmas would be moved back at some point in the future, but right now I am thinking that sadly we need faster arch-split progress. I'll change the comment, and make an issue from your comments, because I agree it's a bit of a mess. I also think that as arch-split proceeds, you sometimes need to dig out more stuff to be put in the interfaces (i.e. stuff only works later on because things magically leak out of Arch). I think this should definitely be cleaned up before it heads towards MCS. It can even be done in parallel once the CSpace* theories are split, I think.

- [ ] make issue for ArchAcc_R vs KHeap_R cleanup

proof/refine/AARCH64/ArchArchAcc_R.thy
corlewis: Thank you for investigating in such detail. I think you're right in that this can be left for this PR but it would be good to work out what we would want the behaviour to be.

I've come around a bit to the idea of what Gerwin mentioned, with it not being unexpected to have the arch version be the unqualified name when inside the Arch locale. This would have the benefit of making consistent proof text across architectures easier, which actually is pretty nice the more I think of it. If there was a way to set it up like that without depending on the weird `named_theorems` behaviour that might actually be my preference.

Regardless of that, I guess the priority might be updating the arch-split documentation to make things clearer around consts vs facts, and possibly including a warning about `named_theorems`.

corlewis:
> > For the second part, what led to my opinion is that the lemmas in the interface don't seem to have anything to do with the commented intent of `ArchAcc_R`, in that instead of being lemmas to do with arch get/set object they are unrelated cross lemmas and `obj_relation_cuts_range_mask_range`. That's why my suggestion was to follow X64's approach and move specifically those lemmas and their interface to `ArchKHeap_R` and `KHeap_R`.
> 
> I think we should create an issue for this, mostly because I don't really know what is supposed to be "ArchAcc" and not. It would mean those poor X64 lemmas would be moved back at some point in the future, but right now I am thinking that sadly we need faster arch-split progress. I'll change the comment, and make an issue from your comments, because I agree it's a bit of a mess. I also think that as arch-split proceeds, you sometimes need to dig out more stuff to be put in the interfaces (i.e. stuff only works later on because things magically leak out of Arch). I think this should definitely be cleaned up before it heads towards MCS. It could even be done in parallel once the CSpace* theories are split.
> 
>     * [ ]  make issue for ArchAcc_R vs KHeap_R cleanup

That sounds like a reasonable approach. I do still like the idea of not having another arch-generic file when not needed but you're definitely right in that the needed interface might change as more progress is made.

proof/refine/AARCH64/ArchArchAcc_R.thy
lsf37: I would want to avoid additional `context Arch` statements, not just because they take time, but also because it is easy to forget them and you only get bitten by the consequences much later.

We could standardise on always using both `Arch.` and `global.` for fact names when there is a clash. As Raf says, we can't really enforce it, but they often (usually?) occur together, so it's relatively easy to spot. And when they do not, it often is clear which one we need (because the other one then usually does no work), so in those cases it is not that problematic if we miss it in review.

Also thumbs up to mentioning it in the docs.

proof/refine/ARM/Retype_R.thy
Xaphiosis: Good catch. Fixed on affected arches (ARM, ARM_HYP)

proof/refine/AARCH64/ArchArchAcc_R.thy
Xaphiosis: created issue for it, will take a look after this PR

proof/refine/X64/ArchArchAcc_R.thy
Xaphiosis: Added note to commit message. Let me know if it suffices

proof/refine/AARCH64/ArchArchAcc_R.thy
Xaphiosis: Thanks for catching that. Migrated to rename commit.

corlewis (approved):
Looks like you've handled all of the comments, thanks for making those issues so that we don't forget them!

PR #912, opened 2025-08-10
misc: bash completion for run_tests

lsf37:
Auto-complete session names and options for the main run_tests script.

I wrote this mainly for myself, but publishing it here in case it's useful for anyone else.

The portable shell check will fail, because this is inherently bash only.

### DIFF
--- a/misc/regression/run_tests-bash-completion.sh
+++ b/misc/regression/run_tests-bash-completion.sh
@@ -1,0 +1,68 @@
+# Copyright 2025 Proofcraft Pty Ltd
+#
+# SPDX-License-Identifier: BSD-2-Clause
+#
+# bash completion for run_tests
+# source this file to use: . ./run_tests-bash-completion.sh
+#
+
+_run_tests_sessions() {
+  # $1 is run_tests
+  # use -L to remove sessions that are excluded by L4V_ARCH selection
+  case "$L4V_ARCH" in
+    ""|"ARM")
+      # prints additional line about setting L4V_ARCH_IS_ARM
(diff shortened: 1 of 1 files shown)


misc/regression/run_tests-bash-completion.sh
```diff
@@ -0,0 +1,68 @@
...
+      "$1" -L | tail -n +3
+      ;;
+    "ARM_HYP"|"X64"|"RISCV64"|"AARCH64")
+      "$1" -L | tail -n +2
+      ;;
+  esac
+}
+
+_run_tests_completion() {
+  local all_opts="--l4v-arches --l4v-arch-all -h --help -d --directory \
```
Xaphiosis, line 24: while I'm here, I've been annoyed multiple times by having to switch between `--l4v-arch-all` and `--l4v-arches=`, to the point I've wanted to rename the former (or add an alias) `--l4v-arches-all`, or be able to feed ALL into `--l4v-arches=`. What's your level of annoyance with this, and thoughts thereon?

lsf37: I rarely use the `--l4v-arches` switches, but I agree with the annoyance of the difference in spelling, it's not streamlined. Adding an alias `--l4v-arches-all` would be very easy to do, I'd be up for that. We can even add a depreciation message for the old one.

Not entirely sure about adding ALL -- would generally be fine, but a bit more special casing and parsing and it might give the impression that other things can deal with ALL as well.

lsf37: I'll put up a separate PR for it.

PR #913, opened 2025-08-12
c-parser: document options in README

lsf37:
Document the existing options of the `install_C_file` command.


corlewis (approved):
Thank you for documenting these options!

tools/c-parser/README.md
```diff
@@ -44,6 +44,56 @@ for more information.
...
 
+Usage
+-----
+
+The C parser provides a command `install_C_file` that parses and translates
+a single `.c` file into Isabelle.
+
+The syntax is as follows
+
+```isablle
```
Xaphiosis: isabelle?

tools/c-parser/README.md
```diff
@@ -44,6 +44,56 @@ for more information.
...
+- `machinety=<type_name>`: define an additional state field
+  `phantom_machine_state` of type `type_name` as part of the state record. This
+  is intended to be used for variable read/write state that is not visible to
+  standard C semantics, and can be read/written in `AUXUPD` annotations, modifies
+  clauses, and `SPEC` annotations.
+- `ghostty=<type_name>`: define an additional state field `ghost'state` of type
+  `type_name` as part of the global state record. This state is intended to be
+  ghost state. It can only be written to from the program via `GHOSTUPD`
+  annotations, but not read by normal C code. It is accessible and useful in
+  Hoare logic assertion.
```
Xaphiosis: assertions?

tools/c-parser/README.md
```diff
@@ -44,6 +44,56 @@ for more information.
...
+- `roots=[fun_name_1, fun_name_2, ...]`: only translate functions `fun_name_1`,
+  `fun_name_2`, etc, ignoring the rest of the functions in the provided file.
+  This is useful for verifying specific functions in a larger C context that
+  is not entirely in the StrictC subset.
```
Xaphiosis: roots implies that it also translates the functions that these functions call, recursively, no?

tools/c-parser/README.md
```diff
@@ -44,6 +44,56 @@ for more information.
 [MLton compiler]: http://mlton.org
 [AutoCorres]: https://trustworthy.systems/projects/OLD/autocorres/
 
+Usage
+-----
+
+The C parser provides a command `install_C_file` that parses and translates
+a single `.c` file into Isabelle.
```
Xaphiosis: was there some functionality in the past that dealt with `#include`, which would technically make it more than a single file?

tools/c-parser/README.md
lsf37: indeed

tools/c-parser/README.md
lsf37: Good point. I have never tried it, but the JIRA issue that mentions it does confirm.

tools/c-parser/README.md
lsf37: I was trying not to mention the includes, because I'm not exactly sure what the parser does with those since we don't use them.  Looks like I didn't get away with that :-) I think the functionality still exists, so I'll have a closer look at what it does with those and describe here.

PR #914, opened 2025-08-12
cparser: add option to skip modifies proofs

lsf37:
Add an option to the C parser to skip the `modifies` proofs, and use that option in `install_C_file` in CKernel.

We're not using the modifies proofs generated by the parser in CRefine -- instead we are reproving them in Substitute.thy. Not generating the first set should remove a few minutes of build time and save some memory.


tools/c-parser/isar_install.ML
```diff
@@ -447,9 +447,9 @@ in
                                             globs
                                             thy
       val thy =
-          if not (Symtab.is_empty (get_defined_functions cse)) then
+          if not (Symtab.is_empty (get_defined_functions cse)) andalso not (o2b no_modifies) then
             Modifies_Proofs.prove_all_modifies_goals thy cse toTranslateP styargs loc2
-          else thy (* like this is ever going to happen *)
+          else thy
```
corlewis, line 452: Haha, a bit amusing to see this comment removed.

tools/c-parser/testfiles/no_modifies_test.thy
```diff
@@ -0,0 +1,26 @@
...
+
+external_file "modifies_assumptions.c"
+install_C_file no_modifies "modifies_assumptions.c"
+
+context modifies_assumptions
+begin
+
+thm f_body_def
+
+(* Would produce a name class error of f_modifies existed: *)
```
corlewis:
```suggestion
(* Would produce a name class error if f_modifies existed: *)
```

corlewis (approved):
I was a bit curious so run some tests to see the actual impact. It does indeed save a small amount of time and it looks like it does use a fair bit less memory, although that is always particularly noisy and hard to tell for sure.

```
master:
  N = 10
  Average ARM CKernel: (00:07:24 real, 00:31:29 cpu, 10.62GB)
  Standard Deviation: (00:00:08 real, 00:00:31 cpu, 1.87GB)
modifies:
  N = 10
  Average ARM CKernel: (00:06:53 real, 00:20:56 cpu, 9.14GB)
  Standard Deviation: (00:00:09 real, 00:00:36 cpu, 2.11GB)
Percentage change for real: -7.0%
```

tools/c-parser/README.md
```diff
@@ -63,6 +63,8 @@ They have to be provided in the order listed below.
              frame rules
 - `c_types`: only define the types in the C code, leave out function definitions
 - `c_defs`: only provide C definitions, leave out type definitions
+- `no_modifies`: do not generate "modifies" proofs (frame theorems about which
+  globals each function can change).
```
Xaphiosis: brain parse trouble, maybe "frame theorems which indicate the globals each function can change"?

tools/c-parser/README.md
```diff
@@ -87,7 +89,7 @@ enclosed in square brackets can be provided. These are:
 An example that uses all options:
 
 ```isabelle
-install_C_file memsafe c_types c_defs "file.c"
+install_C_file memsafe c_types c_defs no_modifies "file.c"
                [machinety=int, ghostty=nat, roots=[f, g, h]]
```
Xaphiosis, line 94: worth explaining *why* we don't bother with modifies theorems during the kernel pass, maybe a pointer to when they do happen?

tools/c-parser/testfiles/no_modifies_test.thy
```diff
@@ -0,0 +1,26 @@
...
+
+(* Test that the "no_modifies" argument does not produce modifies proofs *)
+
+external_file "modifies_assumptions.c"
+install_C_file no_modifies "modifies_assumptions.c"
+
+context modifies_assumptions
+begin
+
+thm f_body_def
```
Xaphiosis, line 19: This is fine to leave, but I wanted to add today's fresh mind-bending discovery. You know how we can say:

```
lemmas [simp] = blah blah
```

well I thought, if we don't want to trip the `thm` situation, but want to show that a theorem exists, could we do

```
lemmas = blah blah
```

turns out that no, we can't... *but* look at this:

```
lemmas _ = exI (* theorem _: ?P ?x ⟹ ∃x. ?P x *)
print_theorems (* theorems:  _: ?P ?x ⟹ ∃x. ?P x *)
thm _ (* PROP ?psi ⟹ PROP ?psi *)
```

lsf37: That is great :-). Thinking about it, `_` is a legal name part in Isabelle, so it is also a legal lemma name like `*` (but only in outer syntax, constants etc are more constrained). I think that syntax was there before `_` became a thing as pattern placeholder and nobody noticed the interaction.

Well, if we ever want to thoroughly confuse people, we now have a new tool available :-)

tools/c-parser/README.md
lsf37: Done (in the Kernel_C files).

tools/c-parser/testfiles/no_modifies_test.thy
lsf37: (No idea why the last line does what it does, though. Maybe symbol table operations/shadowing also gets confused)

PR #916, opened 2025-08-12
make C separation logic notation optional

lsf37:
- make a syntax bundle for C separation logic notation
- remove the bundle by default in CTranslation
- use `unbundle C_seplog_syntax` where needed

This removes the syntax clash with standard Isabelle maps-to notation and reduces warnings in CRefine.

Closes Jira [VER-519](https://sel4.atlassian.net/browse/VER-519)

[VER-519]: https://sel4.atlassian.net/browse/VER-519?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/clib/CCorresLemmas.thy
+++ b/lib/clib/CCorresLemmas.thy
@@ -436,6 +436,10 @@ lemma clift_h_t_validD:
 
 (* Tests *)
 
+context
+includes C_seplog_syntax
+begin
+
 lemma sep_conj_dom:
   "((\<lambda>s. dom s = X) \<and>\<^sup>* (\<lambda>s. dom s = Y)) = (\<lambda>s. dom s = X \<union> Y \<and> X \<inter> Y = {})"
   unfolding sep_conj_def
@@ -447,6 +451,8 @@ lemma sep_conj_dom:
   apply (clarsimp simp: map_disj_def Int_Un_distrib2)
   apply (simp add: Int_ac heap_merge_restrict_dom_un Un_ac)
   done
+
+end
 
 lemma Diff_Diff_Un_eq: "(A - B) - C = A - (B \<union> C)"
   by (simp add: Diff_eq Int_ac)


corlewis (approved):
Very nice improvement, it's always good to see reduced warnings!

Is there anywhere that it would make sense to document using `unbundle C_seplog_syntax` to re-enable the syntax? Currently it's only mentioned in the commit message.

tools/c-parser/CTranslation.thy
```diff
@@ -180,4 +180,7 @@ lemma proc_deps_lvar_nondet_init [simp]:
 
 declare word_neq_0_conv[simp] (* FIXME: for backward compatibility; remove *)
 
+(* Remove separation logic notation, because it leads to too many syntax ambiguities in general *)
+unbundle no C_seplog_syntax
```
Xaphiosis, line 187: not clear on why you need this when I don't see you having unbundled C_seplog_syntax outside of Separation.thy ... is this so that when people include Separation.thy they don't need to additionally unbundle it? CTranslation includes something that unbundles it? Feels like it's not a commonly used feature

Xaphiosis (approved):
Good idea, this has been bothering us for a while.
Agree with Corey, might be nice to document this somewhere.

tools/c-parser/CTranslation.thy
lsf37: CTranslation.thy (transitively) includes Separation.thy, that's how the syntax became global visible in the first place. This one just switches off the bundle from Separation.thy again. Will add that to the comment.

lsf37:
> Is there anywhere that it would make sense to document using `unbundle C_seplog_syntax` to re-enable the syntax? Currently it's only mentioned in the commit message.

I could add it to the comment that removes the syntax at the bottom of CTranslation. If people don't find the bundle in Separation.thy, that's probably the most likely place people will be looking at.

PR #917, opened 2025-08-15
c-parser: add support for _Static_assert statements

midnightveil:
The kernel AST now produces statements like:

    ##StaticAssertion: long_is_32bits
    StaticAssertion [
      "long_is_32bits"
      BinOp [
        "=="
        SizeofTy [
          Unsigned [
            "long"
          ],
        ],
        Constant [
          NUMCONST [
            "4"
            ""
            "DEC"
          ],
        ],
      ],
    ],

This should allow the horrible array size hack to be removed from the kernel source code, i.e. [include/assert.h](
https://github.com/seL4/seL4/blob/13.0.0/include/assert.h#L45-L69).

---

Was this necessary? No
Was this interesting? Yes
Was this fun? Uhhh well I want to do more of it

---

It doesn't seem like any of the kernel proofs actually proved anything against the static asserts, so this should be fine. (I ran the proofs locally after making all the static assertions false and they still succeeded, so I think only the compiler checks these).

---

I've no idea how to write an appropriate test, aside from that it parses.

I also have [a similar one for offsetof](https://github.com/seL4/l4v/compare/master...midnightveil:seL4-l4v:julia/offsetof?expand=1), but I don't want to trigger the CI builds unnecessarily by making another PR (for the moment, at least).

### DIFF
--- a/tools/c-parser/StrictC.grm
+++ b/tools/c-parser/StrictC.grm
@@ -725,6 +725,7 @@ end
             struct_fields wrap * struct_id_decl wrap list
         | struct_declaration of struct_fields wrap * struct_id_decl wrap list
         | type_definition  of declaration
+        | static_assert_declaration of declaration wrap
         | elementary_type of expr ctype
         | integral_type of inttype wrap
         | type_name of expr ctype wrap
@@ -842,6 +843,7 @@ end
     | DONT_TRANSLATE
     | STRING_LITERAL of string
     | SPEC_BLOCKEND
+    | STATIC_ASSERT
     | GCC_ATTRIBUTE
     | YASM
     | YREGISTER
@@ -997,6 +999,13 @@ declaration
                  end) *)
 | declaration_specifiers init_declarator_list YSEMI
                 (make_declaration declaration_specifiers init_declarator_list)
+| static_assert_declaration ([static_assert_declaration])
+
+(* 6.7.10 and/or §A.2.2 ``static_assert-declaration: _Static_assert ( constant-expression , string-literal ) ;'' *)
+(* constant-expression is modelled as rexpression (c.f. "enumerator") *)
+static_assert_declaration
+: STATIC_ASSERT LPAREN rexpression YCOMMA cstring_literal RPAREN YSEMI
+                (wrap(StaticAssertion { condition = rexpression, message = cstring_literal }, eleft rexpression, eright rexpression))
 
 declaration_specifiers
 : storage_class_specifier
@@ -1187,6 +1196,8 @@ struct_declaration
                          left basetype, right struct_declarator_list),
                     sdecls)
                  end)
+(* per the C11 spec this should be allowed but let's not for simplicity *)
+(* | static_assert_declaration () *)
 
 struct_declarator_list :
    struct_declarator (wrap ([node struct_declarator], left struct_declarator,
--- a/tools/c-parser/program_analysis.ML
+++ b/tools/c-parser/program_analysis.ML
@@ -1681,6 +1681,9 @@ in
       (puredecl d, pop_scope (set_proto_params fname (List.rev pm_vis0) e))
     end
   | EnumDecl eninfo => (puredecl d, process_enumdecl eninfo e)
+  (* nothing to be done *)
+  (* XXX: like process_enumdecl, use const_eval on the condition to check it? *)
+  | StaticAssertion {...} => (puredecl d, e)
 end
 
 type asm_stmt_elements = (string * bool * expr option * expr list)
--- a/tools/c-parser/testfiles/static_assert.c
+++ b/tools/c-parser/testfiles/static_assert.c
@@ -1,0 +1,11 @@
+/*
+ * Copyright 2025, UNSW
+ *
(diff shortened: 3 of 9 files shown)


tools/c-parser/program_analysis.ML
```diff
@@ -1681,6 +1681,9 @@ in
       (puredecl d, pop_scope (set_proto_params fname (List.rev pm_vis0) e))
     end
   | EnumDecl eninfo => (puredecl d, process_enumdecl eninfo e)
+  (* nothing to be done *)
+  (* XXX: like process_enumdecl, use const_eval on the condition to check it? *)
+  | StaticAssertion {...} => (puredecl d, e)
```
lsf37, line 1686: I guess this is the main question we have to decide on. I think we should stay with the "do nothing", but I'm keen to hear arguments for/against.

tools/c-parser/testfiles/static_assert.c
```diff
@@ -0,0 +1,11 @@
...
+ * Copyright 2025, UNSW
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+_Static_assert(2 + 3 == 5, "Hello");
+
+// struct hello {
+//   _Static_assert(4 + 4 == 8, "Goodbye");
+// }
```
lsf37, line 11: I'd like to see a few more tests, since the parser is correctness critical for architectures without binary verification.

If this can happen anywhere where we can have a declaration, we should at least add one inside a function body. Anything else that would make sense?

We're generally not so good with negative tests for the parser, but it would be worth adding a few to the `testfiles/errors` directory for this one. E.g. one where the `_Static_assert` has content that does not parse (in either of the arguments, maybe also 0, 1, and > 2 arguments), and one where `_Static_assert` as a whole occurs in a position it is not supposed to occur.

tools/c-parser/testfiles/static_assert.thy
```diff
@@ -0,0 +1,18 @@
...
+begin
+
+external_file "static_assert.c"
+install_C_file "static_assert.c"
+
+context static_assert
+begin
+
+end
+end
```
lsf37, line 18: If we're adding a function body to the test, then we can add a `thm f_body_def` statement here to check that the function body has been defined. In this case it even make sense to state a lemma that shows that it is the expected function body so that the test confirms that the assert has been ignored (with a comment that this is what it does).

lsf37:
If we add this, we should also add a remark to `doc/ctranslation.tex` that the parser handles and ignores `_StaticAssert`. Probably needs a new section somewhere close to the GCC attributes.

lsf37:
> I also have [a similar one for offsetof](https://github.com/seL4/l4v/compare/master...midnightveil:seL4-l4v:julia/offsetof?expand=1), but I don't want to trigger the CI builds unnecessarily by making another PR (for the moment, at least).

Does clang support the same attribute the same way? Our own requirements are GCC first, clang later, but there are other people using the parser mostly on clang code. We don't have much attribute support currently, but if we add more that might become relevant.

tools/c-parser/StrictC.grm
```diff
@@ -997,6 +999,13 @@ declaration
                  end) *)
 | declaration_specifiers init_declarator_list YSEMI
                 (make_declaration declaration_specifiers init_declarator_list)
+| static_assert_declaration ([static_assert_declaration])
+
+(* 6.7.10 and/or §A.2.2 ``static_assert-declaration: _Static_assert ( constant-expression , string-literal ) ;'' *)
+(* constant-expression is modelled as rexpression (c.f. "enumerator") *)
+static_assert_declaration
+: STATIC_ASSERT LPAREN rexpression YCOMMA cstring_literal RPAREN YSEMI
+                (wrap(StaticAssertion { condition = rexpression, message = cstring_literal }, eleft rexpression, eright rexpression))
```
midnightveil, line 1008: I just realised that I forgot to go back and change this, the span for this should probably be STATIC_ASSERTleft and RPARENright not eleft/eright of the rexpression.

Xaphiosis: You're quoting from a C standard; you should say which one, otherwise people will expect C99.

Also, lines too long; this file tries to be 80 wide modulo minor spillage.

tools/c-parser/StrictC.grm
```diff
@@ -1187,6 +1196,8 @@ struct_declaration
                          left basetype, right struct_declarator_list),
                     sdecls)
                  end)
+(* per the C11 spec this should be allowed but let's not for simplicity *)
+(* | static_assert_declaration () *)
```
Xaphiosis, line 1200: What is the issue here? If we're putting functionality in, but not putting in all of it, would be good to know why.

midnightveil: Nothing really, and it's probably pretty easy to do.

I was leaving it out because the kernel doesn't use `_Static_assert` in these positions (i.e. inside a struct) so I didn't see the point (I only learned that this was even possible reading the C spec last year or so, it's not commonly known either).

midnightveil:
I'm going to close this since it seems unlikely that it's going to be merged. And can always reopen later.

Issue #919, opened 2025-08-22
Document annoying `named_theorems` interaction shifting default fact names from `Arch.` to `global.`


Xaphiosis:
In Haskell, the same const name pops up for both generic and Arch (with the former usually calling the latter, e.g. `capUntypedPtr` and `Arch.capUntypedPtr`. However, this leads to ambiguities and the global name not being easily accessible from within the `Arch` locale.

We perform a requalification dance to disambiguate them as `global.` and `Arch.`, after which the default `capUntypedPtr` constant always refers to the global one, `global.capUntypedPtr` to the global one, and `Arch.capUntypedPtr` to the arch-specific one.

The logical next step for this was when dealing with these constants' definitions, i.e. `global.capUntypedPtr_def` pointing to the global definition, `Arch.capUntypedPtr_def` the local definition and the default `capUntypedPtr_def` being the global definition as well. Unfortunately, the default case is broken by `named_theorems` behaviour.

### The problem:

```
context Arch begin arch_global_naming

thm capUntypedPtr_def (* global.capUntypedPtr_def as expected *)

named_theorems SomeThy_assms

thm capUntypedPtr_def (* Arch.capUntypedPtr_def very surprisingly! *)

end

context Arch begin arch_global_naming

thm capUntypedPtr_def (* global.capUntypedPtr_def again! *)

end
```

See also discussion on https://github.com/seL4/l4v/pull/911

### To be done:

- [x] update `doc/arch-split.md` to include `global./Arch.` disambiguation for facts, a warning about `named_theorems`, and a guideline to say `global.` and `Arch.` when unfolding both
- [x] update `lib/test/Requalify_Test.thy` to include const name disambiguation, as well as a demonstration of the `named_theorems` effect

### Other thoughts:

Placing the named_theorems by itself into an Arch context block does address the issue, but it takes 2s on a fast machine to process that block. @lsf37 did not like this one either

Figuring out  why exactly this happens, and manufacturing a `named_theorems` that doesn't do it. Unfortunately I was not able to figure out the reason for why this is happening.

The "please put `global.` and `Arch.` when unfolding both defs" suggestion can't be enforced except via manual review, and some violations of it already appear in the first arch-split pass as a result.

Xaphiosis:
Digging some more into the ML code, it turns out that what I thought was the culprit wasn't. One of the functions of `named_theorems xxx` is to declare an attribute `xxx` such that we can declare lemmas to be in it, or remove lemmas from it. The code that does this is:

[...]

PR #921, opened 2025-08-22
arch-split: fill out facts+consts disambiguation using `global.`, document `named_theorems` anomaly

Xaphiosis:
See https://github.com/seL4/l4v/issues/919 for a description of what's gone wrong, but as this is a documentation PR, hopefully the documentation and examples should explain it. Please let me know if they do, and how they can be improved.

Maybe at some point we'll figure out why named_theorems does this and have a work-around, but for the moment we can only wave yellow flags. Fortunately, this only kicks in during Refine, which is why we hadn't come across it before.

I'm also taking the opportunity to make the indentation policy change on `proof` after `interpretation` that @lsf37 asked for some time ago.


docs/arch-split.md
```diff
@@ -432,19 +438,29 @@ end
...
 term global.deriveCap (* arch-generic alternative *)
 term Arch.deriveCap   (* arch-specific *)
+thm deriveCap_def     (* arch-generic *)
+end
+
+(* For facts, there is an additional caveat: named_theorems perturbs fact name ordering: *)
+context Arch begin
+thm deriveCap_def     (* arch-generic, as expected *)
+named_theorems Some_assms
+thm deriveCap_def     (* arch-specific, until the end of the context block! *)
```
lsf37, line 463: Yeah, seeing it pointed out like this is good. And weird. Really weird.

docs/arch-split.md
```diff
@@ -472,6 +488,20 @@ arch-generic thing can be found with *either* an unqualified name or a
...
+As shown above, for facts, `named_theorems` causes a reordering of fact names,
+meaning that the default name becomes arch-specific in an `Arch` locale or
+interpretation. For this reason, when dealing with fact names which have been
+disambiguated with a `global./Arch.` prefix, we strongly suggest qualifying
+both fact names when unfolding them inside of an Arch locale or interpretation:
+
+```isabelle
+context Arch begin
+(* ... *)
+  apply (simp add: global.deriveCap_def Arch.deriveCap_def)
```
lsf37, line 500: 👍

lsf37 (approved):
Thank you, this will help understand what is going on in the future.

(And thanks for the `proof` indent -- I should put something general about that in the style doc somewhere at some point)

PR #922, opened 2025-08-26

corlewis:
The main new thing from these commits was removing `valid_ioports'` and folding `valid_ioports` into `valid_arch_state`.

### DIFF
--- a/proof/invariant-abstract/Ipc_AI.thy
+++ b/proof/invariant-abstract/Ipc_AI.thy
@@ -446,7 +446,7 @@ end
 
 (* FIXME: can some of these assumptions be proved with lifting lemmas? *)
 locale Ipc_AI_2 = Ipc_AI state_ext_t
-  for state_ext_t :: "'state_ext::state_ext itself"
+  for state_ext_t :: "'state_ext::state_ext itself" +
   assumes is_derived_cap_rights [simp]:
     "\<And>m p R c. is_derived m p (cap_rights_update R c) = is_derived m p c"
   assumes data_to_message_info_valid:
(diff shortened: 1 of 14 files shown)


proof/invariant-abstract/TcbAcc_AI.thy
```diff
@@ -2540,8 +2532,8 @@ lemma thread_set_domain_fault_tcbs_valid_states[wp]:
   "thread_set_domain t d \<lbrace> fault_tcbs_valid_states \<rbrace>"
   by (simp add: thread_set_domain_def thread_set_fault_tcbs_valid_states_trivial)
 
-lemma thread_set_domain_invs[wp]:
-  "thread_set_domain t d \<lbrace>invs\<rbrace>"
+lemma (in TcbAcc_AI) thread_set_domain_invs[wp]:
+  "thread_set_domain t d \<lbrace>invs :: 'state_ext state \<Rightarrow> _\<rbrace>"
```
Xaphiosis, line 2536: can you clarify why the state type annotations are now needed? I'm seeing a fair few of them

proof/invariant-abstract/Ipc_AI.thy
```diff
@@ -446,7 +446,7 @@ end
 
 (* FIXME: can some of these assumptions be proved with lifting lemmas? *)
 locale Ipc_AI_2 = Ipc_AI state_ext_t
-  for state_ext_t :: "'state_ext::state_ext itself"
+  for state_ext_t :: "'state_ext::state_ext itself" +
```
Xaphiosis, line 449: strange. was this a broken outcome of the merge?

proof/invariant-abstract/TcbAcc_AI.thy
corlewis: The commits being merged in from master moved `thread_set_valid_arch_state` into the `TcbAcc_AI` locale and changed it to use the fixed `state_ext` type from it, which meant that these `thread_set_domain` and `thread_set_priority` lemmas needed it as well.

proof/invariant-abstract/Ipc_AI.thy
corlewis: Yep, I made a mistake resolving a conflict and fixing it here was easier than amending the merge commit to fix it there.

corlewis:
> I only reviewed the "rt proof: update after merge" commit. That part looks fine to me. As usual, no clue how to review the rest of a merge. You cleared the conflicts as part of the merge commit, right? Also, I'm guessing that commit message is the usual after merges, and no point adding anything more informative?

That's right, my usual approach is to fix all of the conflicts as best as I can during the merge and then go through and fix anything broken in the commit afterwards.

Looking at past merges it doesn't look like we've added anything to the post-merge commit message. I guess I could try to summarise what has been merged in and required fixing but that could be pretty difficult in larger merges like we've done in the past.

Xaphiosis:
> Looking at past merges it doesn't look like we've added anything to the post-merge commit message. I guess I could try to summarise what has been merged in and required fixing but that could be pretty difficult in larger merges like we've done in the past.

The question is more whether it's worth it in the long run, and what they'll ultimately appear like after the commit linearisation procedure. I guess if something particularly interesting happened in the merge it might be worth taking note, but this one is pretty vanilla. I'm more thinking about the more complex ones like Invariants arch-split.

PR #923, opened 2025-08-31
Verification support for remaining Arm platforms

lsf37:
Add support for `omap3`, `am335x`, and `bcm2837`. This bring Arm platform verification support to 100%.

A number of small kernel changes are necessary for this -- see https://github.com/seL4/seL4/pull/1510

On the verification side:

- new machine interface function `handleSpuriousIRQ_mop`
- platform constants for the 3 platforms above
- new `haveSetTrigger` option (defined, not generated, because it is not a cmake config option, just a macro defined in some particular headers and not in others)
- new `isGICPlatform` option to distinguish these from all other Arm platforms. The other platforms implement Arm's IRQ controller spec, these new ones do not.
- `pptrBase` (= `kernel_base` in ASpec) becomes configurable, although only two values are possible and unlikely that further values will appear (previously only `0xe0000000`, now `0xe0000000` and `0xf0000000`).

The proof updates are split into two larger chunks:

- pptrBase, haveSetTrigger, and numSGIs changes -- these affect mostly ARM and ARM_HYP and are reasonably routine.
- `maybe_handle_interrupt`/`maybeHandleInterrupt` which call `handleSpuriousIRQ`. These affect all architectures, including generic proofs and are required because the infoflow proofs demand that interrupt handing at kernel entry and interrupt handling at preemption are equal. This used to be true only because the body of `handleSpuriousIRQ` was empty in C. This is no longer the case on `omap3` et al, which means a change to the preemption handler was necessary. This is large, but mostly benign and even made some CRefine proofs nicer, because I factored out that code into its own function (`maybeHandleInterrupt`) which can be reused everywhere. 

Test with: https://github.com/seL4/seL4/pull/1510


spec/abstract/ARM/Arch_A.thy
```diff
@@ -22,7 +22,7 @@ fun
   arch_invoke_irq_control :: "arch_irq_control_invocation \<Rightarrow> (unit,'z::state_ext) p_monad"
 where
   "arch_invoke_irq_control (ArchIRQControlIssue irq handler_slot control_slot trigger) = without_preemption (do
-    do_machine_op $ setIRQTrigger irq trigger;
+    when haveSetTrigger $ do_machine_op $ setIRQTrigger irq trigger;
```
corlewis, line 25: Isn't this already checked in the decode? Do we do it again here to avoid having to pass it through `valid_invocation`?

spec/haskell/src/SEL4/Machine/Hardware/ARM.lhs
```diff
@@ -912,6 +915,14 @@ Note that InvalidIOPDE and InvalidPTE do not exist in C, as there is no valid bi
 > config_ARM_GIC_V3 :: Bool
 > config_ARM_GIC_V3 = error "generated from CMake config"
 
+> -- Wether the handleSpuriousIRQ machine op is available
```
corlewis:
```suggestion
> -- Whether the handleSpuriousIRQ machine op is available
```

proof/crefine/ARM/Finalise_C.thy
```diff
@@ -678,6 +678,7 @@ crunch updateRestartPC
   and valid_objs'[wp]: valid_objs'
   and tcb_at'[wp]: "tcb_at' p"
 
+(* short name target_' can be ambiguous for some platforms, use full name instead below *)
```
corlewis, line 681: That must have been painful to figure out, thanks for the comment explaining it.

proof/crefine/ARM/Interrupt_C.thy
```diff
@@ -502,10 +519,19 @@ lemma plat_SGITargetValid_spec:
...
               split: if_split)
   done
 
+lemma valid_tcb_state'_Restart:
+  "valid_tcb_state' Restart s"
+  by (simp add: valid_tcb_state'_def)
+
+lemma ct_active_st_tcb_at_runnable':
+  "ct_active' s \<Longrightarrow> st_tcb_at' runnable' (ksCurThread s) s"
+  by (fastforce simp: ct_in_state'_def elim!: pred_tcb'_weakenE)
```
corlewis: `FIXME: move` candidates?

proof/drefine/CNode_DR.thy
```diff
@@ -1594,63 +1601,6 @@ lemma opt_cap_pd_None:
   by (clarsimp simp: opt_object_page_directory obj_at_def slots_of_def unat_map_def
     opt_cap_def invs_def valid_state_def object_slots_def transform_page_directory_contents_def)+
 
-lemma transform_pde_NullCap:
-  "\<lbrakk>3584 \<le> unat (xa::word32); unat xa < 4096\<rbrakk> \<Longrightarrow>
-            transform_pde (kernel_pde_mask ptx (ucast xa)) = cdl_cap.NullCap"
```
corlewis, line 1599: Out of curiosity, why can this and the following empty_slot lemmas be removed? They don't seem related to anything being changed.

proof/invariant-abstract/ARM_HYP/ArchVCPU_AI.thy
```diff
@@ -606,6 +606,11 @@ lemma handle_recv_valid_cur_vcpu[wp]:
   unfolding handle_recv_def Let_def ep_ntfn_cap_case_helper delete_caller_cap_def
   by (wpsimp wp: hoare_drop_imps)
 
+lemma maybe_handle_interrupt_valid_cur_vcpu[wp]:
+  "\<lbrace>valid_cur_vcpu and invs\<rbrace> maybe_handle_interrupt inKernel \<lbrace>\<lambda>_. valid_cur_vcpu\<rbrace>"
```
corlewis: It would be good to be consistent here with AARCH64 and change one of the lemma statements. AARCH64 is a multiline hoare triple and uses in_kernel instead of inKernel.

proof/invariant-abstract/AARCH64/ArchInterruptAcc_AI.thy
```diff
@@ -27,6 +27,9 @@ lemma dmo_maskInterrupt_invs [InterruptAcc_AI_assms]:
                         cur_tcb_def valid_irq_states_def valid_irq_masks_def)
   done
 
+crunch handle_spurious_irq
+  for invs: invs
```
corlewis, line 31: I guess it wasn't needed anywhere, but `[wp]`?

proof/refine/AARCH64/Syscall_R.thy
```diff
@@ -1726,6 +1726,28 @@ lemmas handleReply_ct_in_state_simple[wp] =
                                      handleReply_st_tcb_at_simple']
 
 
+lemma handleSpuriousIRQ_corres[corres]:
+  "corres dc \<top> \<top> handle_spurious_irq handleSpuriousIRQ"
+  by (simp add: handle_spurious_irq_def handleSpuriousIRQ_def)
+
+lemma contract_all_imp_strg:
+  "P \<and> P' \<and> (\<forall>x. R x \<longrightarrow> Q x) \<Longrightarrow> \<forall>x. R x \<longrightarrow> P \<and> Q x \<and> P'"
+  by blast
```
corlewis, line 1735: This seems to be duplicated here and in Refine.thy (and the same in ARM_HYP).

proof/refine/ARM_HYP/Interrupt_R.thy
```diff
@@ -1030,25 +1035,26 @@ lemma handle_reserved_irq_corres[corres]:
 
 lemma handleInterrupt_corres:
   "corres dc
-     einvs
-     (invs' and (\<lambda>s. intStateIRQTable (ksInterruptState s) irq \<noteq> IRQInactive) and
-       (\<lambda>s. irq \<in> non_kernel_IRQs \<longrightarrow> sch_act_not (ksCurThread s) s))
+     (einvs) (invs' and (\<lambda>s. intStateIRQTable (ksInterruptState s) irq \<noteq> IRQInactive))
```
corlewis:
```suggestion
     einvs (invs' and (\<lambda>s. intStateIRQTable (ksInterruptState s) irq \<noteq> IRQInactive))
```

proof/refine/ARM_HYP/Interrupt_R.thy
```diff
@@ -1030,25 +1035,26 @@ lemma handle_reserved_irq_corres[corres]:
...
      (handle_interrupt irq) (handleInterrupt irq)"
-  (is "corres dc _ (invs' and _ and ?P') _ _")
+  (is "corres dc (einvs) ?P' ?f ?g")
```
corlewis:
```suggestion
  (is "corres dc _ ?P' ?f ?g")
```

proof/crefine/ARM_HYP/Refine_C.thy
```diff
@@ -242,6 +258,44 @@ lemma dmo'_getActiveIRQ_non_kernel[wp]:
...
+lemma handleInterrupt_epilogue:
+  "ccorres dc xfdc invs' UNIV []
+           ((liftE (doMachineOp (getActiveIRQ True) >>=
+                    case_option handleSpuriousIRQ handleInterrupt)))
+           (\<acute>irq___unsigned_long :== CALL getActiveIRQ();;
+            IF \<acute>irq___unsigned_long \<noteq> kernel_all_global_addresses.irqInvalid THEN
+              CALL handleInterrupt(\<acute>irq___unsigned_long)
+            ELSE
+              CALL handleSpuriousIRQ()
+            FI)"
```
corlewis: Out of curiosity, do you remember why this is only needed in ARM_HYP?

corlewis (approved):
Very nice, and a lot more work involved than it had previously sounded like! Thank you for all the cleaning up so many proofs along the way as well.

One minor thing is that the commit message of 'infoflow: updates for pptrBase and haveSetTrigger' needs some changes. I think its first sentence should say "adjust for code variation *of* non-gic platforms", and there's a typo with ompa3 instead of omap3.

Even more minor and probably more a thing of personal preference, I think some of the commits could be combined. At the very least, I don't think that having all of the maybe_handle_interrupt commits separated per session gains us much.

spec/machine/ARM/Platform.thy
```diff
@@ -76,9 +95,23 @@ definition cacheLineBits :: nat where
...
   "cacheLine = 2^cacheLineBits"
 
-(* The first virtual address of the kernel's physical memory window *)
+(* The first virtual address of the kernel's physical memory window.
+   Coincides with seL4_UserTop and USER_TOP in C for AArch32.
+
+   The definition below singles out the omap3 platform, because it is the only AArch32
+   platform that does not put seL4_UserTop at 0xe0000000. The macro definition in C is
+   hard to get at for the kernel config generation script and the definition here is
+   comparatively straightforward. *)
```
Xaphiosis, line 104: a potential concern: if this changes on the C side, will we pick them up? I'm guessing only during CRefine, which would be ok since we don't expect these to change.

spec/cspec/c/gen-config-thy.py
```diff
@@ -71,18 +71,18 @@
...
+    'CONFIG_PLAT_BCM2837': (bool, "config_PLAT_BCM2837"),
     'CONFIG_PLAT_EXYNOS4': (bool, None),
     'CONFIG_PLAT_EXYNOS5': (bool, None),
     'CONFIG_PLAT_HIKEY': (bool, None),
     'CONFIG_PLAT_IMX6': (bool, None),
     'CONFIG_PLAT_IMX7_SABRE': (bool, None),
     'CONFIG_PLAT_IMX8MQ_EVK': (bool, None),
     'CONFIG_PLAT_IMX8MM_EVK': (bool, None),
-    'CONFIG_PLAT_OMAP3': (bool, None),
+    'CONFIG_PLAT_OMAP3': (bool, "config_PLAT_OMAP3"),
```
Xaphiosis, line 85: Could you elaborate on "so we export the platform config instead of trying to figure out pptrBase from the C source" (from the commit message)? What does "export the platform config" mean/imply?

spec/abstract/AARCH64/ArchInterrupt_A.thy
```diff
@@ -87,6 +87,10 @@ definition arch_mask_irq_signal :: "irq \<Rightarrow> (unit,'z::state_ext) s_mon
   "arch_mask_irq_signal irq \<equiv>
      when (\<not>Kernel_Config.config_ARM_GIC_V3) (do_machine_op $ maskInterrupt True irq)"
 
+definition handle_spurious_irq :: "(unit,'z::state_ext) s_monad"
+  where
```
Xaphiosis: doesn't the where fit on the same line?

spec/abstract/Interrupt_A.thy
```diff
@@ -103,4 +104,13 @@ definition
    do_machine_op $ ackInterrupt irq
   od"
 
+definition
+  maybe_handle_interrupt :: "bool \<Rightarrow> (unit, 'z::state_ext) s_monad" where
```
Xaphiosis: style: should go in one line with `definition`

spec/abstract/X64/ArchInterrupt_A.thy
```diff
@@ -28,6 +28,10 @@ definition arch_mask_irq_signal :: "irq \<Rightarrow> (unit,'z::state_ext) s_mon
   where
   "arch_mask_irq_signal irq \<equiv> do_machine_op $ maskInterrupt True irq"
 
+definition handle_spurious_irq :: "(unit,'z::state_ext) s_monad"
+  where
```
Xaphiosis, line 32: style: should go in one line with `definition`

spec/abstract/RISCV64/ArchInterrupt_A.thy
```diff
@@ -25,6 +25,10 @@ definition arch_mask_irq_signal :: "irq \<Rightarrow> (unit,'z::state_ext) s_mon
   where
   "arch_mask_irq_signal irq \<equiv> return ()"
 
+definition handle_spurious_irq :: "(unit,'z::state_ext) s_monad"
+  where
```
Xaphiosis, line 29: style: should go in one line with `definition`

proof/crefine/ARM/Retype_C.thy
```diff
@@ -3634,31 +3634,29 @@ lemma copyGlobalMappings_ccorres:
...
    apply (rule ccorres_rel_imp)
     apply (rule_tac F="\<lambda>_ s. globalPD = armKSGlobalPD (ksArchState s)
-                                \<and> is_aligned globalPD pdBits \<and> valid_pde_mappings' s
-                                \<and> page_directory_at' pd s
-                                \<and> page_directory_at' (armKSGlobalPD (ksArchState s)) s"
-                and i="0xE00"
+                             \<and> is_aligned globalPD pdBits \<and> valid_pde_mappings' s
+                             \<and> page_directory_at' pd s
+                             \<and> page_directory_at' (armKSGlobalPD (ksArchState s)) s"
+                and i="unat pptrBase >> unat ARMSectionBits"
```
Xaphiosis, line 3645: I'm grateful for these magic numbers becoming known

proof/crefine/ARM_HYP/Interrupt_C.thy
```diff
@@ -724,33 +783,19 @@ lemma Arch_decodeIRQControlInvocation_ccorres:
...
-   apply (simp (no_asm) add: numeral_2_eq_2 numeral_3_eq_3)
-  (* ARMIRQIssueSGISignal C *)
-  apply (clarsimp simp: neq_Nil_conv[where xs=extraCaps])
-  apply (drule interpret_excaps_eq[rule_format, where n=0], simp)
-  apply (subst numeral_3_eq_3)
-  apply (subst numeral_2_eq_2)
-  apply (fastforce simp: rf_sr_ksCurThread ucast_ucast_mask)
-  done
+  apply (prop_tac "0 < numSGIs \<longrightarrow> args ! 0 \<le> word_of_nat numSGIs - 1 \<longrightarrow>
+                   unat (args ! 0) < unat (word_of_nat numSGIs::machine_word)")
```
Xaphiosis: nitpick: nicer to have spaces around `::` when there's multiple words to the left of it, otherwise it reads as `numSGIs::machine_word`

proof/crefine/ARM/Refine_C.thy
```diff
@@ -88,7 +89,7 @@ proof -
     apply (solves clarsimp)
    apply (wp getActiveIRQ_le_maxIRQ | simp)+
   apply (clarsimp simp: invs'_def valid_state'_def)
-  done
+  sorry (* handleSpuriosIRQ *)
```
Xaphiosis: sorry!

proof/crefine/ARM/Interrupt_C.thy
Xaphiosis: also ucast_ucast_machine_word_irq above

proof/crefine/ARM/Refine_C.thy
corlewis: I had the same reaction, the sorry goes away in a later commit. There's enough going on that it does make sense to keep it separated into two chunks as described in the PR description, but maybe the sorry should be flagged in the commit message.

proof/infoflow/refine/ARM/Example_Valid_StateH.thy
```diff
@@ -2159,21 +2159,32 @@ lemma valid_caps_s0H[simp]:
...
             simp add: objBitsKO_def,
            simp add: objBitsKO_def s0_ptrs_aligned mask_def,
           rule pspace_distinctD'[OF _ s0H_pspace_distinct', simplified s0H_internal_def],
           simp)+
 
+lemma init_objs_base_neq_0[simp]:
+  "init_objs_base \<noteq> 0"
+  by (simp add: init_objs_base_def)
+
+lemma aligned_addrFromPPtr_kernel_base[simplified, simp]:
```
Xaphiosis: it's strange to see this: can't imagine what gets simplified here, and since the state is hidden, no answers are found here

proof/infoflow/ARM/ArchNoninterference.thy
```diff
@@ -375,6 +375,36 @@ lemma dmo_getActive_IRQ_reads_respect_scheduler[Noninterference_assms]:
...
+
+lemma dmo_handleSpuriousIRQ_mop_scheduler_affects_equiv[wp]:
+  "do_machine_op handleSpuriousIRQ_mop \<lbrace>scheduler_affects_equiv aag l st\<rbrace>"
+  unfolding handleSpuriousIRQ_mop_def scheduler_affects_equiv_def
+  apply wp_pre
+   apply (wps | wp do_machine_op_mol_states_equiv_for hoare_vcg_const_imp_lift)+
+   apply (simp add: scheduler_globals_frame_equiv_def)
+   apply (wpsimp wp: do_machine_op_mol_states_equiv_for hoare_vcg_const_imp_lift
+                 simp: arch_scheduler_affects_equiv_def)
+  apply (clarsimp simp: arch_scheduler_affects_equiv_def)
```
Xaphiosis, line 398: could add a + to previous line

Xaphiosis:
Very nice work. Especially how things are documented. Before going in, I only really knew that there was an infoflow problem related to spurious interrupts, and very little about the extra arches.
There is still a sorry left, so I shouldn't approve yet.

Commit question time:
- arm+arm-hyp aspec: make pptrBase configurable; add haveSetTrigger
  - this should say "use haveSetTrigger" as it doesn't add it, had me confused
  - "C is written underflow safe and will therefore not match the range_check,
    but it will return an unintuitive error message."
    - I don't understand: the C will return an unintuitive error message? It
      doesn't match range_check (which is the spec)?

Now, commit nitpick time:
- the "maybe_handle_interrupt" series of commits is a bit much
  - ainvs, refine are sizeable (refine includes orphanage)
  - at minimum I'd squash sep-abstract with drefine, infoflow with infoflow c,
    possibly even access with infoflow
- the "infoflow C" tag has been mostly seen as "infoflow refine", or wrapped
  into "infoflow" and ignored; "infoflowc" was another invention someone did,
  and "infoflow C" is new; perhaps "infoflow refine" then?
- you write "gic" sometimes (e.g. in "make haveSetTrigger available", do we
  want it to be GIC?
- "arm+arm-hyp ainvs: pptrBase, haveSetTrigger, numSGIs": "majority of the
  changes *are*"

proof/crefine/ARM/Refine_C.thy
Xaphiosis: Crap, I missed that. Am not used to people committing sorries. Still unsure if this is a good idea. Maybe it's unavoidable, but I'll wait for Gerwin to let us know what happened here.

spec/cspec/c/gen-config-thy.py
lsf37: I mean that we export the config option that says "the platform is OMAP3" instead of a config option that says "pptrBase = x" (which would be nicer, but unfortunately only exists as a macro in some of the platform C headers, not as a config option)

proof/crefine/ARM_HYP/Refine_C.thy
lsf37: Oh, well spotted, this should be removed. It's a remnant of how I did the spurious IRQ thing before refactoring.

proof/drefine/CNode_DR.thy
lsf37: They were just unused and had I saw them because they were full of magic numbers.

spec/abstract/ARM/Arch_A.thy
lsf37: Yes, it's mostly so the lemmas about it can have fewer preconditions.

proof/invariant-abstract/ARM_HYP/ArchVCPU_AI.thy
lsf37: Agreed. `in_kernel` would be correct for the abstract invariants. Not sure how this variation crept in.

proof/invariant-abstract/AARCH64/ArchInterruptAcc_AI.thy
lsf37: We could, but I wanted to see where it was actually needed instead of the proof skipping over it. Turns out it's not that many occurrences, so I left it out of `wp`.

spec/machine/ARM/Platform.thy
lsf37: Yes, that's right. We'll pick it up, but only in CRefine. Given that it seems to be standardised for all other platforms that should hopefully be a rare enough occurrence.

spec/abstract/Interrupt_A.thy
lsf37: This was to fit with the style of the definition above, but looking at the file, it's already inconsistent anyway, so we might as well use the new style.

spec/abstract/RISCV64/ArchInterrupt_A.thy
lsf37: In this one, it would actually make the style in this file inconsistent (the rest consistently uses the old style). Happy to change it, but I thought we wanted to keep consistent where possible.

spec/abstract/X64/ArchInterrupt_A.thy
lsf37: (same as for RISCV64)

proof/crefine/ARM/Refine_C.thy
lsf37: Hm, good that you spotted it -- this was originally supposed to disappear in a squash. Need to check where exactly.

proof/infoflow/ARM/ArchNoninterference.thy
lsf37: We could, but that seems a bit unstable to me. If the proof state changes to something with a Hoare triple, `wpsimp` will go and make more progress probably leaving something less recognisable, whereas clarsimp should just stop.

proof/infoflow/refine/ARM/Example_Valid_StateH.thy
lsf37: The `simplified` is acting on the `pageBitsForSize ARMSection` (which has just 20 later in the big goals where things are unfolded). Will add a comment.

proof/crefine/ARM/Refine_C.thy
lsf37: Right, Corey is correct of course. It does need to be sorried in this commit, because the function occurs in C, but not yet in the spec. Will add to the commit message.

lsf37:
> "C is written underflow safe and will therefore not match the range_check,
> but it will return an unintuitive error message."
> * I don't understand: the C will return an unintuitive error message? It
>   doesn't match range_check (which is the spec)?

C does return an unintuitive error message and it does not match the range_check (which is indeed the spec). The spec is wrong in this case, because it underflows and produces no error message, whereas C does not underflow. But the error message that comes out of C in this case is also not great, hence the extra checking and extra error message.

Will rephrase.

lsf37:
> * you write "gic" sometimes (e.g. in "make haveSetTrigger available", do we
>   want it to be GIC?

Yes, GIC would be correct according to Arm, we should use that one.

Xaphiosis:
> I think I have addressed everything now and this is ready to merge. Any last wishes before I hit the button?

I've had a quick look, and don't see anything else to do, but there are still 24+ commits. Any thoughts the squashing a few idea?

spec/abstract/X64/ArchInterrupt_A.thy
Xaphiosis: Um, did this go anywhere?

proof/crefine/ARM_HYP/Interrupt_C.thy
Xaphiosis: You marked this as resolved and thumbs up, but I'm not seeing it changed?

lsf37: Hm, I definitely remember changing it. Let me see

lsf37: Alright, I know what happened: this occurs multiple times, and I fixed it at the first occurrence, but not this one. Will fix.

spec/abstract/X64/ArchInterrupt_A.thy
lsf37: Sorry, that was a bit cryptic (I thought the other occurrence would show up first) -- I wanted to leave it as is because the old style is used consistently in that file.

proof/crefine/ARM_HYP/Interrupt_C.thy
lsf37: And now it properly shows up as fixed.

PR #924, opened 2025-09-09

corlewis:
* refine: arch-split up to InvariantUpdates_H
* ainvs: refactor arch_thread_get/set lemmas
* refine+crefine: remove valid_arch_mdb_ctes
* crefine: adjust for config_set wrapper in C


corlewis:
> This looks good (modulo ARM AInvs which is currently failing, but should not).

Whoops, forgot about ARM AInvs. It's fixed now, with the commit messages updated as well for the relevant changes.

corlewis:
> Have also looked at the CI changes which seem to have come across fine. I've updated the `rt` branch requirements accordingly. This means this PR needs to be merged before any other `rt` PRs can be merged.

Thanks for thinking to do this. We'll presumably need to update them again when we get up to merging the Isabelle2025 update in.

proof/refine/RISCV64/TcbAcc_R.thy
```diff
@@ -5296,73 +5308,64 @@ crunch setQueue, addToBitmap, removeFromBitmap
...
-lemma tcbSchedPrev_update_tcb_cte_cases:
-  "(a, b) \<in> ran tcb_cte_cases \<Longrightarrow> a (tcbSchedPrev_update f tcb) = a tcb"
+lemma update_tcb_cte_cases:
+  "\<And>f. (a, b) \<in> ran tcb_cte_cases \<Longrightarrow> a (tcbInReleaseQueue_update f tcb) = a tcb"
+  "\<And>f. (a, b) \<in> ran tcb_cte_cases \<Longrightarrow> a (tcbQueued_update f tcb) = a tcb"
+  "\<And>f. (a, b) \<in> ran tcb_cte_cases \<Longrightarrow> a (tcbSchedNext_update f tcb) = a tcb"
+  "\<And>f. (a, b) \<in> ran tcb_cte_cases \<Longrightarrow> a (tcbSchedPrev_update f tcb) = a tcb"
+  "\<And>f. (a, b) \<in> ran tcb_cte_cases \<Longrightarrow> a (tcbYieldTo_update f tcb) = a tcb"
+  "\<And>f. (a, b) \<in> ran tcb_cte_cases \<Longrightarrow> a (tcbSchedContext_update f tcb) = a tcb"
+  "\<And>f. (a, b) \<in> ran tcb_cte_cases \<Longrightarrow> a (tcbBoundNotification_update f tcb) = a tcb"
```
Xaphiosis: good idea. is this only worth it on MCS? I don't recall seeing anything like it on master

proof/refine/RISCV64/TcbAcc_R.thy
```diff
@@ -3592,6 +3600,8 @@ lemma sts'_valid_mdb'[wp]:
   "setThreadState st t \<lbrace>valid_mdb'\<rbrace>"
   by (wpsimp simp: valid_mdb'_def)
 
+context begin interpretation Arch . (*FIXME: RT arch-split*)
```
Xaphiosis: strategy observation on these: `context begin interpretation Arch .` is an expensive operation, it can take 7-8s on fast machines. Adding a bunch of them, even if the plan is to fix them later, is sadness.

I'd recommend doing this instead:
```

lemma tcbSchedNext_update_valid_objs'[wp]:
  ...

end

arch_requalify_facts tcbSchedNext_update_valid_objs' (* FIXME: arch-split RT *)
declare tcbSchedNext_update_valid_objs'[wp]

```

This is significantly faster (context Arch is around 2s), but still stands out as weird and so noticeable for fixing.

Xaphiosis: I don't know whether `FIXME: RT arch-split` will cause us to miss grepping for `FIXME: arch-split` in future, or whether `FIXME: arch-split RT` will cause misses on searches for `FIXME: RT`. Which do you think is more likely?

proof/refine/RISCV64/ArchInvsLemmas_H.thy
```diff
@@ -347,35 +345,103 @@ lemma hyp_refs_of_live':
...
-  (case cap of NullCap                         \<Rightarrow> False
-             | DomainCap                       \<Rightarrow> False
-             | IRQControlCap                   \<Rightarrow> False
-             | IRQHandlerCap irq               \<Rightarrow> False
-             | ReplyCap r m cr                 \<Rightarrow> False
-             | ArchObjectCap ASIDControlCap    \<Rightarrow> False
-             | _                               \<Rightarrow> True)"
-  by (simp split: capability.splits arch_capability.splits zombie_type.splits)
 
+(* FIXME RT arch-split: whole typ_at_lift section needs to be split *)
```
Xaphiosis: bikeshed: could I ask that we consistently have `(* FIXME` instead of `(*FIXME` ?

Xaphiosis: and `FIXME RT arch-split` vs `FIXME: RT arch-split` ... consider making an abbreviation in jEdit: in `.isabelle/etc/jedit/abbrevs` I have:
```
FA|(* FIXME arch-split: \| *)
```

proof/refine/RISCV64/ArchInvsDefs_H.thy
```diff
@@ -188,5 +188,18 @@ definition archMakeObjectT :: "arch_kernel_object_type \<Rightarrow> kernel_obje
      of PTET \<Rightarrow> injectKO (makeObject :: pte)
       | ASIDPoolT \<Rightarrow> injectKO (makeObject :: asidpool)"
 
+
```
Xaphiosis: extra newline?

proof/invariant-abstract/ARM/ArchDetSchedSchedule_AI.thy
```diff
@@ -323,7 +324,7 @@ crunch arch_invoke_irq_handler, arch_mask_irq_signal, handle_reserved_irq
 
 lemma arch_invoke_irq_handler_valid_sched_pred_strong[wp]:
   "arch_invoke_irq_handler i \<lbrace> valid_sched_pred_strong P \<rbrace>"
-  by (cases i; wpsimp)
+  by (cases i; (wpsimp | rule conjI)+)
```
Xaphiosis: is this as a result of a misbehaving if splitting?

proof/refine/Invariants_H.thy
```diff
@@ -35,6 +35,9 @@ arch_requalify_facts (H)
   wordsOfTime_def
   parseTimeArg_def
 
+arch_requalify_facts
+ valid_sz_simps
```
Xaphiosis: indent should be 2

proof/refine/RISCV64/Invariants_H.thy
```diff
@@ -2560,80 +2490,6 @@ lemma typ_at_lift_valid_untyped':
...
-   assumptions. The following thm statements should fail if this is not true. *)
-private lemmas check_valid_internal = iffD1[OF refl, where P="valid p g q" for p g q]
-thm typ_at_lifts_all'[atomized, THEN check_valid_internal]
-end
-
-end
-
-(* we expect typ_at' and sc_at'_n lemmas to be [wp], so this should be easy: *)
-method typ_at_props' = unfold_locales; wp?
-
```
Xaphiosis: oof. sad to see so much leave generic space, but as you said the typ_at_lifts system is different

proof/refine/RISCV64/TcbAcc_R.thy
corlewis: Hmm, I hadn't realised it but it looks like some of these lemmas were in master as well. I added and used these more in this update to avoid needing to go into the Arch locale as often to access `objBits_defs`. I think master used a different approach but I don't see any reason not to backport this as well.

proof/refine/RISCV64/TcbAcc_R.thy
corlewis: Hmm, good question. My guess is that `FIXME: arch-split RT` will be more useful, for when master has been arch-split and we are doing an explicit arch-split pass through the rt branch.

proof/invariant-abstract/ARM/ArchDetSchedSchedule_AI.thy
corlewis: Sort of, it's intentional if splitting on if the configuration has GICv3 support. See https://github.com/seL4/l4v/pull/924/commits/de91ea2666c2040fd0f7db4439392feb174ef037 for an example from the original commit of handling the split this way.

proof/refine/RISCV64/Invariants_H.thy
corlewis: I hope that we'll be able to move a lot back to the generic space eventually, but it will need an extra locale and a careful interface.

michaelmcinerney (approved):
Looks good to me, but I'm not overly familiar with many of the trickier sections here, I have to say.

lsf37:
> > I thought we had some fancy `test with` syntax for PRs that lets the CI test with that version? Does this not work on PRs to `rt`?
> 
> My understanding is that that's only for testing with another PR to the seL4 repository, not for testing with specific commits.

That is correct, although it would not be so hard to add. There is an open issue for referencing branches in the CI repo, which is a bit harder, because branch naming makes it harder to parse, but commit hashes would be easier I think.

proof/refine/RISCV64/TcbAcc_R.thy
corlewis: Backported to master in https://github.com/seL4/l4v/pull/927.

proof/refine/RISCV64/KHeap_R.thy
```diff
@@ -3172,7 +3172,7 @@ lemma valid_refs'_def2:
   "valid_refs' R (ctes_of s) = (\<forall>cref. \<not>cte_wp_at' (\<lambda>c. R \<inter> capRange (cteCap c) \<noteq> {}) cref s)"
   by (auto simp: valid_refs'_def cte_wp_at_ctes_of ran_def)
 
-context begin interpretation Arch . (*FIXME: RT arch-split*)
+context begin interpretation Arch . (* FIXME: arch-split RT *)
```
Xaphiosis: can you clarify, is this a new `interpretation Arch .`? I think at this point in the merge queue KHeap_R hasn't been arch-split yet, but you're adding the RT tag. I'm a bit confused at what's happening.
Nonetheless, it would still be good to not introduce new instances of the slow interpretation.

proof/refine/RISCV64/TcbAcc_R.thy
```diff
@@ -3600,7 +3605,7 @@ lemma sts'_valid_mdb'[wp]:
   "setThreadState st t \<lbrace>valid_mdb'\<rbrace>"
   by (wpsimp simp: valid_mdb'_def)
 
-context begin interpretation Arch . (*FIXME: RT arch-split*)
+context begin interpretation Arch . (* FIXME: arch-split RT *)
```
Xaphiosis: same question as the other RT-tagged `interpretation Arch .`

proof/refine/RISCV64/ArchInvsLemmas_H.thy
```diff
@@ -519,7 +519,7 @@ locale typ_at_props' =
   assumes typ': "f \<lbrace>\<lambda>s. P (typ_at' T p' s)\<rbrace>"
 begin
 
-interpretation Arch . (*FIXME RT arch-split *)
+interpretation Arch . (* FIXME arch-split RT *)
```
Xaphiosis: same question as the other RT-tagged `interpretation Arch .`, plus one more in this file

Xaphiosis: oh wait, this one's a locale; so the question is only why is the tag RT, RT-only locale?

proof/refine/RISCV64/KHeap_R.thy
corlewis: Yes, this is a new `interpretation`. To be precise, at the corresponding point on master these lemmas were actually already in an `interpretation` block, but at some point in rt's history for whatever reason the block was shrunk. However, the arch-split process has made these lemmas depend on things in `Arch`, so I've added it back in.

In this case, I didn't put them properly into `Arch` and requalify because it will go away once the whole merge process is completed, and doing that would actually make it more different compared to master and make that merge slightly harder.

proof/refine/RISCV64/TcbAcc_R.thy
corlewis: I left this one as an `interpretation` because it's a large block and requalifying everything in it felt too painful. Also, like with the other one, this block is actually already in a larger `interpretation` on master, so doing it this way also keeps the future merge when this is arch-split simpler.

proof/refine/RISCV64/ArchInvsLemmas_H.thy
corlewis: Yes, these are in RT-only locales, which are part of the `typ_at_lifts` mechanism. Handling it properly will require arch-splitting this whole structure, which is being left for a later pass.

PR #925, opened 2025-09-26
arch-split CSpace_I and CSpace1_R

Xaphiosis:
Please review commit-by-commit. The commit titled "refine: arch-split CSpace1_R" has a longer message that explains some of what happened there and will be useful for reviewing the CSpace1_R split.

This was a lot of work, way more than expected. Results:

```
69 files changed, 13637 insertions(+), 42755 deletions(-)
```

For a total of 29118 lines removed. Most of this is from the massive CSpace1_R which (on AARCH64 for example) was 7.3k lines, and now ArchCSpace1_R only has 1.4k arch-specific lines.

The main obstacle to dealing with these arch-splits was how locales are used. I tried very hard to be able to say "fix this not-yet-existing constant and assume things about it" in one (generic) locale, and then combine it with another mdb-specific locale which also had to be split into generic and arch parts. It was possible to say this, but then trying to recombine it all into a locale that proofs actually expect was intractable. If the proofs were *in* locales, I could do it. However some of the proofs (cteInsert_corres and cteSwap_corres) in particular happily use locales like this (in Isar-style proofs you probably would `interpret` the locale):

```
   apply (rule mdb_swap.intro)
     apply (rule mdb_ptr.intro)
      apply (erule vmdb.intro)
     apply (erule mdb_ptr_axioms.intro)
    apply (rule mdb_ptr.intro)
     apply (erule vmdb.intro)
    apply (erule mdb_ptr_axioms.intro)
   apply (erule mdb_swap_axioms.intro)
```
and refer to specific results in specific locales, e.g. `mdb_swap.descendants`.

So after a long and painful exploration of what locales can and can't do, I had to revert a bunch of this work, move the affected constants (`is_derived'` and ingredients for `weak_derived'`) to the next theory up (`ArchCSpace_I`) and re-knit all the dependencies.

I am still splitting up and re-knitting locales by using a common ancestor with one of the locales holding generic proofs under Arch assumptions (e.g. CSpace1_R_2), so `mdb_insert_child` gets a `mdb_insert_child_gen`, and then the proofs from there get re-inserted into `mdb_insert_child` once the extra assumptions of `mdb_insert_child_gen` are satisfied (in the Arch* theory). I will think about how to write up this technique for the arch-split docs.

Feel free to ask me questions.


proof/refine/AARCH64/ArchCSpace_I.thy
```diff
@@ -11,699 +11,145 @@ theory ArchCSpace_I
...
-  by (rule ext, simp add: fresh_virt_cap_class_def)+
+named_theorems CSpace_I_2_assms
 
-lemma fresh_virt_cap_classD:
-  "\<lbrakk> m x = Some cte; fresh_virt_cap_class C m \<rbrakk>
-     \<Longrightarrow> C \<noteq> PhysicalClass \<longrightarrow> capClass (cteCap cte) \<noteq> C"
-  by (auto simp: fresh_virt_cap_class_def)
+(* for the Arch locale we want the fully expanded version covering all cases, but avoiding the
+   capMasterCap_ArchObjectCap rewrite case for an unspecified ArchObjectCap *)
+declare capMasterCap_simps[simp del]
```
corlewis: Thank you for this comment. Are we ok with the two versions having the same name though?

proof/refine/AARCH64/ArchCSpace_I.thy
```diff
@@ -11,699 +11,145 @@ theory ArchCSpace_I
...
-           (capMasterCap cap) (capMasterCap cap')"
+    (cap = cap'
+     \<and> (\<not> isNullCap cap \<and> \<not> isZombie cap
+        \<and> \<not> isUntypedCap cap \<and> \<not> isArchFrameCap cap)
+     \<and> (\<not> isNullCap cap' \<and> \<not> isZombie cap'
+        \<and> \<not> isUntypedCap cap' \<and> \<not> isArchFrameCap cap'))
+    \<or> (capRange cap' \<noteq> {} \<and> capRange cap' \<subseteq> capRange cap
+       \<and> (isUntypedCap cap \<or> (isArchFrameCap cap \<and> isArchFrameCap cap')))
+    \<or> (isIRQControlCap cap \<and> isIRQHandlerCap cap')
+    \<or> (isIRQControlCap cap \<and> isArchObjectCap cap' \<and> isSGISignalCap (capCap cap')))
```
corlewis: Thank you for improving the indentation but I think there's still some unnecessary brackets making this more confusing than it needs to be.

proof/refine/AARCH64/ArchCSpace_I.thy
```diff
@@ -737,416 +183,91 @@ lemma sameObjectAs_def2:
...
+  by (simp add: sameRegionAs_def2, elim conjE disjE)
+     (auto simp: isCap_simps capRange_def) (* long *)
 
+(* FIXME arch-split: TODO interface *)
 lemma capMasterCap_maskCapRights[simp]:
-  "capMasterCap (maskCapRights msk cap)
-    = capMasterCap cap"
-  apply (cases cap;
-         simp add: maskCapRights_def Let_def isCap_simps capMasterCap_def)
+  "capMasterCap (maskCapRights msk cap) = capMasterCap cap"
```
corlewis: Is this a FIXME that's resolvable now or do you want to come back to it later when the use-cases are better defined?

proof/refine/CSpace_I.thy
```diff
@@ -12,4 +12,1777 @@ theory CSpace_I
...
+definition capMasterCap :: "capability \<Rightarrow> capability" where
+ "capMasterCap cap \<equiv> case cap of
+   EndpointCap ref bdg s r g gr \<Rightarrow> EndpointCap ref 0 True True True True
+ | NotificationCap ref bdg s r \<Rightarrow> NotificationCap ref 0 True True
+ | CNodeCap ref bits gd gs \<Rightarrow> CNodeCap ref bits 0 0
+ | ThreadCap ref \<Rightarrow> ThreadCap ref
+ | ReplyCap ref master g \<Rightarrow> ReplyCap ref True True
+ | UntypedCap d ref n f \<Rightarrow> UntypedCap d ref n 0
+ | ArchObjectCap acap \<Rightarrow> ArchObjectCap (arch_capMasterCap acap)
+ | _ \<Rightarrow> cap"
```
corlewis: While we're arch-splitting this, might as well fix the indentation as well.

proof/refine/AARCH64/ArchCSpace_I.thy
Xaphiosis: I probably should've scrapped all these `TODO interface` FIXMEs, but it does raise why I got a bit stuck on them. This is a lemma that could in theory be used outside of Arch, but it doesn't seem to right now. So the obvious thing is to not make it an interface lemma right now... but on the other hand, for new lemmas/theories/etc in the future, it then won't be in scope for people to use. I'm a bit torn on the issue. What do you think?

proof/refine/AARCH64/ArchCSpace_I.thy
Xaphiosis: Good question. Some lemmas have the same name between Arch and generic (especially in AInvs), but in Refine I've tended to side with adding a `gen_` to the generic ones. I don't remember a clear reason why I didn't in this case. Would it make sense to rename the generic one?

proof/refine/AARCH64/ArchCSpace_I.thy
Xaphiosis: Ahh, there are so many indentation/style crimes in these old lemmas. Will do.

proof/refine/CSpace_I.thy
Xaphiosis: Good idea, will do.

proof/refine/CSpace1_R.thy
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * SPDX-License-Identifier: GPL-2.0-only
+ *)
+
+(* CSpace refinement *)
+
+theory CSpace1_R
+imports
+  ArchCSpace_I
+  AInvs.ArchDetSchedSchedule_AI
```
corlewis: I know a couple of the older architectures imported this previously but I think it would be better to go with the ones that didn't and remove this import. If nothing else, that would be one less place for an unexpected context merge to go wrong and cause problems in the future.

proof/refine/AARCH64/ArchCSpace_I.thy
corlewis: Yeah, I've been ignoring most of them for that reason but it's hard to resist the urge to suggest fixing it when it's a lemma that you're touching already.

proof/refine/AARCH64/ArchCSpace_I.thy
corlewis: Hmm, that's a tough one. I think I lean towards adding lemmas like this that you spot to the interface. It's a bit more boilerplate now but could avoid a lot of pain in the future if they do end up being needed.

corlewis: Haha, after making more progress in reviewing this, looks like you already did add it to the interface and remove this FIXME in a later commit.

Xaphiosis: Damn, did I squash those into the wrong commit?

proof/refine/AARCH64/CNodeInv_R.thy
```diff
@@ -4615,14 +4613,15 @@ lemma distinct_zombies_switchE:
   apply (drule_tac F="\<lambda>cap. (isUntypedCap cap, isZombie cap, isArchFrameCap cap,
                              capClass cap, capUntypedPtr cap, capBits cap)"
                in master_eqE,
-         simp add: isCap_Master capClass_Master capUntyped_Master capBits_Master)+
-  apply (simp add: distinct_zombies_def distinct_zombie_caps_def
+         simp add: gen_isCap_Master capClass_Master capUntyped_Master capBits_Master
+                   isArchFrameCap_capMasterCap)+
+  apply (simp add: distinct_zombies_def distinct_zombie_caps_def AARCH64.arch_isCap_Master
```
corlewis, line 4619: `(* FIXME arch-split *)` (for all architectures)

proof/refine/AARCH64/CSpace_R.thy
```diff
@@ -229,7 +229,8 @@ lemma parent_preserved:
    isMDBParentOf cte' (CTE src_cap src_node)"
   using parency unfolding weak_derived'_def
   apply (cases cte')
-  apply (simp add: isMDBParentOf_CTE sameRegionAs_def2)
+thm isArchFrameCap_capMasterCap
```
corlewis: This `thm` is removed in a later commit but probably should get rid of it here to start with.

proof/refine/ARM/CSpace1_R.thy
```diff
@@ -190,6 +190,8 @@ lemma acap_relation_SGISignalCap:
                     irq = ucast irq' \<and> target = ucast target')"
   by (cases acap) auto
 
+typ arch_irq_control_invocation
```
corlewis: Removed in a later commit, should be squashed into this commit.

proof/refine/AARCH64/CSpace_R.thy
corlewis: Appears in all architectures.

proof/refine/ARM/CSpace1_R.thy
corlewis: Also appears in ARM_HYP

proof/refine/StateRelation.thy
```diff
@@ -438,10 +438,17 @@ text \<open>Architecture-specific requirements\<close>
 
 locale StateRelation_R =
   (* ghost state relation must not depend on machine state *)
+  (* FIXME arch-split: consolidate after arch-split pass; these simps have overlap *)
```
corlewis, line 441: Haha, good thing I looked here, I was about to make a comment suggesting this.

proof/refine/Invariants_H.thy
```diff
@@ -505,6 +505,10 @@ definition
   if capClass cap \<noteq> PhysicalClass then {}
   else {capUntypedPtr cap .. capUntypedPtr cap + 2 ^ capBits cap - 1}"
 
+(* FIXME: poor name, isArchObjectCap this takes a predicate *)
+definition isArchCap :: "(arch_capability \<Rightarrow> bool) \<Rightarrow> capability \<Rightarrow> bool" where
+  "isArchCap P cap \<equiv> case cap of ArchObjectCap acap \<Rightarrow> P acap | _ \<Rightarrow> False"
```
corlewis: I think there's a typo in this comment. Are you meaning to say that this takes a predicate unlike `isArchObjectCap`?

corlewis: More usefully, maybe `testArchObjectCap`? I still don't like it that much but it is at least a bit closer to what this is doing.

proof/refine/CSpace1_R.thy
```diff
@@ -14,4 +14,6145 @@ imports
...
+  unfolding word_bits_def
+  by simp
+
+requalify_facts Arch.word_bits_gt_0
+declare word_bits_gt_0[simp]
+
+(* FIXME arch-split: what do we do with this one? a bundle that kicks out word_bl_Rep' too? *)
+(* Safer evaluation of "length (to_bl x)" for machine words; yields word_bits instead of
+   LENGTH(32) or LENGTH(64). Works best when word_bl_Rep' is removed from simpset. *)
+lemmas word_bl_word_bits = word_bl_Rep'[where 'a=machine_word_len, simplified word_bits_def[symmetric]]
```
corlewis: This FIXME seems like a good question but I'm a bit confused about why it's specifically relevant for arch-split.

proof/refine/CSpace1_R.thy
```diff
@@ -14,4 +14,6145 @@ imports
...
+locale CSpace1_R =
+  assumes archMDBAssertions_cross:
+    "\<And>s s'.
+     \<lbrakk> valid_arch_mdb (is_original_cap s) (caps_of_state s); valid_arch_state s; valid_objs s;
+       (s, s') \<in> state_relation \<rbrakk>
+     \<Longrightarrow> archMDBAssertions s'"
+  assumes parentOf_trans:
+    "\<And>s a b c. \<lbrakk> s \<turnstile> a parentOf b; s \<turnstile> b parentOf c \<rbrakk> \<Longrightarrow> s \<turnstile> a parentOf c"
+  assumes pspace_relation_cte_wp_atI':
+    "\<And>(s::det_ext state) s' cte x.
```
corlewis: Very minor but this could be `det_state` instead of `det_ext state`.

proof/refine/CSpace1_R.thy
```diff
@@ -14,4 +14,6145 @@ imports
...
+
+end
+
+
+locale CSpace1_R =
+  assumes archMDBAssertions_cross:
+    "\<And>s s'.
+     \<lbrakk> valid_arch_mdb (is_original_cap s) (caps_of_state s); valid_arch_state s; valid_objs s;
+       (s, s') \<in> state_relation \<rbrakk>
+     \<Longrightarrow> archMDBAssertions s'"
```
corlewis, line 84: Hmm, should the body of this be indented from the `⋀`?

proof/refine/CSpace1_R.thy
```diff
@@ -1,96 +1,195 @@
...
+  assumes ghost_relation_wrapper_same_concrete_set_cap:
+    "\<And>s s' c cap src.
+     \<lbrakk> ghost_relation_wrapper s c; ((), s') \<in> fst (set_cap cap src s) \<rbrakk>
+     \<Longrightarrow> ghost_relation_wrapper s' c"
+  assumes ghost_relation_wrapper_same_abs_set_cap:
+    "\<And>a a' c c' cap dest.
+     \<lbrakk> ghost_relation_wrapper a c; ((), a') \<in> fst (set_cap cap dest a);
+       ksArchState c' = ksArchState c; gsUserPages c' = gsUserPages c; gsCNodes c' = gsCNodes c \<rbrakk>
+     \<Longrightarrow> ghost_relation_wrapper a' c'"
+  assumes  ghost_relation_wrapper_set_cap_twice:
```
corlewis:
```suggestion
  assumes ghost_relation_wrapper_set_cap_twice:
```

proof/refine/CSpace1_R.thy
```diff
@@ -14,4 +14,6145 @@ imports
...
+      apply (rule corres_trivial, simp)
+     apply (wp | simp)+
+  done
+
+lemma maskCapRights[simp]:
+  "cap_relation c c' \<Longrightarrow>
+  cap_relation (mask_cap msk c) (maskCapRights (rights_mask_map msk) c')"
+  by (simp add: mask_cap_def cap_relation_masks)
+
+
```
corlewis: Extra newline

proof/refine/CSpace1_R.thy
```diff
@@ -14,4 +14,6145 @@ imports
...
+   apply (rule resolveAddressBits_real_cte_at')
+  apply (erule real_cte_at')
+  done
+
+declare AllowSend_def[simp]
+declare AllowRecv_def[simp]
+
+(* arch-split note: this proof was made generic due to very carefully managing to not accidentally
+   unfold word_size, word_bits_def and cte_size_bits_def (which are available in generic context).
+   If you see 64 or 32 show up here, something has gone wrong. *)
```
corlewis, line 476: Thanks for the helpful comment!
Does this suggest that those lemmas shouldn't be generic? Or would that be too painful without first refactoring a lot of the other definitions and proofs that use them?

proof/refine/CSpace1_R.thy
```diff
@@ -578,25 +551,59 @@ lemma corres_stateAssert_assume_stronger:
    apply wp+
   done
 
-lemma cap_table_at_gsCNodes:
-  "cap_table_at bits ptr s \<Longrightarrow> (s, s') \<in> state_relation
-    \<Longrightarrow> gsCNodes s' ptr = Some bits"
-  apply (clarsimp simp: state_relation_def ghost_relation_def
-                        obj_at_def is_cap_table)
-  apply blast
+
```
corlewis:
```suggestion
```

proof/refine/CSpace1_R.thy
```diff
@@ -1574,32 +1409,32 @@ lemma cte_at_cte_map_in_obj_bits:
...
+  apply (subgoal_tac "of_bl (snd p) << cte_level_bits < (2 ^ tcb_bits :: machine_word)")
+   apply (drule (1) pspace_alignedD[rotated])
+   apply (simp add: is_aligned_no_overflow' le_plus' word_le_minus_one_leq x_power_minus_1)
    apply (erule is_aligned_no_wrap')
-   apply simp
-  apply (simp add: tcb_cap_cases_def tcb_cnode_index_def to_bl_1 cte_level_bits_def
-              split: if_split_asm)
+   apply (simp add: word_bits_conv shiftl_t2n mult_ac)
+   \<comment> \<open>we have a rule on tcbBlockSizeBits and cteSizeBits, but need it on tcb_bits and cte_level_bits\<close>
+  apply (simp flip: cteSizeBits_cte_level_bits tcbBlockSizeBits_tcb_bits)
```
corlewis: Should this be a FIXME to make that rule?

proof/refine/CSpace1_R.thy
```diff
@@ -2014,10 +1740,11 @@ lemma caps_of_state_rev_cross:
 
 lemma sameRegion_corres:
   "\<lbrakk> sameRegionAs c' d'; cap_relation c c'; cap_relation d d' \<rbrakk>
-  \<Longrightarrow> same_region_as c d"
+   \<Longrightarrow> same_region_as c d"
   by (simp add: same_region_as_relation)
 
 lemma is_final_cap_unique:
+  fixes s :: "det_ext state"
```
corlewis: `det_state`

proof/refine/CSpace1_R.thy
```diff
@@ -2077,25 +1798,8 @@ proof -
...
-  "cap_relation cap cap' \<Longrightarrow>
-   cap_irqs cap = (case capMasterCap cap' of IRQHandlerCap irq \<Rightarrow> {irq} | _ \<Rightarrow> {})"
-  by (simp split: cap_relation_split_asm arch_cap.split_asm)
-
-lemma arch_gen_refs_relation_Master:
-  "cap_relation cap cap' \<Longrightarrow> arch_gen_refs cap = {}"
-  by (clarsimp split: cap_relation_split_asm)
-
 lemma is_final_cap_unique_sym:
+  fixes s :: "det_ext state"
```
corlewis: `det_state`

proof/refine/CSpace1_R.thy
```diff
@@ -2148,356 +1852,51 @@ proof -
...
-lemma no_loops_no_subtree:
-  "no_loops m \<Longrightarrow> m \<turnstile> x \<rightarrow> x = False"
+lemma is_final_untyped_ptrs:
+  "\<lbrakk> ctes_of s' (cte_map slot) = Some cte; ctes_of s' y = Some cte'; cte_map slot \<noteq> y;
+     pspace_relation (kheap s) (ksPSpace s'); valid_objs s; pspace_aligned s; pspace_distinct s;
+     cte_wp_at (\<lambda>cap. is_final_cap' cap s \<and> obj_refs cap \<noteq> {}) slot s \<rbrakk>
+   \<Longrightarrow> capClass (cteCap cte') \<noteq> PhysicalClass
+      \<or> isUntypedCap (cteCap cte')
+      \<or> capUntypedPtr (cteCap cte) \<noteq> capUntypedPtr (cteCap cte')"
+  for s :: "det_ext state" and s' :: kernel_state
```
corlewis: `det_state`

proof/refine/CSpace1_R.thy
```diff
@@ -2678,86 +2038,260 @@ crunch updateMDB
...
+end (* mdb_preserve_common *)
+
+locale CSpace1_R_2 = CSpace1_R + mdb_preserve_common +
+  assumes arch_mdb_preservation_refl[simp, intro!]:
+    "\<And>cap. arch_mdb_preservation cap cap"
+  assumes arch_mdb_preservation_Untyped[simp]:
+    "\<And>d d' p p' sz sz' idx idx'.
+     arch_mdb_preservation (UntypedCap d p sz idx) (UntypedCap d' p' sz' idx')"
+  assumes updateMDB_pspace_relation:
+    "\<And>x (s :: det_ext state) s' s'' p f.
```
corlewis: `det_state`

proof/refine/CSpace1_R.thy
```diff
@@ -4115,2635 +3237,2408 @@ lemma setUntypedCapAsFull_ctes_of_no_0:
...
-  shows  "valid_dlist m = valid_dlist m'"
-  apply (rule iffI)
-   apply (rule valid_dlist_preserve_oneway[OF dom misc])
-     apply simp+
-  apply (rule valid_dlist_preserve_oneway)
-    apply (simp add:dom misc)+
-done
+(* FIXME: move *)
+lemma next_slot_eq:
+  "\<lbrakk>next_slot p t' m' = x; t' = t; m' = m\<rbrakk> \<Longrightarrow> next_slot p t m = x"
```
corlewis: This seems to have lost some details, the comment used to say "consider putting in AINVS or up above cteInsert_corres". Would that be useful to keep in the FIXME?

proof/refine/CSpace1_R.thy
```diff
@@ -4115,2635 +3237,2408 @@ lemma setUntypedCapAsFull_ctes_of_no_0:
...
+   apply simp
+  apply (clarsimp simp: parentOf_def src dest n)
+  done
+
+lemma n_to_dest[simp]:
+  "n \<turnstile> p \<leadsto> dest = (p = src)"
+  by (simp add: n_direct_eq)
+
+end
+
```
corlewis, line 4295:
```suggestion
```

proof/refine/CSpace1_R.thy
```diff
@@ -4115,2635 +3237,2408 @@ lemma setUntypedCapAsFull_ctes_of_no_0:
...
+end
 
-definition
-  "dest2_node \<equiv> MDB (if dest = mdbPrev src_node then dest else mdbNext dest_node)
-                    (if dest = mdbNext src_node then dest else mdbPrev dest_node)
-                    (mdbRevocable dest_node)
-                    (mdbFirstBadged dest_node)"
+locale CSpace1_R_3 = CSpace1_R_2 +
+  (* mdb_insert_der.dest_no_parent_n (n expanded via mdb_insert), needs Arch *)
+  (* locale-local form is "n \<turnstile> dest \<rightarrow> p = False" *)
```
corlewis: The comments in this locale could use some more detail, or at least an initial explanation. I assume they make sense now with this arch-split fresh in your mind but the intent here is not clear to me reading the diff and I think they will be confusing in the future when we've forgotten the current context.

proof/refine/RISCV64/CSpace_R.thy
```diff
@@ -9,9 +9,54 @@
 *)
 
 theory CSpace_R
-imports CSpace1_R
+imports ArchCSpace1_R
 begin
 
+
```
corlewis:
```suggestion

```

corlewis (commented):
Great work, it's amazing to see so much made generic!

The locales look very painful and confusing, so thank you for the thorough commit message and comments. Hopefully future locales will be easier to work with now that you've gotten through the CSpace1_R ones and have this approach.

proof/crefine/ARM/CSpace_C.thy
```diff
@@ -527,17 +527,23 @@ lemma Arch_isCapRevocable_spec:
            simp: cap_get_tag_isCap cap_get_tag_isCap_unfolded_H_cap isCap_simps
                  Arch_isCapRevocable_def2 from_bool_0)
 
-method revokable'_hammer = solves \<open>(simp add: cap_get_tag_isCap isCap_simps ccorres_cond_iffs,
-                    rule ccorres_guard_imp, rule ccorres_return_C; clarsimp)\<close>
+lemmas isCapRevocable_simps[simp] = Retype_H.isCapRevocable_def[split_simps capability.split]
+
+context begin (* revokable_ccorres *)
+
+private method revokable'_hammer = solves \<open>(
```
lsf37: Should this now be called `isCapRevocable_hammer`?

proof/crefine/ARM_HYP/CSpace_C.thy
```diff
@@ -534,17 +534,23 @@ lemma Arch_isCapRevocable_spec:
            simp: cap_get_tag_isCap cap_get_tag_isCap_unfolded_H_cap isCap_simps
                  Arch_isCapRevocable_def2 from_bool_0)
 
-method revokable'_hammer = solves \<open>(simp add: cap_get_tag_isCap isCap_simps ccorres_cond_iffs,
-                    rule ccorres_guard_imp, rule ccorres_return_C; clarsimp)\<close>
+lemmas isCapRevocable_simps[simp] = Retype_H.isCapRevocable_def[split_simps capability.split]
+
+context begin (* revokable_ccorres *)
+
+private method revokable'_hammer = solves \<open>(
```
lsf37: Marking here in case we do want to rename into `isCapRevocable_hammer`

proof/refine/Invariants_H.thy
lsf37: It lifts a predicate on arch caps to a predicate on caps. Maybe something with `_lifted` or `_lift`?

proof/refine/CSpace_I.thy
```diff
@@ -12,4 +12,1777 @@ theory CSpace_I
...
+  "\<lbrace>P\<rbrace> getSlotCap addr \<lbrace>\<lambda>rv. P\<rbrace>"
+  by (simp add: getSlotCap_def, wp)
+
+lemma cap_case_CNodeCap:
+  "(case cap of CNodeCap _ _ _ _ \<Rightarrow> P
+           | _ \<Rightarrow> Q)
+   = (if isCNodeCap cap then P else Q)"
+  by (cases cap, simp_all add: gen_isCap_simps)
+
+declare resolveAddressBits.simps[simp del] (* FIXME arch-split: do we want to keep this? *)
```
lsf37: Yes, we pretty much never want resolveAddressBits to unfold automatically, because that is not safe and will easily not terminate. We can remove the FIXME.

proof/refine/CSpace1_R.thy
lsf37: +1 from me. It's possible that removing this leads to a bunch of breakage -- in that case I'd say it is fine to do separately and make an issue instead, but if it can be removed without much impact, it would be nice to do now.

proof/refine/CSpace1_R.thy
lsf37: Why it is relevant for arch-split: in the generic files it'd be nice to avoid getting 32 or 64 (and get word_bit instead). In arch files, it matters less.

I think making a bundle would be nice -- then when the word stuff changes in the background we might only have to change one place instead of many.

proof/refine/CSpace1_R.thy
lsf37: They should definitely be generic, they express how `cte_map` works, which should be the same on all architectures. If we change anything in the setup, I think it should be that less of 32/64 should show up automatically. We've accepted that so far from `machine_word` as a thing we hav to live with, but Raf has figured out a way to not get 32/64 and word_bits instead.

If we manage to make that more useable and robust (i.e. requiring less care), we could get those benefits for more proofs.

proof/refine/CSpace1_R.thy
lsf37: Would it make sense to repeat a `defines` for `n = modify_map ..` in the definition of this locale? It's only used in one `assumes`, so maybe not, but it might make it more obvious that we are repeating an assumption of another locale.

proof/refine/CSpace1_R.thy
corlewis: To clarify, I was talking about the lemmas mentioned in the comment (`word_size, word_bits_def and cte_size_bits_def`), not the `cte_map` lemmas.

lsf37: Oh, right. I don't think so: while it is much nicer to do things the way Raf did them, especially in larger proofs, it is still legit to have generic proofs by automation, especially when those disappear in terminal proof methods, i.e. you never actually see the values in the goal or need to refer to them.

proof/refine/CSpace1_R.thy
Xaphiosis: - [x] investigate repeating assumption

proof/refine/CSpace1_R.thy
Xaphiosis: - [x] try make a rule that makes sense

proof/refine/CSpace1_R.thy
Xaphiosis: My personal view is that these kinds of things should be wrapped up inside Arch, but currently doing that would cause changes all over and I wouldn't be able to keep the changes focused on Refine.
WordSetup could potentially already be putting stuff in Arch, at least at a level where we say `Arch.word_bits_def`. Then `cte_size_bits_def` and company could also be hidden. It's a bigger project, but would prevent annoyances like the way this proof is done, while still allowing doing automation for terminal proof methods by using `Arch.`. Do you think it would be worth it to try that one day @lsf37 ?

proof/refine/CSpace1_R.thy
Xaphiosis: I've copied the pattern from AInvs, since I took it to indicate "we need to have these quantifications, but they're not otherwise important". They're as left-aligned as the thing they belong to, and we've been using it when we need the space. There are many of these beyond this PR. Thoughts?

proof/refine/CSpace1_R.thy
Xaphiosis: Should I convert the FIXME to an issue and remove the FIXME?
I want something to remind me/us to try out the find_theorems augmentation in order to be able to search in theorem sets (particularly global simpset), since we need to remove simps that produce `LENGTH('a)`.

proof/refine/Invariants_H.thy
Xaphiosis: Correct, this comment indeed has a brain-eaten missing "unlike". I still have no idea what to call it, might fix the comment and leave it for this PR. Otherwise I'll likely end up with some kind of `lift_ArchCap_pred` monstrosity.

proof/refine/ARM/CSpace1_R.thy
Xaphiosis: - [x] commit surgery

proof/refine/AARCH64/CSpace_R.thy
Xaphiosis: - [x] commit surgery

proof/refine/CSpace1_R.thy
Xaphiosis: Didn't see any breakage, will be removed once I push.

proof/refine/AARCH64/ArchCSpace_I.thy
Xaphiosis: Hmm, I guess I'll leave this one then.

proof/refine/CSpace1_R.thy
corlewis: Yeah, it's interesting. My initial reaction was that it didn't look right with this indentation but then I was reminded that it's a common pattern when I kept on seeing it. I'm fine with leaving it as is, particularly with your convincing argument about the quantifications not being important.

proof/refine/AARCH64/ArchCSpace_I.thy
Xaphiosis: - [x] reduce parens

proof/refine/AARCH64/ArchCSpace_I.thy
Xaphiosis: - [x] investigate adding `gen_` prefix to generic version

proof/refine/CSpace1_R.thy
Xaphiosis: I've done a replacement of all `det_text state` in `CSpace1_R`, thanks for catching these.

proof/refine/AARCH64/ArchCSpace_I.thy
Xaphiosis: Added gen_ prefix to generic version

proof/refine/AARCH64/ArchCSpace_I.thy
Xaphiosis: Updated to reduce parens and re-intent on all arches.

proof/refine/CSpace1_R.thy
Xaphiosis: Done

proof/refine/AARCH64/CNodeInv_R.thy
Xaphiosis: Done, and added another one I missed later the same file(s) (all arches)

proof/refine/CSpace1_R.thy
Xaphiosis: I added `lemmas tcb_slots_less_2p_tcb_bits[simp]` in Bits_R for the ASpec-inspired constants.

proof/refine/CSpace1_R.thy
Xaphiosis: Added amalgam of original comment and FIXME move

proof/refine/CSpace1_R.thy
Xaphiosis: Repeating the defines did not end up making sense, since the `n` captures too many fixed vars from `mdb_insert`. Updated the comment with @corlewis's help instead.

proof/refine/CSpace1_R.thy
lsf37: I would vote for an issue, yes. There are too many FIXMEs that are more comments :-)

Xaphiosis:
I've done pass 2, which should address the remainder of the issues, including the in-place git surgery to remove the `thm` and `typ`. Left the other fixes as squash commits on top. Will look at rebase while people confirm I've addressed everything.

proof/refine/CSpace1_R.thy
lsf37: I'm not sure -- it would be nice, but unclear that it should be high priority.

lsf37 (approved):
:+1: all issues from my side are resolved (still needs rebase, though)

proof/refine/CSpace_I.thy
```diff
@@ -569,22 +569,22 @@ context CSpace_I begin
 
 definition capMasterCap :: "capability \<Rightarrow> capability" where
  "capMasterCap cap \<equiv> case cap of
```
corlewis: Sorry, one last change here, the whole block needs to be indented one more.

proof/refine/CSpace1_R.thy
Xaphiosis: Definitely a post-arch-split mop-up operation kind of possibility. Once everything works, it's easier to tweak.

proof/refine/CSpace_I.thy
Xaphiosis: Reindented, along with two sinners above it.

PR #927, opened 2025-09-29
refine: consolidate update_tcb_cte_cases lemmas

corlewis:
Extracted from MCS, where they were consolidated as part of https://github.com/seL4/l4v/pull/924.


Xaphiosis (approved):
For the commit message, I think a better word than "refactor" would be "consolidate". Otherwise, thumbs up.

PR #930, opened 2025-10-30

corlewis:
* aspec+proof: refactor det_ext to remove scheduler state
* all: update to Isabelle2025

This was a more painful merge than usual, due to both sides having made similar but slightly different changes for the det_ext refactoring. I think I caught most of the cases and the proofs pass, but it's possible that some duplicate lemmas slipped through the merge process.

### DIFF
--- a/tools/c-parser/StrictC.lex
+++ b/tools/c-parser/StrictC.lex
@@ -135,6 +135,7 @@ integer=([1-9]{digit}*|0){intsuffix}?;
 octint=0{digit}+{intsuffix}?;
 hexint=("0x"|"0X")({digit}|[a-fA-F])+{intsuffix}?;
 identifier=({alpha}|"_")({alpha}|{digit}|"_")*;
+isa_identifier=({alpha}|"_")({alpha}|{digit}|"_"|"'")*;
 ws = [\ \t];
 commentbody = ([^*]*(\*[^/])?)*;
 attr_start = "__attribute__"{ws}*"((";
@@ -543,7 +544,7 @@ attr_start = "__attribute__"{ws}*"((";
 <SPECIAL_COMMENT>"[" => (tok(Tokens.LBRACKET,source,yypos,yypos));
 <SPECIAL_COMMENT>"]" => (tok(Tokens.RBRACKET,source,yypos,yypos));
 <SPECIAL_COMMENT>"*" => (tok(Tokens.YSTAR,source,yypos,yypos));
-<SPECIAL_COMMENT>{identifier} =>
+<SPECIAL_COMMENT>{isa_identifier} =>
   (mk_tokident(yytext,getPos(source,yypos),
                getPos(source,yypos + size yytext - 1)));
 <SPECIAL_COMMENT>"*/" => (YYBEGIN INITIAL;


proof/refine/RISCV64/ArchInvsLemmas_H.thy
```diff
@@ -576,24 +576,38 @@ lemmas gen_objBitsT_simps = objBitsT_simps(1-7)
...
          split: arch_kernel_object.split)+
 
+lemma objBitsKO_no_overflow[simp, intro!]:
+  "objBitsKO ko < word_bits \<Longrightarrow> (1::machine_word) < (2::machine_word)^(objBitsKO ko)"
+  by (cases ko; simp add: objBits_simps' pageBits_def pteBits_def scBits_pos_power2
+                   split: arch_kernel_object.splits)
+
 end
 
+(* FIXME: arch split RT *)
```
Xaphiosis: nitpick: `arch-split`

proof/invariant-abstract/TcbAcc_AI.thy
```diff
@@ -2615,7 +2599,7 @@ lemma thread_set_priority_valid_arch_caps[wp]:
   unfolding thread_set_priority_def
   by (wpsimp wp: thread_set_arch_caps_trivial simp: tcb_cap_cases_def) auto
 
-global_interpretation thread_set_priority: tcb_op "thread_set_priority t p"
+sublocale thread_set_priority: tcb_op "thread_set_priority t p"
```
Xaphiosis, line 2602: I'm a bit lost on why the change from global_interpretation to sublocale

proof/invariant-abstract/DetSchedAux_AI.thy
```diff
@@ -562,12 +562,6 @@ lemma retype_region_valid_blocked[wp]:
   apply (blast intro: valid_blocked_fold_update)
   done
 
-=======
-crunch create_cap, cap_insert
-  for valid_queues[wp]: valid_queues
-  (wp: valid_queues_lift)
->>>>>>> 3667ef81cd60b2b814cd28e59231bc7399250fc6
```
Xaphiosis: do we want merge markers in commits? this might be worth squashing back into the merge, if that's even possible

proof/crefine/RISCV64/Ipc_C.thy
```diff
@@ -5643,7 +5643,7 @@ lemma reply_push_ccorres:
                            simp: typ_heap_simps' creply_relation_def option_to_ctcb_ptr_def
                           split: if_splits)
        apply ceqv
-      apply (ctac add: setThreadStateBlockedOnReply_ccorres)
+      apply (ctac add: setThreadStateBlockedOnReply_ccorres, rename_tac xfdc')
```
Xaphiosis, line 5646: I don't seem to remember ever doing a `rename_tac xfdc'`, can you say what's happening here?

Xaphiosis (approved):
I looked over the fixup commit and bounced around the merge commit for a bit, mainly looking at RISCV64. Looks good to me. There's a leftover FIXME move, but not sure if it's worth perturbing this PR more.

proof/invariant-abstract/DetSchedAux_AI.thy
lsf37: We'd usually want to avoid that, but I'm guess this was a mistake. Might be worth trying out on a separate copy if it could be squashed into the merge commit, but if it's too hairy, I'm fine with having it fixed up in the next commit (as here)

proof/crefine/RISCV64/Ipc_C.thy
lsf37: I remember these from the Isabelle update. The name spaces for constants and variables are now separate. That means Isabelle will happily generate the variable name `xfdc` in the goal, which is then inaccessible, because when you type `xfdc` you will get the constant `xfdc` instead. Hence the rename to `xfdc'` which does remain accessible.

Edit:  it's the other way around, I think the bound variable shadows the constant, but you want to be able to use the constant `xfdc` in rule instantiations. In any case, they clash.

proof/invariant-abstract/TcbAcc_AI.thy
corlewis: I didn't investigate why but MCS has this section inside of a `TcbAcc_AI_valid_ipc_buffer_cap_0` context block, so `global_interpretation` doesn't work.

proof/invariant-abstract/DetSchedAux_AI.thy
corlewis: Yeah, this was a mistake. It should be possible to squash it into the merge commit but it's always fiddly to do so, which is why I left in this and the other obvious mistakes I made while resolving conflicts and fixed them in the next commit. I'll give it a go but will leave it if I run into problems.

proof/crefine/RISCV64/Ipc_C.thy
corlewis: Yes, this came from the Isabelle update and the edit is exactly right with what is happening. There were a couple of examples in the Isabelle update commit on master doing this `rename xfdc'` which is where I got the idea from. I didn't investigate why it's needed so much more in the MCS proofs though. I guess we explicitly refer to `xfdc` more in them.

proof/invariant-abstract/TcbAcc_AI.thy
Xaphiosis: Right, and I'm guessing we do want to export everything from the inner locale into the outer one, so `interpretation` isn't going to cut it. If only I knew what `TcbAcc_AI_valid_ipc_buffer_cap_0` et al. do

proof/invariant-abstract/DetSchedAux_AI.thy
corlewis: It wasn't too hard to fix this up, so I squashed some of the other more obvious merge mistakes I made as well while I was there.

PR #932, opened 2025-11-13
arch-split: TcbAcc_R

Xaphiosis:
🦆🦆🦆 This PR includes more experiments to reduce review time, please read

In real life, I did the usual addition of TcbAcc_R, hierarchy update, arch-split and then update. However, since the arch-split part first splits AARCH64, then takes the AARCH64 version of ArchCSpace_R and only fixes what broke, we ended up with a small diff there (again).

- `[squash] PR: copy AARCH64 version into TcbAcc_R` - same as last time: stop TcbAcc_R looking like a wall of green (no need to review)
- `aarch64 refine: arch-split TcbAcc_R` - this is the main bulk of the arch-split, dealing with AARCH64
- `[squash] copy AARCH64 ArchTcbAcc_R to other arches` - wipes the *body* (not the header!) of other arches' ArchTcbAcc_R with a copy of the AARCH64 version with `AARCH64` appropriately substituted (as @corlewis requested last time) - (mostly pointless to review)
- `[squash] refine: arch-split TcbAcc_R (other arches) - result of fixing up the ArchTcbAcc_R of other architectures - shows diff to AARCH64, small for this PR
- `refine+crefine: update for ... arch-split` - not many updates
- EXTRA: `refine: make valid_cap_simps' generic` - check commit message for details

Reminder: DO NOT MERGE squashed version if it still says `aarch64 refine: arch-split` (missed this last time)

Stats: 53 files changed, 6496 insertions(+), 24643 deletions(-)
Total: 18147 lines removed

### DIFF
--- a/proof/refine/AARCH64/ArchInvsLemmas_H.thy
+++ b/proof/refine/AARCH64/ArchInvsLemmas_H.thy
@@ -382,9 +382,6 @@ lemma hyp_refs_of_live':
   "hyp_refs_of' ko \<noteq> {} \<Longrightarrow> live' ko"
   by (cases ko, simp_all add: live'_def hyp_refs_of_hyp_live')
 
-lemmas valid_cap_simps' =
-  valid_cap'_def[split_simps capability.split arch_capability.split]
-
 lemma is_physical_cases:
  "(capClass cap = PhysicalClass) =
   (case cap of NullCap                          \<Rightarrow> False
--- a/proof/refine/ARM/ArchInvsLemmas_H.thy
+++ b/proof/refine/ARM/ArchInvsLemmas_H.thy
@@ -317,9 +317,6 @@ lemma hyp_refs_of_live':
   "hyp_refs_of' ko \<noteq> {} \<Longrightarrow> live' ko"
   by (cases ko, simp_all add: live'_def hyp_refs_of_hyp_live')
 
-lemmas valid_cap_simps' =
-  valid_cap'_def[split_simps capability.split arch_capability.split]
-
 lemma is_physical_cases:
  "(capClass cap = PhysicalClass) =
   (case cap of NullCap                          \<Rightarrow> False
--- a/proof/refine/ARM_HYP/ArchInvsLemmas_H.thy
+++ b/proof/refine/ARM_HYP/ArchInvsLemmas_H.thy
@@ -357,9 +357,6 @@ lemma hyp_refs_of_live':
   "hyp_refs_of' ko \<noteq> {} \<Longrightarrow> live' ko"
   by (cases ko, simp_all add: live'_def hyp_refs_of_hyp_live')
 
-lemmas valid_cap_simps' =
-  valid_cap'_def[split_simps capability.split arch_capability.split]
-
 lemma is_physical_cases:
  "(capClass cap = PhysicalClass) =
   (case cap of NullCap                          \<Rightarrow> False
--- a/proof/refine/CSpace1_R.thy
+++ b/proof/refine/CSpace1_R.thy
@@ -75,6 +75,8 @@ lemmas final_matters_simps[simp]
 
 end
 
+lemmas valid_cap_simps' =
+  valid_cap'_def[split_simps capability.split]
 
 locale CSpace1_R =
   assumes archMDBAssertions_cross:
--- a/proof/refine/CSpace_R.thy
+++ b/proof/refine/CSpace_R.thy
@@ -3175,14 +3175,10 @@ lemma usableUntypedRange_uniq:
   apply blast
   done
 
-(* FIXME arch-split: move? doesn't seem like the ARCH version actually unfolds arch on AARCH64, maybe others, check *)
-lemmas gen_valid_cap_simps' =
-  valid_cap'_def[split_simps capability.split]
-
 lemma usableUntypedRange_empty:
   "valid_cap' cp s \<Longrightarrow> isUntypedCap cp
    \<Longrightarrow> (usableUntypedRange cp = {}) = (capFreeIndex cp = maxFreeIndex (capBlockSize cp))"
-  apply (clarsimp simp: gen_isCap_simps max_free_index_def gen_valid_cap_simps' capAligned_def)
+  apply (clarsimp simp: gen_isCap_simps max_free_index_def valid_cap_simps' capAligned_def)
   apply (rule order_trans, rule word_plus_mono_right)
     apply (rule_tac x="2 ^ capBlockSize cp - 1" in word_of_nat_le)
     apply simp
@@ -3204,7 +3200,7 @@ lemma untypedZeroRange_to_usableCapRange:
         \<and> x \<le> y"
   apply (clarsimp simp: untypedZeroRange_def split: if_split_asm)
   apply (frule(1) usableUntypedRange_empty)
-  apply (clarsimp simp: gen_isCap_simps gen_valid_cap_simps' max_free_index_def)
+  apply (clarsimp simp: gen_isCap_simps valid_cap_simps' max_free_index_def)
   apply (simp add: getFreeRef_def mask_def add_diff_eq)
   done
 
@@ -3285,7 +3281,7 @@ crunch cteInsert
 
 lemma valid_NullCap:
   "valid_cap' NullCap = \<top>"
-  by (rule ext, simp add: gen_valid_cap_simps' capAligned_def word_bits_def)
+  by (rule ext, simp add: valid_cap_simps' capAligned_def word_bits_def)
 
 context CSpace_R begin
 
@@ -5081,7 +5077,7 @@ lemma updateFreeIndex_forward_valid_objs':
   apply clarsimp
   apply (frule(1) CSpace1_R.ctes_of_valid)
   apply (clarsimp simp: cte_wp_at_ctes_of gen_isCap_simps capAligned_def
-                        gen_valid_cap_simps' is_aligned_weaken[OF is_aligned_triv])
+                        valid_cap_simps' is_aligned_weaken[OF is_aligned_triv])
   apply (clarsimp simp add: valid_untyped'_def
                   simp del: usableUntypedRange.simps)
   apply (erule allE, erule notE, erule ko_wp_at'_weakenE)
@@ -5108,7 +5104,7 @@ lemma updateFreeIndex_forward_valid_mdb':
   apply (frule(1) CSpace1_R.ctes_of_valid)
   apply (clarsimp simp: cte_wp_at_ctes_of del: subsetI)
   apply (rule usableUntypedRange_mono2,
-         auto simp add: gen_isCap_simps gen_valid_cap_simps' capAligned_def)
+         auto simp add: gen_isCap_simps valid_cap_simps' capAligned_def)
   done
 
 lemma no_fail_getSlotCap:
--- a/proof/refine/RISCV64/ArchInvsLemmas_H.thy
+++ b/proof/refine/RISCV64/ArchInvsLemmas_H.thy
@@ -355,9 +355,6 @@ lemma hyp_refs_of_live':
   "hyp_refs_of' ko \<noteq> {} \<Longrightarrow> live' ko"
   by (cases ko, simp_all add: live'_def hyp_refs_of_hyp_live')
 
-lemmas valid_cap_simps' =
-  valid_cap'_def[split_simps capability.split arch_capability.split]
-
 lemma is_physical_cases:
  "(capClass cap = PhysicalClass) =
   (case cap of NullCap                         \<Rightarrow> False
--- a/proof/refine/X64/ArchInvsLemmas_H.thy
+++ b/proof/refine/X64/ArchInvsLemmas_H.thy
@@ -398,9 +398,6 @@ lemma hyp_refs_of_live':
   "hyp_refs_of' ko \<noteq> {} \<Longrightarrow> live' ko"
   by (cases ko, simp_all add: live'_def hyp_refs_of_hyp_live')
 
-lemmas valid_cap_simps' =
-  valid_cap'_def[split_simps capability.split arch_capability.split]
-
 lemma is_physical_cases:
  "(capClass cap = PhysicalClass) =
   (case cap of NullCap                         \<Rightarrow> False


proof/refine/TcbAcc_R.thy
```diff
@@ -43,17 +39,86 @@ definition
...
-  unfolding getHighestPrio_def gets_def
+  unfolding getHighestPrio_def
   by (fastforce simp: gets_def get_bind_apply lookupBitmapPriority_def bitmap_fun_defs)
 
+locale TcbAcc_R =
+  (* FIXME arch-split: these versions use word_size_bits and can be made generic *)
+  assumes no_fail_loadWord_bits[wp]:
+    "\<And>p. no_fail (\<lambda>_. is_aligned p word_size_bits) (loadWord p)"
+  assumes no_fail_storeWord_bits:
+    "\<And>p w. no_fail (\<lambda>_. is_aligned p word_size_bits) (storeWord p w)"
```
lsf37: Is that a TODO for later or did you mean to include it in this PR?

proof/refine/TcbAcc_R.thy
```diff
@@ -196,19 +219,17 @@ lemma valid_objs'_valid_tcbs'[elim!]:
 
 lemma invs'_valid_tcbs'[elim!]:
   "invs' s \<Longrightarrow> valid_tcbs' s"
-  by (fastforce intro: valid_objs'_valid_tcbs')
+  by (fastforce del: valid_objs'_valid_tcbs' intro: valid_objs'_valid_tcbs')
```
corlewis, line 222: Is this `del` for avoiding a warning?

proof/refine/TcbAcc_R.thy
```diff
@@ -218,11 +239,11 @@ lemma valid_tcbs'_obj_at':
...
   using assms
-  apply (clarsimp simp add: valid_tcbs'_def ran_def typ_at'_def
-                            ko_wp_at'_def valid_obj'_def valid_tcb'_def obj_at'_def)
-  done
+  by (clarsimp simp add: valid_tcbs'_def ran_def typ_at'_def
+                         ko_wp_at'_def valid_obj'_def valid_tcb'_def obj_at'_def)
 
-lemma update_valid_tcb'[simp]:
+(* same derivation on all architectures *)
+lemma (in Arch) update_valid_tcb'[simp]:
```
corlewis, line 246: Hmm, it almost definitely doesn't make a practical difference but if we're requalifying and making a rule `simp` in the general context, would it make sense to not have it `simp` in `Arch`? I could see it being unnecessarily in `simp` in both having an infinitesimal performance impact.

proof/refine/TcbAcc_R.thy
```diff
@@ -2343,6 +2055,85 @@ lemma in_ready_q_tcbQueued_eq:
                   split: option.splits)
   done
 
+
+locale TcbAcc_R_2 = TcbAcc_R +
+  assumes no_fail_isRunnable[wp]:
+    "\<And>t. no_fail (tcb_at' t) (isRunnable t)"
```
corlewis: Out of curiosity, do you remember why this had to go into the interface? This one in particular feels at surface level like it could be generic.

proof/refine/ARM_HYP/Finalise_R.thy
```diff
@@ -3374,6 +3374,12 @@ lemma archThreadSet_tcbSchedPrevNext[wp]:
   apply auto
   done
 
+(* FIXME arch-split: RAF move to ArchTcbAcc_R on ARM_HYP and AARCH64 *)
+lemma archThreadGet_wp:
+  "\<lbrace>\<lambda>s. \<forall>tcb. ko_at' tcb t s \<longrightarrow> Q (f (tcbArch tcb)) s\<rbrace> archThreadGet f t \<lbrace>Q\<rbrace>"
```
corlewis: This sounds like one you want to do now?

proof/refine/TcbAcc_R.thy
Xaphiosis: I'm not really sure what to do with this comment. I guess I could remove the FIXME from it, but the long-term idea would be to not have `is_aligned p 3` style lemmas running around. To do that would need updating from Machine downwards though, and that's a lot of noise.
For now I'll keep the comment, drop the FIXME.

proof/refine/TcbAcc_R.thy
Xaphiosis: Yeah, I'm trying to remove as many warnings as I can from Refine

proof/refine/TcbAcc_R.thy
Xaphiosis: Not sure about performance, but if I'm immediately requalifying and applying `[simp]` I agree it's redundant and will remove it. It's less clear when we're properly in Arch, because there it might need to be `[simp]` to solve other Arch lemmas before we do the interpretation.

proof/refine/ARM_HYP/Finalise_R.thy
Xaphiosis: It turns out I put them into ArchTcbAcc_R but forgot to remove them from the respective Finalise_R and IpcCancel_R ... thank you so much for catching this!

proof/refine/TcbAcc_R.thy
Xaphiosis: I'll take another look. I think this is related to the broken `no_fail_return` you helped me debug (and remove). Now that it's removed, the crunch should work and be generic, I'll try clear that up.

Xaphiosis: Yes, you are right! Thank you for catching that

proof/refine/TcbAcc_R.thy
```diff
@@ -43,17 +39,86 @@ definition
...
+  assumes prioToL1Index_complement_nth_w2p:
+    "\<And>(p::priority) (p'::priority).
+     (~~ 2 ^ prioToL1Index p :: machine_word) !! prioToL1Index p'
+     = (prioToL1Index p \<noteq> prioToL1Index p')"
+  assumes prioToL1Index_size[simp]:
+    "\<And>w. prioToL1Index w < l2BitmapSize"
+  assumes ghost_relation_wrapper_ksSchedulerAction_upd[simp]:
+    "\<And>s s' v. ghost_relation_wrapper s (s'\<lparr>ksSchedulerAction := v\<rparr>) = ghost_relation_wrapper s s'"
+  assumes ghost_relation_wrapper_scheduler_action_upd[simp]:
+    "\<And>s s' v. ghost_relation_wrapper (s\<lparr>scheduler_action := v\<rparr>) s' = ghost_relation_wrapper s s'"
```
corlewis, line 113: I'm having to deal with lemmas like these as well, is there anything nicer we can do or is the expectation that we'll need to add lemmas for each of the state fields that come up? If we do need to add lots of them, they probably should be collected in the `StateRelation_R` interface.

corlewis: Hmm, how about something like this for the interface? We would still need additional lemmas for arch state fields but I think these would handle all of the generic ones.
```
lemma ghost_relation_wrapper_upd_id[simp]:
  "⟦⋀s. kheap (f s) = kheap s⟧
   ⟹ ghost_relation_wrapper (f s) s' = ghost_relation_wrapper s s'"
  by simp

lemma ghost_relation_wrapper_upd_id'[simp]:
  "⟦⋀s'. gsUserPages (f' s') = gsUserPages s'; ⋀s'. gsCNodes (f' s') = gsCNodes s';
    ⋀s'. ksArchState (f' s') = ksArchState s'⟧
   ⟹ ghost_relation_wrapper s (f' s') = ghost_relation_wrapper s s'"
  by simp
  ```

File: proof/refine/TcbAcc_R.thy (hunk 3 above)
Comment (corlewis) [reply proof/refine/TcbAcc_R.thy:246]: Yeah, agreed that within a proper Arch context it would make sense to keep the `[simp]`.

File: proof/refine/TcbAcc_R.thy (hunk 6 above)
Comment (Xaphiosis) [reply proof/refine/TcbAcc_R.thy:113]: Hmm. Good point, I'll have a think. At some point I thought of some conditional lemma, like "if the function doesn't change any of these then it's the same", but then we might not want that in the simpset, as it's conditional and matches many possibilities.
One possibility to not write them all out is some kind of ML hack to generate them, but then I'm less sure given that the whole reason for the wrapper is that the ghost state relation changes between arches.

Comment (Xaphiosis) [reply proof/refine/TcbAcc_R.thy:113]: Somehow our messages overlapped. For the first one, I think that might be ok, though Gerwin is the best arbiter of when it's safe to place a conditional rule into the simpset.

For the second one, check the arch variations carefully! AARCH64 has more ghost state than other arches due to last-level page table wonders, and that extra field doesn't exist on other arches.

Comment (corlewis) [reply proof/refine/TcbAcc_R.thy:113]: Yeah, that's why I included `ksArchState` in the second one. I think it's as general as possible, but I'm working in MCS at the moment so I'm only looking at RISCV64 and will need to confirm that this approach makes sense for other architectures when I backport it to master.

Comment (Xaphiosis) [reply proof/refine/TcbAcc_R.thy:113]: I'd prefer not introducing it on MCS like that if it's not going to work on all arches / master...any chance you'll be able to check that before it gets merged?

Comment (corlewis) [reply proof/refine/TcbAcc_R.thy:113]: Yes, that's the plan.

Comment (lsf37) [reply proof/refine/TcbAcc_R.thy:113]: For the second one, if all of `ksArchState` is the same, the lemmas as such will be true for AArch64. The question is whether you'll be able to prove that all of `ksArchState` remains the same if the ghost state has parts in it it that are being modified. That depends on which function you need it for.

@corlewis which functions are you lifting over with those lemmas?

We could also try to see if there is a way to write it positively, such that this becomes one lemma in the interface that talks about a set of update functions. I haven't tried this out, but something along the lines of

```
"⋀s s' v. f : { ksSchedulerAction_update, ksSomeOtherState_update} ==>  ghost_relation_wrapper s (f s') = ghost_relation_wrapper s s'"
```

That would not be nicely usable directly, but the forms that are needed could be derived from it generically (i.e. the original forms that we now have in the interface). On the arch side, the interface lemma can probably proved directly with `auto` (assuming a suitable unfolding of ghost_relation_wrapper), because the set should turn into disjunctions that `auto` will split on.

Comment (lsf37) [reply proof/refine/TcbAcc_R.thy:113]: > Hmm. Good point, I'll have a think. At some point I thought of some conditional lemma, like "if the function doesn't change any of these then it's the same", but then we might not want that in the simpset, as it's conditional and matches many possibilities.

I think that is correct, you would not want these as simp lemmas. But you could use them to derive the original forms once in the generic theory from them, and then make them simp lemmas. That still means one lemma for each field update, but only one generically, not one in each Arch.

Comment (Xaphiosis) [reply proof/refine/TcbAcc_R.thy:113]: I like the `f : { ksSchedulerAction_update, ksSomeOtherState_update}` idea, it has a chance of being the most compact interface lemma, and then we can split it up for simp consumption in the generic context.

Comment (corlewis) [reply proof/refine/TcbAcc_R.thy:113]: Are my proposed lemmas not good as simp rules because they could be unified as `ghost_relation_wrapper s = ghost_relation_wrapper s` and diverge? I thought the simplifier was smart enough not to do that?

I can add that I've tried applying them to both MCS RISCV64 and master AARCH64 now and that they work perfectly with the existing proofs. Currently, all of the functions that we want to simplify are state field updates (i.e. `ghost_relation_wrapper (s⦇scheduler_action := v⦈) s'`), which are very easy for the simplifier to handle. I can imagine a more complicated function would be harder to deal with but I'm pretty sure that that would be true for any approach we take.

Comment (corlewis) [reply proof/refine/TcbAcc_R.thy:113]: To get all of the context in one place, on AARCH64 the ghost relation wrapper looks like this
```
definition ghost_relation_wrapper :: "det_state ⇒ kernel_state ⇒ bool" where
  "ghost_relation_wrapper s s' ≡
     ghost_relation (kheap s) (gsUserPages s') (gsCNodes s') (gsPTTypes (ksArchState s'))"
```
and my lemmas are
```
lemma ghost_relation_wrapper_upd_id[StateRelation_R_assms, simp]:
  "⟦⋀s. kheap (f s) = kheap s⟧
   ⟹ ghost_relation_wrapper (f s) = ghost_relation_wrapper s"
  by auto

lemma ghost_relation_wrapper_upd_id'[StateRelation_R_assms, simp]:
  "⟦⋀s'. gsUserPages (f' s') = gsUserPages s'; ⋀s'. gsCNodes (f' s') = gsCNodes s';
    ⋀s'. ksArchState (f' s') = ksArchState s'⟧
   ⟹ ghost_relation_wrapper s (f' s') = ghost_relation_wrapper s s'"
  by simp
  ```
Within the `Arch` locale we actually have
```
(* inside Arch locale, we have no need for the wrapper *)
lemmas [simp] = ghost_relation_wrapper_def
```
so we do not have to do anything extra to handle an `Arch.kernel_state` field being updated and modifying `ksArchState`.

corlewis: > For the second one, if all of `ksArchState` is the same, the lemmas as such will be true for AArch64. The question is whether you'll be able to prove that all of `ksArchState` remains the same if the ghost state has parts in it it that are being modified. That depends on which function you need it for.

I'm not quite sure what you mean with this. In my mind, in a generic context the most you can say about a function is whether it modifies `ksArchState` or not, you can't talk about whether it modifies the specific part of the arch state that `ghost_relation` cares about.

If we are dealing with such a function then to prove anything about it we will need to be in the `Arch` locale, in which case the wrapper will be automatically unfolded and, for example on AARCH64, we will have from the definition that `ghost_relation` only cares about `gsPTTypes`.

lsf37: > In my mind, in a generic context the most you can say about a function is whether it modifies ksArchState or not, you can't talk about whether it modifies the specific part of the arch state that ghost_relation cares about.

That's a valid point, you would in a generic context not reason about things that modify arch state -- lemmas about those functions would indeed be in the Arch locale.

> Are my proposed lemmas not good as simp rules because they could be unified as ghost_relation_wrapper s = ghost_relation_wrapper s and diverge? I thought the simplifier was smart enough not to do that?

It is smart enough for that part. The problem is that they always apply whenever `ghost_relation_wrapper` occurs in the goal, which means the simplifier will always try to satisfy the preconditions and most of the time will fail. If they are working in the situations where they are needed we could have them as interface rules, but not put them in the global simp set (only use explicitly locally where we know they work). Then they won't slow down the rest and will also not diverge when the attempted proof of the precondition diverges.

corlewis: Ah, I hadn't thought about performance. I do want something in the simp set though, in my mind the wrapper is something that should be handled automatically as much as possible, it isn't something that we want to have to think about.

With that in mind I've gone back to one of the earlier suggestions and written some hacky ML to produce individual simp rules for all of the appropriate state fields. I'll clean that up a bit and make a separate PR.

corlewis: Having said that, I've just realised that a much simpler solution would be to use our `_2` approach to have an abbreviation that exposes to generic proofs that `ghost_relation_wrapper` only depends on certain fields. With that the simplifier should 'just work' with no additional lemmas from us and it would be easy to keep updated if any architecture needs new fields.

lsf37: Yeah, that's probably the best solution. Although also the one that is most work to implement.

PR #934, opened 2025-11-19
Update SGI proofs for seL4 change

lsf37:
See https://github.com/seL4/seL4/pull/1552 for the original seL4 PR. The change is allowing more SGI targets for GICv3. The proof update required another small tweak in seL4/seL4#1559.

Proof changes:

- update ASpec/Haskell definitions to reflect C change
- change comparisons to `_ <= _ - 1` instead of `_ < _` to avoid overflow condition on 32 bit platforms
- fix proofs; this includes:
  - add word lemma for the `<=` comparison and `ucast` to avoid detour through more casting and plain `<`
  - cherry-pick word lemmas from `rt` that turn `0xFFFFFFFF` into `-1` for 32 bit (and same for 64). I have not found a nicer way to do this inline. The automation for `-1` turns everything you set up inline immediately into `True` and won't rewrite.
 
The one that failed (`imx8mm`) is not actually in the PR checks, but I have been running it locally (final check still underway).

Test with: seL4/seL4#1559

### DIFF
--- a/proof/crefine/AARCH64/Interrupt_C.thy
+++ b/proof/crefine/AARCH64/Interrupt_C.thy
@@ -358,14 +358,13 @@ lemma Arch_invokeIRQControl_IssueIRQHandler_ccorres:
   done
 
 lemma sgi_target_valid_sgi_target_len:
-  "isSGITargetValid w = (w < 2^sgi_target_len)"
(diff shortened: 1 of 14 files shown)


corlewis (approved):
Thanks for making this as generic as possible. Pity that the tests are too slow for the platforms to be included in the PR checks.

PR #936, opened 2025-11-24
refine: make ghost_relation_wrapper projections generic

corlewis:
Use the `_2` approach to make a `ghost_relation_wrapper` abbreviation that exposes to generic proofs the fields that `ghost_relation` can depend on. This allows generic proofs to automatically simplify most appearances of `ghost_relation_wrapper` without additional lemmas.

Stats:  24 files changed, 49 insertions(+), 269 deletions(-)
Total: 220 lines removed

### DIFF
--- a/proof/refine/AARCH64/ArchStateRelation.thy
+++ b/proof/refine/AARCH64/ArchStateRelation.thy
@@ -119,21 +119,23 @@ definition is_other_obj_relation_type :: "a_type \<Rightarrow> bool" where
     | _ \<Rightarrow> True"
 
 definition ghost_relation ::
-  "Structures_A.kheap \<Rightarrow> (machine_word \<rightharpoonup> vmpage_size) \<Rightarrow> (machine_word \<rightharpoonup> nat) \<Rightarrow> (machine_word \<rightharpoonup> pt_type) \<Rightarrow> bool" where
+  "kheap \<Rightarrow> (machine_word \<rightharpoonup> vmpage_size) \<Rightarrow> (machine_word \<rightharpoonup> nat) \<Rightarrow> (machine_word \<rightharpoonup> pt_type) \<Rightarrow> bool"
+  where
   "ghost_relation h ups cns pt_types \<equiv>
      (\<forall>a sz. (\<exists>dev. h a = Some (ArchObj (DataPage dev sz))) \<longleftrightarrow> ups a = Some sz) \<and>
      (\<forall>a n. (\<exists>cs. h a = Some (CNode n cs) \<and> well_formed_cnode_n n cs) \<longleftrightarrow> cns a = Some n) \<and>
      (\<forall>a pt_t. (\<exists>pt. h a = Some (ArchObj (PageTable pt)) \<and> pt_t = pt_type pt) \<longleftrightarrow> pt_types a = Some pt_t)"
 
-(* FIXME arch-split: provided only so that ghost_relation can be used within generic definition
+(* provided only so that ghost_relation can be used within generic definition
    of state_relation (since arch_state_relation doesn't have access to kheap, and
    gsPTTypes on AARCH64 isn't generic) *)
-definition ghost_relation_wrapper :: "det_state \<Rightarrow> kernel_state \<Rightarrow> bool" where
-  "ghost_relation_wrapper s s' \<equiv>
-     ghost_relation (kheap s) (gsUserPages s') (gsCNodes s') (gsPTTypes (ksArchState s'))"
+definition ghost_relation_wrapper_2 ::
+  "kheap \<Rightarrow> (machine_word \<rightharpoonup> vmpage_size) \<Rightarrow> (machine_word \<rightharpoonup> nat) \<Rightarrow> Arch.kernel_state \<Rightarrow> bool"
+  where
+  "ghost_relation_wrapper_2 h ups cns as \<equiv> ghost_relation h ups cns (gsPTTypes as)"
 
 (* inside Arch locale, we have no need for the wrapper *)
-lemmas [simp] = ghost_relation_wrapper_def
+lemmas ghost_relation_wrapper_def[simp] = ghost_relation_wrapper_2_def
 
 definition arch_state_relation :: "(arch_state \<times> AARCH64_H.kernel_state) set" where
   "arch_state_relation \<equiv> {(s, s') .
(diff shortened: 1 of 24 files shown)


corlewis:
> The main thing we are committing to here is that any future architecture-dependent ghost state is part of arch state only and makes no use of additional generic fields to construct its arch-specific state.
> 
> I think that is very reasonable. We have so far not needed additional generic fields in any of the extensions, and anything that was not generic, we have added to arch state, so this meshes with what has happened in the past. (And in the worst case, we'd have to update the _2 definition, which is not the end of the world either).

Yeah, in that case I think it would be pretty easy to update the definition and abbreviation. If anything it would be easier than what we would need to update with our alternative ideas.

proof/refine/AARCH64/Retype_R.thy
```diff
@@ -1377,7 +1377,7 @@ lemma retype_state_relation:
 
   from sr
   have gr: "ghost_relation (kheap s) (gsUserPages s') (gsCNodes s') (gsPTTypes (ksArchState s'))"
-    by (rule state_relationE[simplified ghost_relation_wrapper_def]) (* FIXME arch-split *)
+    by (rule state_relationE[simplified]) (* FIXME arch-split *)
```
Xaphiosis: I have a strange feeling that this `simplified` can now be removed, resolving the FIXME (same in Detype_R, and same on other arches)

proof/refine/CSpace1_R.thy
```diff
@@ -5177,7 +5177,7 @@ lemma cteInsert_corres:
             apply (thin_tac "ksIdleThread t = p" for p t)+
             apply (thin_tac "ksSchedulerAction t = p" for p t)+
             apply (rule conjI)
-             apply (erule (4) ghost_relation_wrapper_same_abs_set_cap)
+             apply (erule (1) ghost_relation_wrapper_same_abs_set_cap; rule refl)
```
Xaphiosis, line 5180: I'm a bit amazed that it changed the proof like this, yet without altering the lemma statements... can you enlighten me on what happened that resulted in this?

Xaphiosis (approved):
I'm a bit embarrassed that I did not notice this myself until a bunch of cruft accumulated. For some reason I wanted to safeguard potential for more future arch-specific ghost-state, which likely isn't going to materialise. Nice catch.

Could you do a `git diff --stat` to master and add that plus the total line delta to the PR description? It'll be useful for reporting yet more lines removed.

proof/refine/AARCH64/Retype_R.thy
corlewis: I had that feeling as well but the `simplified` was still needed. I didn't investigate why as I didn't have the appropriate sessions open at the time to compare what was happening. Do you want me to do so?

corlewis: Separately, I actually intended to remove the FIXME anyway, since I thought that it was mainly referring to unfolding `ghost_relation_wrapper_def`. Should I do that regardless of what happens with the simplified?

proof/refine/CSpace1_R.thy
corlewis: I was also surprised by this but again, didn't have the sessions open to compare and work out exactly what is happening. My guess is that the new approach leads to some simplification happening earlier than it previously did, resulting in goals here that don't exactly match assumptions.

Xaphiosis: Cool, more simplification is good to see.

proof/refine/AARCH64/Retype_R.thy
Xaphiosis: Remember that `simplified` does not do anything additional unless it's lonely. So `simplified blah` is not the same as `simp add: blah`, but `simplified` is analogous to `simp`.
i.e. both the `simplified` and the `FIXME` should be removed.

corlewis: Yes, but I meant that I did try removing the `simplified` but the proof stopped working.

corlewis: Having investigated further, what is happening is that the current proof does want to unfold the `ghost_relation_wrapper` to get to the `ghost_relation`. As this is in the `Arch` locale, where `ghost_relation_wrapper_def` is already `[simp]`, both `simplified` and `simplified ghost_relation_wrapper_def` are doing the same thing.

My feeling is that this won't be too hard to fix but that there will be some arch-splitting needed here, so I think I'll leave the `FIXME`. I'm also thinking that I might put the `ghost_relation_wrapper_def` back in as well, to make it easier to notice what is happening when this is being arch-split.

Xaphiosis: Ahh, ok now I understand. I guess now with the extra simplification from the ghost state wrapper being broken up it does something, meaning it's needed again.
It might be worth seeing what the simplified `state_relationE` actually is to see whether it should be rephrased, but honestly am fine with only removing the FIXMEs.

Xaphiosis: Wow we typed past each other again. Your solution to leave the def and the FIXME makes sense. Thanks for investigating.

proof/refine/AARCH64/ArchStateRelation.thy
```diff
@@ -128,12 +129,13 @@ definition ghost_relation ::
 (* FIXME arch-split: provided only so that ghost_relation can be used within generic definition
    of state_relation (since arch_state_relation doesn't have access to kheap, and
    gsPTTypes on AARCH64 isn't generic) *)
```
corlewis: It's unclear to me if this FIXME is still needed. Should I remove it, or do we think that we might revisit the idea of the ghost_relation_wrapper?

Xaphiosis: Dropping the FIXME makes sense I guess, but does keeping the comment make sense?

corlewis: I think it does? I guess my new comment in `StateRelation.thy` duplicates part of it but this still explains why it isn't in `arch_state_relation` and indicates that `AARCH64` is a bit special.

Xaphiosis: Cool, then let's do that (drop FIXME, keep comment)

corlewis:
Updated with final changes to comments, tests previously passed.

PR #939, opened 2025-12-01
refine+crefine: make valid_arch_obj' stateless

Xaphiosis:
Due to ARM_HYP, all architectures were conforming to an interface which required valid_arch_obj' to depend on state. It turns out that this is not necessary. Remove the dependency on state and clean up.

I need this one for arch-splitting Retype_R, due to some assumptions in the interface. Also, this inconsistency was annoying in general.

43 files changed, 195 insertions(+), 523 deletions(-)
Total: 328 lines deleted (i.e. very minor cleanup)

### DIFF
--- a/proof/crefine/ARM_HYP/Finalise_C.thy
+++ b/proof/crefine/ARM_HYP/Finalise_C.thy
@@ -2089,29 +2089,21 @@ lemma associateVCPUTCB_ccorres:
      apply (rule ccorres_pre_getObject_vcpu, rename_tac vcpu)
      apply (rule ccorres_move_c_guard_vcpu)
      apply (rule ccorres_split_nothrow[where r'=dc and xf'=xfdc])
-         apply (rule_tac Q="(\<lambda>s. \<exists>tcb. ko_at' vcpu vcpuptr s) and
+         apply (rule_tac Q="ko_at' vcpu vcpuptr and
                             no_0_obj' and
                             valid_objs'"
                      and Q'=UNIV
                      and C'="{s. (vcpuTCBPtr vcpu \<noteq> None)}"
                      in ccorres_rewrite_cond_sr)
+
           apply clarsimp
-          apply (frule cmap_relation_vcpu)
-          apply (erule cmap_relationE1)
-           apply (erule ko_at_projectKO_opt, rename_tac cvcpu)
-          apply (clarsimp simp add: typ_heap_simps')
-          apply (case_tac "vcpuTCBPtr vcpu \<noteq> None")
-           apply(frule valid_objs_valid_vcpu', simp)
-           apply (clarsimp simp add: cvcpu_relation_def
-                                     option_to_ctcb_ptr_def
-                                     tcb_ptr_to_ctcb_ptr_def
-                                     typ_at_tcb'
-                                     valid_vcpu'_def)
-           apply (frule tcb_aligned')
-           apply (frule_tac y=ctcb_offset and n=tcbBlockSizeBits in aligned_offset_non_zero
-                 ; clarsimp simp: ctcb_offset_defs objBits_defs)
-          apply (clarsimp simp: cvcpu_relation_def option_to_ctcb_ptr_def)
-         apply (wpc ; clarsimp ; ccorres_rewrite)
+          apply (frule (1) ko_at_valid_objs', simp)
+          apply (clarsimp simp: valid_obj'_def valid_vcpu'_def)
+          apply (drule (1) vcpu_at_rf_sr)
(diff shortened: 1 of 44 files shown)


proof/crefine/ARM_HYP/Finalise_C.thy
```diff
@@ -2055,6 +2055,26 @@ lemma dissociateVCPUTCB_vcpu_ccorres:
      (dissociateVCPUTCB vcpuptr tptr) (Call dissociateVCPUTCB_'proc)"
   by (rule ccorres_guard_imp2[OF dissociateVCPUTCB_ccorres]) clarsimp
 
+(* FIXME arch-split: move to Wellformed above tcb_at_not_NULL / generic lemmas with arch-specific ... RAF *)
+lemma aligned_tcb_ctcb_not_NULL:
```
lsf37: Did you want to move this one now or leave for later?

proof/crefine/ARM_HYP/Finalise_C.thy
```diff
@@ -2089,29 +2109,21 @@ lemma associateVCPUTCB_ccorres:
      apply (rule ccorres_pre_getObject_vcpu, rename_tac vcpu)
      apply (rule ccorres_move_c_guard_vcpu)
      apply (rule ccorres_split_nothrow[where r'=dc and xf'=xfdc])
-         apply (rule_tac Q="(\<lambda>s. \<exists>tcb. ko_at' vcpu vcpuptr s) and
+         apply (rule_tac Q="ko_at' vcpu vcpuptr and
```
lsf37, line 2092: Interesting existential we had there.. I'm glad we're cleaning up some of this old stuff.

proof/crefine/ARM_HYP/Finalise_C.thy
Xaphiosis: Oh yes, definitely now, I yoinked it from AARCH64 so should match the location there. Thanks.

proof/refine/ARM/ArchInvsDefs_H.thy
```diff
@@ -138,27 +138,27 @@ lemmas [simp] = arch_valid_irq_def
...
- "valid_mapping' x sz s \<equiv> is_aligned x (pageBitsForSize sz)
-                          \<and> ptrFromPAddr x \<noteq> 0"
+definition valid_mapping' :: "word32 \<Rightarrow> vmpage_size \<Rightarrow> bool" where
+ "valid_mapping' x sz \<equiv> is_aligned x (pageBitsForSize sz)
+                        \<and> ptrFromPAddr x \<noteq> 0"
 
-primrec valid_pte' :: "ARM_H.pte \<Rightarrow> kernel_state \<Rightarrow> bool" where
-  "valid_pte' (InvalidPTE) = \<top>"
+primrec valid_pte' :: "ARM_H.pte \<Rightarrow> bool" where
+  "valid_pte' (InvalidPTE) = \<top>rue"
```
corlewis: Is this an issue with our unicode script or is the `T` in this actually `\<top>`?

It looks like it is, but then how are the tests passing?

corlewis: Huh, I just tested it out and Isabelle accepts that as `True`. Is that something that we did or is `\<top>` set up somewhere deeper in Isabelle to be read as `T`?

Xaphiosis: That... is terrifying. On my screen the font makes T and \<top> look basically the same. Thanks for spotting that! I'll use vim to fix it

corlewis: Yes, even if Isabelle is happy with it, I'd definitely prefer T. Just in case you don't notice, it's also in `valid_pde'` and the equivalent `ARM_HYP` and `X64` definitions.

lsf37: It makes perfect sense unfortunately: \<top> is an abbrievation for `%_. True`, and `\<top>rue` is `\<top> rue` which reduces as lambda term to just `True`, which is what Isabelle actually sees.

lsf37: You don't want this one to be `\<top>` (it doesn't take any state argument any more), it should be just `True`.

Xaphiosis: These were meant to be `True`, issue was my not noticing that `True` and `⊤rue` are different.

proof/crefine/ARM_HYP/Finalise_C.thy
Xaphiosis: Must be from an era where there was a tcb_at' around, but yes, unusual.

PR #940, opened 2025-12-03
arch-split Retype_R

Xaphiosis:
🦆🦆🦆 This PR has the same diff-reducing tech as the TcbAcc_R PR, copied instructions:

In real life, I did the usual addition of Retype_R, hierarchy update, arch-split and then update. However, since the arch-split part first splits AARCH64, then takes the AARCH64 version of ArchCSpace_R and only fixes what broke, we ended up with a small(er) diff there (again).

- `[squash] PR: copy AARCH64 version into Retype_R` - same as last time: stop Retype_R looking like a wall of green (no need to review)
- `aarch64 refine: arch-split Retype_R` - this is the main bulk of the arch-split, dealing with AARCH64
- `[squash] copy AARCH64 ArchRetype_R to other arches` - wipes the *body* (not the header!) of other arches' ArchRetype_R with a copy of the AARCH64 version with `AARCH64` appropriately substituted (as @corlewis requested last time) - (mostly pointless to review)
- `[squash] refine: arch-split Retype_R (other arches) - result of fixing up the ArchRetype_R of other architectures - shows diff to AARCH64, small for this PR
- `refine+crefine: update for ... arch-split`

Reminder: DO NOT MERGE squashed version if it still says `aarch64 refine: arch-split`

Stats: 61 files changed, 11343 insertions(+), 24510 deletions(-)
Total: 13167 lines removed

### DIFF
--- a/proof/crefine/AARCH64/Retype_C.thy
+++ b/proof/crefine/AARCH64/Retype_C.thy
@@ -4381,11 +4381,11 @@ lemma getObjectSize_symb:
   apply (case_tac newType)
    apply simp_all
    apply (simp_all add: object_type_from_H_def Kernel_C_defs
-                        APIType_capBits_def objBits_simps')
+                        APIType_capBits_def APIType_capBits_gen_def objBits_simps')
    apply (rename_tac apiobject_type)
    apply (case_tac apiobject_type)
    apply (simp_all add: object_type_from_H_def Kernel_C_defs frameSizeConstants_defs
-                        APIType_capBits_def objBits_simps' bit_simps
+                        objBits_simps' bit_simps
                    split: if_split)
    apply unat_arith
   (* True in either config with corresponding max vm level change *)
@@ -6323,7 +6323,7 @@ proof -
                           word_le_nat_alt split:
                           apiobject_type.splits object_type.splits)
    apply (subgoal_tac "\<exists>apiType. newType = APIObjectType apiType")
-    apply clarsimp
+    apply (clarsimp simp: APIType_capBits_gen_def)
     apply (rule ccorres_guard_imp)
       apply (rule_tac apiType=apiType in ccorres_apiType_split)
 
@@ -6383,10 +6383,9 @@ proof -
          apply (frule invs_pspace_distinct')
          apply (frule invs_sym')
          apply (simp add: getObjectSize_def objBits_simps word_bits_conv
-                          apiGetObjectSize_def
+                          apiGetObjectSize_def APIType_capBits_gen_def
                           tcbBlockSizeBits_def new_cap_addrs_def projectKO_opt_tcb)
-         apply (clarsimp simp: range_cover.aligned
-                               region_actually_is_bytes_def APIType_capBits_def)
+         apply (clarsimp simp: range_cover.aligned region_actually_is_bytes_def)
          apply (frule(1) ghost_assertion_size_logic_no_unat)
          apply (clarsimp simp: ccap_relation_def cap_to_H_def
                                getObjectSize_def apiGetObjectSize_def
@@ -6427,7 +6426,7 @@ proof -
         apply (frule invs_pspace_distinct')
         apply (frule invs_sym')
         apply (auto simp: getObjectSize_def objBits_simps apiGetObjectSize_def
-                          epSizeBits_def word_bits_conv
+                          epSizeBits_def word_bits_conv APIType_capBits_gen_def
                    elim!: is_aligned_no_wrap'
                   intro!: range_cover_simpleI)[1]
 
@@ -6463,7 +6462,7 @@ proof -
        apply (frule invs_pspace_aligned')
        apply (frule invs_pspace_distinct')
        apply (frule invs_sym')
-       apply (auto simp: getObjectSize_def objBits_simps
+       apply (auto simp: getObjectSize_def objBits_simps APIType_capBits_gen_def
                    apiGetObjectSize_def
                    ntfnSizeBits_def word_bits_conv
                 elim!: is_aligned_no_wrap'  intro!: range_cover_simpleI)[1]
@@ -6505,7 +6504,7 @@ proof -
        apply (frule invs_sym')
        apply (frule(1) ghost_assertion_size_logic_no_unat)
        apply (clarsimp simp: getObjectSize_def objBits_simps
-                  apiGetObjectSize_def
+                  apiGetObjectSize_def APIType_capBits_gen_def
                   cteSizeBits_def word_bits_conv add.commute createObject_c_preconds_def
                   region_actually_is_bytes_def
                   invs_valid_objs' invs_urz
@@ -6515,7 +6514,7 @@ proof -
       apply (frule range_cover.strong_times_64[folded addr_card_wb], simp+)
       apply (subst h_t_array_valid_retyp, simp+)
        apply (simp add: power_add cte_C_size cteSizeBits_def)
-      apply (clarsimp simp: ccap_relation_def cap_to_H_def cap_cnode_cap_lift
+      apply (clarsimp simp: ccap_relation_def cap_to_H_def cap_cnode_cap_lift  APIType_capBits_gen_def
                             getObjectSize_def apiGetObjectSize_def cteSizeBits_def
                             objBits_simps field_simps is_aligned_power2
                             addr_card_wb is_aligned_weaken[where y=2]
@@ -6532,7 +6531,7 @@ proof -
        apply (clarsimp simp: word_bits_conv)
       apply simp
      apply auto[1]
-    apply (clarsimp simp: createObject_c_preconds_def)
+    apply (clarsimp simp: createObject_c_preconds_def APIType_capBits_gen_def)
     apply (clarsimp simp:nAPIOBjects_object_type_from_H)?
     apply (intro impI conjI, simp_all)[1]
    apply (clarsimp simp: nAPIObjects_def object_type_from_H_def Kernel_C_defs
@@ -7502,6 +7501,7 @@ lemma createObject_untypedRange:
             (toAPIType ty \<noteq> Some apiobject_type.Untyped \<longrightarrow> Q {} s)\<rbrace>"
   shows "\<lbrace>P\<rbrace> createObject ty ptr us dev\<lbrace>\<lambda>m s. Q (untypedRange m) s\<rbrace>"
   including no_pre
+  supply toAPIType_Some[simp del]
   using split
   apply (simp add: createObject_def)
   apply (case_tac "toAPIType ty")
@@ -7516,7 +7516,7 @@ lemma createObject_capRange:
 shows "\<lbrace>P\<rbrace>createObject ty ptr us dev \<lbrace>\<lambda>m s. capRange m = {ptr.. ptr + 2 ^ (APIType_capBits ty us) - 1}\<rbrace>"
   apply (simp add:createObject_def)
   apply (case_tac "ty")
-    apply (simp_all add:toAPIType_def AARCH64_H.toAPIType_def)
+    apply (simp_all add: toAPIType_def APIType_capBits_gen_def)
         apply (rule hoare_pre)
          apply wpc
              apply wp
@@ -7787,8 +7787,8 @@ shows
   {}\<rbrace>"
   apply (rule createObject_untypedRange)
   apply (clarsimp | wp)+
-  apply (clarsimp simp: blah toAPIType_def APIType_capBits_def
-    AARCH64_H.toAPIType_def split: object_type.splits)
+  apply (clarsimp simp: blah toAPIType_def APIType_capBits_gen_def
+                        toAPIType_def split: object_type.splits)
   apply (clarsimp simp:shiftl_t2n field_simps)
   apply (drule word_eq_zeroI)
   apply (drule(1) range_cover_no_0[where p = "Suc n"])
@@ -7998,13 +7998,13 @@ lemma createObject_cnodes_have_size:
       \<and> cnodes_retype_have_size R (APIType_capBits newType userSize) (gsCNodes s)\<rbrace>
     createObject newType ptr userSize dev
   \<lbrace>\<lambda>rv s. cnodes_retype_have_size R (APIType_capBits newType userSize) (gsCNodes s)\<rbrace>"
+  supply toAPIType_Some[simp del]
   apply (simp add: createObject_def)
   apply (rule hoare_pre)
    apply (wp mapM_x_wp' | wpc | simp add: createObjects_def)+
-  apply (cases newType, simp_all add: AARCH64_H.toAPIType_def)
-  apply (clarsimp simp: APIType_capBits_def objBits_simps'
-                              cnodes_retype_have_size_def cte_level_bits_def
-                       split: if_split_asm)
+  apply (cases newType, simp_all add: toAPIType_def APIType_capBits_gen_def)
+  apply (clarsimp simp: objBits_simps' cnodes_retype_have_size_def cte_level_bits_def
+                  split: if_split_asm)
   done
 
 crunch placeNewDataObject
@@ -8017,6 +8017,7 @@ lemma createObject_cnodes_have_size_pt:
     createObject newType ptr userSize dev
   \<lbrace>\<lambda>rv s. cnodes_retype_have_size R (APIType_capBits newType userSize) (gsPTTypes (ksArchState s) ||> (\<lambda>pt_t. pt_bits pt_t - cte_level_bits))\<rbrace>"
   supply fun_upd_apply[simp del] fun_upd_def[symmetric, simp]
+  supply toAPIType_Some[simp del]
   apply (simp add: createObject_def)
   apply (rule hoare_pre)
    apply (wp mapM_x_wp' | wpc | simp add: createObjects_def AARCH64_H.createObject_def updatePTType_def)+
@@ -8570,8 +8571,7 @@ shows  "ccorres dc xfdc
             apply simp
            apply (simp only: unat_eq_0, clarsimp simp: unat_of_nat)
           apply (frule range_cover_sz')
-          apply (clarsimp simp: Let_def hrs_htd_update
-                                APIType_capBits_def[where ty="APIObjectType ArchTypes_H.apiobject_type.Untyped"])
+          apply (clarsimp simp: Let_def hrs_htd_update APIType_capBits_gen_def)
          apply (simp, subst range_cover.unat_of_nat_n)
           apply (erule range_cover_le)
           subgoal by simp
--- a/proof/refine/X64/Detype_R.thy
+++ b/proof/refine/X64/Detype_R.thy
@@ -1020,38 +1020,41 @@ lemma deleteObjects_corres:
     apply (simp add: valid_pspace'_def)
     apply (rule state_relation_null_filterE, assumption,
            simp_all add: pspace_aligned'_cut pspace_distinct'_cut)[1]
-           apply (simp add: detype_def)
-          apply (intro exI, fastforce)
-         apply (rule ext, clarsimp simp add: null_filter_def)
+            apply (simp add: detype_def)
+           apply clarsimp
+           (* unification can't guess we want identity update on ksArchState s' *)
+           apply (repeat 3 \<open>rule exI\<close>, rule_tac x=id in exI)
+           apply fastforce
+          apply (rule ext, clarsimp simp add: null_filter_def)
+          apply (rule sym, rule ccontr, clarsimp)
+          apply (drule(4) cte_map_not_null_outside')
+           apply (fastforce simp add: cte_wp_at_caps_of_state)
+          apply simp
+         apply (rule ext, clarsimp simp add: null_filter'_def
+                            map_to_ctes_delete[simplified field_simps])
          apply (rule sym, rule ccontr, clarsimp)
-         apply (drule(4) cte_map_not_null_outside')
-          apply (fastforce simp add: cte_wp_at_caps_of_state)
+         apply (frule(2) pspace_relation_cte_wp_atI
+                         [OF state_relation_pspace_relation])
+         apply (elim exE)
+         apply (frule(4) cte_map_not_null_outside')
+          apply (rule cte_wp_at_weakenE, erule conjunct1)
+          apply (case_tac y, clarsimp)
+          apply (clarsimp simp: valid_mdb'_def valid_mdb_ctes_def
+                                valid_nullcaps_def)
+         apply clarsimp
+         apply (frule_tac cref="(aa, ba)" in cte_map_untyped_range,
+                erule cte_wp_at_weakenE[OF _ TrueI], assumption+)
          apply simp
-        apply (rule ext, clarsimp simp add: null_filter'_def
-                           map_to_ctes_delete[simplified field_simps])
-        apply (rule sym, rule ccontr, clarsimp)
-        apply (frule(2) pspace_relation_cte_wp_atI
-                        [OF state_relation_pspace_relation])
-        apply (elim exE)
-        apply (frule(4) cte_map_not_null_outside')
-         apply (rule cte_wp_at_weakenE, erule conjunct1)
-         apply (case_tac y, clarsimp)
-         apply (clarsimp simp: valid_mdb'_def valid_mdb_ctes_def
-                               valid_nullcaps_def)
-        apply clarsimp
-        apply (frule_tac cref="(aa, ba)" in cte_map_untyped_range,
-               erule cte_wp_at_weakenE[OF _ TrueI], assumption+)
-        apply simp
-       apply (rule detype_pspace_relation[simplified],
-              simp_all add: state_relation_pspace_relation valid_pspace_def)[1]
-        apply (simp add: valid_cap'_def capAligned_def)
-       apply (clarsimp simp: valid_cap_def, assumption)
-      apply (rule detype_ready_queues_relation; blast?)
-       apply (clarsimp simp: deletionIsSafe_delete_locale_def)
-      apply (frule state_relation_ready_queues_relation)
-      apply (simp add: ready_queues_relation_def Let_def)
-     apply (clarsimp simp: state_relation_def ghost_relation_of_heap
-                           detype_def)
+        apply (rule detype_pspace_relation[simplified],
+               simp_all add: state_relation_pspace_relation valid_pspace_def)[1]
+         apply (simp add: valid_cap'_def capAligned_def)
+        apply (clarsimp simp: valid_cap_def, assumption)
+       apply (rule detype_ready_queues_relation; blast?)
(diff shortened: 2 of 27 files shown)


proof/refine/Retype_R.thy
```diff
@@ -12,88 +12,37 @@ theory Retype_R
...
-    | Inr (APIObjectType ArchTypes_H.NotificationObject) \<Rightarrow> Structures_A.NotificationObject
-    | Inr (APIObjectType ArchTypes_H.CapTableObject) \<Rightarrow> Structures_A.CapTableObject
-    | Inr LargePageObject \<Rightarrow> ArchObject LargePageObj
-    | Inr HugePageObject \<Rightarrow> ArchObject HugePageObj
-    | Inr PageTableObject \<Rightarrow> ArchObject PageTableObj
-    | Inr VSpaceObject \<Rightarrow> ArchObject VSpaceObj
-    | Inl (KOArch (KOASIDPool _)) \<Rightarrow> ArchObject ASIDPoolObj
-    | Inr VCPUObject \<Rightarrow> ArchObject AARCH64_A.VCPUObj
-    | _ \<Rightarrow> ArchObject SmallPageObj"
+lemmas [simp] = objBitsKO_less_word_bits
```
corlewis: Should we make this `[simp]` directly in the interface in `Bits_R`?

proof/refine/Retype_R.thy
```diff
@@ -1,92 +1,48 @@
...
 
 lemma placeNewObject_def2:
- "placeNewObject ptr val gb = createObjects' ptr 1 (injectKO val) gb"
-   apply (clarsimp simp:placeNewObject_def placeNewObject'_def
-     createObjects'_def shiftL_nat)
-  done
+  "placeNewObject ptr val gb = createObjects' ptr 1 (injectKO val) gb"
+  by (clarsimp simp:placeNewObject_def placeNewObject'_def createObjects'_def shiftL_nat)
```
corlewis: Since we're already cleaning this up.
```suggestion
  by (clarsimp simp: placeNewObject_def placeNewObject'_def createObjects'_def shiftL_nat)
```

proof/refine/Retype_R.thy
```diff
@@ -484,6 +334,318 @@ lemma ps_clearD:
...
+definition obj_relation_retype :: "Structures_A.kernel_object \<Rightarrow> Structures_H.kernel_object \<Rightarrow> bool"
+  where
+  "obj_relation_retype ko ko' \<equiv>
+    obj_bits ko \<ge> objBitsKO ko'
+    \<and> (\<forall>p. fst ` obj_relation_cuts ko p
+           = {p + x * 2 ^ (objBitsKO ko') | x. x < 2 ^ (obj_bits ko - objBitsKO ko')}
+           \<and> (\<forall>x \<in> obj_relation_cuts ko p. snd x ko ko'))"
+
+abbreviation
+ "injectKOS \<equiv> (injectKO :: ('a :: pspace_storable) \<Rightarrow> kernel_object)"
```
corlewis: I know you just moved it but while we're here
```suggestion
  "injectKOS \<equiv> (injectKO :: ('a :: pspace_storable) \<Rightarrow> kernel_object)"
```

proof/refine/Retype_R.thy
```diff
@@ -501,6 +334,318 @@ lemma ps_clearD:
...
+
+lemma createObjects_createObjects'_inv:
+  "createObjects ptr numObjects val gSize \<lbrace>P\<rbrace>
+   = createObjects' ptr numObjects val gSize \<lbrace>P\<rbrace>"
+  unfolding createObjects_def
+  by (fastforce simp: createObjects_def valid_def bind_def return_def)
+
+locale Retype_R =
+  assumes valid_obj_makeObject_tcb[simp]:
+    "valid_obj' (KOTCB makeObject) s"
```
corlewis: Would it make sense to have a new lemma along the lines of `valid_arch_tcb' newArchTCB s` in the interface instead of this? Or is that just giving us more work for not much gain.

proof/refine/Retype_R.thy
```diff
@@ -501,6 +334,318 @@ lemma ps_clearD:
...
+  assumes APIType_map2_Untyped[simp]:
+    "\<And>tp. (APIType_map2 tp = Structures_A.Untyped) = (tp = Inr (APIObjectType ArchTypes_H.Untyped))"
+  assumes APIType_map2_TCBObject[simp]:
+    "\<And>tp. (APIType_map2 tp = Structures_A.TCBObject) = (tp = Inr (APIObjectType ArchTypes_H.TCBObject))"
+  assumes APIType_capBits_generic[simp]:
+    "\<And>api us. APIType_capBits (APIObjectType api) us = APIType_capBits_gen api us"
+  assumes toAPIType_Some[simp]:
+    "\<And>ty x. (toAPIType ty = Some x) = (ty = APIObjectType x)"
+  assumes obj_relation_cuts_trivial:
+    "\<And>ptr ty. ptr \<in> fst ` obj_relation_cuts ty ptr"
```
corlewis: Should this be moved to an earlier interface?

proof/refine/Retype_R.thy
```diff
@@ -1607,1195 +1350,904 @@ next
...
-               split del: if_split)
-  apply (erule allEI, erule allEI)
-  apply (clarsimp split del: if_split)
-  apply (clarsimp split: if_split_asm simp: makeObject_cte)
-  apply (clarsimp simp: isCap_simps)
-  done
+lemma tcb_relation_retype:
+  "obj_relation_retype (default_object Structures_A.TCBObject dev n d)
+                       (KOTCB (tcbDomain_update (\<lambda>_. d) makeObject))"
+  supply tcb_bits_def[simp del] (* FIXME arch-split: we might not want this in the simpset *)
```
corlewis, line 2193: Hmm, yeah, it does seem strange to have this in the simpset when it's an arch constant

proof/refine/Retype_R.thy
```diff
@@ -1607,1195 +1350,904 @@ next
...
+  sum.split_asm kernel_object.split_asm
 
-definition
-  caps_no_overlap'' :: "machine_word \<Rightarrow> nat \<Rightarrow> kernel_state \<Rightarrow> bool"
-where
- "caps_no_overlap'' ptr sz s \<equiv> \<forall>cte \<in> ran (ctes_of s).
-               untypedRange (cteCap cte) \<inter> {ptr .. (ptr && ~~ mask sz) + 2 ^ sz - 1} \<noteq> {}
-               \<longrightarrow> {ptr .. (ptr && ~~ mask sz) + 2 ^ sz - 1} \<subseteq> untypedRange (cteCap cte)"
+declare hoare_in_monad_post[wp del]
+declare univ_get_wp[wp del]
```
corlewis, line 2259: Not relevant to this PR at all but these declares should probably have a FIXME on them to be cleaned up.

proof/refine/AARCH64/ArchRetype_R.thy
```diff
@@ -30,109 +32,80 @@ where
...
-lemma valid_obj_makeObject_tcb [simp]:
+lemma valid_obj_makeObject_tcb[Retype_R_assms, simp]:
   "valid_obj' (KOTCB makeObject) s"
   unfolding valid_obj'_def valid_tcb'_def valid_tcb_state'_def valid_arch_tcb'_def
   by (clarsimp simp: makeObject_tcb makeObject_cte tcb_cte_cases_def minBound_word newArchTCB_def
                      cteSizeBits_def)
 
-lemma valid_obj_makeObject_tcb_tcbDomain_update [simp]:
+lemma valid_obj_makeObject_tcb_tcbDomain_update[simp]:
   "d \<le> maxDomain \<Longrightarrow> valid_obj' (KOTCB (tcbDomain_update (\<lambda>_. d) makeObject)) s"
```
corlewis: Similarly to `valid_obj_makeObject_tcb`, not sure if it's worth doing but this is another lemma that feels like it could be made generic with a lower-level interface.

proof/refine/AARCH64/ArchRetype_R.thy
```diff
@@ -0,0 +1,1869 @@
...
+
+  assumes ptr_cn: "canonical_address (ptr && ~~ mask sz)"
+  assumes sz_constrained: "sz \<le> maxUntypedSizeBits"
+
+  shows "\<lbrace>\<lambda>s. pspace_no_overlap' ptr sz s \<and> valid_pspace' s\<rbrace>
+           createNewCaps ty ptr n us dev
+         \<lbrace>\<lambda>r s. (\<forall>cap \<in> set r. s \<turnstile>' cap)\<rbrace>"
+proof -
+  note blah[simp del] = untyped_range.simps usable_untyped_range.simps atLeastAtMost_iff atLeastatMost_subset_iff atLeastLessThan_iff
+          Int_atLeastAtMost atLeastatMost_empty_iff split_paired_Ex
```
corlewis: Potential cleanup
```suggestion
  note [simp del] = untyped_range.simps usable_untyped_range.simps atLeastAtMost_iff atLeastatMost_subset_iff atLeastLessThan_iff
          Int_atLeastAtMost atLeastatMost_empty_iff split_paired_Ex
```

proof/refine/AARCH64/ArchRetype_R.thy
```diff
@@ -3466,572 +669,659 @@ lemma createNewCaps_cte_wp_at':
...
-  and     not_0: "n\<noteq> 0"
-  shows  "\<lbrace>\<lambda>s. obj_at' P p s \<and> valid_pspace' s \<and> pspace_no_overlap' ptr sz s\<rbrace>
-  createObjects ptr n val gbits \<lbrace>\<lambda>_. obj_at' P p\<rbrace>"
-  apply (simp add: createObjects_def)
-  apply (wp createObjects_orig_obj_at'[where sz = sz])
-  using cover not_0
-  apply (clarsimp simp: cover not_0 valid_pspace'_def pspace_no_overlap'_def)
+(* FIXME arch-split: example of arch-split attempt of this kind of proof;
+   unfortunately splitting off the arch-specific part doesn't actually save space, so we will leave
+   these in Arch *)
```
corlewis: Ah, that's a pity. I guess we should keep some sort of comment like this to avoid trying to split it again, but should it stay as a FIXME? I'm not sure what is left to fix.

proof/refine/AARCH64/ArchRetype_R.thy
```diff
@@ -5130,6 +1641,7 @@ lemma corres_retype_region_createNewCaps:
                 init_arch_objects (APIType_map2 (Inr ty)) dev y n us x;
                 return x od)
             (createNewCaps ty y n us dev)"
+  supply APIType_map2_generic[simp del]
```
corlewis, line 1621: This and `APIType_capBits_generic[simp del]` were needed a few times, I'm guessing to avoid warnings. Does that mean that the default simpset should be slightly tweaked inside `Arch` to avoid duplication?

proof/refine/ARM/ArchRetype_R.thy
```diff
@@ -0,0 +1,2122 @@
...
+         (\<lambda>cns x. if x \<in> ptrs then Some us else cns x)
+     | ArchObject (SmallPageObj) \<Rightarrow> gsUserPages_update
+         (\<lambda>ups x. if x \<in> ptrs then Some ARMSmallPage else ups x)
+     | ArchObject (LargePageObj) \<Rightarrow> gsUserPages_update
+         (\<lambda>ups x. if x \<in> ptrs then Some ARMLargePage else ups x)
+     | ArchObject (SectionObj) \<Rightarrow> gsUserPages_update
+         (\<lambda>ups x. if x \<in> ptrs then Some ARMSection else ups x)
+     | ArchObject (SuperSectionObj) \<Rightarrow> gsUserPages_update
+         (\<lambda>ups x. if x \<in> ptrs then Some ARMSuperSection else ups x)
+     | _ \<Rightarrow> id"
```
corlewis, line 166:
```suggestion
  "update_gs ty us ptrs \<equiv> case ty of
       Structures_A.CapTableObject \<Rightarrow> gsCNodes_update
         (\<lambda>cns x. if x \<in> ptrs then Some us else cns x)
     | ArchObject SmallPageObj \<Rightarrow> gsUserPages_update
         (\<lambda>ups x. if x \<in> ptrs then Some ARMSmallPage else ups x)
     | ArchObject LargePageObj \<Rightarrow> gsUserPages_update
         (\<lambda>ups x. if x \<in> ptrs then Some ARMLargePage else ups x)
     | ArchObject SectionObj \<Rightarrow> gsUserPages_update
         (\<lambda>ups x. if x \<in> ptrs then Some ARMSection else ups x)
     | ArchObject SuperSectionObj \<Rightarrow> gsUserPages_update
         (\<lambda>ups x. if x \<in> ptrs then Some ARMSuperSection else ups x)
     | _ \<Rightarrow> id"
```

proof/refine/ARM/ArchRetype_R.thy
```diff
@@ -0,0 +1,2122 @@
...
+         apply (simp_all add:objBits_simps' pageBits_def)
+  apply (rename_tac arch_kernel_object)
+  apply (case_tac arch_kernel_object)
+    apply (simp_all add: objBits_simps' vspace_bits_defs)
+  done
+
+lemma cwo_ret:
+  assumes  cover:"range_cover ptr sz v n"
+  assumes not_0:"n\<noteq> 0"
+  shows result: "\<lbrace>pspace_no_overlap' ptr sz and valid_pspace' and K (v = 12 + bs)\<rbrace>
```
corlewis: Might as well make this consistent across architectures
```suggestion
  assumes not_0: "n\<noteq> 0"
  assumes cover: "range_cover ptr sz v n"
  shows "\<lbrace>pspace_no_overlap' ptr sz and valid_pspace' and K (v = 12 + bs)\<rbrace>
```

proof/refine/ARM/ArchRetype_R.thy
```diff
@@ -301,25 +293,21 @@ lemma range_cover_canonical_address':
   using unat_less_helper by blast
 
 lemma createNewCaps_valid_cap:
-  fixes ptr :: machine_word
+  fixes ptr :: word32
```
corlewis: I'm guessing that the `machine_word` used by `AARCH64` would be better than the `word32`?

proof/refine/ARM/ArchRetype_R.thy
```diff
@@ -301,25 +293,21 @@ lemma range_cover_canonical_address':
...
            createNewCaps ty ptr n us dev
          \<lbrace>\<lambda>r s. (\<forall>cap \<in> set r. s \<turnstile>' cap)\<rbrace>"
 proof -
   note blah[simp del] = untyped_range.simps usable_untyped_range.simps atLeastAtMost_iff atLeastatMost_subset_iff atLeastLessThan_iff
           Int_atLeastAtMost atLeastatMost_empty_iff split_paired_Ex
-  note if_split_def[split del] = if_split
-  note createObjects_nonzero' = createObjects_nonzero[OF not_0]
-  note cwo_ret' = cwo_ret[OF not_0]
+  note if_split[split del]
+  note projectKOs[simp del]
```
corlewis: This should possibly have a FIXME, I feel like we now want to move towards all of the proofs consistently using `projectKOs[simp]`.

proof/refine/ARM/ArchRetype_R.thy
corlewis: Same for `ARM_HYP`

proof/refine/ARM/ArchRetype_R.thy
corlewis: Same for `ARM_HYP`

proof/refine/ARM/ArchRetype_R.thy
corlewis: Same for `ARM_HYP`

proof/refine/ARM/ArchRetype_R.thy
corlewis: Same for `ARM_HYP`

proof/crefine/AARCH64/Retype_C.thy
```diff
@@ -7502,6 +7501,7 @@ lemma createObject_untypedRange:
             (toAPIType ty \<noteq> Some apiobject_type.Untyped \<longrightarrow> Q {} s)\<rbrace>"
   shows "\<lbrace>P\<rbrace> createObject ty ptr us dev\<lbrace>\<lambda>m s. Q (untypedRange m) s\<rbrace>"
   including no_pre
+  supply toAPIType_Some[simp del]
```
corlewis, line 7504: Any thoughts on this needing to be removed from the simpset in quite a few proofs? Is it a sign that something isn't set-up quite right, or is it more to do with these being old proofs?

proof/refine/X64/Detype_R.thy
```diff
@@ -961,6 +961,10 @@ lemma detype_ready_queues_relation:
   apply (fastforce simp add: detype_def)
   done
 
+lemma insert_id_ksArchState_upd:
+  "s = s\<lparr>ksArchState := id (ksArchState s)\<rparr>"
+  by simp
```
corlewis: This doesn't seem to be used.

corlewis:
> This PR has the same diff-reducing tech as the TcbAcc_R PR, copied instructions:

This is still very much appreciated, but just commenting that unfortunately it hasn't been quite as effective in this PR. I think it's due to more content being moved around inside of files, which confuses the diff and makes it harder to see where the real changes are. Even more unfortunately, I don't have any suggestions for how to improve this.

proof/refine/Retype_R.thy
Xaphiosis: very well spotted, will do

proof/refine/Retype_R.thy
Xaphiosis: done for next commit

proof/refine/Retype_R.thy
Xaphiosis: done for next commit

proof/refine/Retype_R.thy
Xaphiosis: done in next commit

proof/refine/Retype_R.thy
Xaphiosis: in terms of arch-split concepts, yours is the better approach. I can prove valid_obj_makeObject_tcb in the locale and make that simp as well, hopefully it won't break much.

proof/refine/Retype_R.thy
Xaphiosis: I'll look for candidate spots

proof/refine/Retype_R.thy
Xaphiosis: can't do too much about it now since I'd have to plunge into AInvs, and the tcb_bits/word_bits/cte_size_bits unfolding sins are stronger there, hence the FIXME

proof/refine/Retype_R.thy
Xaphiosis: ~~I'll try a cleanup pass today~~ actually I have no idea, will go with the FIXME

proof/refine/AARCH64/ArchRetype_R.thy
Xaphiosis: well spotted, this can be made generic it seems. done on AARCH64 so far, will have to update the rest

proof/refine/AARCH64/ArchRetype_R.thy
Xaphiosis: will do

proof/refine/AARCH64/ArchRetype_R.thy
Xaphiosis: This FIXME was left in to attract attention during the PR and hopefully have some fresher minds throw some advice my way. As you can see, I've only done it for this one lemma, despite the shape of it occurring a fair few times. I guess I can leave the comment to remember why this one is different?

proof/refine/AARCH64/ArchRetype_R.thy
Xaphiosis: This is an ongoing issue... the problem with the non-arch-split stuff that comes after the current PR is that it uses `interpretation Arch .` ... which dumps Arch stuff into the generic context, and therefore making it look like we need to [simp del] the parts that are a generic simp. However, I will only know the situation when I get to try arch-split the lemmas where that [simp del] is happening, and having the simp there will hopefully yield statements that are no longer arch-specific (which the full definition of `APIType_map2` definitely is).
The gamble feels worthwhile for arch-split, even if it does require a few simp dels at the moment... but I could still turn out to be wrong about it in the end, once the potentially generic proof parts are assessed. Thoughts?

proof/refine/ARM/ArchRetype_R.thy
Xaphiosis: will do

proof/refine/ARM/ArchRetype_R.thy
Xaphiosis: will do

proof/refine/ARM/ArchRetype_R.thy
Xaphiosis: cheers, will do

proof/refine/ARM/ArchRetype_R.thy
Xaphiosis: will try to clear these, since this is a Retype_R PR, now's the time to fix them

proof/crefine/AARCH64/Retype_C.thy
Xaphiosis: I'm not sure. It's a similar situation to `APIType_map2_generic` where it'll split off the generic part from the arch part. At the moment it's a bit messy, and so this is the quickest fix without digging much into old proofs. I think saying `simp del` to this one is a bit like saying "I give up on arch-split in this lemma". What do you think?

proof/refine/X64/Detype_R.thy
Xaphiosis: thanks, debugging leftover

Xaphiosis:
Thank you for the very thorough review! Especially thanks for reading the commit messages and finding both a typo and a squash-derp.

> This is still very much appreciated, but just commenting that unfortunately it hasn't been quite as effective in this PR. I think it's due to more content being moved around inside of files, which confuses the diff and makes it harder to see where the real changes are. Even more unfortunately, I don't have any suggestions for how to improve this.

The problem with Retype_R is that it was harder than it appears. Things kept breaking, had to go back and forth so many times between arches to finally get it to converge, and even back out of some things I thought would be generic. So the diff could only be so nice in the end.

What are your thoughts on the oddball interface at the end of ArchCSpace_R and the unfortunate case of `createNewCaps_obj_at''`?

proof/refine/AARCH64/ArchRetype_R.thy
lsf37: Leaving in the split and the comment as an example for posterity sounds good. It shows that we thought about it and why we didn't do more of it. That's valuable info.

I would remove the FIXME tag, because we don't actually want to fix anything here.

proof/refine/AARCH64/ArchRetype_R.thy
lsf37: Happy to gamble on this one. If your gut feeling says this might be worthwile, let's try it. We can always go back and tweak things if we really care strongly about it.

proof/refine/Retype_R.thy
lsf37: Let's keep the FIXME, good reminder that we might want to do something about this.

There was a proposal from the Microkit side to print object sizes to a json file during build, because they are trying to run the same code on different kernel configs. In the distant future we could read object sizes from that same file, and then we'd really want all of these to not be unfolded all the time.

But not for now.

proof/refine/Retype_R.thy
lsf37: Agree with Corey, something cleanup-worthy is going on here. FIXME appreciated.

proof/crefine/AARCH64/Retype_C.thy
lsf37: Is this to avoid warnings or do things break if we have this `[simp]`?

lsf37: (If warnings only, we could add a comment)

lsf37 (approved):
I have nothing to add to Corey's comments (very thorough).

Also agree that 13k lines removed is not bad at all. It was a lot of work, but it's Retype, that's some of less pleasant stuff around.

proof/refine/Retype_R.thy
Xaphiosis: moved to StateRelationLemmas

proof/refine/AARCH64/ArchRetype_R.thy
Xaphiosis: will drop the FIXME tag then

proof/crefine/AARCH64/Retype_C.thy
Xaphiosis: It breaks this proof, because it does not expecting toAPIType to unfold in any way - but `toAPIType_Some` does unfold the generic part in the case when we know it's Some; which forces dealing with generic and arch-specific differently. The proof wants to deal with everything as if it's arch-specific, and then unfold at the leafs.

proof/refine/AARCH64/ArchRetype_R.thy
corlewis: You have by far the most experience here so happy to go with your feeling as well. Would it make sense to mark any of them with a FIXME though, to make sure we look at it again once they're arch-split?

proof/refine/ARM/ArchRetype_R.thy
corlewis: Minor enough to not bother with if you don't feel like it, but it looks like you can still remove the `result:` from the `shows` line on those architectures.

corlewis:
> What are your thoughts on the oddball interface at the end of ArchCSpace_R and the unfortunate case of `createNewCaps_obj_at''`?

No particular insights or suggestions for improvements. It's a bit hard to get a good feeling for them just from the diff but they seem reasonable and are nicely commented. I guess the requalifies at the end of ArchCSpace_R could be moved to a generic file to reduce duplication (probably to Retype_R), but then they would be separated from their definitions and that could just cause more confusion.

proof/refine/ARM/ArchRetype_R.thy
Xaphiosis: ah, missed that. thanks. will do

proof/refine/AARCH64/ArchRetype_R.thy
Xaphiosis: Well, for the ones in ArchRetype_R I think the only uses are where I wanted to keep the old proof and couldn't arch-split, so not much point adding them there. In the later theories, there's only one or two uses, and it'll become apparent when I try to arch-split those whether to keep them or not.

Issue #905, opened 2025-06-25
Make run_tests print a warning when skip_proofs is set in .isabelle/etc


Xaphiosis:
In rare circumstances in our Isabelle fork, the "Skip proofs" checkbox's value can be saved to `~/.isabelle/etc/settings`. This leads to all commandline builds resulting in proofs being skipped, and it's very hard to spot this situation (interactive proofs failing, but commandline passing, or one machine's tests passing while another's fail).

We should check for this in run_tests and print a big red warning when running with skip proofs on.

`isabelle options -g skip_proofs` should produce `true` or `false` accordingly

PR #941, opened 2025-12-05
run_tests: --l4v-arch= and skip_proofs check

Xaphiosis:
run_tests: l4v-arch option deprecates l4v-arches

It's annoying when jumping between --l4v-arch-all and --l4v-arches when
re-using command history.
Introduce an --l4v-arch= option makes things more consistent, i.e.:
    run_tests --l4v-arch=X64,RISCV64
    run_tests --l4v-arch-all
Keep the old --l4v-arches as an alias for --l4v-arch for compatibility:
    run_tests --l4v-arches=X64,RISCV64

---

regression: warn when skip_proofs is set

When using the skip_proofs checkbox in our version of the JEdit Isabelle
plugin and closing/killing JEdit, sometimes the skip_proofs setting is
persistently set in ~/.isabelle/etc/preferences

Running run_tests after that makes it look like everything worked when
no proofs were checked.

Caveat: Isabelle internally respects ISABELLE_BUILD_OPTIONS, which can
contain skip_proofs, but `isabelle options -g skip_proofs` will ignore
that and only print the etc/preferences setting.

Closes https://github.com/seL4/l4v/issues/905

### DIFF
--- a/misc/regression/run_tests.py
+++ b/misc/regression/run_tests.py
@@ -43,6 +43,7 @@ ANSI_RESET = "\033[0m"
 ANSI_RED = "\033[31;1m"
 ANSI_GREEN = "\033[32m"
 ANSI_YELLOW = "\033[33m"
+ANSI_BRIGHT_YELLOW = "\033[0;93m"
 ANSI_WHITE = "\033[37m"
 ANSI_BOLD = "\033[1m"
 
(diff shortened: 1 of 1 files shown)


misc/regression/run_tests.py
```diff
@@ -67,6 +68,25 @@ def which(filename):
             return candidate
     return None
 
+
+# Warn user if skip_proofs is set (Isabelle won't prove anything)
+
+def warn_isabelle_skip_proofs():
+    # Check for skip_proofs and warn user
+    isabelle = "./isabelle/bin/isabelle"
```
lsf37: I'm never clear on relative to what these paths are. It does seem to work fine if I invoke `run_tests` from another directory, so everything is fine here. Should we add a comment that we are expecting this to work?

Or should we assume stronger python-fu than mine?

Xaphiosis: It is indeed the case that I tested it like you did, but I'm not sure why exactly it works even from another directory. Assumed that run_tests will set its current directory relative to its location, but will investigate.

Xaphiosis: Given this, I should try to use unqualified "isabelle" I think:
```
# Fetch directory this script is stored in.
DIR=os.path.dirname(os.path.realpath(__file__))

# Add repo version of Isabelle to our path.
os.environ["PATH"] += os.pathsep + DIR
```

Xaphiosis: Nope, run_tests does not pass DIR on, but it does set cwd to DIR, so that's why my original works. I'll update the comment.

Xaphiosis: tweaked, take a look

lsf37: Excellent, thank you. That comment will reduce future pain.

corlewis:
> Introduce an --l4v-arch= option makes things more consistent, i.e.:

Commit message suggestion, I think the above should be either `Introducing ...` or `... to make things ...`

corlewis (approved):
Nice improvement!

PR #942, opened 2025-12-05

corlewis:
* refine: use ARM version of Corres.thy
* refine: arch-split Bits_R
* refine: arch-split EmptyFail
* refine: unify Invocations_R
* refine+crefine: arch-split Machine_R
* spec+proof: use semi-lazy FPU switching

Tested with seL4 commit a117712c19.

### DIFF
--- a/proof/invariant-abstract/Tcb_AI.thy
+++ b/proof/invariant-abstract/Tcb_AI.thy
@@ -38,10 +38,18 @@ locale Tcb_AI_1 =
        \<lbrace>\<lambda>b s. ex_nonz_cap_to a s\<rbrace>"
   assumes arch_post_set_flags[wp]:
   "\<And>t flags. arch_post_set_flags t flags \<lbrace>invs :: 'state_ext state \<Rightarrow> _\<rbrace>"
+  assumes arch_post_set_flags_cur_thread[wp]:
+  "\<And>t d. arch_post_set_flags t d \<lbrace>\<lambda>s::'state_ext state. P (cur_thread s)\<rbrace>"
+  assumes arch_post_set_flags_st_tcb_at[wp]:
+  "\<And>t d Q P t'. arch_post_set_flags t d \<lbrace>\<lambda>s::'state_ext state. Q (st_tcb_at P t' s)\<rbrace>"
   assumes arch_prepare_set_domain_invs[wp]:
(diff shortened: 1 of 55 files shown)


proof/invariant-abstract/Tcb_AI.thy
```diff
@@ -38,10 +38,18 @@ locale Tcb_AI_1 =
...
+  assumes arch_post_set_flags_st_tcb_at[wp]:
+  "\<And>t d Q P t'. arch_post_set_flags t d \<lbrace>\<lambda>s::'state_ext state. Q (st_tcb_at P t' s)\<rbrace>"
   assumes arch_prepare_set_domain_invs[wp]:
   "\<And>t d. arch_prepare_set_domain t d \<lbrace>invs :: 'state_ext state \<Rightarrow> _\<rbrace>"
   assumes arch_prepare_set_domain_typ_at[wp]:
   "\<And>t d P T p. arch_prepare_set_domain t d \<lbrace>\<lambda>s::'state_ext state. P (typ_at T p s)\<rbrace>"
+  assumes arch_prepare_set_domain_cur_thread[wp]:
+  "\<And>t d. arch_prepare_set_domain t d \<lbrace>\<lambda>s::'state_ext state. P (cur_thread s)\<rbrace>"
+  assumes arch_prepare_set_domain_st_tcb_at[wp]:
+  "\<And>t d Q P t'. arch_prepare_set_domain t d \<lbrace>\<lambda>s::'state_ext state. Q (st_tcb_at P t' s)\<rbrace>"
```
Xaphiosis, line 52: Small confusion: these interface changes did not merge properly? Or is this an extra improvement post-merge?

corlewis: Extra improvement post-merge, these lemmas weren't needed in master.

corlewis:
> Silly idea: You know how you suggested virtual commits for the arch-split PRs to show a work delta that ultimately gets squashed away? Given that you're doing a merge, would it make any sense at all to create a virtual commit with the merge conflicts in it, then show the delta to your actual merge, and finally the cleanup commit? Then the conflict resolution can get squashed into the merge commit (if git permits that). It might well be the case that the conflict resolution is just a ton of noise and so the idea isn't practical, but what are your thoughts?

Interesting idea, I'll give it a go next merge. My guess is that the conflict resolution will be too noisy to be very useful but it's definitely worth a try.

PR #945, opened 2025-12-14
AArch64: support SMC calls in verified kernel

lsf37:
This adds support for SMC (Secure Monitor Calls), see also [RFC-9](https://sel4.github.io/rfcs/implemented/0090-smc-cap.html).

Currently, this forces `KernelAllowSMCCalls` to be ON for AArch64 builds. The plan is to remove all SMC ifdefs in the C code apart from the one that provides the initial SMC master cap at boot. (Or, if people are happy with that, just have SMC always on, because the initialiser will not pass on the SMC master cap if it is not explicitly requested).

The main proof change on the generic interface side is the introduction of `arch_cap_badge`/`arch_capBadge`, which was necessary because the proofs previously assumed that arch caps never have a badge that is managed with the standard badge mechanisms.

Since we know nothing about the actual content of SMC operations, the best we can do is assume that the `doSMC_mop` will only modify machine state the kernel does not manage, and leave kernel-visible memory unchanged. This appears to be a true assumption in how SMC is used so far in seL4 (power management and device access). Other uses will have to make their own assessment.

There is a companion seL4 PR (https://github.com/seL4/seL4/pull/1572) with a minor code refactor that aligns the C more to the usual decode/invoke split and argument passing. Not behaviour change is intended.

During the proofs it came up that the `capclass` definition was placed in a generic ASpec file, but has architecture-specific content (e.g. IOPorts). It turns out that we only make use of 3 actual cap classes: physical, reply, and the rest. Since this still works in the generic setting, instead of arch splitting the type, I have adjusted it into these 3 classes in a separate commit at the end. The initial ASpec commit for SMC caps introduces an SMCClass which is later removed in that consolidation.

Test with: https://github.com/seL4/seL4/pull/1572

[RFC-9]: https://sel4.atlassian.net/browse/RFC-9?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/spec/abstract/CSpace_A.thy
+++ b/spec/abstract/CSpace_A.thy
@@ -842,7 +842,12 @@ definition
 
 section "Cap classification used to define invariants"
 
+text \<open>@{term PhysicalClass} is for caps to objects that consume physical
+memory. This includes anything that can be retyped from Untyped caps.
+@{term ReplyClass} is specifically for reply caps. @{term OtherCapClass} for
+anything else.\<close>
+
 datatype capclass =
-  PhysicalClass | ReplyClass "obj_ref" | IRQClass | ASIDMasterClass | NullClass | DomainClass | IOPortClass | SMCClass
+  PhysicalClass | ReplyClass "obj_ref" | OtherCapClass
 
 end
--- a/proof/invariant-abstract/AARCH64/ArchInvariants_AI.thy
+++ b/proof/invariant-abstract/AARCH64/ArchInvariants_AI.thy
@@ -277,12 +277,12 @@ lemmas is_nondevice_page_cap_simps[simp] =
 
 primrec acap_class :: "arch_cap \<Rightarrow> capclass" where
   "acap_class (ASIDPoolCap _ _)     = PhysicalClass"
-| "acap_class (ASIDControlCap)      = ASIDMasterClass"
-| "acap_class (SGISignalCap _ _)    = IRQClass"
+| "acap_class (ASIDControlCap)      = OtherCapClass"
+| "acap_class (SGISignalCap _ _)    = OtherCapClass"
 | "acap_class (FrameCap _ _ _ _ _)  = PhysicalClass"
 | "acap_class (PageTableCap _ _ _)  = PhysicalClass"
 | "acap_class (VCPUCap _)           = PhysicalClass"
-| "acap_class (SMCCap _)            = SMCClass"
+| "acap_class (SMCCap _)            = OtherCapClass"
 
 definition valid_ipc_buffer_cap_arch :: "arch_cap \<Rightarrow> machine_word \<Rightarrow> bool" where
   [simp]: "valid_ipc_buffer_cap_arch ac bufptr \<equiv>
--- a/proof/invariant-abstract/ARM/ArchInvariants_AI.thy
+++ b/proof/invariant-abstract/ARM/ArchInvariants_AI.thy
@@ -224,11 +224,11 @@ primrec
   acap_class :: "arch_cap \<Rightarrow> capclass"
 where
   "acap_class (ASIDPoolCap x y)      = PhysicalClass"
-| "acap_class (ASIDControlCap)       = ASIDMasterClass"
+| "acap_class (ASIDControlCap)       = OtherCapClass"
 | "acap_class (PageCap dev x y sz z) = PhysicalClass"
 | "acap_class (PageTableCap x y)     = PhysicalClass"
 | "acap_class (PageDirectoryCap x y) = PhysicalClass"
-| "acap_class (SGISignalCap _ _ )    = IRQClass"
+| "acap_class (SGISignalCap _ _ )    = OtherCapClass"
 
 definition
   valid_ipc_buffer_cap_arch :: "arch_cap \<Rightarrow> word32 \<Rightarrow> bool"
--- a/proof/invariant-abstract/ARM_HYP/ArchInvariants_AI.thy
+++ b/proof/invariant-abstract/ARM_HYP/ArchInvariants_AI.thy
@@ -235,12 +235,12 @@ primrec
   acap_class :: "arch_cap \<Rightarrow> capclass"
 where
   "acap_class (ASIDPoolCap _ _)      = PhysicalClass"
-| "acap_class (ASIDControlCap)       = ASIDMasterClass"
+| "acap_class (ASIDControlCap)       = OtherCapClass"
 | "acap_class (PageCap _ _ _ _ _)    = PhysicalClass"
 | "acap_class (PageTableCap _ _)     = PhysicalClass"
 | "acap_class (PageDirectoryCap _ _) = PhysicalClass"
 | "acap_class (VCPUCap _)            = PhysicalClass"
-| "acap_class (SGISignalCap _ _)     = IRQClass"
+| "acap_class (SGISignalCap _ _)     = OtherCapClass"
 
 definition
   valid_ipc_buffer_cap_arch :: "arch_cap \<Rightarrow> word32 \<Rightarrow> bool"
--- a/proof/invariant-abstract/Invariants_AI.thy
+++ b/proof/invariant-abstract/Invariants_AI.thy
@@ -400,18 +400,18 @@ definition
 primrec
   cap_class :: "cap \<Rightarrow> capclass"
 where
-  "cap_class (cap.NullCap)                          = NullClass"
-| "cap_class (cap.UntypedCap dev p n f)             = PhysicalClass"
-| "cap_class (cap.EndpointCap ref badge r)          = PhysicalClass"
-| "cap_class (cap.NotificationCap ref badge r)     = PhysicalClass"
-| "cap_class (cap.CNodeCap ref n bits)              = PhysicalClass"
-| "cap_class (cap.ThreadCap ref)                    = PhysicalClass"
-| "cap_class (cap.DomainCap)                        = DomainClass"
-| "cap_class (cap.Zombie r b n)                     = PhysicalClass"
-| "cap_class (cap.IRQControlCap)                    = IRQClass"
-| "cap_class (cap.IRQHandlerCap irq)                = IRQClass"
-| "cap_class (cap.ReplyCap tcb m rights)            = ReplyClass tcb"
-| "cap_class (cap.ArchObjectCap cap)                = acap_class cap"
+  "cap_class (NullCap)                     = OtherCapClass"
+| "cap_class (UntypedCap dev p n f)        = PhysicalClass"
+| "cap_class (EndpointCap ref badge r)     = PhysicalClass"
+| "cap_class (NotificationCap ref badge r) = PhysicalClass"
+| "cap_class (CNodeCap ref n bits)         = PhysicalClass"
+| "cap_class (ThreadCap ref)               = PhysicalClass"
+| "cap_class (DomainCap)                   = OtherCapClass"
+| "cap_class (Zombie r b n)                = PhysicalClass"
+| "cap_class (IRQControlCap)               = OtherCapClass"
+| "cap_class (IRQHandlerCap irq)           = OtherCapClass"
+| "cap_class (ReplyCap tcb m rights)       = ReplyClass tcb"
+| "cap_class (ArchObjectCap cap)           = acap_class cap"
 
 
 definition
--- a/proof/invariant-abstract/RISCV64/ArchInvariants_AI.thy
+++ b/proof/invariant-abstract/RISCV64/ArchInvariants_AI.thy
@@ -255,7 +255,7 @@ lemmas is_nondevice_page_cap_simps[simp] =
 
 primrec acap_class :: "arch_cap \<Rightarrow> capclass" where
   "acap_class (ASIDPoolCap _ _)     = PhysicalClass"
-| "acap_class (ASIDControlCap)      = ASIDMasterClass"
+| "acap_class (ASIDControlCap)      = OtherCapClass"
 | "acap_class (FrameCap _ _ _ _ _)  = PhysicalClass"
 | "acap_class (PageTableCap _ _)    = PhysicalClass"
 
--- a/proof/invariant-abstract/X64/ArchInvariants_AI.thy
+++ b/proof/invariant-abstract/X64/ArchInvariants_AI.thy
@@ -300,14 +300,14 @@ primrec
   acap_class :: "arch_cap \<Rightarrow> capclass"
 where
   "acap_class (ASIDPoolCap x y)       = PhysicalClass"
-| "acap_class (ASIDControlCap)        = ASIDMasterClass"
+| "acap_class (ASIDControlCap)        = OtherCapClass"
 | "acap_class (PageCap d x y m sz z)  = PhysicalClass"
 | "acap_class (PageTableCap x y)      = PhysicalClass"
 | "acap_class (PageDirectoryCap x y)  = PhysicalClass"
 | "acap_class (PDPointerTableCap x y) = PhysicalClass"
 | "acap_class (PML4Cap x y)           = PhysicalClass"
-| "acap_class (IOPortCap x y)         = IOPortClass"
-| "acap_class (IOPortControlCap)      = IOPortClass"
+| "acap_class (IOPortCap x y)         = OtherCapClass"
+| "acap_class (IOPortControlCap)      = OtherCapClass"
 
 definition
   valid_ipc_buffer_cap_arch :: "arch_cap \<Rightarrow> machine_word \<Rightarrow> bool"
--- a/proof/refine/AARCH64/ArchInvsDefs_H.thy
+++ b/proof/refine/AARCH64/ArchInvsDefs_H.thy
@@ -199,12 +199,12 @@ primrec
   acapClass :: "arch_capability \<Rightarrow> capclass"
 where
   "acapClass (ASIDPoolCap _ _)    = PhysicalClass"
-| "acapClass ASIDControlCap       = ASIDMasterClass"
+| "acapClass ASIDControlCap       = OtherCapClass"
 | "acapClass (FrameCap _ _ _ _ _) = PhysicalClass"
 | "acapClass (PageTableCap _ _ _) = PhysicalClass"
 | "acapClass (VCPUCap _)          = PhysicalClass"
-| "acapClass (SGISignalCap _ _)   = IRQClass"
-| "acapClass (SMCCap _)           = SMCClass"
+| "acapClass (SGISignalCap _ _)   = OtherCapClass"
+| "acapClass (SMCCap _)           = OtherCapClass"
 
 locale_abbrev isArchSMCCap :: "capability \<Rightarrow> bool" where
   "isArchSMCCap \<equiv> arch_cap'_pred is_SMCCap"
--- a/proof/refine/ARM/ArchInvsDefs_H.thy
+++ b/proof/refine/ARM/ArchInvsDefs_H.thy
@@ -165,11 +165,11 @@ primrec valid_arch_obj' :: "arch_kernel_object \<Rightarrow> bool" where
 
 primrec acapClass :: "arch_capability \<Rightarrow> capclass" where
   "acapClass (ASIDPoolCap x y)      = PhysicalClass"
-| "acapClass ASIDControlCap         = ASIDMasterClass"
+| "acapClass ASIDControlCap         = OtherCapClass"
 | "acapClass (PageCap d x y sz z)   = PhysicalClass"
 | "acapClass (PageTableCap x y)     = PhysicalClass"
 | "acapClass (PageDirectoryCap x y) = PhysicalClass"
-| "acapClass (SGISignalCap _ _ )    = IRQClass"
+| "acapClass (SGISignalCap _ _ )    = OtherCapClass"
 
 definition valid_arch_badges :: "capability \<Rightarrow> capability \<Rightarrow> mdbnode \<Rightarrow> bool" where
   "valid_arch_badges cap cap' node' \<equiv>
--- a/proof/refine/ARM_HYP/ArchInvsDefs_H.thy
+++ b/proof/refine/ARM_HYP/ArchInvsDefs_H.thy
@@ -201,12 +201,12 @@ primrec valid_arch_obj' :: "arch_kernel_object \<Rightarrow> bool" where
 
 primrec acapClass :: "arch_capability \<Rightarrow> capclass" where
   "acapClass (ASIDPoolCap x y)      = PhysicalClass"
-| "acapClass ASIDControlCap         = ASIDMasterClass"
+| "acapClass ASIDControlCap         = OtherCapClass"
 | "acapClass (PageCap d x y sz z)   = PhysicalClass"
 | "acapClass (PageTableCap x y)     = PhysicalClass"
 | "acapClass (PageDirectoryCap x y) = PhysicalClass"
 | "acapClass (VCPUCap x)            = PhysicalClass"
-| "acapClass (SGISignalCap _ _ )    = IRQClass"
+| "acapClass (SGISignalCap _ _ )    = OtherCapClass"
 
 definition valid_arch_badges :: "capability \<Rightarrow> capability \<Rightarrow> mdbnode \<Rightarrow> bool" where
   "valid_arch_badges cap cap' node' \<equiv>
--- a/proof/refine/Invariants_H.thy
+++ b/proof/refine/Invariants_H.thy
@@ -482,18 +482,18 @@ where
 primrec
   capClass :: "capability \<Rightarrow> capclass"
 where
-  "capClass (NullCap)                          = NullClass"
-| "capClass (DomainCap)                        = DomainClass"
-| "capClass (UntypedCap d p n f)               = PhysicalClass"
-| "capClass (EndpointCap ref badge s r g gr)   = PhysicalClass"
-| "capClass (NotificationCap ref badge s r)    = PhysicalClass"
-| "capClass (CNodeCap ref bits g gs)           = PhysicalClass"
-| "capClass (ThreadCap ref)                    = PhysicalClass"
-| "capClass (Zombie r b n)                     = PhysicalClass"
-| "capClass (IRQControlCap)                    = IRQClass"
-| "capClass (IRQHandlerCap irq)                = IRQClass"
-| "capClass (ReplyCap tcb m g)                 = ReplyClass tcb"
-| "capClass (ArchObjectCap cap)                = acapClass cap"
+  "capClass (NullCap)                        = OtherCapClass"
+| "capClass (DomainCap)                      = OtherCapClass"
+| "capClass (UntypedCap d p n f)             = PhysicalClass"
+| "capClass (EndpointCap ref badge s r g gr) = PhysicalClass"
+| "capClass (NotificationCap ref badge s r)  = PhysicalClass"
+| "capClass (CNodeCap ref bits g gs)         = PhysicalClass"
+| "capClass (ThreadCap ref)                  = PhysicalClass"
+| "capClass (Zombie r b n)                   = PhysicalClass"
+| "capClass (IRQControlCap)                  = OtherCapClass"
+| "capClass (IRQHandlerCap irq)              = OtherCapClass"
+| "capClass (ReplyCap tcb m g)               = ReplyClass tcb"
+| "capClass (ArchObjectCap cap)              = acapClass cap"
 
 definition
   "capRange cap \<equiv>
--- a/proof/refine/RISCV64/ArchInvsDefs_H.thy
+++ b/proof/refine/RISCV64/ArchInvsDefs_H.thy
@@ -150,7 +150,7 @@ primrec
   acapClass :: "arch_capability \<Rightarrow> capclass"
 where
   "acapClass (ASIDPoolCap _ _)    = PhysicalClass"
-| "acapClass ASIDControlCap       = ASIDMasterClass"
+| "acapClass ASIDControlCap       = OtherCapClass"
 | "acapClass (FrameCap _ _ _ _ _) = PhysicalClass"
 | "acapClass (PageTableCap _ _)   = PhysicalClass"
 
--- a/proof/refine/X64/ArchInvsDefs_H.thy
+++ b/proof/refine/X64/ArchInvsDefs_H.thy
@@ -187,15 +187,15 @@ definition pspace_in_kernel_mappings' :: "kernel_state \<Rightarrow> bool" where
  "pspace_in_kernel_mappings' s \<equiv> \<forall>p \<in> dom (ksPSpace s). p \<in> kernel_mappings"
 
 primrec acapClass :: "arch_capability \<Rightarrow> capclass" where
-  "acapClass (ASIDPoolCap x y)      = PhysicalClass"
-| "acapClass ASIDControlCap         = ASIDMasterClass"
-| "acapClass (PageCap x y t sz d z) = PhysicalClass"
-| "acapClass (PageTableCap x y)     = PhysicalClass"
-| "acapClass (PageDirectoryCap x y) = PhysicalClass"
+  "acapClass (ASIDPoolCap x y)       = PhysicalClass"
+| "acapClass ASIDControlCap          = OtherCapClass"
+| "acapClass (PageCap x y t sz d z)  = PhysicalClass"
+| "acapClass (PageTableCap x y)      = PhysicalClass"
+| "acapClass (PageDirectoryCap x y)  = PhysicalClass"
 | "acapClass (PDPointerTableCap x y) = PhysicalClass"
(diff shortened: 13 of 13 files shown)


proof/crefine/AARCH64/Arch_C.thy
```diff
@@ -4405,6 +4406,298 @@ lemma decodeSGISignalInvocation_ccorres:
...
+lemma word_of_nat_8_n_msgRegisters:
+  "(word_of_nat (if option_to_ptr buf \<noteq> (NULL :: machine_word ptr) then 8 else unat n_msgRegisters)
+     :: machine_word) =
+   (if option_to_ptr buf \<noteq> (NULL :: machine_word ptr) then 8 else scast n_msgRegisters)"
+  by (simp add: n_msgRegisters_def split: if_split)
+
+context begin
+
+(* unroll a counting while loop by one step *)
+private method while_unfold_step for step::machine_word =
```
lsf37: Forgot to clean up that method (it's duplicated below). Will push an update to this file.

lsf37: done.

spec/design/skel/AARCH64/ArchRetype_H.thy
```diff
@@ -19,7 +19,7 @@ begin
 
 context Arch begin arch_global_naming (H)
 
-#INCLUDE_HASKELL SEL4/Object/ObjectType/AARCH64.hs CONTEXT AARCH64_H Arch.Types=ArchTypes_H ArchInv=ArchRetypeDecls_H NOT bodies_only
+#INCLUDE_HASKELL SEL4/Object/ObjectType/AARCH64.hs CONTEXT AARCH64_H Arch.Types=ArchTypes_H ArchInv= NOT bodies_only
```
Xaphiosis, line 22: Looking in this commit and reading the commit message, I'm wondering if the other way of going about these would be to do the `global.` vs `Arch.` renaming dance as we do for arch-split (the `context begin global_naming global` one).

Xaphiosis: The reason  I bring it up is because the commit message has a "it works by coincidence now" vibes, but I'm not 100% on what's going on.

Xaphiosis:
commit nitpicks/suggestions while I warm up:
* `Haskell is using` -> `Haskell uses`
* `It appears that "||" has been correctly` -> should that be `\|`? Confused
* `This introduces` -> `Introduces`
* `This commit introduces` -> `Introduce`
* `orphanage` -> `Orphanage`
* `arm+arm-hyp+riscv_x64` should have a `+` there, and `arch_cap_bade` should be badge

Also, is the plan to squash any of these at the end, or leave as is?

spec/machine/AARCH64/MachineOps.thy
```diff
@@ -485,5 +485,32 @@ consts' setVSpaceRoot_impl :: "paddr \<Rightarrow> machine_word \<Rightarrow> un
...
 
+
+subsection "Secure Monitor Calls (SMC)"
+
+definition numSMCRegs :: nat where
+  "numSMCRegs \<equiv> 8"
+
+(* Tempting to use a list, but encoding the length requirement would make it ugly, and numSCMRegs
+   is only going to change when the SMC spec changes. *)
+type_synonym 'a eight_tuple = "'a \<times> 'a \<times>'a \<times>'a \<times>'a \<times>'a \<times>'a \<times>'a"
```
Xaphiosis: might want to add spaces after the `×` consistently, I was staring at this for a moment thinking what `x'a` is

tools/haskell-translator/lhs_pars.py
```diff
@@ -701,7 +701,8 @@ def typename_transform(line, header, d):
         [oldtype] = line.split()
     except:
         if header not in known_type_assignments:
-            warning('type assignment with parameters not supported %s' % d.body, filename)
+            warning('Type assignment with parameters not supported %s\nTranslated string was "%s"'
```
Xaphiosis: is there any benefit to having an indent for the second line to indicate they go together?

spec/haskell/src/SEL4/Object/ObjectType/AARCH64.hs
```diff
@@ -74,6 +83,10 @@ isArchMDBParentOf (ArchObjectCap (SGISignalCap _ _)) (ArchObjectCap (SGISignalCa
 isArchMDBParentOf _ _ _ = True
 
 updateCapData :: Bool -> Word -> ArchCapability -> Capability
+updateCapData preserve newBadge (SMCCap badge) =
+    if not preserve &&  badge == 0
```
Xaphiosis: two spaces?

spec/haskell/src/SEL4/Object/ObjectType/AARCH64.hs
```diff
@@ -263,17 +278,35 @@ decodeInvocation label args capIndex slot cap extraCaps =
...
+       SMCCap {} -> decodeSMCInvocation label args cap
        -- TODO AARCH64 SMMU: SID/CB control caps, SID/CB caps
        _ -> decodeARMMMUInvocation label args capIndex slot cap extraCaps
 
+decodeSMCInvocation :: Word -> [Word] -> ArchCapability -> KernelF SyscallError ArchInv.Invocation
+decodeSMCInvocation label args (SMCCap badge) = do
+    unless (invocationType label == ArchInvocationLabel ARMSMCCall) $ throw IllegalOperation
+    unless (length args >= numSMCRegs) $ throw TruncatedMessage
+    smcFuncID <- return $ args !! 0
+    when (badge /= 0 && badge /= smcFuncID) $ throw IllegalOperation
```
Xaphiosis, line 290: can you refresh my memory: what is the significance of badge 0?

proof/invariant-abstract/AARCH64/ArchTcb_AI.thy
```diff
@@ -42,12 +42,67 @@ lemma same_object_obj_refs[Tcb_AI_assms]:
...
+    is_cnode_cap cap \<or>
+    is_arch_cap cap \<and> (is_pt_cap cap \<longrightarrow> cap_asid cap \<noteq> None) \<and>
+      (\<not> arch_cap_fun_lift is_SMCCap False cap)"
+
+crunch arch_get_sanitise_register_info, arch_post_modify_registers
+  for tcb_at[wp, Tcb_AI_assms]: "tcb_at a"
+crunch arch_get_sanitise_register_info, arch_post_modify_registers
+  for invs[wp, Tcb_AI_assms]: "invs"
+crunch arch_get_sanitise_register_info, arch_post_modify_registers
+  for ex_nonz_cap_to[wp, Tcb_AI_assms]: "ex_nonz_cap_to a"
```
Xaphiosis: combine crunches?

Xaphiosis: (I know it's moved, but while we're here)

proof/invariant-abstract/Tcb_AI.thy
```diff
@@ -371,6 +373,9 @@ lemma zombies_final_helperE:
   apply (fastforce simp: cte_wp_at_caps_of_state)
   done
 
+context Tcb_AI_1
+begin
```
Xaphiosis: For these ones (that don't define a locale), I'd prefer the `begin` to be on the same line, or at least that's how I've been doing it in arch-split of Refine...

proof/refine/CSpace1_R.thy
```diff
@@ -39,6 +39,8 @@ arch_requalify_consts
   arch_mdb_preservation
   is_derived'
   capASID cap_asid_base' cap_vptr' (* needed to define weak_derived' *)
+  arch_capMasterCap
+  arch_capBadge
```
Xaphiosis, line 43: interesting that this was not needed before

proof/refine/CSpace1_R.thy
```diff
@@ -2407,12 +2411,25 @@ lemma zbits_map_eq[simp]:
   "(zbits_map zbits = zbits_map zbits') = (zbits = zbits')"
   by (simp add: zbits_map_def split: option.split sum.split)
 
+context CSpace1_R
+begin
```
Xaphiosis: at least in Refine, on same line please

proof/refine/AARCH64/CNodeInv_R.thy
```diff
@@ -5736,16 +5765,17 @@ lemma make_zombie_invs':
...
+     apply (erule_tac x=sl in allE)
+     apply simp
+     apply (erule_tac x=p' in allE)
+     apply (rule conjI; solves \<open>clarsimp simp: mdb_next_unfold\<close>)
+    apply (fastforce simp: mdb_next_unfold)
    apply (erule_tac x=p in allE)
    apply simp
    apply (erule_tac x=p' in allE)
-   apply (solves \<open>clarsimp simp: mdb_next_unfold\<close>)
+   apply (rule conjI; solves \<open>clarsimp simp: mdb_next_unfold\<close>)
```
Xaphiosis: do you really need the conjI here? surprising

proof/refine/AARCH64/CNodeInv_R.thy
```diff
@@ -4165,6 +4176,24 @@ lemma SGI_scap_neq_cap:
...
   by (auto simp: weak_derived'_def isCap_simps)
 
+lemma isSMC_src:
+  "isArchSMCCap scap = isArchSMCCap src_cap"
+  using src_derived
+  by (fastforce simp: isCap_simps weak_derived'_def)
+
+lemma isSMC_dest:
+  "isArchSMCCap dcap = isArchSMCCap dest_cap"
+  using dest_derived by (fastforce simp: isCap_simps weak_derived'_def)
```
Xaphiosis: one of these is not like the others (should we be wrapping the `by` on the same line, or saving space? either way, these four should be the same)

proof/refine/AARCH64/CNodeInv_R.thy
```diff
@@ -73,22 +73,28 @@ where
...
     apply simp
    apply (clarsimp simp: word_size word_size_def cnode_padding_bits_def cnode_guard_size_bits_def
                          cteRightsBits_def cteGuardBits_def)
-  apply (clarsimp simp: AARCH64_H.updateCapData_def isCap_simps split del: if_split)
+  apply (clarsimp simp: isCap_simps acap_relation_arch_update_cap_data_NullCap split del: if_split)
   done
 
 (* Sometimes I need something about the state. This is neater (IMHO) and req *)
-lemma whenE_throwError_corres':
+lemma whenE_throwError_corres': (* FIXME: move *)
```
Xaphiosis, line 97: would be good to move it indeed, maybe after arch-split, we should have most of the "oh by the way heres's a generic corres rule" isolated to the generic theories

proof/refine/AARCH64/ArchCSpace_I.thy
```diff
@@ -386,20 +401,25 @@ qed
...
 
 definition vs_cap_ref_arch' :: "arch_capability \<Rightarrow> (asid \<times> vspace_ref) option" where
   "vs_cap_ref_arch' acap \<equiv>
      case acap of
        ASIDPoolCap _ asid \<Rightarrow> Some (asid, 0)
      | ASIDControlCap \<Rightarrow> None
      | FrameCap _ _ _ _ m \<Rightarrow> m
-     | PageTableCap _ _ m \<Rightarrow> m"
+     | PageTableCap _ _ m \<Rightarrow> m
+     | _ \<Rightarrow> None"
```
Xaphiosis, line 417: hmm, I think we had this discussion before, whether we want to have the benefit of complete pattern match without dummy; in this case I think I'd like having the SMC cap here and no default

proof/refine/AARCH64/ArchCSpace1_R.thy
```diff
@@ -228,51 +248,45 @@ qed
 
 lemmas setCTE_typ_ats[wp] = typ_at_lifts[OF setCTE_typ_at']
 
+context
+begin
```
Xaphiosis: I have a vague feeling I'd want a comment here to indicate the purpose of the context, or maybe at the end, but it's hard to say what would help exactly. I guess if it's this short, we can leave it unadorned, but I never know what to do with these things.

proof/crefine/AARCH64/Machine_C.thy
```diff
@@ -19,6 +19,17 @@ instance virq_C :: array_inner_packed
...
+context kernel
+begin
+
+(* for doSMC_mop_ccorres *)
+definition smc_args_to_C :: "machine_word eight_tuple \<Rightarrow> smc_args_t__C" where
+  "smc_args_to_C \<equiv>
+     \<lambda>(x0, x1, x2, x3, x4, x5, x6, x7).
+       smc_args_t__C (ARRAY i. [x0, x1, x2, x3, x4, x5, x6, x7] ! i)"
+
+end
```
Xaphiosis, line 31: why `context kernel` for this one, instead of dumping it into kernel_m below? I don't remember how long it takes to enter/leave kernel context, but I don't think it's fast; might be a little faster to say `definition (in kernel)` if you really want it to be in kernel here

proof/crefine/AARCH64/CSpace_C.thy
```diff
@@ -3330,12 +3375,19 @@ lemma updateCapData_spec:
...
-   apply (simp add: isArchCap_tag_def2)
-   apply (simp add: AARCH64_H.updateCapData_def)
-
+   apply (clarsimp simp: isArchCap_tag_def2 AARCH64_H.updateCapData_def)
+   apply (case_tac acap;
+          simp add: cap_get_tag_isCap_unfolded_H_cap2[unfolded cap_tag_defs, simplified])
+    apply (rename_tac p pt_t m)
+    apply (case_tac pt_t;
+           simp add: cap_get_tag_isCap_unfolded_H_cap2[unfolded cap_tag_defs, simplified])
+   (* smc_cap *)
```
Xaphiosis: should match case with other comments (`CNodeCap` vs `smc_cap`)

proof/crefine/AARCH64/Arch_C.thy
```diff
@@ -4405,6 +4406,300 @@ lemma decodeSGISignalInvocation_ccorres:
...
+                apply (rule ccorres_expand_while_iff_Seq[THEN iffD1])
+                apply (rule ccorres_cond_false)
+                (* This recreates the return value of the last setMR that we discarded above.
+                   Note that msg_len also occurs as parameter of the schematic C guard and the
+                   clarsimp below will substitute this term there as well. This wreaks havoc later
+                   with the rule application for setMessageInfo_ccorres where the term will be
+                   unified back into just msg_len which is the variable we wanted to get rid of.
+                   Rewriting with word_of_nat_8_n_msgRegisters makes the schematic parameter
+                   different from the message length parameter in the program and prevents that
+                   unwanted unification *)
```
Xaphiosis, line 4532: I am getting the feeling I would have changed the C somehow to prevent this loop unrolling and related mess, especially since the number is fixed in the SMC spec to 8. Out of curiosity, was this not possible?

proof/crefine/AARCH64/Arch_C.thy
```diff
@@ -4405,6 +4406,300 @@ lemma decodeSGISignalInvocation_ccorres:
...
+                 apply wpsimp
+                apply clarsimp
+               apply (wpsimp simp: setMR_def length_of_msgRegisters)
+              apply wpsimp
+             apply wpsimp
+            apply wpsimp
+           apply wpsimp
+          apply wpsimp
+         apply wpsimp
+        apply wpsimp
```
Xaphiosis: worth using a `((wpsimp ...)+)[n]` here?

proof/crefine/AARCH64/Arch_C.thy
```diff
@@ -4405,6 +4406,300 @@ lemma decodeSGISignalInvocation_ccorres:
...
+  (* 8 times array update = smc_args_to_C *)
+  apply (rule drop_imp, rule allI)
+  apply (rename_tac smc_arg, case_tac smc_arg)
+  apply (clarsimp simp: smc_args_to_C_def intro!: array_ext)
+  apply (rename_tac i, case_tac i, simp)
+  apply (repeat 7 \<open>rename_tac i', case_tac i', simp add: eval_nat_numeral\<close>)
+  apply simp
+  done
+
+end (* of private method context *)
```
Xaphiosis, line 4694: yes, like this! (related to my other comment about a method context, forgot where it was again)

spec/abstract/CSpace_A.thy
```diff
@@ -843,6 +843,11 @@ definition
 section "Cap classification used to define invariants"
 
 datatype capclass =
-  PhysicalClass | ReplyClass "obj_ref" | IRQClass | ASIDMasterClass | NullClass | DomainClass | IOPortClass | SMCClass
+  PhysicalClass | ReplyClass "obj_ref" | OtherCapClass
```
Xaphiosis: I like the idea, and it does clean things up, but should there be a comment that says something about what goes into OtherCapClass or not?

Xaphiosis: oh, actually, there is text right below here... maybe it should go above the datatype definition instead?

spec/abstract/CSpace_A.thy
```diff
@@ -843,6 +843,11 @@ definition
...
+
+text \<open>@{const PhysicalClass} is for caps to objects that consume physical
+memory. This includes anything that can be retyped from Untyped caps.
+@{const ReplyClass} is specifically for reply caps. @{const OtherCapClass} for
+anything else.\<close>
```
Xaphiosis, line 848: might also want to indent per style guide

proof/refine/AARCH64/ArchCSpace1_R.thy
Xaphiosis: After side-channel discussion: this should probably go on the same line, we're going with favouring vertical compression over noise when there's nothing interesting in the context we're entering. Same argumentation as `if (....) {` in C, but it's more verbose here.

proof/invariant-abstract/ARM_HYP/ArchCSpaceInvPre_AI.thy
```diff
@@ -66,6 +66,11 @@ definition
 
 lemmas is_ap_cap_simps [simp] = is_ap_cap_def [split_simps cap.split arch_cap.split]
 
+definition arch_cap_badge :: "arch_cap \<Rightarrow> machine_word option" where
```
Xaphiosis, line 70: for some of these kinds of things, we've added comments about how feature <blah> is either not present or trivial on this architecture; thoughts?

proof/refine/X64/Tcb_R.thy
```diff
@@ -1556,7 +1548,9 @@ proof -
             apply (clarsimp cong: imp_cong conj_cong)
             apply (rule_tac Q'="\<lambda>_. ?T2_pre' and (\<lambda>s. valid_option_prio p_auth)"
                          in hoare_strengthen_postE_R[simplified validE_R_def, rotated])
-             apply (case_tac g'; clarsimp simp: isCap_simps ; clarsimp elim: invs_valid_objs' cong:imp_cong)
+             apply (case_tac g';
+                    clarsimp simp: isCap_simps arch_capBadge_def;
+                    clarsimp elim: invs_valid_objs' cong:imp_cong)
```
Xaphiosis: if you're cleaning up, `cong:imp_cong` : space (same on other arches)

proof/invariant-abstract/ARM_HYP/ArchCSpaceInvPre_AI.thy
Xaphiosis: Oh. There is one in ArchCSpace_I, so that might be good enough

Xaphiosis (approved):
I have the usual pile of questions and nitpicks, but looks good from my side. That C loop thing though, wow. Wouldn't have come up with that easily.

proof/refine/AARCH64/ArchCSpace1_R.thy
lsf37: > I have a vague feeling I'd want a comment here to indicate the purpose of the context, or maybe at the end, but it's hard to say what would help exactly. I guess if it's this short, we can leave it unadorned, but I never know what to do with these things.

Not sure a comment would make much sense here. The comment would say that it is for the `private method` constructs below, but the `context/begin` is immediately followed by `private method` in the next line, so it would just be saying what the next line is saying.

proof/crefine/AARCH64/Machine_C.thy
lsf37: I think I had to do that, because it is also used in theories that are not using `kernel_m`, but just `kernel` (I was surprised to find out these existed, but there is a whole bunch of theories that don't need any machine interface assumptions, and they are not using `kernel_m`).

proof/refine/AARCH64/CNodeInv_R.thy
lsf37: I think I copy/pasted these, but yes, I agree.

spec/abstract/CSpace_A.thy
lsf37: Yes, it probably should. I just updated the exiting comment there, but it'd be better above than below.

proof/crefine/AARCH64/Arch_C.thy
lsf37: You mean unroll the loop in C directly? That could be done, but it is kinda ugly and unidiomatic and doesn't really help much, because the issue is mostly on the abstract/Haskell side where `setMR` is supposed to return how many registers where actually transferred. We'd still be ignoring the first 7 of these return values and then match up the 8th one. 

We could also unroll fully in Haskell/abstract, but it's a lot repetition. This way, the repetition is confined to the proof. The main annoyance is that we can't avoid the repetition in the proof without proving new rules that do not really warrant the complexity they would need. Hence the proof methods and comments.

Xaphiosis: If you want to confine it to the proof, would unrolling via monadic_rewrite or something like that work?

proof/refine/AARCH64/ArchCSpace1_R.thy
Xaphiosis: Yup, I can buy that. Adding a comment to the `end` like you did in the other situation I applauded would work too.

lsf37: I thought the end comment was exactly for this one. At leat that was the intention :-)

spec/design/skel/AARCH64/ArchRetype_H.thy
lsf37: It worked purely by coincidence before (because it was basically unused) and works more reliably now, but it does not yet fully work without restrictions. It would make sense to look at this a bit more globally separately, because you are right, it would be nice to have a proper documented general solution for this.

proof/refine/CSpace1_R.thy
lsf37: `arch_capBadge` is new and the interface lemma about it talks about `arch_capMasterCap`. We don't have that many separate generic properties about arch caps, so I guess this is the first time we make an interface that manages to go by master cap instead of something more specific.

proof/refine/AARCH64/CNodeInv_R.thy
lsf37: It does for speed. `fastforce` and the other more aggressive tactics go down a wrong path and take forever.

proof/refine/AARCH64/ArchCSpace_I.thy
lsf37: We did have the discussion for Haskell, but I don't think we did for Isabelle.

For Isabelle, I definitely don't agree (for Haskell I partially I agree, depending on context). In this case, it is specifically an Isabelle function for extracting info from VM caps. You don't get a warning for missing cases, you just get `undefined`, which is worse (I actually hit that).  I would only be adding noise if I added cases for other caps. I would in fact argue that we should remove the ASIDControlCap case, because it obscures what is going on.

If we had written it with the default case in the first place, I would not have had to backtrack from the failed proof and edit this definition. If I had been adding a VM cap instead, I might still have missed this case, and have had to backtrack from a failed proof, but that is no worse than what we have now. So the default case is strictly better for this.

I don't want to go as far as saying that default cases are always better for Isabelle definitions, but for these extractor functions, I'm pretty sure.

tools/haskell-translator/lhs_pars.py
lsf37: Good idea, done.

spec/haskell/src/SEL4/Object/ObjectType/AARCH64.hs
lsf37: do you mean indent by 2 instead of 4? Haskell uses 4, no? (at least everything else in this file does)

spec/haskell/src/SEL4/Object/ObjectType/AARCH64.hs
lsf37: It's like for endpoints and notifications: badge 0 means it is the unbadged "master" cap which has more authority than badged caps. For SMC caps that additional authority means it can invoke any SMC function.

proof/invariant-abstract/AARCH64/ArchTcb_AI.thy
lsf37: Done.

proof/refine/AARCH64/ArchCSpace1_R.thy
lsf37: Hm, while I was searching for this, there does seem to be a relatively strong upstream preference of not putting `context` and `begin` on the same line (95 hits with line break, 22 without in isabelle/**). I'll still do it, because in l4v we do prefer the other way: 38 vs 16 so far for not breaking the line (> 100 if you count the "context begin interpretation" incantation, but that is special).

lsf37: > I thought the end comment was exactly for this one. At leat that was the intention :-)

It was indeed not, because there are 2, and I had done it only for one.

proof/crefine/AARCH64/Arch_C.thy
lsf37: rewriting was my first idea for this as well, but it doesn't work, because you have to unroll both sides simultaneously, and we don't have good infrastructure for doing that on the C side. I was halfway through trying to do something with ccores_rewrite, when I realised that we have all of the necessary infrastructure available to do this directly in the proof. It is a bit ugly, but the alternatives are a lot bigger and not prettier.

proof/crefine/AARCH64/Arch_C.thy
lsf37: Hm, 4 is roughly my threshold for these. I find the double parenthesis thing not so nice to read and it is a bit unwieldy in maintenance, because you tend to have to unroll them when something breaks. If there were 7 of these, sure. 4 I think are borderline, but still fine.

I'm happy to be overruled on this one if people have strong opinions about it, though. I'm more explaining how I got there than saying that this is how it should be.

spec/abstract/CSpace_A.thy
lsf37: Moved.

spec/abstract/CSpace_A.thy
lsf37: The rest of this file doesn't indent on text, and since it's using it consistently I think I'd prefer leaving it until we update the entire thing at some point.

proof/invariant-abstract/ARM_HYP/ArchCSpaceInvPre_AI.thy
lsf37: This is the first occurrence, so it would actually fit better here. Will add one.

lsf37:
> commit nitpicks/suggestions while I warm up:
> 
> * `declare constants for SMC caps` doesn't declare the consts, only flags them for import?
> * `Haskell is using` -> `Haskell uses`
> * `It appears that "||" has been correctly` -> should that be `\|`? Confused
> * `Similarly to AInvs in the previous commit, this commit introduces` -> `..., introduce` ?
> * `This introduces` -> `Introduces`
> * `This commit introduces` -> `Introduce`
> * `orphanage` -> `Orphanage`
> * `arm+arm-hyp+riscv_x64` should have a `+` there, and `arch_cap_bade` should be badge

Thanks, all done.

> Also, is the plan to squash any of these at the end, or leave as is? 

I was planning to leave them as is, but happy to squash more if there are suggestions. We could squash all of the aarch64 proofs into one commit, for instance. I don't have a big preference either way.

> The other thing is that the order is a bit strange. I was expecting a story, like aspec, then haskell / design infrastructure, then aarch64 proofs, and then mop-up operations for other arches. The later parts seem in order, but at the start it's a bit confusing to follow since I'm reviewing commit-by-commit.

True, the current version did not have a good flow. I have now tried to group it as:

- general tool and import fixes
- machine
- Haskell
- tool fixes that depend on Haskell update, caseconvs
- design
- aspec
- aarch64 proofs
- other proofs

(ASpec doesn't build without the Haskell update, hence Haskell first)

spec/abstract/CSpace_A.thy
lsf37: And I know now why the comment was below before and not above :-): it uses `@{const ..}` to refer to the constructors and if the comment is above the declaration, those are not defined. Have replace const with term, which is less checking, but better text flow.

spec/design/skel/AARCH64/ArchRetype_H.thy
Xaphiosis: Well, if you do end up understanding it, please do document it so we can join you

proof/refine/AARCH64/CNodeInv_R.thy
Xaphiosis: ah. worth a comment?

proof/refine/AARCH64/ArchCSpace_I.thy
Xaphiosis: I would have thought that if you don't have a default case and you miss a case, you get a warning at the definition of this function, which allows you to fix this function. You don't get any signs in other proofs where the function is used, that's true, it'll spit out undefined. So my argument was along the lines of getting a warning that the definition of this function has become wrong, because a new case has appeared and isn't being handled. Whereas with the default case, no more warning when that happens, and hope that's what you wanted for whatever case shows up in the future. That's why I thought the warning would be useful.

If you're confident that in future situations where a new case pops up it'll almost certainly be the None case and the spirit of the caches swings that way, well, sure, but I still don't feel that it's generally a future-safe approach.

Then again if you don't look at the spec in Isabelle, and so don't see any warnings, then you will indeed be backtracking from a proof. Sigh. I'm willing to concede this if you at least understand what I'm trying to get at :)

spec/haskell/src/SEL4/Object/ObjectType/AARCH64.hs
Xaphiosis: on my monitor there are two spaces between `&&` and `badge`

proof/crefine/AARCH64/Arch_C.thy
Xaphiosis: Would a `apply (solves wpsimp)+` feel better for you? Otherwise I can live with a threshold of 4.

proof/refine/AARCH64/ArchCSpace_I.thy
lsf37: > I would have thought that if you don't have a default case and you miss a case, you get a warning at the definition of this function, which allows you to fix this function.

You don't get a warning for `case` statements Isabelle (only for primrec/fun), though. If there was a warning, I'd agree that enumerating all cases would often be more useful. I might still disagree about this specific case, but not in general :-)

> Then again if you don't look at the spec in Isabelle, and so don't see any warnings, 

Even if you do look at the spec, there is no warning for these.

> then you will indeed be backtracking from a proof. Sigh. I'm willing to concede this if you at least understand what I'm trying to get at :)

I think I do :-)

proof/crefine/AARCH64/Arch_C.thy
lsf37: I'll give that shot.

spec/haskell/src/SEL4/Object/ObjectType/AARCH64.hs
lsf37: Yup, there are indeed. Thank you.

proof/refine/AARCH64/ArchCSpace_I.thy
Xaphiosis: Well, turns out not only AIs hallucinate. I guess primrec/function will give warnings, but indeed case won't. Sigh. Point conceded, I'll try remember in future.

proof/crefine/AARCH64/Arch_C.thy
lsf37: That does work.

PR #946, opened 2025-12-17
Fix task enabling FPU for itself

corlewis:
Restore the FPU immediately when setting the TCB flag to enabled if it is the current thread. Otherwise the FPU would stay disabled, as switch_to_thread would not be called.

Test with: https://github.com/seL4/seL4/pull/1574

### DIFF
--- a/proof/crefine/AARCH64/Tcb_C.thy
+++ b/proof/crefine/AARCH64/Tcb_C.thy
@@ -4505,6 +4505,35 @@ lemma decodeSetTLSBase_ccorres:
   apply (auto simp: unat_eq_0 le_max_word_ucast_id)+
   done
 
+lemma lazyFPURestore_ccorres:
+  "ccorres dc xfdc
+     (no_0_obj' and tcb_at' t)
+     (\<lbrace>\<acute>thread = tcb_ptr_to_ctcb_ptr t\<rbrace>) hs
+     (lazyFpuRestore t) (Call lazyFPURestore_'proc)"
+  apply (cinit lift: thread_')
+   apply (rule ccorres_pre_threadGet, rename_tac flags)
+   apply (rule ccorres_move_c_guard_tcb)
+   apply (rule ccorres_if_lhs)
+    apply (rule ccorres_cond_true)
+    apply (ctac add: disableFpu_ccorres)
+   apply (rule ccorres_cond_false)
+   apply (ctac add: nativeThreadUsingFPU_ccorres[where thread=t])
+     apply (rule ccorres_cond[where R=\<top>])
+       apply (clarsimp simp: to_bool_neq_0 split: if_split)
+      apply (ctac add: enableFpu_ccorres)
+     apply (ctac add: switchLocalFpuOwner_ccorres)
+    apply wpsimp
+   apply (vcg exspec=nativeThreadUsingFPU_modifies)
+  apply (clarsimp simp: isFlagSet_def typ_heap_simps' word_bw_comms ctcb_relation_def
+                        option_to_ctcb_ptr_def)
+  done
+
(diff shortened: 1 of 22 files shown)


proof/crefine/AARCH64/Tcb_C.thy
```diff
@@ -4505,6 +4505,35 @@ lemma decodeSetTLSBase_ccorres:
...
+  apply (clarsimp simp: isFlagSet_def typ_heap_simps' word_bw_comms ctcb_relation_def
+                        option_to_ctcb_ptr_def)
+  done
+
+crunch lazyFpuRestore
+  for st_tcb_at'[wp]: "\<lambda>s. P (st_tcb_at' Q p s)"
+  and sch_act_wf[wp]: "\<lambda>s. sch_act_wf (ksSchedulerAction s) s"
+  and aligned'[wp]: "pspace_aligned'"
+  and distinct'[wp]: "pspace_distinct'"
+
```
Indanz, line 4536: This bit is exactly the same as for X86, I suppose there is no easy way of avoiding the duplication?

Xaphiosis: We wish, but no. For the C refinement we don't really have an arch-splitting story, and even for higher abstraction levels we still don't have a "arches with FPU" wrapper (AARCH64 and X64 are the only ones with FPU currently).

proof/invariant-abstract/AARCH64/ArchFPU_AI.thy
```diff
@@ -234,6 +225,21 @@ lemma option_update_thread_no_etcb_change_cur_fpu_in_cur_domain:
...
 
+lemma arch_post_set_flags_in_cur_domain[wp]:
+  "\<lbrace>cur_fpu_in_cur_domain and (\<lambda>s. in_cur_domain (cur_thread s) s)\<rbrace>
+   arch_post_set_flags t flags
+   \<lbrace>\<lambda>_. cur_fpu_in_cur_domain\<rbrace>"
+  unfolding arch_post_set_flags_def
+  by wpsimp
+
+lemma set_flags_in_cur_domain[wp]:
+  "set_flags param_a param_b \<lbrace>in_cur_domain t\<rbrace>"
```
Xaphiosis: crunch?

Xaphiosis: if not, maybe better var names `t flags` as above

Xaphiosis (approved):
One minor nitpick, looks good otherwise!

lsf37 (approved):
All good from my side. 

Also good to see that this is contained to just functional correctness. For a second I was afraid this would have infoflow/access control impact.

proof/invariant-abstract/AARCH64/ArchFPU_AI.thy
corlewis: Thanks for spotting, definitely should have better var names.

The proof here isn't following the usual crunch pattern, due to unfolding the property definition and not wanting to recurse through any other functions. I could probably turn it into a crunch with the right ignores and wp and simp rules, but I don't think it would even save a line and it would probably make it harder to see what is happening.

proof/crefine/AARCH64/Tcb_C.thy
corlewis: We don't have time to do it now but there actually is a plausible way for us to make this generic. The FPU code is self-contained enough that we could wrap it in `if (config_set(CONFIG_HAVE_FPU))` instead of ifdefs. If we're careful with the proofs this would then allow us to treat FPU being supported as a generic feature in the same way the code does, instead of tying it to architectures like the specifications and proofs currently do.

PR #948, opened 2025-12-18
arch-split Schedule_R

Xaphiosis:
🦆🦆🦆 This PR has the same diff-reducing tech as the TcbAcc_R PR, copied instructions:

In real life, I did the usual addition of Schedule_R, hierarchy update, arch-split and then update. However, since the arch-split part first splits AARCH64, then takes the AARCH64 version of ArchSchedule_R and only fixes what broke, we ended up with a small(er) diff there (again).

- `[squash] PR: copy AARCH64 version into Schedule_R` - same as last time: stop Schedule_R looking like a wall of green (no need to review)
- `aarch64 refine: arch-split Schedule_R` - this is the main bulk of the arch-split, dealing with AARCH64
- `[squash] copy AARCH64 ArchSchedule_R to other arches` - wipes the *body* (not the header!) of other arches' ArchSchedule_R with a copy of the AARCH64 version with `AARCH64` appropriately substituted (as @corlewis requested last time) - (mostly pointless to review)
- `[squash] refine: arch-split Schedule_R (other arches) - result of fixing up the ArchSchedule_R of other architectures - shows diff to AARCH64, small for this PR
- `refine+crefine: update for ... arch-split`

Reminder: DO NOT MERGE squashed version if it still says `aarch64 refine: arch-split`

Stats: 30 files changed, 3108 insertions(+), 10756 deletions(-)
Total: 7648 lines removed

### DIFF
--- a/proof/refine/AARCH64/IpcCancel_R.thy
+++ b/proof/refine/AARCH64/IpcCancel_R.thy
@@ -1853,7 +1853,7 @@ lemma rescheduleRequired_all_invs_but_ct_not_inQ:
   done
 
 lemma cancelAllIPC_invs'[wp]:
-  "\<lbrace>invs'\<rbrace> cancelAllIPC ep_ptr \<lbrace>\<lambda>rv. invs'\<rbrace>"
+  "cancelAllIPC ep_ptr \<lbrace>invs'\<rbrace>"
   apply (simp add: cancelAllIPC_def ep'_Idle_case_helper cong del: if_cong)
   apply (rule bind_wp[OF _ stateAssert_sp])
   apply (wp rescheduleRequired_all_invs_but_ct_not_inQ
--- a/proof/crefine/AARCH64/Retype_C.thy
+++ b/proof/crefine/AARCH64/Retype_C.thy
@@ -6262,8 +6262,7 @@ lemma threadSet_domain_ccorres [corres]:
 
 lemma createObject_ccorres:
   notes APITypecapBits_simps[simp] =
-          APIType_capBits_def[split_simps
-          object_type.split apiobject_type.split]
+          APIType_capBits_gen_def[split_simps object_type.split apiobject_type.split]
   shows
     "ccorres ccap_relation ret__struct_cap_C_'
      (createObject_hs_preconds regionBase newType userSize isdev)
@@ -6323,7 +6322,7 @@ proof -
                           word_le_nat_alt split:
                           apiobject_type.splits object_type.splits)
    apply (subgoal_tac "\<exists>apiType. newType = APIObjectType apiType")
-    apply (clarsimp simp: APIType_capBits_gen_def)
+    apply clarsimp
     apply (rule ccorres_guard_imp)
       apply (rule_tac apiType=apiType in ccorres_apiType_split)
 
@@ -6382,8 +6381,7 @@ proof -
          apply (frule invs_pspace_aligned')
          apply (frule invs_pspace_distinct')
          apply (frule invs_sym')
-         apply (simp add: getObjectSize_def objBits_simps word_bits_conv
-                          apiGetObjectSize_def APIType_capBits_gen_def
+         apply (simp add: getObjectSize_def objBits_simps word_bits_conv apiGetObjectSize_def
                           tcbBlockSizeBits_def new_cap_addrs_def projectKO_opt_tcb)
          apply (clarsimp simp: range_cover.aligned region_actually_is_bytes_def)
          apply (frule(1) ghost_assertion_size_logic_no_unat)
@@ -6426,7 +6424,7 @@ proof -
         apply (frule invs_pspace_distinct')
         apply (frule invs_sym')
         apply (auto simp: getObjectSize_def objBits_simps apiGetObjectSize_def
-                          epSizeBits_def word_bits_conv APIType_capBits_gen_def
+                          epSizeBits_def word_bits_conv
                    elim!: is_aligned_no_wrap'
                   intro!: range_cover_simpleI)[1]
 
@@ -6462,10 +6460,9 @@ proof -
        apply (frule invs_pspace_aligned')
        apply (frule invs_pspace_distinct')
        apply (frule invs_sym')
-       apply (auto simp: getObjectSize_def objBits_simps APIType_capBits_gen_def
-                   apiGetObjectSize_def
-                   ntfnSizeBits_def word_bits_conv
-                elim!: is_aligned_no_wrap'  intro!: range_cover_simpleI)[1]
+       apply (auto simp: getObjectSize_def objBits_simps apiGetObjectSize_def
+                         ntfnSizeBits_def word_bits_conv
+                  elim!: is_aligned_no_wrap'  intro!: range_cover_simpleI)[1]
 
       (* CapTable *)
       apply (clarsimp simp: createObject_c_preconds_def)
@@ -6503,18 +6500,17 @@ proof -
        apply (frule invs_pspace_distinct')
        apply (frule invs_sym')
        apply (frule(1) ghost_assertion_size_logic_no_unat)
-       apply (clarsimp simp: getObjectSize_def objBits_simps
-                  apiGetObjectSize_def APIType_capBits_gen_def
-                  cteSizeBits_def word_bits_conv add.commute createObject_c_preconds_def
-                  region_actually_is_bytes_def
-                  invs_valid_objs' invs_urz
-                 elim!: is_aligned_no_wrap'
-                dest: word_of_nat_le  intro!: range_coverI)[1]
+       apply (clarsimp simp: getObjectSize_def objBits_simps apiGetObjectSize_def
+                             cteSizeBits_def word_bits_conv add.commute createObject_c_preconds_def
+                             region_actually_is_bytes_def
+                             invs_valid_objs' invs_urz
+                      elim!: is_aligned_no_wrap'
+                       dest: word_of_nat_le  intro!: range_coverI)[1]
       apply (clarsimp simp: createObject_hs_preconds_def hrs_htd_update isFrameType_def)
       apply (frule range_cover.strong_times_64[folded addr_card_wb], simp+)
       apply (subst h_t_array_valid_retyp, simp+)
        apply (simp add: power_add cte_C_size cteSizeBits_def)
-      apply (clarsimp simp: ccap_relation_def cap_to_H_def cap_cnode_cap_lift  APIType_capBits_gen_def
+      apply (clarsimp simp: ccap_relation_def cap_to_H_def cap_cnode_cap_lift
                             getObjectSize_def apiGetObjectSize_def cteSizeBits_def
                             objBits_simps field_simps is_aligned_power2
                             addr_card_wb is_aligned_weaken[where y=2]
@@ -6531,7 +6527,7 @@ proof -
        apply (clarsimp simp: word_bits_conv)
       apply simp
      apply auto[1]
-    apply (clarsimp simp: createObject_c_preconds_def APIType_capBits_gen_def)
+    apply (clarsimp simp: createObject_c_preconds_def)
     apply (clarsimp simp:nAPIOBjects_object_type_from_H)?
     apply (intro impI conjI, simp_all)[1]
    apply (clarsimp simp: nAPIObjects_def object_type_from_H_def Kernel_C_defs
@@ -7448,17 +7444,17 @@ lemma createObject_sch_act_simple[wp]:
 
 lemma createObject_ct_active'[wp]:
   "\<lbrace>\<lambda>s. ct_active' s \<and> pspace_aligned' s \<and> pspace_distinct' s
-     \<and>  pspace_no_overlap' ptr (APIType_capBits ty us) s
-     \<and>  is_aligned ptr (APIType_capBits ty us) \<and> APIType_capBits ty us < word_bits
-    \<rbrace>createObject ty ptr us dev\<lbrace>\<lambda>r s. ct_active' s \<rbrace>"
- apply (simp add:ct_in_state'_def createObject_def3)
- apply (rule hoare_pre)
- apply wp
- apply wps
- apply (wp createNewCaps_pred_tcb_at')
- apply (intro conjI)
- apply (auto simp:range_cover_full)
- done
+        \<and> pspace_no_overlap' ptr (APIType_capBits ty us) s
+        \<and>  is_aligned ptr (APIType_capBits ty us) \<and> APIType_capBits ty us < word_bits\<rbrace>
+   createObject ty ptr us dev
+   \<lbrace>\<lambda>r s. ct_active' s \<rbrace>"
+  apply (simp add: ct_in_state'_def createObject_def3)
+  apply (rule hoare_pre)
+   apply wp
+   apply wps
+  apply (wp createNewCaps_pred_tcb_at')
+  apply (auto simp: range_cover_full createNewCaps_arch_ko_type_pre_non_arch)
+  done
 
 lemma createObject_notZombie[wp]:
   "\<lbrace>\<top>\<rbrace>createObject ty ptr us dev \<lbrace>\<lambda>r s. \<not> isZombie r\<rbrace>"
--- a/proof/crefine/ARM/Retype_C.thy
+++ b/proof/crefine/ARM/Retype_C.thy
@@ -4973,8 +4973,7 @@ lemma threadSet_domain_ccorres [corres]:
 
 lemma createObject_ccorres:
   notes APITypecapBits_simps[simp] =
-          APIType_capBits_def[split_simps
-          object_type.split apiobject_type.split]
+          APIType_capBits_gen_def[split_simps object_type.split apiobject_type.split]
   shows
     "ccorres ccap_relation ret__struct_cap_C_'
      (createObject_hs_preconds regionBase newType userSize isdev)
@@ -5023,7 +5022,7 @@ proof -
                           word_le_nat_alt
                    split: apiobject_type.splits object_type.splits)
    apply (subgoal_tac "\<exists>apiType. newType = APIObjectType apiType")
-    apply (clarsimp simp: APIType_capBits_gen_def)
+    apply clarsimp
     apply (rule ccorres_guard_imp)
       apply (rule_tac apiType=apiType in ccorres_apiType_split)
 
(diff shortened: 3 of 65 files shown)


proof/refine/Schedule_R.thy
```diff
@@ -2220,10 +1576,23 @@ crunch reschedule_required
...
+     scheduleChooseNewThread
+     \<lbrace>\<lambda>_. invs' \<rbrace>"
+  assumes scheduleChooseNewThread_ct_activatable'[wp]:
+    "\<lbrace> invs' and (\<lambda>s. ksSchedulerAction s = ChooseNewThread) \<rbrace>
+     scheduleChooseNewThread
+     \<lbrace>\<lambda>_. ct_in_state' activatable'\<rbrace>"
+begin
 
-lemma schedule_corres:
+lemma  schedule_corres:
```
lsf37: Looks like an accidental space

proof/crefine/ARM_HYP/ArchMove_C.thy
```diff
@@ -423,7 +423,7 @@ crunch readVCPUReg
   for obj_at'[wp]: "\<lambda>s. P (obj_at' a p s)"
 
 crunch readVCPUReg
-  for pred_tcb_at'[wp]: "\<lambda>s. P (pred_tcb_at' a b p s)"
+  for pred_tcb_at'2[wp]: "\<lambda>s. P (pred_tcb_at' a b p s)"
```
lsf37, line 426: Is the `2` intentional here?

proof/refine/AARCH64/IpcCancel_R.thy
```diff
@@ -1220,7 +1222,7 @@ lemma (in delete_one) suspend_corres:
             apply (simp add: AARCH64.nextInstructionRegister_def AARCH64.faultRegister_def
                              AARCH64_H.nextInstructionRegister_def AARCH64_H.faultRegister_def)
             apply (simp add: AARCH64_H.Register_def)
-            apply (fold dc_def, subst unit_dc_is_eq)
```
lsf37, line 1223: nice to see the need for these disappearing

proof/refine/Schedule_R.thy
```diff
@@ -10,23 +10,22 @@ theory Schedule_R
...
-context begin interpretation Arch . (*FIXME: arch-split*)
-
 declare hoare_weak_lift_imp[wp_split del]
+declare sts_rel_idle[simp]
+declare objBitsT_koTypeOf[simp]
+declare gq_wp[wp]
+declare setQueue_obj_at[wp]
+declare doMachineOp_obj_at[wp]
+declare Cons_eq_tails[simp]
+declare hoare_weak_lift_imp_conj[wp_split del]
```
corlewis, line 21: Not really relevant to this PR but I can't see where either `hoare_weak_lift_imp` or `hoare_weak_lift_imp_conj` are added to `wp_split`. Are these lines needed?

corlewis: Some of the other lemmas also look unnecessary. It might be worth going through and removing any that have already been declared, and moving the others to their sources (or at least adding a FIXME).

proof/refine/Schedule_R.thy
```diff
@@ -788,6 +516,76 @@ lemma tcbSchedAppend_valid_bitmaps[wp]:
...
+  assumes switchToThread_it[wp]:
+    "\<And>t P. switchToThread t \<lbrace>\<lambda>s. P (ksIdleThread s)\<rbrace>"
+  assumes switchToIdleThread_it[wp]:
+    "\<And>P. switchToIdleThread \<lbrace>\<lambda>s. P (ksIdleThread s)\<rbrace>"
+  assumes switchToIdleThread_ct_not_queued_no_cicd':
+    "\<lbrace>invs_no_cicd'\<rbrace> switchToIdleThread \<lbrace>\<lambda>_ s. obj_at' (Not \<circ> tcbQueued) (ksCurThread s) s \<rbrace>"
+  assumes switchToIdleThread_activatable_2[wp]:
+    "\<lbrace>invs_no_cicd'\<rbrace> switchToIdleThread \<lbrace>\<lambda>_. ct_in_state' activatable'\<rbrace>"
+  assumes switchToIdleThread_invs_no_cicd':
+    "\<lbrace>invs_no_cicd'\<rbrace> switchToIdleThread \<lbrace>\<lambda>rv. invs'\<rbrace>"
```
corlewis: Is it worth trying to make any of these generic, with an interface lemma for `Arch.switchToThread` / `Arch.switchToIdleThread` instead?

proof/refine/Schedule_R.thy
```diff
@@ -788,6 +516,76 @@ lemma tcbSchedAppend_valid_bitmaps[wp]:
...
+
+context Schedule_R begin
```
corlewis: Out of curiosity, do we have a style or performance preference between
```
locale Schedule_R =
  ...
context Schedule_R begin
```
and
```
locale Schedule_R =
  ...
begin
```

proof/refine/Schedule_R.thy
```diff
@@ -788,6 +516,76 @@ lemma tcbSchedAppend_valid_bitmaps[wp]:
   apply (clarsimp simp: valid_bitmaps_def)
   done
 
+locale Schedule_R =
+  assumes tcbSchedAppend_state_hyp_refs_of'[wp]:
+    "\<And>P t. tcbSchedAppend t \<lbrace>\<lambda>s. P (state_hyp_refs_of' s)\<rbrace>"
+  assumes tcbSchedDequeue_state_hyp_refs_of'[wp]:
+    "\<And>P t. tcbSchedDequeue t \<lbrace>\<lambda>s. P (state_hyp_refs_of' s)\<rbrace>"
+  assumes arch_switchToThread_corres_interface:
```
corlewis, line 525: For future reference, what convention do you have in mind with this naming scheme of adding interface as a suffix?

proof/refine/Schedule_R.thy
```diff
@@ -1206,43 +1059,15 @@ lemma setThreadState_rct:
...
-   apply (rule word_of_nat_less)
-    apply (fastforce intro: word_of_nat_less simp: wordRadix_def' unat_of_nat word_size)+
-  done
+context Schedule_R begin
 
 lemma bitmapQ_from_bitmap_lookup:
-  "\<lbrakk> ksReadyQueuesL1Bitmap s d \<noteq> 0 ;
-     valid_bitmapQ s ; bitmapQ_no_L1_orphans s
-     \<rbrakk>
+  "\<lbrakk> ksReadyQueuesL1Bitmap s d \<noteq> 0 ; valid_bitmapQ s ; bitmapQ_no_L1_orphans s \<rbrakk>
```
corlewis: I know it's existing but while we're here
```suggestion
  "\<lbrakk> ksReadyQueuesL1Bitmap s d \<noteq> 0; valid_bitmapQ s; bitmapQ_no_L1_orphans s \<rbrakk>
```

corlewis: And same for `bitmapL1_highest_lookup` later on

proof/refine/AARCH64/ArchSchedule_R.thy
```diff
@@ -0,0 +1,745 @@
...
+  (simp: crunch_simps wp: crunch_wps)
+
+crunch switchToThread
+  for cap_to'[wp]: "ex_nonz_cap_to' p"
+  (simp: crunch_simps)
+
+crunch "ThreadDecls_H.switchToThread"
+  for ksCurDomain[wp]: "\<lambda>s. P (ksCurDomain s)"
+
+ (* neater unfold, actual unfold is really ugly *)
```
corlewis:
```suggestion
(* neater unfold, actual unfold is really ugly *)
```

corlewis: And in other arches

proof/refine/ARM_HYP/ArchSchedule_R.thy
```diff
@@ -12,58 +13,60 @@ context Arch begin arch_global_naming
...
-  (simp: crunch_simps wp: crunch_wps)
-
-lemma vcpu_switch_valid_global_arch_objs[wp]:
-  "vcpu_switch v \<lbrace>valid_global_arch_objs\<rbrace>"
-  by (wp valid_global_arch_objs_lift)
-
-crunch set_vm_root
-  for pspace_distinct[wp]: pspace_distinct
-  (simp: crunch_simps)
+(*  and valid_asid_table[wp]: "valid_asid_table at" *) (* FIXME arch-split:  *)
```
corlewis: Is this FIXME resolvable now or are we waiting to see if it's needed in later files?

proof/refine/RISCV64/ArchSchedule_R.thy
```diff
@@ -0,0 +1,427 @@
...
+crunch arch_switch_to_thread, arch_switch_to_idle_thread
+  for pspace_aligned[Schedule_R_assms, wp]: pspace_aligned
+  and pspace_distinct[Schedule_R_assms, wp]: pspace_distinct
+  and ready_queues[Schedule_R_assms, wp]: "\<lambda>s. P (ready_queues s)"
+  and ready_qs_distinct[wp]: ready_qs_distinct
+  (wp: ready_qs_distinct_lift crunch_wps simp: crunch_simps)
+
+lemma arch_switchToThread_corres:
+  "corres dc (valid_arch_state and valid_objs and pspace_aligned and pspace_distinct
+                and valid_vspace_objs and tcb_at t)
```
corlewis:
```suggestion
              and valid_vspace_objs and tcb_at t)
```

proof/refine/X64/ArchSchedule_R.thy
```diff
@@ -0,0 +1,559 @@
...
+              and valid_vspace_objs and pspace_in_kernel_window and valid_cur_fpu and tcb_at t)
+             (no_0_obj')
+             (arch_switch_to_thread t) (Arch.switchToThread t)"
+  by (corres corres: arch_switchToThread_corres; simp add: valid_arch_caps_def)
+
+lemma arch_switchToIdleThread_corres:
+  "corres dc
+        (valid_arch_state and valid_objs and valid_asid_map and unique_table_refs \<circ> caps_of_state
+           and valid_vs_lookup and valid_global_objs and pspace_aligned and pspace_distinct
+           and valid_vspace_objs and valid_idle)
```
corlewis:
```suggestion
lemma arch_switchToIdleThread_corres:
  "corres dc
        (valid_arch_state and valid_objs and valid_asid_map and unique_table_refs \<circ> caps_of_state
         and valid_vs_lookup and valid_global_objs and pspace_aligned and pspace_distinct
         and valid_vspace_objs and valid_idle)
```

corlewis (approved):
Looks very good to me, just a minor suggestions!

proof/crefine/ARM_HYP/ArchMove_C.thy
Xaphiosis: Yes, sadly. Due to arch-splitting introducing a crunch that as part of its action generates a `readVCPUReg_pred_tcb_at'` without the `P`, we need a new name. I looked around this file, and this happened before and was resolved with a `2`, so I stuck with that.

proof/refine/Schedule_R.thy
Xaphiosis: If you look at the arch files, you'll see comments like:
```
(* use superset of arch_switchToThread_corres preconditions across the architectures as interface *)
...
(* use superset of arch_switchToIdleThread_corres preconditions across the architectures as interface *)
```

What this means is that the one tagged `_interface` is not what you want on any architecture, but can be used as a compromise for (a) generic proof(s). In other words, it's artificial. E.g. on X64 you don't need `sym_refs (state_hyp_refs_of s)` for `arch_switchToThread_corres`, but as part of this work I had to clean up fictitious preconditions of `arch_switchToThread_corres` et. al. across architectures already in order to get it to work, so I don't want to add them back to the actual corres rule, hence I add it back to ones I specifically constructed for the task. Does that make any sense? Can you think of a better name scheme?

proof/refine/Schedule_R.thy
Xaphiosis: Honestly not sure what we should be aiming for with this one. I think in this case there was something in between them which I then had to move due to some dependency or other. So this form is definitely more flexible, but it does end up looking a bit silly. Thoughts?

proof/refine/Schedule_R.thy
Xaphiosis: Good question. Not sure, I'll have to investigate, because I've forgotten my thought process when I did this.

proof/refine/Schedule_R.thy
Xaphiosis: They don't generate a warning, because I'm cleaning up as many warnings as I can and wouldn't leave it in if it did nothing. However, current time constraints mean I'm unlikely to go for a hunt for where all these got added.

Xaphiosis: Unless I'm assuming something incorrect here... do `[wp_split del]` even generate warnings?

corlewis: Yeah, I don't think any of our `wp` attributes generate warnings. Probably something we should fix.

proof/crefine/ARM_HYP/ArchMove_C.thy
corlewis: Would it be possible to change that other crunch to have a `P`? Or will that be too much rebuilding for this moment.

proof/refine/Schedule_R.thy
corlewis: Ah, that makes sense. I guess that's a direct example of the increased flexibility so maybe we should lean towards that form. I agree it does look a bit silly though.

proof/refine/Schedule_R.thy
corlewis: That makes sense, and is probably a pattern we could use for quite a few other lemmas in our various interfaces. There's been a few times where there's been some confusion around different preconditions being required across architectures.

proof/crefine/ARM_HYP/ArchMove_C.thy
Xaphiosis: I already tried that, and could not get it stable across the arches. The original crunch doesn't pass with a `P`

lsf37: Alright, let's leave it. Not even sure it needs a comment, just wanted to make sure it wasn't an accident.

proof/refine/Schedule_R.thy
lsf37: Generally I think we should prefer the direct `begin`, because first declaring the locale then doing `context begin` means first leaving then entering the context again. Depending on how large that context is, that can take considerable time. Probably does not in this case, but as a general rule of thumb, I'd avoid the unnecessary leave/enter.

proof/refine/Schedule_R.thy
Xaphiosis: I'll try keep an eye out as well, but I don't remember having encountered something like this before. In this case, I had to figure out what the actual preconditions are, because there were unnecessary ones on old arches that got in the way, but then there was also a superset. I'm guessing there are some around, but without a deep dive like this I missed them.

proof/crefine/ARM_HYP/ArchMove_C.thy
corlewis: Yeah, that doesn't seem worth trying to fix. I guess there must be some intermediate crunches or lemmas that are needed for it to work.

proof/refine/AARCH64/ArchSchedule_R.thy
Xaphiosis: nice catch

proof/refine/ARM_HYP/ArchSchedule_R.thy
Xaphiosis: resolvable, thanks

proof/refine/Schedule_R.thy
Xaphiosis: I dug into this and refreshed my memory. The answer is sadly no, because there is not a wrapped "Arch" part of these. In other words:

```
thm switchToIdleThread_def switchToThread_def
```

yields

```
  Arch.switchToIdleThread ≡ do vcpuSwitch Nothing;
                               setGlobalUserVSpace
                            od
  Arch.switchToThread ?tcb ≡ do tcbobj <- getObject ?tcb;
                                vcpuSwitch (atcbVCPUPtr $ tcbArch tcbobj);
                                setVMRoot ?tcb;
                                lazyFpuRestore ?tcb
                             od
```

Since we can't process `vcpuSwitch` generically as it doesn't exist on non-VCPU arches, and FPU on non-FPU arches I have to handle the next thing up. Normally the solution would be to create an Arch wrapper and shove it before or after, but in this case we'd need one for before and after? No idea. Any further thoughts?

proof/refine/X64/ArchSchedule_R.thy
Xaphiosis: done

proof/refine/RISCV64/ArchSchedule_R.thy
Xaphiosis: done

proof/refine/Schedule_R.thy
Xaphiosis: done

proof/refine/Schedule_R.thy
Xaphiosis: cheers, removed

proof/refine/Schedule_R.thy
Xaphiosis: Now that we have a way of suppressing warnings in Eisbach methods, might be a good idea indeed.

For now, I added `(* FIXME: review which of these are necessary/redundant and move *)` as a reminder.

proof/refine/Schedule_R.thy
corlewis: Ah, looks like we've messed up the naming of these definitions. On ARM I have
```
thm Thread_H.switchToIdleThread_def switchToIdleThread_def
```
yielding
```
  global.switchToIdleThread ≡ do stateAssert ready_qs_runnable [];
                                  thread <- getIdleThread;
                                  Arch.switchToIdleThread;
                                  setCurThread thread
                               od
  Arch.switchToIdleThread ≡ getIdleThread >>= setVMRoot
```
That is, we have a generic and Arch part but the names aren't the way we normally want them.

corlewis: Which I think means that this is the interface that I expected, I just didn't realise it due to the names making it look like it would be for the generic constants.

Xaphiosis: Now you've made me paranoid, I will check again to confirm.

Xaphiosis: Yeah, you're right. I got confused between the Arch and non-Arch versions between the locales. I will try change the interfaces to be on the `Arch.` versions

proof/refine/Schedule_R.thy
Xaphiosis: Removed the unnecessary leave/enter.

proof/refine/Schedule_R.thy
Xaphiosis: Can you check the latest commit? I managed to extract a bit more generic stuff by doing it as you suggested.

I also took a look at tcbSchedAppend et al over hyp refs, but that requires rules related to VCPUs to process, and those can't really be written generically.

corlewis: Looks good to me!

Yeah, making those lemmas  more generic but `state_hyp_refs_of'` being harder is understandable.

PR #949, opened 2025-12-19
monads: make invariant notation bidirectional

lsf37:
Remove the (input)-only restriction from Hoare triple invariant notation and its variants in the trace monad.

We initially had this as input-only, because we were concerned that newcomers may find the output confusing without being able to find its definition (since it was an abbreviation only). Now that abbreviation notation in ctrl-clickable, this concern is no longer relevant and we can benefit from less cluttered output and better input/output alignment.


lsf37:
> For commit message: `notation in ctrl-clickable` -> is?

dammit, I clicked too fast, sorry.

PR #950, opened 2026-01-13
lib: add "declaring" method

lsf37:
Add proof method "declaring" with syntax
    
        declaring (no_warn) <attribs> in <method>
    
where (no_warn) is optional, `<attribs>` is an attribute expression as in
the declare command (without "for"), and `<method>` is a method closure
(single name or cartouche).
    
The method is executed and parsed in the context after attribute
evaluation. If (no_warn) is present, the context for attribute
evaluation is set to invisible, suppressing duplicate rule warnings.
    
The intention is to provide a command that can be used like
"declare", but on the right-hand side of Eisbach method definitions.

### DIFF
--- a/lib/Corres_Method.thy
+++ b/lib/Corres_Method.thy
@@ -54,16 +54,16 @@ method corres_split declares corres_splits = no_name_eta, rule corres_splits
    applied. At that point, there should be no schematic variables in those side condition goals.
    Despite that, we are still careful with simp etc here, in case the user does provide a corres
    rule that generates a schematic in those side condition goals. *)
-method corres_cleanup methods m uses simp simp_del split split_del cong intro =
+method corres_cleanup methods m uses simp simp_del simp_split split_del cong intro =
   #break "corres_cleanup",
   ( m
   | assumption
   | rule refl TrueI
-  | clarsimp simp del: corres_no_simp simp_del simp: simp split: split split del: split_del
+  | clarsimp simp del: corres_no_simp simp_del simp: simp split: simp_split split del: split_del
              cong: cong intro!: intro
   (* enables passing in conjI for terminal goals: *)
   | (rule intro;
-     corres_cleanup m simp: simp simp_del: simp_del split: split split_del: split_del
+     corres_cleanup m simp: simp simp_del: simp_del simp_split: simp_split split_del: split_del
                       cong: cong intro: intro))
 
 (* Apply a single corres rule and attempt to solve non-corres and non-wp side conditions.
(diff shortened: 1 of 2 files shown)


lsf37:
In the current form, `uses` Eisbach parameters do not seem to be visible to the attribute part in `noting`. Which makes this much less useful.

lsf37:
Ok, updated, now fact references work. The `notes` parser can't deal with those fact references, but the parser of the `rule` command can, so I'm using that one.

lsf37:
Now with application to `corres` method.

lib/Eisbach_Tools/Eisbach_Methods.thy
```diff
@@ -28,6 +28,83 @@ method_setup not_visible =
...
+        SIMPLE_METHOD (method_evaluate m (Context_Position.set_visible was_visible ctxt) facts) facts)\<close>
+  \<open>Evaluate method in supplied attribute context. Provides the equivalent of declare command as
+   a method for Eisbach method definitions.\<close>
+
+experiment
+  fixes x y
+  assumes foo: "x = y"
+begin
+
+
```
corlewis:
```suggestion

```

lib/Eisbach_Tools/Eisbach_Methods.thy
```diff
@@ -28,6 +28,82 @@ method_setup not_visible =
...
+  apply (fails \<open>m2 test_del: foo\<close>)
+  (* deletion overrides addition if done this way *)
+  apply (fails \<open>m2 test_del: foo test: foo\<close>)
+  apply m2
+  done
+
+(* Beware of name shadowing for attributes that are not named_theorems, i.e. when you have to use
+   "uses" instead of "declares". The name of the "uses" parameter will shadow the existing
+   attribute name, and "declaring" will be referring to the parameter set instead of the original
+   attribute. *)
```
Xaphiosis, line 91: thank you for documenting this, it is only obvious in retrospect

lib/Eisbach_Tools/Eisbach_Methods.thy
```diff
@@ -28,6 +28,82 @@ method_setup not_visible =
...
+  (* Turn a parser that returns a proof context transformer into a unit context_parser *)
+  fun proof_context_parser (p : (Proof.context -> Proof.context) parser) : unit context_parser =
+    Scan.depend (fn context => p >> (fn f => (Context.map_proof f context, ())))
+
+  fun set_invisible invisible = Context_Position.set_visible (not invisible)
+
+  (* Set context visibility according to no_warn and return whether original context was visible *)
+  val no_warn_parser =
+    (Args.context >> Context_Position.is_visible)
+      --| proof_context_parser (Args.mode "no_warn" >> set_invisible)
```
Xaphiosis: silly question: in ML, do combinators like this get extra indentation, or do they end up flush with `Args.` like we would do with `|>`?

lsf37: Excellent question. The isabelle sources do both, but the majority is flush, so I'll adjust this.

PR #987, opened 2026-04-10
spec+proof: move call to setThreadState within suspend

michaelmcinerney:
Test with https://github.com/seL4/seL4/pull/1642


lsf37 (approved):
Looks like Orphanage sessions still needs updating, but otherwise all good from my side.

lsf37 (approved):

corlewis (approved):

PR #989, opened 2026-04-30
Add cong section to crunch

corlewis:
I also tried adding a `cong_del` section using `Simplifier.del_cong` but it didn't seem to do anything, so I removed it again.


Xaphiosis (approved):
Looks good! Is it worth adding `Allows using cong: <cong_rule> in crunch proofs` or something like that to the commit message?

corlewis:
> Looks good! Is it worth adding `Allows using cong: <cong_rule> in crunch proofs` or something like that to the commit message?

Good suggestion, I'll do that once the tests pass.

lsf37:
> I also tried adding a `cong_del` section using `Simplifier.del_cong` but it didn't seem to do anything, so I removed it again.

Interesting, I didn't even know that `del_cong` existed (although it makes sense). The standard way from the user side for the simplifier is to override with a different `cong` rule, so there is no loss of generality if we don't have a `del`.

lsf37 (approved):
Looks good!

PR #991, opened 2026-04-30
arch-split Finalise_R

Xaphiosis:
🦆🦆🦆 Apart from some locale fails/optimisations (for which I left comments), and an increasing tendency to try avoid interpreting Arch since it's so slow, nothing really novel. Instructions below to remind people of the flow of this.
Extra commit added for cleanup now possible thanks to `crunch cong:`

**Instructions:**

In real life, I did the usual addition of Finalise_R, hierarchy update, arch-split and then update. However, since the arch-split part first splits AARCH64, then takes the AARCH64 version of Finalise_R and only fixes what broke, we ended up with a small(er) diff there (again).

- `aarch64 refine: arch-split `..._R` - this is the main bulk of the arch-split, dealing with AARCH64, as well as updates
- `[squash][fake] refine: arch-split ..._R (other arches) - result of fixing up the ArchSchedule_R of other architectures - shows diff to AARCH64

Stats: 86 files changed, 8107 insertions(+), 18227 deletions(-)
Total: 10120 lines removed

### DIFF
--- a/proof/refine/AARCH64/ArchFinalise_R.thy
+++ b/proof/refine/AARCH64/ArchFinalise_R.thy
@@ -1099,13 +1099,9 @@ lemma archThreadSet_tcbSchedPrevNext[wp]:
   apply auto
   done
 
-context notes if_cong[cong] begin
-
 crunch asUser
   for tcbSchedPrevNext[wp]: "obj_at' (\<lambda>tcb. P (tcbSchedNext tcb) (tcbSchedPrev tcb)) t"
-  (wp: threadGet_wp)
-
-end
+  (wp: threadGet_wp cong: if_cong)
 
 crunch prepareThreadDelete
   for tcbSchedPrevNext[wp]: "obj_at' (\<lambda>tcb. P (tcbSchedNext tcb) (tcbSchedPrev tcb)) t"
@@ -1383,14 +1379,11 @@ crunch dissociateVCPUTCB, unmapPageTable
   (wp: crunch_wps getVCPU_wp getObject_inv hoare_vcg_all_lift hoare_vcg_if_lift3
    simp: loadObject_default_def updateObject_default_def)
 
-context notes if_cong[cong] begin
-
 crunch Arch_finaliseCap
   for nosch[wp]: "\<lambda>s. P (ksSchedulerAction s)"
   (wp: crunch_wps getObject_inv simp: loadObject_default_def updateObject_default_def
-   rule: AARCH64_H.finaliseCap_def)
-
-end
+   rule: AARCH64_H.finaliseCap_def
+   cong: if_cong)
 
 crunch vcpuFinalise, deletingIRQHandler, finaliseCap
   for sch_act_simple[wp]: sch_act_simple
--- a/proof/refine/AARCH64/ArchFinalise_R.thy
+++ b/proof/refine/AARCH64/ArchFinalise_R.thy
@@ -5,11 +5,9 @@
  * SPDX-License-Identifier: GPL-2.0-only
  *)
 
-theory Finalise_R
+theory ArchFinalise_R
 imports
-  ArchIpcCancel_R
-  InterruptAcc_R
-  ArchRetype_R
+  Finalise_R
 begin
 
 context begin interpretation Arch . (*FIXME: arch-split*)
--- a/proof/refine/AARCH64/ArchFinalise_R.thy
+++ b/proof/refine/AARCH64/ArchFinalise_R.thy
@@ -10,502 +10,81 @@ imports
   Finalise_R
 begin
 
-context begin interpretation Arch . (*FIXME: arch-split*)
+context Arch begin arch_global_naming
 
-declare doUnbindNotification_def[simp]
+named_theorems Finalise_R_assms
 
 lemma isArchSGISignalCap_NullCap[simp]:
   "\<not>isArchSGISignalCap NullCap"
   by (simp add: isCap_simps)
 
-text \<open>Properties about empty_slot/emptySlot\<close>
-
-lemma case_Null_If:
-  "(case c of NullCap \<Rightarrow> a | _ \<Rightarrow> b) = (if c = NullCap then a else b)"
-  by (case_tac c, simp_all)
-
-crunch emptySlot
-  for aligned'[wp]: pspace_aligned'
-  and distinct'[wp]: pspace_distinct'
-  and pspace_canonical'[wp]: pspace_canonical'
-  (simp: case_Null_If)
-
-lemma updateCap_cte_wp_at_cases:
-  "\<lbrace>\<lambda>s. (ptr = ptr' \<longrightarrow> cte_wp_at' (P \<circ> cteCap_update (K cap)) ptr' s) \<and> (ptr \<noteq> ptr' \<longrightarrow> cte_wp_at' P ptr' s)\<rbrace>
-     updateCap ptr cap
-   \<lbrace>\<lambda>rv. cte_wp_at' P ptr'\<rbrace>"
-  apply (clarsimp simp: valid_def)
-  apply (drule updateCap_stuff)
-  apply (clarsimp simp: cte_wp_at_ctes_of modify_map_def)
-  done
-
-crunch postCapDeletion, updateTrackedFreeIndex
-  for cte_wp_at'[wp]: "cte_wp_at' P p"
-
-lemma updateFreeIndex_cte_wp_at:
-  "\<lbrace>\<lambda>s. cte_at' p s \<and> P (cte_wp_at' (if p = p' then P'
-      o (cteCap_update (capFreeIndex_update (K idx))) else P') p' s)\<rbrace>
-    updateFreeIndex p idx
-  \<lbrace>\<lambda>rv s. P (cte_wp_at' P' p' s)\<rbrace>"
-  apply (simp add: updateFreeIndex_def updateTrackedFreeIndex_def
-        split del: if_split)
-  apply (rule hoare_pre)
-   apply (wp updateCap_cte_wp_at' getSlotCap_wp)
-  apply (clarsimp simp: cte_wp_at_ctes_of)
-  apply (cases "p' = p", simp_all)
-  apply (case_tac cte, simp)
-  done
-
-lemma emptySlot_cte_wp_cap_other:
-  "\<lbrace>(\<lambda>s. cte_wp_at' (\<lambda>c. P (cteCap c)) p s) and K (p \<noteq> p')\<rbrace>
-  emptySlot p' opt
-  \<lbrace>\<lambda>rv s. cte_wp_at' (\<lambda>c. P (cteCap c)) p s\<rbrace>"
-  apply (rule hoare_gen_asm)
-  apply (simp add: emptySlot_def clearUntypedFreeIndex_def getSlotCap_def)
-  apply (rule hoare_pre)
-   apply (wp updateMDB_weak_cte_wp_at updateCap_cte_wp_at_cases
-             updateFreeIndex_cte_wp_at getCTE_wp' hoare_vcg_all_lift
-              | simp add:  | wpc
-              | wp (once) hoare_drop_imps)+
-  done
+lemma arch_postCapDeletion_ksArchState_lift[Finalise_R_assms]:
+  "\<lbrakk>\<And>s as. P (s\<lparr>ksArchState := as\<rparr>) = P s\<rbrakk> \<Longrightarrow> Arch.postCapDeletion ac \<lbrace>P\<rbrace>"
+  unfolding postCapDeletion_def
+  by wpsimp
 
 lemmas clearUntypedFreeIndex_typ_ats[wp] = typ_at_lifts[OF clearUntypedFreeIndex_typ_at']
 
-crunch postCapDeletion
-  for tcb_at'[wp]: "tcb_at' t"
-crunch emptySlot
-  for ct[wp]: "\<lambda>s. P (ksCurThread s)"
-crunch clearUntypedFreeIndex
-  for cur_tcb'[wp]: "cur_tcb'"
-  (wp: cur_tcb_lift)
+crunch setIRQState
+  for umm[Finalise_R_assms, wp]: "\<lambda>s. P (underlying_memory (ksMachineState s))"
+  (wp: dmo_lift')
 
-crunch emptySlot
-  for ksRQ[wp]: "\<lambda>s. P (ksReadyQueues s)"
-crunch emptySlot
-  for ksRQL1[wp]: "\<lambda>s. P (ksReadyQueuesL1Bitmap s)"
-crunch emptySlot
-  for ksRQL2[wp]: "\<lambda>s. P (ksReadyQueuesL2Bitmap s)"
-crunch postCapDeletion
-  for obj_at'[wp]: "obj_at' P p"
+(* better crunch names for Arch.postCapDeletion *)
+abbreviation (input)
+  "Arch_postCapDeletion \<equiv> Arch.postCapDeletion"
 
-crunch clearUntypedFreeIndex
-  for inQ[wp]: "\<lambda>s. P (obj_at' (inQ d p) t s)"
-crunch clearUntypedFreeIndex
-  for tcbDomain[wp]: "obj_at' (\<lambda>tcb. P (tcbDomain tcb)) t"
-crunch clearUntypedFreeIndex
-  for tcbPriority[wp]: "obj_at' (\<lambda>tcb. P (tcbPriority tcb)) t"
+crunch Arch_postCapDeletion
+  for valid_global_refs[wp]: valid_global_refs'
+  and valid_arch_state'[wp]: valid_arch_state'
+  (rule: AARCH64_H.postCapDeletion_def)
 
-crunch emptySlot
-  for nosch[wp]: "\<lambda>s. P (ksSchedulerAction s)"
-crunch emptySlot
-  for ksCurDomain[wp]: "\<lambda>s. P (ksCurDomain s)"
+lemma arch_postCapDeletion_corres[Finalise_R_assms]:
+  "acap_relation cap cap' \<Longrightarrow> corres dc \<top> \<top> (arch_post_cap_deletion cap) (AARCH64_H.postCapDeletion cap')"
+  by (clarsimp simp: arch_post_cap_deletion_def AARCH64_H.postCapDeletion_def)
 
-lemma emptySlot_sch_act_wf [wp]:
-  "\<lbrace>\<lambda>s. sch_act_wf (ksSchedulerAction s) s\<rbrace>
-  emptySlot sl opt
-  \<lbrace>\<lambda>rv s. sch_act_wf (ksSchedulerAction s) s\<rbrace>"
-  apply (simp add: emptySlot_def case_Null_If)
-  apply (wp sch_act_wf_lift tcb_in_cur_domain'_lift | wpcw | simp)+
-  done
+(* better crunch names for Arch.finaliseCap *)
+abbreviation (input)
+  "Arch_finaliseCap \<equiv> Arch.finaliseCap"
 
-lemma updateCap_valid_objs' [wp]:
-  "\<lbrace>valid_objs' and valid_cap' cap\<rbrace>
-  updateCap ptr cap \<lbrace>\<lambda>r. valid_objs'\<rbrace>"
-  unfolding updateCap_def
-  by (wp setCTE_valid_objs getCTE_wp) (clarsimp dest!: cte_at_cte_wp_atD)
+crunch Arch_finaliseCap, prepareThreadDelete
+  for typ_at'[Finalise_R_assms, wp]: "\<lambda>s. P (typ_at' T p s)"
+  and aligned'[Finalise_R_assms, wp]: "pspace_aligned'"
+  and distinct'[Finalise_R_assms, wp]: "pspace_distinct'"
+  (wp: crunch_wps getObject_inv loadObject_default_inv
+   simp: crunch_simps unless_def o_def
+   ignore_del: setObject
+   rule: AARCH64_H.finaliseCap_def)
 
-lemma updateFreeIndex_valid_objs' [wp]:
-  "\<lbrace>valid_objs'\<rbrace> clearUntypedFreeIndex ptr \<lbrace>\<lambda>r. valid_objs'\<rbrace>"
-  apply (simp add: clearUntypedFreeIndex_def getSlotCap_def)
-  apply (wp getCTE_wp' | wpc | simp add: updateTrackedFreeIndex_def)+
-  done
+crunch prepareThreadDelete, Arch_finaliseCap
+  for it'[Finalise_R_assms, wp]: "\<lambda>s. P (ksIdleThread s)"
+  (wp: hoare_drop_imps
+   simp: crunch_simps updateObject_default_def
+   rule: AARCH64_H.finaliseCap_def)
 
-crunch emptySlot
-  for valid_objs'[wp]: "valid_objs'"
+definition
+  arch_final_matters' :: "arch_capability \<Rightarrow> bool" where
+ "arch_final_matters' acap \<equiv> case acap of
+    FrameCap ref rghts sz d mapdata \<Rightarrow> False
+  | ASIDControlCap \<Rightarrow> False
+  | SMCCap _ \<Rightarrow> False
+  | SGISignalCap _ _ \<Rightarrow> False
+  | _ \<Rightarrow> True"
 
-crunch setInterruptState
-  for state_refs_of'[wp]: "\<lambda>s. P (state_refs_of' s)"
-  (simp: state_refs_of'_pspaceI)
-crunch emptySlot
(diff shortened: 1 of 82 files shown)


Xaphiosis:
Deployed the `crunch cong:` tech as an extra commit, which removed 131 lines. Thanks @corlewis

proof/refine/Finalise_R.thy
```diff
@@ -13,152 +13,12 @@ imports
...
-    apply (erule trancl.intros(1))
-   apply (simp add: is_chunk_def)
-   apply (drule spec, drule mp, erule trancl.intros(1))
-   apply (drule mp, rule trancl_into_rtrancl)
-    apply (erule trancl.intros(1))
-   apply clarsimp
-  apply (clarsimp simp: mdb_next_unfold)
-  apply (case_tac z, simp)
-  done
+declare suspend_unqueued[wp]
```
lsf37: should this merge into the `lemmas [wp]` above?

proof/refine/AARCH64/ArchFinalise_R.thy
```diff
@@ -3356,12 +1262,89 @@ lemma archThreadSet_tcbSchedPrevNext[wp]:
   apply auto
   done
 
+context notes if_cong[cong] begin
```
lsf37: Is this a thing we're doing? I'd have expected

```
context
  notes if_cong[cong]
begin
```

I guess for a single `notes` this makes sense.

proof/refine/ARM/ArchFinalise_R.thy
```diff
@@ -92,7 +92,8 @@ proof -
   show ?thesis
     supply if_cong[cong]
     supply to_slot_eq[simp del]
-    apply (simp add: valid_badges_def3)
+    (* FIXME arch-split; AARCH64 has a different setup due to having unordered_valid_arch_badges *)
+    apply (simp add: valid_badges_def2)
```
lsf37: Is the comment for future attempts of splitting this further? If we've now decided it's not worth it, because it doesn't save anything, we could remove it.

proof/refine/AARCH64/ArchFinalise_R.thy
Xaphiosis: This is all cleaned up anyway in the last crunch cong cleanup commit.

proof/refine/X64/ArchFinalise_R.thy
```diff
@@ -28,6 +25,16 @@ crunch setIRQState
...
 
+lemma setIOPortMask_valid_arch'[wp]:
+  "\<lbrace>valid_arch_state'\<rbrace> setIOPortMask f l b \<lbrace>\<lambda>rv. valid_arch_state'\<rbrace>"
+  by (wpsimp simp: valid_arch_state'_def setIOPortMask_def)
+
+lemma freeIOPortRange_valid_global_refs[wp]:
+  "\<lbrace>valid_global_refs'\<rbrace> freeIOPortRange f l \<lbrace>\<lambda>rv. valid_global_refs'\<rbrace>"
+  apply (clarsimp simp: freeIOPortRange_def setIOPortMask_def)
+  apply wpsimp
+  by (clarsimp simp: valid_global_refs'_def global_refs'_def table_refs'_def bit_simps)
```
lsf37: There's an apply/by style creeping in here. Are all these `by` really necessary?

proof/refine/ARM/ArchFinalise_R.thy
Xaphiosis: It's a response to your comment that said something like "take the AARCH64 version". Since I didn't, I'd want your opinion before I remove these.

proof/refine/X64/ArchFinalise_R.thy
```diff
@@ -37,9 +44,38 @@ crunch Arch_postCapDeletion
...
+
 lemma arch_postCapDeletion_corres[Finalise_R_assms]:
   "acap_relation cap cap' \<Longrightarrow> corres dc \<top> \<top> (arch_post_cap_deletion cap) (X64_H.postCapDeletion cap')"
-  by (clarsimp simp: arch_post_cap_deletion_def X64_H.postCapDeletion_def)
+  apply (clarsimp simp: arch_post_cap_deletion_def X64_H.postCapDeletion_def)
+  apply (rule corres_guard_imp)
+    apply (case_tac cap; clarsimp simp: acap_relation_def)
+    apply (clarsimp simp: free_ioport_range_def freeIOPortRange_def)
+    apply (rule set_ioport_mask_corres)
+   by clarsimp+
```
lsf37: apply/done

proof/refine/X64/ArchFinalise_R.thy
```diff
@@ -37,9 +44,38 @@ crunch Arch_postCapDeletion
...
+  "corres dc \<top> \<top>
+     (set_ioport_mask f l b)
+     (setIOPortMask f l b)"
+  apply (clarsimp simp: set_ioport_mask_def setIOPortMask_def)
+  apply (rule corres_guard_imp)
+    apply (rule corres_split_eqr[OF corres_gets_allocated_io_ports])
+      apply (rule corres_modify)
+      apply (clarsimp simp: state_relation_def arch_state_relation_def foldl_map
+                            foldl_fun_upd)
+     by wpsimp+
```
lsf37: should probably be apply/done, not by

proof/refine/X64/ArchFinalise_R.thy
Xaphiosis: These are copied from the original version on that architecture, where the difference from AARCH64 was significant.

Xaphiosis: Unsurprisingly, most of these are from X64

proof/refine/X64/ArchFinalise_R.thy
```diff
@@ -772,36 +754,38 @@ lemma isFinalCapability_corres'[Finalise_R_2_assms]:
...
-  done
+  apply (clarsimp simp: cap_irqs_def cap_irq_opt_def sameObjectAs_def3 isCap_simps arch_gen_obj_refs_def
+                        acap_relation_def
+                 split: cap.split_asm arch_cap.split_asm)
+  apply (case_tac cap; clarsimp simp: arch_gen_obj_refs_def)
+  apply (case_tac x12; clarsimp simp: get_cap_caps_of_state)
+  apply (drule invs_valid_ioports, clarsimp simp: valid_ioports_def ioports_no_overlap_def)
+  apply (drule_tac x="cap.ArchObjectCap (arch_cap.IOPortCap x31a x32)" in bspec, fastforce)
+  apply (drule_tac x="cap.ArchObjectCap (arch_cap.IOPortCap x31a x32a)" in bspec, fastforce)
+  by (clarsimp simp: cap_ioports_def valid_cap_def)
```
lsf37: is this clarsimp slow? otherwise it should be apply/done

proof/refine/X64/ArchFinalise_R.thy
```diff
@@ -772,36 +754,38 @@ lemma isFinalCapability_corres'[Finalise_R_2_assms]:
...
-                 split: cap.split_asm)
-  done
+  apply (clarsimp simp: cap_irqs_def cap_irq_opt_def sameObjectAs_def3 isCap_simps arch_gen_obj_refs_def
+                        acap_relation_def
+                 split: cap.split_asm arch_cap.split_asm)
+  apply (case_tac cap; clarsimp simp: arch_gen_obj_refs_def)
+  apply (case_tac x12; clarsimp simp: get_cap_caps_of_state)
+  apply (drule invs_valid_ioports, clarsimp simp: valid_ioports_def ioports_no_overlap_def)
+  apply (drule_tac x="cap.ArchObjectCap (arch_cap.IOPortCap x31a x32)" in bspec, fastforce)
+  apply (drule_tac x="cap.ArchObjectCap (arch_cap.IOPortCap x31a x32a)" in bspec, fastforce)
```
lsf37: should not use generated names

proof/refine/X64/ArchFinalise_R.thy
```diff
@@ -772,36 +754,38 @@ lemma isFinalCapability_corres'[Finalise_R_2_assms]:
...
+                   tcb_in_cur_domain'_def)
   apply (auto simp add: ran_def split: if_split_asm)
   done
 
-crunch invalidateTLBByASID
-  for asidTable[wp]: "\<lambda>s. P (armKSASIDTable (ksArchState s))"
+lemma hwASIDInvalidate_invs'[wp]:
+  "\<lbrace>invs'\<rbrace> hwASIDInvalidate asid vs \<lbrace>\<lambda>r. invs'\<rbrace>"
+  apply (simp add: hwASIDInvalidate_def)
+  by wpsimp
```
lsf37: `by (wpsimp ..)`?

This is probably all from old proofs, but it does stick out in the diff.

proof/refine/X64/ArchFinalise_R.thy
```diff
@@ -814,11 +798,27 @@ lemma deleteASIDPool_invs[wp]:
...
+     apply (simp add: fun_upd_def[symmetric] valid_obj'_def)
+     apply (case_tac rv, simp)
+     apply (subst inv_f_f, rule inj_onI, simp)+
+     apply (rule conjI)
+      apply clarsimp
+      apply (drule subsetD, blast)
+      apply clarsimp
+     apply (auto dest!: ran_del_subset)[1]
+    apply (wp getObject_valid_obj getObject_inv loadObject_default_inv
+             | simp add: X64.objBits_simps pageBits_def)+
```
lsf37: Indent of the second line looks off

proof/refine/ARM_HYP/EmptyFail_H.thy
```diff
@@ -105,6 +105,7 @@ lemma ArchRetypeDecls_H_deriveCap_empty_fail[intro!, wp, simp]:
   "isPageTableCap y \<or> isPageDirectoryCap y \<or> isPageCap y
    \<or> isASIDControlCap y \<or> isASIDPoolCap y \<or> isVCPUCap y \<or> isSGISignalCap y
    \<Longrightarrow> empty_fail (Arch.deriveCap x y)"
+  supply if_cong[cong]
   apply (simp add: ARM_HYP_H.deriveCap_def)
   by auto
```
lsf37: I'm tempted to collapse this to 
` by (auto simp: ARM_HYP_H.deriveCap_def cong: if_cong)`

proof/refine/X64/EmptyFail_H.thy
```diff
@@ -105,6 +105,7 @@ lemma ArchRetypeDecls_H_deriveCap_empty_fail[intro!, wp, simp]:
   "isPageTableCap y \<or> isPageDirectoryCap y \<or> isPageCap y \<or> isIOPortControlCap y
    \<or> isASIDControlCap y \<or> isASIDPoolCap y \<or> isPDPointerTableCap y \<or> isPML4Cap y \<or> isIOPortCap y
    \<Longrightarrow> empty_fail (Arch.deriveCap x y)"
+  supply if_cong[cong]
   apply (simp add: X64_H.deriveCap_def)
   by auto
```
lsf37: (marking this in case you decide to do the other one)

proof/refine/ARM/ArchFinalise_R.thy
lsf37: Ok, yes. That comment was more hopeful than anything. Given that this seems to affect pretty much only this one lemma, I agree that it's not worth it to introduce a finer-grained interface.

That means it's fine to remove the FIXME.

lsf37 (approved):
Very solid and another 10k lines we don't have to maintain in the future.

While I can understand your frustration with the locales you had to give up on, I don't think they are a problem at all. Maybe there is a solution of some kind, but nothing I can see that would not be super complex and hard to maintain/understand in the future. I think you have the right trade-off.

I don't have many detail comments. I got triggered a bit by some of the old X64 stuff, because it sticks out in the diff view. If you don't feel like cleaning up that old code, that's fine too. It's not a lot, though, so if you can bear it, we could leave the proofs slightly nicer than before.

proof/refine/Finalise_R.thy
```diff
@@ -1185,51 +599,352 @@ lemma reply_masters_rvk_fb_n [simp]: "reply_masters_rvk_fb n"
...
                     slot)
   apply (subst ball_ran_modify_map_eq)
    apply (clarsimp simp add: modify_map_def)
    apply fastforce
   apply (simp add: ball_ran_modify_map_eq)
   done
 
+end (* mdb_empty *)
+
+declare doUnbindNotification_def[simp]
```
corlewis: Preexisting, but thoughts on marking this `FIXME: remove`? It doesn't quite seem simple enough to want to unfold it everywhere.

proof/refine/Finalise_R.thy
```diff
@@ -1185,51 +599,352 @@ lemma reply_masters_rvk_fb_n [simp]: "reply_masters_rvk_fb n"
...
+  apply (clarsimp simp: valid_def)
+  apply (drule updateCap_stuff)
+  apply (clarsimp simp: cte_wp_at_ctes_of modify_map_def)
+  done
+
+lemma updateFreeIndex_cte_wp_at:
+  "\<lbrace>\<lambda>s. cte_at' p s \<and> P (cte_wp_at' (if p = p' then P'
+      o (cteCap_update (capFreeIndex_update (K idx))) else P') p' s)\<rbrace>
+    updateFreeIndex p idx
+  \<lbrace>\<lambda>rv s. P (cte_wp_at' P' p' s)\<rbrace>"
```
corlewis: The indentation of that precondition is horrible enough that it would be nice to fix while we're touching this file.

proof/refine/Finalise_R.thy
```diff
@@ -1417,38 +1113,92 @@ lemma emptySlot_cteCaps_of:
...
+     final_matters' (cteCap cte) \<Longrightarrow>
+     corres (=) (invs and cte_wp_at ((=) cap) ptr)
+                 (invs' and cte_wp_at' ((=) cte) (cte_map ptr))
+         (is_final_cap cap) (isFinalCapability cte)"
+  assumes Arch_finaliseCap_irq_node'[wp]:
+    "\<And>acap final P. Arch.finaliseCap acap final \<lbrace>\<lambda>s. P (Invariants_H.irq_node' s)\<rbrace>"
+  assumes prepareThreadDelete_irq_node'[wp]:
+    "\<And>t P. prepareThreadDelete t \<lbrace>\<lambda>s. P (Invariants_H.irq_node' s)\<rbrace>"
+  assumes cancelAllIPC_cte_wp_at'[wp]:
+    "\<And>epPtr P p. cancelAllIPC epPtr \<lbrace>cte_wp_at' P p\<rbrace>"
```
corlewis: There's a few generic functions in this locale that could possibly be moved to the generic file if their arch-equivalents were in the interface instead. At first glance their proofs look to be pretty simple and mostly crunches. This probably wouldn't actually save that much so feel free to ignore and leave as is.

proof/refine/Finalise_R.thy
```diff
@@ -3689,57 +2470,46 @@ lemma cteDeleteOne_tcbDomain_obj_at':
...
+     ThreadDecls_H.suspend t
+     \<lbrace>\<lambda>_. invs'\<rbrace>"
+  assumes delete_one_conc_fr_suspend_objs': (* obtained later by delete_one_conc_fr interpretation *)
+    "\<And>t.
+     \<lbrace>invs' and sch_act_simple and tcb_at' t and (\<lambda>s. t \<noteq> ksIdleThread s)\<rbrace>
+     ThreadDecls_H.suspend t
+     \<lbrace>\<lambda>rv. valid_objs'\<rbrace>"
+  assumes deletingIRQHandler_corres: (* obtained later by interpreting delete_one *)
+    "\<And>irq. corres dc einvs invs' (deleting_irq_handler irq) (deletingIRQHandler irq)"
+  assumes suspend_corres:
```
corlewis: I think this could have the `obtained later by interpreting delete_one` comment as well.

proof/refine/ARM_HYP/ArchMachine_R.thy
```diff
@@ -51,10 +51,8 @@ lemma dmo_maskInterrupt_True:
                    ct_not_inQ_def ct_idle_or_in_cur_domain'_def tcb_in_cur_domain'_def)
   done
 
-lemma setIRQState_irq_states':
-  "\<lbrace>valid_irq_states'\<rbrace>
-      setIRQState state irq
-   \<lbrace>\<lambda>rv. valid_irq_states'\<rbrace>"
+lemma setIRQState_irq_states'[Machine_R_assms, wp]:
+  "setIRQState state irq \<lbrace>valid_irq_states'\<rbrace>"
```
corlewis, line 55: Very minor but it looks like AARCH64 is the only arch that doesn't have the nicer version of this lemma. Might as well make it consistent as well while we're here.

proof/refine/AARCH64/Arch_R.thy
```diff
@@ -1877,9 +1877,9 @@ lemma performASIDControlInvocation_st_tcb_at':
   apply auto
   done
 
-lemmas arch_finalise_cap_aligned' = ArchRetypeDecls_H_AARCH64_H_finaliseCap_aligned'
+lemmas arch_finalise_cap_aligned' = Arch_finaliseCap_aligned'
 
-lemmas arch_finalise_cap_distinct' = ArchRetypeDecls_H_AARCH64_H_finaliseCap_distinct'
+lemmas arch_finalise_cap_distinct' = Arch_finaliseCap_distinct'
```
corlewis: Now that you've improved the original name, should we mark these as `FIXME: remove`?

corlewis (approved):
Very nice! Not that much to add on top of Gerwin's comments.

proof/refine/Finalise_R.thy
Xaphiosis: Can you clarify? I think I missed this, and would like to know a bit more details of your idea to check whether it might make things more generic.

proof/refine/Finalise_R.thy
Xaphiosis: Tricky. The lemmas [wp] is indicating that the requalified lemma needs to be [wp], but the declare is a declare I moved and have no clue whatsoever whether it should've been [wp] in the first place. I might try make it [wp] at the original site and see.

proof/refine/ARM/ArchFinalise_R.thy
Xaphiosis: Will remove on non-AARCH64

proof/refine/Finalise_R.thy
Xaphiosis: Tuned to this:
```
lemma updateFreeIndex_cte_wp_at:
  "⦃λs. cte_at' p s
        ∧ P (cte_wp_at' (if p = p' then P' ∘ (cteCap_update (capFreeIndex_update (K idx))) else P')
                        p' s)⦄
   updateFreeIndex p idx
   ⦃λrv s. P (cte_wp_at' P' p' s)⦄"
  apply (simp add: updateFreeIndex_def updateTrackedFreeIndex_def split del: if_split)
  apply (wpsimp wp: updateCap_cte_wp_at' getSlotCap_wp simp: cte_wp_at_ctes_of)
  apply (cases "p' = p"; simp)
  apply (case_tac cte, simp)
  done
```
does that work?

proof/refine/Finalise_R.thy
Xaphiosis: Trying to remove it now, seeing how many breakages that leaves. Will use it as an opportunity for further cleanup.

proof/refine/X64/EmptyFail_H.thy
Xaphiosis: I'll do all of these that don't have the `| wpc |` chain

proof/refine/ARM_HYP/ArchMachine_R.thy
Xaphiosis: I'm looking through all of the arches but can't find it in the final result...

proof/refine/AARCH64/Arch_R.thy
Xaphiosis: Very good eye! These are unused now

proof/refine/X64/ArchFinalise_R.thy
Xaphiosis: yup, will fix

proof/refine/Finalise_R.thy
corlewis: A lot better, I'm happy with that

proof/refine/AARCH64/Arch_R.thy
corlewis: I think there might have been something similar for another `Arch_` function if those ones are also unused.

proof/refine/Finalise_R.thy
corlewis: My thinking was that this interface has some assumptions about generic functions preserving generic properties. For example, this `cancelAllIPC` one and the ones following it about functions like `unbindNotification`.

From looking at the text my guess is that the existing proofs do big crunches about both generic and arch-specific parts, which then need to go in the arch file and force the interface to be like this. If we can split these proofs up appropriately then it feels like it should be possible to have only the arch functions in this interface and all the generic parts of the proof then in the generic file.

What's less clear is that with all of this being relatively simple proofs, I don't know if that will actually make things nicer.

proof/refine/ARM_HYP/ArchMachine_R.thy
corlewis: Oh, I see what's happened, I confused myself with the broken apart commit history! I forgot that it's only `Finalise_R` and `ArchFinalise_R` that are adjusted to reduce the diff size. Nothing to see here.

proof/refine/Finalise_R.thy
```diff
@@ -1812,7 +1806,7 @@ lemma unbindNotification_invs[wp]:
   apply (wpsimp wp: sbn'_valid_pspace'_inv sbn_sch_act' valid_irq_node_lift valid_dom_schedule'_lift
                     irqs_masked_lift setBoundNotification_ct_not_inQ sym_heap_sched_pointers_lift
                     untyped_ranges_zero_lift
-                simp: cteCaps_of_def o_def)
+                simp: cteCaps_of_def o_def  doUnbindNotification_def)
```
corlewis, line 1786: Extra space here in the `[squash] PR comments for split` commit

proof/refine/Finalise_R.thy
Xaphiosis: Oh I see now! I'll take a look. I remember there might have been a reason why I did one of those, and it might have been frustration.

proof/refine/Finalise_R.thy
Xaphiosis: Ah indeed, thanks, will fix.

Xaphiosis:
CRefine is currently a bit broken due to that `doUnbindNotification_def` simpset change. Am fixing as fast as rebuilds permit.

proof/refine/Finalise_R.thy
Xaphiosis: This ended up sending me on a merry adventure of education and locales again. I was able to migrate the crunch-related ones, that was very well spotted, it was a stupid copy-paste mistake from me.
The isFinal I was able to make generic, but it caused a circular dependency between the delete_one* locale interpretations and the interface locale. After bashing my head against it for a while, I ended up figuring out that `sublocale` for once does the expected thing, and was able to use that.
Will push that once I get CRefine fixed.

proof/refine/AARCH64/ArchFinalise_R.thy
```diff
@@ -1399,19 +1232,15 @@ lemma (in delete_one_conc_pre) finaliseCap_replaceable:
...
 
+named_theorems Finalise_R_3_assms
+
 lemma finaliseCap_cte_refs:
   "\<lbrace>\<lambda>s. s \<turnstile>' cap\<rbrace>
      finaliseCap cap final flag
    \<lbrace>\<lambda>rv s. fst rv \<noteq> NullCap \<longrightarrow> cte_refs' (fst rv) = cte_refs' cap\<rbrace>"
-  apply (simp  add: finaliseCap_def Let_def getThreadCSpaceRoot
-                    AARCH64_H.finaliseCap_def
+  apply (simp add: global.finaliseCap_def Let_def getThreadCSpaceRoot finaliseCap_def
```
corlewis, line 1243: Ah, I was confused by the names flipping here,but is this due to the moved `named_theorems` causing that annoying name reordering?

corlewis (approved):
Reviewed the `reduce Finalise_R_2 locale` commit, very nice! That's a fair bit better than what I was thinking might be possible!

proof/refine/AARCH64/ArchFinalise_R.thy
Xaphiosis: Yeah, the infamous named_theorems Arch flip. We still haven't figured out how it comes about exactly.

lsf37:
> Reviewed the `reduce Finalise_R_2 locale` commit, very nice! That's a fair bit better than what I was thinking might be possible!

I agree, that is very nice indeed!

lsf37 (approved):

Xaphiosis:
Squashed relevant commits, updated lines removed in PR comment (ended up over 10k, woo). Will merge once tests pass.

PR #993, opened 2026-05-12
arch-split Ipc_R

Xaphiosis:
🦆🦆🦆 In the process of doing Ipc_R, I discovered that TcbAcc_R could be improved, so there is an extra commit with that improvement in this PR before the Ipc_R arch-splitting starts. Again, please review commit-by-commit.

**Instructions:**

In real life, I did the usual addition of Ipc_R, hierarchy update, arch-split and then update. However, since the arch-split part first splits AARCH64, then takes the AARCH64 version of Ipc_R and only fixes what broke, we ended up with a small(er) diff there (again).

- `[squash][fake] refine: arch-split ..._R (other arches) - result of fixing up the ArchSchedule_R of other architectures - shows diff to AARCH64

Stats: 76 files changed, 3118 insertions(+), 20037 deletions(-)
Total: 16919 lines removed


proof/refine/TcbAcc_R.thy
```diff
@@ -10,6 +10,13 @@ theory TcbAcc_R
 imports ArchCSpace_R
 begin
 
+(* FIXME arch-split: move up *)
+arch_requalify_consts pageBitsForSize
+
+arch_requalify_facts mab_pb (* FIXME arch-split: from ArchTcbAcc_AI *)
```
corlewis, line 16: Are these FIXMEs that you want to resolve as part of this round or leave for later?

proof/refine/AARCH64/ArchTcbAcc_R.thy
```diff
@@ -1112,6 +872,67 @@ context Arch begin arch_global_naming
 
 named_theorems TcbAcc_R_3_assms
 
+lemma setMRs_corres:
+  assumes m: "mrs' = mrs"
+  shows
+  "corres (=) (tcb_at t and pspace_aligned and pspace_distinct and case_option \<top> in_user_frame buf)
+              (case_option \<top> valid_ipc_buffer_ptr' buf)
+              (set_mrs t buf mrs) (setMRs t buf mrs')"
```
corlewis, line 884: Out of curiosity, why is `setMRs_corres` the only one out of the MRs family that had to stay in the arch file? The definition of `setMRs` doesn't seem to do anything more arch-specific than `getMRs`.

corlewis: Oh, is it because its proof refers to `user_fpu_state` on architectures that have it? Fair enough then, probably not worth trying to abstract that.

proof/refine/CSpace_R.thy
```diff
@@ -3291,7 +3291,8 @@ lemma deriveCap_valid[wp]:
 
 end (* CSpace_R *)
 
-lemma lookup_cap_valid':
+(* FIXME: poor name, eliminate as wp argument *)
+lemma lookup_cap_valid'[wp]:
```
corlewis: It took me a moment to work out what you meant with the second part of this FIXME, and I'm not sure I would have without the context of this being added to `wp`. Possible reword?
```
(* FIXME: change name, eliminate existing uses as argument to wp method *)
```

proof/refine/TcbAcc_R.thy
Xaphiosis: I want to leave these for a second pass, as it involves diving into AInvs, including figuring out where the Machine_AI exports are, and having a clearer head to maybe improve something over there.

There's a FIXME a bit below this that says "review for sanity", which I'd appreciate an opinion on. I'm not sure if I captured the right idea.

proof/refine/AARCH64/ArchTcbAcc_R.thy
Xaphiosis: Yeah, that kept popping up, but even ignoring that it kept throwing up an update to the TCB arch state for which I was unable produce a nice interface.

proof/refine/CSpace_R.thy
Xaphiosis: Oh yes, much better, I'll take your version.

proof/refine/Ipc_R.thy
```diff
@@ -9,26 +9,103 @@ theory Ipc_R
...
+  and lim: "unat k * unat j < 2 ^ len_of TYPE ('a)"
+  shows  "k * i \<le> k * j"
+  using word_mult_le_mono1[OF ij] lim
+  by (simp add: mult.commute)
+
+(* FIXME arch-split: move, review for sanity *)
+lemma no_fail_zipWithM_x:
+  assumes P: "\<And>x y. no_fail P (f x y)"
+  assumes inv: "\<And>x y. f x y \<lbrace>P\<rbrace>"
+  shows "no_fail P (zipWithM_x f xs ys)"
```
corlewis: This looks reasonable to me. If you really want to though, you could strengthen it to be more like `no_fail_mapM_wp`.

proof/refine/Ipc_R.thy
```diff
@@ -39,94 +116,167 @@ definition
...
+  "msgMaxLength = word_of_nat msg_max_length"
+  by (simp add: msg_max_extra_caps_def msg_max_length_def
+                msgMaxExtraCaps_def msgExtraCapBits_def msgMaxLength_def)+
+
+locale Ipc_R =
+  assumes getMessageInfo_corres:
+    "\<And>t.
+     corres ((=) \<circ> message_info_map)
+            (tcb_at t and pspace_aligned and pspace_distinct) \<top>
+            (get_message_info t) (getMessageInfo t)"
```
corlewis: Probably not worth the effort but this looks like a generic function where the interface could be a bit lower, if we introduced a lemma for `msgInfoRegister`.

proof/refine/Ipc_R.thy
```diff
@@ -39,94 +116,167 @@ definition
...
+     \<lbrakk>isUntypedCap cap'\<rbrakk>
+     \<Longrightarrow> is_derived' m src cap' cap
+        = (isUntypedCap cap \<and> badge_derived' cap' cap \<and> descendants_of' src m = {})"
+  assumes is_derived'_Reply:
+    "\<And>cap cap' m src.
+     \<lbrakk>isReplyCap cap'\<rbrakk>
+     \<Longrightarrow> is_derived' m src cap' cap
+        = (isReplyCap cap \<and> capTCBPtr cap = capTCBPtr cap' \<and> capReplyMaster cap \<and> \<not> capReplyMaster cap')"
+  assumes maskCapRights_eq_null[simp]:
+    "(RetypeDecls_H.maskCapRights r cap = capability.NullCap) = (cap = capability.NullCap)"
```
corlewis: Again probably not worth it but this could be generic with an interface lemma for `Arch.maskCapRights` instead.

proof/refine/Ipc_R.thy
```diff
@@ -39,94 +116,167 @@ definition
...
+    "\<And>ep buffer n caps slots mi.
+     transferCapsToSlots ep buffer n caps slots mi \<lbrace>pspace_in_kernel_mappings'\<rbrace>"
+  assumes makeArchFaultMessage_sch_act:
+    "\<And>af t. makeArchFaultMessage af t \<lbrace>\<lambda>s. P (ksSchedulerAction s)\<rbrace>"
+  assumes is_derived'_IRQHandlerCap:
+    "\<And>cap cap' src (s::kernel_state).
+     \<lbrakk>isIRQHandlerCap cap'\<rbrakk> \<Longrightarrow> is_derived' (ctes_of s) src cap' cap =
+     (isIRQHandlerCap cap \<and> badge_derived' cap' cap)"
+  assumes storeWordUser_vms'[wp]:
+    "\<And>a w. storeWordUser a w \<lbrace>valid_machine_state'\<rbrace>"
```
corlewis: Looks possibly generalisable?

proof/refine/Ipc_R.thy
```diff
@@ -39,94 +116,167 @@ definition
...
+            (handleArchFaultReply (arch_fault_map ft) t label msg)"
+  assumes ctes_of_mdbNext_parentOf:
+    "\<And>(s'::kernel_state) cptr slot t master rights n master' rights' n' cte.
+     \<lbrakk> ctes_of s' \<turnstile> cte_map cptr \<rightarrow> cte_map slot;
+       ctes_of s' (cte_map cptr) = Some (CTE (capability.ReplyCap t master rights) n);
+       ctes_of s' (mdbNext (cteMDBNode cte)) = Some (CTE (capability.ReplyCap t master' rights') n');
+       ctes_of s' \<turnstile> cte_map slot \<leadsto> mdbNext (cteMDBNode cte)\<rbrakk>
+     \<Longrightarrow> ctes_of s' \<turnstile> cte_map cptr parentOf mdbNext (cteMDBNode cte)"
+  assumes get_mrs_inv'[wp]:
+    "\<And>t buf info P. getMRs t buf info \<lbrace>P\<rbrace>"
```
corlewis: Generalisable?

proof/refine/Ipc_R.thy
```diff
@@ -827,19 +922,62 @@ lemma tcts_idle'[wp]:
...
+  by (wp transferCapsToSlots_pres2 | simp)+
+
+crunch setExtraBadge
+  for if_unsafe'[wp]: if_unsafe_then_cap'
+
+lemma tcts_ifunsafe[wp]:
+  "\<lbrace>\<lambda>s. if_unsafe_then_cap' s \<and> distinct slots \<and>
+         (\<forall>x\<in>set slots.  cte_wp_at' (\<lambda>cte. cteCap cte = capability.NullCap) x s \<and>
+             ex_cte_cap_to' x s)\<rbrace> transferCapsToSlots ep buffer n caps slots mi
+   \<lbrace>\<lambda>rv. if_unsafe_then_cap'\<rbrace>"
```
corlewis, line 1023: I'm trying not to comment on style of proofs that you've just moved but this one is bad enough that it would be nice to fix now.

proof/refine/Ipc_R.thy
```diff
@@ -878,7 +1020,9 @@ lemma transferCapsToSlots_irq_handlers[wp]:
      transferCapsToSlots ep buffer n caps slots mi
   \<lbrace>\<lambda>rv. valid_irq_handlers'\<rbrace>"
   apply (wpsimp wp: transferCapsToSlots_presM[where vo=True and emx=False and drv=True and pad=False])
-     apply (clarsimp simp: is_derived'_def cte_wp_at_ctes_of badge_derived'_def)
+using [[show_types]]
```
corlewis: Should we add `show_types` to the theory linter so that it's pointed out by the tests?

proof/refine/Ipc_R.thy
```diff
@@ -907,69 +1053,52 @@ lemma transferCapsToSlots_irqs_masked'[wp]:
...
+  apply (wp | simp only: distinct.simps list.simps empty_iff)+
+  apply simp
+  done
+
+context Ipc_R begin
 
 lemma setExtraBadge_vms'[wp]:
   "\<lbrace>valid_machine_state'\<rbrace> setExtraBadge buffer b n \<lbrace>\<lambda>_. valid_machine_state'\<rbrace>"
-by (simp add: setExtraBadge_def) wp
+  by (simp add: setExtraBadge_def) (wp storeWordUser_vms')
```
corlewis: Isn't `storeWordUser_vms'` already `wp`?

proof/refine/Ipc_R.thy
```diff
@@ -1429,6 +1478,32 @@ crunch copy_mrs
   for valid_arch_state[wp]: valid_arch_state
   (wp: crunch_wps)
 
+lemma corres_liftE_lift:
+  "corres r1 P P' m m' \<Longrightarrow>
+  corres (f1 \<oplus> r1) P P' (liftE m) (withoutFailure m')"
```
corlewis: `FIXME: move`

proof/refine/ARM/ArchIpc_R.thy
```diff
@@ -82,23 +87,23 @@ lemma is_derived'_IRQHandlerCap[Ipc_R_assms]:
...
   by (clarsimp simp add: ARM.is_derived'_def gen_isCap_simps)
-     (cases cap; clarsimp simp: badge_derived'_def capMasterCap_def)
+     (cases cap; clarsimp simp: badge_derived'_def capMasterCap_def vsCapRef_generic isCap_simps)
 
 lemma storeWordUser_vms'[Ipc_R_assms, wp]:
   "storeWordUser a w \<lbrace>valid_machine_state'\<rbrace>"
 proof -
   have aligned_offset_ignore:
-    "\<And>(l::machine_word) (p::machine_word) sz. l<8 \<Longrightarrow> p && mask 3 = 0 \<Longrightarrow>
+    "\<And>(l::machine_word) (p::machine_word) sz. l<4 \<Longrightarrow> p && mask 2 = 0 \<Longrightarrow>
```
corlewis: Is there anything we can do about the numbers in this proof?

proof/refine/ARM/ArchIpc_R.thy
```diff
@@ -125,7 +130,7 @@ lemma isArchObjectCap_maskCapRights[Ipc_R_assms]:
   by (cases acap; simp add: ARM_H.maskCapRights_def isCap_simps)
 
 lemma isFrameCap_maskCapRights[simp]:
-  "isArchCap isFrameCap (global.maskCapRights R c) = isArchCap isFrameCap c"
+  "isArchCap isPageCap (global.maskCapRights R c) = isArchCap isPageCap c"
```
corlewis: Probably want to change the name of this lemma back to `isPageCap` as well.

corlewis: I guess this also works as yet another reminder to us that it would be nice to make the `FrameCap` and `PageCap` terminology more consistent across architectures.

proof/refine/ARM/ArchIpc_R.thy
```diff
@@ -292,12 +298,16 @@ lemma max_message_size_less_max_ipc_words[Ipc_R_assms]:
    \<Longrightarrow> word_size * (word_of_nat msg_max_extra_caps + (word_of_nat msg_max_length + n))
       < max_ipc_words * word_size"
   apply (simp add: msg_max_extra_caps_def msg_max_length_def max_ipc_words word_size_def)
-  apply (rule_tac y="0x3D8 + 8 * 4" in order_le_less_trans)
+  apply (rule_tac y="0x1EC + 4 * 4" in order_le_less_trans)
```
corlewis: Could be painful since we've unfolded everything just before this, but it would be nice to improve these magic numbers.

proof/refine/ARM/ArchIpc_R.thy
corlewis: (If there is, same for all architectures)

proof/refine/ARM/ArchIpc_R.thy
corlewis: > Probably want to change the name of this lemma back to `isPageCap` as well.

And on ARM_HYP and X64 as well

proof/refine/ARM/ArchIpc_R.thy
corlewis: (If it is possible, same for all architectures)

corlewis (approved):
Once again, very nice to see so much able to be generalised and so much good cleanup done along the way! I have a few suggestions for extra lemmas that are possibly able to be made general, but if not then the interface as it is looks good to me.

proof/refine/Ipc_R.thy
Xaphiosis: Hmm, looking at that, I'm not sure I want to introduce/debug that complexity. Would it make sense to rename this one to `no_fail_zipWithM_x_lift` then, as it's a straight up lift?

proof/refine/Ipc_R.thy
Xaphiosis: Yeah, this is generic, thanks for `msgInfoRegister_msg_info_register`. Thank you!

proof/refine/Ipc_R.thy
Xaphiosis: will do

proof/refine/Ipc_R.thy
Xaphiosis: I don't see it. It depends on depending on word_size_bits, using storeWord_um_inv which directly exposes the number of bytes in a word as a list.

Xaphiosis: Best I can do is replace some magic numbers so the text is (nearly) the same on all arches

proof/refine/Ipc_R.thy
Xaphiosis: Oh yes, turns out `getRegister_inv` was already exported, thank you!

proof/refine/Ipc_R.thy
Xaphiosis: Good idea! Can I delegate this?

proof/refine/Ipc_R.thy
Xaphiosis: Indeed. I'll poke a few in the general area also.

proof/refine/Ipc_R.thy
Xaphiosis: Yes, and will swap to wpsimp here. Sometimes it would be nice to get a warning for duplicate wp insertions...

proof/refine/Ipc_R.thy
Xaphiosis: cheers

proof/refine/ARM/ArchIpc_R.thy
Xaphiosis: Yeah, I've managed to clean this up to:
```
lemma storeWordUser_vms'[Ipc_R_assms, wp]:
  "storeWordUser a w ⦃valid_machine_state'⦄"
proof -
  have aligned_offset_ignore:
    "⋀(l::machine_word) (p::machine_word) sz. l < word_size ⟹ p && mask word_size_bits = 0 ⟹
       p+l && ~~ mask pageBits = p && ~~ mask pageBits"
  proof -
    fix l p sz
    assume al: "(p::machine_word) && mask word_size_bits = 0"
    assume "(l::machine_word) < word_size"
    hence less: "l < 2 ^ word_size_bits" by (simp add: word_size_bits_def word_size_def)
    have le: "word_size_bits ≤ pageBits" by (simp add: word_size_bits_def pageBits_def)
    show "?thesis l p sz"
      by (rule is_aligned_add_helper[simplified is_aligned_mask,
          THEN conjunct2, THEN mask_out_first_mask_some,
          where n=word_size_bits, OF al less le])
  qed

  show ?thesis
    apply (simp add: valid_machine_state'_def storeWordUser_def
                     doMachineOp_def split_def)
    apply wp
    apply clarsimp
    apply (drule use_valid)
    apply (rule_tac x=p in storeWord_um_inv, simp+)
    apply (subst (asm) word_size_bits_def[symmetric]) (* "is_aligned a _" from storeWord_um_inv *)
    apply (drule_tac x=p in spec)
    apply (erule disjE, simp_all)
    apply (erule conjE)
    apply (erule disjE, simp)
    apply (simp add: pointerInUserData_def)
    apply (subgoal_tac "a && ~~ mask pageBits = p && ~~ mask pageBits", simp)
    apply (simp only: is_aligned_mask[of _ word_size_bits])
    apply (elim disjE, simp_all)
      apply (rule aligned_offset_ignore[simplified word_size_def, symmetric]; simp)+
    done
qed
```
unfortunately the `word_size_def` and `storeWord_um_inv` make this non-generic and I don't know how to get around that currently, especially since `storeWord_um_inv` comes from AInvs.

proof/refine/ARM/ArchIpc_R.thy
Xaphiosis: Interestingly on several arches it looked promising to have an arbitrary `isArchCap P` here, but then SMC caps broke it, so I'll do the renames.

proof/refine/ARM/ArchIpc_R.thy
Xaphiosis: Managed to get it to the same proof on all arches (but not generic, obviously):
```
lemma max_message_size_less_max_ipc_words[Ipc_R_assms]:
  "n ≤ 4
   ⟹ word_size * (word_of_nat msg_max_extra_caps + (word_of_nat msg_max_length + n))
      < max_ipc_words * word_size"
  apply (rule_tac y="word_size * (word_of_nat msg_max_extra_caps + (word_of_nat msg_max_length + 4))"
               in order_le_less_trans)
   apply (rule word_mult_le_mono1' | rule word_plus_mono_right)+
       apply (simp add: msg_max_extra_caps_def msg_max_length_def max_ipc_words word_size_def)+
  done
```

proof/refine/Ipc_R.thy
Xaphiosis: (made an attempt, see below on the ArchIpc_R comment)

proof/refine/Ipc_R.thy
corlewis: Sure. Do we want a fixme pointing at the `mapM_wp` lemma and saying that there could be an equivalent one for `zipWithM`?

proof/refine/Ipc_R.thy
corlewis: Sure, https://github.com/seL4/ci-actions/pull/468

proof/refine/Ipc_R.thy
corlewis: That probably wouldn't be hard to add, although I imagine it would be a tsunami of warnings when first turned on.

proof/refine/Ipc_R.thy
corlewis: Right, of course, it's working directly on the words. It's a lot nicer with your changes though, thanks for making them!

proof/refine/Ipc_R.thy
Xaphiosis: No idea. Will wait for @lsf37's thoughts

proof/refine/ARM/ArchIpc_R.thy
corlewis: Nice, thank you! Now that you've made it the same text on all of them, do we want to do the trick of proving and requalifying it immediately in the generic file? Or would that be too optimistic that it will always stay the same text.

proof/refine/Ipc_R.thy
Xaphiosis: I guess the thing to do would be to enable/disable the capability for it, so people can turn it on for cleanup. Then do some cleanup, and make it the default with places where we turn it off (e.g. CRefine) with a FIXME?

proof/refine/ARM/ArchIpc_R.thy
Xaphiosis: I'm never sure. On the one hand, it'll make things shorter, but on the other any changes to it would need to be checked on all the architectures. But then you're also correct that drift is possible... but on the fourth hand, will this thing ever change again? Hmm.

proof/refine/Ipc_R.thy
Xaphiosis: Thank you!

proof/refine/AARCH64/ArchTcbAcc_R.thy
```diff
@@ -1100,6 +857,9 @@ lemma tcbSchedAppend_pspace_in_kernel_mappings'[TcbAcc_R_2_assms]:
   "tcbSchedAppend t \<lbrace>pspace_in_kernel_mappings'\<rbrace>"
   by wpsimp
 
+(* length_type is machine_word on all architectures *)
```
lsf37: It's probably worth a FIXME or issue: we should declare length_type generically, not per architecture in ASpec. The code also assumes everywhere that it is word_t, so it can't really vary from that.

proof/refine/Ipc_R.thy
```diff
@@ -9,26 +9,103 @@ theory Ipc_R
 imports ArchFinalise_R
 begin
 
-context begin interpretation Arch . (*FIXME: arch-split*)
+(* FIXME arch-split: sendSignal_def already leaks that this is an OR *)
+arch_requalify_facts (A) combine_ntfn_badges_def (* Machine_AI *)
```
lsf37: The fact that notification badges are combined via OR is not arch specific as a concept. It's the specified generic mechanism described in the manual. I assume it ended up arch specific in ASpec by accident or lack of generic base theory (and leaked early, because it never was intended to be arch specific).

It is also generic that they are words, although how many bits are usable in a specific badge is indeed arch specific and cap specific.

Not sure we need to do anything about that, but I would rephrase the FIXME as: make the definition of combine_ntfn_badges generic so we don't get confused about the direction later.

proof/refine/Ipc_R.thy
lsf37: I would vote for `no_fail_zipWithM_x_inv`, since we're demanding invariance of P. A pointer to `no_fail_mapM_wp` in a comment (not necessarily FIXME) might be useful for future selves, so I'd also be for that if we're tweaking it.

proof/refine/Ipc_R.thy
```diff
@@ -1883,36 +1848,35 @@ lemma handle_fault_reply_registers_corres:
...
-      apply (rule corres_split)
-         apply (rule asUser_corres')
-         apply(simp add: setRegister_def sanitise_register_def
-                         sanitiseRegister_def syscallMessage_def Let_def cong: register.case_cong)
-         apply(subst zipWithM_x_modify)+
-         apply(rule corres_modify')
-          apply (simp|wp)+
+  apply (corres corres: arch_getSanitiseRegisterInfo_corres)
+      apply (simp add: sanitiseRegister_sanitise_register)
+      apply (corres corres: asUser_corres' wp: no_fail_zipWithM_x det_no_fail det_setRegister)
```
lsf37: Would it make sense to put the `simp` into the first `corres` and then combine the two `corres` incantations or does that conflict with something? `corres` is fairly conservative with `simp`, so this might also just not work.

proof/refine/Ipc_R.thy
```diff
@@ -3276,6 +3211,7 @@ lemma receiveIPC_corres:
               apply (clarsimp simp add: valid_ep_def valid_pspace_def)
               apply (drule(1) sym_refs_obj_atD[where P="\<lambda>ko. ko = Endpoint e" for e])
               apply (fastforce simp: st_tcb_at_refs_of_rev elim: st_tcb_weakenE)
+             (* FIXME weak elimination rule warning *)
```
lsf37, line 3302: I assume there is already the warning, do we need a separate FIXME  for it?

proof/refine/ARM/ArchIpc_R.thy
lsf37: Just removing the magic numbers is already nice. If that's all we can reasonably do here, that's fine. (I don't see anything obvious either)

proof/refine/ARM/ArchIpc_R.thy
lsf37: Interesting, I wouldn't have expected this to work on other architectures either, since `maskCapRights` does do something to the cap, so it should not preserve any P, just the kinds of P that only depend on the  cap type.

lsf37 (approved):
Nice! I was hoping Ipc_R would be fairly generic, but this is more than I expected. And the TcbAcc improvement is also excellent. 

No substantial new things from my side, only minor comments. Corey has a very keen eyes :-) Well done on spotting those generalisations.

proof/refine/ARM/ArchIpc_R.thy
lsf37: My jet lagged brain does not remember what role the 4 plays and if that could bite us in the future, but the rest of it does indeed look like it could be one of those "generic statement with same proof text on all architectures" kind of lemma. It's not something that I would expect much change for.

Xaphiosis: I investigated. The 4 is a result of the 3 words loaded in captransfer_from_words being simplified with some other offset added on in load_cap_transfer. So the term is generic, and likely to stay the same on all arches in the future. I'll make `max_message_size_less_max_ipc_words` a same-on-all-arch proof.

proof/refine/Ipc_R.thy
Xaphiosis: Not sure about these, I'm trying to remove all warnings if possible, and this one is not possible. I've left a bunch of these in the past, should I stop doing that?

proof/refine/Ipc_R.thy
Xaphiosis: I can add the `sanitiseRegister_sanitise_register` to the corres as `simp:`, but the reason the proof doesn't work is that it needs the `simp` in between those two corres steps. I am not sure why. Tracing it only shows `subst_all(2): (⋀x. ?a = x ⟹ PROP ?P x) ≡ PROP ?P ?a`

proof/refine/AARCH64/ArchTcbAcc_R.thy
Xaphiosis: will add comment

proof/refine/Ipc_R.thy
Xaphiosis: will add comment

proof/refine/RISCV64/ArchIpc_R.thy
```diff
@@ -70,45 +70,24 @@ lemma is_derived'_IRQHandlerCap[Ipc_R_assms]:
...
-    apply (subgoal_tac "a && ~~ mask pageBits = p && ~~ mask pageBits", simp)
-    apply (simp only: is_aligned_mask[of _ word_size_bits])
-    apply (elim disjE, simp_all)
-      apply (rule aligned_offset_ignore[simplified word_size_def, symmetric]; simp)+
-    done
-qed
+(* FIXME: the code assumes that it is word_t, so length_type should be defined generically in ASpec,
+   not per architecture *)
+(* variant of storeWord_um_inv which does not expose architecture-specific information *)
+lemma storeWord_um_inv'[Ipc_R_assms]:
```
corlewis: That first FIXME seems to have been copied here accidentally

proof/refine/RISCV64/ArchIpc_R.thy
```diff
@@ -70,45 +70,24 @@ lemma is_derived'_IRQHandlerCap[Ipc_R_assms]:
...
+  apply (clarsimp simp add: word_size_bits_def word_size_def)
+  apply (frule is_aligned_no_overflow_mask)
+  apply (subst add.commute) (* want "\<le> a + _" *)
+  apply (erule disjE, clarsimp simp: mask_def)
+  apply (simp (no_asm) add: word_le_nat_alt)
+  apply (auto simp add: unat_plus_simple[THEN iffD1] word_plus_mono_right2 mask_def)
+  done
 
-lemma isArchObjectCap_maskCapRights[Ipc_R_assms]:
+lemma lemma isArchObjectCap_maskCapRights[Ipc_R_assms]:
```
corlewis: Accidental extra lemma

proof/refine/X64/ArchIpc_R.thy
```diff
@@ -74,43 +74,22 @@ lemma is_derived'_IRQHandlerCap[Ipc_R_assms]:
...
-    apply (subgoal_tac "a && ~~ mask pageBits = p && ~~ mask pageBits", simp)
-    apply (simp only: is_aligned_mask[of _ word_size_bits])
-    apply (elim disjE, simp_all)
-      apply (rule aligned_offset_ignore[simplified word_size_def, symmetric]; simp)+
-    done
-qed
+(* FIXME: the code assumes that it is word_t, so length_type should be defined generically in ASpec,
+   not per architecture *)
+(* variant of storeWord_um_inv which does not expose architecture-specific information *)
+lemma storeWord_um_inv'[Ipc_R_assms]:
```
corlewis: FIXME here as well, not on other three architectures

proof/refine/RISCV64/ArchIpc_R.thy
Xaphiosis: done, squashed into the review commit. thank you

proof/refine/RISCV64/ArchIpc_R.thy
Xaphiosis: done, squashed into the review commit. thank you

proof/refine/X64/ArchIpc_R.thy
Xaphiosis: done, squashed into the review commit. thank you

lsf37 (approved):
Happy with the fixes, Corey already commented on the bits that slipped in.

Xaphiosis:
Squashed, will merge once tests succeed.

PR #995, opened 2026-05-21
crefine: trivial adjustment for update to the handling of the ipc queues

michaelmcinerney:
Test with https://github.com/seL4/seL4/pull/1576


corlewis (approved):

lsf37 (approved):
👍 

Let's merge #993 first, but after that this is ready to go.

Xaphiosis (approved):
I think historically we used "trivial" in commits which don't change spec/proofs. I'd drop the "trivial" from the commit message.

PR #996, opened 2026-05-21
Add `ccorres_exec_l_pre` method

michaelmcinerney:
I'd be happy to squash commits however people would like. I did not add a body to the commit message of the commit containing the new method, since there is already a comment where the method is defined. I could add the comment to the body of the commit message if that's preferable.

I could do this for `master`, too, if everyone is happy with this method. There are only a few uses of this pattern there, but it may be good to consolidate the tech between `master` and `rt`.

### DIFF
--- a/proof/crefine/RISCV64/IpcCancel_C.thy
+++ b/proof/crefine/RISCV64/IpcCancel_C.thy
@@ -522,7 +522,7 @@ lemma tcb_queue_prepend_ccorres:
      (tcbQueuePrepend queue tcbPtr) (Call tcb_queue_prepend_'proc)"
   supply if_split[split del]
   unfolding tcbQueuePrepend_def
-  apply (rule ccorres_symb_exec_l'[OF _ _ stateAssert_sp]; (solves wpsimp)?)
+  apply ccorres_exec_l_pre
   apply (cinit lift: tcb_')
    \<comment> \<open>cinit is not able to lift queue_' because queue_' is later modified in the C program\<close>
    apply (rule_tac xf'=queue_' in ccorres_abstract, ceqv, rename_tac cqueue)
@@ -607,7 +607,7 @@ lemma tcb_queue_append_ccorres:
      (tcbQueueAppend queue tcbPtr) (Call tcb_queue_append_'proc)"
   supply if_split[split del]
   unfolding tcbQueueAppend_def
-  apply (rule ccorres_symb_exec_l'[OF _ _ stateAssert_sp]; (solves wpsimp)?)
+  apply ccorres_exec_l_pre
   apply (cinit lift: tcb_')
    \<comment> \<open>cinit is not able to lift queue_' because queue_' is later modified in the C program\<close>
    apply (rule_tac xf'=queue_' in ccorres_abstract, ceqv, rename_tac cqueue)
@@ -740,8 +740,8 @@ proof -
   note word_less_1[simp del]
 
   show ?thesis
-    unfolding tcbSchedEnqueue_def K_bind_apply
-    apply (rule ccorres_symb_exec_l'[OF _ _ stateAssert_sp]; (solves wpsimp)?)+
+    unfolding tcbSchedEnqueue_def
+    apply ccorres_exec_l_pre+
     apply (cinit' lift: tcb_')
      apply (rule ccorres_symb_exec_l)
         apply (rule ccorres_assert)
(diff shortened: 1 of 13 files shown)


proof/crefine/RISCV64/Ipc_C.thy
```diff
@@ -5844,9 +5834,9 @@ lemma tcbEPAppend_ccorres[corres]:
      (\<lbrace>\<acute>thread = tcb_ptr_to_ctcb_ptr thread\<rbrace> \<inter> \<lbrace>\<acute>epptr = ep_Ptr epPtr\<rbrace>
       \<inter> \<lbrace>cendpoint_state_relation state \<acute>ep_state\<rbrace>) hs
      (tcbEPAppend thread epPtr state) (Call tcbEPAppend_'proc)"
-  unfolding tcbEPAppend_def haskell_assert_def K_bind_apply fun_app_def
-  apply (rule ccorres_symb_exec_l'[OF _ _ stateAssert_sp]; (solves wpsimp)?)
-  apply (rule ccorres_symb_exec_l'[OF _ _ get_ep_sp']; (solves wpsimp)?)
+  unfolding tcbEPAppend_def
+  apply (ccorres_exec_l_pre ccorres_exec_l_pre: get_ep_sp')+
+  apply clarsimp
```
michaelmcinerney: I should remove this `clarsimp`, since it's before the `cinit'`.

michaelmcinerney: Done in a separate commit, which I'll squash.

proof/crefine/RISCV64/IpcCancel_C.thy
```diff
@@ -3407,8 +3407,9 @@ lemma tcbEPDequeue_ccorres[corres]:
      (\<lbrace>\<acute>thread = tcb_ptr_to_ctcb_ptr thread\<rbrace> \<inter> \<lbrace>\<acute>epptr = ep_Ptr epPtr\<rbrace>) hs
      (tcbEPDequeue thread epPtr) (Call tcbEPDequeue_'proc)"
   supply if_split[split del]
-  unfolding tcbEPDequeue_def haskell_assert_def K_bind_apply fun_app_def
-  apply (rule ccorres_symb_exec_l'[OF _ _ get_ep_sp']; (solves \<open>wpsimp wp: empty_fail_getEndpoint\<close>)?)
+  unfolding tcbEPDequeue_def
+  apply (ccorres_exec_l_pre ccorres_exec_l_pre: get_ep_sp'
+                                            wp: empty_fail_getEndpoint)
```
corlewis, line 3412: It's valid style of course but this is a case where aligning on the `:` feels particularly wasteful of space. It's fine here but what would we want to do in a case where more `wp` rules are needed and it starts needing multiple lines?

corlewis (approved):
Thanks for adding this! Method looks good to me, with it being particularly helpful having all the examples of you using it to improve the existing proofs.

I'm happy with the commit structure, although I would add something like the first sentence of your comment to the commit adding the method, so that we can have an idea of what it is when looking at the log.

michaelmcinerney:
I'd like to add just a little bit more to the comment for the method. Instead of just saying "consuming a line via a strongest postcondition rule", I'd like to say "consuming a line of the abstract function via a strongest postcondition rule", since these are for `ccorres` rules with abstract and concrete functions, and I wouldn't want anyone to think we're consuming a line of both the abstract and concrete function at the same time.

lsf37:
> I'd like to add just a little bit more to the comment for the method. Instead of just saying "consuming a line via a strongest postcondition rule", I'd like to say "consuming a line of the abstract function via a strongest postcondition rule", since these are for `ccorres` rules with abstract and concrete functions, and I wouldn't want anyone to think we're consuming a line of both the abstract and concrete function at the same time.

Good idea.

michaelmcinerney:
The tests are being messed up by using the same C branch for various different L4V branches. The latest commit on the C branch here is necessary for `handleInvocation_ccorres`, but the proofs here are older and don't work with the latest commit. I can make another branch to test this with, or we can just ignore the tests for now and make sure they all pass later on.

lsf37:
Happy to merge this as is and make sure the proofs work in the final PR of this chain.

We should check the AutoCorres session manually because it is not in the run_tests set for MCS and we have touched AutoCorres files.

Edit: have done that now and the AutoCorres tests are working fine. Ready to merge.

lsf37 (approved):

PR #997, opened 2026-05-21
arch-split CNodeInv_R

Xaphiosis:
🦆🦆🦆 CNodeInv_R is huge, over 9k lines, and accumulated a number of needed interface changes in previous theories. I have isolated those to a separate commit for review (`interface changes needed for CNodeInv_R split`), hopefully it'll help. Again, please review commit-by-commit.

Locales were a bit annoying in that I needed another _gen locale for mdb_swap and mdb_move, so those ended up with funky suffixed names.

**Instructions:**

In real life, I did the usual addition of CNodeInv_R, hierarchy update, arch-split and then update. However, since the arch-split part first splits AARCH64, then takes the AARCH64 version of CNodeInv_R and only fixes what broke, we ended up with a small(er) diff there (again).

- `[squash] interface changes needed for CNodeInv_R split` - interface changes and improvements in imported theories
- updates in downstream theories: only refine+orphanage this time

Stats: 61 files changed, 4006 insertions(+), 38544 deletions(-)
Total: 34538 lines removed


proof/refine/AARCH64/ArchCSpace1_R.thy
```diff
@@ -196,14 +196,14 @@ lemma can_be_is[CSpace1_R_assms]:
   apply (auto simp: Let_def isCap_simps is_cap_simps dest!: acap_relation_SGISignalCapD)[1]
   done
 
-lemma maskCap_valid[simp]:
+lemma maskCap_valid[CSpace1_R_assms, simp]:
   "s \<turnstile>' global.maskCapRights R cap = s \<turnstile>' cap"
   by (clarsimp simp: valid_cap'_def global.maskCapRights_def isCap_simps
```
corlewis, line 201: Not sure if it would be worth it but this could probably be split into a generic and an `Arch.maskCapRights` lemma.

proof/refine/AARCH64/ArchCSpace1_R.thy
```diff
@@ -543,7 +537,7 @@ lemma arch_gen_refs_cap_relation_Master_eq[CSpace1_R_assms]:
    \<Longrightarrow> arch_gen_refs c = arch_gen_refs c'"
   by (simp split: cap_relation_split_asm arch_cap.split_asm)
 
-lemma descendants_of_update_ztc:
+lemma descendants_of_update_ztc[CSpace1_R_assms]:
```
corlewis, line 540: Hard to tell because it's so big but this looks possibly generalisable?

proof/refine/AARCH64/ArchCSpace1_R.thy
```diff
@@ -907,10 +901,10 @@ context Arch begin arch_global_naming
 
 named_theorems CSpace1_R_2_assms
 
-lemma weak_derived_updateCapData:
+lemma weak_derived_updateCapData[CSpace1_R_2_assms]:
   "\<lbrakk> updateCapData P x c \<noteq> NullCap; weak_derived' c c';
      capBadge (updateCapData P x c) = capBadge c' \<rbrakk>
-  \<Longrightarrow> weak_derived' (updateCapData P x c) c'"
+   \<Longrightarrow> weak_derived' (updateCapData P x c) c'"
```
corlewis: Also possibly generalisable.

proof/refine/AARCH64/ArchCSpace_I.thy
```diff
@@ -22,7 +22,7 @@ lemma arch_capUntypedPtr_simps[simp]:
   "Arch.capUntypedPtr (VCPUCap r) = r"
   by (auto simp: AARCH64_H.capUntypedPtr_def)
 
-lemma maskCapRights_allRights[simp]:
+lemma maskCapRights_allRights[CSpace_I_assms, simp]:
   "maskCapRights allRights c = c"
```
corlewis, line 26: Potential for split between `global.maskCapRights` and `Arch.maskCapRights`.

proof/refine/AARCH64/ArchCSpace_R.thy
```diff
@@ -549,7 +549,7 @@ lemma create_reply_master_corres[CSpace_R_2_assms]:
   apply (clarsimp simp: valid_pspace'_def cte_wp_at'_def)
   done
 
-lemma cte_map_nat_to_cref:
+lemma cte_map_nat_to_cref[CSpace_R_2_assms]:
   "\<lbrakk> n < 2 ^ b; b < word_bits \<rbrakk> \<Longrightarrow>
    cte_map (p, nat_to_cref b n) = p + (of_nat n * 2^cte_level_bits)"
```
corlewis: Hard to tell if this proof depends on the word size but it looks like the same text on all architectures, so could be moved to generic file even if it does.

proof/refine/AARCH64/ArchFinalise_R.thy
corlewis: It would need a bit of work but this and `finaliseCap_cte_refs` could probably be split into generic and `Arch.finaliseCap` parts.

proof/refine/ARM/ArchCSpace1_R.thy
```diff
@@ -233,53 +233,40 @@ qed
...
+  "Arch.updateCapData P d acap \<noteq> NullCap \<Longrightarrow>
+   capMasterCap (Arch.updateCapData P d acap) = capMasterCap (ArchObjectCap acap)"
+  by (cases acap; simp add: ARM_H.updateCapData_def split: if_split_asm)
+
+context begin (* private method *)
+
+private method updateCapData_cases for c =
+  (cases c; simp add: global.updateCapData_def isCap_simps Let_def),
+  (rename_tac arch_capability),
+  (case_tac arch_capability; simp add: ARM_H.updateCapData_def isCap_simps Let_def)
```
corlewis, line 246: Backporting this to older architectures is nice, thanks for doing that!

proof/refine/CNodeInv_R.thy
```diff
@@ -155,31 +155,257 @@ lemma cnode_invok_case_cleanup:
...
+  shows   "\<lbrace>\<lambda>s. cap \<noteq> NullCap \<longrightarrow> P s\<rbrace> deriveCap x cap \<lbrace>\<lambda>rv s. rv \<noteq> NullCap \<longrightarrow> Q rv s\<rbrace>,-"
+  apply (cases "cap = NullCap")
+   apply (simp add: deriveCap_def gen_isCap_simps)
+   apply (wp | simp)+
+  apply (rule hoare_strengthen_postE_R, rule assms)
+  apply simp
+  done
+
+lemma hasCancelSendRights_not_Null:
+  "hasCancelSendRights cap \<Longrightarrow> isEndpointCap cap"
```
corlewis: Feel free to ignore since this PR is just moving this lemma, but the name doesn't seem to match the lemma statement and this could probably be moved even further to an earlier file.

proof/refine/CNodeInv_R.thy
```diff
@@ -155,31 +155,257 @@ lemma cnode_invok_case_cleanup:
...
+  apply (clarsimp simp: updateCapData_def gen_isCap_simps)
+  done
+
+lemma untyped_derived_eq_refl:
+  "untyped_derived_eq c c"
+  by (simp add: untyped_derived_eq_def)
+
+definition not_recursive_ctes :: "kernel_state \<Rightarrow> machine_word set" where
+ "not_recursive_ctes s \<equiv> {ptr. \<exists>cap. cteCaps_of s ptr = Some cap
+                             \<and> \<not> (isZombie cap \<and> capZombiePtr cap = ptr)}"
```
corlewis: Existing lemma with very bad indentation.

proof/refine/CNodeInv_R.thy
```diff
@@ -155,31 +155,257 @@ lemma cnode_invok_case_cleanup:
...
+  "no_cte_prop P = (if \<forall>sl cte. \<lbrace>P\<rbrace> setCTE sl cte \<lbrace>\<lambda>_. P\<rbrace> then P else \<top>)"
+
+definition capToRPO :: "capability \<Rightarrow> machine_word option \<times> nat" where
+ "capToRPO cap \<equiv> case cap of
+    NullCap \<Rightarrow> (None, 0)
+  | Zombie p zt n \<Rightarrow> (Some p, 2)
+  | _ \<Rightarrow> (None, 3)"
+
+locale CNodeInv_R =
+  fixes arch_finalise_prop_stuff :: "(kernel_state \<Rightarrow> bool) \<Rightarrow> bool"
```
corlewis, line 225: Unusual name, could probably use a comment indicating what it's for, particularly in the architectures where it's not used.

proof/refine/CNodeInv_R.thy
```diff
@@ -155,31 +155,257 @@ lemma cnode_invok_case_cleanup:
...
+  assumes arch_finalise_prop_stuff_top[simp]:
+    "arch_finalise_prop_stuff \<top>"
+  assumes rightsFromWord_correspondence:
+    "\<And>w. rightsFromWord w = rights_mask_map (data_to_rights w)"
```
corlewis: Looks generalisable

proof/refine/CNodeInv_R.thy
```diff
@@ -155,31 +155,257 @@ lemma cnode_invok_case_cleanup:
...
+  assumes prepareThreadDelete_ctes_of[wp]:
+    "\<And>P t. prepareThreadDelete t \<lbrace>\<lambda>s. P (ctes_of s)\<rbrace>"
+  assumes prepareThreadDelete_not_recursive_ctes:
+    "\<And>P t. prepareThreadDelete t \<lbrace>\<lambda>s. P (not_recursive_ctes s)\<rbrace>"
+  assumes cteSwap_refs[wp]:
+    "\<And>c c1 c' c2.
+     \<lbrace>valid_global_refs' and cte_wp_at' (weak_derived' c \<circ> cteCap) c1
+      and cte_wp_at' (weak_derived' c' \<circ> cteCap) c2\<rbrace>
+     cteSwap c c1 c' c2
+     \<lbrace>\<lambda>rv. valid_global_refs'\<rbrace>"
```
corlewis: Possibly generalisable

proof/refine/CNodeInv_R.thy
```diff
@@ -155,31 +155,257 @@ lemma cnode_invok_case_cleanup:
...
+  assumes sameRegionAs_Zombie[simp]:
+    "\<And>p zb n cap. \<not> sameRegionAs (Zombie p zb n) cap"
```
corlewis, line 255: This and other `sameRegionAs` lemmas feel like they should be generalisable but are probably more work than is worth it to refactor the proofs.

proof/refine/CNodeInv_R.thy
corlewis: Ah, makes more sense now that I've gotten to `finalise_prop_stuff`. The name still feels a bit off though.

proof/refine/CNodeInv_R.thy
```diff
@@ -6134,51 +5955,63 @@ lemma updateCap_sch_act_simple[wp]:
...
+crunch postCapDeletion
+  for no_cte_prop[wp]: "no_cte_prop P"
+  (ignore: doMachineOp wp: dmo_maskInterrupt_no_cte_prop arch_postCapDeletion_no_cte_prop)
 
 crunch emptySlot, capSwapForDelete
   for no_cte_prop[wp]: "no_cte_prop P"
   (ignore: doMachineOp wp: dmo_maskInterrupt_no_cte_prop)
 
+lemma finalise_prop_stuff_archD:
+  "finalise_prop_stuff Q \<Longrightarrow> arch_finalise_prop_stuff Q"
```
corlewis, line 6025: Could be moved earlier and marked `[elim]`

proof/refine/CNodeInv_R.thy
```diff
@@ -6249,16 +6082,10 @@ lemma reduceZombie_invs'':
      apply clarsimp
     apply clarsimp
    apply clarsimp
-  apply (clarsimp dest!: isCapDs)
-  apply (fastforce dest!: ex_Zombie_to2 simp: cte_level_bits_def objBits_defs)
+  apply (clarsimp dest!: isCapDs) (* doesn't combine with fastforce below *)
```
corlewis, line 6098: Useful comment, thank you!

proof/refine/AARCH64/ArchCSpace1_R.thy
lsf37: We could, but given that it's a single `by` I don't think it's really worth it

proof/refine/AARCH64/ArchCNodeInv_R.thy
```diff
@@ -704,7661 +76,383 @@ lemma in_preempt':
...
-    }
-    ultimately show ?thesis by (cases "p = ptr") auto
-  qed
-  thus ?thesis using Some
-    by (simp add: modify_map_def)
-qed
+lemma cap_relation_trans:
+  "\<lbrakk> cap_relation cap cap'; cap_relation cap cap'' \<rbrakk>
+   \<Longrightarrow> cap' = cap''"
+  by (clarsimp split: cap_relation_split_asm arch_cap.split_asm)
```
corlewis: Would this be useful generically and earlier? If so, I guess it would need an interface lemma for `acap_relation`.

proof/refine/RISCV64/ArchCNodeInv_R.thy
```diff
@@ -295,51 +280,15 @@ crunch prepareThreadDelete, Arch_finaliseCap
    rule: RISCV64_H.finaliseCap_def)
 
 lemma arch_recycleCap_improve_cases:
-  "\<lbrakk> \<not>isFrameCap cap; \<not>isPageTableCap cap; \<not>isVCPUCap cap; \<not>isASIDControlCap cap;
-     \<not>isSGISignalCap cap; \<not>isSMCCap cap \<rbrakk>
-   \<Longrightarrow> (if isASIDPoolCap cap then v else undefined) = v"
+   "\<lbrakk> \<not> isFrameCap cap; \<not> isPageTableCap cap;
+         \<not> isASIDControlCap cap \<rbrakk> \<Longrightarrow> (if isASIDPoolCap cap then v else undefined) = v"
```
corlewis: Might as well fix this style while we're making the architectures more consistent.

proof/refine/X64/ArchCNodeInv_R.thy
```diff
@@ -15,8 +16,14 @@ context Arch begin arch_global_naming
 
 named_theorems CNodeInv_R_assms
 
+definition "ioport_state_independent_H (P :: kernel_state \<Rightarrow> bool)\<equiv>
+              \<forall>(s :: kernel_state) f. P s \<longrightarrow> P (s\<lparr>ksArchState := ksArchState s
+                                \<lparr>x64KSAllocatedIOPorts := f (x64KSAllocatedIOPorts (ksArchState s))\<rparr>\<rparr>)"
```
corlewis: It's old but it would be nice to fix this style.

proof/refine/RISCV64/ArchCNodeInv_R.thy
corlewis: Same on X64

corlewis (approved):
Very nice as always! Feel free to say that any of the potential lemmas I've pointed out are too difficult to generalise, it's hard to tell how much work they'll be just from the text.

proof/refine/CNodeInv_R.thy
```diff
@@ -2706,10 +2792,12 @@ lemma (in mdb_swap) n_dest:
...
   assumes md: "m (mdbPrev dest_node) = Some cte"
   shows "\<exists>cte'. n (mdbPrev dest_node) = Some cte'
-  \<and> mdbNext (cteMDBNode cte') = (if dest = mdbNext src_node then mdbNext dest_node else src)
-  \<and> mdbPrev (cteMDBNode cte') =
-  (if (mdbNext src_node = mdbPrev dest_node \<or> dest = mdbNext src_node) then dest else
-  mdbPrev (cteMDBNode cte))"
+                \<and> mdbNext (cteMDBNode cte')
+                  = (if dest = mdbNext src_node then mdbNext dest_node else src)
+                \<and> mdbPrev (cteMDBNode cte')
+                  = (if (mdbNext src_node = mdbPrev dest_node \<or> dest = mdbNext src_node)
```
lsf37: very optional: could leave out parentheses for if condition

lsf37: (a few more of these below)

proof/refine/ARM/Syscall_R.thy
```diff
@@ -262,7 +262,7 @@ lemma decodeInvocation_corres:
                     split del: if_split cong: if_cong)
         apply (clarsimp simp add: o_def)
         apply (rule corres_guard_imp)
-          apply (rule_tac F="length list \<le> 32" in corres_gen_asm)
+          apply (rule_tac F="length list \<le> word_bits" in corres_gen_asm)
```
lsf37, line 265: nice to see these gradually disappear

proof/refine/CNodeInv_R.thy
lsf37: Yeah, true. If someone has a flash of inspiration for a better name, it would be welcome.

lsf37 (approved):
Excellent! I did not dare hope that it would go this well. Very nice interface work.

Review-wise, as usual, Corey is too fast for me :-), I don't have anything deep in addition to what is already commented on. I have added minor comments to some of the existing ones.

proof/refine/AARCH64/ArchCSpace_I.thy
Xaphiosis: We end up with this:
```
lemma arch_maskCapRights_allRights[CSpace_I_assms, simp]:
  "Arch.maskCapRights allRights c = ArchObjectCap c"
  unfolding isCap_defs allRights_def maskCapRights_def maskVMRights_def
  by (cases c) (simp_all split: arch_capability.split vmrights.split)

lemma maskCapRights_allRights[simp]:
  "maskCapRights allRights c = c"
  by (cases c;
      clarsimp simp: global.maskCapRights_def Let_def gen_isCap_simps;
      clarsimp simp: allRights_def)
```

which is twice as long and doesn't really give us anything. Should I still do it, or leave it?

proof/refine/AARCH64/ArchCSpace_R.thy
Xaphiosis: This should work as `(* same proof on all architectures *)`

proof/refine/AARCH64/ArchFinalise_R.thy
Xaphiosis: mostly done, need to make sure it works on all arches again

proof/refine/CNodeInv_R.thy
Xaphiosis: Will rename to `hasCancelSendRights_isEndpointCap`

proof/refine/CNodeInv_R.thy
Xaphiosis: I honestly have no idea what it's for, so I just moved it and added the arch part.

proof/refine/CNodeInv_R.thy
Xaphiosis: `sameRegionAs` is a thorn in the foot of arch-split. It feels like there should be nice generic properties about it, but it's definition isn't generic. Maybe on pass two I'll get a better handle on it.

proof/refine/CNodeInv_R.thy
Xaphiosis: It's used exactly once, and there the elim doesn't help, so I'll leave it

proof/refine/CNodeInv_R.thy
Xaphiosis: Very good eye! I thought the unfolding of the `global_refs'` meant that this had to stay in Arch, but it turns out that it didn't do anything and could be removed. In next push it'll be fully generic.

proof/refine/CNodeInv_R.thy
Xaphiosis: If `data` is always machine_word, then `data_to_16` is the same ucast. I guess I can requalify that.

proof/refine/CNodeInv_R.thy
Xaphiosis: definition... will fix

proof/refine/AARCH64/ArchCNodeInv_R.thy
Xaphiosis: I don't think we'll ever see an acap_relation trans in the wild, and it's a one-line proof. I'd suggest leaving it

proof/refine/RISCV64/ArchCNodeInv_R.thy
Xaphiosis: will do

proof/refine/CNodeInv_R.thy
Xaphiosis: should be done

proof/refine/AARCH64/ArchCSpace1_R.thy
Xaphiosis: The trouble here is that this does not reduce nicely into a form that is helpful. i.e. trying to make it generic, the remaining goal is:
```
       ⟦cap = capability.ArchObjectCap x8; Arch.maskCapRights R x8 = capability.ArchObjectCap x8a⟧
       ⟹ (is_aligned (Arch.capUntypedPtr x8a) (acapBits x8a) ∧ acapBits x8a < word_bits ∧ valid_arch_cap' x8a s) =
         (is_aligned (Arch.capUntypedPtr x8) (acapBits x8) ∧ acapBits x8 < word_bits ∧ valid_arch_cap' x8 s)
```

Which means that one `by` becomes:

```
lemma arch_maskCap_valid[CSpace1_R_assms]:
  "s ⊢' Arch.maskCapRights R acap = s ⊢' ArchObjectCap acap"
  by (clarsimp simp: valid_cap'_def isCap_simps
                     capAligned_def AARCH64_H.maskCapRights_def
              split: arch_capability.split)

end

lemma maskCap_valid[simp]:
  "s ⊢' global.maskCapRights R cap = s ⊢' cap"
  apply (cases cap;
         clarsimp simp: valid_cap'_def gen_isCap_simps global.maskCapRights_def capAligned_def
         AARCH64.non_arch_maskCapRights split: capability.split)
  apply (rename_tac acap acap')
  apply (cut_tac acap=acap and R=R in AARCH64.arch_maskCap_valid)
  apply (auto simp: AARCH64.arch_maskCap_valid valid_cap'_def capAligned_def)
  done
```

and on top of that I need to backport `non_arch_maskCapRights` and `isArchObjectCap_maskCapRights` from Ipc_R. So I would be inclined to leave this one as is, unless one of you two see something I don't.

proof/refine/AARCH64/ArchCSpace1_R.thy
Xaphiosis: After two hours, the answer is that I don't know if there's a way to do this.
The issue is that we end up with `isMDBParentOf cte' (cteCap_update (λ_. cap) cte)` as the conclusion for arbitrary `cte'`. If we could show that `cte'` is not an arch cap, then there exists a simpler formulation of `isMDBParentOf`. But if we don't know, then we need to handle the SGI cap case on platforms that have it. The SGI case is super annoying because it only checks the left hand side and allows an arbitrary cap type on the right. So we can't reason backwards from knowing that `cap` is not an arch cap to knowing `cteCap cte'` is not an arch cap.
If the SGI cap case of `isMDBParentOf` required both sides to be SGI caps, or at least required the right side to be an arch cap as well, then I could reason backwards and say well if the right side is not an arch cap, then the left isn't an arch cap, and we can use nicer formulations. Maybe. @lsf37 thoughts on this?

Unfortunately, I don't think this PR is the place to do it at any rate. There's also a bunch of `sameRegionAs_def3` and `sameRegionAsE` in this proof, which might still pose a problem even if the above `isMDBParentOf` was addressed.

proof/refine/AARCH64/ArchCSpace1_R.thy
Xaphiosis: missed this one due to accidentally commenting here with the wrong thing, will take a look

Xaphiosis:
Hopefully addressed (or replied to) all PR comments now and pushed 3 extra commits (also rebased). Let me know if that resolves matters.

proof/refine/AARCH64/ArchCSpace1_R.thy
corlewis: Definitely sounds best to leave, thanks for investigating so thoroughly.

proof/refine/AARCH64/ArchCSpace_I.thy
corlewis: Right, that doesn't seem like it would help us much then, definitely reasonable to leave.

proof/refine/CNodeInv_R.thy
corlewis: Yeah, the definitions for `sameRegionAs` do seem confusing, with some being generic and others not.

proof/refine/AARCH64/ArchCNodeInv_R.thy
corlewis: I was more thinking about this lemma only appearing inside `Arch`. I guess it currently isn't used in any generic proofs but it feels to me like it should at least be accessible to them.

The simpler alternative I didn't think of before was that this lemma itself could be added to an interface, instead of splitting off an `acap_relation` part.

Xaphiosis: Oh. I was assuming this whole time that it was an interface lemma. But clearly my brain isn't working today. Yes, that's a good idea. I will make it an interface lemma.

Xaphiosis:
Squashed commits. Will merge when tests pass.

PR #998, opened 2026-05-21
run_tests: abnormal exit codes, adjust spacing

Xaphiosis:
When running long sessions using run_tests, abnormal failure (such as a Poly/ML segmentation fault) looked identical to normal failure (such as a proof not working), and you wouldn't find out which it was until all the other tasks were done (which could be an hour later).

We now print the process exit code as `[exit N]` for FAILED results when the exit code differs from normal Isabelle failure.

Here is what a normal run with normal failure looks like now:
```
Testing for L4V_ARCH='ARM', L4V_FEATURES='', L4V_PLAT='', INPUT_NUM_DOMAINS=''
Setting L4V_ARCH_IS_ARM
Running 2 test(s)...
  Started isabelle ...
  Finished isabelle                  passed      (0:00:04 real, 0:00:09 cpu,  1.09GB)
  Started Lib ...
  Finished Lib                       FAILED *    (0:00:28 real, 0:03:00 cpu,  6.45GB)
```

🦆🦆🦆 I am honestly unsure whether the asterisk after non-passed states is of much use to people. `passed` shows up in lowercase and green, everything else in caps and red. Accounting for this adds two columns in all runs.

And here is an abnomal failure (killed):
```
Testing for L4V_ARCH='ARM', L4V_FEATURES='', L4V_PLAT='', INPUT_NUM_DOMAINS=''
Setting L4V_ARCH_IS_ARM
Running 2 test(s)...
  Started isabelle ...
  Finished isabelle                  passed      (0:00:03 real, 0:00:10 cpu,  1.04GB)
  Started Lib ...
  Finished Lib                       FAILED *    (0:00:10 real, 0:01:00 cpu,  6.15GB) [exit 143]
```

The `[exit 143]` is as short as I could make it and still be comprehensible. Other options were `[code 143]`, `[exit code 143]`, and my original `[143]` which people were not a fan of. On my terminal (set to 90 wide by default) the wrap is currently at `[exi`.
In my original version, the code appeared right after the proof state, e.g. `FAILED 143`, but people did not like that one (might mess with anyone parsing the output of run_tests?).
It is also not coloured in any way. Would it make sense to make it red in the same colour as `FAILURE` is? 🦆🦆🦆

Finally, space tuning. If we have an image/task that runs for over 10 hours and doesn't time out, column alignment will be the least of our problems. A before and after of a normal line that shows the <10h assumption:
```
  Finished isabelle                  passed      ( 0:00:03 real,  0:00:10 cpu,  1.06GB)

  Finished isabelle                  passed      (0:00:04 real, 0:00:09 cpu,  1.09GB)
```

Oh, and I checked to make sure nothing extra gets printed in other states:
```
  Finished Lib                       TIMEOUT *   (0:00:01 real, 0:00:03 cpu,  0.36GB)
```


Xaphiosis:
Not sure what to do about style failure, I blended in with what was there already.

lsf37:
> Not sure what to do about style failure, I blended in with what was there already.

The style failure is complaining about argument type spacing (the colon):
```diff
diff --git a/misc/regression/run_tests.py b/misc/regression/run_tests.py
index 187372bed..a4778774a 100755
--- a/misc/regression/run_tests.py
+++ b/misc/regression/run_tests.py
@@ -358,7 +358,7 @@ def run_test(test, status_queue, kill_switch,
 
     status_queue.put({'name': test.name,
                       'status': test_status[0],
-                      'returncode' : returncode,
+                      'returncode': returncode,
                       'output': output,
                       'real_time': datetime.datetime.now() - start_time,
                       'cpu_time': cpu_usage,
```

lsf37:
Thank you for adding that. From the code side and general output, this looks good to me.

The new spacing for time might be ambitious. While we do not expect any real time values to go over 10h, we are already at `8:20:50 cpu` for `X64` in CI, so 10h does seem possible in the future. Of course we could deal with that when it eventually happens (e.g. MCS/X64 could get there, or maybe if VT-X is added at some point in the future -- none of these are very soon).

For the asterisk: I'd be fine with removing it, it has in fact confused me in the past.

Colour: I'd slightly prefer it to be not red (just standard output colour). It will already stick out and the red may be hard to read on some terminal backgrounds.

lsf37 (approved):

corlewis (approved):
Thanks for doing this! I don't have strong opinions on the asterisk or colour but do have a slight preference for the reduced spacing and dealing with misaligned 10 hour tests if they happen.

Xaphiosis:
Hopefully addressed style issue (thanks @lsf37 , I was staring right at it and yet didn't see it somehow), and added an extra commit to remove the asterisk printing.

We now have lines that look like this:
```
  Finished isabelle                  passed    (0:00:03 real, 0:00:09 cpu,  1.04GB)
  Finished Lib                       TIMEOUT   (0:00:01 real, 0:00:02 cpu,  0.34GB)
  Finished Lib                       FAILED    (0:00:05 real, 0:00:16 cpu,  1.31GB) [exit 143]
```

At 90 columns it now wraps at `exit ` (after the space, so that `143]` is on next line).

lsf37:
This looks good from my side, and style check has passed as well now.

PR #999, opened 2026-05-26
Use `ccorres_exec_l_pre` in `ccorres_pre*` lemmas

michaelmcinerney:


corlewis (approved):
These look much nicer! Thanks for the bits of style clean-up along the way as well.

lsf37 (approved):

PR #1000, opened 2026-05-26
Prove `handleInvocation_ccorres`

michaelmcinerney:


michaelmcinerney:
Because this is targeting a branch other than `rt` or whatever, the tests aren't running correctly, so I've cancelled them. When the branch I've targeted has been merged, I can test again. This does all pass locally, I'm quite sure.

proof/crefine/RISCV64/Syscall_C.thy
```diff
@@ -1215,209 +1225,213 @@ lemma handleInvocation_ccorres:
...
-                      apply (clarsimp simp: word_less_nat_alt)
-                      apply (rule conjI)
-                       apply clarsimp
-                       apply (case_tac rvd, clarsimp simp: option_to_ptr_def option_to_0_def min_def n_msgRegisters_def)
-                       apply (clarsimp simp: option_to_0_def option_to_ptr_def)
+           apply (rule_tac xf'="\<lambda>s. (status_' s, current_extra_caps_' (globals s))"
+                       and ef'=fst and vf'=snd and es=errstate
+                        in ccorres_split_nothrow_case_sum)
+                apply (rule ccorres_call, rule lookupExtraCaps_ccorres, simp+)
+               apply (rule ceqv_tuple2, ceqv, ceqv)
```
lsf37, line 1257: This is just reindented, I think, but it made me see the `ceqv_tuple2` which is a bit weird. Could you double check if a single `ceqv` invocation works? It's possible that this a left-over from times where `ceqv` was still in development. Also entirely possible that the manual `xf'` instantiation above (also old) still makes this necessary, though.

lsf37 (approved):
One sorry closer..

Nice to see that the sendFaultIPC lemma now has a sane interface.

lsf37:
> Because this is targeting a branch other than `rt` or whatever, the tests aren't running correctly, so I've cancelled them. When the branch I've targeted has been merged, I can test again. This does all pass locally, I'm quite sure.

👍

We can re-run before we merge it all into `rt`.

proof/crefine/RISCV64/Syscall_C.thy
michaelmcinerney: Yep, just a single `ceqv` does work. I've updated it and will push either when I do further updates after more comments or when we're ready to merge.

proof/refine/RISCV64/Ipc_R.thy
```diff
@@ -2979,6 +2979,11 @@ lemma bound_sc_tcb_at_cross':
    \<Longrightarrow> bound_sc_tcb_at' P t s'"
   by (fastforce dest!: bound_sc_tcb_at_cross simp: pred_tcb_at'_def obj_at'_def opt_map_red)
 
+defs bound_sc_tcb_at_asrt_def:
+  "bound_sc_tcb_at_asrt flag t s \<equiv> flag \<longrightarrow> bound_sc_tcb_at' (\<lambda>a. \<exists>y. a = Some y) t s"
```
corlewis, line 2983: I've lost track, do we still want to use the `bound` input abbreviation for conditions like this?

corlewis (approved):
Very nice!

proof/refine/RISCV64/Ipc_R.thy
michaelmcinerney: I will never use `bound` ever again for the rest of my life.

corlewis: Haha, fair enough, it did seem to cause a lot more confusion for us all than it helped with. If we're consistent in using this form with exists then we already have the main benefit from it anyway.

lsf37: I'd also prefer not using `bound`

proof/crefine/RISCV64/Syscall_C.thy
michaelmcinerney: I must not have checked the entire proof or something, because just a single `ceqv` doesn't work. Should I just leave it with the `ceqv_tuple2, ceqv, ceqv` (which does work) or should I investigate further?

lsf37: It's fine, let's leave it as is. Looks like it is not just a left-over.

PR #1004, opened 2026-05-29
Prove `handleRecv_ccorres` and `handleSyscall_ccorres`

michaelmcinerney:


spec/haskell/src/SEL4/Model/StateData.lhs
```diff
@@ -578,6 +578,11 @@ An assert that will say that there is a TCB with active' thread state at the giv
 > active_tcb_at'_asrt :: PPtr TCB -> KernelState -> Bool
 > active_tcb_at'_asrt _ _ = True
 
+An assert that will say that there is a schedulable' TCB at the given pointer
```
lsf37: stray `'`
```suggestion
An assert that will say that there is a schedulable TCB at the given pointer
```

lsf37: oh, it's actual `schedulable'`. Hm. That's still going to look like a stray `'` in the future. Does the comment still make sense if we leave it out or does it just become wrong?

lsf37: Or we could put `"schedulable'`"` in the comment. Then it's clearly intentional.

michaelmcinerney: Yep, I'll go with that. Just to double check, though, you do mean the back tick things, right? Or quotation marks?

proof/crefine/RISCV64/Refine_C.thy
```diff
@@ -290,157 +292,327 @@ lemma handleUserLevelFault_ccorres:
...
 
 lemma handleSyscall_ccorres:
   "ccorres dc xfdc
-           (invs' and
-               sch_act_simple and ct_running' and
-               (\<lambda>s. ksSchedulerAction s = ResumeCurrentThread))
-           (UNIV \<inter> {s. syscall_' s = syscall_from_H sysc }) []
-           (callKernel (SyscallEvent sysc)) (Call handleSyscall_'proc)"
+     (invs' and sch_act_simple and ct_running'
+      and (\<lambda>s. ksSchedulerAction s = ResumeCurrentThread))
```
lsf37: This was there before, but does it make sense to have both `sch_act_simple` and `(λs. ksSchedulerAction s = ResumeCurrentThread)` in the precondition? I guess it's not contradictory, but usually I'd expect just the stronger one.

proof/crefine/RISCV64/Refine_C.thy
```diff
@@ -290,157 +292,327 @@ lemma handleUserLevelFault_ccorres:
...
+           apply (clarsimp simp: if_to_top_of_bind)
+           apply (rule ccorres_cond[where R=\<top>])
+             apply (clarsimp simp: to_bool_def split: if_splits)
+            prefer 2
+            apply (clarsimp simp: returnOk_def)
+            apply (rule ccorres_returnOk_skip[simplified returnOk_def comp_def])
+           apply (rule ccorres_Catch)
+           apply ccorres_rewrite
+           apply wpc
+                     apply (find_goal \<open>match premises in "sysc = SysSend" \<Rightarrow> succeed\<close>)
```
lsf37: This should also work and is a bit clearer to read:
```suggestion
                     apply (find_case "SysSend")
```

proof/crefine/RISCV64/Refine_C.thy
```diff
@@ -290,157 +292,327 @@ lemma handleUserLevelFault_ccorres:
...
+                         apply (rule ccorres_handlers_weaken2)
+                         apply (rule ccorres_call[where xf'=xfdc, OF checkInterrupt_ccorres]; simp)
+                        apply fastforce
+                       apply wpsimp
+                      apply (rule_tac Q'=invs' in hoare_post_impE_E_dc)
+                       apply wpsimp
+                      apply simp
+                     apply clarsimp
+                     apply (vcg exspec=handleInvocation_modifies)
+                    apply (find_goal \<open>match premises in "sysc = SysNBSend" \<Rightarrow> succeed\<close>)
```
lsf37:
```suggestion
                    apply (find_case "SysNBSend")
```

lsf37: (a few more below)

proof/crefine/RISCV64/Refine_C.thy
michaelmcinerney: Yeah I thought about changing this, but decided against it for some reason. I can definitely go with just the `ResumeCurrentThread` conjunct.

proof/crefine/RISCV64/Refine_C.thy
michaelmcinerney: Cool, didn't even know about that one.

lsf37: I like selecting the cases by name, that should be more stable than relying on goal order.

proof/crefine/RISCV64/Syscall_C.thy
```diff
@@ -1498,178 +1508,230 @@ lemma not_obj_at'_ntfn:
...
+         apply (ctac add: receiveIPC_ccorres)
+        apply (clarsimp simp: case_bool_If capFault_bindE liftE_bindE bind_bindE_assoc bind_assoc)
+        apply (rule ccorres_cond_both'[where Q=\<top> and Q'=\<top>])
+          apply clarsimp
+         \<comment> \<open>cap is a NotificationCap\<close>
+         apply (rule ccorres_rhs_assoc)+
+         apply csymbr
+         apply csymbr
+         apply csymbr
+         apply csymbr
```
lsf37: I'd prefer `(repeat 4 cysmbr)`. Or just `csymbr+` if it doesn't do too much.

lsf37 (approved):
0 sorries!!

proof/crefine/RISCV64/Refine_C.thy
lsf37: If it's messy for some reason somewhere down the track it's also fine to add a comment to curb our enthusiasm for the future.

spec/haskell/src/SEL4/Model/StateData.lhs
lsf37: This is a .lhs file, so it's going to be LaTeX text, which means I would actually use quotes, because I'm not sure what the backtick does to LaTeX.

michaelmcinerney:
I should mention for anyone else wanting to review that all of the changes in Refine are just for adding the asserts, and that a lot of the proofs are really just reindented after the asserts are handled.

proof/crefine/RISCV64/Refine_C.thy
michaelmcinerney: Actually, the guard here can be reduced all the way down to `invs'`. I've done this and weakened the guards of some other proofs in this theory file.

lsf37: Excellent, very nice!

proof/crefine/RISCV64/Refine_C.thy
```diff
@@ -290,157 +292,324 @@ lemma handleUserLevelFault_ccorres:
...
+                  apply (wpsimp wp: hoare_drop_imps)
+                 apply (vcg exspec=handleInvocation_modifies)
+                apply wpsimp
+               apply (find_case SysNBSendRecv)
+               apply (clarsimp simp: syscall_from_H_def syscall_defs)
+               apply (rule ccorres_cond_empty | rule ccorres_cond_univ)+
+               apply (clarsimp simp: liftE_bindE bind_assoc)
+               apply (rule ccorres_rhs_assoc)+
+               apply (ctac add: getNBSendRecvDest_ccorres)
+                 \<comment> \<open>isolate handleInvocation\<close>
```
Xaphiosis, line 487: comment above already says `isolate handleInvocation`, I'm confused

Xaphiosis: oh, the next comment also says the same thing

proof/crefine/RISCV64/Syscall_C.thy
```diff
@@ -1498,178 +1508,227 @@ lemma not_obj_at'_ntfn:
...
+   apply (rule ccorres_pre_getCurThread)
+   apply (ctac (no_vcg) add: getMessageInfo_ccorres)
+    apply (rule ccorres_pre_getCurThread)
+    apply (ctac add: capFaultOnFailure_ccorres[OF lookupCap_ccorres])
+       apply ccorres_rewrite
+       apply csymbr
+       apply (rule ccorres_cond_seq)
+       apply (rule ccorres_cond_both'[where Q=\<top> and Q'=\<top>])
+         apply clarsimp
+         apply (frule cap_get_tag_isCap(10))
```
Xaphiosis: nitpick: comment here would be good to indicate which case this is, my best guess would be Reply?

Xaphiosis: you can also try find_names to see if there's a specific name, rather than an indexed one

proof/crefine/RISCV64/Syscall_C.thy
```diff
@@ -1498,178 +1508,227 @@ lemma not_obj_at'_ntfn:
...
+               apply wp
+              apply (vcg exspec=handleFault_modifies)
+             apply vcg
+            apply (rule conseqPre, vcg)
+            apply (clarsimp simp: rf_sr_upd_safe)
+           apply vcg
+          apply (rule conseqPre, vcg)
+          apply (clarsimp simp: rf_sr_upd_safe)
+         apply (vcg exspec=handleFault_modifies)
+        \<comment> \<open>the cap is neither an EndpointCap nor a NotificationCap; throw\<close>
```
Xaphiosis, line 1719: thanks for these, very useful

Xaphiosis (approved):
Very nice. One proof where the comments are very confusing, and the `monads:` commit could maybe use some kind of `lib` prefix?

proof/crefine/RISCV64/Refine_C.thy
michaelmcinerney: These are in different cases.

proof/crefine/RISCV64/Syscall_C.thy
michaelmcinerney: There isn't another name for this. If we don't like the index so much, I could do a `fastforce dest: cap_get_tag_isCap`. Though I'm not sure whether that's better. Maybe it's "less informative", though the indexing could change further down the track...

proof/crefine/RISCV64/Refine_C.thy
Xaphiosis: Right, but there's no documentation that there are cases or what they might be

proof/crefine/RISCV64/Syscall_C.thy
Xaphiosis: Nah, if there isn't keep the 10 but put in a comment hint as to which one you mean, in case the numbering ever changes (e.g. when moving to a different arch)

michaelmcinerney:
> Very nice. One proof where the comments are very confusing, and the `monads:` commit could maybe use some kind of `lib` prefix?

We're inconsistent with this. Corey seems to use `lib/monads` but Gerwin has used both recently. My most recent commits for this use just `monads`. The commit message doc does mention `lib/monads` explicitly, so I suppose I should go with that.

proof/crefine/RISCV64/Refine_C.thy
michaelmcinerney: There are numerous `find_case` statements throughout this proof.

Xaphiosis: They don't jump out as nicely as comments, but I will concede the point.

proof/crefine/RISCV64/Syscall_C.thy
```diff
@@ -1498,178 +1508,227 @@ lemma not_obj_at'_ntfn:
...
-             apply (ctac  add: receiveSignal_ccorres)
+         apply (simp add: case_bool_If if_to_top_of_bind)
+         apply ccorres_rewrite
+         apply (rule ccorres_cond_seq)
+         apply (subst if_swap)
+         apply (rule ccorres_cond[where R=\<top>])
+           apply (frule ccap_relation_ep_helpers)
+            apply (fastforce dest: cap_get_tag_isCap)
+           apply (clarsimp simp: from_bool_def split: bool.splits)
+          \<comment> \<open>capCanReceive is false; update faults then call handleFault\<close>
```
corlewis, line 1591: Thanks for these comments, very helpful!

corlewis (approved):
Very nice once again!

corlewis:
> > Very nice. One proof where the comments are very confusing, and the `monads:` commit could maybe use some kind of `lib` prefix?
> 
> We're inconsistent with this. Corey seems to use `lib/monads` but Gerwin has used both recently. My most recent commits for this use just `monads`. The commit message doc does mention `lib/monads` explicitly, so I suppose I should go with that.

I've gone back and forth on that tag quite a few times and in more recent work have settled on just using `monads:` (or `monads/nondet:` if it's more specific). In my mind the monads directory is big enough now to be its own thing, and keeping the `lib` tag in it made commit titles very unwieldy.

michaelmcinerney:
> > > Very nice. One proof where the comments are very confusing, and the `monads:` commit could maybe use some kind of `lib` prefix?
> > 
> > 
> > We're inconsistent with this. Corey seems to use `lib/monads` but Gerwin has used both recently. My most recent commits for this use just `monads`. The commit message doc does mention `lib/monads` explicitly, so I suppose I should go with that.
> 
> I've gone back and forth on that tag quite a few times and in more recent work have settled on just using `monads:` (or `monads/nondet:` if it's more specific). In my mind the monads directory is big enough now to be its own thing, and keeping the `lib` tag in it made commit titles very unwieldy.

Oh dear I've just gone and made it `lib/monads`. I guess we should update the commit messages doc if we want to do something else.

lib/Monads/nondet/Nondet_Lemmas.thy
```diff
@@ -166,6 +166,10 @@ lemma liftE_handleE[simp]:
   "(liftE a <handle> b) = liftE a"
   unfolding handleE_def by simp
 
+lemma liftE_case_sum:
+  "liftE f >>= case_sum e g = f >>= g"
+  by (simp add:liftE_def)
```
corlewis, line 171: Actually, talking about the monads directory has reminded me, any chance you could add this to the equivalent place in `Monads/trace/Trace_Lemmas.thy`? It should almost definitely just work for the trace monad as well and we want to keep `nondet` and `trace` consistent as much as possible.

michaelmcinerney: These commits are just for the `rt` branch. I could do a similar thing for `master`, but the commit would be different because I'd be removing this lemma from CRefine in all architectures. Maybe I could add it for the trace monad when I do that, if we'd like. Thoughts?

corlewis: Sure, that works.

proof/crefine/RISCV64/Refine_C.thy
lsf37: That might be my fault: I suggested the `find_case` statements because they were shorter compared to the previous longer `find_goal` statements that did actually stand out. No sure what the best way is to do these. Comments stand out because they are syntax highlighted and the long statement stood out because it used a long string/cartouche that is also syntax highlighted. `find_case` reduces it to just the necessary part and no longer stands out. Would it make sense to start a convention that we want an empty line before a `find_case` statement or something like that?

PR #1005, opened 2026-06-02
arch-split Tcb_R

Xaphiosis:
🦆🦆🦆 Again, needed a number of needed interface changes in previous theories. Since people liked it before, and it makes my setup easier, I've isolated those to a separate commit for review (`adjust interfaces pre ...`), hopefully it'll help. Again, please review commit-by-commit.

**Instructions:**

In real life, I did the usual addition of Tcb_R, hierarchy update, arch-split and then update. However, since the arch-split part first splits AARCH64, then takes the AARCH64 version of Tcb_R and only fixes what broke, we ended up with a small(er) diff there (again).

- `[squash] ... adjust interfaces ...` - interface changes and improvements in imported theories
- updates in downstream theories: only crefine this time

Stats: 50 files changed, 2350 insertions(+), 12007 deletions(-)
Total: 9657 lines removed


proof/refine/AARCH64/ArchCSpace_R.thy
corlewis: Hard to tell from the text if this is generalisable but it is at least the same on all architectures.

proof/refine/AARCH64/ArchCSpace_R.thy
```diff
@@ -639,7 +639,7 @@ lemma ex_nonz_cap_not_global':
   apply (clarsimp simp: ctes_of_valid_cap')
   done
 
-lemma setupReplyMaster_invs'[wp]:
+lemma setupReplyMaster_invs'[CSpace_R_2_assms, wp]:
   "\<lbrace>invs' and tcb_at' t and ex_nonz_cap_to' t\<rbrace>
    setupReplyMaster t
    \<lbrace>\<lambda>rv. invs'\<rbrace>"
```
corlewis, line 618: Same for this lemma.

proof/refine/AARCH64/ArchTcbAcc_R.thy
```diff
@@ -362,46 +363,48 @@ lemma tcb_hyp_refs'_atcbVCPUPtr_eq:
   by (simp add: tcb_vcpu_refs'_def)
      (metis Times_eq_cancel2 empty_not_insert ex_in_conv option.set_cases set_empty_eq)
 
+crunch threadSet
+  for irqs_masked'[wp]: "irqs_masked'"
+  (rule: irqs_masked_lift)
```
corlewis: Feels like it should be able to be done generally?

proof/refine/AARCH64/ArchTcbAcc_R.thy
```diff
@@ -959,6 +959,7 @@ lemma asUser_invs[wp]:
   apply (wpsimp wp: threadSet_invs_trivial threadGet_wp)
   done
 
+
```
corlewis: Extra newline, here and in ARM_HYP

proof/refine/Tcb_R.thy
```diff
@@ -20,11 +41,90 @@ lemma asUser_setNextPC_corres:
...
+    "\<And>t ct.
+     corres dc (tcb_at t and pspace_aligned and pspace_distinct) (tcb_at' t and tcb_at' ct)
+       (arch_post_modify_registers ct t)
+       (asUser t $ postModifyRegisters ct t)"
+  assumes threadSet_state_hyp_refs_of':
+    "\<And>F t P.
+     \<lbrakk>\<And>tcb. tcb_hyp_refs' (tcbArch (F tcb)) = tcb_hyp_refs' (tcbArch tcb) \<rbrakk>
+     \<Longrightarrow> threadSet F t \<lbrace>\<lambda>s. P (state_hyp_refs_of' s)\<rbrace>"
+  assumes threadSet_tcbPriority_if_live_then_nonz_cap'[wp]:
+    "\<And>f t. threadSet (tcbPriority_update f) t \<lbrace>if_live_then_nonz_cap'\<rbrace>"
```
corlewis, line 74: Probably more work than wanted for now but I'm pretty sure that this could be generalised if we also made a general version of `threadSet_iflive'T` that had an assumption on `tcb_hyp_refs'` being unchanged. Maybe something for a FIXME?

proof/refine/Tcb_R.thy
```diff
@@ -20,11 +41,90 @@ lemma asUser_setNextPC_corres:
...
+  assumes arch_post_modify_registers_pspace_distinct:
+    "\<And>cur t. arch_post_modify_registers cur t \<lbrace>pspace_distinct :: det_state \<Rightarrow> bool\<rbrace>"
+  assumes prepare_thread_delete_pspace_distinct[wp]:
+    "\<And>t. prepare_thread_delete t \<lbrace>pspace_distinct :: det_state \<Rightarrow> bool\<rbrace>"
+  assumes prepare_thread_delete_pspace_aligned[wp]:
+    "\<And>t. prepare_thread_delete t \<lbrace>pspace_aligned :: det_state \<Rightarrow> bool\<rbrace>"
+  assumes arch_finalise_cap_pspace_distinct[wp]:
+    "\<And>acap b. arch_finalise_cap acap b \<lbrace>pspace_distinct :: det_state \<Rightarrow> bool\<rbrace>"
+  assumes arch_finalise_cap_pspace_aligned[wp]:
+    "\<And>acap b. arch_finalise_cap acap b \<lbrace>pspace_aligned :: det_state \<Rightarrow> bool\<rbrace>"
```
corlewis, line 89: Do we want a `FIXME: move to AInvs` on these?

proof/refine/Tcb_R.thy
```diff
@@ -20,11 +41,90 @@ lemma asUser_setNextPC_corres:
...
+    "\<And>p cap P. checkValidIPCBuffer p cap \<lbrace>P\<rbrace>"
+  assumes checkValidIPCBuffer_ArchObject_wp:
+    "\<And>p cap P.
+     \<lbrace>\<lambda>s. isArchObjectCap cap \<and> capBadge cap = None \<and> is_aligned p msg_align_bits \<longrightarrow> P s\<rbrace>
+     checkValidIPCBuffer p cap
+     \<lbrace>\<lambda>rv s. P s\<rbrace>,-"
+  assumes isValidVTableRoot_eq:
+    "\<And>cap cap'. cap_relation cap cap' \<Longrightarrow> isValidVTableRoot cap' = is_valid_vtable_root cap"
+  assumes decodeTransfer_inv:
+    "\<And>w P. decodeTransfer w \<lbrace>P\<rbrace>"
```
corlewis: If we're already assuming `decodeTransfer_def'` can't this one be generalised?

proof/refine/Tcb_R.thy
```diff
@@ -465,37 +523,36 @@ proof -
...
 lemma readreg_invs':
   "\<lbrace>invs' and sch_act_simple and tcb_at' src and ex_nonz_cap_to' src\<rbrace>
-     invokeTCB (tcbinvocation.ReadRegisters src susp n arch)
+   invokeTCB (tcbinvocation.ReadRegisters src susp n arch)
    \<lbrace>\<lambda>rv. invs'\<rbrace>"
-  by (simp add: invokeTCB_def performTransfer_def | wp
-       | clarsimp simp: invs'_def valid_state'_def
+  unfolding invokeTCB_def performTransfer_def
+  by (simp | wp (* somehow much faster than wpsimp *)
+      | clarsimp simp: invs'_def valid_state'_def
```
corlewis, line 544: Thanks for the comment to save us from repeatedly trying this in the future.

proof/refine/Tcb_R.thy
```diff
@@ -794,29 +840,14 @@ lemma cte_map_tcb_0:
...
-    apply (fastforce simp: sameRegionAs_def isCap_simps split: arch_cap.splits)
-   subgoal by (fastforce simp: add_mask_fold sameRegionAs_def isCap_simps
-                         split: arch_cap.splits)
-  by (fastforce simp: sameRegionAs_def isCap_simps split: arch_cap.splits)
+  by (simp add: shiftl_t2n cte_map_def to_bl_1 tcb_cnode_index_def cteSizeBits_cte_level_bits)
+
+defs
+  assertDerived_def:
+  "assertDerived src cap f \<equiv>
+   do stateAssert (\<lambda>s. cte_wp_at' (is_derived' (ctes_of s) src cap o cteCap) src s) []; f od"
```
corlewis: Not new but would be nice to fix this style.

proof/refine/Tcb_R.thy
```diff
@@ -2170,7 +2073,9 @@ lemma decodeSetMCPriority_wf[wp]:
   apply (rule hoare_pre)
   apply (wp checkPrio_lt_ct_weak | wpc | simp | wp (once) checkPrio_inv)+
   apply (clarsimp simp: maxPriority_def numPriorities_def)
-  using max_word_max [of \<open>UCAST(64 \<rightarrow> 8) x\<close> for x]
+  (* FIXME arch-split: this leaks machine word size, but since maxPriority fits in 8 bits, this
+     works for all machine word sizes *)
+  using max_word_max [of \<open>UCAST(machine_word_len \<rightarrow> 8) x\<close> for x]
```
corlewis: Just checking, should this and the following one still be a FIXME? The comment makes it sound like it might be ok to keep it like this.

proof/refine/Tcb_R.thy
```diff
@@ -2645,11 +2531,13 @@ lemma decodeTCBInvocation_corres:
                                  (invs' and tcb_at' t and (\<lambda>s. \<forall>x \<in> set extras'. s \<turnstile>' fst x \<and> cte_at' (snd x) s))
          (decode_tcb_invocation label args c slot extras)
          (decodeTCBInvocation label args c' (cte_map slot) extras')"
+  (* FIXME arch-split: this should not be in the simpset! *)
+  supply tcb_bits_def[simp del]
```
corlewis, line 2541: Are you happy with this as a 'normal' FIXME or should we mark it as a more important one in some way?

corlewis (approved):
Like always, very nice to see so many lines removed! Just a few suggestions this time.

proof/refine/Tcb_R.thy
michaelmcinerney: If we wanted to remove `if_live_then_nonz_cap'` from `invs'`, as I did for MCS recently, then we could remove this assumption. I think there were a few other assumptions about this that I was able to remove from the locales for MCS. I think it simplified a fair few `invs'` rules and various other rules we have about `ex_nonz_cap_to'`, too. Maybe something to ponder later.

proof/refine/Tcb_R.thy
```diff
@@ -1514,58 +1431,54 @@ proof -
...
+                           arch_cap_fun_lift_def
                            invs_psp_aligned invs_distinct
+                           is_cnode_cap_is_cnode_or_valid_arch
                      cong: conj_cong imp_cong
                     split: option.split_asm)
-    by (clarsimp simp: invs'_def valid_state'_def valid_pspace'_def objBits_defs
-                       cte_map_tcb_0 cte_map_tcb_1[simplified] tcb_at_cte_at' cte_at_tcb_at_32'
-                       isCap_simps domIff valid_tcb'_def tcb_cte_cases_def
+    apply (clarsimp simp: invs'_def valid_state'_def valid_pspace'_def gen_objBits_simps
+                       cte_map_tcb_0 cte_map_tcb_1[simplified] tcb_at_cte_at'
```
michaelmcinerney: Nitpick: indentation is off here, now.

proof/refine/Tcb_R.thy
```diff
@@ -1580,9 +1493,32 @@ lemma cteInsert_sa_simple[wp]:
...
+  shows
+    "\<lbrace>invs' and K (\<forall>tcb. is_aligned (tcbIPCBuffer (F tcb)) msg_align_bits)\<rbrace>
+     threadSet F t
+     \<lbrace>\<lambda>_. invs'\<rbrace>"
+  by (rule hoare_gen_asm)
+     (wp threadSet_invs_trivialT; simp add: assms)
 
-lemma tc_invs':
+lemmas threadSet_invs_trivial2 =
+  threadSet_invs_trivialT2 [OF all_tcbI all_tcbI all_tcbI all_tcbI, OF ball_tcb_cte_casesI]
```
michaelmcinerney: Nitpick: would we like to have the square brackets without the space? I'm guessing this is old.

proof/refine/ARM/ArchTcbAcc_R.thy
```diff
@@ -338,46 +339,51 @@ lemma tcb_hyp_refs'_valid_arch_tcb'_eq[TcbAcc_R_2_assms]:
...
-            sym_heap_sched_pointers_lift threadSet_valid_sched_pointers
-            threadSet_tcbQueued
-            threadSet_tcbSchedPrevs_of threadSet_tcbSchedNexts_of valid_bitmaps_lift
-         | clarsimp simp: assms cteCaps_of_def valid_arch_tcb'_def | rule refl)+
-  apply (clarsimp simp: o_def)
-  by (auto simp: assms obj_at'_def)
+  supply tcb_hyp_refs_of'_simps[simp del]
+  apply (simp add: invs'_def valid_state'_def)
+  apply (rule hoare_pre)
+   apply (wp threadSet_valid_pspace'T
```
lsf37: Is there a reason for the `hoare_pre` or did this slip in from another architecture? If there is, we should probably add a comment

lsf37: The change is the same in the other architectures, so probably another source then.

proof/refine/Tcb_R.thy
lsf37: should indent by 2

proof/refine/Tcb_R.thy
```diff
@@ -2491,6 +2372,23 @@ lemma decodeSetSpace_tc_slot[wp]:
   apply simp
   done
 
+lemma isThreadControl_def2:
+  "isThreadControl tinv = (\<exists>a b c d e f g h. tinv = ThreadControl a b c d e f g h)"
+  by (cases tinv, simp_all add: isThreadControl_def)
+
+lemma decodeSetSpaceSome[wp]:
+  "\<lbrace>\<top>\<rbrace> decodeSetSpace xs cap y zs
+   \<lbrace>\<lambda>rv s. tcNewCRoot rv \<noteq> None\<rbrace>,-"
```
lsf37: it's old, but would benefit from style update (very optional)

proof/refine/Tcb_R.thy
lsf37: Would it make sense to have a lemma somewhere `8 < LENGTH(machine_word_len)` or something like that, which we then use for these? The numbers still leak, but the reason the proof holds would be encoded in one place, somewhere high up. Not sure it's worth it.

proof/refine/Tcb_R.thy
lsf37: I'm amazed how you spot these. I stared at the same thing and did not see it.

lsf37 (approved):
Cool, another 10k lines less. This was quite nice to review, too.

proof/refine/ARM/ArchTcb_R.thy
```diff
@@ -8,2763 +8,323 @@ theory ArchTcb_R
...
-               in hoare_post_imp, clarsimp)
-    apply (wp)+
-  apply assumption
-  done
-
-crunch setup_reply_master, Tcb_A.restart, arch_post_modify_registers
-  for pspace_aligned[wp]: "pspace_aligned :: det_ext state \<Rightarrow> _"
-  and pspace_distinct[wp]: "pspace_distinct :: det_ext state \<Rightarrow> _"
+crunch arch_post_modify_registers
+  for pspace_aligned[Tcb_R_assms, wp]: "pspace_aligned"
```
michaelmcinerney: Nitpick: we can leave the quotations marks off these.

proof/refine/ARM/ArchTcb_R.thy
```diff
@@ -8,2763 +8,323 @@ theory ArchTcb_R
...
-  "\<lbrakk> cap_relation c c'; cap_relation d d' \<rbrakk>
-     \<Longrightarrow> same_region_as c d = sameRegionAs c' d'"
-  by (erule(1) same_region_as_relation)
-
-lemma sameObject_corres2:
+    apply (clarsimp simp: arch_post_modify_registers_def postModifyRegisters_def when_def)
+    apply safe
+    apply (subst submonad_asUser.return)
+    apply (rule corres_stateAssert_assume)
+  by simp+
```
michaelmcinerney: I'm guessing this `by` needs to be indented more.

michaelmcinerney (approved):
Just a few nitpicks from me. I think that if I were more familiar with these proofs and the differences between the architectures, more would jump out at me. I might need a bit more first hand experience trying to get arch-split stuff working myself.

proof/refine/Tcb_R.thy
michaelmcinerney: I might be totally wrong here, but could we introduce a `priority_len` and `domain_len` and then define the types in terms of these? Then we'd have something more like `LENGTH(priority_len) < LENGTH(machine_word_len)` which might be a nicer requirement. 

I think Isabelle would still treat the types for domains and priorities exactly the same, which is a bit of a nuisance as far as wrangling domains and properties with, for instance, `spec` or other lemmas. I wonder if there is a way around that. It's a minor point, though.

proof/refine/AARCH64/ArchCSpace_R.thy
Xaphiosis: Yes! With some head-scratching this is a go.

proof/refine/AARCH64/ArchCSpace_R.thy
Xaphiosis: In principle probably yes, but it would require interfacing the lemmas for setupReplyMaster and valid_pspace', valid_global_refs', etc. I think it might also throw up a dependency problem but I'm not sure since it's a lot of work to re-interface this one. I'd like to leave it at this time. Would some kind of FIXME work for the next pass?

proof/refine/AARCH64/ArchTcbAcc_R.thy
Xaphiosis: Yes! It initially didn't work where I tried it. I bisected my way to where it works now. Cheers

proof/refine/Tcb_R.thy
Xaphiosis: I think this one is a FIXME, yes. Very impressive you found it. I don't know off the top of my head to rephrase it with tcb_hyp_refs'. Too many things at once. I'll add a FIXME above `threadSet_iflive'T`:
`(* FIXME arch-split: if this lemma statement can be phrased generically by using tcb_hyp_refs',
   then it can become an assumption, and make threadSet_tcbPriority_if_live_then_nonz_cap generic *)`

@michaelmcinerney Yeah, you cleared out several of those nicely. Not sure which ones needed infrastructure like adding assertions, and which ones could just be crossed. Maybe after first arch-split pass (or when a slot appears) we could backport some of the "just" ones.

proof/refine/Tcb_R.thy
Xaphiosis: I don't know. There's quite a bit of AInvs-related stuff in Refine, and we haven't pushed to move all those either. I can add a FIXME.

Xaphiosis: @lsf37 any thoughts on the wider picture?

proof/refine/Tcb_R.thy
Xaphiosis: Incredible, that's another one!

proof/refine/Tcb_R.thy
Xaphiosis: I indented by two, and changed `o` to `∘`, do you want me to multi-line it?

proof/refine/Tcb_R.thy
Xaphiosis: The FIXME was so we could talk about it, because I wasn't sure if this is the right move.
8 is one byte, but Michael is right, it's coincidentally priority/domain size.
What the proof spits out at this point is:
`UCAST(64 → 8) xb ≤ 0xFF ∧ UCAST(64 → 8) x ≤ 0xFF` if we don't unfold `maxPriority_def numPriorities_def` then we still get stuck with the ucast...

Adding priority_len and domain_len potentially sounds like something for after this round or as a separate PR, but I can make an issue or change the FIXME to say that. Further advice welcome...

proof/refine/Tcb_R.thy
Xaphiosis: I want to get to the end of the first pass of arch-split, and then review the FIXMEs to see what we can clean up. Potentially with some assistance, depending on available time. Doesn't the exclamation mark already make it important? :)
The issue is I can't reliably remove it from the simpset without going back to AInvs to never have added it in the first place, but then that'll break AInvs, etc.

proof/refine/ARM/ArchTcbAcc_R.thy
Xaphiosis: Nah, it's just old. Will fix.

proof/refine/Tcb_R.thy
Xaphiosis: yes, see my reply to Corey's comment also

proof/refine/Tcb_R.thy
Xaphiosis: will do

proof/refine/Tcb_R.thy
Xaphiosis: very. I also like it without the space. will fix

proof/refine/Tcb_R.thy
Xaphiosis: cheers

proof/refine/ARM/ArchTcb_R.thy
Xaphiosis: wasn't sure, the ones without quotes don't have the same colours... but will drop them

proof/refine/ARM/ArchTcb_R.thy
Xaphiosis: yup, and will add a done since it's not slow

Xaphiosis:
Pushed two commits that should address most of the comments (one for pre-Tcb_R, the other one for Tcb_R split).
I replied to the rest, let me know what's left to do.

proof/refine/AARCH64/ArchCSpace_R.thy
corlewis: I'm leaning a bit towards the optimistic side with my generalisation suggestions, so I'm always happy for you to leave it if it isn't easy. A FIXME works but also happy to skip that if you think there's fundamental difficulties.

proof/refine/Tcb_R.thy
lsf37: It's one of those things we are undecided on. On the side of moving it up would be consistency, separation of concerns and keeping similar things together. On the side of leaving it here is keeping lemmas close to where they are used, so that if they need to change you don't need to rebuild images, and the fact that it is easier to see if a lemma is used or not. Both sides have merit.

In this case, I think we should move them, because they are pretty basic properties (distinct/aligned) that probably fit in well in AInvs somewhere.

proof/refine/Tcb_R.thy
lsf37: Adding priority_len and domain_len to the FIXME sounds like a good idea. We can otherwise leave it for this PR, it's not like this one is actually dangerous, it's just not as nice as we'd like it.

proof/refine/Tcb_R.thy
corlewis: I don't think it needs to be multi-lined, but should the first two lines be combined into `defs assertDerived_def:`?

proof/refine/Tcb_R.thy
corlewis: Haha, I guess the exclamation mark works, since that's what made me stop and ask this question. And right, I hadn't realised that it comes from AInvs, that's a pain and definitely isn't something that we want to fix now.

proof/refine/Tcb_R.thy
corlewis: I was hesitant to make this comment for most of the mentioned reasons, but I think it was that last one about them being such basic properties that decided me to suggest it.

corlewis:
> Pushed two commits that should address most of the comments (one for pre-Tcb_R, the other one for Tcb_R split). I replied to the rest, let me know what's left to do.

Rereviewed, one very minor style suggestion and otherwise it looks to me like all comments have been addressed.

proof/refine/AARCH64/ArchCSpace_R.thy
Xaphiosis: Let's skip this one, since the interface for the lemmas to prove it would be bigger that this entire proof, and then they'd get used only once anyway.

proof/refine/Tcb_R.thy
Xaphiosis: I added `(* FIXME arch-split: consider moving several of these to AInvs *)`, will that be sufficient for this round?

proof/refine/Tcb_R.thy
Xaphiosis: ah, yes, will do

proof/refine/Tcb_R.thy
Xaphiosis: Adjusted FIXME:
```
  (* FIXME arch-split: this leaks machine word size, but since maxPriority fits in 8 bits, this
     works for all machine word sizes; we should add priority_len/domain_len to platform types
     and show elsewhere that they are smaller than machine_word_len *)
```

lsf37:
Also looks good from my side. I didn't spot a priority_len/domain_len addition to that ucast FIXME (maybe I missed it), but re-reading the current FIXME, it does actually give enough of a hint, so happy for you to merge. 

Edit: it just came in. All good :-)

Xaphiosis:
Cheers, then I will squash and merge this tomorrow morning unless anything else comes up.

Xaphiosis:
Squashed, will merge when tests pass.

PR #1006, opened 2026-06-02
refine: improve typ_at_lifts and add typ_at_props' locale

corlewis:
These changes are based on ones that were previously made in the `rt` branch. Some additional cleanup was done along the way and the `typ_at_props'` locale was arch-split and applied to all architectures.

The main benefit for master is the improved `typ_at_lifts` lemmas; the `typ_at_props'` locales themselves are not as immediately useful as in `rt`. However, backporting them does make the two branches more similar and should ease future merges.

See the following commits for where some of these changes originally happened in the `rt` branch:
6710b9a271a327aee25d8a1c2e95042524a3ba5a
64b6edf5143056cc1dc8a526370019b1e1d4752c
79e2e2b714017d91f202d5c918440fd8d110c08a
91a2023934bc549b9abd3e5dd35bab15e984f752
78bb847ddd78386d22d9eda29ade6807a767af71
8a01dd4b4edc2e7ae219e7dfc4aa146776d3177d
336bd0b1d63b3165f28cfef73dc4c70067fb35f3

### DIFF
--- a/proof/refine/AARCH64/ArchCNodeInv_R.thy
+++ b/proof/refine/AARCH64/ArchCNodeInv_R.thy
@@ -532,4 +532,23 @@ proof goal_cases
   case 1 show ?case by (intro_locales; (unfold_locales; (fact CNodeInv_R_assms)?)?)
 qed
 
+context Arch begin arch_global_naming
+
+sublocale cteDelete: typ_at_props' "cteDelete slot exposed"
+  by typ_at_props'
+
+sublocale reduceZombie: typ_at_props' "reduceZombie cap slot x"
+  by typ_at_props'
+
+sublocale finaliseSlot: typ_at_props' "finaliseSlot ptr exposed"
+  by typ_at_props'
+
+sublocale invokeCNode: typ_at_props' "invokeCNode i"
+  by typ_at_props'
+
+sublocale cteMove: typ_at_props' "cteMove cap src dest"
+  by typ_at_props'
+
+end  (* Arch *)
+
 end
--- a/proof/refine/AARCH64/ArchInvsLemmas_H.thy
+++ b/proof/refine/AARCH64/ArchInvsLemmas_H.thy
@@ -479,6 +479,25 @@ lemmas typ_at_lifts = gen_typ_at_lifts typ_at_lifts_strong_internal
 
 end
 
+end (* Arch *)
+
+locale typ_at_props' = Arch +
+  fixes f :: "_ kernel"
+  assumes typ': "f \<lbrace>\<lambda>s. P (typ_at' T p' s)\<rbrace>"
+begin
+
+lemmas typ_ats[wp] = typ_at_lifts[REPEAT [OF typ']]
+
+context begin
+(* We want to enforce that typ_ats only contains lemmas that have no
+   assumptions. The following thm statement should fail if this is not true. *)
+private lemmas check_valid_internal = iffD1[OF refl, where P="valid p g q" for p g q]
+thm typ_ats[atomized, THEN check_valid_internal]
+end
+
+end (* typ_at_props' *)
+
+context Arch begin arch_global_naming
 
 lemma valid_arch_state_armKSGlobalUserVSpace:
   "valid_arch_state' s \<Longrightarrow> canonical_address (addrFromKPPtr (armKSGlobalUserVSpace (ksArchState s)))"
--- a/proof/refine/AARCH64/ArchInvsLemmas_H.thy
+++ b/proof/refine/AARCH64/ArchInvsLemmas_H.thy
@@ -395,62 +395,90 @@ lemma is_physical_cases:
              | _                                \<Rightarrow> True)"
   by (simp split: capability.splits arch_capability.splits zombie_type.splits)
 
-lemma typ_at_lift_page_table_at':
-  assumes x: "\<And>T p. f \<lbrace>typ_at' T p\<rbrace>"
-  shows "f \<lbrace>page_table_at' pt_t p\<rbrace>"
+named_theorems Invariants_H_typ_at_lifts_assms
+
+lemma page_table_at'_typ_at_lift_strong:
+  "(\<And>p. f \<lbrace>\<lambda>s. P (typ_at' (ArchT PTET) p s)\<rbrace>) \<Longrightarrow> f \<lbrace>\<lambda>s. P (page_table_at' pt_t p s)\<rbrace>"
   unfolding page_table_at'_def
-  by (wpsimp wp: hoare_vcg_all_lift hoare_vcg_imp_lift' x)
+  apply (rule bool_to_bool_cases[where f=P]; clarsimp)
+  apply (wpsimp wp: hoare_vcg_const_Ball_lift hoare_vcg_imp_lift
+                    hoare_vcg_all_lift hoare_vcg_ex_lift
+         | assumption)+
+  done
 
-lemma typ_at_lift_asid_at':
-  "(\<And>T p. \<lbrace>typ_at' T p\<rbrace> f \<lbrace>\<lambda>_. typ_at' T p\<rbrace>) \<Longrightarrow> \<lbrace>asid_pool_at' p\<rbrace> f \<lbrace>\<lambda>_. asid_pool_at' p\<rbrace>"
-  by assumption
-
-lemma typ_at_lift_vcpu_at':
-  "(\<And>T p. \<lbrace>typ_at' T p\<rbrace> f \<lbrace>\<lambda>_. typ_at' T p\<rbrace>) \<Longrightarrow> \<lbrace>vcpu_at' p\<rbrace> f \<lbrace>\<lambda>_. vcpu_at' p\<rbrace>"
-  by assumption
-
-lemma typ_at_lift_frame_at':
-  assumes "\<And>T p. f \<lbrace>typ_at' T p\<rbrace>"
-  shows "f \<lbrace>frame_at' p sz d\<rbrace>"
+lemma frame_at'_typ_at_lift_strong:
+  "\<lbrakk>\<And>T p. f \<lbrace>\<lambda>s. P (typ_at' T p s)\<rbrace>\<rbrakk> \<Longrightarrow> f \<lbrace>\<lambda>s. P (frame_at' p sz d s)\<rbrace>"
+  supply if_split[split del]
   unfolding frame_at'_def
-  by (wpsimp wp: hoare_vcg_all_lift hoare_vcg_const_imp_lift assms split_del: if_split)
-
-lemma typ_at_lift_valid_cap':
-  assumes P: "\<And>P T p. \<lbrace>\<lambda>s. P (typ_at' T p s)\<rbrace> f \<lbrace>\<lambda>rv s. P (typ_at' T p s)\<rbrace>"
-  shows      "\<lbrace>\<lambda>s. valid_cap' cap s\<rbrace> f \<lbrace>\<lambda>rv s. valid_cap' cap s\<rbrace>"
-  including no_pre
-  apply (simp add: valid_cap'_def)
-  apply wp
-  apply (case_tac cap;
-         simp add: valid_cap'_def P[of id, simplified] typ_at_lift_tcb'
-                   hoare_vcg_prop typ_at_lift_ep'
-                   typ_at_lift_ntfn' typ_at_lift_cte_at'
-                   hoare_vcg_conj_lift [OF typ_at_lift_cte_at'])
-     apply (rename_tac zombie_type nat)
-     apply (case_tac zombie_type; simp)
-      apply (wp typ_at_lift_tcb' P hoare_vcg_all_lift typ_at_lift_cte')+
-    apply (rename_tac arch_capability)
-    apply (case_tac arch_capability,
-           simp_all add: P[of id, simplified] vspace_table_at'_defs
-                         hoare_vcg_prop All_less_Ball
-                    split del: if_split)
-       apply (wp hoare_vcg_const_Ball_lift P typ_at_lift_valid_untyped'
-                 hoare_vcg_all_lift typ_at_lift_cte' typ_at_lift_frame_at')+
+  apply (rule bool_to_bool_cases[where f=P]; clarsimp)
+   apply (wpsimp wp: hoare_vcg_all_lift hoare_vcg_const_imp_lift hoare_vcg_ex_lift
+          | assumption)+
   done
 
-(* interface lemma *)
-lemma valid_arch_tcb_lift':
-  assumes x: "\<And>T p. \<lbrace>typ_at' T p\<rbrace> f \<lbrace>\<lambda>rv. typ_at' T p\<rbrace>"
-  shows "\<lbrace>\<lambda>s. valid_arch_tcb' tcb s\<rbrace> f \<lbrace>\<lambda>rv s. valid_arch_tcb' tcb s\<rbrace>"
-  apply (clarsimp simp add: valid_arch_tcb'_def)
-  apply (cases "atcbVCPUPtr tcb"; simp)
-   apply (wp x)+
+lemma asid_pool_at'_typ_at_lift_strong:
+  "(\<And>T p. f \<lbrace>\<lambda>s. P (typ_at' T p s)\<rbrace>) \<Longrightarrow> f \<lbrace>\<lambda>s. P (asid_pool_at' p s)\<rbrace>"
+  by assumption
+
+lemma vcpu_at'_typ_at_lift_strong:
+  "(\<And>T p. f \<lbrace>\<lambda>s. P (typ_at' T p s)\<rbrace>) \<Longrightarrow> f \<lbrace>\<lambda>s. P (vcpu_at' p s)\<rbrace>"
+  by assumption
+
+lemma valid_arch_tcb'_typ_at_lift_strong[Invariants_H_typ_at_lifts_assms]:
+  "(\<And>T p. f \<lbrace>\<lambda>s. P (typ_at' T p s)\<rbrace>) \<Longrightarrow> f \<lbrace>\<lambda>s. P (valid_arch_tcb' tcb s)\<rbrace>"
+  unfolding valid_arch_tcb'_def
+  apply (rule bool_to_bool_cases[where f=P]; clarsimp)
+  apply (wpsimp wp: hoare_vcg_imp_lift hoare_vcg_all_lift hoare_vcg_ex_lift
+         | assumption)+
   done
 
-lemmas typ_at_lifts =
-           typ_at_lift_tcb' typ_at_lift_ep' typ_at_lift_ntfn' typ_at_lift_cte' typ_at_lift_cte_at'
-           typ_at_lift_valid_untyped' typ_at_lift_valid_cap' valid_bound_tcb_lift
-           typ_at_lift_page_table_at' typ_at_lift_asid_at' typ_at_lift_vcpu_at'
+lemma valid_arch_cap'_typ_at_lift[Invariants_H_typ_at_lifts_assms]:
+  assumes P: "\<And>P T p. f \<lbrace>\<lambda>s. P (typ_at' T p s)\<rbrace>"
+  shows      "f \<lbrace>\<lambda>s. valid_arch_cap' cap s\<rbrace>"
+  apply (case_tac cap,
+         simp_all add: P[where P=id, simplified] All_less_Ball
+            split del: if_split)
+    apply (wp hoare_vcg_const_Ball_lift P
+              page_table_at'_typ_at_lift_strong frame_at'_typ_at_lift_strong)+
+  done
+
+end (* Arch *)
+
+global_interpretation Invariants_H_typ_at_lifts?: Invariants_H_typ_at_lifts
+proof goal_cases
+  interpret Arch .
+  case 1 show ?case by (intro_locales; unfold_locales; (fact Invariants_H_typ_at_lifts_assms)?)
+qed
+
+context Arch begin arch_global_naming
+
+lemmas typ_at_lifts_strong =
+  gen_typ_at_lifts_strong
+  page_table_at'_typ_at_lift_strong
+  frame_at'_typ_at_lift_strong
+  asid_pool_at'_typ_at_lift_strong
+  vcpu_at'_typ_at_lift_strong
+
+context begin
+
+\<comment>\<open> See the comment above @{ML weaken_typ_at_lifts_thms} for an explanation of what
+    this @{command ML_goal} is for. Here we are constructing the combined list of
+    both generic and arch-specific typ_at_lifts lemmas.\<close>
+ML \<open>
+local
+  val strong_thms = subtract Thm.eq_thm gen_strong_thms @{thms typ_at_lifts_strong[no_vars]};
+in
+  val typ_at_lifts_internal_goals = weaken_typ_at_lifts_thms strong_thms
+end
+\<close>
+
+private ML_goal typ_at_lifts_strong_internal:
+  \<open>typ_at_lifts_internal_goals\<close>
+  by (auto simp: typ_at_lifts_strong)
+
+lemmas typ_at_lifts = gen_typ_at_lifts typ_at_lifts_strong_internal
+
+end
+
 
 lemma valid_arch_state_armKSGlobalUserVSpace:
   "valid_arch_state' s \<Longrightarrow> canonical_address (addrFromKPPtr (armKSGlobalUserVSpace (ksArchState s)))"
--- a/proof/refine/AARCH64/ArchIpc_R.thy
+++ b/proof/refine/AARCH64/ArchIpc_R.thy
@@ -260,6 +260,33 @@ lemma is_derived_mask'[simp]:
   "is_derived' m p (maskCapRights R c) = is_derived' m p c"
   by (rule ext, simp add: is_derived'_def badge_derived'_def)
 
+sublocale setExtraBadge: typ_at_props' "setExtraBadge buffer badge n"
+  by typ_at_props'
+
+sublocale transferCaps: typ_at_props' "transferCaps info caps endpoint receiver receiveBuffer"
+  by typ_at_props'
+
+sublocale copyMRs: typ_at_props' "copyMRs s sb r rb n"
+  by typ_at_props'
+
+sublocale doNormalTransfer: typ_at_props' "doNormalTransfer s sb e b g r rb"
+  by typ_at_props'
+
+sublocale doIPCTransfer: typ_at_props' "doIPCTransfer s e b g r"
+  by typ_at_props'
+
+sublocale handleFaultReply: typ_at_props' "handleFaultReply x t l m"
+  by typ_at_props'
+
+sublocale sendFaultIPC: typ_at_props' "sendFaultIPC tptr fault"
+  by typ_at_props'
+
+sublocale receiveIPC: typ_at_props' "receiveIPC t cap b"
+  by typ_at_props'
+
+sublocale receiveSignal: typ_at_props' "receiveSignal t cap b"
+  by typ_at_props'
+
 end (* Arch *)
 
 end
--- a/proof/refine/ARM/ArchInvsLemmas_H.thy
+++ b/proof/refine/ARM/ArchInvsLemmas_H.thy
@@ -405,6 +405,25 @@ lemmas typ_at_lifts = gen_typ_at_lifts typ_at_lifts_strong_internal
 
 end
 
+end (* Arch *)
+
+locale typ_at_props' = Arch +
+  fixes f :: "_ kernel"
+  assumes typ': "f \<lbrace>\<lambda>s. P (typ_at' T p' s)\<rbrace>"
+begin
+
+lemmas typ_ats[wp] = typ_at_lifts[REPEAT [OF typ']]
+
+context begin
+(* We want to enforce that typ_ats only contains lemmas that have no
+   assumptions. The following thm statement should fail if this is not true. *)
+private lemmas check_valid_internal = iffD1[OF refl, where P="valid p g q" for p g q]
+thm typ_ats[atomized, THEN check_valid_internal]
+end
+
+end (* typ_at_props' *)
+
+context Arch begin arch_global_naming
 
 lemmas bit_simps' = pteBits_def pdeBits_def asidHighBits_def asid_low_bits_def word_size_bits_def
                     asid_high_bits_def bit_simps
--- a/proof/refine/ARM/ArchInvsLemmas_H.thy
+++ b/proof/refine/ARM/ArchInvsLemmas_H.thy
@@ -329,58 +329,82 @@ lemma is_physical_cases:
              | _                                \<Rightarrow> True)"
   by (simp split: capability.splits arch_capability.splits zombie_type.splits)
 
-lemma typ_at_lift_page_directory_at':
-  assumes x: "\<And>T p. \<lbrace>typ_at' T p\<rbrace> f \<lbrace>\<lambda>rv. typ_at' T p\<rbrace>"
-  shows      "\<lbrace>page_directory_at' p\<rbrace> f \<lbrace>\<lambda>rv. page_directory_at' p\<rbrace>"
-  unfolding page_directory_at'_def All_less_Ball
-  by (wp hoare_vcg_const_Ball_lift x)
+named_theorems Invariants_H_typ_at_lifts_assms
 
-lemma typ_at_lift_page_table_at':
-  assumes x: "\<And>T p. \<lbrace>typ_at' T p\<rbrace> f \<lbrace>\<lambda>rv. typ_at' T p\<rbrace>"
-  shows      "\<lbrace>page_table_at' p\<rbrace> f \<lbrace>\<lambda>rv. page_table_at' p\<rbrace>"
-  unfolding page_table_at'_def All_less_Ball
-  by (wp hoare_vcg_const_Ball_lift x)
-
-lemma typ_at_lift_asid_at':
-  "(\<And>T p. \<lbrace>typ_at' T p\<rbrace> f \<lbrace>\<lambda>_. typ_at' T p\<rbrace>) \<Longrightarrow> \<lbrace>asid_pool_at' p\<rbrace> f \<lbrace>\<lambda>_. asid_pool_at' p\<rbrace>"
-  by assumption
-
-lemma typ_at_lift_valid_cap':
-  assumes P: "\<And>P T p. \<lbrace>\<lambda>s. P (typ_at' T p s)\<rbrace> f \<lbrace>\<lambda>rv s. P (typ_at' T p s)\<rbrace>"
-  shows      "\<lbrace>\<lambda>s. valid_cap' cap s\<rbrace> f \<lbrace>\<lambda>rv s. valid_cap' cap s\<rbrace>"
-  including no_pre
-  apply (simp add: valid_cap'_def)
-  apply wp
-  apply (case_tac cap;
-         simp add: valid_cap'_def P[of id, simplified] typ_at_lift_tcb'
-                   hoare_vcg_prop typ_at_lift_ep'
-                   typ_at_lift_ntfn' typ_at_lift_cte_at'
-                   hoare_vcg_conj_lift [OF typ_at_lift_cte_at'])
-     apply (rename_tac zombie_type nat)
-     apply (case_tac zombie_type; simp)
-      apply (wp typ_at_lift_tcb' P hoare_vcg_all_lift typ_at_lift_cte')+
-    apply (rename_tac arch_capability)
-    apply (case_tac arch_capability,
-           simp_all add: P[of id, simplified] vspace_table_at'_defs
-                         hoare_vcg_prop All_less_Ball
-                    split del: if_split)
-       apply (wp hoare_vcg_const_Ball_lift P typ_at_lift_valid_untyped'
-                 hoare_vcg_all_lift typ_at_lift_cte')+
+lemma page_directory_at'_typ_at_lift_strong:
+  "(\<And>p. f \<lbrace>\<lambda>s. P (typ_at' (ArchT PDET) p s)\<rbrace>) \<Longrightarrow> f \<lbrace>\<lambda>s. P (page_directory_at' p s)\<rbrace>"
+  unfolding page_directory_at'_def
+  apply (rule bool_to_bool_cases[where f=P]; clarsimp)
+  apply (wpsimp wp: hoare_vcg_const_Ball_lift hoare_vcg_imp_lift
+                    hoare_vcg_all_lift hoare_vcg_ex_lift
+         | assumption)+
   done
 
-(* interface lemma *)
-lemma valid_arch_tcb_lift':
-  assumes x: "\<And>T p. \<lbrace>typ_at' T p\<rbrace> f \<lbrace>\<lambda>rv. typ_at' T p\<rbrace>"
-  shows "\<lbrace>\<lambda>s. valid_arch_tcb' tcb s\<rbrace> f \<lbrace>\<lambda>rv s. valid_arch_tcb' tcb s\<rbrace>"
-  by (clarsimp simp add: valid_arch_tcb'_def, wp)
+lemma page_table_at'_typ_at_lift_strong:
+  "(\<And>p. f \<lbrace>\<lambda>s. P (typ_at' (ArchT PTET) p s)\<rbrace>) \<Longrightarrow> f \<lbrace>\<lambda>s. P (page_table_at' p s)\<rbrace>"
+  unfolding page_table_at'_def
+  apply (rule bool_to_bool_cases[where f=P]; clarsimp)
+  apply (wpsimp wp: hoare_vcg_const_Ball_lift hoare_vcg_imp_lift
+                    hoare_vcg_all_lift hoare_vcg_ex_lift
+         | assumption)+
+  done
 
-lemmas typ_at_lifts =
-           typ_at_lift_tcb' typ_at_lift_ep' typ_at_lift_ntfn' typ_at_lift_cte' typ_at_lift_cte_at'
-           typ_at_lift_valid_untyped' typ_at_lift_valid_cap' valid_bound_tcb_lift
-           typ_at_lift_page_table_at'
-           typ_at_lift_page_directory_at'
-           typ_at_lift_asid_at'
-           valid_arch_tcb_lift'
+lemma asid_at'_typ_at_lift_strong:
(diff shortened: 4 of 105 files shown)


proof/refine/Invariants_H.thy
```diff
@@ -2056,17 +2041,157 @@ lemma typ_at_lift_valid_untyped':
...
+  apply (wp hoare_vcg_all_lift P typ_at_lift_cte'_strong)
   done
 
-lemma valid_bound_tcb_lift:
-  "(\<And>T p. \<lbrace>typ_at' T p\<rbrace> f \<lbrace>\<lambda>_. typ_at' T p\<rbrace>) \<Longrightarrow>
-  \<lbrace>valid_bound_tcb' tcb\<rbrace> f \<lbrace>\<lambda>_. valid_bound_tcb' tcb\<rbrace>"
-  by (auto simp: valid_bound_tcb'_def valid_def typ_at_tcb'[symmetric] split: option.splits)
+(* use b_itself parameter to fix free type variable in interface lemmas *)
+locale typ_at_gen = Invariants_H_cte_ats +
+  fixes b_itself :: 'b
```
corlewis: This `fixes` was a pain but is necessary to be able to make this interface. The alternative would be to move all of these `gen` constructions to `InvariantUpdates_H`, after `ArchInvsLemmas_H` and where the needed arch lemmas could be requalified.

The parameter is generally resolved once the `typ_at_gen` locale is interpreted in `ArchInvsLemmas_H`, but the downside is that it appears in all future interpretations of the `gen_typ_at_props` locale as a parameter that can be ignored.

corlewis: See https://github.com/seL4/l4v/pull/1006/changes#diff-1e2b2cf051529198334e7ad45c0bebf1785e31a833e15cb37429a0610a6d82c0R1337 for an example of what I mean by the parameter appearing as something that can be ignored. Adding the `_` is very easy if you know about it or are copying an existing interpretation but could be confusing otherwise.

proof/refine/AARCH64/ArchInvsLemmas_H.thy
```diff
@@ -395,62 +395,130 @@ lemma is_physical_cases:
...
-           typ_at_lift_tcb' typ_at_lift_ep' typ_at_lift_ntfn' typ_at_lift_cte' typ_at_lift_cte_at'
-           typ_at_lift_valid_untyped' typ_at_lift_valid_cap' valid_bound_tcb_lift
-           typ_at_lift_page_table_at' typ_at_lift_asid_at' typ_at_lift_vcpu_at'
+end
+
+global_interpretation typ_at_gen?: typ_at_gen
+  proof goal_cases
+  interpret Arch .
+  case 1 show ?case by (intro_locales; unfold_locales; (fact typ_at_gen_assms)?)
+  qed
```
lsf37:
```suggestion
proof goal_cases
  interpret Arch .
  case 1 show ?case by (intro_locales; unfold_locales; (fact typ_at_gen_assms)?)
qed
```

proof/refine/Invariants_H.thy
lsf37: We might need a longer comment somewhere central that explains how to use the infrastructure (then this bit could go there). It's fairly clear now while we're focussing on it, but I can well imagine being confused later what sublocale I need to do when and which methods I should be using.

(the usual name for the kernel ext type would be `z` -- any particular reason you didn't pick that one?)

proof/refine/Invariants_H.thy
```diff
@@ -2056,17 +2041,157 @@ lemma typ_at_lift_valid_untyped':
...
+      val prems = List.map abstract_P (Thm.prems_of thm);
+      fun imp [] = Thm.concl_of thm
+        | imp (p :: pms) = @{const Pure.imp} $ p $ imp pms
+    in
+      imp prems
+    end
+in
+  val gen_typ_at_lifts_internal_goals = List.map abstract strong_thms
+end
+\<close>
```
lsf37, line 2124: The comment and the code are currently duplicated for the architectures. If we give `gen_typ_at_lifts_strong` as a parameter, the code as such could be generic in a separate theory file somewhere, together with the long comment, which we could then point to here.

proof/refine/ARM/ArchInvsLemmas_H.thy
```diff
@@ -329,58 +329,122 @@ lemma is_physical_cases:
...
-           typ_at_lift_page_directory_at'
-           typ_at_lift_asid_at'
-           valid_arch_tcb_lift'
+end
+
+global_interpretation typ_at_gen?: typ_at_gen
+  proof goal_cases
+  interpret Arch .
+  case 1 show ?case by (intro_locales; unfold_locales; (fact typ_at_gen_assms)?)
+  qed
```
lsf37:
```suggestion
proof goal_cases
  interpret Arch .
  case 1 show ?case by (intro_locales; unfold_locales; (fact typ_at_gen_assms)?)
qed
```

proof/refine/ARM/ArchInvsLemmas_H.thy
```diff
@@ -329,58 +329,122 @@ lemma is_physical_cases:
...
+
+private ML_goal typ_at_lifts_internal:
+  \<open>typ_at_lifts_internal_goals\<close>
+  by (auto simp: typ_at_lifts_strong)
+
+lemmas typ_at_lifts = gen_typ_at_lifts typ_at_lifts_internal
+
+end
+
+end
```
lsf37: end comments would be good on these ones (also in the other architectures)

proof/refine/ARM/ArchKHeap_R.thy
```diff
@@ -397,9 +395,14 @@ lemma setObject_ko_wp_at':
 
 lemmas [KHeap_R_assms_2] = setEndpoint_valid_arch' setNotification_valid_arch'
 
-lemmas doMachineOp_typ_ats[wp] = typ_at_lifts[OF doMachineOp_typ_at']
+sublocale setObject: typ_at_props' "setObject p v"
+  by typ_at_props'
```
lsf37, line 404: after looking at many of these, I'm starting to wonder if it would be worth adding a little ML command that only needs the constant name (or list of names) and does the rest. So that we could write

```
typ_at_props doMachineOp, setEndpoint
```

and it figures out parameters etc itself. Not for this PR, but could be a follow-up if we like the idea.

proof/refine/ARM_HYP/ArchInvsLemmas_H.thy
```diff
@@ -362,66 +362,130 @@ lemma is_physical_cases:
...
-           typ_at_lift_page_directory_at'
-           typ_at_lift_asid_at'
-           valid_arch_tcb_lift'
+end
+
+global_interpretation typ_at_gen?: typ_at_gen
+  proof goal_cases
+  interpret Arch .
+  case 1 show ?case by (intro_locales; unfold_locales; (fact typ_at_gen_assms)?)
+  qed
```
lsf37:
```suggestion
proof goal_cases
  interpret Arch .
  case 1 show ?case by (intro_locales; unfold_locales; (fact typ_at_gen_assms)?)
qed
```

proof/refine/RISCV64/ArchFinalise_R.thy
```diff
@@ -567,11 +568,17 @@ lemma isFinal_no_descendants[Finalise_R_2_assms]:
   apply (simp add: valid_mdb'_def)
   done
 
-lemmas cancelAllIPC_typs[wp] = typ_at_lifts[OF cancelAllIPC_typ_at']
-lemmas cancelAllSignals_typs[wp] = typ_at_lifts[OF cancelAllSignals_typ_at']
-lemmas suspend_typs[wp] = typ_at_lifts[OF suspend_typ_at']
+sublocale cancelIPC: typ_at_props' "cancelIPC tptr" by typ_at_props'
 
-lemmas finaliseCap_typ_ats[wp] = typ_at_lifts[OF finaliseCap_typ_at']
+sublocale cancelAllIPC: typ_at_props' "cancelAllIPC epptr" by typ_at_props'
```
lsf37: uses a different style to the rest. Should we one-line all of them?

proof/refine/RISCV64/ArchInvsLemmas_H.thy
```diff
@@ -366,55 +366,122 @@ lemma is_physical_cases:
...
-           typ_at_lift_tcb' typ_at_lift_ep' typ_at_lift_ntfn' typ_at_lift_cte' typ_at_lift_cte_at'
-           typ_at_lift_valid_untyped' typ_at_lift_valid_cap' valid_bound_tcb_lift
-           typ_at_lift_page_table_at' typ_at_lift_asid_at'
+end
+
+global_interpretation typ_at_gen?: typ_at_gen
+  proof goal_cases
+  interpret Arch .
+  case 1 show ?case by (intro_locales; unfold_locales; (fact typ_at_gen_assms)?)
+  qed
```
lsf37:
```suggestion
proof goal_cases
  interpret Arch .
  case 1 show ?case by (intro_locales; unfold_locales; (fact typ_at_gen_assms)?)
qed
```

proof/refine/X64/ArchFinalise_R.thy
```diff
@@ -595,11 +596,17 @@ lemma isFinal_no_descendants[Finalise_R_2_assms]:
   apply (simp add: valid_mdb'_def)
   done
 
-lemmas cancelAllIPC_typs[wp] = typ_at_lifts[OF cancelAllIPC_typ_at']
-lemmas cancelAllSignals_typs[wp] = typ_at_lifts[OF cancelAllSignals_typ_at']
-lemmas suspend_typs[wp] = typ_at_lifts[OF suspend_typ_at']
+sublocale cancelIPC: typ_at_props' "cancelIPC tptr" by typ_at_props'
```
lsf37: (different style here as well)

proof/refine/X64/ArchInvsLemmas_H.thy
```diff
@@ -400,72 +400,146 @@ lemma is_physical_cases:
...
+              typ_at_lift_pd_pointer_table_at' typ_at_lift_page_map_l4_at')+
+  done
+
+end
+
+global_interpretation typ_at_gen?: typ_at_gen
+  proof goal_cases
+  interpret Arch .
+  case 1 show ?case by (intro_locales; unfold_locales; (fact typ_at_gen_assms)?)
+  qed
```
lsf37:
```suggestion
proof goal_cases
  interpret Arch .
  case 1 show ?case by (intro_locales; unfold_locales; (fact typ_at_gen_assms)?)
qed
```

lsf37 (approved):
Thanks for doing that. The setup/instantiating is a bit more verbose, but proofs all become simpler. I think this is good. We might try to simplify making instances at some point, but I don't think we need to do that immediately.

One thing we should add is documentation. With the locale and arch split this has become complex enough that it should be explained somewhere in a reasonably central/findable location (what it does, why it exists, how to extend properties, when to do which kind of sublocale).

proof/refine/AARCH64/ArchInvsLemmas_H.thy
```diff
@@ -395,62 +395,130 @@ lemma is_physical_cases:
...
-  assumes x: "\<And>T p. \<lbrace>typ_at' T p\<rbrace> f \<lbrace>\<lambda>rv. typ_at' T p\<rbrace>"
-  shows "\<lbrace>\<lambda>s. valid_arch_tcb' tcb s\<rbrace> f \<lbrace>\<lambda>rv s. valid_arch_tcb' tcb s\<rbrace>"
-  apply (clarsimp simp add: valid_arch_tcb'_def)
-  apply (cases "atcbVCPUPtr tcb"; simp)
-   apply (wp x)+
+lemma typ_at_lift_valid_arch_cap'[typ_at_gen_assms]:
+  assumes P: "\<And>P T p. f \<lbrace>\<lambda>s. P (typ_at' T p s)\<rbrace>"
+  shows      "f \<lbrace>\<lambda>s. valid_arch_cap' cap s\<rbrace>"
+  apply (case_tac cap,
+         simp_all add: P [where P=id, simplified] All_less_Ball
```
Xaphiosis: while we're here, no space before `[`

proof/refine/AARCH64/ArchInvsLemmas_H.thy
```diff
@@ -395,62 +395,130 @@ lemma is_physical_cases:
...
+
+private ML_goal typ_at_lifts_internal:
+  \<open>typ_at_lifts_internal_goals\<close>
+  by (auto simp: typ_at_lifts_strong)
+
+lemmas typ_at_lifts = gen_typ_at_lifts typ_at_lifts_internal
+
+end
+
+end
```
Xaphiosis: consider `(* Arch *)` closing tags

proof/refine/AARCH64/ArchInvsLemmas_H.thy
```diff
@@ -395,62 +395,130 @@ lemma is_physical_cases:
...
+      imp prems
+    end
+in
+  val typ_at_lifts_internal_goals = List.map abstract strong_thms
+end
+\<close>
+
+private ML_goal typ_at_lifts_internal:
+  \<open>typ_at_lifts_internal_goals\<close>
+  by (auto simp: typ_at_lifts_strong)
```
Xaphiosis, line 476: I'm a bit lost here. Firstly, I'm wondering why this tech has to be duplicated across all the arches, along with the big comment, etc.
Second, I'm wondering, the deployment here only works *once*, right, to make the weaker form from the strong? So if someone wants to add more, they have to add both a strong and a weak? How does that work?

proof/refine/AARCH64/ArchInvsLemmas_H.thy
Xaphiosis: (and same for other arches)

proof/refine/Invariants_H.thy
Xaphiosis: Yeah, I think I commented the same. Definitely would be good to de-duplicate the comment and tech.

proof/refine/Invariants_H.thy
```diff
@@ -1939,81 +1939,66 @@ lemma locateSlot_conv:
                          split: zombie_type.split cong: option.case_cong)
   done
 
+context
+begin
+
+private method typ_at_proof =
```
Xaphiosis, line 1945: small comment, or add a docstring?

Xaphiosis: (there's no such thing as a truly private method, they seem to leak when you do `print_methods`)

proof/refine/Invariants_H.thy
```diff
@@ -2056,17 +2041,157 @@ lemma typ_at_lift_valid_untyped':
...
+
 lemma typ_at_lift_valid_irq_node':
-  assumes P: "\<And>P T p. \<lbrace>\<lambda>s. P (typ_at' T p s)\<rbrace> f \<lbrace>\<lambda>rv s. P (typ_at' T p s)\<rbrace>"
-  shows      "\<lbrace>valid_irq_node' p\<rbrace> f \<lbrace>\<lambda>_. valid_irq_node' p\<rbrace>"
+  assumes P: "\<And>P T p. f \<lbrace>\<lambda>s. P (typ_at' T p s)\<rbrace>"
+  shows      "f \<lbrace>valid_irq_node' p\<rbrace>"
   apply (simp add: valid_irq_node'_def)
-  apply (wp hoare_vcg_all_lift P typ_at_lift_cte')
+  apply (wp hoare_vcg_all_lift P typ_at_lift_cte'_strong)
   done
```
Xaphiosis, line 2068: why do all these proofs appear so much nicer than the Arch proofs that seem to go via `bool_to_bool_cases`? am I not comparing the right kinds of proofs?

proof/refine/Invariants_H.thy
Xaphiosis: I don't think this is the kernel ext type, the 'kernel' here is a monad?

These assumptions appear to be saying:
"given a monadic function with a return value of 'b, assume that you can do a typ_at' lift over valid_arch_tcb' (strong version, for some reason) and valid_arch_cap' (weak version, for some reason)"

Then for every function you want the typ_at' lifting tech for, you have to instantiate this locale via interpretation?
Also, why these specific assumptions?

Xaphiosis: What I'm also not understanding is how you'd get rid of the type parameter by moving things to InvariantUpdates_H. I'd've thought if you instantiate with a function, it would fill in the type parameter, so you would still need it, but maybe not `'b itself`?

proof/refine/Ipc_R.thy
```diff
@@ -3797,13 +3797,12 @@ lemma ri_invs' [wp]:
    apply (case_tac ep)
      \<comment> \<open>endpoint = RecvEP\<close>
      apply (simp add: invs'_def valid_state'_def)
-     apply (rule hoare_pre, wpc, wp valid_irq_node_lift valid_dom_schedule'_lift)
-      apply (simp add: valid_ep'_def)
-      apply (wp sts_sch_act' hoare_vcg_const_Ball_lift valid_irq_node_lift valid_dom_schedule'_lift
-                setThreadState_ct_not_inQ
-                asUser_urz
+     apply (rule hoare_pre, wpc, wp sts_sch_act' valid_irq_node_lift valid_dom_schedule'_lift)
```
Xaphiosis: while we're here (and below in this file where you're making changes), do we need the hoare_pre?

proof/refine/AARCH64/ArchCNodeInv_R.thy
```diff
@@ -533,4 +533,23 @@ proof goal_cases
...
+sublocale finaliseSlot: typ_at_props' "finaliseSlot ptr exposed"
+  by typ_at_props'
+
+sublocale invokeCNode: typ_at_props' "invokeCNode i"
+  by typ_at_props'
+
+sublocale cteMove: typ_at_props' "cteMove cap src dest"
+  by typ_at_props'
+
+end
```
Xaphiosis: end tags please. also, consistency, in some files you have empty lines in between, in others they run into each other

Also: while `context Arch` is cheaper than `interpretation Arch .`, it's still not free; any chance these could go into an Arch block above?

proof/refine/AARCH64/ArchIpc_R.thy
```diff
@@ -260,6 +260,25 @@ lemma is_derived_mask'[simp]:
...
+sublocale doIPCTransfer: typ_at_props' "doIPCTransfer s e b g r"
+  by typ_at_props'
+sublocale handleFaultReply: typ_at_props' "handleFaultReply x t l m"
+  by typ_at_props'
+sublocale sendFaultIPC: typ_at_props' "sendFaultIPC tptr fault"
+  by typ_at_props'
+sublocale receiveIPC: typ_at_props' "receiveIPC t cap b"
+  by typ_at_props'
+sublocale receiveSignal: typ_at_props' "receiveSignal t cap b"
+  by typ_at_props'
```
Xaphiosis, line 288: out of curiosity, how fast does a block like this run?
(also this is an example of no empty lines between)

clarification question: with sublocale in Arch here, you're saying that you only want the lift rules to be visible in Arch? that doesn't feel right for all of these

proof/refine/AARCH64/Tcb_R.thy
```diff
@@ -761,12 +761,15 @@ crunch setPriority, setMCPriority
   for typ_at'[wp]: "\<lambda>s. P (typ_at' T p s)"
   (simp: crunch_simps)
 
-lemmas setPriority_typ_ats [wp] = typ_at_lifts [OF setPriority_typ_at']
+end
```
Xaphiosis: another reminder to incorporate end tags into repertoire

proof/refine/ARM/ArchKHeap_R.thy
Xaphiosis: it would be convenient to use, but then we'd end up with the `param_a` and `param_b` that crunch does... then again this is a bit like crunching, so would be worth investigating at some point

proof/refine/RISCV64/ArchFinalise_R.thy
Xaphiosis: I'd say given how many there are, one-lining if it fits on one line is something I'd support

proof/refine/RISCV64/Syscall_R.thy
```diff
@@ -634,25 +634,32 @@ lemma invokeTCB_typ_at'[wp]:
...
+crunch restart, bindNotification, performTransfer, setFlags, postSetFlags, invokeTCB, doReplyTransfer,
+         performIRQControl, InterruptDecls_H.invokeIRQHandler, sendIPC, handleFault
   for typ_at'[wp]: "\<lambda>s. P (typ_at' T p s)"
+  (simp: crunch_simps
+   wp: crunch_wps checkCap_inv hoare_vcg_all_lift
+   ignore: checkCapAt)
 
-lemmas invokeIRQControl_typ_ats[wp] =
-  typ_at_lifts [OF performIRQControl_typ_at']
+end
```
Xaphiosis: This is why end tags are important, in the diff I have no idea what context the next sublocale commands are in.

proof/refine/Syscall_R.thy
```diff
@@ -634,25 +634,32 @@ lemma invokeTCB_typ_at'[wp]:
...
+sublocale Arch < performIRQControl: typ_at_props' "performIRQControl i"
+  by typ_at_props'
+sublocale Arch < arch_invokeIRQHandler: typ_at_props' "Arch.invokeIRQHandler i"
+  by typ_at_props'
+sublocale Arch < invokeIRQHandler: typ_at_props' "invokeIRQHandler i"
+  by typ_at_props'
+sublocale Arch < sendIPC: typ_at_props' "sendIPC bl call bdg cg cgr t' ep"
+  by typ_at_props'
+sublocale Arch < handleFault: typ_at_props' "handleFault t ex"
+  by typ_at_props'
```
Xaphiosis, line 839: Looking at this, and the fact this will probably hit before I get to Syscall arch-split (and maybe some others, I dunno), how does one go about arch-splitting these?

Also, `sublocale Arch` is something I try to avoid because it is slow. A whole block of them feels like it will be very slow. Can you let me know how long this runs for? Will the sublocale Arch go away as part of arch-splitting? This is new to me.

proof/refine/TcbAcc_R.thy
```diff
@@ -960,11 +960,11 @@ crunch threadSet
...
   for ksArchState[wp]: "\<lambda>s. P (ksArchState s)"
 
-lemma threadSet_typ_at'[wp]:
-  "\<lbrace>\<lambda>s. P (typ_at' T p s)\<rbrace> threadSet t F \<lbrace>\<lambda>rv s. P (typ_at' T p s)\<rbrace>"
-  by (simp add: threadSet_def, wp setObject_typ_at')
+crunch threadSet
+  for typ_at'[wp]: "\<lambda>s. P (typ_at' T p s)"
 
-lemmas threadSet_typ_at_lifts[wp] = gen_typ_at_lifts[OF threadSet_typ_at']
+global_interpretation threadSet: gen_typ_at_props' _ "threadSet tptr f"
```
Xaphiosis: Yeah, this is definitely confusing to see without the context of this particular PR.
See previous comments about whether we are sure we need the itself parameter. Maybe you should walk us through it... some things appear unintuitive, but have no alternative.

corlewis:
Thank you for the reviews and sorry for the delayed response, the comments led me to think of an improved way to structure and describe these changes. In particular, I've changed the commits to better reflect that there are two relatively separate changes going on here, both based on changes that have been made in the rt branch.

The first is improving and arch-splitting the typ_at_lift lemmas themselves. This involved doing a better job of centralising all of the relevant lemmas in one location, along with making some of the lemmas into 'strong' versions, which hold for any `P`. For technical reasons, which have an earlier attempt at an explanation in the big comment above the `ML_goal`, we additionally want slightly weaker versions of those 'strong' lemmas. An interface locale is also added to make it possible to prove the generic lemmas in the generic `Invariants_H` file. This is a relatively standard arch-split interface locale, with a slight complication to handle a free type variable in the lemmas being interfaced.

The second half of the changes is the addition and arch-split of the `typ_at_props'` locale. In the master branch this locale isn't actually that useful, it mostly just lets us interpret it for a given function to get the lifted typ_at lemmas, instead of doing commands like `lemmas archThreadSet_typ_ats[wp] = typ_at_lifts[OF archThreadSet_typ_at']`. In the rt branch where it was first developed, it is more useful to handle there being multiple sets of lemmas depending on whether they also depend on `sc_at'` being preserved.

The way this locale has been arch-split follows how the `typ_at_lifts` lemmas had previously been arch-split, there is a generic version containing only the generic lemmas and then arch-specific versions containing the total set, both arch-specific and generic lemmas. The generic locale can then be globally interpreted in generic files and the arch-specific locale is interpreted as a sublocale of `Arch`. Similarly to before with the `typ_at_lifts` lemmas, generic functions often have both a global interpretation in a generic file and an arch-specific one in the equivalent arch file.

proof/refine/Invariants_H.thy
corlewis: Sorry for the confusion, hopefully the new commit structure and big comment in this PR makes it a bit easier to understand.

This locale is the arch-split interface for the following `typ_at_lifts` lemmas. The assumptions are the arch-specific `typ_at_lift` lemmas that hold for all architectures and are needed to prove the generic ones.

Once the locale is interpreted in `ArchInvsLemmas_H` then we never need to think about it again, like with other arch-split interface locales.

corlewis: >> (the usual name for the kernel ext type would be `z` -- any particular reason you didn't pick that one?)

> I don't think this is the kernel ext type, the 'kernel' here is a monad?

@Xaphiosis is right, this is the free return type of the monads in the interface lemmas, not the kernel ext type. I fix it so that the generic lemmas in this locale can have the same type and make use of the interface lemmas.

However, it should maybe be `'a`. I think I only went with `'b` because `'a` was already used by something else in the goal when I was first working out what the problem was and how I could solve it.

corlewis: > What I'm also not understanding is how you'd get rid of the type parameter by moving things to InvariantUpdates_H. I'd've thought if you instantiate with a function, it would fill in the type parameter, so you would still need it, but maybe not `'b itself`?

I think your confusion here is due to mixing up this locale with the later one which gets instantiated with a function. In this case, what I was referring to is that instead of introducing this locale as an interface in `Invariants_H`, we could move the generic lemmas that depend on the arch-specific ones to `InvariantUpdates_H`, where those arch-specific lemmas could just be requalified.

proof/refine/Invariants_H.thy
corlewis: I've done this now, the comment and ML function are in the generic `Invariants_H` and pointed at from each `ArchInvsLemmas_H` file.

proof/refine/ARM/ArchKHeap_R.thy
corlewis: Yeah, while going through all of the files and doing these I definitely had the feeling that a command like that would be useful. @Xaphiosis is right though, it was annoying to have to work out sensible parameters for each function but it does end up with much nicer lemmas than crunch ones with `param_a` and `param_b`.

proof/refine/RISCV64/ArchFinalise_R.thy
corlewis: A lot of these were being copied across from the rt branch but yeah, I noticed all the different styles. At the time I thought it was better to be consistent with rt to make merges easier but now I'm not sure if we should just pick a style now and deal with the merge the one time.

proof/refine/Invariants_H.thy
corlewis: This was taken directly from rt. I guess I could try to come up with a comment, I think it's just wrapping up the duplicated proofs for the following `typ_ats'` lemmas into something that can be used for all of them.

corlewis: > (there's no such thing as a truly private method, they seem to leak when you do `print_methods`)

And yeah, I guess still useful though as an indication that it's not intended for general use.

proof/refine/AARCH64/ArchInvsLemmas_H.thy
corlewis: > Second, I'm wondering, the deployment here only works _once_, right, to make the weaker form from the strong? So if someone wants to add more, they have to add both a strong and a weak? How does that work?

Yes, this only works once, to make the weaker form of the specific 'strong' lemmas that were introduced here.

I'm not quite sure what you mean with the second half. If someone has new 'strong' lemmas then they could add them here easily by proving them and then adding them to `typ_at_lifts_strong`. If they proved them in a later file for whatever reason then they could still add them to the `typ_at_props'` locale, and construct the weaker version as well if needed.

proof/refine/Invariants_H.thy
corlewis: The proofs with `bool_to_bool_cases` are all lemmas I made into 'strong' versions. For these lemmas I was either unable to or didn't think to make 'strong', and the proofs could stay simpler.

proof/refine/Ipc_R.thy
corlewis: Probably not, will investigate

proof/refine/AARCH64/ArchCNodeInv_R.thy
corlewis: > Also: while `context Arch` is cheaper than `interpretation Arch .`, it's still not free; any chance these could go into an Arch block above?

Hmm, I have conflicted opinions here. I agree that moving them into an already existing Arch block would be much nicer, but doing that would almost definitely result in them being duplicated when this is merged into the rt branch.

proof/refine/AARCH64/ArchIpc_R.thy
corlewis: > out of curiosity, how fast does a block like this run? (also this is an example of no empty lines between)

These interpretations and sublocales are practically instant, they're below the threshold of my timing panel.

> clarification question: with sublocale in Arch here, you're saying that you only want the lift rules to be visible in Arch? that doesn't feel right for all of these

This is the larger locale that contains both generic and arch-specific lifted rules, and so can only be visible from Arch. There is normally an equivalent generic interpretation that contains only the generic rules and is visible globally. This approach follows my understanding of how `typ_at_lifts` was previously arch-split and handled.

proof/refine/Syscall_R.thy
corlewis: When these later files get arch-split my expectation is that in the generic file these cases will be replaced with a global_interpretation of the `gen_typ_at_props'` locale. We would then make the larger `typ_at_props'` a sublocale of `Arch` in the arch files. Again, I think this is basically the same as what would have been needed to arch-split the previous `typ_at_lifts` lemmas commands. I've also tried to mark each case of these with a `FIXME: arch-split` which will hopefully help.

The current `sublocale Arch` commands are a bit slow, but once these files are arch-split then the equivalents will almost definitely already be in `Arch` and so could just look like `sublocale handleFault: typ_at_props' "handleFault t ex"`, which is quick.

corlewis: When I rebase this PR I'm also going to have to go through this exact process for `Tcb_R`, so that will be a good test case to confirm that it makes sense.

proof/refine/Ipc_R.thy
corlewis: Removed the hoare_pre touched by this PR, although there's lots of other ones in this file that I've left for a future pass.

proof/refine/AARCH64/ArchCNodeInv_R.thy
corlewis: Ah, actually in this case I put them here because they need to go after the interpretation of `CNodeInv_R`. I don't see an easy way around this.

proof/refine/RISCV64/Syscall_R.thy
corlewis: This is in a file that hasn't been arch-split yet, so the context being ended is an anonymous one where we do `context begin interpretation Arch`. I'm not sure what end tag would be appropriate for that.

Once it is arch-split the context will almost definitely no longer be there, and these sublocale commands will be able to be replaced with global interpretations.

proof/refine/Syscall_R.thy
corlewis: Having now rebased and done this for `Tcb_R`, I can confirm that the above process is correct and is relatively straightforward.

One note for the future though, I impressively already forgot about the locale type parameter and it meaning that interpreting `gen_typ_at_props'` needs a `_`, which led to a very confusing failure. I'm not sure what to do about this, beyond adding a comment to the definition of `gen_typ_at_props'` for anyone who clicks through to there.

lsf37: A comment at whatever place you get to when you click though would be fairly helpful, I think. The other discovery mechanism would be `grep`, so it would be good if it mentioned `gen_typ_at_props'` somewhere.

corlewis:
This is now ready for re-review. I think I've resolved all suggestions that were made in both our online and offline discussions. Once we're happy with this state of it I'll squash, rebase, and update the remaining files that have since been arch-split.

proof/refine/RISCV64/ArchInvsLemmas_H.thy
```diff
@@ -445,22 +445,21 @@ end
...
 begin
 
 lemmas typ_ats[wp] = typ_at_lifts[REPEAT [OF typ']]
 
 context begin
 (* We want to enforce that typ_ats only contains lemmas that have no
-   assumptions. The following thm statements should fail if this is not true. *)
+   assumptions. The following thm statement should fail if this is not true. *)
 private lemmas check_valid_internal = iffD1[OF refl, where P="valid p g q" for p g q]
 thm typ_ats[atomized, THEN check_valid_internal]
```
corlewis, line 458: I tried turning this `thm` into a `lemmas _ = ...` but it pollutes the namespace and led to confusing `find_theorems` results, so I changed it back to `thm`.

corlewis:
> Once we're happy with this state of it I'll squash, rebase, and update the remaining files that have since been arch-split.

Proof failures are expected in the current state, CRefine won't work until that rebase happens.

proof/refine/RISCV64/ArchInvsLemmas_H.thy
Xaphiosis: Ah, thanks for trying it out anyway.

proof/refine/Invariants_H.thy
```diff
@@ -2041,47 +2041,51 @@ lemma typ_at_lift_valid_untyped':
...
-  fixes b_itself :: 'b
-  assumes valid_arch_tcb_lift'_strong:
-    "\<And>(f :: 'b kernel) P tcb. (\<And>T p. f \<lbrace>\<lambda>s. P (typ_at' T p s)\<rbrace>) \<Longrightarrow> f \<lbrace>\<lambda>s. P (valid_arch_tcb' tcb s)\<rbrace>"
-  assumes typ_at_lift_valid_arch_cap':
-    "\<And>(f :: 'b kernel) cap. (\<And>P T p. f \<lbrace>\<lambda>s. P (typ_at' T p s)\<rbrace>) \<Longrightarrow> f \<lbrace>valid_arch_cap' cap\<rbrace>"
+(* arch-interface locale for typ_at lifting lemmas *)
+locale Invariants_H_typ_at_lifts = Invariants_H_cte_ats +
+(* Use f_rvt_itself parameter to fix free type variable in interface lemmas.
+   This is the loosest constraint that we have come up with to make this locale work, and is
+   able to be used here because nothing within the locale looks at f. *)
```
Xaphiosis: shouldn't this be indented to the same level as `fixes` below?

Xaphiosis: do we care to say that rvt is the return type of f?

proof/refine/TcbAcc_R.thy
```diff
@@ -1289,7 +1284,7 @@ lemma setObject_valid_tcbs':
   apply (rename_tac s s' ptr' tcb)
   apply (prop_tac "\<forall>tcb'. valid_tcb' tcb s \<longrightarrow> valid_tcb' tcb s'")
    apply clarsimp
-   apply (erule (1) use_valid[OF _ setObject_valid_tcb'])
+   apply (erule (1) use_valid[OF _ setObject.gen_typ_ats(18)])
```
Xaphiosis: hmm. the (18) here and (14) in KHeap_R doesn't feel like an improvement, and doesn't indicate any longer that we want the tcb version. Either having a `lemmas` to bind it, or a comment to indicate which one it is in case the number changes (e.g. for port to MCS)

proof/refine/InvariantUpdates_H.thy
```diff
@@ -18,6 +18,11 @@ arch_requalify_facts
   objBitsT_koTypeOf (* FIXME arch-split: consider declaring [simp] here rather than Schedule_R *)
   hyp_refs_of_live'
 
+(* transfer facts from partial locale (with extra assumptions) into complete locale
+   (since we can now satisfy these assumptions) *)
```
Xaphiosis: this is true in a bit of a `i = i + 1; // add 1 to i` kind of way, but what were we missing to satisfy these assumptions? Arch proofs? something else? would be good to say that

proof/refine/Invariants_H.thy
```diff
@@ -2176,12 +2176,14 @@ end
 
 end (* typ_at_gen *)
 
-locale gen_typ_at_props' = typ_at_gen "b_itself :: 'b" for b_itself +
-  fixes f :: "'b kernel"
-  assumes typ': "f \<lbrace>\<lambda>s. P (typ_at' T p' s)\<rbrace>"
+(* Will be a sublocale of gen_typ_at_props' once Invariants_H_typ_at_lifts is interpreted.
+   Done this way to avoid b_itself parameter.*)
```
Xaphiosis: again, would be good to know what is missing here. the mention of b_itself might not be clear to someone not looking at the diff where `'b kernel` is mentioned; are we talking about the kernel ext type? if so, then that's worth mentioning

Xaphiosis (approved):
I'm very happy with the removal of the gen_typ_at_props' dummy parameter!

My remaining comments are related to increased clarity of comments for future historians.

proof/refine/TcbAcc_R.thy
lsf37: +1 from me on that, these numbers shifting can be pretty hard to debug. It's fine if it's in some sledgehammer result that can be found automatically again, but otherwise if we can we should avoid them or guard them.

lsf37 (approved):
Also very happy with the changes, I don't have anything beyond Raf's comments.

proof/refine/Invariants_H.thy
corlewis: Both good points, indenting it like that would make it extra clear what the comment is referring to and it's always nice to have a hint about what abbreviations are intended to be short for.

proof/refine/TcbAcc_R.thy
corlewis: Yeah, I was a bit hesitant to change to these numbers but also didn't like the ad-hoc `lemmas` being added for random lemmas out of the `typ_ats` set. There's a bit of a conflict here, with these proofs wanting to use the rules manually while the `typ_ats` are more constructed to be used automatically from `wp`. I didn't really want to try and rework these proofs now though.

I'll have a think to see if I can find a more systematic way of handling these.

corlewis: Ok, I have vague ideas for something nicer but that I think would be too much work for not enough benefit, so instead I've given these specific typ_at_lift lemmas actual names to start with and changed this line to
```
apply (erule (1) use_valid[OF _ valid_tcb'_typ_at_lift[OF setObject_typ_at']])
```
Not perfect but much more maintainable.

proof/refine/InvariantUpdates_H.thy
corlewis: Does this help and sound accurate?
```
(* transfer facts from partial locale (with extra assumptions) into complete locale
   (since we can satisfy these assumptions now that the Invariants_H_typ_at_lifts interface has
   been interpreted) *)
   ```

File: proof/refine/Invariants_H.thy (hunk 28 above)
Comment (corlewis) [reply proof/refine/Invariants_H.thy]: Whoops, thanks for spotting, I thought I had removed all references to b. This should be `f_rvt_itself`.

File: proof/refine/InvariantUpdates_H.thy (hunk 27 above)
Comment (Xaphiosis) [reply proof/refine/InvariantUpdates_H.thy]: I'm still a bit lost, I admit. Let me try again: I understand that we now have *something* from the interpretation of Invariants_H_typ_at_lifts, which allows us to satisfy gen_typ_at_props'_interface, but is there any coherent name that encompasses what that is that isn't just "assumptions"?

I was thinking it would be something like "After interpretation of Invariants_H_typ_at_lifts, The arch-specific assumptions of gen_typ_at_props'_interface are now satisfiable, meaning the typ_at properties can now be placed in a complete locale with no further type assumption" ... is that close to the truth? Does something like that sound better?

File: proof/refine/Invariants_H.thy [hunk 29]
```diff
@@ -2189,7 +2189,7 @@ end
 end (* typ_at_gen *)
 
 (* Will be a sublocale of gen_typ_at_props' once Invariants_H_typ_at_lifts is interpreted.
-   Done this way to avoid b_itself parameter.*)
+   Done this way to avoid f_rvt_itself parameter.*)
```
Xaphiosis: maybe add so it's "avoid f_rvt_itself parameter in final locale"?

proof/refine/TcbAcc_R.thy
Xaphiosis: Yes, that looks much better to me, thank you!

proof/refine/InvariantUpdates_H.thy
corlewis: I think our use of the word assumptions when we wrote this comment is making this seem more complicated than it actually is. I can't think of how to express this with the right terminology, but in my mind what we're working around is that `Invariants_H_typ_at_lifts` has a `fixes` that `gen_typ_at_props'_interface` has to inherit. Then, once `Invariants_H_typ_at_lifts` has been interpreted, we can resolve that `fixes` by moving everything into `gen_typ_at_props'`.

Xaphiosis: Oh, that sounds informative! What if the comment was something like:

"Invariants_H_typ_at_lifts has a fixes that gen_typ_at_props'_interface had to inherit. Now that Invariants_H_typ_at_lifts has been instantiated, we can drop that fixes assumption by moving everything into gen_typ_at_props'."

This explains a lot. I think previously you were kind of explaining more what sublocale does than what we were trying to do.

corlewis: Thanks, that reads well to me. I think between the comment here and the comments in Invariants_H where these locales are defined there's now enough for future us to work out what's going on here.

corlewis:
Squashed, rebased and new arch-split files updated. Small bits of new content in Syscall_R and Untyped_R.

lsf37:
Looks good and ready to merge from my side.

PR #1008, opened 2026-06-03
capDL style update

lsf37:
Restyle capDL spec according to current style guide.

- definition/function style
- indentation
- parentheses
- comment style
- minor adjustments to comment contents
- some (rare) changes might affect proofs:
  - changes to some bound names
  - removal/addition of $
  - change of = into ==
- no other content changes, in particular no change of if/case into when/maybeM or other monad idioms or other things that could be expressed much more nicely than they are. 

In a separate commit:
- remove imports that are already included transitively by other imports.

- very minor proof fixup in DRefine

Main problem is how to review this one. The diff is probably terrible. I might be best to open up the branch in Isabelle and look at the finished files if they are now in style.

### DIFF
--- a/spec/capDL/KHeap_D.thy
+++ b/spec/capDL/KHeap_D.thy
@@ -4,246 +4,182 @@
  * SPDX-License-Identifier: GPL-2.0-only
  *)
 
-(*
- * Accessor functions for objects and caps.
- *)
+(* Accessor functions for objects and caps. *)
 
 theory KHeap_D
-imports Monads_D
+  imports Monads_D
 begin
 
 (* Return an item from the heap. Fail if no such object exists. *)
-abbreviation
-  get_object :: "cdl_object_id \<Rightarrow> cdl_object k_monad"
-where
+abbreviation get_object :: "cdl_object_id \<Rightarrow> cdl_object k_monad" where
   "get_object p \<equiv> gets_the (\<lambda>s. cdl_objects s p)"
 
 (* Set an item on the heap to the given object. *)
-definition
-  set_object :: "cdl_object_id \<Rightarrow> cdl_object \<Rightarrow> unit k_monad"
-where
-  "set_object p obj \<equiv>
-    modify (\<lambda>s. s \<lparr> cdl_objects := (cdl_objects s) (p \<mapsto> obj) \<rparr> )"
+definition set_object :: "cdl_object_id \<Rightarrow> cdl_object \<Rightarrow> unit k_monad" where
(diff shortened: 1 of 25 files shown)


spec/capDL/KHeap_D.thy
```diff
@@ -4,246 +4,183 @@
...
+     slots \<leftarrow> return $ object_slots obj;
+     obj' \<leftarrow> return $ update_slots (slots (slot \<mapsto> cap)) obj;
+     obj'' \<leftarrow> case obj' of
+                Tcb t \<Rightarrow> if slot = tcb_ipcbuffer_slot \<and> slots slot \<noteq> Some cap
+                         then do
+                           intent' \<leftarrow> select UNIV;
+                           return $ Tcb (t \<lparr> cdl_tcb_intent := intent' \<rparr>)
+                         od
+                         else
+                           return obj'
```
lsf37: the still looks off

Xaphiosis: the only other way I see is to indent within the `do od`:
```
then do
       ...
       ...
     od
else
```
but I'm not sure it's that much better

Xaphiosis (approved):
Not sure what prompted this, but everything looks like it's an improvement, and I can't see any misses.
Did you do all of this by hand? O_o

Xaphiosis:
Oh, the proof update commit looks a bit... teensy. You sure you want to keep it separated?

lsf37:
> Oh, the proof update commit looks a bit... teensy. You sure you want to keep it separated?

I can squash it in. I was expecting it to be larger when I worked on it.

> Not sure what prompted this, but everything looks like it's an improvement, and I can't see any misses.

The deeper prompt was the capDL proofs recently and having to read this over and over again. The current prompt is that Corey needs to be make changes to the capDL spec for the new monads. So better to do this first.

> Did you do all of this by hand? O_o

I did. Tried to automate some of it and it was a complete failure.

spec/capDL/KHeap_D.thy
lsf37: I thought the indent was off, but that might just be GitHub printing. I think the `do` part is Ok. Maybe the last `return` can move up, because it doesn't need multiple lines, but I don't think it's a style violation. I'll have a look at the surrounding ones if it fits in.

corlewis: Could be GitHub but it looks like the `then` and `else` sections of the `if` are indented one more than they should be?

corlewis (approved):
So much nicer, thanks for suffering through this and cleaning it all up!

spec/capDL/KHeap_D.thy
lsf37: Can confirm that it is GitHub rendering, the indentation is correct in jEdit.

PR #1009, opened 2026-06-04
spec+proofs: consolidate maxIRQ definition in Platform.thy

Xaphiosis:
On Arm architectures, maxIRQ defaulted to Kernel_Config.maxIRQ, while on other architectures, it defaulted to Platform.L4VARCH.maxIRQ. This caused divergence when attempting to make interfaces that work on all architectures when arch-splitting Interrupt_R.

By forcing a Platform.L4VARCH.maxIRQ definition to exist on all platforms as an numeral, and pointing it to Kernel_Config.maxIRQ on architectures that have it, we can now point to the same constant on all architectures in interfaces.

For the design spec, we keep the abbreviation confining maxIRQ to type irq.

For CRefine, add maxIRQ_def to [simp] in ArchMove_R for Kernel_Config architectures, in order to keep the proof as close to what it used to be, which expected to only see Kernel_Config.maxIRQ.

Refine update is the minimal necessary to proceed to Interrupt_R arch-split, where we can have a go at proper interfacing.

This also means that Arch_Kernel_Config_Lemmas now needs to say which maxIRQ it's talking about. Apart from being verbose, this now causes another friction point: we might want to have those lemmas on the arch-common maxIRQ, and they only exist on Kernel_Config-enabled arches. So if we add such lemmas for Platform.L4VARCH.maxIRQ, we immediately get a "what do I call this thing" head-scratcher.

Merge-wise, this can go in after Tcb_R arch-split. It's necessary to arch-split Interrupt_R.

### DIFF
--- a/proof/crefine/RISCV64/Finalise_C.thy
+++ b/proof/crefine/RISCV64/Finalise_C.thy
@@ -2187,7 +2187,7 @@ lemma finaliseCap_ccorres:
    apply (rule ccorres_if_lhs)
     apply (simp add: Collect_False Collect_True Let_def
                 del: Collect_const)
-    apply (rule_tac P="(capIRQ cap) \<le>  RISCV64.maxIRQ" in ccorres_gen_asm)
+    apply (rule_tac P="(capIRQ cap) \<le> maxIRQ" in ccorres_gen_asm)
     apply (rule ccorres_rhs_assoc)+
     apply csymbr
     apply csymbr
--- a/proof/crefine/X64/Finalise_C.thy
+++ b/proof/crefine/X64/Finalise_C.thy
@@ -2682,7 +2682,7 @@ lemma finaliseCap_ccorres:
    apply (rule ccorres_if_lhs)
     apply (simp add: Collect_False Collect_True Let_def
                 del: Collect_const)
-    apply (rule_tac P="(capIRQ cap) \<le>  X64.maxIRQ" in ccorres_gen_asm)
+    apply (rule_tac P="(capIRQ cap) \<le> maxIRQ" in ccorres_gen_asm)
     apply (rule ccorres_rhs_assoc)+
     apply csymbr
     apply csymbr
@@ -2743,13 +2743,13 @@ lemma finaliseCap_ccorres:
    apply (frule cap_get_tag_to_H, erule(1) cap_get_tag_isCap [THEN iffD2])
    apply (frule(1) ccap_relation_IRQHandler_mask)
    apply (clarsimp simp: isCap_simps irqInvalid_def
-                      valid_cap'_def X64.maxIRQ_def
+                      valid_cap'_def maxIRQ_def
                       Kernel_C.maxIRQ_def)
     apply (rule irq_opt_relation_Some_ucast'[simplified Kernel_C.maxIRQ_def, simplified])
      apply fastforce
     apply simp
     apply (clarsimp simp: isCap_simps irqInvalid_def
-                      valid_cap'_def X64.maxIRQ_def
+                      valid_cap'_def maxIRQ_def
                       Kernel_C.maxIRQ_def)
    apply fastforce
   apply clarsimp
--- a/proof/refine/RISCV64/Interrupt_R.thy
+++ b/proof/refine/RISCV64/Interrupt_R.thy
@@ -164,7 +164,7 @@ lemma checkIRQ_corres:
   "corres (ser \<oplus> dc) \<top> \<top> (arch_check_irq irq) (checkIRQ irq)"
   unfolding arch_check_irq_def checkIRQ_def rangeCheck_def
   apply (rule corres_guard_imp)
-    apply (clarsimp simp: minIRQ_def unlessE_whenE not_le)
+    apply (clarsimp simp: minIRQ_def maxIRQ_def unlessE_whenE not_le)
     apply (rule corres_whenE)
       apply (fastforce simp: ucast_nat_def)+
   done
@@ -183,10 +183,10 @@ crunch arch_check_irq, checkIRQ
 
 lemma arch_check_irq_valid:
   "\<lbrace>\<top>\<rbrace> arch_check_irq y \<lbrace>\<lambda>_. (\<lambda>s. unat y \<le> unat maxIRQ \<and> unat y \<noteq> unat irqInvalid)\<rbrace>, -"
-  unfolding arch_check_irq_def
+  unfolding arch_check_irq_def maxIRQ_def
   apply (wpsimp simp: validE_R_def wp: whenE_throwError_wp)
   apply (rule conjI)
-   apply (metis unat_ucast_up_simp[where 'a=6 and 'b=64, simplified] word_le_nat_alt word_le_not_less)
+   apply (clarsimp simp: not_less word_le_nat_alt)
   apply (simp add: unat_ucast_up_simp[where 'a=6 and 'b=64, simplified] irqInvalid_def)
   apply (rule unat_mono[where a=0, simplified])
   apply (simp add: word_neq_0_conv)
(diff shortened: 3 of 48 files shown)


lsf37:
> This also means that Arch_Kernel_Config_Lemmas now needs to say which maxIRQ it's talking about. Apart from being verbose, this now causes another friction point: we might want to have those lemmas on the arch-common maxIRQ, and they only exist on Kernel_Config-enabled arches. So if we add such lemmas for Platform.L4VARCH.maxIRQ, we immediately get a "what do I call this thing" head-scratcher.

We need to be careful with those. I would not expect all of them to be true for all architectures, because they do handle IRQs differently. In particular `irqInvalid`, but I'm sure there are other details that I'm forgetting.

Some of the basic facts should be exportable, though. E.g. that the size of the IRQ CNode is always sufficient for all IRQs (`irqBits`/`irq_len` relationship to `maxIRQ`), and therefore the `ucast`/`unat`/`enum` rules that are based on that. `irq_len < machine_word_len` is also universally true. `irq_len` must be significantly smaller than `machine_word_len`, otherwise the IRQ CNode would take up all memory of the machine.

proof/refine/RISCV64/Interrupt_R.thy
```diff
@@ -183,10 +183,10 @@ crunch arch_check_irq, checkIRQ
 
 lemma arch_check_irq_valid:
   "\<lbrace>\<top>\<rbrace> arch_check_irq y \<lbrace>\<lambda>_. (\<lambda>s. unat y \<le> unat maxIRQ \<and> unat y \<noteq> unat irqInvalid)\<rbrace>, -"
-  unfolding arch_check_irq_def
+  unfolding arch_check_irq_def maxIRQ_def Kernel_Config.maxIRQ_def
```
lsf37: This one is concerning, we should not be unfolding Kernel_Config.maxIRQ_def, especially not globally at the beginning of a proof. Do you remember why that became necessary?

(It's not just arch specific, but platform specific, so can't be unfolded without danger in arch proofs either)

lsf37: Just saw that this is a RISCV64 file: Kernel_Config.maxIRQ should be unused here still.

lsf37 (approved):
👍 I think this gives us a sane setup to work from. It's good to see that some of the changes I'm making for RISCV on my branch are equal to what you needed here. That's encouraging.

There is a Kernel_Config unfolding that slipped in somewhere, which can probably just be left out. If not, I can deal with it when I get to it on my branch (I thought it was Arm when I first commented).

The Kernel_Config prefixes in Arch_Kernel_Config_Lemmas are not the prettiest, but we can think about that at some later point. One idea was to give the config value a slightly different name now that we have a universal proper `maxIRQ` in Platform.thy. Later, though.

proof/refine/RISCV64/Interrupt_R.thy
Xaphiosis: Ah yes, indeed. I can probably remove it then,

Xaphiosis:
> We need to be careful with those. I would not expect all of them to be true for all architectures, because they do handle IRQs differently. In particular `irqInvalid`, but I'm sure there are other details that I'm forgetting.
> 
> Some of the basic facts should be exportable, though. E.g. that the size of the IRQ CNode is always sufficient for all IRQs (`irqBits`/`irq_len` relationship to `maxIRQ`), and therefore the `ucast`/`unat`/`enum` rules that are based on that. `irq_len < machine_word_len` is also universally true. `irq_len` must be significantly smaller than `machine_word_len`, otherwise the IRQ CNode would take up all memory of the machine.

What I was saying didn't quite go that far, I was only complaining about the naming part. If all the expected names are taken by Kernel_Config lemmas, then if you want the same name for something that spans all arches (or in fact just a Platform constant), then what do you call the new lemma. With what you're proposing, that'll still be an issue.

For one such lemma we went with maxIRQ_H instead of max_IRQ in Refine, but it's going to come up again.

lsf37:
> What I was saying didn't quite go that far, I was only complaining about the naming part. If all the expected names are taken by Kernel_Config lemmas, then if you want the same name for something that spans all arches (or in fact just a Platform constant), then what do you call the new lemma. With what you're proposing, that'll still be an issue.
> 
> For one such lemma we went with maxIRQ_H instead of max_IRQ in Refine, but it's going to come up again.

Got it, yes. That is indeed annoying. It'd be Ok to rename the Arch_Kernel_Config_Lemmas ones from `name` to `conf_name` or something like that, but they are used, so that will likely have a fairly large diff. Maybe instead a prefix for the interface lemmas, e.g. `plat_name`.

Xaphiosis:
> Got it, yes. That is indeed annoying. It'd be Ok to rename the Arch_Kernel_Config_Lemmas ones from `name` to `conf_name` or something like that, but they are used, so that will likely have a fairly large diff. Maybe instead a prefix for the interface lemmas, e.g. `plat_name`.

I'd probably want to rename the Kernel_Config ones. Explaining to someone why the normal "0 < maxIRQ" is named something like `zero_less_plat_maxIRQ` will not be intuitive, and also they can go anywhere, while Kernel_Config lemmas should probably stay near Kernel_Config...

lsf37:
> > Got it, yes. That is indeed annoying. It'd be Ok to rename the Arch_Kernel_Config_Lemmas ones from `name` to `conf_name` or something like that, but they are used, so that will likely have a fairly large diff. Maybe instead a prefix for the interface lemmas, e.g. `plat_name`.
> 
> I'd probably want to rename the Kernel_Config ones. Explaining to someone why the normal "0 < maxIRQ" is named something like `zero_less_plat_maxIRQ` will not be intuitive, and also they can go anywhere, while Kernel_Config lemmas should probably stay near Kernel_Config...

Fair enough. If we only do this for the ones where we need a generic version, maybe the fallout is not that large anyway.

proof/crefine/RISCV64/Finalise_C.thy
```diff
@@ -2187,7 +2187,7 @@ lemma finaliseCap_ccorres:
    apply (rule ccorres_if_lhs)
     apply (simp add: Collect_False Collect_True Let_def
                 del: Collect_const)
-    apply (rule_tac P="(capIRQ cap) \<le>  RISCV64.maxIRQ" in ccorres_gen_asm)
+    apply (rule_tac P="(capIRQ cap) \<le>  maxIRQ" in ccorres_gen_asm)
```
ryybrr: Extra space

proof/crefine/X64/Finalise_C.thy
```diff
@@ -2682,7 +2682,7 @@ lemma finaliseCap_ccorres:
    apply (rule ccorres_if_lhs)
     apply (simp add: Collect_False Collect_True Let_def
                 del: Collect_const)
-    apply (rule_tac P="(capIRQ cap) \<le>  X64.maxIRQ" in ccorres_gen_asm)
+    apply (rule_tac P="(capIRQ cap) \<le>  maxIRQ" in ccorres_gen_asm)
```
ryybrr: Extra space

proof/infoflow/refine/RISCV64/Example_Valid_StateH.thy
```diff
@@ -3416,7 +3416,7 @@ lemma s0H_invs:
   apply (rule conjI)
    apply (clarsimp simp: valid_machine_state'_def s0H_internal_def machine_state0_def)
   apply (rule conjI)
-   apply (clarsimp simp: irqs_masked'_def s0H_internal_def maxIRQ_def timer_irq_def irqInvalid_def)
+   apply (clarsimp simp: irqs_masked'_def s0H_internal_def maxIRQ_def maxIRQ_def timer_irq_def irqInvalid_def)
```
ryybrr: Duplicate definition

ryybrr (approved):
A few minor tweaks but all good otherwise

Xaphiosis:
Fixed @ryybrr 's latest keen eye finds, squashed and rebased.

PR #1011, opened 2026-06-05
Aarch64 asid map 

lsf37:
Remove the `VMID` entry from the ASID pool in ASpec and use a global `asid_map` instead, like for ARM. The implementation in Haskell and below is unchanged.

The idea is to remove the VMID entries from kheap so that the infoflow proof does not need to introduce additional kheap infrastructure to show that the VMID is not observable. This means infoflow can for the most part continue to assume kheap equality.

Proofs updated accordingly, with the largest change in Refine where we need to add an `aobjs_of'` parameter to `arch_state_relation` so that we can construct the abstract `asid_map` from the concrete VMID entries in ASID pools. This means generic theories are affected (new interface) and the other architectures need small updates to ignore the additional parameter.

Because it was useful in a few places, I have taken the opportunity to finally implement that option filter for projections that I've mentioned a few times. This is in a separate commit and then used in Refine. There are a few more places where it could be useful, but I have left the rest of the proofs alone for now to reduce noise.

### DIFF
--- a/lib/Monads/reader_option/Reader_Option_Monad.thy
+++ b/lib/Monads/reader_option/Reader_Option_Monad.thy
@@ -39,6 +39,8 @@ definition map_set :: "('a \<Rightarrow> 'b set option) \<Rightarrow> 'a \<Right
 definition opt_pred :: "('a \<Rightarrow> bool) \<Rightarrow> ('b \<Rightarrow> 'a option) \<Rightarrow> ('b \<Rightarrow> bool)" (infixl "|<" 55) where
   "P |< proj \<equiv> (\<lambda>x. case_option False P (proj x))"
 
+definition opt_filter :: "('a \<Rightarrow> bool) \<Rightarrow> 'a \<Rightarrow> 'a option" where
+  "opt_filter P x \<equiv> if P x then Some x else None"
 
 subsection \<open>Lemmas for @{const opt_map}\<close>
 
@@ -190,6 +192,101 @@ lemma opt_pred_proj_upd_eq[simp]:
   "(P |< proj (p \<mapsto> v)) p = P v"
   by (simp add: in_opt_pred)
 
+subsection \<open>Lemmas for @{const opt_filter}\<close>
+
+lemma opt_filter_True[simp]:
+  "opt_filter \<top> = Some"
+  by (simp add: opt_filter_def[abs_def])
+
+lemma opt_filter_False[simp]:
+  "opt_filter \<bottom> = (\<lambda>_. None)"
+  by (simp add: opt_filter_def[abs_def])
+
+lemma opt_filter_map_eq:
+  "opt_filter P |> f = (\<lambda>x. if P x then f x else None)"
+  by (auto simp: opt_filter_def[abs_def] opt_map_def)
+
+lemma neg_opt_filter_map_eq:
+  "opt_filter (not P) |> f = (\<lambda>x. if P x then None else f x)"
+  by (auto simp: opt_filter_def[abs_def] opt_map_def)
+
+lemma opt_filter_map_apply:
+  "(opt_filter P |> f) x = (if P x then f x else None)"
+  by (auto simp: opt_filter_def opt_map_def)
+
+lemma opt_filter_Some_map_eq:
+  "opt_filter P ||> f = (\<lambda>x. if P x then Some (f x) else None)"
+  by (auto simp: opt_filter_def[abs_def] opt_map_def)
+
+lemma in_opt_filter_eq:
+  "opt_filter P x = Some y \<longleftrightarrow> P x \<and> y = x"
+  by (auto simp: opt_filter_def)
+
+lemma in_opt_filter_None_eq:
+  "opt_filter P x = None \<longleftrightarrow> \<not> P x"
+  by (simp add: opt_filter_def)
+
+lemma in_opt_map_filter_Some:
+  "(f |> opt_filter P) x = Some y \<longleftrightarrow> f x = Some y \<and> P y"
+  by (simp add: in_opt_map_eq in_opt_filter_eq)
+
+lemma in_opt_filter_map_Some:
+  "(opt_filter P |> f) x = Some y \<longleftrightarrow> P x \<and> f x = Some y"
+  by (simp add: in_opt_map_eq in_opt_filter_eq)
+
+lemma opt_filter_conj:
+  "opt_filter P |> opt_filter Q = opt_filter (P and Q)"
+  by (auto simp: opt_filter_def[abs_def] opt_map_def)
+
+lemma opt_filter_idem[simp]:
+  "opt_filter P |> opt_filter P = opt_filter P"
+  by (simp add: opt_filter_conj)
+
+lemma opt_filter_comm:
+  "opt_filter P |> opt_filter Q = opt_filter Q |> opt_filter P"
+  by (simp add: opt_filter_conj inf_commute)
+
+lemma opt_pred_opt_filter[simp]:
+  "(P |< opt_filter Q) = (P and Q)"
+  by (auto simp: in_opt_pred in_opt_filter_eq)
+
+lemma opt_pred_opt_filter_map[simp]:
+  "(P |< (f |> opt_filter Q)) = (P and Q) |< f"
+  by (simp add: opt_pred_unfold_map)
+
+lemma opt_pred_is_opt_filter:
+  "(P |< f) x \<longleftrightarrow> (\<exists>y. (f |> opt_filter P) x = Some y)"
+  by (auto simp: in_opt_pred in_opt_filter_eq in_opt_map_eq)
+
+lemma opt_pred_top_Some:
+  "(\<top> |< f) = (\<lambda>x. \<exists>y. f x = Some y)"
+  by (auto simp: in_opt_pred)
+
+lemma opt_filter_mono:
+  "\<lbrakk> \<And>x. P x \<Longrightarrow> Q x; opt_filter P x = Some y \<rbrakk> \<Longrightarrow> opt_filter Q x = Some y"
+  by (simp add: in_opt_filter_eq)
+
+lemma opt_map_filter_id:
+  "\<lbrakk> \<And>x y. f x = Some y \<Longrightarrow> P y \<rbrakk> \<Longrightarrow> f |> opt_filter P = f"
+  by (auto simp: opt_map_def in_opt_filter_eq split: option.splits)
+
+lemma opt_map_Some_map_eq:
+  "g ||> f |> h = g |> (h \<circ> f)"
+  by (auto simp add: opt_map_def split: option.splits)
+
+lemma opt_filter_upd_None_id:
+  "\<not>P x \<Longrightarrow> (opt_filter P) (x := None) = opt_filter P"
+  by (auto simp: opt_filter_def)
+
+lemma opt_filter_apply_None:
+  "\<not>P x \<Longrightarrow> opt_filter P x = None"
+  by (auto simp: opt_filter_def)
+
+lemma opt_filter_apply_Some:
+  "P x \<Longrightarrow> opt_filter P x = Some x"
+  by (auto simp: opt_filter_def)
+
+lemmas opt_filter_apply = opt_filter_apply_None opt_filter_apply_Some
 
 section \<open>The Reader Option Monad\<close>
 
--- a/proof/invariant-abstract/AARCH64/ArchAcc_AI.thy
+++ b/proof/invariant-abstract/AARCH64/ArchAcc_AI.thy
@@ -749,37 +749,10 @@ lemma valid_asid_table_inj:
   unfolding valid_asid_table_def
   by clarsimp (drule inj_onD; fastforce)
 
-lemma set_asid_pool_None_vmid[wp]:
-  "\<lbrace>\<lambda>s. valid_asid_table s \<and>
-        asid_pools_of s p = Some ap \<and>
-        (\<exists>asid_high. asid_table s asid_high = Some p \<and>
-                     P ((vmid_for_asid s) (asid_of asid_high asid_low := None)))\<rbrace>
-   set_asid_pool p (ap(asid_low := None))
-   \<lbrace>\<lambda>_ s. P (vmid_for_asid s)\<rbrace>"
-  apply (wp_pre, wps, wp)
-  apply (clarsimp simp del: fun_upd_apply)
-  apply (erule rsubst[where P=P])
-  apply (rule ext)
-  apply (clarsimp simp: vmid_for_asid_def obind_def simp del: fun_upd_apply split: option.splits)
-  apply (clarsimp simp: opt_map_left_None)
-  apply (rule opt_map_apply_left_eq)
-  apply (clarsimp simp: entry_for_pool_def obind_def split: option.splits)
-  apply (drule (2) valid_asid_table_inj, simp)
-  apply (erule notE)
-  apply (simp add: asid_bits_of_defs)
-  apply (word_eqI_solve simp: asid_high_bits_def asid_low_bits_def)
-  done
-
 lemma set_asid_pool_None_vmid_inv:
-  "\<lbrace>\<lambda>s. vmid_inv s \<and> valid_asid_table s \<and>
-        asid_pools_of s p = Some ap \<and>
-        (\<exists>asid_high. asid_table s asid_high = Some p \<and>
-                     vmid_for_asid s (asid_of asid_high asid_low) = None) \<rbrace>
-   set_asid_pool p (ap(asid_low := None))
-   \<lbrace>\<lambda>_. vmid_inv\<rbrace>"
+  "set_asid_pool p (ap(asid_low := None)) \<lbrace>vmid_inv\<rbrace>"
   unfolding vmid_inv_def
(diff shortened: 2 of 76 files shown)


lib/Monads/reader_option/Reader_Option_Monad.thy
```diff
@@ -39,6 +39,8 @@ definition map_set :: "('a \<Rightarrow> 'b set option) \<Rightarrow> 'a \<Right
 definition opt_pred :: "('a \<Rightarrow> bool) \<Rightarrow> ('b \<Rightarrow> 'a option) \<Rightarrow> ('b \<Rightarrow> bool)" (infixl "|<" 55) where
   "P |< proj \<equiv> (\<lambda>x. case_option False P (proj x))"
 
+definition opt_filter :: "('a \<Rightarrow> bool) \<Rightarrow> 'a \<Rightarrow> 'a option" where
+  "opt_filter P x \<equiv> if P x then Some x else None"
```
Xaphiosis, line 43: Interesting that this didn't already have a definition upstream (like option_map or such).

lib/Monads/reader_option/Reader_Option_Monad.thy
```diff
@@ -190,6 +192,101 @@ lemma opt_pred_proj_upd_eq[simp]:
...
+lemma opt_filter_map_apply:
+  "(opt_filter P |> f) x = (if P x then f x else None)"
+  by (auto simp: opt_filter_def opt_map_def)
+
+lemma opt_filter_Some_map_eq:
+  "opt_filter P ||> f = (\<lambda>x. if P x then Some (f x) else None)"
+  by (auto simp: opt_filter_def[abs_def] opt_map_def)
+
+lemma in_opt_filter_eq:
+  "opt_filter P x = Some y \<longleftrightarrow> P x \<and> y = x"
```
Xaphiosis, line 222: curious, why the `y = x` ordering?

lib/Monads/reader_option/Reader_Option_Monad.thy
```diff
@@ -190,6 +192,101 @@ lemma opt_pred_proj_upd_eq[simp]:
...
+  "\<not>P x \<Longrightarrow> (opt_filter P) (x := None) = opt_filter P"
+  by (auto simp: opt_filter_def)
+
+lemma opt_filter_apply_None:
+  "\<not>P x \<Longrightarrow> opt_filter P x = None"
+  by (auto simp: opt_filter_def)
+
+lemma opt_filter_apply_Some:
+  "P x \<Longrightarrow> opt_filter P x = Some x"
+  by (auto simp: opt_filter_def)
```
Xaphiosis, line 287: Not for this PR, but I wonder how we're to handle these "almost-simp" rules. Obviously they can't be in the simpset in general due to performance, but at the same time having to use the `_apply_None` or `_apply_Some` separately when we've got these monadic things to clean up always feels a bit like we should have a better way. Bundles?

proof/invariant-abstract/AARCH64/ArchAcc_AI.thy
```diff
@@ -794,21 +767,56 @@ lemma set_asid_pool_dom[wp]:
...
+
+lemma set_asid_pool_None_valid_asid_table[wp]:
+  "\<lbrace>valid_asid_table and (\<lambda>s. asid_pools_of s p = Some ap)\<rbrace>
+   set_asid_pool p (ap(asid_low := None))
+   \<lbrace>\<lambda>_. valid_asid_table\<rbrace>"
   unfolding valid_asid_table_def
-  using set_asid_pool_asid_pools_of[wp del]
-  by (wp_pre, wps, wp, clarsimp)
+  supply set_asid_pool_asid_pools_of[wp del] fun_upd_apply[simp del]
+  apply (wp_pre, wps)
```
Xaphiosis, line 797: random question, should wps include hoare_pre?

proof/invariant-abstract/AARCH64/ArchInvariants_AI.thy
```diff
@@ -763,11 +763,12 @@ lemmas hyp_refs_of_simps[simp] = hyp_refs_of_def[split_simps kernel_object.split
 definition state_hyp_refs_of :: "'z::state_ext state \<Rightarrow> obj_ref \<Rightarrow> (obj_ref \<times> reftype) set" where
   "state_hyp_refs_of \<equiv> \<lambda>s p. case_option {} (hyp_refs_of) (kheap s p)"
 
-
-(* Mostly covered by ASIDPool case of valid_vspace_obj and vmid_inv, but we still need to make sure
-   that ASID 0 is never mapped. *)
+(* Compared to ARM we do not duplicate the vspace pointer in the asid_map, but instead use the
+   existing vspace_for_asid function The invariant records that we only allocate a VMID for an
```
Xaphiosis: missing `.` after `function`.

proof/refine/X64/ArchCSpace_R.thy
```diff
@@ -465,6 +466,12 @@ proof -
     "\<And>P. \<lbrace>\<lambda>s. P (new_irqn s) (new_irqs s)\<rbrace> f
              \<lbrace>\<lambda>rv s. \<exists>irn. interrupt_irq_node s = irn \<and> P irn (interrupt_states s)\<rbrace>"
     by (wp hoare_vcg_ex_lift updates, simp)
+  have g_as_ao'[wp]:
+    "\<And>P. \<lbrace>\<lambda>s'. P (new_as' s') (new_ao' s')\<rbrace> g \<lbrace>\<lambda>_ s'. P (ksArchState s') (aobjs_of' s')\<rbrace>"
+    apply wp_pre
```
Xaphiosis: given that wp goes first, does it need the wp_pre?

proof/refine/StateRelationPre.thy
```diff
@@ -25,4 +25,18 @@ text \<open>
 type_synonym obj_relation_cut = "Structures_A.kernel_object \<Rightarrow> Structures_H.kernel_object \<Rightarrow> bool"
 type_synonym obj_relation_cuts = "(machine_word \<times> obj_relation_cut) set"
 
+text \<open>
+  Arch object projection for predicates that depend on arch objects only.\<close>
```
Xaphiosis: can't this fit on one line? looks like 80 chars

proof/refine/ARM_HYP/ArchStateRelation.thy
```diff
@@ -175,8 +175,9 @@ definition ghost_relation_wrapper_2 ::
 (* inside Arch locale, we have no need for the wrapper *)
 lemmas ghost_relation_wrapper_def[simp] = ghost_relation_wrapper_2_def
 
-definition arch_state_relation :: "(arch_state \<times> ARM_HYP_H.kernel_state) set" where
-  "arch_state_relation \<equiv> {(s, s') .
+definition arch_state_relation ::
+  "(obj_ref \<rightharpoonup> arch_kernel_object) \<Rightarrow> (arch_state \<times> ARM_HYP_H.kernel_state) set" where
+  "arch_state_relation aobjs \<equiv> {(s, s') .
```
Xaphiosis: hmm, in this case aobjs is named but unused; would make sense to have a comment (same on other arches where it's unused)

proof/refine/AARCH64/ArchDetype_R.thy
```diff
@@ -267,9 +267,43 @@ end (* detype_locale' *)
...
+   apply (simp add: mapM_storeWord_clear_um[unfolded word_size_def] word_size_def
+                    intvl_range_conv[where 'a=machine_word_len, folded word_bits_def])
+   apply wp
+  apply (clarsimp simp: invs_def valid_state_def valid_irq_states_def clear_um_def cur_tcb_def
+                        valid_machine_state_def)
+  done
+
 (* FIXME arch-split: some of this can be generalised; 3 is same on all arches *)
+(* FIXME: is_aligned base magnitude; magnitude \<ge> 3: both already follow from valid_cap in
+          the abstract precondition *)
```
Xaphiosis, line 304: Now I'm even more confused. What is 3 then? Are you saying this can't be (partly) generalised?

Xaphiosis: Oh wait, that's just you adding an extra comment. Whoops. But have you figured out what 3 is?

proof/refine/ARM_HYP/ArchStateRelation.thy
Xaphiosis: Also this argument is names `aobjs'` on AARCH64

proof/refine/AARCH64/ArchStateRelation.thy
```diff
@@ -139,16 +190,18 @@ definition ghost_relation_wrapper_2 ::
 (* inside Arch locale, we have no need for the wrapper *)
 lemmas ghost_relation_wrapper_def[simp] = ghost_relation_wrapper_2_def
 
-definition arch_state_relation :: "(arch_state \<times> AARCH64_H.kernel_state) set" where
-  "arch_state_relation \<equiv> {(s, s') .
+definition arch_state_relation ::
+  "(obj_ref \<rightharpoonup> arch_kernel_object) \<Rightarrow> (arch_state \<times> AARCH64_H.kernel_state) set" where
+  "arch_state_relation aobjs' \<equiv> {(s, s') .
```
Xaphiosis, line 195: I'm a bit confused. On other arches that don't use `aobjs'`, the argument is called `aobjs`. I'm also surprised that while this PR is to change the ASID map to be "like ARM", ARM does not seem to use aobjs' in the arch_state_relation. Could you comment on how that worked out that way?

proof/refine/AARCH64/ArchVSpace_R.thy
```diff
@@ -1359,35 +1359,53 @@ lemma mask_is_asid_low_bits_of[simp]:
...
+              pool (ucast asid && mask asid_low_bits) \<noteq> None)"
+  apply (clarsimp simp: vspace_for_asid_def ex_abs_def entry_for_asid_def
+                        pool_for_asid_def entry_for_pool_def if_option_eq
+                        ucast_down_ucast_id is_down)
+  apply (erule state_relationE)
+  apply (fastforce simp: arch_state_relation_def asid_pool_relation_def
+                   dest!: asid_pool_of_ko_at'_cross)
+  done
+
+lemma abs_asid_leq_asidRange[simp]:
```
Xaphiosis: what does the "abs" stand for?

proof/refine/AARCH64/ArchVSpace_R.thy
```diff
@@ -1651,42 +1730,155 @@ lemma invalidate_vmid_entry_entry_for_asid[wp]:
   unfolding invalidate_vmid_entry_def
   by wpsimp
 
+
```
Xaphiosis: extra newline?

proof/refine/AARCH64/ArchVSpace_R.thy
```diff
@@ -1651,42 +1730,155 @@ lemma invalidate_vmid_entry_entry_for_asid[wp]:
...
+  assumes a: "is_ArchObj ob"
+  assumes arch:
+    "\<And>s s'. \<lbrakk> (arch_state s, ksArchState s') \<in> arch_state_relation (aobjs_of' s');
+               typ_at' (koTypeOf (injectKO ob')) ptr s'; obj_at' Q  ptr s'; P s \<rbrakk> \<Longrightarrow>
+              (arch_state s, ksArchState s') \<in>
+                arch_state_relation ((aobjs_of' s')(ptr := aobj_of' (injectKO ob')))"
+  shows      "other_aobj_relation ob (injectKO (ob' :: 'a :: pspace_storable)) \<Longrightarrow>
+  corres dc (obj_at (\<lambda>ko. a_type ko = a_type ob) ptr and obj_at (same_caps ob) ptr and P)
+            (obj_at' (Q :: 'a \<Rightarrow> bool) ptr)
+            (set_object ptr ob) (setObject ptr ob')"
```
Xaphiosis: this indentation does not look right; if there's no room, then `"other_aobj_relation` should migrate to the next line at least

proof/refine/AARCH64/ArchVSpace_R.thy
```diff
@@ -1651,42 +1730,155 @@ lemma invalidate_vmid_entry_entry_for_asid[wp]:
...
+  supply image_cong_simp [cong del] projectKOs[simp del]
```
Xaphiosis: inconsistent, remove space before `[`

proof/refine/AARCH64/ArchVSpace_R.thy
```diff
@@ -1651,42 +1730,155 @@ lemma invalidate_vmid_entry_entry_for_asid[wp]:
...
+  apply (prop_tac "koTypeOf (injectKO ob') \<noteq> TCBT")
+   subgoal
+     by (clarsimp simp: other_aobj_relation_def; cases ob; cases "injectKO ob'";
+         simp split: arch_kernel_obj.split_asm)
+  by (fastforce dest: tcbs_of'_non_tcb_update)
+
+lemma setObject_ASIDPool_None_corres[corres]:
+  "\<lbrakk> p = p'; asid_pool_relation pool (asidpool.ASIDPool pool');
+     asid_low' = ucast (asid_low_bits_of asid) \<rbrakk> \<Longrightarrow>
+  corres dc ((\<lambda>s. asid_pools_of s p = Some pool \<and> asid_table s (asid_high_bits_of asid) = Some p \<and>
```
Xaphiosis: I think these are old and you're mostly moving them, but while we're here, indentation on `corres`

lib/Monads/reader_option/Reader_Option_Monad.thy
lsf37: If you use this in rewriting, you would usually want the `y` to disappear, not the `x`, because the `y` in `Some y` will at least often (not always) be a generated name.

proof/invariant-abstract/AARCH64/ArchAcc_AI.thy
lsf37: I've been asking myself that many times. Probably yes. At least I can't see any reason why it would hurt.

lib/Monads/reader_option/Reader_Option_Monad.thy
lsf37: There is the `lemmas opt_filter_apply` set. We could also do a bundle, but I didn't think it would be used that extremely often.

proof/refine/StateRelationPre.thy
lsf37: Might fit, will try.

proof/refine/AARCH64/ArchDetype_R.thy
lsf37: Unfortunately not, only that it is redundant. It smells like word_size or something like that, but I'm not sure.

proof/refine/X64/ArchCSpace_R.thy
lsf37: Will check, can possibly go, yes.

proof/refine/AARCH64/ArchStateRelation.thy
lsf37: It should indeed just be called `aobjs'` everywhere. `aobjs` would be the ASpec name.

On ARM, the entire asid_map thing is pulled all the way down to Haskell, so the state relation between ASpec and ExecSpec doesn't need the heap, because it can just compare the two asid_map functions. The actual asid_map reasoning is happening in CRefine there, which is not a good place to do it, but that's what we did.

proof/refine/AARCH64/ArchVSpace_R.thy
lsf37: yes, that does indeed look terrible. I hope that was copy/paste or moved..

proof/refine/AARCH64/ArchVSpace_R.thy
lsf37: abstract asid (of type AARCH64_A.asid)

Xaphiosis: confusing. after word proof trauma I keep thinking of rep/abs, never even occurred to me it could refer to abstract spec

lsf37: right, I didn't even think of that. Happy to tweak if you have a good idea.

Xaphiosis: is AARCH64_H.asid a different type? hmm. recently something like this came up and we ended up with a `_H_` in the name... I dunno, does `ucast_asid_A_le_asidRange` sound plausible? If this lemma is unlikely to appear for the haskell one, then `ucast_asid_le_asidRange` is probably fine anyway.

lsf37: yeah, `ucast_asid_le_asidRange` sounds sufficient, I'll go with that

proof/invariant-abstract/AARCH64/ArchAcc_AI.thy
lsf37: https://github.com/seL4/l4v/issues/1019

proof/refine/X64/ArchCSpace_R.thy
lsf37: That does in deed work.

Xaphiosis (approved):
Seems fine now

PR #1013, opened 2026-06-09
Set VTIMER_UPDATE_VOFFSET default to OFF

lsf37:
Switch specifications to config CONFIG_VTIMER_UPDATE_VOFFSET=OFF, because a) nobody is using the CONFIG_VTIMER_UPDATE_VOFFSET=ON setting and b) it causes a directly observable information leak in the kernel.

It is a leak, because the user can directly read the register CNTVCT (virtual time) as well as ask the kernel for the value of CNTVOFF (offset). Together that means the user can observe the actual global physical time, which in the current model is too underspecified to derive that domain switch happens at always-determined times. Since in reality domain switches do have some jitter (until the time protection project finishes), that assumption would also not be true.

Without CONFIG_VTIMER_UPDATE_VOFFSET, the kernel does not read or store the global physical time. The user can still derive it, but that fact is no longer visible in the kernel model.

In both cases, the same actual time leak occurs (slightly more indirect with CONFIG_VTIMER_UPDATE_VOFFSET off). In both cases this time leak is of the same magnitude and kind as other time leaks the kernel already has, where the proof does not model the presence of other timer devices and explicitly does not talk about timing channels. That is, we make no more use of the no-timing assumption in the infoflow proof than without VCPUs.

Test with: https://github.com/seL4/seL4/pull/1687

(Needs to be merged after #1011)


proof/access-control/AARCH64/ArchAccess.thy
```diff
@@ -133,23 +141,23 @@ where
...
                        vgic_apr := vgic_apr (vcpu_vgic vst),
-                       vgic_lr := vgic_lr (vcpu_vgic vst)\<rparr>\<rparr>
-          else vcpu))" \<comment> \<open>Not current VCPU\<close>
+                       vgic_lr := vgic_lr (vcpu_vgic vst)\<rparr>\<rparr>))"
 
 definition vcpu_integrity where
-   "vcpu_integrity aag subjects x hv hv' cv cv' n n' vcpus vcpus' \<equiv>
-    (pasObjectAbs aag x \<notin> subjects \<longrightarrow> vcpu_of_state hv cv n vcpus x = vcpu_of_state hv' cv' n' vcpus' x)"
+   "vcpu_integrity hv hv' cv cv' n n' vopt vopt' \<equiv>
+      vcpu_of_state hv cv n vopt = vcpu_of_state hv' cv' n' vopt'"
```
Xaphiosis, line 148: Not sure how excited I am about `vopt` as a name. All around there's `vcpu` for VCPU stuff, but then from `cur_vcpu_of_2` and these changes we see `vopt` show up, so it feels like it's a new invention for this PR. I guess technically it's `vcpu_ptr_opt`, which would be a mouthful, but at least something with `vcpu` in it I think would be nicer.

Xaphiosis (approved):
Bikeshed on the `vopt`, otherwise looks good from my side.

proof/access-control/AARCH64/ArchAccess.thy
lsf37: Now that you mention it, I would also prefer `vcpu_opt` at least if `vcpu_ptr_opt` is too long.

lsf37:
I had forgotten to add the entire arm-hyp part, which is now there and squashed in. Not sure you need to re-review @Xaphiosis, the content in ARM_HYP is the same as in AARCH64. Just letting you know that there has been change.

@ryybrr are you happy for me to do the renaming fixes for `vopt` in Access?

proof/access-control/AARCH64/ArchAccess.thy
lsf37: After looking at this, I have to change my opinion: the `vopt` is used a lot in the access control files, especially in ArchArch_C. When I search for it, I get over 60 hits and all look VCPU related. So this is just fitting in with that.

lsf37:
(rebased after merge of base branch)

proof/access-control/AARCH64/ArchAccess.thy
```diff
@@ -133,23 +141,23 @@ where
...
      arm_gicvcpu_numlistregs as = arm_gicvcpu_numlistregs as' \<and>
-     vcpu_integrity aag subjects x (vcpu_state ms) (vcpu_state ms')
-                                        (arm_current_vcpu as) (arm_current_vcpu as')
-                                        (arm_gicvcpu_numlistregs as) (arm_gicvcpu_numlistregs as')
-                                        (aobjs |> vcpu_of) (aobjs' |> vcpu_of)"
+     (pasObjectAbs aag x \<notin> subjects \<longrightarrow>
+      vcpu_integrity (vcpu_state ms) (vcpu_state ms')
+                     (cur_vcpu_of_2 (arm_current_vcpu as) x) (cur_vcpu_of_2 (arm_current_vcpu as') x)
+                     (arm_gicvcpu_numlistregs as) (arm_gicvcpu_numlistregs as')
+                     ((aobjs |> vcpu_of) x) ((aobjs' |> vcpu_of) x))"
```
corlewis, line 160: Just checking, are the changes here and with the introduction of `cur_vcpu_of` 'just' improvements to the definitions? Except for `vcpu_mask`, the rest doesn't seem particularly related to the `VTIMER_UPDATE_VOFFSET` update.

corlewis (approved):
Looks good to me.

Very minor and more of a personal git preference, but if it was me I'd combine the `ainvs`, `refine` and `crefine` proof updates into one commit. They feel simple enough to me that them being separated doesn't gain as anything.

proof/access-control/AARCH64/ArchAccess.thy
lsf37: Yes, most of these are just improvements.

lsf37:
> Looks good to me.

👍 

> Very minor and more of a personal git preference, but if it was me I'd combine the `ainvs`, `refine` and `crefine` proof updates into one commit. They feel simple enough to me that them being separated doesn't gain as anything.

I'd prefer to leave it, tbh, tags like `aarch64+arm-hyp ainv+refine+crefine:` use up almost all the available space and just `proofs` is too coarse since we have specific `access` commits as well. It's true that there is nothing particularly forcing a separation, but they are also not trivial or something like that when separate.

corlewis:
> I'd prefer to leave it, tbh, tags like `aarch64+arm-hyp ainv+refine+crefine:` use up almost all the available space and just `proofs` is too coarse since we have specific `access` commits as well. It's true that there is nothing particularly forcing a separation, but they are also not trivial or something like that when separate.

Yeah, fair enough. That does make me think though, `ainvs+refine+crefine`  is a pretty common combination, do we want a shorter tag specifically for that? Maybe something like `fc` since it's updating the functional correctness proofs? I guess the downside would be that it's less obvious of a connection, so if you came across it in the logs and didn't already know what it represented then it wouldn't be helpful.

Indanz:
> That does make me think though, `ainvs+refine+crefine` is a pretty common combination, do we want a shorter tag specifically for that? Maybe something like `fc` since it's updating the functional correctness proofs? I guess the downside would be that it's less obvious of a connection, so if you came across it in the logs and didn't already know what it represented then it wouldn't be helpful.

You could use the short tag in the title and write it out as `ainvs+refine+crefine`  in the commit message.

(What about `arc`? It bridges the spec with the implementation.)

PR #1016, opened 2026-06-10
AARCH64 MCS Specs

michaelmcinerney:
Introduce the theory file Time and collect many definitions, lemmas,
and axiomatizations related to time in the MCS kernel that were
previously in separate architecture-dependent files.

Update for AARCH64 MCS.


michaelmcinerney:
Force pushed to rebase and fix typo in commit message.

michaelmcinerney:
And I suppose we'll want to turn on testing for AARCH64 ASpec and Haskell, but I'm not sure how to do that.

michaelmcinerney:
I did try testing the `ASpec` for AARCH64  with a couple of the commits I pushed, but it didn't work. I'm really not sure how to test `HaskellKernel`. It doesn't seem to be covered in `run_tests`.

lsf37:
> I did try testing the `ASpec` for AARCH64 with a couple of the commits I pushed, but it didn't work. I'm really not sure how to test `HaskellKernel`. It doesn't seem to be covered in `run_tests`.

Will have a look later today on Haskell (it's definitely covered, the question is how to switch on only ASpec/ExecSpec/Haskell for now). The test matrix unfortunately needs to be changed on the master branch, because GitHub is great.

spec/abstract/AARCH64/ArchCSpace_A.thy
```diff
@@ -8,8 +8,7 @@
 chapter "Architecture-specific Functions for CSpace"
 
 theory ArchCSpace_A
-imports
-  ArchVSpace_A
+imports ArchVSpace_A
```
lsf37: The correct style according to the style guide theory is

```
theory ArchCSpace_Atheory ArchCSpace_A
  imports
    ArchVSpace_A
begin
```

(or on the same line if it's just one, but still indented)

Given that there are no other changes in this file, I would remove the change entirely, though.

spec/abstract/AARCH64/ArchInterrupt_A.thy
```diff
@@ -21,53 +21,58 @@ definition virqSetEOIIRQEN :: "virq \<Rightarrow> machine_word \<Rightarrow> vir
...
-
-            ct \<leftarrow> gets cur_thread;
-            st \<leftarrow> get_thread_state ct;
-            \<comment> \<open>until a proof links active current vcpu to runnable current thread, we need this
-               check: @{term handle_fault} needs a runnable current thread\<close>
-            when (runnable st) $ handle_fault ct $ ArchFault fault
-          od
-        | _ \<Rightarrow> return ()
+     ct \<leftarrow> gets cur_thread;
+     ct_schdble \<leftarrow> gets (schedulable ct);
```
lsf37:
```suggestion
     ct_schedulable \<leftarrow> gets (schedulable ct);
```

spec/abstract/AARCH64/Machine_A.thy
```diff
@@ -128,6 +128,9 @@ text \<open>
 definition idle_thread_ptr :: vspace_ref where
   "idle_thread_ptr = pptr_base + 0x1000"
 
+definition idle_sc_ptr :: vspace_ref where
+  "idle_sc_ptr = pptr_base + 0x3000"
```
lsf37, line 134: We should probably repeat the comment from `idle_thread_ptr` above here, even if we didn't do that on the other architectures.

spec/design/skel/AARCH64/Arch_Structs_B.thy
```diff
@@ -12,15 +12,24 @@ chapter "Common, Architecture-Specific Data Types"
...
 imports
   Setup_Locale
+  Lib.HaskellLib_H
   MachineExports
 begin
 
 context Arch begin arch_global_naming (H)
 
+definition usToTicks :: "word64 \<Rightarrow> word64" where
+  "usToTicks \<equiv> us_to_ticks"
```
lsf37: Should we make this one an (input) abbreviation? (bit like maxIRQ)

spec/design/skel/AARCH64/Hardware_H.thy
```diff
@@ -13,16 +13,35 @@ begin
...
+  "maxTicksToUs \<equiv> max_ticks_to_us"
+
+definition kernelWCETTicks :: "word64" where
+  "kernelWCETTicks \<equiv> kernelWCET_ticks"
+
+definition kernelWCETUs :: "word64" where
+  "kernelWCETUs \<equiv> kernelWCET_us"
+
+definition maxPeriodUs :: "word64" where
+  "maxPeriodUs \<equiv> MAX_PERIOD_US"
```
lsf37: same here, if we make these input abbreviations, we don't have to carry around simplification rules that they are equal, and we still don't need to make any unidiomatic changes to the Haskell code.

This would need an update for RISCV64 and ARM to do the same, but it's easier to do this now than later when we have more architectures.

spec/haskell/src/SEL4/Object/VCPU/AARCH64.hs
```diff
@@ -521,43 +521,45 @@ vcpuFlushIfCurrent t = do
...
-            if eisr0 /= 0 then countTrailingZeros eisr0
-                          else (countTrailingZeros eisr1) + 32
-        setIndex vcpuPtr irq_idx = (do
-                virq <- doMachineOp $ get_gic_vcpu_ctrl_lr (fromIntegral irq_idx)
-                virqen <- return $ virqSetEOIIRQEN virq 0
-                doMachineOp $ set_gic_vcpu_ctrl_lr (fromIntegral irq_idx) virqen
-                vgicUpdateLR vcpuPtr irq_idx virqen
-                )
+    curThread <- getCurThread
+    isschedulable <- getSchedulable curThread
```
lsf37: either

```suggestion
    isSchedulable <- getSchedulable curThread
```

or if that name is taken, then

```
    is_schedulable <- getSchedulable curThread
```

spec/haskell/src/SEL4/Object/VCPU/AARCH64.hs
```diff
@@ -568,14 +570,16 @@ irqVPPIEventIndex irq =
...
-            doMachineOp $ maskInterrupt True irq
-            let vppi = fromJust $ irqVPPIEventIndex irq
-            vcpuUpdate vcpuPtr
-                       (\vcpu -> vcpu{ vcpuVPPIMasked = vcpuVPPIMasked vcpu // [(vppi, True)] })
-            curThread <- getCurThread
-            runnable <- isRunnable curThread
-            when runnable $ handleFault curThread $ ArchFault $ VPPIEvent irq
-        _ -> return ()
+    curThread <- getCurThread
+    isschedulable <- getSchedulable curThread
```
lsf37: (same name change as above)

spec/machine/Time.thy
```diff
@@ -0,0 +1,258 @@
+(*
+ * Copyright 2026, Proofcraft Pty Ltd
+ *
+ * SPDX-License-Identifier: GPL-2.0-only
+ *)
+
+theory Time
+imports
+  Word_Lib.WordSetup
+  Kernel_Config \<comment> \<open>currently unnecessary, but added in anticipation of future use\<close>
```
lsf37:
```suggestion
  Kernel_Config \<comment> \<open>currently unnecessary, but added in anticipation of future use\<close>
```

(Kernel_Config already imports Word_Lib.WordSetup)

proof/crefine/RISCV64/ADT_C.thy
```diff
@@ -33,7 +33,11 @@ where
            tsType_CL (thread_state_lift (tcbState_C tcb)) = scast ThreadState_Running"
 
 definition
-  "handleHypervisorEvent_C = (CALL schedule();; CALL activateThread())"
+  "handleHypervisorEvent_C =
+     (CALL updateTimestamp();;
+      \<acute>ret__unsigned_long :== CALL checkBudgetRestart();;
+      CALL schedule();;
+      CALL activateThread())"
```
lsf37, line 40: This should really be a separate PR for just RISCV64 -- it's a fix for us forgetting to properly update the ADT on the C side to model hypervisor event entries. On RISCV64, we might be able to show that this is never called, but I think it's nicer to keep the proof of this function working for whenever RISCV64 gets hyp extensions.

proof/refine/RISCV64/Syscall_R.thy
```diff
@@ -2793,6 +2793,7 @@ proof -
      apply clarsimp
      apply (fastforce simp: simple_sane_strg sch_act_simple_def ct_in_state'_def
                       elim: st_tcb_ex_cap'' pred_tcb'_weakenE)
+    apply (rule updateTimeStamp_checkBudgetRestart_helper)
```
lsf37, line 2796: This one I'm not sure which PR it belongs to. If it's related to the hypervisor event change, then in that separate PR.

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -20993,11 +20993,6 @@ lemma handle_event_valid_sched:
 
       apply handle_event_valid_sched_fault+
 
-   (* HypervisorEvent *)
-   apply wpsimp
-   apply (clarsimp simp: ct_in_state_def schact_is_rct_def valid_sched_def ct_not_in_q_def
-                         pred_tcb_at_def obj_at_def)
-
   (* Interrupt *)
```
lsf37, line 20996: This removal looks strange, why was the case there before and is now gone in a generic file? Indentation of the previous cases is likely wrong now if these lines can be just removed.

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -25594,15 +25584,12 @@ lemma handle_event_cur_sc_in_release_q_imp_zero_consumed:
...
-  \<comment> \<open>handle_hypervisor_fault\<close>
-  apply wpsimp
-  apply (prop_tac "cur_sc_tcb_are_bound s")
-   apply (rule invs_strengthen_cur_sc_tcb_are_bound; fastforce?)
-  apply (intro conjI impI)
-    apply (rule cur_sc_tcb_are_bound_cur_sc_in_release_q_imp_zero_consumed; assumption?)
-    apply (erule invs_sym_refs)
-   apply (fastforce simp: ct_in_state_def pred_tcb_at_def obj_at_def)
-  apply fastforce
+  apply (wpsimp wp: check_budget_restart_if_lift
```
lsf37, line 25589: Is this still the `handle_hypervisor_fault` case? If yes, we should not remove the comment.

proof/crefine/RISCV64/ADT_C.thy
lsf37: That would include the generic change to hypervisor fault function in ASpec and Haskell. Maybe not necessarily a separate PR, but definitely a separate commit. It's not really anything to do with AARCH64, apart from the fact that we noticed it while doing the AARCH64 specs.

spec/design/skel/AARCH64/Arch_Structs_B.thy
michaelmcinerney: Yeah that sounds good

proof/invariant-abstract/DetSchedSchedule_AI.thy
michaelmcinerney: I think this case is now handled by the method `handle_event_valid_sched_fault` just before it, because it's shape is now the same as the other events.

lsf37 (approved):
Cool, other than teasing apart the generic code change that affects RISCV64 from the AARCH64 update, I think this is pretty good. I double-checked against a search on CONFIG_KERNEL_MCS and I can't find anything else that even looks arch specific. Which is a good sign.

Main query that might cost some work now, but reduce work later is whether we should make some definitions in ExecSpec input abbreviations instead. That'll save a few trivial equality lemmas and remove annoyances about having stated some property with the constant from the wrong spec level.

proof/invariant-abstract/DetSchedSchedule_AI.thy
lsf37: ah, that makes a lot of sense. I did not think it would be able to. In that case the indentation is likely still correct.

proof/crefine/RISCV64/ADT_C.thy
michaelmcinerney: This was the part I mentioned before, where I said I wasn’t sure about the changes. This does work, but my thoughts were that we would prefer to say that we just don’t receive this type of event. This is saying that we do, and that the specific code for this specific event is a noop, and then we just wrap it in the surrounding code. When RISCV does get hyp extensions then I think we’d get rid of this custom definition and use the definition from the CParser instead. But I could be wrong.

lsf37: Right, this is only correct, because `handle_hypervisor_fault` does nothing on RISCV, and we can't add the C version of that there, because that doesn't exist in RISCV64.

Arguing that the event doesn't occur is not so easy, it'd either be an assumption on do_user_op, or some kind of operation in the automaton that disallows it and that makes the automaton more awkward and arch specific (which we'd want to avoid for multikernel for instance). Given that the kernel entry/exit protocol is generic for this event in the specs, I would still go with what you have now, even if it is a somewhat large no-op. You're right that on AARCH64 (or when RISCV gets hyp extensions), this will be mostly `handleVCPUFault_proc` instead.

michaelmcinerney: Alright great. Thanks for clarifying that.

michaelmcinerney:
I should be able to address all the comments made so far by this afternoon or tonight. Should I split the work up into the commit for the change for the hypervisor event, as well as a separate commit for making the timer constants input abbreviations, and then having the remainder of the changes for AARCH64? It might make it a little harder to review for those who have already seen the diff, and it would mean a force push, so I thought I should ask first.

lsf37:
> I should be able to address all the comments made so far by this afternoon or tonight. 

👍 

> Should I split the work up into the commit for the change for the hypervisor event, as well as a separate commit for making the timer constants input abbreviations, and then having the remainder of the changes for AARCH64? It might make it a little harder to review for those who have already seen the diff, and it would mean a force push, so I thought I should ask first.

Happy with the split and force push.

michaelmcinerney:
I've now addressed all of the comments received so far, and have split up the work into several commits as suggested by Gerwin. I would recommend reviewing commit-by-commit.

michaelmcinerney:
AARCH64 tests are failing, maybe because I didn't pull after updates were made to this branch and left some things out. Should  `CBaseRefine` and `c-kernel` be added to the list of exclusions for testing on `AARCH64` in `run_tests`?

lsf37:
> AARCH64 tests are failing, maybe because I didn't pull after updates were made to this branch and left some things out. Should `CBaseRefine` and `c-kernel` be added to the list of exclusions for testing on `AARCH64` in `run_tests`?

Just `c-kernel` should be enough, that should exclude the rest.

We should also diagnose and fix that one, but that's for another PR.

lsf37:
On thing that I didn't look at before: now that we have multiple architecture on the `rt` branch, we do need to add architecture tags for things that are arch specific. I think we had decided on the pattern `rt aarch64 ainvs:`.

The commits that just factor out `Time.thy` probably don't need an arch tag (they affect all architectures + generic), but the ones that are aarch64 specific spec and proof things, should get one.

michaelmcinerney:
Forced pushed to update the `run_tests` file (testing now seems to be working as we'd like) and to add arch tags.

spec/machine/AARCH64/MachineOps.thy
```diff
@@ -12,6 +12,7 @@ imports
   Word_Lib.WordSetup
   Monads.Nondet_Monad
   MachineMonad
+  Time
```
corlewis, line 15: Incredibly minor but I feel like the addition of this import here belongs in the AARCH64 commit, not in the initial commit. At this point in the commits we're still only dealing with ARM and RISCV64.

spec/machine/Time.thy
```diff
@@ -0,0 +1,257 @@
+(*
+ * Copyright 2026, Proofcraft Pty Ltd
+ *
+ * SPDX-License-Identifier: GPL-2.0-only
+ *)
+
+theory Time
+imports
+  Kernel_Config \<comment> \<open>currently unnecessary, but added in anticipation of future use\<close>
```
corlewis, line 9: Will it be obvious when we reach that future use and can remove this comment?

proof/refine/StateRelation.thy
```diff
@@ -263,9 +263,8 @@ lemma
 
 lemma maxReleaseTime_equiv:
   "maxReleaseTime = MAX_RELEASE_TIME"
-  apply (clarsimp simp: maxReleaseTime_def MAX_RELEASE_TIME_def maxBound_max_word RISCV64_H.maxPeriodUs_def
-                        RISCV64_H.usToTicks_def MAX_PERIOD_def) (* FIXME arch-split RT *)
-  done
+  (* FIXME arch-split RT *)
+  by (clarsimp simp: maxReleaseTime_def MAX_RELEASE_TIME_def maxBound_max_word MAX_PERIOD_def)
```
corlewis: This might now not need the arch-split FIXME, since it looks like it no longer uses any lemmas from the Arch locale.

proof/refine/StateRelation.thy
```diff
@@ -1087,7 +1086,7 @@ lemma refill_ready_relation:
   "\<lbrakk>sc_relation sc n sc'; sc_valid_refills' sc'\<rbrakk>
    \<Longrightarrow> refill_ready time (refill_hd sc) = (rTime (refillHd sc') \<le> time)"
   apply (frule (1) refill_hd_relation)
-  by (clarsimp simp: refill_ready_def RISCV64_H.kernelWCETTicks_def refill_map_def) (* FIXME arch-split RT *)
+  by (clarsimp simp: refill_ready_def refill_map_def) (* FIXME arch-split RT *)
```
corlewis: Same here for this FIXME

proof/refine/StateRelation.thy
```diff
@@ -1099,8 +1098,8 @@ lemma refill_sufficient_relation:
   "\<lbrakk>sc_relation sc n sc'; sc_valid_refills' sc'\<rbrakk> \<Longrightarrow>
    refill_sufficient usage (refill_hd sc) = refillSufficient usage (refillHd sc')"
   apply (frule (1) refill_capacity_relation[where usage=usage])
-  by (clarsimp simp: refillSufficient_def refill_sufficient_def minBudget_def MIN_BUDGET_def
-                     RISCV64_H.kernelWCETTicks_def) (* FIXME arch-split RT *)
+  (* FIXME arch-split RT *)
+  by (clarsimp simp: refillSufficient_def refill_sufficient_def minBudget_def MIN_BUDGET_def)
```
corlewis: And this FIXME

spec/machine/RISCV64/MachineOps.thy
```diff
@@ -34,16 +34,14 @@ section "The Operations"
 
 subsection "Memory"
 
-definition loadWord :: "machine_word \<Rightarrow> machine_word machine_monad"
-  where
+definition loadWord :: "machine_word \<Rightarrow> machine_word machine_monad" where
```
corlewis, line 37: If we're fixing style of the other architectures here should we do ARM as well?

spec/machine/AARCH64/MachineOps.thy
```diff
@@ -28,7 +28,7 @@ text \<open>
 
   All this is done only to avoid a large number of axioms (2 for each operation).\<close>
 
-context Arch begin global_naming AARCH64
+context Arch begin arch_global_naming
```
corlewis, line 31: Huh, we probably want to do this to all architectures and on master as well. @Xaphiosis thoughts? It looks like there's a few other spec files with older `global_naming` commands as well.

spec/machine/AARCH64/MachineOps.thy
michaelmcinerney: Oh yep, good point. I'll put this bit into the other commit, thanks.

spec/machine/Time.thy
michaelmcinerney: It won't be immediately obvious when this import becomes necessary, but the comment should prevent people  removing this before then. This was Gerwin's suggestion, so maybe he has a better idea for a comment, or whether a comment is necessary or a good idea.

proof/refine/StateRelation.thy
michaelmcinerney: Oh great. I should have paid a bit more attention to whether this was still necessary, but I will remove these now, thanks.

spec/machine/RISCV64/MachineOps.thy
michaelmcinerney: Sure, can do. I think I had some more style fixes originally but then found it hard to separate out the commits and had to add some back.

spec/abstract/AARCH64/Init_A.thy
```diff
@@ -71,21 +71,28 @@ definition init_kheap :: kheap where
...
          tcb_flags = {},
          tcb_arch = init_arch_tcb
          \<rparr>,
+      idle_sc_ptr \<mapsto> SchedContext (default_sched_context
+                                    \<lparr>sc_tcb := Some idle_thread_ptr,
+                                     sc_refills := [\<lparr>r_time = 0, r_amount = MAX_PERIOD\<rparr>,
+                                                    \<lparr>r_time = MAX_PERIOD, r_amount = 0\<rparr>],
+                                     sc_budget := MAX_PERIOD,
+                                     sc_refill_max := MIN_REFILLS\<rparr>) 0,
+
```
corlewis: I think this is indented one too far and shouldn't have the empty line afterwards.

corlewis (approved):
Just a few minor comments, otherwise this looks very good!

It's also nice to see that as we've been hoping, the arch-specific parts of seL4 are relatively unaffected by MCS. Hopefully the proofs are similar.

spec/machine/AARCH64/MachineOps.thy
michaelmcinerney: This particular context block does have `arch_global_naming` on all arches, so I'll leave it as is for now, and we can decide on an approach and maybe clean up other instances separately.

spec/machine/RISCV64/MachineOps.thy
michaelmcinerney: Actually, these files have quite a lot of old style in them, and I don't really fix much at all of it in this PR. So if it's OK, I think I'll leave it as is for now and we can go through these files all together and update style another time.

corlewis: Sure, that's fine

PR #1020, opened 2026-06-14
Clean up old Python 2 compatibility features

timretout:
This started as just removing the 'six' module (as mentioned in passing in seL4/seL4#1590) but I ended up trying to clean up various python 2 compatibility features - but I've tried to keep the changes small.

Rather than get stuck in to the unicode/string handling in git-commit-email.py, I'm suggesting that that script's most likely unused and should be removed.

I've found one module (misc/pysymbols) that's still python2-only which I haven't touched.

I've successfully run the haskell-translator with these applied, which covers most of the diff:

$ L4V_ARCH=X64 python3 ./misc/regression/run_tests.py haskell-translator
Running 2 test(s)...
  Started kernel-config ...         
  Finished kernel-config             passed    (0:00:00 real, 0:00:00 cpu,  0.05GB)
  Started haskell-translator ...    
  Finished haskell-translator        passed    (0:00:00 real, 0:00:00 cpu,  0.05GB)

2/2 tests succeeded.

All tests passed.


lsf37 (approved):
Thank you, it's good to finally remove all this old python2 stuff.

I'm happy with the changes and agree that we should remove the commit email code.

I'd like to reword and combine some of the commits, though. Are you happy for me to just do that?

At that opportunity I would also squash-in the one space style fix the is unrelated to your changes but pops up because it is in a file that was touched.

timretout:
Yes, please do rearrange/reword as you think best!

PR #1021, opened 2026-06-15

corlewis:
* refine: arch-split Schedule_R
* spec+proof: SMC caps
* aspec+ainvs+refine: consolidate cap_class/capClass
* spec+proof: changes needed for an executable initialiser


lsf37 (approved):
Looks good from my side. I have reviewed the post-merge commit in detail (all good) and scrolled through the merge commit with a few point checks.

lsf37:
Since we haven't merged #1016 yet, we'd expect the AARCH64 run to fail. Fine to ignore that one.

michaelmcinerney (approved):
I only looked at the last commit, but looks good!

proof/refine/KHeap_R.thy
```diff
@@ -1417,6 +1417,14 @@ lemma dmo_return' [simp]:
                    bind_def modify_def put_def)
   done
 
+lemma no_fail_dmo'[wp]:
+  "no_fail P f \<Longrightarrow> no_fail (P o ksMachineState) (doMachineOp f)"
+  apply (simp add: doMachineOp_def split_def)
+  apply (rule no_fail_pre, wp)
+  apply simp
+  apply (simp add: no_fail_def)
+  done
```
Xaphiosis, line 1425: is this correctly indented? not sure where it came from, but the simp + simp looks odd

proof/refine/Bits_R.thy
```diff
@@ -24,13 +24,6 @@ lemma (in Arch) wordBits_word_bits:
 
 requalify_facts Arch.wordBits_word_bits
 
-(* same derivation on all architectures *)
-lemma (in Arch) wordBits_word_bits:
-  "wordBits = word_bits"
-  by (simp add: wordBits_def' word_bits_def)
-
-requalify_facts Arch.wordBits_word_bits
```
Xaphiosis: where did this go?

proof/refine/KHeap_R.thy
corlewis: Whoops, this was something that was deleted for some reason during the merge and then I realised was still needed when trying to update the proofs afterwards. I meant to squash it into the merge commit but forgot to. Will do once you're done with your review.

proof/refine/Bits_R.thy
corlewis: Similar here, the merge duplicated this, presumably because it had been added in both branches, and I found and removed it during the update. Will squash also.

proof/refine/KHeap_R.thy
corlewis: Oh, and I guess while here I'll open it up to check the indentation. It's the same on master but might as well fix it now if it does need to be.

proof/refine/Schedule_R.thy
```diff
@@ -1210,18 +1206,17 @@ lemma schact_bind_inside: "do x \<leftarrow> f; (case act of resume_cur_thread \
...
   "\<mu>s_in_ms = usInMs"
   by (simp add: usInMs_def \<mu>s_in_ms_def)
 
-lemma us_to_ticks_equiv:
+(*FIXME arch-split RT*)
+lemma (in Arch) us_to_ticks_equiv:
   "us_to_ticks = usToTicks"
   by (simp add: usToTicks_def)
 
+requalify_facts Arch.us_to_ticks_equiv
```
Xaphiosis, line 1218: you might want to keep this as a `(* same proof on all architectures *)` ... incidentally, I've been writing them the other way around, so the way I've been doing them would look like:
```
lemma usToTicks_us_to_ticks:
  "usToTicks = us_to_ticks"
```

proof/refine/Schedule_R.thy
```diff
@@ -1911,12 +1900,22 @@ crunch checkDomainTime
...
-lemma (in Schedule_R_3) schedule_invs':
+locale Schedule_R_3 = Schedule_R_2 +
+  assumes scheduleChooseNewThread_corres:
+    "corres dc
+       (\<lambda>s. invs s \<and> valid_ready_qs s \<and> ready_or_release s \<and> scheduler_action s = choose_new_thread)
+       (\<lambda>s. invs' s \<and> ksSchedulerAction s = ChooseNewThread)
+       schedule_choose_new_thread scheduleChooseNewThread"
+  assumes scheduleChooseNewThread_invs':
+    "scheduleChooseNewThread \<lbrace>invs'\<rbrace>"
+begin
```
Xaphiosis, line 1911: hard to see from the diff: is this basically the same interface for Schedule_R_3 as on master? I didn't see you tweaking the interface in ArchSchedule_R theories

proof/refine/Schedule_R.thy
```diff
@@ -3352,7 +3347,7 @@ lemma mergeNonoverlappingHeadRefill_corres:
                          simp: refill_map_def)
       apply (wpsimp simp: update_refill_hd_def wp: update_sched_context_wp)
      apply (clarsimp simp: vs_all_heap_simps active_sc_def is_active_sc2_def obj_at_def opt_map_def)
-     apply (wpsimp simp: updateRefillHd_def simp: objBits_simps)
```
Xaphiosis, line 3548: how come on RT so many objBits_simps can go away?

Xaphiosis (approved):
Looks good to me (checking the fixup commit). Only questions and a minor observation.

proof/refine/Schedule_R.thy
corlewis: Oh, I didn't even think about this being a lemma like that, I was just making the minimal change to keep it working on the one architecture we're currently working on. That's why I marked it with the `FIXME arch-split RT`, I was assuming that we'd look at it again when doing a proper arch-split pass for the rt branch.

Now that you've pointed it out though, I guess this is something that had better be true for all architectures. In that case, maybe it just needs a `FIXME move`.

proof/refine/Schedule_R.thy
corlewis: From memory, both lemmas have different preconditions now but otherwise yes, it's the same interface just moved to a different location in `Schedule_R`.

proof/refine/Schedule_R.thy
corlewis: I actually don't know? When going through the file I was just trying to remove them first before changing them to the gen version. I didn't investigate but maybe they weren't all needed in the first place.

proof/refine/Schedule_R.thy
Xaphiosis: Cool. Thoughts on the direction/naming thing?

I'm still tagging this kind of thing with `(* FIXME arch-split: move *)` given I have some hope for a second pass.

michaelmcinerney: I’m just about to remove this lemma in the AARCH64 updates. I made `usToTicks` an input abbreviation so this lemma is trivial now. So you can leave it and I’ll get rid of it when I rebase.

Indanz: Also, we're thinking about removing the microseconds interface and let user space do the conversion, because there are too many problems with it, see https://github.com/seL4/seL4/issues/1509.

corlewis: Oh, thanks for both pointing those out! In that case I'll leave this as is for now.

PR #1023, opened 2026-06-24

corlewis:
* spec+proof: dynamic domain scheduler
* spec+proof: fix domain schedule wrap


proof/refine/RISCV64/Syscall_R.thy
```diff
@@ -199,6 +199,11 @@ lemma fst_snd_pairI:
   "\<lbrakk> fst p = x; snd p = y \<rbrakk> \<Longrightarrow> p = (x, y)"
   by auto
 
+(* FIXME: move *)
+lemma timeArgSize_TIME_ARG_SIZE[simp]:
+  "timeArgSize = TIME_ARG_SIZE"
+  by (clarsimp simp: TIME_ARG_SIZE_def timeArgSize_def wordBits_word_bits)
```
corlewis: This being needed was a bit sad. On master both the abstract and design specs use the same constant, `timeArgLen`. I tried making `timeArgSize` visible to the abstract spec but its current definition depends on `wordBits`, which then pulls in a lot more of the design spec than we want.

proof/crefine/RISCV64/Invoke_C.thy
```diff
@@ -163,7 +164,6 @@ lemma invokeDomainScheduleSetStart_ccorres:
     apply wpsimp
    apply (simp add: guard_is_UNIV_def)
   apply clarsimp
-  apply (frule invs_weak_sch_act_wf)
```
lsf37: I assume this rule not being available is the reason we need `warch_sch_act_wf` in the precondition. Just to double check: we're not just missing the rule, right? It's not implied by `invs'` on MCS?

corlewis: Correct, `sch_act_wf` was removed from `invs'` on MCS.

proof/refine/RISCV64/Syscall_R.thy
lsf37: Is there a different definition we could use instead of `wordBits`? I think I was working around the same thing at one point in master. I guess we need something that's visible in Haskell here, though. Unless we use `wordBits` in Haskell and just ignore the entire Haskell definition, taking something more primitive defined earlier. Does that make any sense?

lsf37 (approved):
I'm happy with the merge, no issues from the domain scheduler side. I've looked at the merge fixup commit in detail and over the entire final state of the merge.

It'd be nice to sort out the `timeArgLen` so that we only have one of these. Maybe the choice of name on `master` wasn't so great, because all caps is not a legal constant name in Haskell. We could rename to something Haskell compatible in a separate commit and then apply that to master as well so that we are consistent on both branches.

corlewis:
> It'd be nice to sort out the `timeArgLen` so that we only have one of these. Maybe the choice of name on `master` wasn't so great, because all caps is not a legal constant name in Haskell. We could rename to something Haskell compatible in a separate commit and then apply that to master as well so that we are consistent on both branches.

To clarify, master has `timeArgLen = 1` or `2` in the Haskell and `timeArgLen ≡ LENGTH(ticks_len) div LENGTH(machine_word_len)` shared between the abstract and design specs, while MCS has `timeArgSize ≡ 64 div wordBits` in Haskell and design and `TIME_ARG_SIZE ≡ LENGTH(ticks_len) div word_bits` in the abstract.

When we talked about this before we decided to go with the constants already in MCS, i.e. remove `timeArgLen` in the merge and use `timeArgSize` and `TIME_ARG_SIZE` instead. Thinking about it now, should we instead go the other way and only keep `timeArgLen`?

proof/crefine/RISCV64/Invoke_C.thy
```diff
@@ -130,7 +130,8 @@ lemma invokeDomainSetSet_ccorres:
   done
 
 lemma invokeDomainScheduleSetStart_ccorres:
-  "ccorres dc xfdc invs' \<lbrace>unat \<acute>index___unsigned_long = idx\<rbrace> hs
+  "ccorres dc xfdc (invs' and (\<lambda>s. weak_sch_act_wf (ksSchedulerAction s) s))
```
michaelmcinerney: The formatting of the `ccorres` statement is a little bit off. It's not major, but we try to group the guards together where possible, and then the functions together where possible, indented by 2 from the original `ccorres`.

proof/crefine/RISCV64/Refine_C.thy
```diff
@@ -317,7 +317,9 @@ lemma getNBSendRecvDest_ccorres[corres]:
   done
 
 lemma handleSyscall_ccorres:
-  "ccorres dc xfdc invs' \<lbrace>\<acute>syscall___unsigned_long = syscall_from_H sysc\<rbrace> []
+  "ccorres dc xfdc
+     (invs' and (\<lambda>s. ksSchedulerAction s = ResumeCurrentThread))
```
michaelmcinerney, line 321: This one is perfection.

proof/invariant-abstract/DetSchedSchedule_AI.thy
```diff
@@ -16601,53 +16597,62 @@ lemma valid_blocked_valid_ready_qs_ready_and_sufficient:
...
+   apply (wpsimp wp: hoare_vcg_imp_lift' thread_set_domain_not_idle_valid_sched)
+  apply (clarsimp simp: tcb_at_kh_simps[symmetric] valid_sched_def)
+  apply (wpsimp wp: tcb_sched_dequeue_valid_ready_qs tcb_dequeue_not_queued
+                    tcb_sched_action_valid_sched_misc
+                    tcb_sched_dequeue_valid_blocked_except_set_const
+                    hoare_vcg_conj_lift hoare_vcg_imp_lift' tcb_sched_dequeue_ct_not_in_q)
+   apply (clarsimp simp: valid_blocked_thread_def)
+   apply (intro conjI)
+  using valid_blocked_valid_ready_qs_ready_and_sufficient schedulable_def2 apply blast
+   apply (clarsimp simp: runnable_eq_active tcb_at_kh_simps schedulable_def2)
```
michaelmcinerney: Indentation on this one?

lsf37:
> To clarify, master has `timeArgLen = 1` or `2` in the Haskell and `timeArgLen ≡ LENGTH(ticks_len) div LENGTH(machine_word_len)` shared between the abstract and design specs, while MCS has `timeArgSize ≡ 64 div wordBits` in Haskell and design and `TIME_ARG_SIZE ≡ LENGTH(ticks_len) div word_bits` in the abstract.

Thanks for that summary, I didn't have the full picture in mind any more.
 
> When we talked about this before we decided to go with the constants already in MCS, i.e. remove `timeArgLen` in the merge and use `timeArgSize` and `TIME_ARG_SIZE` instead. Thinking about it now, should we instead go the other way and only keep `timeArgLen`?

Yes, looking at this, we should go with the setup from master. It's more important that the specs share the setup and it is consistent with master than the Haskell code itself being generic. The Haskell side can basically be independent as long as the Isabelle definitions slot it.

michaelmcinerney (approved):
Looks great!

lsf37 (approved):
I'm happy with the new `timeArgLen` handling. Ready to merge from my side when the tests are passing.

PR #1024, opened 2026-07-02
Make proof generic over current seL4 RISC-V platforms 

lsf37:
This PR mostly makes the treatment of `maxIRQ` generic in the proofs, which is the only not-yet-generic parameter that still varies between RISC-V platforms. This means, after this PR is merged, all seL4 RISC-V platforms are supported by the full RISC-V proof stack.

Part of this update is fixing the machine interface definition `getActiveIRQ` for RISC-V. It was suppressing IRQ 0xFF, which was an invalid IRQ for a previous architecture, not for RISC-V. There is no effect on the proof, since this is an assumption that was not made use of anywhere (invalidIRQ = 0 on RISC-V).

PR for verified configs here: https://github.com/seL4/seL4/pull/1698. I'm planning to make a separate PR for adding some of these to CI one both PRs are merged.

### DIFF
--- a/proof/crefine/RISCV64/CSpace_C.thy
+++ b/proof/crefine/RISCV64/CSpace_C.thy
@@ -1930,7 +1930,7 @@ definition
 where
   "cleanup_info_wf' cap \<equiv> case cap of
       IRQHandlerCap irq \<Rightarrow>
-        UCAST(6\<rightarrow>machine_word_len) irq \<le>  SCAST(32 signed\<rightarrow>machine_word_len) Kernel_C.maxIRQ
+        UCAST(irq_len\<rightarrow>machine_word_len) irq \<le>  SCAST(32 signed\<rightarrow>machine_word_len) Kernel_C.maxIRQ
     | ArchObjectCap acap \<Rightarrow> arch_cleanup_info_wf' acap
     | _ \<Rightarrow> True"
 
@@ -2236,9 +2236,11 @@ lemma postCapDeletion_ccorres:
   apply (clarsimp simp: isCap_simps)
   apply (frule cap_get_tag_isCap_unfolded_H_cap(5))
   apply (clarsimp simp: cap_irq_handler_cap_lift ccap_relation_def cap_to_H_def
-                        cleanup_info_wf'_def c_valid_cap_def cl_valid_cap_def mask_def)
-  apply (rule mask_eq_ucast_eq[where 'a="6" and 'b="64" and 'c="64", symmetric, simplified])
-  by (simp add: mask_def)
+                        cleanup_info_wf'_def c_valid_cap_def cl_valid_cap_def)
+  apply (rule mask_eq_ucast_eq[where 'a="irq_len" and 'b="machine_word_len" and 'c="machine_word_len",
+                               symmetric, simplified])
+  apply (simp flip: LENGTH_irq_len_irqBits)
+  done
 
 lemma emptySlot_ccorres:
   "ccorres dc xfdc
@@ -2829,7 +2831,8 @@ lemma sameRegionAs_spec:
            apply (simp add: cap_to_H_def)
            apply (clarsimp simp: up_ucast_inj_eq c_valid_cap_def ucast_eq_mask
                                  cl_valid_cap_def mask_twice from_bool_0
-                          split: if_split bool.split
+                           simp flip: LENGTH_irq_len_irqBits
+                           split: if_split bool.split
                   | intro impI conjI
                   | simp)
           apply (frule_tac cap'=cap_b in cap_get_tag_isArchCap_unfolded_H_cap)
--- a/proof/crefine/RISCV64/Ctac_lemmas_C.thy
+++ b/proof/crefine/RISCV64/Ctac_lemmas_C.thy
@@ -161,11 +161,11 @@ lemmas ccorres_move_c_guard_ap = ccorres_move_c_guards [OF move_c_guard_ap]
 
 lemma array_assertion_abs_irq:
   "\<forall>s s'. (s, s') \<in> rf_sr \<and> True
-        \<and> (n s' \<le> 64 \<and> (x s' \<noteq> 0 \<longrightarrow> n s' \<noteq> 0))
+        \<and> (n s' \<le> 2 ^ irqBits \<and> (x s' \<noteq> 0 \<longrightarrow> n s' \<noteq> 0))
     \<longrightarrow> (x s' = 0 \<or> array_assertion intStateIRQNode_Ptr (n s') (hrs_htd (t_hrs_' (globals s'))))"
   apply (intro allI impI disjCI2)
   apply (clarsimp simp: rf_sr_def cstate_relation_def Let_def)
-  apply (clarsimp simp: h_t_valid_clift_Some_iff)
+  apply (clarsimp simp: h_t_valid_clift_Some_iff simp flip: LENGTH_irq_len_irqBits)
   apply (erule clift_array_assertion_imp, (simp add: exI[where x=0])+)
   done
 
--- a/proof/crefine/RISCV64/Interrupt_C.thy
+++ b/proof/crefine/RISCV64/Interrupt_C.thy
@@ -51,7 +51,7 @@ lemma ptr_add_assertion_irq_guard:
 
 lemma cte_at_irq_node':
   "invs' s \<Longrightarrow>
-    cte_at' (irq_node' s + 2 ^ cte_level_bits * ucast (irq :: 6 word)) s"
+    cte_at' (irq_node' s + 2 ^ cte_level_bits * ucast (irq :: irq)) s"
   by (clarsimp simp: invs'_def valid_state'_def valid_irq_node'_def
                      cte_level_bits_def real_cte_at' cteSizeBits_def shiftl_t2n)
 
@@ -91,18 +91,17 @@ proof -
    apply (simp add: guard_is_UNIV_def ghost_assertion_data_get_def
                     ghost_assertion_data_set_def)
   apply (clarsimp simp: cte_at_irq_node' ucast_nat_def)
-  apply (clarsimp simp: cte_wp_at_ctes_of badge_derived'_def
-                        Collect_const_mem unat_gt_0 valid_cap_simps' maxIRQ_def)
+  apply (clarsimp simp: cte_wp_at_ctes_of badge_derived'_def maxIRQ_def
+                        Collect_const_mem unat_gt_0 valid_cap_simps')
   apply (drule word_le_nat_alt[THEN iffD1])
-  apply clarsimp
-  apply (drule valid_globals_ex_cte_cap_irq[where irq=irq])
-  apply auto
+  using maxIRQ_less_2p_irqBits
+  apply (auto dest: valid_globals_ex_cte_cap_irq[where irq=irq])
   done
 qed
 
 lemma invokeIRQHandler_ClearIRQHandler_ccorres:
   "ccorres dc xfdc
-          (invs' and (\<lambda>s. weak_sch_act_wf (ksSchedulerAction s) s) and K(irq \<le> 0xFF))
+          (invs' and (\<lambda>s. weak_sch_act_wf (ksSchedulerAction s) s))
           (UNIV \<inter> {s. irq_' s = ucast irq}) []
       (InterruptDecls_H.invokeIRQHandler (ClearIRQHandler irq))
       (Call invokeIRQHandler_ClearIRQHandler_'proc)"
@@ -120,12 +119,9 @@ lemma invokeIRQHandler_ClearIRQHandler_ccorres:
     apply wp
    apply (simp add: guard_is_UNIV_def ghost_assertion_data_get_def
                     ghost_assertion_data_set_def)
-  apply (clarsimp simp: cte_at_irq_node' ucast_nat_def)
-  apply (drule word_le_nat_alt[THEN iffD1])
-  apply (auto simp add:Word.uint_up_ucast)
-  apply (case_tac "of_int (uint irq) \<noteq> 0 \<longrightarrow> 0 < unat irq")
-   by (auto simp: Collect_const_mem unat_eq_0)
-
+  apply (clarsimp simp: cte_at_irq_node' split: if_split)
+  apply (auto simp: unat_eq_0 unat_le_2p_irqBits)
+  done
 
 lemma ntfn_case_can_send:
   "(case cap of NotificationCap x1 x2 x3 x4 \<Rightarrow> f x3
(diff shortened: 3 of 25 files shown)


spec/machine/RISCV64/Arch_Kernel_Config_Lemmas.thy
```diff
@@ -42,5 +42,123 @@ lemma kernelELFBase_no_overflow:
   unfolding kernelELFBase_def kernelELFPAddrBase_def pptrTop_def
   by (simp add: mask_def Kernel_Config.physBase_def)
 
+(* maxIRQ lemmas *)
```
Xaphiosis, line 46: when we use `(* *)` comments for a grouping, typically we have an empty line after... if there's no empty line, we assume the comment is only for the thing it's attached to

spec/machine/RISCV64/Arch_Kernel_Config_Lemmas.thy
```diff
@@ -42,5 +42,123 @@ lemma kernelELFBase_no_overflow:
...
+
+lemma maxIRQ_ucast_toEnum_eq_irq:
+  "x \<le> Kernel_Config.maxIRQ \<Longrightarrow> toEnum (unat x) = (ucast x :: irq)" for x::machine_word
+  by (simp add: word_le_nat_alt Kernel_Config.maxIRQ_def)
+
+lemma Kernel_Config_maxIRQ_ucast_toEnum_eq:
+  "x \<le> Kernel_Config.maxIRQ \<Longrightarrow> toEnum (unat x) = x" for x::machine_word
+  by (simp add: word_le_nat_alt Kernel_Config.maxIRQ_def)
+
+(* The following are derived and no longer need to unfold, but are small enough to include here *)
```
Xaphiosis: Same comment as above on needing an extra line after if this applies to more than just the next lemma.
Also I don't really understand what this comment says. The following [...] no longer need to unfold?

spec/machine/RISCV64/Arch_Kernel_Config_Lemmas.thy
```diff
@@ -42,5 +42,123 @@ lemma kernelELFBase_no_overflow:
...
+
+lemma ucast_eq_irqInvalid_conv:
+  "x \<le> Kernel_Config.maxIRQ \<Longrightarrow> (ucast x = irqInvalid) = (x = ucast irqInvalid)" for x :: machine_word
+  unfolding irqInvalid_def
+  using leq_maxIRQ_leq_mask_irq_len
+  by (simp add: ucast_0_eq_left)
+
+lemma ucast_irq_bounded_machine_word:
+  "(ucast irq :: machine_word) < 2 ^ irq_len" for irq::irq
+  by (simp add: word_less_nat_alt unat_ucast_upcast is_up unat_2p_irq_len_machine unat_irq_bounded)
```
Xaphiosis, line 166: aside: how close are these Kernel_Config lemmas to the other Arms'?

spec/machine/RISCV64/MachineOps.thy
```diff
@@ -135,9 +135,9 @@ definition getActiveIRQ :: "bool \<Rightarrow> (irq option) machine_monad"
      is_masked \<leftarrow> gets $ irq_masks;
      modify (\<lambda>s. s \<lparr> irq_state := irq_state s + 1 \<rparr>);
      active_irq \<leftarrow> gets $ irq_oracle \<circ> irq_state;
-     if is_masked active_irq \<or> active_irq = 0xFF \<or> (in_kernel \<and> active_irq \<in> non_kernel_IRQs)
+     if is_masked active_irq \<or> active_irq = irqInvalid \<or> (in_kernel \<and> active_irq \<in> non_kernel_IRQs)
      then return None
-     else return ((Some active_irq) :: irq option)
```
Xaphiosis, line 140: For this commit (`riscv machine: fix getActiveIRQ`) I have an irrational bias towards a non-fix title. Something like `riscv machine: use irqInvalid in getActiveIRQ` maybe? I have traumatic experiences where saying something was "fixed" was like taunting the gods. Saying it was updated to use irqInvalid instead of the incorrect 0xFF feels less drastic.

Also typo in commit message: `RSICV64`

spec/abstract/RISCV64/Init_A.thy
```diff
@@ -25,10 +25,17 @@ definition riscv_global_pt_ptr :: obj_ref
   where
   "riscv_global_pt_ptr = pptr_base + 0x2000"
 
-(* Sufficiently aligned for irq type + cte_level_bits *)
+(* Sufficiently aligned for irq type + cte_level_bits. irq_len can be up to 10, depending on
+   platform, so we need alignment to at least 15 bits. *)
```
Xaphiosis: `alignment to at least 15 bits` -> `an alignment of at least 15 bits`?
If you want "to at least", then I think it would fit in a "we need X to be aligned to at least ..." style sentence

proof/crefine/RISCV64/CSpace_C.thy
```diff
@@ -2829,6 +2831,7 @@ lemma sameRegionAs_spec:
            apply (simp add: cap_to_H_def)
            apply (clarsimp simp: up_ucast_inj_eq c_valid_cap_def ucast_eq_mask
                                  cl_valid_cap_def mask_twice from_bool_0
+                           simp flip: LENGTH_irq_len_irqBits
                           split: if_split bool.split
```
Xaphiosis: indentation edge case: alignment is now to argument start for first two args (simp, simp flip), but `:` for third (split)

proof/crefine/RISCV64/Ctac_lemmas_C.thy
```diff
@@ -161,11 +161,11 @@ lemmas ccorres_move_c_guard_ap = ccorres_move_c_guards [OF move_c_guard_ap]
 
 lemma array_assertion_abs_irq:
   "\<forall>s s'. (s, s') \<in> rf_sr \<and> True
-        \<and> (n s' \<le> 64 \<and> (x s' \<noteq> 0 \<longrightarrow> n s' \<noteq> 0))
+        \<and> (n s' \<le> 2^irqBits \<and> (x s' \<noteq> 0 \<longrightarrow> n s' \<noteq> 0))
```
Xaphiosis: did we ever decide if we prefer `2 ^ n` over `2^n`? I feel like we talked about this at some point, but completely forgot what the outcome was

proof/crefine/RISCV64/CSpace_C.thy
```diff
@@ -2236,9 +2236,11 @@ lemma postCapDeletion_ccorres:
...
   apply (frule cap_get_tag_isCap_unfolded_H_cap(5))
   apply (clarsimp simp: cap_irq_handler_cap_lift ccap_relation_def cap_to_H_def
-                        cleanup_info_wf'_def c_valid_cap_def cl_valid_cap_def mask_def)
-  apply (rule mask_eq_ucast_eq[where 'a="6" and 'b="64" and 'c="64", symmetric, simplified])
-  by (simp add: mask_def)
+                        cleanup_info_wf'_def c_valid_cap_def cl_valid_cap_def)
+  apply (rule mask_eq_ucast_eq[where 'a="irq_len" and 'b="machine_word_len" and 'c="machine_word_len",
+                               symmetric, simplified])
+  apply (simp flip: LENGTH_irq_len_irqBits)
+  done
```
Xaphiosis, line 2243: waaay nicer now; do the Arms use this nicer tech too?

proof/crefine/RISCV64/Interrupt_C.thy
```diff
@@ -352,7 +341,7 @@ lemma isIRQActive_ccorres:
         (isIRQActive irq) (Call isIRQActive_'proc)"
   apply (cinit lift: irq_')
    apply (simp add: getIRQState_def getInterruptState_def)
-   apply (rule_tac P="irq \<le> ucast Kernel_C.maxIRQ \<and> unat irq < (126::nat)" in ccorres_gen_asm)
+   apply (rule_tac P="irq \<le> ucast Kernel_C.maxIRQ \<and> unat irq < Kernel_Config.maxIRQ+1" in ccorres_gen_asm)
```
Xaphiosis: would prefer ` + 1`

proof/crefine/RISCV64/Interrupt_C.thy
```diff
@@ -443,33 +433,31 @@ lemma checkIRQ_wpE:
...
-  by (simp add: word_le_nat_alt maxIRQ_def Kernel_C.maxIRQ_def ucast_nat_def unat_ucast)
+  unfolding maxIRQ_def
+  by (simp add: Kernel_C_maxIRQ maxIRQ_ucast_toEnum_eq_irq irq_machine_le_maxIRQ_irq)
 
 lemma ucast_toEnum_unat_irq_t_leq_scast:
   "x \<le> ucast maxIRQ \<Longrightarrow> UCAST(_ \<rightarrow> machine_word_len) (toEnum (unat x)::irq) \<le> scast Kernel_C.maxIRQ"
   for x::machine_word
-  by (simp add: word_le_nat_alt maxIRQ_def Kernel_C.maxIRQ_def ucast_nat_def unat_ucast)
+  unfolding maxIRQ_def
+  by (simp add: Kernel_C_maxIRQ maxIRQ_ucast_toEnum_eq_irq irq_machine_le_maxIRQ_irq)
```
Xaphiosis, line 454: I'm a little confused by this block. unfolding maxIRQ_def will take it from the Platform one to the Kernel_Config one, and then `maxIRQ_ucast_toEnum_eq_irq` and `irq_machine_le_maxIRQ_irq` are Kernel_Config lemmas? If so, why don't they need the `Kernel_Config.` prefix here?

proof/crefine/RISCV64/Wellformed_C.thy
```diff
@@ -498,8 +498,11 @@ text \<open>maxIRQ interface\<close>
 
 declare Kernel_C.maxIRQ_def[code]
 
-(* FIXME: compute maxIRQ for kernel config instead *)
-value_type irq_array_size = "Suc (unat Kernel_C.maxIRQ)"
+lemma Kernel_C_maxIRQ:
+  "Kernel_C.maxIRQ = Kernel_Config.maxIRQ"
+  by (simp add: Kernel_C.maxIRQ_def Kernel_Config.maxIRQ_def)
```
Xaphiosis, line 503: I'm wondering... this equation bypasses the Platform maxIRQ, which means it can only be applied after unfolding maxIRQ. Is this related to only having some properties on Platform maxIRQ, and the rest on Kernel_Config.maxIRQ, and so we end up in this strange land where we have to unfold to use the Kernel_Config lemmas, but if we unfolded we now need to relate the Kernel_C maxIRQ to Kernel_Config.maxIRQ?

proof/infoflow/Retype_IF.thy
```diff
@@ -560,7 +560,11 @@ lemma invoke_untyped_reads_respects_g:
    apply (clarsimp simp: is_cap_simps)
    apply (rule equiv_valid_guard_imp, rule invoke_untyped_reads_respects_g_wcap)
    apply (cases ui, clarsimp simp: cte_wp_at_caps_of_state valid_untyped_inv_wcap)
-   apply auto[1]
+   (* resolve schematics first, auto and fastforce only take forever *)
```
Xaphiosis: I first read this as `(only take forever)`, which seemed very deep and poetic... on second read, I realised it's `auto and (fastforce only)` :P

proof/infoflow/refine/RISCV64/Example_Valid_StateH.thy
```diff
@@ -2199,13 +2150,14 @@ lemma s0H_pspace_distinct':
...
-                          drule_tac b=ya and a="_ + _" in dual_order.trans, assumption)?,
-                         simp add: s0_ptr_defs)+)[16]
+   apply (elim disjE; clarsimp simp: kh0H_obj_def split: if_splits)
+                 apply (repeat 16
+                          \<open>solves \<open>clarsimp simp: objBitsKO_def archObjSize_def bit_simps
+                                                  irq_node_offs_range_def2 cte_level_bits_def
+                                                  pt_offs_range_def cnode_offs_range_def
+                                                  page_offs_range_def irq_len_val
+                                                  s0_ptr_defs mask_def is_aligned_def,
+                                   unat_arith?\<close>\<close>)
```
Xaphiosis: While this is bravery award territory, there's also a lot of repetition, and the differences appear to only be 16 vs 17. Would it make sense here to deploy `supply_local_method` technology here?

Xaphiosis:
Commits:
* `riscv aspec: adjustments for generic maxIRQ` has the following in the comment:
`- make sure we are referring to Kernel_Config.maxIRQ` however the contents of the commit does not have anything related to Kernel_Config. Removing that reduces the bullet point list to one point, and so it can stop being a list.

* `riscv proofs: update for generic maxIRQ`: the phrase `which also simplifies most proofs that are involved.` could use some clarification... do you maybe mean "proofs involving *them*" (where them = the properties?)

spec/machine/RISCV64/Arch_Kernel_Config_Lemmas.thy
lsf37: many are just copies of the Arm ones

spec/machine/RISCV64/Arch_Kernel_Config_Lemmas.thy
lsf37: None of the lemmas that follow unfold any `Kernel_Config.*dev`. This means they don't really have to be in this theory. (But they are similar enough to the other ones and small enough that we want to keep them together).

proof/crefine/RISCV64/Ctac_lemmas_C.thy
lsf37: yeah we did, and I think you won :)

proof/crefine/RISCV64/Interrupt_C.thy
lsf37: The lemmas are fine to use, that's why we make them. They never need the Kernel_Config prefix (unless the name is ambiguous somehow)

proof/infoflow/refine/RISCV64/Example_Valid_StateH.thy
lsf37: We could, and maybe I'm wrong, but I somewhat wanted to keep this ugly so that it doesn't look like this needs no cleanup. The current approach in these lemmas is "unfold everything to raw numbers, then use unat_arith to solve". Previously it was "unfold most of everything to raw numbers, do some word rules on magic numbers, then try to solve with big hammers". Neither of them is good. Ideally this would be mostly symbolic, which would make the whole theory a lot faster and hopefully easier to understand. It's a bit of a project, though.

proof/crefine/RISCV64/Interrupt_C.thy
Xaphiosis: Right, so it's a case of "we don't have all the same theorems on Platform.maxIRQ because that would be duplication, and we can't just have everyone use the Kernel_Config ones uniformly because x86"

proof/crefine/RISCV64/Ctac_lemmas_C.thy
Xaphiosis: Ahh, if I won then `2 ^ n` please (also elsewhere in PR)

spec/machine/RISCV64/Arch_Kernel_Config_Lemmas.thy
Xaphiosis: Ah, then you might also want to add something to the comment, at least something after the "unfold", otherwise it's a complete mystery when reading it.

proof/crefine/RISCV64/Interrupt_C.thy
lsf37: Correct, yes. This is CRefine, so everything is fully arch specific, and going via Platform.maxIRQ wouldn't really help anything.

proof/crefine/RISCV64/CSpace_C.thy
lsf37: yes, most of that is from Arm :-)

corlewis (approved):

proof/infoflow/refine/RISCV64/Example_Valid_StateH.thy
lsf37: (now fixed)

PR #1025, opened 2026-07-05
[drop] testing only, do not merge

lsf37:
bisecting where InfoFlowC hangs in CI


lsf37:
Failure pinned down to ExampleStateH and the pspace_distinct' lemma in there.

PR #1027, opened 2026-07-08
monads + ainvs: strengthen hoare_strengthen_pre_via_assert lemmas

michaelmcinerney:


lsf37 (approved):
Looks good. Thanks for updating the trace monad as well.

Xaphiosis (approved):
Change itself looks ok, but I'm curious why it was needed.

michaelmcinerney:
> Change itself looks ok, but I'm curious why it was needed.

In changing `set_tcb_obj_ref` to be an abbreviation in terms of `thread_set`, the lemma `sched_context_donate_valid_objs` broke. It used the previous version of this lemma, which I had not really come across before. (I'm not entirely sure how the previous proof worked, actually. It was quite complicated and involved some matching on the particular binds used in the function.) The relevant bit of `sched_context_donate_valid_objs` involves a `thread_set` and an `update_sched_context`, and to show `valid_objs`, we need there to be a sched context at the sc pointer and a TCB at the TCB pointer. Now, `thread_set` will fail if there is not a TCB at the pointer, but for `update_sched_context`, because it keeps the size of the scheduling context, the fact that it does not fail only says that there is some `sc` and some `n` such that `kheap s sc_ptr = Some (SchedContext sc n)`. In order for `sc_at sc_ptr s` to be true, we also need `valid_objs` so that we know that `n` is a `valid_sched_context_size`. So this is why I wanted to be able to assume `valid_objs` in the precondition of the "negative" assumption of this rule I strengthened. There wasn't a way to be more careful about the specification of the schematics and still get the right precondition, so I think this change was necessary for this approach to work.

lib/Monads/nondet/Nondet_VCG.thy
```diff
@@ -1376,7 +1376,7 @@ text \<open>Like @{thm hoare_strengthen_pre_via_assert_forward}, strengthen a pr
       failure. This version is intended for backward reasoning. Apply it to a goal to
       obtain a stronger precondition after proving the side condition.\<close>
 lemma hoare_strengthen_pre_via_assert_backward:
-  assumes neg: "\<lbrace> Not \<circ> E \<rbrace> f \<lbrace> \<bottom>\<bottom> \<rbrace>"
+  assumes neg: "\<lbrace> P and Not \<circ> E \<rbrace> f \<lbrace> \<bottom>\<bottom> \<rbrace>"
   assumes pos: "\<lbrace> P and E \<rbrace> f \<lbrace> Q \<rbrace>"
   shows "\<lbrace> P \<rbrace> f \<lbrace> Q \<rbrace>"
```
corlewis, line 1380: This doesn't look quite right to me for the backwards version, or at least isn't as general as it should be. I think something like the following would be better
 
```suggestion
  assumes neg: "\<lbrace> P and Not \<circ> E \<rbrace> f \<lbrace> \<bottom>\<bottom> \<rbrace>"
assumes pos: "\<lbrace> P' and E \<rbrace> f \<lbrace> Q \<rbrace>"
shows "\<lbrace> P and P' \<rbrace> f \<lbrace> Q \<rbrace>"
```

michaelmcinerney: I didn't think too much about forwards versus backwards, even though they're right there in the names of these lemmas. The backwards version is not used at all in `master`, and while the "backwards" version is used a couple times for MCS, it's used in a forwards way. Is it even a good lemma for backwards reasoning? The assumptions don't have preconditions that are just schematic. I don't think my brain is sufficiently backwards to figure out how a backwards version would work nicely. To me, it seems like we'd specify the `E` schematic manually, and then we'd have to weaken the precondition anyway. I for the life of me cannot figure out how to prove this `sched_context_donate_valid_objs` lemma without this kind of rule.

corlewis: True, I was just going off of the name but now that you've made me think about it properly, it is a weird rule for backward reasoning.

I'd almost be tempted to delete it but if it's used a couple of times in MCS then it's probably fine to ignore and leave in as it is.

michaelmcinerney: To me the “backwards” version isn’t so much a backwards version of the forwards rule, but just a standard form where the “negative” part is the negation of the positive part. So I could make the original version a stronger primed version and then make the version currently named backwards just the standard unprimed version. Making this change is much easier now than any other time. Thoughts?

lsf37: `hoare_strengthen_via_assert_Not`?

lsf37: Also not great. Maybe `hoare_strengthen_pre_via_assert_neg` to indicate that pos/neg assumptions are flipped.

michaelmcinerney: As far as the `via_assert` part goes, we don't have any other "hoare strengthen pre" rules, so I would get rid of the `via_assert` bit. But if that's not quite concrete enough, then I would say `hoare_strengthen_pre_via_fail`, or `via_bot` or something. And if we ignore forwards versus backwards, the `Not` version is just a weakening of the first version, so couldn't we just have `hoare_strengthen_pre'` and `hoare_strengthen_pre`?

corlewis: > To me the “backwards” version isn’t so much a backwards version of the forwards rule, but just a standard form where the “negative” part is the negation of the positive part. So I could make the original version a stronger primed version and then make the version currently named backwards just the standard unprimed version

That makes sense to me. I'd be happy with either the primed and unprimed versions or `_neg`, but now that we're disagreeing with the comment about this being a backward reasoning rule we should either change or remove that as well.

lsf37: I'm happy with either name as well (and agree that the comment needs updating)

michaelmcinerney:
Consider that the formatting of the comment was changed, along with the comment's body itself, and the naming, I just decided to force push. Hope that's not a problem.

corlewis (approved):

lsf37 (approved):

PR #1028, opened 2026-07-09
Make `set_tcb_obj_ref` an abbreviation

michaelmcinerney:
Also includes a slightly modified version of the commit for the `hoare_strengthen_pre` rules.

### DIFF
--- a/proof/invariant-abstract/IpcDet_AI.thy
+++ b/proof/invariant-abstract/IpcDet_AI.thy
@@ -360,18 +360,6 @@ lemma receive_ipc_preamble_st_tcb_at:
   by (wpsimp simp: receive_ipc_preamble_def
                wp: cancel_ipc_st_tcb_at hoare_vcg_if_lift2 get_simple_ko_wp)
 
-global_interpretation set_tcb_obj_ref: non_reply_op "set_tcb_obj_ref f ref new"
-  apply unfold_locales
-  apply (wpsimp simp: set_tcb_obj_ref_def wp: set_object_wp)
-  by (clarsimp simp: reply_at_ppred_def obj_at_def get_tcb_def
-              split: option.splits kernel_object.splits)
-
-global_interpretation set_tcb_obj_ref: non_ntfn_op "set_tcb_obj_ref f ref new"
-  apply unfold_locales
-  apply (wpsimp simp: set_tcb_obj_ref_def wp: set_object_wp)
-  by (clarsimp simp: ntfn_at_ppred_def obj_at_def get_tcb_def
-              split: option.splits kernel_object.splits)
-
 global_interpretation reply_unlink_tcb: non_reply_sc_op "reply_unlink_tcb t r"
   apply unfold_locales
   apply (wpsimp simp: reply_unlink_tcb_def wp: get_simple_ko_wp update_sk_obj_ref_wps gts_wp)
@@ -1253,7 +1241,7 @@ crunch maybe_return_sc
   and valid_ioc[wp]: valid_ioc
   and state_hyp_refs_of[wp]: "\<lambda>s. P (state_hyp_refs_of s)"
   and valid_cur_fpu[wp]: valid_cur_fpu
-  (simp: crunch_simps wp: crunch_wps ignore: set_tcb_obj_ref)
+  (simp: crunch_simps wp: crunch_wps ignore: thread_set)
 
 lemma set_sc_obj_ref_ko_not_tcb_at[wp]:
   "set_sc_obj_ref f scp v \<lbrace>\<lambda>s. \<not> ko_at (TCB tcb) t s\<rbrace>"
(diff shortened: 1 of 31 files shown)


michaelmcinerney:
Force pushed just to remove an `apply_trace` that the linter complained about.

michaelmcinerney:
I think ARM AInvs will fail, but it should be easy to fix. I'll fix locally and push when the other tests pass.

spec/abstract/KHeap_A.thy
```diff
@@ -204,13 +204,11 @@ definition
...
-where
-  "set_tcb_obj_ref f ref new \<equiv> do
-     tcb \<leftarrow> gets_the $ get_tcb ref;
-     set_object ref (TCB (f (K new) tcb))
-   od"
+abbreviation set_tcb_obj_ref ::
+  "((obj_ref option \<Rightarrow> obj_ref option) \<Rightarrow> tcb \<Rightarrow> tcb) \<Rightarrow> obj_ref \<Rightarrow> obj_ref option
+   \<Rightarrow> (unit, 'z::state_ext) s_monad"
+  where
+  "set_tcb_obj_ref f ref new \<equiv> thread_set (f (\<lambda>_. new)) ref"
```
lsf37: Should we make it (input) only?

I'm not really sure what I prefer. Making it input-only would help with figuring out what kind of rules one should be using with this and how to unfold it, but it would also mean people copy/paste the long form from goals and use it in lemma statements instead of using either only `thread_set` or the `set_tcb_obj_ref` syntax. Reading what I wrote, I think I slightly prefer input-only, but I'm interested in other opinions.

lsf37 (approved):
Nice. That does remove a whole lot of boiler plate.

spec/abstract/KHeap_A.thy
michaelmcinerney: I think I prefer it being an input abbreviation. I don’t really like having the abbreviation at all, even if it makes some definitions look a little prettier, so for me, the less it’s seen, the better. I also think it would encourage people to write lemmas for `thread_set` that are as general as possible.

Xaphiosis (approved):
`monads + ainvs` -> `monads+ainvs` in first commit message
otherwise this seems good to me. same question about it being an input-only abbreviation that Gerwin has, because those have a tendency of disappearing as people forget that they exist

michaelmcinerney:
> `monads + ainvs` -> `monads+ainvs` in first commit message otherwise this seems good to me. same question about it being an input-only abbreviation that Gerwin has, because those have a tendency of disappearing as people forget that they exist

I think the only benefit I can see of keeping the `set_tcb_obj_ref` abbreviation is that it's a little nicer to type out in a spec, so I wouldn't mind if people really do just forget about it. But I'm open to other arguments for why it should be a regular abbreviation and then seen in proof states. Maybe if there are no other good arguments soon, I can make in `(input)` and we can merge this when AARCH64 AInvs is about to be ready.

proof/invariant-abstract/IpcDet_AI.thy
```diff
@@ -360,17 +360,9 @@ lemma receive_ipc_preamble_st_tcb_at:
   by (wpsimp simp: receive_ipc_preamble_def
                wp: cancel_ipc_st_tcb_at hoare_vcg_if_lift2 get_simple_ko_wp)
 
-global_interpretation set_tcb_obj_ref: non_reply_op "set_tcb_obj_ref f ref new"
-  apply unfold_locales
-  apply (wpsimp simp: set_tcb_obj_ref_def wp: set_object_wp)
-  by (clarsimp simp: reply_at_ppred_def obj_at_def get_tcb_def
-              split: option.splits kernel_object.splits)
-
 global_interpretation set_tcb_obj_ref: non_ntfn_op "set_tcb_obj_ref f ref new"
```
corlewis: Out of curiosity, do you remember why you were able to remove the `non_reply_op` interpretation but kept this `non_ntfn_op` one?

corlewis (approved):
So much nicer, and thank you for all of the cleanup done along the way!

proof/invariant-abstract/IpcDet_AI.thy
michaelmcinerney: No, sorry, can't remember. I just saw the blue warning message and happily deleted it without thinking much.

corlewis: Is it worth trying to remove the second one as well then? Not a big deal but it feels a bit strange to keep one and not the other.

michaelmcinerney: The assumptions of the `non_reply_op` locale had already been shown by interpreting the `tcb_op` locale for the constant `thread_set`, but nothing had been done for `non_ntfn_op` for `thread_set` before this point. I have made an interpretation for `non_ntfn_op` for `thread_set` where the `tcb_op` locale is interpreted for `thread_set`, so it's a little more streamlined now.

While these locales work fine, I don't think things are done systematically. I do quite like having a `setObject` in Refine depend on the type, and I think ultimately, it would be great if we had some sort of general `getObject/setObject/updateObject` thing for every type, and then instead of having `obj_at'`, we exclusively use the `opt_pred/opt_map` style definitions with the heap projections (along with `pspace_aligned'/pspace_distinct'` for the other bits of `obj_at'`). And then have all that at the AInvs level, too. But this is just wishful thinking at this point.

PR #1029, opened 2026-07-13
arch-split Arch_R and Syscall_R

Xaphiosis:
🦆🦆🦆 Interface changes isolated again, and an extra commit for adding dummy `vs_valid_duplicates'` which we need for interfacing. This PR is a bit strange because we're "arch-splitting" Arch_R which doesn't make sense the usual way. I put the interface and whatever is generic (for now) in Arch_R, and then ArchArch_R has the rest (naming mirrors AInvs). The Arch_R work doesn't do all the fake commit juggling, because it's not a proper arch-split. Syscall_R does do the normal commit juggling like you're used to by now. Again, please review commit-by-commit.

The infoflow refinement fixes are a bit special, because when stuff broke, it turned out this was already what @ryybrr was working on as part of AArch64 infoflow and needing to be properly arch-generic there. I took only the minimum of the changes needed to make it go. @ryybrr please check you're ok with that commit.

**Instructions:**

In real life, I did the usual addition of Syscall_R, hierarchy update, arch-split and then update. However, since the arch-split part first splits AARCH64, then takes the AARCH64 version of Syscall_R and only fixes what broke, we ended up with a small(er) diff there (again).

- Arch_R commits: first given arch prefix, then extraction of generic stuff plus adding an interface which will be used by Syscall_R
- `refine: adjust arch interfaces for Syscall_R arch-split` and `add dummy vs_valid_duplicates'` groups all the changes to earlier theories that were needed
- `[squash] PR: copy AARCH64 version of ..._R` - same as last time: stop `..._R` looking like a wall of green (no need to review)
- `refine: arch-split `..._R (AARCH64 ONLY)` - this is the main bulk of the arch-split, dealing with AARCH64
- updates in downstream theories: refine+crefine this time, but also infoflow refinement in a commit of its own since I'm borrowing some of Ryan's work

Stats: 75 files changed, 2297 insertions(+), 11666 deletions(-)
Total: 9369 lines removed


proof/infoflow/refine/ADT_IF_Refine_C.thy
```diff
@@ -428,14 +438,60 @@ lemma kernelEntry_corres_C:
...
+   apply (rule ccorres_add_return2)
+   apply (ctac (no_vcg) add: checkInterrupt_ccorres')
+    apply (rule_tac R="\<top>" in ccorres_return[where R'=UNIV])
+    apply (rule conseqPre, vcg)
+    apply (clarsimp simp: return_def)
+   apply wpsimp
+  apply simp
+  done
+
+(* FIXME AARCH64 IF: getActiveIRQ_neq_non_kernel should be exposed *)
```
ryybrr: Can remove this comment

ryybrr (approved):

proof/refine/AARCH64/ArchArch_R.thy
```diff
@@ -434,6 +369,7 @@ where
    | InvokeSGISignal _ \<Rightarrow> \<top>
    | InvokeSMCCall _ \<Rightarrow> \<top>"
 
+(* FIXME arch-split: do we have anything about this on the other include chain? *)
 lemma mask_vmrights_corres:
```
lsf37: What's the other include chain?

proof/refine/ARM_HYP/ArchArch_R.thy
```diff
@@ -828,13 +744,14 @@ lemma decodeARMVCPUInvocation_corres:
...
   apply (rename_tac vcpui)
-  apply (case_tac vcpui; simp split del: if_split)
+  apply (case_tac vcpui; simp)
       (* set_tcb *)
       apply (simp add: decode_vcpu_set_tcb_def decodeVCPUSetTCB_def Let_def isVCPUCap_def)
       apply (cases excaps; simp add: null_def)
       apply (frule list_all2_Cons)
       apply clarsimp
       apply (case_tac a; clarsimp simp add: cap_relation_def)
+      supply if_split[split] (* FIXME: this avoids corresK if_split not in set warnings *)
```
lsf37, line 754: Thanks for the FIXME, that's probably something we should look at in corresK now that we can suppress such warnings.

proof/refine/RISCV64/ArchArch_R.thy
```diff
@@ -1053,6 +946,17 @@ lemma decode_page_inv_wf[wp]:
                            vmsz_aligned_user_region not_le))
   done
 
+lemma arch_cap_exhausted:
+  "\<lbrakk>\<not> isFrameCap param_e; \<not> isPageTableCap param_e; \<not> isASIDControlCap param_e; \<not> isASIDPoolCap param_e\<rbrakk>
+    \<Longrightarrow> undefined \<lbrace>P\<rbrace>"
```
lsf37: Is it worth cleaning up the `param_e` into `cap` while we're here? These lemmas always look a bit strange, but they really are just `==> P` or `==> False` (which we don't want to write, because we want `crunch` to match the conclusion)

proof/refine/X64/ArchArch_R.thy
```diff
@@ -2196,6 +2058,31 @@ lemma arch_performInvocation_invs':
...
+             det_getRestartPC asUser_inv
+          | wpcw)+
+  apply simp
+  done
+
+end (* Arch *)
+
+arch_requalify_consts
+  valid_arch_inv'
+  archinv_relation
```
lsf37, line 2080: Shouldn't we be doing that in the theory that includes this? (Syscall_R I guess)

proof/refine/Arch_R.thy
```diff
@@ -12,4 +12,245 @@ theory Arch_R
...
 begin
 
+arch_requalify_facts
+  arch_prepare_set_domain_pspace_aligned (* FIXME arch-split: ArchTcb_AI *)
+  arch_prepare_set_domain_pspace_distinct (* FIXME arch-split: ArchTcb_AI *)
+  arch_prepare_set_domain_typ_ats (* FIXME arch-split: ArchTcb_AI *)
+lemmas [wp] =
+  arch_prepare_set_domain_pspace_aligned
+  arch_prepare_set_domain_pspace_distinct
+  arch_prepare_set_domain_typ_ats(3) (*want tcb_at*)
```
lsf37: This one looks somewhat dangerous for maintenance. The index might change and then it'll take us quite a while to figure out it is here where it broke. Is there a sanity check we could do against `tcb_at`?

lsf37: Or alternatively, would it be Ok to have all `arch_prepare_set_domain_typ_ats` as `[wp]`? Usually that's what `typ_ats` are designed for, so it should be safe.

proof/refine/Arch_R.thy
```diff
@@ -12,4 +12,245 @@ theory Arch_R
...
+       (arch_perform_invocation ai) (Arch.performInvocation ai')"
+  assumes arch_decodeInvocation_corres:
+    "\<And>arch_cap arch_cap' excaps excaps' slot mi args cptr'.
+     \<lbrakk> acap_relation arch_cap arch_cap';
+       list_all2 cap_relation (map fst excaps) (map fst excaps');
+       list_all2 (\<lambda>s s'. s' = cte_map s) (map snd excaps) (map snd excaps') \<rbrakk> \<Longrightarrow>
+     corres
+     (ser \<oplus> archinv_relation)
+     (invs and valid_cap (cap.ArchObjectCap arch_cap) and
+          cte_wp_at ((=) (cap.ArchObjectCap arch_cap)) slot and
```
lsf37: indent is a bit off on this line

proof/refine/AARCH64/ArchArch_R.thy
Xaphiosis: I should remove this comment. Arch_R is included in parallel to Untyped_R, so doesn't see Tcb_R etc. I wondered whether there was something I couldn't see that I was duplicating.

proof/refine/RISCV64/ArchArch_R.thy
Xaphiosis: I literally have no clue what this is doing, I just moved it and maybe fixed indent. Looking at it more closely, yeah, the conclusion should be `False`. Can fix, unless someone is abusing this by using it as a wp rule.

proof/refine/X64/ArchArch_R.thy
Xaphiosis: We can, but:
- they have nothing to do with Syscall_R
- if I move them to Syscall_R, I will have to change the parameters I pass to the Arch_R interpretation to have specific arch prefixes.

Thoughts?

proof/refine/Arch_R.thy
Xaphiosis: Yes, that should be ok. I think I got confused because in AInvs, typ_ats is the generic one, whereas in Refine, `gen_typ_ats` is the generic one. If anything breaks it should be easy enough to fix.

proof/infoflow/refine/ADT_IF_Refine_C.thy
Xaphiosis: will do

proof/refine/AARCH64/ArchSyscall_R.thy
```diff
@@ -422,1376 +25,39 @@ lemma vcpuFlushIfCurrent_corres[corres]:
...
-     apply (rule hoare_vcg_conj_lift)
-      apply (rule tcbSchedAppend_all_invs_but_ct_not_inQ')
-     apply wpsimp
-    apply wpsimp
-   apply wpsimp
-  apply (simp add:ct_active_runnable'[unfolded ct_in_state'_def])
-  done
+(* FIXME arch-split: consider moving to where other msgRegisters stuff goes... Tcb_R? Ipc_R? *)
+lemma len_msg_registes_le_max_length[Syscall_R_assms]:
+  "length msg_registers \<le> msg_max_length"
```
lsf37: It could even go into AInvs, since it only talks about ASpec stuff.

lsf37 (approved):
Looking good!

proof/refine/X64/ArchArch_R.thy
lsf37: > * if I move them to Syscall_R, I will have to change the parameters I pass to the Arch_R interpretation to have specific arch prefixes.

I hadn't thought about that one. Yes, makes sense to keep them here then, because the interpretation also forces them to exist, so there is no risk that we'd forget one of them in a change somewhere.

proof/refine/RISCV64/ArchArch_R.thy
lsf37: It's for the crunch just below, I think (hence also the `param_e`). I don't think we want to change it to `==> False`, because the crunch will then refuse to match. We could make the `undefined` a variable instead if we wanted, though.

proof/refine/Arch_R.thy
```diff
@@ -12,4 +12,245 @@ theory Arch_R
...
+lemma sts_cte_cap_to'[wp]:
+  "\<lbrace>ex_cte_cap_to' p\<rbrace> setThreadState st t \<lbrace>\<lambda>rv. ex_cte_cap_to' p\<rbrace>"
+  by (wp ex_cte_cap_to'_pres)
+
+crunch setMRs
+  for nosch[wp]: "\<lambda>s. P (ksSchedulerAction s)"
+    (ignore: getRestartPC setRegister transferCapsToSlots
+   wp: hoare_drop_imps hoare_vcg_split_case_option
+        mapM_wp'
+   simp: split_def zipWithM_x_mapM)
```
corlewis, line 154: This old indentation is bad enough that I couldn't ignore it. If you have a chance, thoughts on fixing it while we're here?

proof/refine/AARCH64/ArchArch_R.thy
```diff
@@ -497,35 +433,15 @@ lemma ARMMMU_improve_cases:
...
-   apply (rule nonec)
-  apply simp
-  apply (rule somec)
-  done
-
-
-lemma cap_relation_Untyped_eq:
-  "cap_relation c (UntypedCap d p sz f) = (c = cap.UntypedCap d p sz f)"
-  by (cases c) auto
+lemmas Arch_decodeInvocation_inv = ArchRetypeDecls_H_AARCH64_H_decodeInvocation_inv
```
corlewis: Have we decided on a preference on this approach for nicer crunch names, or the alternative done in a previously arch-split theory where you added an abbreviation for the constant? In this case I guess that would be adding
```
abbreviation (input)
  "Arch_decodeInvocation \<equiv> Arch.decodeInvocation"
```
and crunching that instead of `AARCH64_H.decodeInvocation`.

proof/refine/ARM_HYP/PageTableDuplicates.thy
```diff
@@ -1265,7 +1265,7 @@ lemma unmapPage_valid_duplicates'[wp]:
           apply (rule_tac ptr = "p && ~~ mask ptBits" and word = p
             in mapM_x_storePTE_update_helper[where sz = 7])
           apply (wp checkMappingPPtr_inv lookupPTSlot_page_table_at'
-                    Arch_R.lookupPTSlot_aligned | simp)+
+                    ARM_HYP.lookupPTSlot_aligned | simp)+
```
corlewis: Do you remember what's going on here? Are there multiple versions of `lookupPTSlot_aligned`? It makes sense that the name has changed but I'm surprised that the lemma needs to be qualified at all.

proof/refine/AARCH64/ArchTcbAcc_R.thy
```diff
@@ -774,35 +728,6 @@ lemma setBoundNotification_state_hyp_refs_of'[TcbAcc_R_2_assms, wp]:
   by (simp add: setBoundNotification_def fun_upd_def
         | wp threadSet_state_hyp_refs_of')+
 
-lemma tcbSchedNext_update_iflive'[TcbAcc_R_2_assms]:
-  "\<lbrace>\<lambda>s. if_live_then_nonz_cap' s \<and> ex_nonz_cap_to' t s\<rbrace>
-   threadSet (tcbSchedNext_update f) t
-   \<lbrace>\<lambda>_. if_live_then_nonz_cap'\<rbrace>"
```
corlewis, line 780: Very nice that you were able to make all of these `threadSet` lemmas generic

proof/refine/ARM_HYP/PageTableDuplicates.thy
```diff
@@ -37,7 +37,7 @@ crunch setupReplyMaster
   (wp: crunch_wps simp: crunch_simps)
 
 
-(* we need the following lemma in Syscall_R *)
+(* we need the following lemma in ArchSyscall_R *)
 crunch getRegister
   for inv[wp]: "P"
```
corlewis, line 42: It isn't new but I'm confused by this comment. This has always come after `Syscall_R` hasn't it?

proof/refine/Syscall_R.thy
```diff
@@ -11,7 +11,220 @@ theory Syscall_R
...
+  assumes prepareSetDomain_sch_act_simple[wp]:
+    "\<And>t d. prepareSetDomain t d \<lbrace>sch_act_simple\<rbrace>"
+  assumes prepareSetDomain_tcb_at'[wp]:
+    "\<And>t d p. prepareSetDomain t d \<lbrace>tcb_at' p\<rbrace>"
+  assumes postSetFlags_typ_at'[wp]:
+    "\<And>t flags P T p. postSetFlags t flags \<lbrace>\<lambda>s. P (typ_at' T p s)\<rbrace>"
+  assumes arch_performIRQControl_typ_at'[wp]:
+    "\<And>inv P T p. Arch.performIRQControl inv \<lbrace>\<lambda>s. P (typ_at' T p s)\<rbrace>"
+  assumes Arch_performIRQControl_inv_EE[wp]:
+    "\<And>irqc P. \<lbrace>\<top>\<rbrace> Arch.performIRQControl irqc -, \<lbrace>P\<rbrace>"
```
corlewis, line 189: Not new but it looks like we're inconsistent with naming of lemmas of these `Arch.` constants. I'm guessing we prefer the versions starting with a lowercase a?

proof/refine/Syscall_R.thy
```diff
@@ -11,7 +11,220 @@ theory Syscall_R
...
+  assumes inv_arch_IRQInactive:
+    "\<And>invocation. \<lbrace>\<top>\<rbrace> Arch.performInvocation invocation -, \<lbrace>P\<rbrace>"
```
corlewis: Old but this name doesn't look right. `arch_performInvocation_inv`?

proof/refine/Syscall_R.thy
```diff
@@ -1110,20 +1150,28 @@ lemma setDomain_invs':
...
+                              \<and> (ptr \<noteq> curThread \<longrightarrow> ct_not_inQ s \<and> sch_act_not ptr s)"
+                 in hoare_strengthen_post[rotated])
      apply (clarsimp simp:invs'_def valid_state'_def)
     apply (wp hoare_vcg_imp_lift)+
   apply (clarsimp simp:invs'_def valid_pspace'_def valid_state'_def valid_domain_inv'_def)+
   done
 
+context Arch begin arch_global_naming
+
 crunch prepareSetDomain
```
corlewis: Do we want the usual comment here about this being the same on all arches?

proof/refine/Syscall_R.thy
```diff
@@ -1646,54 +1702,54 @@ lemma handleRecv_isBlocking_corres':
...
+                     prefer 3 (* throwError *)
+                     apply (rule corres_trivial)
+                     apply (clarsimp simp: lookup_failure_map_def)+
+                  apply (wp get_simple_ko_wp getNotification_wp | wpcw | simp)+
+                apply clarsimp+
+                apply (clarsimp simp: valid_cap_def ct_in_state_def)
+               apply (clarsimp simp: valid_cap'_def capAligned_def)
+              apply (clarsimp simp: lookup_failure_map_def)
+           (* neither endpoint nor notification cap, so clear up throwError case *)
+             apply (case_tac rv; clarsimp simp: lookup_failure_map_def)
```
corlewis: nitpick: indentation

proof/refine/Syscall_R.thy
```diff
@@ -1957,25 +1958,29 @@ lemma hr_ct_active'[wp]:
...
 
 lemma hc_invs'[wp]:
-  "\<lbrace>invs' and
-      (\<lambda>s. ksSchedulerAction s = ResumeCurrentThread) and
-      ct_active'\<rbrace>
-     handleCall
+  "\<lbrace>invs' and (\<lambda>s. vs_valid_duplicates' (ksPSpace s)) and
+      (\<lambda>s. ksSchedulerAction s = ResumeCurrentThread) and ct_active'\<rbrace>
+   handleCall
    \<lbrace>\<lambda>rv. invs'\<rbrace>"
```
corlewis, line 1964: Since we're already improving the indentation of this lemma, shouldn't the second line of the precondition be indented less?

proof/refine/ARM/orphanage/Orphanage.thy
```diff
@@ -2084,6 +2084,12 @@ lemma sts_tcb_at'_preserve':
 crunch handleSpuriousIRQ
   for no_orphans[wp]: no_orphans
 
+lemma hv_inv':
+  "\<lbrace>P\<rbrace> handleVMFault p t \<lbrace>\<lambda>_. P\<rbrace>"
+  unfolding ARM_H.handleVMFault_def
+  by (wpsimp wp: dmo_inv' getDFSR_inv getFAR_inv getIFSR_inv getRestartPC_inv det_getRestartPC
+                 asUser_inv)
```
corlewis, line 2091: Feels a bit unexpected that this is needed now. Maybe a crunch used to produce it and now doesn't?

corlewis (approved):
Very nice!

proof/refine/ARM/orphanage/Orphanage.thy
Xaphiosis: It used to be produced on some architectures in the old Syscall_R. I can move it back to ArchSyscall_R for those architectures (AARCH64 does not have this), but it's definitely not generic, and only used in Orphanage. Do I move it, or leave it here until Orphanage is arch-split?

proof/refine/Syscall_R.thy
Xaphiosis: Um, as far as I can tell for Refine I've been going with uppercase `Arch_` for the renames from crunch, and I no longer remember if that was on purpose or now. It seems on AInvs, `arch_` might be more popular. But now thinking about it more, when writing a lemma, it feels silly to write `lemma Arch_...`, so the `arch_performIRQControl_typ_at'` made sense. Except now they aren't consistent. Whoops.
Any thoughts/advice?

proof/refine/ARM_HYP/PageTableDuplicates.thy
Xaphiosis: I only adjusted the comment. Honestly, this lemma shouldn't even be in PageTableDuplicates

proof/refine/ARM_HYP/PageTableDuplicates.thy
Xaphiosis: Good point. I will check.

proof/refine/AARCH64/ArchArch_R.thy
Xaphiosis: I think the abbreviation (input) is definitely good when it's used multiple times, because multiple lots of `ArchRetypeDecls_H_AARCH64_H_` hurt to look at. For a one-off, adding an abbreviation is less clear. I've forgotten already whether this one's used multiple times.

corlewis: I think this one is only crunched once, although the constant is referred to in a couple of other lemmas.

corlewis: And oh, this has reminded me of the other things I wasn't sure about. In this crunch we use `AARCH64_H.decodeInvocation`, while I think the pattern from others would be `Arch.decodeInvocation`. It's minor but I think the `Arch` version would be nicer, since then the text is more consistent between architectures and it might even be able to be moved to a generic file.

proof/refine/Syscall_R.thy
corlewis: Hmm, I'm actually not sure which one I prefer but if nothing else it looks like `lemma arch_` is a lot more common than `lemma Arch_`, so I'd be tempted to go that way for consistency.

proof/refine/ARM/orphanage/Orphanage.thy
corlewis: Right, that makes sense. Probably fine to leave in Orphanage then.

proof/refine/Arch_R.thy
Xaphiosis: will do

proof/refine/AARCH64/ArchArch_R.thy
Xaphiosis: I'll change it to `Arch.` and use `arch_` in this file. I guess for other files where I was inconsistent I'll have to take down a note for second pass.

proof/refine/Syscall_R.thy
Xaphiosis: will go with `arch_` then

proof/refine/Syscall_R.thy
Xaphiosis: will change

proof/refine/Syscall_R.thy
Xaphiosis: Wow, thanks for spotting that. It turns out that's a mistake on my part which happened to work. These should've been interfaced.
So I should probably go and interface them, except `pred_tcb_at'` can't be interfaced so I'll have to think of something.

proof/refine/Syscall_R.thy
Xaphiosis: On my screen it looks like the parenthesis is exactly under the `v` of `invs`, meaning the second line is indented by two from `invs`. I thought for right-operator wrapping indent by two was standard?

lsf37: No, in right-operator style, an `and` chain should be

```
"⦃invs' and (λs. vs_valid_duplicates' (ksPSpace s)) and
  (λs. ksSchedulerAction s = ResumeCurrentThread) and ct_active'⦄
```

(Edit: the `(` should be under `i`, I can't seem to make that happen on GitHub)

proof/refine/RISCV64/ArchArch_R.thy
Xaphiosis: I changed it to `acap`, but indeed the crunch needs that specific form.

proof/refine/Syscall_R.thy
Xaphiosis: should be done now

proof/refine/ARM_HYP/PageTableDuplicates.thy
Xaphiosis: yup, removed qualification

proof/refine/AARCH64/ArchArch_R.thy
```diff
@@ -433,12 +432,12 @@ lemma ARMMMU_improve_cases:
...
   done
 
-crunch AARCH64_H.decodeInvocation
+crunch Arch.decodeInvocation
   for inv[Arch_R_assms, wp]: "P"
   (wp: crunch_wps mapME_x_inv_wp getASID_wp hoare_vcg_imp_lift'
    simp: crunch_simps ARMMMU_improve_cases)
 
-lemmas Arch_decodeInvocation_inv = ArchRetypeDecls_H_AARCH64_H_decodeInvocation_inv
+lemmas arch_decodeInvocation_inv = ArchRetypeDecls_H_AARCH64_H_decodeInvocation_inv
```
corlewis, line 440: Huh, I thought that this would end up `ArchRetypeDecls_H_Arch_decodeInvocation_inv` on the right. I guess `crunch` still produces this qualified form of the name internally, regardless of what is passed in.

proof/refine/Syscall_R.thy
corlewis: If it was working before with the same text for all architectures, couldn't we have kept it in the generic file? I guess in this case there isn't that much being duplicated across architectures, so maybe it doesn't make enough of a difference to bother.

Xaphiosis: Well, it *happened* to work across all architectures, without any brain usage as to what `prepareSetDomain` was actually doing. I prefer the "on all arches" to be in cases where it's very obvious. I guess the differentiator here is that I have no real idea about prepareSetDomain and everything it's supposed to be doing (e.g. flushing VCPU/FPU which doesn't exist on all arches?)

proof/refine/AARCH64/ArchArch_R.thy
Xaphiosis: Ahh, I think you forgot the link between this and why we use the `(input)` abbreviation if we have to interface multiple of these. Without the abbreviation, it always barfs out stuff like that. I think I asked once whether there's some way `crunch` could not do that, and we even talked about it, but I don't think we came up with anything resembling a decent solution.
Part of the problem is that the constant is `Arch.blah`, but the definition is `AARCH64_H.blah_def`. It would be nice to be able to give crunch a hint to get around this mess.

Xaphiosis:
Rebased and squashed.

lsf37 (approved):
Squashed commits look good to go when the tests have passed.

PR #1030, opened 2026-07-14
arm-hyp+aarch64: sync VCPU context switching

lsf37:
Minor spec tweak and trivial proof update to sync VCPU context switching with seL4 PR seL4/seL4#1318:

- context switch PAR on AArch64 and PARhigh/PARlow on AArch32
- remove ISR from VCPU context switch, because it is read only

Test with: https://github.com/seL4/seL4/pull/1318

### DIFF
--- a/proof/crefine/ARM_HYP/ArchMove_C.thy
+++ b/proof/crefine/ARM_HYP/ArchMove_C.thy
@@ -432,10 +432,9 @@ crunch readVCPUReg
 crunch readVCPUReg
   for ksCurThread[wp]: "\<lambda>s. P (ksCurThread s)"
 
-(* schematic_goal leads to Suc (Suc ..) form only *)
-lemma fromEnum_maxBound_vcpureg_def:
-  "fromEnum (maxBound :: vcpureg) = 43"
-  by (clarsimp simp: fromEnum_def maxBound_def enum_vcpureg)
+schematic_goal fromEnum_maxBound_vcpureg_def:
+  "fromEnum (maxBound :: vcpureg) = numeral ?n"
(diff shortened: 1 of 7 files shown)


proof/crefine/ARM_HYP/ArchMove_C.thy
```diff
@@ -434,7 +434,7 @@ crunch readVCPUReg
 
 (* schematic_goal leads to Suc (Suc ..) form only *)
 lemma fromEnum_maxBound_vcpureg_def:
-  "fromEnum (maxBound :: vcpureg) = 43"
+  "fromEnum (maxBound :: vcpureg) = 45"
```
lsf37: I tried looking at this again, but still didn't find a good way to go from `Suc*` to `numeral ?n`. Everything is geared towards the other direction.

Xaphiosis: possibly related:
```
schematic_goal enum_apiobject_type_length:
  "length (enum :: apiobject_type list) = numeral ?n"
  by (simp add: enum_apiobject_type flip: One_nat_def)
```

Xaphiosis (approved):
If you want to have another go at the Suc enum lemma issue, might be interesting. Otherwise good to go.

proof/crefine/ARM_HYP/ArchMove_C.thy
lsf37: I will definitely try that.

lsf37: Wow, that works. The additional key thing not to do is `clarsimp`, because that will refuse to instantiate. But flipping `One_nat_def` seems to be doing what I was looking for. Thanks for finding that.

PR #1031, opened 2026-07-16
aarch64 machine: sync kernelELFBase with seL4

lsf37:
seL4/seL4#1516 provides a fixed virtual address for the kernel ELF region. Previously the virtual address of the kernel ELF region was dependent on physBase, and the offset was fixed.

Test with: seL4/seL4#1516


Indanz:
After this change it should be fairly straightforward to make `kernelELFPAddrBase` variable, so that we don't have to hard-code the kernel load address. It looks like the main requirement from the proofs is that it's aligned. That could be solved by `kernelELFPAddrBase = (kernelELFLoadAddr & alignment_mask)` or `kernelELFPAddrBase = kernelELFLoadAddrHighbits << alignment` and a check in boot code to ensure it is (before MMU config/enable). It would be nice if the proofs supported semi-constant values which are determined at early boot and then stay constant, that would make it easier to not hard-code compile-time constants like config values and other platform settings.

lsf37:
> After this change it should be fairly straightforward to make `kernelELFPAddrBase` variable, so that we don't have to hard-code the kernel load address. It looks like the main requirement from the proofs is that it's aligned. 

The main proof obligation is that, yes. I'm not entirely sure if there aren't some others in there. We've been a lot more disciplined on AArch64 than in the past, so chances are good that this is it, but if it's important to know now, I can double check.


> That could be solved by `kernelELFPAddrBase = (kernelELFLoadAddr & alignment_mask)` or `kernelELFPAddrBase = kernelELFLoadAddrHighbits << alignment` and a check in boot code to ensure it is (before MMU config/enable). It would be nice if the proofs supported semi-constant values which are determined at early boot and then stay constant, that would make it easier to not hard-code compile-time constants like config values and other platform settings.

Yes, making it runtime constant (e.g. determined at boot, then unchanging) changes the model and first needs the new environment monads that we have been developing for multikernel, so it's still a bit early. They do exist now though, and we're working towards merging those in. That'll still be a while, but after that it should not be hard to do this.

Xaphiosis (approved):

PR #1033, opened 2026-07-19
crefine: adjust for ptr name mangling change

lsf37:
seL4 PR seL4/seL4#728 removes a use of the pointer name `p` under a different type and changes the mangled name.

While this was already robust for references in Simpl, the pointer name produced by vcg was still dependent on order and other types, because people relied and generated names and not use rename_tac.

Make this robust against internal name changes as it should have been for the start.

Test with: seL4/seL4#728

(this should also work with current master)


lsf37:
I seem to have missed another occurrence, will fix up.

(Edit: now done)

corlewis (approved):

PR #1035, opened 2026-07-20
arm spec+proof: stale cap check for PageFlush

lsf37:
Sync specs and proof with seL4 PR seL4/seL4#1633 which aborts the flush operation when the cap that is used to invoke it has stale mapping info.

When the check passes, we cannot guarantee that the cap is the original authority to this mapping, but we do know that it is at least authority to the same frame.

Test with: seL4/seL4#1633

### DIFF
--- a/proof/crefine/ARM/Arch_C.thy
+++ b/proof/crefine/ARM/Arch_C.thy
@@ -2145,6 +2145,10 @@ where
                       x (to_option (to_bool \<circ> resolve_ret_C.valid_C) y)"
 
 
+lemma to_option_SomeD:
+  "to_option f x = Some y \<Longrightarrow> f x \<and> y = x"
+  by (simp add: to_option_def split: if_split_asm)
+
 lemma resolve_ret_rel_None[simp]:
   "resolve_ret_rel None y = (valid_C y = scast false)"
   by (clarsimp simp: resolve_ret_rel_def to_option_def to_bool_def split: if_splits)
@@ -2297,6 +2301,7 @@ lemma decodeARMFrameInvocation_ccorres:
        (Call decodeARMFrameInvocation_'proc)"
   supply if_cong[cong] option.case_cong[cong] tl_drop_1[simp]
   apply (clarsimp simp only: isCap_simps)
+  apply (rename_tac dev p rights vmsz m_opt)
   apply (cinit' lift: invLabel_' length___unsigned_long_' cte_' current_extra_caps_' cap_' buffer_' call_'
                 simp: decodeARMMMUInvocation_def decodeARMPageFlush_def)
    apply (simp add: Let_def isCap_simps invocation_eq_use_types split_def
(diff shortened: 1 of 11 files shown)


proof/crefine/ARM/Arch_C.thy
```diff
@@ -2145,6 +2145,10 @@ where
                       x (to_option (to_bool \<circ> resolve_ret_C.valid_C) y)"
 
 
+lemma to_option_SomeD:
+  "to_option f x = Some y \<Longrightarrow> f x \<and> y = x"
+  by (simp add: to_option_def split: if_split_asm)
```
Xaphiosis, line 2150: FIXME move candidate

spec/haskell/src/SEL4/Kernel/VSpace/ARM.lhs
```diff
@@ -1450,6 +1450,13 @@ The function should not be called on any other cap types.
 >             throw $ InvalidArgument 1
 >         let pageSize = bit (pageBitsForSize (capVPSize cap))
 >         let pageBase = addrFromPPtr $ capVPBasePtr cap
+#ifndef CONFIG_ARM_HYPERVISOR_SUPPORT
+>         frameInfo <- withoutFailure $ resolveVAddr pd vaddr
```
Xaphiosis, line 1461: any benefit to adding a comment here (or in the abstract spec) about what's happening here? we know right now from the PR and commit message, but in future we won't

proof/crefine/ARM/Arch_C.thy
lsf37: You'd think so, but the definition of it is just above, so this is the highest point it can move. The whole thing should be moved, though (and does already have a FIXME for it)

spec/haskell/src/SEL4/Kernel/VSpace/ARM.lhs
lsf37: Good idea, yes. Will add.

proof/infoflow/ARM/ArchDecode_IF.thy
```diff
@@ -193,6 +193,19 @@ lemma decode_sgi_signal_invocation_reads_respects_f[wp]:
   "reads_respects_f aag l \<top> (decode_sgi_signal_invocation (SGISignalCap irq target))"
   by (wpsimp simp: decode_sgi_signal_invocation_def)
 
+lemma is_subject_asid_trans:
+  "\<lbrakk> is_subject_asid aag x; pas_refined aag s;
+     (pasASIDAbs aag x, Control, pasObjectAbs aag y) \<in> pasPolicy aag \<rbrakk>
+     \<Longrightarrow> is_subject aag y"
```
Xaphiosis: indentation

proof/crefine/ARM/Arch_C.thy
```diff
@@ -2371,46 +2376,127 @@ lemma decodeARMFrameInvocation_ccorres:
...
+                                      2 ^ pageBitsForSize vmsz \<le> gsMaxObjectSize s" and
+                               P'="\<lbrace>\<acute>ksCurThread = tcb_ptr_to_ctcb_ptr thread \<and>
+                                     flushtype_relation (labelToFlushType label) (ucast label)\<rbrace>"
+                               in ccorres_inst)
+               apply (case_tac resolve_rv; clarsimp)
+                apply (simp add: injection_handler_throwError)
+                apply (rule ccorres_symb_exec_r2)
+                  apply (rule ccorres_cond_true_seq)
+                  apply ccorres_rewrite
+                  apply (rule ccorres_cond_true_seq)
```
Xaphiosis, line 2418: using cond_true/cond_false could use a comment as to what's going on, or at least a comment somewhere that we're dealing with checking for a stale cap

proof/crefine/ARM/Arch_C.thy
```diff
@@ -2371,46 +2376,127 @@ lemma decodeARMFrameInvocation_ccorres:
...
+                                      thread \<noteq> ksIdleThread s \<and>
+                                      (obj_at' tcbQueued thread s \<longrightarrow> st_tcb_at' runnable' thread s) \<and>
+                                      fst (the m_opt) \<le> mask asid_bits \<and>
+                                      is_aligned p (pageBitsForSize vmsz) \<and>
+                                      hd args + snd (the m_opt) && mask cacheLineBits =
+                                        addrFromPPtr p + hd args && mask cacheLineBits \<and>
+                                      2 ^ pageBitsForSize vmsz \<le> gsMaxObjectSize s" and
+                               P'="\<lbrace>\<acute>ksCurThread = tcb_ptr_to_ctcb_ptr thread \<and>
+                                     flushtype_relation (labelToFlushType label) (ucast label)\<rbrace>"
+                               in ccorres_inst)
```
Xaphiosis, line 2409: are you sure this is the right way? this looks like a nuke

Xaphiosis:
second commit message:
`know that it is at least authority` -> `... an authority`

proof/crefine/ARM/Arch_C.thy
lsf37: I have looked at it again this morning, and I managed to pull out the first `if` on the abstract side, but it then descends into chaos again. I can have another look at it again after the release. I agree that it should possible to do this without instantiation, just haven't found the right knob to turn yet.

proof/infoflow/ARM/ArchDecode_IF.thy
lsf37: Serves me right for just copying this out of AARCH64 , where I had complained about exactly this :-)

lsf37:
Feedback addressed (I tweaked the commit message slightly differently).

proof/crefine/ARM/Arch_C.thy
Xaphiosis: maybe add FIXME then?

Xaphiosis (approved):

PR #1036, opened 2026-07-22
Let deployment jobs finish when new commits are pushed

lsf37:
- Don't cancel deployment jobs that have started when a new commit is pushed. Instead let it finish so that the job in the new commit can benefit from the images built in the first one. This would have sped up the seL4 release cycle by at least one full proof run today (maybe two).

- Make the concurrency group ID that controls this more robust in both deployment and PR jobs. The previous strategy.job-index was dependent on the order GitHub expanded the matrix in.


lsf37:
(Should be merged after the release has finished)

Xaphiosis (approved):

PR #1038, opened 2026-07-26

corlewis:
* refine+crefine: arch-split Detype_R
* refine: expose threadSet_invs_trivialT as generic interface
* refine: arch-split IpcCancel_R
* cspec/c: extract maxIRQ for RISCV64
* refine: arch-split Init_R
* refine: arch-split InterruptAcc_R
* refine+crefine: arch-split Finalise_R
* refine: improve TcbAcc_R arch-split
* refine+crefine: arch-split Ipc_R
* refine+crefine: arch-split CNodeInv_R
* refine+crefine: arch-split Tcb_R
* spec+proofs: consolidate maxIRQ definition in Platform.thy
* aarch64: remove plic_complete_claim
* refine: arch-split Interrupt_R
* monads: introduce opt_filter projection
* aarch64 aspec+ainvs: introduce asid_map
* refine+crefine: update proofs for asid_map addition
* aarch64: CONFIG_VTIMER_UPDATE_VOFFSET=OFF

As always with these merges, the main commit to review is the 'update after merge' one. Quite a lot of work went into the merge commit itself, trying to resolve conflicts in a way that minimised the overall effort, but I don't know how feasible that is to review. The final crefine commit has already been reviewed elsewhere and is only cherry-picked here to try to avoid merge-related git complications.

### DIFF
--- a/proof/crefine/AARCH64/Retype_C.thy
+++ b/proof/crefine/AARCH64/Retype_C.thy
@@ -203,13 +203,15 @@ lemma memzero_spec:
                                                (replicate (unat (n_' s - n_' t)) 0))
                                                       (t_hrs_' (globals s))\<rparr> }"
             and V1=undefined in subst [OF whileAnno_def])
+  apply (rename_tac s)
   apply vcg
     apply (clarsimp simp add: hrs_mem_update_def)
 
    apply clarsimp
-   apply (case_tac s, case_tac p___ptr_to_unsigned_char)
+   apply (rename_tac p n')
+   apply (case_tac s, case_tac p)
 
-   apply (subgoal_tac "8 \<le> unat na")
+   apply (subgoal_tac "8 \<le> unat n'")
     apply (intro conjI)
            apply (simp add: ptr_safe_def s_footprint_def s_footprint_untyped_def
                             typ_uinfo_t_def typ_info_word)
@@ -237,7 +239,7 @@ lemma memzero_spec:
     apply (subst to_bytes_machine_word_0)
     apply (rule heap_update_list_replicate)
      apply clarsimp
-    apply (rule_tac s="unat ((n - na) + 8)" in trans)
+    apply (rule_tac s="unat ((n - n') + 8)" in trans)
      apply (simp add: field_simps)
     apply (subst Word.unat_plus_simple[THEN iffD1])
      apply (rule is_aligned_no_overflow''[where n=3, simplified])
@@ -290,6 +292,7 @@ lemma memset_spec:
      apply (rule is_aligned_and_2_to_k, clarsimp simp: mask_def)
     apply (rule is_aligned_and_2_to_k, clarsimp simp: mask_def)
    apply clarsimp
+   apply (rename_tac p n')
    apply (intro conjI)
         apply (simp add: ptr_safe_def s_footprint_def s_footprint_untyped_def
                          typ_uinfo_t_def typ_info_word)
@@ -306,14 +309,14 @@ lemma memset_spec:
         apply assumption
        apply (erule order_trans [rotated])
        apply (simp add: lt1_neq0)
-      apply (case_tac p___ptr_to_unsigned_char, simp add: CTypesDefs.ptr_add_def unat_minus_one field_simps)
+      apply (case_tac p, simp add: CTypesDefs.ptr_add_def unat_minus_one field_simps)
      apply (metis word_must_wrap word_not_simps(1) linear)
     apply (erule order_trans[rotated])
     apply (clarsimp simp: ptr_val_case split: ptr.splits)
     apply (erule subsetD[OF intvl_sub_offset, rotated])
     apply (simp add: unat_sub word_le_nat_alt word_less_nat_alt)
    apply (clarsimp simp: ptr_val_case unat_minus_one hrs_mem_update_def split: ptr.splits)
-   apply (subgoal_tac "unat (n - (na - 1)) = Suc (unat (n - na))")
+   apply (subgoal_tac "unat (n - (n' - 1)) = Suc (unat (n - n'))")
     apply (erule ssubst, subst replicate_Suc_append)
     apply (subst heap_update_list_append)
     apply (simp add: heap_update_word8)
--- a/proof/crefine/ARM/Retype_C.thy
+++ b/proof/crefine/ARM/Retype_C.thy
@@ -161,9 +161,10 @@ lemma memzero_spec:
     apply (clarsimp simp add: hrs_mem_update_def)
 
    apply clarsimp
-   apply (case_tac s, case_tac p___ptr_to_unsigned_char)
+   apply (rename_tac p n')
+   apply (case_tac s, case_tac p)
 
-   apply (subgoal_tac "4 \<le> unat na")
+   apply (subgoal_tac "4 \<le> unat n'")
     apply (intro conjI)
            apply (simp add: ptr_safe_def s_footprint_def s_footprint_untyped_def
                             typ_uinfo_t_def typ_info_word)
@@ -191,7 +192,7 @@ lemma memzero_spec:
     apply (subst to_bytes_word32_0)
     apply (rule heap_update_list_replicate)
      apply clarsimp
-    apply (rule_tac s="unat ((n - na) + 4)" in trans)
+    apply (rule_tac s="unat ((n - n') + 4)" in trans)
      apply (simp add: field_simps)
     apply (subst Word.unat_plus_simple[THEN iffD1])
      apply (rule is_aligned_no_overflow''[where n=2, simplified])
@@ -250,6 +251,7 @@ lemma memset_spec:
      apply (rule is_aligned_and_2_to_k, clarsimp simp: mask_def)
     apply (rule is_aligned_and_2_to_k, clarsimp simp: mask_def)
    apply clarsimp
+   apply (rename_tac p n')
    apply (intro conjI)
         apply (simp add: ptr_safe_def s_footprint_def s_footprint_untyped_def
                          typ_uinfo_t_def typ_info_word)
@@ -266,14 +268,14 @@ lemma memset_spec:
         apply assumption
        apply (erule order_trans [rotated])
        apply (simp add: lt1_neq0)
-      apply (case_tac p___ptr_to_unsigned_char, simp add: CTypesDefs.ptr_add_def unat_minus_one field_simps)
+      apply (case_tac p, simp add: CTypesDefs.ptr_add_def unat_minus_one field_simps)
      apply (metis word_must_wrap word_not_simps(1) linear)
     apply (erule order_trans[rotated])
     apply (clarsimp simp: ptr_val_case split: ptr.splits)
     apply (erule subsetD[OF intvl_sub_offset, rotated])
     apply (simp add: unat_sub word_le_nat_alt word_less_nat_alt)
    apply (clarsimp simp: ptr_val_case unat_minus_one hrs_mem_update_def split: ptr.splits)
-   apply (subgoal_tac "unat (n - (na - 1)) = Suc (unat (n - na))")
+   apply (subgoal_tac "unat (n - (n' - 1)) = Suc (unat (n - n'))")
     apply (erule ssubst, subst replicate_Suc_append)
     apply (subst heap_update_list_append)
     apply (simp add: heap_update_word8)
--- a/proof/crefine/ARM_HYP/Retype_C.thy
+++ b/proof/crefine/ARM_HYP/Retype_C.thy
@@ -187,9 +187,10 @@ lemma memzero_spec:
     apply (clarsimp simp add: hrs_mem_update_def)
 
    apply clarsimp
-   apply (case_tac s, case_tac p___ptr_to_unsigned_char)
+   apply (rename_tac p n')
+   apply (case_tac s, case_tac p)
 
-   apply (subgoal_tac "4 \<le> unat na")
+   apply (subgoal_tac "4 \<le> unat n'")
     apply (intro conjI)
            apply (simp add: ptr_safe_def s_footprint_def s_footprint_untyped_def
                             typ_uinfo_t_def typ_info_word)
@@ -217,7 +218,7 @@ lemma memzero_spec:
     apply (subst to_bytes_word32_0)
     apply (rule heap_update_list_replicate)
      apply clarsimp
-    apply (rule_tac s="unat ((n - na) + 4)" in trans)
+    apply (rule_tac s="unat ((n - n') + 4)" in trans)
      apply (simp add: field_simps)
     apply (subst Word.unat_plus_simple[THEN iffD1])
      apply (rule is_aligned_no_overflow''[where n=2, simplified])
@@ -264,6 +265,7 @@ lemma memset_spec:
      apply (rule is_aligned_and_2_to_k, clarsimp simp: mask_def)
     apply (rule is_aligned_and_2_to_k, clarsimp simp: mask_def)
    apply clarsimp
+   apply (rename_tac p n')
    apply (intro conjI)
         apply (simp add: ptr_safe_def s_footprint_def s_footprint_untyped_def
                          typ_uinfo_t_def typ_info_word)
@@ -280,14 +282,14 @@ lemma memset_spec:
         apply assumption
        apply (erule order_trans [rotated])
        apply (simp add: lt1_neq0)
-      apply (case_tac p___ptr_to_unsigned_char, simp add: CTypesDefs.ptr_add_def unat_minus_one field_simps)
+      apply (case_tac p, simp add: CTypesDefs.ptr_add_def unat_minus_one field_simps)
      apply (metis word_must_wrap word_not_simps(1) linear)
     apply (erule order_trans[rotated])
     apply (clarsimp simp: ptr_val_case split: ptr.splits)
     apply (erule subsetD[OF intvl_sub_offset, rotated])
     apply (simp add: unat_sub word_le_nat_alt word_less_nat_alt)
    apply (clarsimp simp: ptr_val_case unat_minus_one hrs_mem_update_def split: ptr.splits)
-   apply (subgoal_tac "unat (n - (na - 1)) = Suc (unat (n - na))")
+   apply (subgoal_tac "unat (n - (n' - 1)) = Suc (unat (n - n'))")
     apply (erule ssubst, subst replicate_Suc_append)
     apply (subst heap_update_list_append)
     apply (simp add: heap_update_word8)
--- a/proof/crefine/RISCV64/Retype_C.thy
+++ b/proof/crefine/RISCV64/Retype_C.thy
@@ -203,13 +203,15 @@ lemma memzero_spec:
                                                (replicate (unat (n_' s - n_' t)) 0))
                                                       (t_hrs_' (globals s))\<rparr> }"
             and V1=undefined in subst [OF whileAnno_def])
+  apply (rename_tac s)
   apply vcg
     apply (clarsimp simp add: hrs_mem_update_def)
 
    apply clarsimp
-   apply (case_tac s, case_tac p___ptr_to_unsigned_char)
+   apply (rename_tac p n')
+   apply (case_tac s, case_tac p)
 
-   apply (subgoal_tac "8 \<le> unat na")
+   apply (subgoal_tac "8 \<le> unat n'")
     apply (intro conjI)
            apply (simp add: ptr_safe_def s_footprint_def s_footprint_untyped_def
                             typ_uinfo_t_def typ_info_word)
@@ -237,7 +239,7 @@ lemma memzero_spec:
     apply (subst to_bytes_machine_word_0)
     apply (rule heap_update_list_replicate)
      apply clarsimp
-    apply (rule_tac s="unat ((n - na) + 8)" in trans)
+    apply (rule_tac s="unat ((n - n') + 8)" in trans)
      apply (simp add: field_simps)
     apply (subst Word.unat_plus_simple[THEN iffD1])
      apply (rule is_aligned_no_overflow''[where n=3, simplified])
@@ -290,6 +292,7 @@ lemma memset_spec:
      apply (rule is_aligned_and_2_to_k, clarsimp simp: mask_def)
     apply (rule is_aligned_and_2_to_k, clarsimp simp: mask_def)
    apply clarsimp
+   apply (rename_tac p n')
    apply (intro conjI)
         apply (simp add: ptr_safe_def s_footprint_def s_footprint_untyped_def
                          typ_uinfo_t_def typ_info_word)
@@ -306,14 +309,14 @@ lemma memset_spec:
         apply assumption
        apply (erule order_trans [rotated])
        apply (simp add: lt1_neq0)
-      apply (case_tac p___ptr_to_unsigned_char, simp add: CTypesDefs.ptr_add_def unat_minus_one field_simps)
+      apply (case_tac p, simp add: CTypesDefs.ptr_add_def unat_minus_one field_simps)
      apply (metis word_must_wrap word_not_simps(1) linear)
     apply (erule order_trans[rotated])
     apply (clarsimp simp: ptr_val_case split: ptr.splits)
     apply (erule subsetD[OF intvl_sub_offset, rotated])
     apply (simp add: unat_sub word_le_nat_alt word_less_nat_alt)
    apply (clarsimp simp: ptr_val_case unat_minus_one hrs_mem_update_def split: ptr.splits)
-   apply (subgoal_tac "unat (n - (na - 1)) = Suc (unat (n - na))")
+   apply (subgoal_tac "unat (n - (n' - 1)) = Suc (unat (n - n'))")
     apply (erule ssubst, subst replicate_Suc_append)
     apply (subst heap_update_list_append)
     apply (simp add: heap_update_word8)
--- a/proof/crefine/X64/Retype_C.thy
+++ b/proof/crefine/X64/Retype_C.thy
@@ -203,13 +203,15 @@ lemma memzero_spec:
                                                (replicate (unat (n_' s - n_' t)) 0))
                                                       (t_hrs_' (globals s))\<rparr> }"
             and V1=undefined in subst [OF whileAnno_def])
+  apply (rename_tac s)
   apply vcg
     apply (clarsimp simp add: hrs_mem_update_def)
 
    apply clarsimp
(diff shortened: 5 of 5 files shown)


proof/refine/Invariants_H.thy
```diff
@@ -1904,17 +1904,19 @@ lemmas tcb_cte_cases_neqs_n =
...
 
+lemma tcb_cte_cases_simps[simp]:
+  "tcb_cte_cases (tcbCTableSlot << cteSizeBits) = Some (tcbCTable, tcbCTable_update)"
+  "tcb_cte_cases (tcbVTableSlot << cteSizeBits) = Some (tcbVTable, tcbVTable_update)"
+  "tcb_cte_cases (tcbIPCBufferSlot << cteSizeBits) = Some (tcbIPCBufferFrame, tcbIPCBufferFrame_update)"
+  "tcb_cte_cases (tcbFaultHandlerSlot << cteSizeBits) = Some (tcbFaultHandler, tcbFaultHandler_update)"
+  "tcb_cte_cases (tcbTimeoutHandlerSlot << cteSizeBits) = Some (tcbTimeoutHandler, tcbTimeoutHandler_update)"
+  by (simp add: tcb_cte_cases_neqs_n tcb_cte_cases_def tcbSlot_defs)+
+
+lemmas tcb_cte_cases_simps'[simp] = tcb_cte_cases_simps[simplified tcbSlot_defs shiftl_0]
```
corlewis: This change is a bit unrelated to the rest of the merge but it's something that I found made some of the proofs I needed to update bit nicer and easier to keep symbolic. I didn't include it here but I was also tempted to add a version that uses `_ * 2 ^ cteSizeBits` instead of `_ << cteSizeBits`, as that form seems to be produced by the simplifier quite often.

Assuming that none of this is contentious, I might copy this specific change back to master, to see if it makes any proofs nicer there and to keep the branches more in sync.

lsf37: I like the symbolic form, but I'd keep `<<` -- we should usually not automatically simplify to 2^n, it's just that a lot of the existing proofs do that. Better to remove the latter than the former.

proof/refine/RISCV64/ArchDetype_R.thy
```diff
@@ -221,8 +227,8 @@ lemma deleteObjects_corres:
    apply (frule invs_iflive)
    apply (frule (1) if_live_then_nonz_cap_to_cross, fastforce+)[1]
   apply (rule corres_stateAssert_add_assertion[rotated])
-   apply (rule delete_locale.deletionIsSafe_holds;
-          fastforce simp: delete_locale_def valid_cap_simps sch_act_simple_def state_relation_def
+   apply (rule detype_locale'.deletionIsSafe_holds;
+          fastforce simp: detype_locale'_def delete_locale_def invs_valid_pspace valid_cap_simps sch_act_simple_def state_relation_def
```
lsf37: Does this fit on a line? (hard to see in the diff)

proof/refine/Invariants_H.thy
corlewis: I didn't investigate too hard but some of the `2^n` seemed to come out of the specification, which is why I was thinking to duplicate these lemmas to handle it as well. That way we would at least not have to do things like `shiftl_t2n[symmetric]` for easy cases.

lsf37: yes, that would be perfectly fine, especially if we can do a `lemmas ... = ...[simplified ...]` thing

corlewis:
I forgot to mention, while doing this update I did come across a lot of proofs and crunches that broke due to `if_cong` no longer being added to `cong` in Refine. I probably should have investigated more to see if they could be reworked not to need it, but for now I just added it back in as needed.

proof/refine/CNodeInv_R.thy
```diff
@@ -5800,7 +5801,7 @@ proof (induct rule: finalise_spec_induct)
           apply (wp | simp)+
        apply (rule hoare_strengthen_post)
         apply (rule hoare_vcg_conj_lift[where Q="\<lambda>rv. cte_at' slot"])
-         apply (wp gen_typ_at_lifts [OF finaliseCap_typ_at'])[1]
+         apply (wp RISCV64.typ_at_lifts [OF finaliseCap_typ_at'])[1] (* FIXME: rt arch-split *)
```
lsf37, line 5804: Now that we've done the `typ_at_lifts` in master, do you remember if there is anything left that needs thinking for this in `rt` or can we just apply what we did on master?

lsf37: (not for this PR, I'm just looking ahead to the rt arch split)

proof/refine/RISCV64/ArchDetype_R.thy
corlewis: Good spotting, that line is too long. I'll change it to
```
   apply (rule detype_locale'.deletionIsSafe_holds;
          fastforce simp: detype_locale'_def delete_locale_def invs_valid_pspace valid_cap_simps
                          sch_act_simple_def state_relation_def sched_act_relation_def)
```

proof/refine/CNodeInv_R.thy
corlewis: Just apply what we did on master, that was one of my goals with doing it there first.

lsf37: Excellent

proof/invariant-abstract/Ipc_AI.thy
```diff
@@ -110,6 +110,37 @@ lemma ensure_no_children_wp:
...
+    | NullCap \<Rightarrow> NullCap
+    | ReplyCap t rights \<Rightarrow> ReplyCap t rights
+    | SchedContextCap s n \<Rightarrow> SchedContextCap s n
+    | SchedControlCap \<Rightarrow> SchedControlCap
+    | IRQControlCap \<Rightarrow> IRQControlCap
+    | IRQHandlerCap irq \<Rightarrow> IRQHandlerCap irq
+    | Zombie r b n \<Rightarrow> Zombie r b n
+    | ArchObjectCap cap \<Rightarrow> arch_update_cap_data pres w cap)"
+  by (cases cap,
+         simp_all only: cap.simps update_cap_data_def is_ep_cap.simps if_False if_True
```
michaelmcinerney: I think this was just moved but the indentation off here, and maybe we can do `cases cap; simp only: ` instead.

lsf37:
> I forgot to mention, while doing this update I did come across a lot of proofs and crunches that broke due to `if_cong` no longer being added to `cong` in Refine. I probably should have investigated more to see if they could be reworked not to need it, but for now I just added it back in as needed.

From my side, that's Ok. `if_cong` is easy to find if we want to reduce its use later, and locally it's not like using `if_cong` is a problem, it's just having it `[cong]` globally that it becomes slow and/or termination dangerous.

proof/refine/Detype_R.thy
```diff
@@ -2265,21 +2192,17 @@ lemma locateCTE_ko_wp_at':
   "\<lbrace>cte_at' src and pspace_distinct'\<rbrace>
    locateCTE src
    \<lbrace>\<lambda>rv. ko_wp_at' \<top> rv \<rbrace>"
-  apply (clarsimp simp: locateCTE_def split_def)
+  apply (clarsimp simp:locateCTE_def split_def)
```
michaelmcinerney: Nitpick: add space (back in)

lsf37 (approved):
A massive amount of work again (just the overall diff 36k+/134k- is gigantic), thank you for doing these. Our main problem with these merges is that we are so productive :-)

I have reviewed the post-merge commits closely and skimmed the merge commit. Looks all good from my side.

proof/refine/Finalise_R.thy
```diff
@@ -2071,6 +2065,20 @@ lemma sameObjectAs_Null[simp]:
 
 context Finalise_R begin
 
+crunch finaliseCap
+  for aligned'[wp]: pspace_aligned'
```
michaelmcinerney, line 2069: Not for this PR, but this reminds me of the constant battle with shortened names and abbreviations we have for a lot of invariants and functions. I really dislike them, but maybe it's more of a personal bias than anything else. Though for `crunch`, I know that the naming is involved somehow. Remembering the abbreviations is like an extra burden and some of them are not really all that easy to figure out, either. For `pspace_aligned'`, or maybe `pspace_aligned` in AInvs, we have at least `psp_aligned`, but there are probably several others floating around.

proof/refine/Ipc_R.thy
```diff
@@ -4242,10 +4219,10 @@ lemma ntfnSetActive_corres:
   apply wpsimp
   done
 
-crunch maybe_donate_sc
+crunch  maybe_donate_sc
```
michaelmcinerney: Nitpick: remove extra space

proof/refine/IpcCancel_R.thy
```diff
@@ -2644,7 +2639,7 @@ crunch as_user
   (wp: crunch_wps ep_queues_blocked_lift ntfn_queues_blocked_lift ready_queues_runnable_lift
        release_q_runnable_lift)
 
-lemma suspend_corres:
+lemma (in delete_one_gen)  suspend_corres:
```
michaelmcinerney: Nitpick: extra space after the closed parenthesis.

proof/refine/IpcCancel_R.thy
```diff
@@ -4322,30 +4317,20 @@ lemma setThreadState_unlive_other:
...
-  apply (clarsimp simp: replyUnlink_def updateReply_def)
-  apply (wpsimp wp: setThreadState_Inactive_unlive set_reply'.set_wp gts_wp')
-  apply (clarsimp simp: live'_def ko_wp_at'_def live_reply'_def obj_at'_def ps_clear_upd fun_upd_apply)
-  done
+lemma possibleSwitchTo_unlive_other:
+  "\<lbrace>ko_wp_at' (Not \<circ> live') p and K (p \<noteq> t) and valid_tcbs'\<rbrace>
+   possibleSwitchTo t
+   \<lbrace>\<lambda>_. ko_wp_at' (Not \<circ> live') p\<rbrace>"
+  unfolding possibleSwitchTo_def inReleaseQueue_def
+  by (wpsimp wp: tcbSchedEnqueue_unlive_other threadGet_wp rescheduleRequired_unlive)+
```
michaelmcinerney: Is the `+` necessary here?

proof/refine/IpcCancel_R.thy
```diff
@@ -4711,4 +4697,4 @@ lemma asUser_tcbQueued[wp]:
 
 lemmas asUser_st_tcb_at'[wp] = asUser_obj_at [folded st_tcb_at'_def]
 
-end
+end
```
michaelmcinerney, line 4699: Maybe not for this PR, but what's the convention with having an empty line at the end of a file? Do we have one? I just checked and `Style.thy` itself does not have an empty line at the end of the file.

proof/refine/Tcb_R.thy
```diff
@@ -247,13 +254,18 @@ crunch maybe_donate_sc
...
+  and pspace_distinct[wp]: "pspace_distinct :: det_state \<Rightarrow> _"
+  (wp: crunch_wps simp: crunch_simps)
+
 lemma restart_corres:
   "corres dc
      (einvs and tcb_at t and ex_nonz_cap_to t and current_time_bounded) invs'
      (Tcb_A.restart t) (ThreadDecls_H.restart t)"
   apply (simp add: Tcb_A.restart_def Thread_H.restart_def test_possible_switch_to_def
                    get_tcb_obj_ref_def)
-  apply (rule corres_stateAssert_ignore, simp add: sch_act_wf_asrt_def)+
```
michaelmcinerney: Thanks for removing these assert definitions.

michaelmcinerney (approved):
Looks great! Just a few minor nitpicks and a few questions that we can discuss elsewhere, if you'd like to merge this soon.

proof/refine/Finalise_R.thy
lsf37: You are right, we should really standardise these, it's always a drag to try to remember which one is used in which context. I'm not sure we want to always have the full long form, but if there is some kind of abbreviating going on, it should always be the same and should be used consistently over different proofs.

proof/refine/IpcCancel_R.thy
lsf37: I thought gitlint would be complaining about this (whatever we decided we wanted), but apparently not.

Edit: sorry, not gitlint, but the trailing whitespace check

proof/refine/CNodeInv_R.thy
```diff
@@ -7626,18 +7620,18 @@ qed
...
-
 crunch invokeCNode
   for typ_at'[wp]: "\<lambda>s. P (typ_at' T p s)"
   and sc_at'_n[wp]: "\<lambda>s. P (sc_at'_n n p s)"
   (ignore: filterM finaliseSlot
      simp: crunch_simps filterM_mapM unless_def
        wp: crunch_wps undefined_valid finaliseSlot_preservation)
 
-global_interpretation invokeCNode: typ_at_all_props' "invokeCNode i"
+sublocale invokeCNode: typ_at_all_props' "invokeCNode i"
```
Xaphiosis, line 7630: Is this an example of how the typ_all_at_props' shifting into an arch-splitted theory? I'm expecting that in the generic theory this is what drops the Arch interpretation?

proof/refine/InvariantsPre_H.thy
```diff
@@ -143,10 +143,17 @@ lemma (in Arch) cteSizeBits_cte_level_bits:
...
+
+lemmas aobj_of'_simps[simp] = aobj_of'_def[split_simps kernel_object.split]
+
+lemma aobj_of'_Some[iff]:
+  "(aobj_of' a = Some ao) = (a = KOArch ao)"
+  by (simp add: aobj_of'_def split: kernel_object.splits)
 
 abbreviation aobjs_of' :: "kernel_state \<Rightarrow> obj_ref \<rightharpoonup> arch_kernel_object" where
-  "aobjs_of' s \<equiv> ksPSpace s |> aobj_of'"
+  "aobjs_of' \<equiv> \<lambda>s. ksPSpace s |> aobj_of'"
```
Xaphiosis, line 157: interesting, where did this change originate? something backportable to master?

proof/refine/Invariants_H.thy
Xaphiosis: I approve of this change and backporting. One thing to check for is to make sure it doesn't introduce more warnings for duplicate simplification rules. It looks like it'll be fine, but keep an eye out.

For the 2^cteSizeBits, I'm ok if you make a rule for this case, but I would prefer that it not end up in [simp]. A long-term strategy is to get rid of the powers of two in favour of shifts and masks when they related to transforms like this, and when that happens seeing the explicit rule being added would be better than mystery failures or successes.

proof/refine/IpcCancel_R.thy
```diff
@@ -1053,13 +1055,6 @@ locale delete_one_gen = delete_one +
     "\<And>t. corres dc (tcb_at t and pspace_aligned and pspace_distinct) \<top>
                 (update_restart_pc t) (updateRestartPC t)"
   (* results of IpcCancel_R locale *)
-  assumes reply_descendants_of_mdbNext:
-    "\<lbrakk> (s, s') \<in> state_relation; valid_reply_caps s; valid_reply_masters s;
-       valid_objs s; valid_mdb s; valid_mdb' s'; pspace_aligned' s';
-       pspace_distinct' s'; st_tcb_at (Not \<circ> halted) t s;
-       ctes_of s' (t + 2*2^cte_level_bits) = Some cte \<rbrakk>
-     \<Longrightarrow> (descendants_of (t, tcb_cnode_index 2) (cdt s) = {})
-        = (mdbNext (cteMDBNode cte) = nullPointer)"
```
Xaphiosis: interesting, where did this one go?

proof/refine/KHeap_R.thy
```diff
@@ -1522,6 +1522,21 @@ lemma setObject_state_refs_of':
                   cong: if_cong option.case_cong)
   done
 
+lemma setObject_state_hyp_refs_of':
+  assumes x: "updateObject val = updateObject_default val"
+  shows
+  "\<lbrace>\<lambda>s. P ((state_hyp_refs_of' s) (ptr := hyp_refs_of' (injectKO val)))\<rbrace>
+     setObject ptr val
```
Xaphiosis, line 1529: indentation

proof/refine/KHeap_R.thy
```diff
@@ -2338,13 +2344,35 @@ lemma setSchedContext_state_refs_of'[wp]:
...
+lemma state_hyp_refs_of'_sc:
+  "sc_at' sc s \<Longrightarrow> (state_hyp_refs_of' s) (sc := {}) = state_hyp_refs_of' s"
+  by (rule ext) (clarsimp simp: state_hyp_refs_of'_def obj_at'_def)
+
+lemma set_sc_state_hyp_refs_of'[wp]:
+  "setSchedContext p reply \<lbrace>\<lambda>s. P (state_hyp_refs_of' s)\<rbrace>"
+  unfolding setSchedContext_def
+  apply (rule setObject_gen_obj_at, simp)
+  apply (wp setObject_state_hyp_refs_of'; simp add: gen_objBits_simps state_hyp_refs_of'_sc)
+  done
```
Xaphiosis, line 2356: looks collapsible via `rule ... ; wpsimp`, same for `set_reply_state_hyp_refs_of'` below

proof/refine/RISCV64/ArchCNodeInv_R.thy
```diff
@@ -259,6 +259,8 @@ crunch Arch_finaliseCap, prepareThreadDelete
    simp: crunch_simps o_def pteAtIndex_def
    rule: RISCV64_H.finaliseCap_def)
 
+declare prepareThreadDelete_nosch[CNodeInv_R_assms]
```
Xaphiosis: what's happening here?

proof/refine/RISCV64/ArchIpcCancel_R.thy
```diff
@@ -106,17 +102,8 @@ context delete_one begin
...
 
 end (* delete_one *)
 
-context delete_one_conc_pre begin
-
-sublocale delete_one_conc_pre_gen
-  by unfold_locales
-     (fact emptySlot_st_tcb_at')
-
-end (* delete_one_conc_pre *)
```
Xaphiosis: interesting that this is disappearing. did it move elsewhere during the merge? no longer necessary on rt?

proof/refine/RISCV64/ArchIpc_R.thy
```diff
@@ -158,6 +147,8 @@ lemma makeArchFaultMessage_inv[Ipc_R_assms, wp]:
   unfolding makeArchFaultMessage_def
   by (wpsimp wp: asUser_inv getRestartPC_inv split: arch_fault.split)
 
+declare lookupIPCBuffer_valid_ipc_buffer[Ipc_R_assms]
```
Xaphiosis: another one of these declares... not sure why you need to do them like this; also aside: for arch-split I never use `declare`, but `lemmas [Ipc_R_assms] = ...` because then they're easier to combine. And furthermore, it's much more convenient to put them as late as possible, right in front of the actual locale interpretation. Scattering them in the middle like this makes it hard to clean up. (same comment goes for any other declares like this)

proof/refine/RISCV64/ArchTcb_R.thy
```diff
@@ -18,10 +18,7 @@ lemma activateIdleThread_corres[Tcb_R_assms]:
     (arch_activate_idle_thread t) (activateIdleThread t)"
   by (simp add: arch_activate_idle_thread_def activateIdleThread_def)
 
-crunch arch_post_modify_registers
-  for pspace_aligned[Tcb_R_assms, wp]: pspace_aligned
-  and pspace_distinct[Tcb_R_assms, wp]: pspace_distinct
-  (wp: crunch_wps simp: crunch_simps)
+declare arch_post_modify_registers_inv[Tcb_R_assms]
```
Xaphiosis: another one of these declares

proof/refine/RISCV64/ArchTcb_R.thy
```diff
@@ -74,10 +66,12 @@ lemma isValidVTableRootD:
...
                 option.split_asm)
 
-crunch prepare_thread_delete, arch_finalise_cap
+crunch arch_finalise_cap
   for pspace_aligned[Tcb_R_assms, wp]: "pspace_aligned :: det_ext state \<Rightarrow> _"
   and pspace_distinct[Tcb_R_assms, wp]: "pspace_distinct :: det_ext state \<Rightarrow> _"
-  (simp: crunch_simps preemption_point_def wp: crunch_wps OR_choiceE_weak_wp)
+  (simp: crunch_simps wp: crunch_wps)
+
+declare prepare_thread_delete_inv[Tcb_R_assms]
```
Xaphiosis: and another one

proof/refine/Tcb_R.thy
```diff
@@ -49,8 +49,11 @@ locale Tcb_R =
   assumes checked_insert_tcb_invs_gen:
     "\<And>target ref new_cap src_slot slot.
      \<lbrace>invs and cte_wp_at (\<lambda>c. c = cap.NullCap) (target, ref)
-      and K (is_cnode_or_valid_arch new_cap) and valid_cap new_cap
+      and K (is_cnode_or_valid_arch new_cap \<or> valid_fault_handler new_cap)
+      and valid_cap new_cap
       and tcb_cap_valid new_cap (target, ref)
+      and (\<lambda>s. valid_fault_handler new_cap
+                       \<longrightarrow> cte_wp_at (\<lambda>c. c = new_cap \<or> c = cap.NullCap) src_slot s)
```
Xaphiosis: indentation of -->

proof/refine/Finalise_R.thy
Xaphiosis: It's been a plague for ages, with just about every person that tried to do an abbreviation coming up with a different way to abbreviate. I've seen many flavours of just about everything in crunches when arch-splitting. I'm actually fine for writing out the long name for crunches, and especially crunches that go straight into [wp].

proof/refine/Detype_R.thy
```diff
@@ -890,6 +810,40 @@ lemma null_filter':
...
+    by (rule arch_deletionIsSafe[OF al vs vu sr])
+
+  thus ?thesis
+    apply -
+    apply (clarsimp simp: deletionIsSafe_def)
+    apply (fastforce simp: add_mask_fold dest!: live_notRange)
+    done
+qed
+
+end (* detype_locale'_gen *)
```
Xaphiosis, line 845: is this identical to what is on master? did it move or something? it seems very familiar

Xaphiosis (approved):
Well done on getting through all that.
I have mainly questions, and one gripe: those declares make some sense being where they are in the diff, but the style and location approach is differing from what I was trying to do on master during arch-split. It would be nice if we were on the same frequency for those.

proof/refine/InvariantsPre_H.thy
michaelmcinerney: This projection was done on `rt` as part of the IPC queues work. We could backport the projections to `master` either separately, or certainly if/when we did the IPC queues work on `master`.

lsf37: There's nothing to back port, the change makes the definition equal to the one in master.

proof/refine/RISCV64/ArchIpc_R.thy
lsf37: They are from crunches, no? (collecting them at the end with `lemmas` is good, though, that's a nice method to settle on)

proof/refine/CNodeInv_R.thy
corlewis: This is actually an example of one shifting into an interface locale, and needing to do a `sublocale` for that reason.

The `rt` version of `typ_at_all_props'` is in a bit of a weird state where it hasn't been arch-split yet. Once I get back to and finish the master version and the merges into `rt` catch up to that then this should all be consistent.

In this case, this would still be a `sublocale` but as this is a generic file, it would be the generic version of `typ_at_all_props'`.

proof/refine/InvariantsPre_H.thy
corlewis: This projection has actually already been added to master, as part of Gerwin's changes for the `asid_map` addition. During the merge I took master's version of the projection with the `λ` but kept it in the location that rt had it. In master, the abbreviation is instead in `StateRelationPre`.

proof/refine/IpcCancel_R.thy
corlewis: Reply caps don't exist in rt in the same way as they do in master, so this and many other lemmas to do with them have been removed.

proof/refine/KHeap_R.thy
corlewis: Unfortunately doesn't work out nicely, something goes wrong with the application of `state_hyp_refs_of'_sc`.

proof/refine/RISCV64/ArchIpc_R.thy
corlewis: In master there are interface lemmas that are produced here by crunches, while in rt the crunches aren't doing anything because it finds preexisting lemmas that can be used instead. I didn't want to adjust the interface because some of those earlier lemmas won't exist for other architectures, and then without thinking too hard about it just put the `declare` where each `crunch` used to be.

I'll consolidate them and change to `lemmas`.

proof/refine/RISCV64/ArchIpcCancel_R.thy
corlewis: I'm actually not sure what happened here. Looking at my different versions of the repository, it looks like I was able to remove `delete_one_conc_pre_gen` completely from `rt`?

I'm guessing that I might have noticed that all the lemmas in maser that needed to be proved inside of it had been deleted in `rt`, and then decided that there was no reason to keep the locale.

michaelmcinerney: My guess is that `cancelIPC` doesn't have the `cteDeleteOne` in `rt`, so a lot of the deletion lemmas weren't necessary in the IPC cancel files.

proof/refine/Detype_R.thy
corlewis: The same locale structure as master, but moved and the `deletionIsSafe` lemma is a bit different on `rt`.

This one almost caused a whole bunch of problems but I was able to move some other lemmas that the `rt` version depends on into locales that let me keep the same overall structure.

Xaphiosis:
I think my comments are all answered or updated. Thanks!

PR #1039, opened 2026-07-27
clean up manual instantiation in CRefine

lsf37:
As promised in #1035, clean up that manual instantiation. The right combination of `ccorres_if_bindE_lhs` and `ccorres_symb_exec_r2` does the trick in the end (plus avoiding further `ccorres_symb_exec_r2` steps later by  using `ccorres_cond_both'` later).

### DIFF
--- a/lib/clib/CCorresLemmas.thy
+++ b/lib/clib/CCorresLemmas.thy
@@ -630,6 +630,22 @@ lemma ccorres_if_lhs:
           hs (if P then f else g) conc"
   by (simp split: if_split)
 
+lemma ccorres_if_bind_lhs:
+  "\<lbrakk> P \<Longrightarrow> ccorres_underlying sr Gamm r xf arrel axf Q S hs (f >>= m) conc;
+     \<not> P \<Longrightarrow> ccorres_underlying sr Gamm r xf arrel axf R T hs (g >>= m) conc \<rbrakk>
+   \<Longrightarrow> ccorres_underlying sr Gamm r xf arrel axf (\<lambda>s. (P \<longrightarrow> Q s) \<and> (\<not> P \<longrightarrow> R s))
+                      {s. (P \<longrightarrow> s \<in> S) \<and> (\<not> P \<longrightarrow> s \<in> T)}
+          hs ((if P then f else g) >>= m) conc"
+  by (simp split: if_split)
+
+lemma ccorres_if_bindE_lhs:
+  "\<lbrakk> P \<Longrightarrow> ccorres_underlying sr Gamm r xf arrel axf Q S hs (f >>=E m) conc;
+     \<not> P \<Longrightarrow> ccorres_underlying sr Gamm r xf arrel axf R T hs (g >>=E m) conc \<rbrakk>
+   \<Longrightarrow> ccorres_underlying sr Gamm r xf arrel axf (\<lambda>s. (P \<longrightarrow> Q s) \<and> (\<not> P \<longrightarrow> R s))
+                      {s. (P \<longrightarrow> s \<in> S) \<and> (\<not> P \<longrightarrow> s \<in> T)}
+          hs ((if P then f else g) >>=E m) conc"
+  by (simp split: if_split)
+
 lemma ccorres_if_bindE:
   "ccorres_underlying sr Gamm r xf arrel axf G G' hs (if a then (b >>=E f) else (c >>=E f)) d
   \<Longrightarrow> ccorres_underlying sr Gamm r xf arrel axf G G' hs ((if a then b else c) >>=E f) d"
--- a/proof/crefine/ARM/Arch_C.thy
+++ b/proof/crefine/ARM/Arch_C.thy
@@ -2335,7 +2335,7 @@ lemma decodeARMFrameInvocation_ccorres:
         apply (wpsimp wp: sts_invs_minor' ct_in_state'_set)+
       apply (vcg exspec=setThreadState_modifies)
 
-      \<comment> \<open>PageUnify_Instruction | PageCleanInvalidate_Data | Page Invalidate_Data | PageClean_Data\<close>
+     \<comment> \<open>PageUnify_Instruction | PageCleanInvalidate_Data | Page Invalidate_Data | PageClean_Data\<close>
      apply (rule ccorres_rhs_assoc)+
      apply csymbr+
      apply (simp add: ivc_label_flush_case decodeARMPageFlush_def list_case_If2 if3_fold2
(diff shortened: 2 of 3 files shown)


lib/clib/CCorresLemmas.thy
```diff
@@ -630,6 +630,22 @@ lemma ccorres_if_lhs:
           hs (if P then f else g) conc"
   by (simp split: if_split)
 
+lemma ccorres_if_bind_lhs:
+  "\<lbrakk> P \<Longrightarrow> ccorres_underlying sr Gamm r xf arrel axf Q S hs (f >>= m) conc;
+     \<not> P \<Longrightarrow> ccorres_underlying sr Gamm r xf arrel axf R T hs (g >>= m) conc \<rbrakk>
+   \<Longrightarrow> ccorres_underlying sr Gamm r xf arrel axf (\<lambda>s. (P \<longrightarrow> Q s) \<and> (\<not> P \<longrightarrow> R s))
+                      {s. (P \<longrightarrow> s \<in> S) \<and> (\<not> P \<longrightarrow> s \<in> T)}
+          hs ((if P then f else g) >>= m) conc"
```
Xaphiosis, line 638: something worth considering for these is whether you want a `>>= (λrv. m rv)` for name binding to carry through

proof/crefine/ARM/Arch_C.thy
```diff
@@ -2392,122 +2393,81 @@ lemma decodeARMFrameInvocation_ccorres:
...
-                    apply (rule syscall_error_throwError_ccorres_n)
-                    apply (simp add: syscall_error_to_H_cases)
-                   apply clarsimp
-                  apply assumption
-                 apply clarsimp
-                 apply vcg
-                 apply clarsimp
+               apply (simp add: injection_handler_whenE cong: ccorres_prog_only_cong)
+               apply csymbr
+               apply ccorres_rewrite
```
Xaphiosis, line 2423: csymbr followed by ccorres_rewrite seems unusual, wouldn't the rewrite be possible earlier already?

proof/crefine/ARM/Arch_C.thy
```diff
@@ -2392,122 +2393,81 @@ lemma decodeARMFrameInvocation_ccorres:
...
+               apply (rule ccorres_if_cond_throws[rotated -1,where Q = \<top> and Q' = \<top>])
+                  apply vcg
+                 apply (clarsimp simp: hd_drop_conv_nth hd_conv_nth)
+                 apply (clarsimp dest!: ccap_relation_PageCap_generics)
+                apply (simp add: injection_handler_throwError)
+                apply (rule syscall_error_throwError_ccorres_n)
+                apply (simp add: syscall_error_to_H_cases)
+               apply csymbr
+               apply csymbr
+               apply csymbr
```
Xaphiosis, line 2453: does csymbr have an argument for the number of times it should apply?

proof/crefine/ARM/Arch_C.thy
```diff
@@ -2392,122 +2393,81 @@ lemma decodeARMFrameInvocation_ccorres:
...
+               apply (simp add: performARMMMUInvocations bindE_assoc)
+               apply (ctac add: setThreadState_ccorres)
+                 apply (ctac(no_vcg) add: performPageFlush_ccorres)
```
Xaphiosis: I think style-wise we're more prone to `ctac (no_vcg)`

Xaphiosis (approved):

proof/crefine/ARM/Arch_C.thy
lsf37: no, we have this kinds of pattern everywhere. Maybe we should add an argument, though. I guess once could write `repeat 3 csymbr`, but then again these proofs are painful enough that you want to be able to single-step through and see what is happening.

proof/crefine/ARM/Arch_C.thy
lsf37: this is just reindented, I didn't write any of that

lib/clib/CCorresLemmas.thy
lsf37: Hm, interesting. I don't think this should be necessary for these, but need to confirm -- the bound name there does not appear as another bound name in the goal, so there is nothing for Isabelle to transport. But the whole mechanism is obscure enough that I'm not entirely sure. Will try it out.

proof/crefine/ARM/Arch_C.thy
lsf37: (happy to fix this one up, I just don't want to fix up the entire thing because it's large and old and very terrible and I don't want to touch more than necessary)

proof/crefine/ARM/Arch_C.thy
lsf37: I don't think you can. The csymbr evaluates a `ret__int = <constant value>` and the `corres_rewrite` evaluates the `IF` with that constant value. An initial `ccorres_rewrite` wouldn't evaluate the `ret_int` assignment.

You could also do a `simp` after the `csymbr`, but `corres_rewrite` is a bit less error prone.

lib/clib/CCorresLemmas.thy
lsf37: Confirmed, it's not needed. The name survives without the explicit lambda (because the lambda is never taken apart)

Xaphiosis: I would like to get an intuition for this. Can you clarify what you mean by the lambda not being taken apart?

lsf37: This all started with rules like `(!!x. P x) ==> All (%x. P x)`. Note that the two `x` are different bound `x`. I.e. `(!!x. P x) ==> All (%y. P y)` would be the same rule by term equality. And they are different cases of bound-ness. The first (`!!`) is a goal parameter, the second is a normal lambda.

When you use the original rule (where the names are the same) on a goal that is `All (%z. P z)`,  you ideally want to see `!!z. P z` as the next subgoal. Operationally, the rule takes apart the lambda in `All (%z. P z)`, notes that the name was `z` and uses `z` as the name of the goal parameter. Overall, since the rule reused the name `x`, the rule application is also trying to reuse the instance of what `x` is in the goal. This applies only to names bound under lambda that become goal parameters.

In our rule above, this case doesn't apply, because the lambda still exists as the same lambda that it was before, the name is not reused as a goal parameter. Since the lambda sub-term just gets reused as a whole, it doesn't need any special handling for the name to survive, and it doesn't even need to mention that the sub-term is a lambda term.

PR #1041, opened 2026-07-29
arch-split Refine + EmptyFail_H, add Refine toplevel, clean up small theories

Xaphiosis:
🦆🦆🦆 This is the last of arch-split Refine pass 1, except LevityCatch which I'll deal with as part of cleanup. There are quite a few commits in this one, due to dealing with the smaller theories. The main action is in Refine.thy and EmptyFail_H.thy, which follow the usual arch-split copy-AARCH64-REST process. Review commit-by-commit.

Things to consider for future: do we get rid of LevityCatch by distributing it into later theories? Having both [Arch]EmptyFail.thy and [Arch]EmptyFail_H.thy is confusing, but the latter isn't imported for CRefine, which provides a very small time save.

Commits tagged [fake] are generated to make PR review easier.
For Refine and EmptyFail_H we have:
- move Theory to ArchTheory, create a new generic Theory
- [fake] copy AARCH64 version of original Theory to new generic Theory (prevent wall of green)
- arch-split AARCH64 only
- copy AARCH64 arch-split version of ArchTheory to other arches
- arch-split other arches

Apart from those, we have:
- `refine riscv64+x64: remove archThreadSet_if_live'` : cleanup, thanks @corlewis for pointing this out
- `refine: make IncKernelInit.thy generic, use ARM version`
- `refine: make KernelInit_R generic, use ARM version`
- `refine: remove PageTableDuplicates on 64-bit architectures`
- `refine: remove InitLemmas`
- `refine: add new RefineToplevel_R to Refine session`, plus commits updating include hierachy
- `refine+crefine: make RAB_FN generic`
- updating imports for ArchEmptyFail_H replacing EmptyFail_H in the hierarchy

Stats: 56 files changed, 2294 insertions(+), 6739 deletions(-)
Total: 4445 lines removed


proof/refine/ARM_HYP/KernelInit_R.thy
```diff
@@ -1,45 +0,0 @@
...
-begin
-
-(* Axiomatisation of the rest of the initialisation code *)
-axiomatization where
-  init_refinement:
-  "Init_H \<subseteq> lift_state_relation state_relation `` Init_A"
-
-axiomatization where
-  ckernel_init_valid_duplicates':
-  "\<forall>((tc,s),x) \<in> Init_H. vs_valid_duplicates' (ksPSpace s)"
```
lsf37, line 23: I assume that is the one that ARM also has but the other ones do not. I think assuming this generically is fine, just want to point out that technically this means we have an unnecessary for some architectures. The axiom just assumes "True" in the end, so it's Ok, but we should be aware that we're doing this.

proof/refine/Refine.thy
```diff
@@ -283,15 +182,19 @@ lemma kernel_entry_invs:
...
-                                 (m = KernelMode \<longrightarrow> e \<noteq> None) \<and>
-                                 (m = UserMode \<longrightarrow> ct_running s) \<and>
-                                 (m = IdleMode \<longrightarrow> ct_idle s) \<and>
-                                 (e \<noteq> None \<and> e \<noteq> Some Interrupt \<longrightarrow> ct_running s) \<and>
-                                 0 < domain_time s \<and> valid_domain_list s \<and>
-                                 (scheduler_action s = resume_cur_thread)}"
+end (* Refine *)
+
+(* FIXME arch-split: is this type signature sane? *)
+definition full_invs :: "(('a \<times> det_ext state) \<times> mode \<times> event option) set" where
```
lsf37: It is, but I have forgotten (not for the first time) what `tc` is. Possibly worth a comment here when we find out, since it is ignored in this definition.

proof/refine/Refine.thy
```diff
@@ -794,6 +638,8 @@ lemma valid_domain_list_from_invs':
   "\<lbrakk> (s, s') \<in> state_relation; invs' s' \<rbrakk> \<Longrightarrow> valid_domain_list s"
   by (clarsimp simp: valid_domain_list_2_cross invs'_def valid_state'_def elim!: state_relationE)
 
+context Refine begin
+
 lemma ckernel_invariant:
   "ADT_H uop \<Turnstile> full_invs'"
```
lsf37, line 653: I wonder if it is worth it spelling out these top-level theorems again manually in the top-level Refine theory. Just without name and proved by `fact` from the locale interpretation. Doesn't really do anything apart from confirming without doubt that the top-level theorem holds without hidden assumptions in the way we want it.

lsf37 (approved):
Excellent! Congratulations on reaching the end of Refine 🎉 🎉 

I don't have any complaints this this time, looks all very good.

>Things to consider for future: do we get rid of LevityCatch by distributing it into later theories? 

That would be nice to do, yes. It's not really needed any longer and everything in there is there because it should be distributed.

> Having both [Arch]EmptyFail.thy and [Arch]EmptyFail_H.thy is confusing, but the latter isn't imported for CRefine, which provides a very small time save.

I'd be fine with either option: leaving them as is or integrating them into one and take a small time hit for CRefine. The latter is potentially less error prone for the future because it collects everything in one place and we won't have these situations any more where we've included the wrong EmptyFail and something isn't visible because of that. There is a chance that integrating them fails in first instance because they build on different contexts.

proof/refine/Refine.thy
```diff
@@ -15,178 +15,70 @@ imports
...
+begin
 
-lemma device_mem_relation:
-  "\<lbrakk>(s,s') \<in> state_relation; valid_state' s'; valid_state s\<rbrakk>
-   \<Longrightarrow> device_mem' s' = device_mem s"
-  apply (rule ext)
-  apply (clarsimp simp: device_mem_def device_mem'_def pointerInUserData_relation
-                        pointerInDeviceData_relation)
-  done
+lemmas [simp] = valid_sched_init[simplified]
```
corlewis: This seems a little bit unexpected, why do we need the `simplified` version in `simp` now?

proof/refine/Refine.thy
corlewis: I think `tc` is `user_context` in the main place that this is used. I guess it maybe stands for 'thread context'?

proof/refine/Refine.thy
```diff
@@ -428,62 +289,57 @@ lemma doMachineOp_sch_act_simple[wp]:
   "doMachineOp f \<lbrace>sch_act_simple\<rbrace>"
   by (wp sch_act_simple_lift)
 
+lemma device_update_invs':
+  "\<lbrace>invs'\<rbrace>doMachineOp (device_memory_update ds)
+   \<lbrace>\<lambda>_. invs'\<rbrace>"
```
corlewis: Not new but while we're cleaning so much else up here, fixing this style would be good.

proof/refine/Refine.thy
```diff
@@ -687,23 +549,21 @@ lemma ct_idle_related:
...
-  "full_invs' \<equiv> {((tc,s),m,e). invs' s \<and>
-                          ex_abs (einvs::det_ext state \<Rightarrow> bool) s \<and>
-                          ksSchedulerAction s = ResumeCurrentThread \<and>
-                          (ct_running' s \<or> ct_idle' s) \<and>
-                          (m = KernelMode \<longrightarrow> e \<noteq> None) \<and>
-                          (m = UserMode \<longrightarrow> ct_running' s) \<and>
-                          (m = IdleMode \<longrightarrow> ct_idle' s) \<and>
-                          (e \<noteq> None \<and> e \<noteq> Some Interrupt \<longrightarrow> ct_running' s) \<and>
-                          0 < ksDomainTime s}"
+  "full_invs' \<equiv> {((tc,s),m,e).
```
corlewis, line 561: Adding a type_signature to `full_invs` was a good idea, should we do the same here?

proof/refine/Refine.thy
```diff
@@ -782,7 +626,7 @@ lemma domain_time_rel_eq:
 crunch doUserOp, checkActiveIRQ
   for valid_objs': valid_objs'
   (wp: crunch_wps
-   rule: doUserOp_def) (* FIXME: crunch can't find the rule of locale-defined constant *)
+   rule: doUserOp_def) (* crunch can't find the rule of locale-defined constant *)
```
corlewis: I think you accidentally reverted this FIXME in a rebase.

proof/refine/AARCH64/ArchRefine.thy
```diff
@@ -118,209 +119,74 @@ lemma typ_at_DeviceDataI:
...
-  apply (subst mult.commute)
-  apply (subst shiftl_t2n [symmetric])
-  apply (simp add: shiftr_shiftl1)
-  apply (subst mask_out_add_aligned)
-   apply (rule is_aligned_neg_mask)
-   apply (simp add: pbfs_atleast_pageBits)
-  apply (subst add.commute)
-  apply (simp add: word_plus_and_or_coroll2)
+  apply (drule_tac sz=sz and n="(p && mask (pageBitsForSize sz)) >> pageBits"
+               in typ_at_ADeviceDataI[where s=s and s'=s'])
```
corlewis: I think the second line should be indented one more?

proof/refine/EmptyFail_H.thy
```diff
@@ -101,129 +108,151 @@ lemma constOnFailure_empty_fail[intro!, wp, simp]:
...
 
-lemma decodeCNodeInvocation_empty_fail[intro!, wp, simp]:
-  "empty_fail (decodeCNodeInvocation label args cap exs)"
-  apply (rule_tac label=label and args=args and exs=exs in decode_cnode_cases2)
-         apply (simp_all add: decodeCNodeInvocation_def
-                         split_def cnode_invok_case_cleanup unlessE_whenE
-                   cong: if_cong bool.case_cong list.case_cong)
-         by (simp | wp | wpc | safe)+
+lemmas finalise_spec_empty_fail_induct = finaliseSlot'.induct[where P=
+    "\<lambda>sl exp s. spec_empty_fail (finaliseSlot' sl exp) s"]
```
corlewis: I think this was just moved but style could be improved.

proof/refine/EmptyFail_H.thy
```diff
@@ -101,129 +108,151 @@ lemma constOnFailure_empty_fail[intro!, wp, simp]:
...
+  assumes arch_deriveCap_empty_fail[intro!, wp, simp]:
+    "\<And>x y. empty_fail (Arch.deriveCap x y)"
+  assumes Arch_postCapDeletion_empty_fail[intro!, wp, simp]:
+    "\<And>c. empty_fail (Arch.postCapDeletion c)"
+  assumes prepareThreadDelete_empty_fail[intro!, wp, simp]:
+    "\<And>t. empty_fail (prepareThreadDelete t)"
+  (* FIXME arch-split: if we adopt lowercase arch_ naming for these, then the fact that we also
+     adopted Arch_ prefix for design-spec abbreviations to assist crunch becomes inconsistent *)
+  assumes Arch_finaliseCap_empty_fail[intro!, wp, simp]:
+    "\<And>x y. empty_fail (Arch.finaliseCap x y)"
```
corlewis: It would be good to come to a consensus on this now, or, since this doesn't make that much difference, at least make an arbitrary decision so that we can start working towards consistency.

In my mind, the `Arch_` prefix for constant names is consistent with and comes from the Haskell code using `Arch.` to qualify a lot of its names of arch functions. Since I don't think we're going to change the Haskell, it makes sense to me to keep `Arch_` for prefixing abbreviations in Refine.

The next question is whether we're more annoyed about lemma names starting with a capital or with lemma names not matching the constants that they are about. I personally don't have a strong preference either direction beyond a desire for consistency.

corlewis:
Minor commit suggestion: for "refine: remove InitLemmas" I think it should either be "Content was generic" or "Contents were generic"

corlewis:
> Having both [Arch]EmptyFail.thy and [Arch]EmptyFail_H.thy is confusing, but the latter isn't imported for CRefine, which provides a very small time save.

I was leaning towards combining them but a quick experiment suggests that it might be more work than we want to do now to sort out the dependencies. Parts of Refine depends on lemmas in [Arch]EmptyFail, while some of the more complicated [Arch]EmptyFail_H proofs use lemmas from Refine. That means we can't easily consolidate all of the `empty_fail` lemmas into one of the two pairs of files.

corlewis (approved):
Amazing to see the end of Refine, and with so much still being able to made generic in these last files! Just a few minor comments and questions from my side.

Xaphiosis:
> Minor commit suggestion: for "refine: remove InitLemmas" I think it should either be "Content was generic" or "Contents were generic"

Will go with "were", thank you!

Also thanks for looking at the EmptyFail situation. That does suck a bit, unsure where that leaves us for future situations. Maybe the bits of Refine these depend on can be migrated upstream to EmptyFail also, but without looking deeper I don't know. Anyway, looks like not a near future cleanup.

proof/refine/EmptyFail_H.thy
Xaphiosis: Talked this over with @lsf37 as well, and it seems we have a three-way consensus. The `Arch.` to `Arch_` for naming makes some sense, but I don't think any of us much like lemma names starting with caps. So I will make this lowercase here, but sadly I did not do this in past bits of Refine arch-split, so might have to go back for that during cleanup.

proof/refine/Refine.thy
Xaphiosis: Many thanks for finding this. I got a *huge* conflict in the rebase and couldn't figure out what the actual issue was. This line! Will put the FIXME back

proof/refine/Refine.thy
Xaphiosis: So should I change 'a to user_context, or what? I feel bad leaving it as 'a

Xaphiosis: but `tc` appears free in both full_invs and full_invs' ... hmm

proof/refine/Refine.thy
Xaphiosis: This one is super annoying. When you take `valid_sched (init_A_st :: det_state)` and simplify it, it explodes with `valid_sched_2  ......... ......  ..   ......  ......`. Not something I'd want to type out, but also not in simp normal form.
I don't know why exactly it blows up, and not sure now's the time to fix it, but I agree this is unexpected.
I will attempt to make the interface assumption [simplified, simp] and add a comment.

proof/refine/Refine.thy
Xaphiosis: I am not opposed to the principle, but don't feel very motivated to do the execution :)
Feel free anytime after this is merged.

proof/refine/ARM_HYP/KernelInit_R.thy
Xaphiosis: Should I put in a comment then? Since the assumption is there for all architectures, but trivial for some, it doesn't harm anything... but not sure what would clear it up.

proof/refine/Refine.thy
michaelmcinerney: Similar things happen a lot in MCS and there it is due to `scheduler_action s = resume_current_thread` being in the assumptions, and so the `scheduler_action s` field of `valid_sched_2` will be rewritten since `valid_sched` is an abbreviation. Maybe this or something similar is happening for you. It's really annoying, since then the elim rules for `valid_sched` won't fire and unfolding the definition takes a very long time. So we made `schact_is_rct` a definition to avoid this particular case and it's worked well.

Xaphiosis: Right, it's because `sched_act_init: scheduler_action init_A_st = resume_cur_thread` is in the simpset, meaning that:
`valid_sched (init_A_st :: det_state)`
is rewritten to
`valid_sched_2 (ready_queues init_A_st) (etcbs_of init_A_st) resume_cur_thread (cur_domain init_A_st) (kheap init_A_st) (cur_thread init_A_st) (idle_thread init_A_st)`

Xaphiosis: Yeah, I have to move the sched_act_init assumption up a bit, and then we end up with:
```
  assumes sched_act_init[simp]:
    "scheduler_action (init_A_st :: det_state) = resume_cur_thread"
  (* sched_act_init being in the simpset automatically expands this to a no-longer-abbreviated
     valid_sched_2 term we don't want to type out, but it's still used to simplify the proofs *)
  assumes valid_sched_init[simplified, simp]:
    "valid_sched (init_A_st :: det_state)"
```

proof/refine/Refine.thy
Xaphiosis: Will do, but this has the same 'a problem in the type signature as with full_invs, see earlier comments in this PR

proof/refine/Refine.thy
Xaphiosis: @michaelmcinerney yeah, I think this is not ideal since it foot-guns itself directly with sched_act_init being put into simp, and then valid_sched_init ending up in non-simp-form because of it. if we ever do work in this file again, might be worth unscrambling

proof/refine/Refine.thy
corlewis: If I'm right then I think changing it to `user_context` makes sense. I haven't looked at how they're used in CRefine though.

proof/refine/EmptyFail_H.thy
corlewis: Sounds like a plan. No rush for that cleanup but should we make an issue so that we don't forget this decision that we've just made?

proof/refine/ARM_HYP/KernelInit_R.thy
lsf37: I think a comment would be good, yes. At least that clears up that we are aware that it's an axiom that is not strictly necessary everywhere, and also that this is harmless.

proof/refine/Refine.thy
lsf37: I'd be fine leaving it as `'a`, but adding a comment that this is the user context, which the invariants don't constrain.

Xaphiosis: Somewhere between the two of you is calling it `'user_context`, which clearly indicates what it is, but also that it's free. Would that work?

corlewis: That works for me.

proof/refine/EmptyFail_H.thy
Xaphiosis: Will do

Xaphiosis:
Squashed and rebased. Will do a final read over the commits and merge once tests pass.
Let me know if any other commit / commit-message tuning is in order.

PR #1043, opened 2026-07-31
ignore isabelle.in.tum.de in link check

lsf37:
I got fed up with the constant link check failures, so

- fix one slightly broken link
- ignore isabelle.in.tum.de entirely


Xaphiosis (approved):
Sad, but I am also tired of seeing tests failed on every single PR.

PR #1047, opened 2026-08-21
Refine+AInvs: performance improvements (clear_named_theorems, avoid interpretation Arch and interpret Arch)

Xaphiosis:
Please review commit-by-commit. The "big" commits really consist of tiny changes to many files.

Introduce the `clear_named_theorems` command plus tests/documentation, then use it to get rid of all the named_theorems declarations throughout Refine, in lieu of a single Arch_assms which we re-use every time we want to accumulate Arch interface locale assumptions. Then we export it to a lemma name which we can access directly from outside Arch in order to interpret the interface locale.

This means `print_attributes` isn't going to be a flood of all the named_theorems.

The name perturbations that named_theorems produced are now gone, so I had to put back the arch prefixes I previously had to remove when I introduced the named_theorems uses in Refine.

A few small cleanups snuck in, and then I managed to remove every single `interpret Arch .` and `interpretation Arch .` in Refine.

Then I did the same thing for AInvs. There was more turbulence, but once again, no more `interpret Arch .` or `interpretation Arch .` remains.

Benchmarks are in progress (I wrote a tool, Corey is helping me test it), but it's definitely an improvement. E.g. internal isabelle timing reports around 8.5 minutes for RISCV64 Refine vs around 9 before this change (on my fast main machine).

@ryybrr : tagging you since you'll probably want to copy this for InfoFlow at some point; I can assist.

### Benchmark

Comparison of current master (f49402738) and named_theorems optimisation branch (August 2026). Times are *session processing times as reported by Isabelle* in seconds, do not confuse them with run_tests times. Benchmark ran on `pcr` ( i9-12900K) using 8 max threads. Times are average of 5 runs after an initial warm-up.

| ARM      | Time before | after | improved s | % | CPU time before | after | improved s | %
|----------|------------:|------:|-----------:|--:|----------------:|------:|-----------:|--:|
| AInvs       | 366  | 353 | 13 | 3.6% | 2023 | 1922 | 101 | 5.0% |
| Refine      | 761  | 721 | 40 | 5.3% | 4235 | 3929 | 306 | 7.2% |
| CBaseRefine | 714  | 698 | 16 | 2.2% | 1123 | 1100 | 23  | 2.0% |

| ARM_HYP  | Time before | after | improved s | % | CPU time before | after | improved s | %
|----------|------------:|------:|-----------:|--:|----------------:|------:|-----------:|--:|
| AInvs       | 540  | 525 | 15 | 2.8% | 2482 | 2351 | 131 | 5.3% |
| Refine      | 945  | 871 | 74 | 7.8% | 5017 | 4600 | 417 | 8.3% |
| CBaseRefine | 1014 | 968 | 46 | 4.5% | 1612 | 1525 | 87  | 5.4% |

| RISCV64  | Time before | after | improved s | % | CPU time before | after | improved s | %
|----------|------------:|------:|-----------:|--:|----------------:|------:|-----------:|--:|
| AInvs       | 296  | 295 | 1  | 0.3% | 1782 | 1761 | 21  | 1.2% |
| Refine      | 713  | 672 | 41 | 5.8% | 4069 | 3770 | 299 | 7.3% |
| CBaseRefine | 606  | 591 | 15 | 2.5% | 945  | 917  | 28  | 3.0% |

| X64      | Time before | after | improved s | % | CPU time before | after | improved s | %
|----------|------------:|------:|-----------:|--:|----------------:|------:|-----------:|--:|
| AInvs       | 437  | 430 | 7   | 1.6%  | 2408 | 2325 | 83   | 3.4% |
| Refine      | 1066 | 895 | 171 | 16.0% | 5774 | 4659 | 1115 | 19.3% |
| CBaseRefine | 936  | 918 | 18  | 1.9%  | 1491 | 1456 | 35   | 2.3% |

| AARCH64  | Time before | after | improved s | % | CPU time before | after | improved s | %
|----------|------------:|------:|-----------:|--:|----------------:|------:|-----------:|--:|
| AInvs       | 447  | 425  | 22 | 4.9% | 2522 | 2355 | 167 | 6.6% |
| Refine      | 971  | 902  | 69 | 7.1% | 5255 | 4730 | 525 | 10.0% |
| CBaseRefine | 1084 | 1080 | 4  | 0.4% | 1790 | 1687 | 103 | 5.8% |

| Combined | Time before | after | improved s | % | CPU time before | after | improved s | %
|----------|------------:|------:|-----------:|--:|----------------:|------:|-----------:|--:|
| AInvs       | 2086 | 2028 | 58  | 2.8% | 11217 | 10714 | 503  | 4.5% |
| Refine      | 4456 | 4061 | 395 | 8.9% | 24350 | 21688 | 2662 | 10.9% |
| CBaseRefine | 4354 | 4255 | 99  | 2.3% | 6961  | 6685  | 276  | 4.0% |

### Benchmark Discussion

Overall, this saves 9.2 minutes of real time across the three images, and 57 minutes of CPU time. As expected, the major improvement is Refine, with between 5.8% to 16% real time speedup. AInvs still benefited, with 0.3% to 4.9% speedup, and CBaseRefine got a marginal boost.

Apparently, clearing the pipeline stall during an `interpret Arch .` when interpreting an interface locale only results in a larger improvement when there is something to shove into the pipes before the next stall. In `CBaseRefine` which is running with skip proofs on, there isn't a lot. In `AInvs`, the `interpret` doesn't take as long, it only really becomes more expensive when we hit `Refine`. Nonetheless the 0.3% improvement on RISCV64 AInvs is unexpectedly low. Standard deviation on that run is under 1s, so it's not jitter.

Another possibility is that the Refine arch-split is very aggressive, significantly reducing the amount of arch-specific proofs, and so relying on interfaces more.

### DIFF
--- a/lib/Clear_Named_Theorems.thy
+++ b/lib/Clear_Named_Theorems.thy
@@ -1,0 +1,35 @@
+(*
+ * Copyright 2026, Proofcraft Pty Ltd
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ *)
+
+(* Clearing an existing named_theorems within the current context, allowing the same named_theorems
+   to be re-used multiple times as an accumulator. *)
+
+theory Clear_Named_Theorems
+imports Main
+keywords "clear_named_theorems" :: thy_decl
+begin
+
+ML \<open>
+local
+
+(* We need to lift from a Context.generic transformer to a local_theory transformer.
+   Context.proof_map looks like it should do this, but using it with Named_Theorems.clear does not
+   have any effect. It is unclear why that's the case.
+   Going via Local_Theory.declaration does work, even if we have to give it a "morphism" that
+   doesn't actually contain a morphism. *)
+fun local_declare f =
+  Local_Theory.declaration {syntax = false, pervasive = false, pos = \<^here>} (K f);
+
+val _ =
+  Outer_Syntax.local_theory \<^command_keyword>\<open>clear_named_theorems\<close>
+    "clear named collection of theorems"
+    (Parse.name_position >>
+      (fn (b,pos) => fn ctxt =>
+       local_declare (Named_Theorems.clear (Named_Theorems.check ctxt (b, pos))) ctxt));
+
+in end\<close>
+
+end
--- a/proof/access-control/AARCH64/ArchRetype_AC.thy
+++ b/proof/access-control/AARCH64/ArchRetype_AC.thy
@@ -16,6 +16,10 @@ lemma invs_mdb_cte':
 
 context retype_region_proofs begin interpretation Arch .
 
+(* FIXME arch-split: clean up interfaces in Retype_AI so that vs_lookup_table'
+   is properly exported *)
+interpretation retype_region_proofs_arch ..
+
 lemma state_vrefs_eq:
   "\<lbrakk> valid_vspace_objs s; valid_arch_state s \<rbrakk>
      \<Longrightarrow> state_vrefs s' = state_vrefs s"
--- a/proof/invariant-abstract/AARCH64/ArchInvariants_AI.thy
+++ b/proof/invariant-abstract/AARCH64/ArchInvariants_AI.thy
@@ -44,6 +44,10 @@ end_qualify
 
 context Arch begin arch_global_naming
 
+(* used to accumulate theorems for satisfying Arch interface assumptions;
+   remember to clear before starting a new accumulation *)
+named_theorems Arch_assms
+
 definition arch_tcb_to_iarch_tcb :: "arch_tcb \<Rightarrow> iarch_tcb" where
   "arch_tcb_to_iarch_tcb arch_tcb \<equiv>
      \<lparr> itcb_vcpu = tcb_vcpu arch_tcb, itcb_cur_fpu = tcb_cur_fpu arch_tcb \<rparr>"
--- a/proof/invariant-abstract/AARCH64/ArchRetype_AI.thy
+++ b/proof/invariant-abstract/AARCH64/ArchRetype_AI.thy
@@ -266,12 +266,7 @@ locale retype_region_proofs_arch
   + Arch
   for s :: "'state_ext :: state_ext state"
   and ty us ptr sz n ps s' dev
-
-
-context retype_region_proofs begin
-
-(* FIXME arch_split: is there any way to optimise this interpretation out? we can't nest contexts *)
-interpretation Arch .
+begin
 
 lemma valid_cap:
   assumes cap:
@@ -492,11 +487,6 @@ lemma wellformed_default_obj[Arch_assms]:
     arch_valid_obj ao s'"
   by (cases ao; clarsimp elim!: obj_at_pres simp: valid_vcpu_def
                          split: arch_kernel_obj.splits option.splits)+
-
-end
-
-
-context retype_region_proofs_arch begin
 
 lemma hyp_refs_eq:
   "state_hyp_refs_of s' = state_hyp_refs_of s"
--- a/proof/invariant-abstract/AARCH64/ArchRetype_AI.thy
+++ b/proof/invariant-abstract/AARCH64/ArchRetype_AI.thy
@@ -15,19 +15,19 @@ begin
 
 context Arch begin arch_global_naming
 
-named_theorems Retype_AI_assms
+clear_named_theorems Arch_assms (* accumulate assumptions for Retype_AI locale *)
 
-lemma arch_kobj_size_cong[Retype_AI_assms]:
+lemma arch_kobj_size_cong[Arch_assms]:
 "\<lbrakk>a = a1; c=c1\<rbrakk> \<Longrightarrow> arch_kobj_size (default_arch_object a b c)
   = arch_kobj_size (default_arch_object a1 b1 c1)"
   by (simp add: default_arch_object_def split: aobject_type.splits)
 
 
-lemma clearMemoryVM_return[simp, Retype_AI_assms]:
+lemma clearMemoryVM_return[simp, Arch_assms]:
   "clearMemoryVM a b = return ()"
   by (simp add: clearMemoryVM_def)
 
-lemma slot_bits_def2 [Retype_AI_assms]: "slot_bits = cte_level_bits"
+lemma slot_bits_def2 [Arch_assms]: "slot_bits = cte_level_bits"
   by (simp add: slot_bits_def cte_level_bits_def)
 
 definition
@@ -35,7 +35,7 @@ definition
                          ArchObject SmallPageObj, ArchObject LargePageObj, ArchObject HugePageObj,
                          ArchObject PageTableObj, ArchObject VSpaceObj}"
 
-lemma no_gs_types_simps [simp, Retype_AI_assms]:
+lemma no_gs_types_simps [simp, Arch_assms]:
   "Untyped \<in> no_gs_types"
   "TCBObject \<in> no_gs_types"
   "EndpointObject \<in> no_gs_types"
@@ -43,7 +43,7 @@ lemma no_gs_types_simps [simp, Retype_AI_assms]:
   "ArchObject ASIDPoolObj \<in> no_gs_types"
   by (simp_all add: no_gs_types_def)
 
-lemma retype_region_ret_folded [Retype_AI_assms]:
+lemma retype_region_ret_folded [Arch_assms]:
   "\<lbrace>\<top>\<rbrace> retype_region y n bits ty dev
    \<lbrace>\<lambda>r s. r = retype_addrs y ty n bits\<rbrace>"
   unfolding retype_region_def
@@ -112,7 +112,7 @@ crunch reserve_region
 crunch reserve_region
   for invs[wp]: "invs"
 
-lemma dmo_eq_kernel_restricted [wp, Retype_AI_assms]:
+lemma dmo_eq_kernel_restricted [wp, Arch_assms]:
   "\<lbrace>\<lambda>s. equal_kernel_mappings (kheap_update (f (kheap s)) s)\<rbrace>
        do_machine_op m
    \<lbrace>\<lambda>rv s. equal_kernel_mappings (kheap_update (f (kheap s)) s)\<rbrace>"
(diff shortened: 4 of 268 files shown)


lib/Clear_Named_Theorems.thy
```diff
@@ -0,0 +1,35 @@
...
+ML \<open>
+local
+
+(* We need to lift from a Context.generic transformer to a local_theory transformer.
+   Context.proof_map looks like it should do this, but using it with Named_Theorems.clear does not
+   have any effect. It is unclear why that's the case.
+   Going via Local_Theory.declaration does work, even if we have to give it a "morphism" that
+   doesn't actually contain a morphism. *)
+fun alt_proof_map f =
+  Local_Theory.declaration {syntax = false, pervasive = false, pos = \<^here>} (fn _ => f);
```
lsf37: I'd probably replace `(fn _ => f)` by `(K f)` as the more usual form in the Isabelle sources.  

`alt_proof_map` is not really the right name, because `proof_map` is exactly not the thing we're doing. Looking at the code of `Local_Theory.declaration`, `proof_map` is just one of multiple primitive steps in there. Basically `proof_map` is a primitive lifting operation, and `declaration` does all of the actual context handling of the locale. What we're doing really is a declaration command like a `declare` (and we're also giving the keyword type `thy_decl`, which is correct), so maybe `local_declare` as a name? The morphism is safe to ignore in this case, because we are not referring to any locale `fixes` or anything like that, so there is nothing to transform.

lib/Clear_Named_Theorems.thy
```diff
@@ -0,0 +1,35 @@
...
+
+val _ =
+  Outer_Syntax.local_theory \<^command_keyword>\<open>clear_named_theorems\<close>
+    "clear named collection of theorems"
+    ((Parse.name_position) >>
```
lsf37: `(Parse.name_position)` -> `Parse.name_position`

proof/refine/AARCH64/ArchInvsLemmas_H.thy
```diff
@@ -567,16 +567,19 @@ instance user_data_device :: no_vcpu by intro_classes auto
 
 end_qualify
 
+(* FIXME arch-split: koType_asidpool and koType_pte are somehow [simp] even outside Arch on AARCH64
+   due to how pre_storable *)
```
lsf37: The sentence is cut off. Due to how pre_storable works?

proof/refine/AARCH64/ArchInvsLemmas_H.thy
```diff
@@ -567,16 +567,19 @@ instance user_data_device :: no_vcpu by intro_classes auto
...
+
 instantiation AARCH64_H.asidpool :: no_vcpu
 begin
-interpretation Arch .
-instance by intro_classes auto
+instance by intro_classes
+             (auto simp: AARCH64_H.arch_kernel_object_type.distinct)
```
lsf37: Indent seems slightly off in the second line in GitHub view. The `(` should be under the `I` of `intro_classes` (globally in this commit). If that's the case already, please ignore.

lsf37 (approved):
Very nice!

So much name space pollution removed, I'm glad you pushed for looking into this.

Xaphiosis:
Updated PR description with benchmark. Will deal with PR comments after weekend.

lsf37:
Those are some pretty nice numbers, X64 is getting a real boost. Not sure why RISCV64 isn't seeing that much improvement, but overall it is an excellent picture.

proof/invariant-abstract/AARCH64/ArchInvariants_AI.thy
```diff
@@ -44,6 +44,10 @@ end_qualify
 
 context Arch begin arch_global_naming
 
+(* used to accumulate theorems for satisfying Arch interface assumptions;
+   remember to clear before starting a new accumulation *)
+named_theorems Arch_assms
```
corlewis, line 49: If we want to minimise duplication, the declaration of this named theorem could happen in a generic file.

An argument against this is that there isn't really an obvious place that it would happen. I guess it could be in either `InvariantsPre_AI` or `Include_AI`?

corlewis:
> Please recombine commands into a single line, the first is split to satisfy gitlint:

Not that big of a deal but in the past we've been ok with failing gitlint due to long commands like this, so if you want to I think it would be fine to do so again in this commit message.

corlewis:
How would you feel about splitting the 'use Arch_assms in lieu of named_theorems' commits each into two, separating out the parts that are from the sed scripts and commands?

It would make it more feasible to review, and I think could help when we're transferring this to the rt branch.

proof/invariant-abstract/AARCH64/ArchRetype_AI.thy
```diff
@@ -266,12 +266,7 @@ locale retype_region_proofs_arch
   + Arch
   for s :: "'state_ext :: state_ext state"
   and ty us ptr sz n ps s' dev
-
-
-context retype_region_proofs begin
-
-(* FIXME arch_split: is there any way to optimise this interpretation out? we can't nest contexts *)
-interpretation Arch .
+begin
```
corlewis, line 270: Out of curiosity, any idea what has changed that we can avoid this interpretation now?

proof/access-control/AARCH64/ArchRetype_AC.thy
```diff
@@ -16,6 +16,8 @@ lemma invs_mdb_cte':
 
 context retype_region_proofs begin interpretation Arch .
 
+interpretation retype_region_proofs_arch ..
```
corlewis, line 21: I think a FIXME here would help, saying something along the lines of what you have in the commit message for this change. We're unlikely to remember the current context if we come across this interpretation again in the future, and depending on us finding the commit message feels optimistic.

corlewis (approved):
Very nice cleanup, that until you started talking about I hadn't even considered as a possibility! Like Gerwin said, thank you for pushing for this.

It's also very nice to see the continued improved performance, thank you for the thorough benchmarking. It's hard to tell how much to read into some of the outliers; the 0.4% improvement for AARCH64 CBaseRefine is also surprising, particularly with it having the largest change in CPU time.

lib/Clear_Named_Theorems.thy
Xaphiosis: Will do. With these changes, is the comment acceptable?

lib/Clear_Named_Theorems.thy
Xaphiosis: will do

proof/refine/AARCH64/ArchInvsLemmas_H.thy
Xaphiosis: Thanks, should've been "how pre_storable instantiation is done"; it's not how it works in general, it was a questionable choice on AARCH64. The FIXME is here because on other arches I use `L4V_ARCH.koType_pte` (and same for asid pool) but here it causes a duplicate simp warning.

Xaphiosis: well, other arches. only ARM_HYP.

proof/refine/AARCH64/ArchInvsLemmas_H.thy
Xaphiosis: Yes, it was also off on ARM_HYP. Will fix while we're here.

proof/invariant-abstract/AARCH64/ArchInvariants_AI.thy
Xaphiosis: I had considered this, but then had a bad feeling about advocating for `context Arch begin arch_global_naming` in a generic theory. Normally I only allow `context Arch` followed by a requalify with `Arch.`. Technically your proposal works, but I am unsure about the precedent it sets. In terms of line saving, we'd be removing 3 lines for 5 arches, then adding back 5 lines, saving 10 lines.

Xaphiosis:
> > Please recombine commands into a single line, the first is split to satisfy gitlint:
> 
> Not that big of a deal but in the past we've been ok with failing gitlint due to long commands like this, so if you want to I think it would be fine to do so again in this commit message.

Good point. Yeah, I think having the proper command for people to copy without a weird comment about commands being broken up is better than a gitlint violation. I'll change it.

Xaphiosis:
> How would you feel about splitting the 'use Arch_assms in lieu of named_theorems' commits each into two, separating out the parts that are from the sed scripts and commands?
> 
> It would make it more feasible to review, and I think could help when we're transferring this to the rt branch.

I thought about it, but it ended up more convoluted. For AInvs, there is much more mess and so the sed substitutions didn't land a bunch of the time and had to be hand-held. For Refine it might be useful, but then I also had to touch up after. I'm not sure how much help it'll be in an actual merge. In the interest of lack of time this week, unless you really want me to, I'll keep it as is. Worst case scenario if you need it from me in future, I can regenerate the diff for you.

proof/invariant-abstract/AARCH64/ArchRetype_AI.thy
Xaphiosis: The locale hierarchy got very tangled here, I don't think this interpretation was ever necessary. I was not able to untangle it more, because the approach uses pulls in the arch version of the locale whenever it feels like it, resulting is slow things like `interpretation retype_region_proofs_arch ..`. Removing it did break the AARCH64 Access proof, because we needed another lemma pulled out of the arch locale. That said, interpreting locales like that is actually the "proper" way, just that Arch is weird since it has no assumptions, opening up the situation to optimisations.
One day maybe we'll overhaul this for a bit of a speedup.

proof/access-control/AARCH64/ArchRetype_AC.thy
Xaphiosis: I'll add
 ```
 (* FIXME arch-split: clean up interfaces in Retype_AI so that vs_lookup_table'
   is properly exported *)
```

All of these `interpretation Arch .` should've had a `(* FIXME arch-split *)` on them anyhow, but the comment explains the extra interpretation hopefully. Let me know if it's enough.

Xaphiosis:
Hopefully addressed PR comments.
gitlint violation is expected due to that long sed line, and theory linter is complaining about `lib/test/Clear_Named_Theorems.thy` (I'm starting to wonder whether the theory linter should not bother with theories in `lib/test/`)

lib/Clear_Named_Theorems.thy
lsf37: Yes, absolutely

lsf37 (approved):
Happy with the changes, good to merge from my side!

proof/invariant-abstract/AARCH64/ArchInvariants_AI.thy
corlewis: Yeah, fair enough. Not worth adding that confusion for such a minor line gain.

proof/access-control/AARCH64/ArchRetype_AC.thy
corlewis: That looks good, thanks!

corlewis (approved):
Same, looks good to merge to me as well!

Xaphiosis:
Well, the gitlint violation means I can't merge this, as the gitlint test is marked as Required. So either @lsf37 overrides this, or I have to revert my commit message change.

lsf37:
Override done!

PR #1050, opened 2026-08-25
update to Isabelle 2025-2

lsf37:
Update all proofs to Isabelle-2025-2 (which is the same as Isabelle-2025-1 plus one bug fix).

This includes a sync of `Word_Lib` and `Simpl` with afp-2025-2 in separate commits. Please keep in mind that these are upstream changes when reviewing.

The rest is was comparatively painless. Main noise is that `null_def` is being rewritten automatically now, and the usual few theorem renames. The ML updates happened a long time ago, so I don't quite remember the details any more, but I don't think there was anything big.

I did retire `Try_Methods.thy` in `lib/`, because it was too annoying to update and it is unused these days and superseded by `try` and `try0`. I haven't added new methods to `try0`, but if people are keen, we could do that, based on the diff . (Personally I use it very rarely). 

There is currently one commit per major proof session (all architectures squashed already). We could squash more, but this is probably nicer for review.

Test with: seL4/isabelle@ts-2025-2

(I might need to update the AWS VMs first with the new Isabelle2025-2 components for the test to work, but we'll see. Edit: it's downloading the new components, so it seems to be generally working. Might time out, though. Will update the VMs tomorrow.)


lsf37:
The update should rebase fairly easily over #1047 and #1047 should probably go in first, but otherwise I'd like to merge this fairly soon before it goes out of date.

lib/Word_Lib/Norm_Words.thy
```diff
@@ -92,13 +92,13 @@ private lemma "- 2 = (13 + 1 :: 'a::len word)"
 private lemma "7 = (3 :: 2 word)"
   by simp
 
-private lemma "- 2 = (22 :: 3 word)" 
+private lemma "f (- 2) = f (22 :: 3 word)"
```
corlewis, line 95: Are you pushing these fixed tests back to the afp version? It would be nice if we didn't have to do this again next Isabelle update.

lib/concurrency/Prefix_Refinement.thy
```diff
@@ -1937,7 +1937,7 @@ lemma prefix_refinement_mapM[rule_format]:
         apply (rule prefix_refinement_triv_pre, rule prefix_refinement_return_imp, simp)
        apply wp
        apply fastforce
-     apply (simp | wp | blast dest: validI_prefix_closed)+
+     apply (simp | wp | blast del: validI_prefix_closed dest: validI_prefix_closed)+
```
corlewis, line 1940: Any memory of what changed to cause this to be needed? It looks like `validI_prefix_closed` is declared `elim!`, which I guess doesn't work for this proof, but it's then not clear why you now need to `del` it first.

corlewis (approved):
Looks good, thanks for doing this!

proof/refine/ARM/ArchArch_R.thy
```diff
@@ -1491,7 +1491,7 @@ lemma arch_decodeInvocation_wf[wp]:
        apply (erule cte_wp_at_weakenE')
        apply (simp, drule_tac t="cteCap c" in sym, simp add: isPDCap_def)
       apply (subst (asm) conj_assoc [symmetric])
-      apply (subst (asm) assocs_empty_dom_comp [symmetric])
+      apply (subst (asm) assocs_empty_dom_comp[simplified,symmetric])
```
Xaphiosis: space after simplified to match others

proof/refine/ARM/ArchArch_R.thy
```diff
@@ -784,7 +784,7 @@ shows
          apply (simp only: bindE_assoc)
          apply (rule corres_splitEE)
             apply (rule corres_whenE)
-              apply (subst assocs_empty_dom_comp [symmetric])
+              apply (subst assocs_empty_dom_comp[simplified,symmetric])
```
Xaphiosis: space after simplified to match others.
also: if `assocs_empty_dom_comp` is no longer in simp-normal form, should we update the lemma so that it is in simp-normal form?
I've noticed that in this commit you're sometimes doing `[simplified, symmetric]` and sometimes `[symmetric, simplified]` for `assocs_empty_dom_comp`. It would be good to pick one over the other.

Xaphiosis: Checking the rest of this commit, it appears that it's backwards in ARM ArchArch_R compared to the other arches.

proof/crefine/ARM_HYP/Invoke_C.thy
```diff
@@ -243,7 +243,7 @@ lemma decodeDomainSetSetInvocation_ccorres:
                  cong: StateSpace.state.fold_congs globals.fold_congs)
       apply (rule syscall_error_throwError_ccorres_n)
       apply (simp add: syscall_error_to_H_cases)
-     apply (simp add: null_def excaps_map_def unat_numDomains_to_H)
```
Xaphiosis, line 246: for more understanding: what happened to null_def?

proof/crefine/ARM/Ipc_C.thy
```diff
@@ -1481,8 +1479,8 @@ proof -
      apply (frule (1) option_to_0_imp)
      apply (subst drop_zip)
      apply (subst drop_n)
-     apply (clarsimp simp:  n_msgRegisters_def numeral_eqs
-                            mapM_cong[OF msg_aux, simplified numeral_eqs])
+     apply (clarsimp simp: n_msgRegisters_def mapM_cong[OF msg_aux]
+                     simp del: upt_rec_numeral)
```
Xaphiosis, line 1483: this is mildly concerning for CRefine and multi-config proofs, can you say more what happened here?

Xaphiosis (approved):
Some questions and a minor consistency nitpick. Small update, yay. Well done.

lib/Word_Lib/Norm_Words.thy
lsf37: I believe I have already pushed them back, but I will double check before the next one

lib/concurrency/Prefix_Refinement.thy
lsf37: It's been a while, but AFAIR the problem was exactly the `elim!` declaration which marks it as a safe rule. The `del` removes that, and `dest` adds it an unsafe rule that `blast` will do backtracking on.

Strictly speaking this means the rule is not safe and only be `elim` at most, but it seems to be working fine everywhere else, so I didn't touch the global declaration.

proof/refine/ARM/ArchArch_R.thy
lsf37: Well spotted. I suspect those ARM changes are from the initial run I did earlier this year and I forgot about cleaning that up. Will do that now.

I haven't changed the goal statement, because there are still occurrences of it that work unchanged, so I didn't want to disturb them. Looking at the grep output now, though, these are all as part of a `simp` call which would rewrite into `null` and then rewrite the `null`. So it should be safe to change. I'll give it a go and see if anything breaks. (That would mean we can remove all those `simplified`.)

proof/crefine/ARM_HYP/Invoke_C.thy
lsf37: It got renamed and the replacement was made `[simp]`, so `null xs` now automatically rewrites to `xs =[]`.

```
* Consolidated auxiliary operations intended for code generation:
...
    const null ~> List.null
    thm null_def ~> List.null_iff [simp]
```

Overall it turns out better, I have to say, although I'm a bit annoyed that there is now no predicate any more for `= []` that won't automatically become an equation. Sometimes you want that. In our case, the `null` occurrences are from Haskell and rewriting them automatically is fine.

proof/crefine/ARM/Ipc_C.thy
lsf37: Unfortunately nothing precise. As usual, Word_Lib has been disturbed a bit, and automation works slightly differently, this time mostly with numerals especially between Nat and Word numerals. Overall, it seems more stable than before (less need for `numeral_eqs` and things like that), but there may well be wrinkles, especially in the platform variations, yes.

We'll have to see when it's merged and the fuller set runs. I can try a few locally, but they are too many to check exhaustively first.

PR #1054, opened 2026-09-06
spec: remove stale FIXMEs and dead code

lsf37:
- remove all FIXMEs that are no longer true, e.g. because completed or code has changed.
- remove FIXMEs that are questions that are answered, e.g. "unclear if verification needs X"
- remove machine ops and constants that are not mentioned anywhere else
- remove out-of-date kernel init code for ARM_HYP
- remove FIXME for unimplemented machine ops (now standard)
- make FIXME x64-vtd tags consistent
- make FIXME ARMHYP_SMMU tags consistent

Hopefully this will make the remaining FIXMEs more actionable. (-127 FIXMEs, +12 FIXMEs)

Actionable FIXMEs are good. Going through these led to https://github.com/seL4/seL4/issues/1744


spec/haskell/src/SEL4/Kernel/VSpace/AARCH64.hs
```diff
@@ -68,15 +68,13 @@ lookupIPCBuffer isReceiver thread = do
...
 
--- FIXME: make this a Reader Monad when we move to MCS
 getPoolPtr :: ASID -> Kernel (Maybe (PPtr ASIDPool))
 getPoolPtr asid = do
     assert (asid > 0) "ASID 0 is used for objects that are not mapped"
     assert (asid <= snd asidRange) "ASID out of range"
     asidTable <- gets (armKSASIDTable . ksArchState)
     return $ asidTable!(asidHighBitsOf asid)
 
--- FIXME: make this a Reader Monad when we move to MCS
```
Xaphiosis, line 79: did this happen?

spec/haskell/src/SEL4/Kernel/VSpace/X64.lhs
```diff
@@ -180,19 +180,20 @@ Locating the page directory for a given ASID is necessary when updating or delet
 >             return ptr
 >         Nothing -> throw InvalidRoot
 
-> -- FIXME x64: these are all now unused.
+These are used in proofs only, and will be translated to the corresponding table_at
+predicates.
 
 > checkPML4At :: PPtr PML4E -> Kernel ()
-> checkPML4At _ = return ()
+> checkPML4At _ = error "defined in Isabelle"
```
Xaphiosis, line 187: Before this PR, there is exactly *one* occurrence of `error "defined in Isabelle"` (for `archMDBAssertions`, all other commands and constants take a different form. Are you sure this is the form you want to choose to propagate?

spec/haskell/src/SEL4/Machine/Hardware/ARM.lhs
```diff
@@ -313,16 +313,16 @@ caches must be done separately.
 #ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
 
 > setCurrentPDPL2 :: PAddr -> MachineMonad ()
-> setCurrentPDPL2 = error "FIXME ARMHYP machine callback unimplemented"
+> setCurrentPDPL2 = error "ARMHYP machine callback unimplemented"
 
 > writeContextIDAndPD :: HardwareASID -> PAddr -> MachineMonad ()
-> writeContextIDAndPD = error "FIXME ARMHYP  machine callback unimplemented"
+> writeContextIDAndPD = error "ARMHYP  machine callback unimplemented"
```
Xaphiosis, line 319: while we're here, clear the double-space

spec/haskell/src/SEL4/Model/StateData/ARM.lhs
```diff
@@ -99,31 +99,7 @@ FIXME ARMHYP_SMMU ARMHYP missing IO ASID to PD map for SMMU
...
->             armKSNextASID = minBound,
->             armKSASIDMap = funPartialArray (const Nothing) asidRange,
->             armHSCurVCPU = Nothing,
->             armKSGICVCPUNumListRegs = error "FIXME ARMHYP read from platform",
->             armUSGlobalPD = error "FIXME ARMHYP address of C global constant",
->             armKSKernelVSpace =
->                 (\vref -> if vref < mask 20 then ArmVSpaceKernelWindow
->                                             else ArmVSpaceInvalidRegion)
->             }
+> newKernelState data_start = error "no initial state defined for ARM_HYP"
```
Xaphiosis, line 103: would we consider no initial state defined as a FIXME, given one day we'd like to verify kernel init again?

spec/haskell/src/SEL4/Object/Interrupt.lhs
```diff
@@ -106,8 +106,6 @@ An IRQ handler capability allows a thread possessing it to set an endpoint which
 > toBool :: Word -> Bool
 > toBool w = w /= 0
 
-%FIXME x64 naming: this should be called perform, not invoke, same for CNode
-
```
Xaphiosis, line 110: do we have any further thoughts on this, given it is still named invoke?

spec/haskell/src/SEL4/Object/Notification.lhs
```diff
@@ -32,7 +32,7 @@ This module specifies the behavior of notification objects.
 
 \subsection{Sending Signals}
 
-> -- helper function, FIXME redundant with Structure.isReceive
+> -- redundant with Structure.isReceive, but kept because written as case
```
Xaphiosis, line 35: what is written as case? maybe a word or two extra would help here (or went missing)

spec/haskell/src/SEL4/Object/Structures.lhs
```diff
@@ -356,8 +356,7 @@ The basic structure is a double-linked list. The algorithm used to determine the
 
 A user thread may be in the following states:
 
-%FIXME: Mangled for datatype constructor order
-
+%Note: keep datatype constructor order
```
Xaphiosis, line 359: I don't understand/remember what this comment means.

spec/haskell/src/SEL4/Object/Structures/X64.lhs
```diff
@@ -99,8 +99,8 @@ size.
 >                 KOPDPTE _ -> 3
 >                 KOPML4E _ -> 3
 >--                KOIOPTE _ -> 3
->--                KOIOCTE _ -> 3
->--                KOIORTE _ -> 3 -- FIXME: Not correct ?
+>--                KOIOCTE _ -> 4 -- FIXME x64-vtd: check IOCTE size
+>--                KOIORTE _ -> 4 -- FIXME x64-vtd: check IORTE size
```
Xaphiosis, line 103: I was wondering whether `vtd` or `vt-d` would be the tag. It turns out that in master there is exactly one `vt-d`, so while we are here if you haven't caught it might be good to take care of it as well:
```
spec/machine/X64/MachineOps.thy
351:(* FIXME x64: VT-d
```

spec/machine/X64/MachineOps.thy
```diff
@@ -348,18 +348,6 @@ where
   "invalidateLocalPageStructureCacheASID vspace asid \<equiv>
      machine_op_lift (invalidateLocalPageStructureCacheASID_impl vspace asid)"
 
-(* FIXME x64: VT-d
```
Xaphiosis, line 351: ah, you did catch it after all!

Xaphiosis (approved):
Usual minor bikesheds and questions. Thanks for the cleanup.

spec/haskell/src/SEL4/Kernel/VSpace/X64.lhs
lsf37: That is probably exactly the one I looked at :) (I do remember looking at something). Happy to use anything else there. Using grep, we seem to mostly have either "Unimplemented" or "Unimplemented - machine op".

If there are no other preferences, I'd take the latter.

spec/haskell/src/SEL4/Model/StateData/ARM.lhs
lsf37: In that case it doesn't need a FIXME, because it'll be an obvious part of the many things that need to be defined there. A FIXME should be something that actually needs to be fixed. I left the ones for SMMU/VT-d in, because those are about 70% done and the FIXMEs there are good hints at what to look at, but here it'd just be like everything else that needs to be done.

spec/haskell/src/SEL4/Object/Interrupt.lhs
lsf37: It's consistently named "invoke" on all other architectures (nothing to do with x64) and for multiple functions. It is indeed not consistent with other top-level functions in Haskell that are named "perform", but there is also 0% likelihood that we will ever want the pain of renaming something as deeply ingrained as CNode invoke. So there is no point to keeping it as an x64 FIXME for something like Interrupt.

We could add an issue for making the naming consistent there overall, but I don't think the cost/benefit equation will ever come out in favour of doing a rename (just think of all the lemmas that have invoke in their name that will then be inconsistent if the function is renamed).

spec/haskell/src/SEL4/Object/Notification.lhs
lsf37: Yeah, that is a bit short. I was counting on it being visible that the line directly below is a case statement. Would "case statement" in the text be the right trigger?

spec/haskell/src/SEL4/Object/Structures.lhs
lsf37: The original FIXME refers to the order of the constructors (and maybe hence the text) being mangled/weird, because usually you would start with the simple ones. There is nothing to fix, though, we want it this way for the proof.

The note is trying to say that the weird order is intentional and should not be changed. It seems like that is not coming across. How about "Note: datatype constructor order below is determined by the proof, do not change for aesthetics".

spec/haskell/src/SEL4/Object/Structures/X64.lhs
lsf37: Good check, and I did get that one :-)

spec/haskell/src/SEL4/Kernel/VSpace/AARCH64.hs
lsf37: In the sense that the Haskell spec for AArch64 is now done on rt, the FIXME is now resolved for master. In the sense of "did we actually make it a reader monad", I'm not sure, but if not it should be a FIXME on rt, not master.

Just checked, it is still a FIXME on `rt`, and could stay one there, but we could also decide now, then we don't need to be careful in the merge of this commit to rt -- @michaelmcinerney would it make sense for `getPoolPtr` and `getASIDPoolEntry` to become reader monads in Haskell on `rt`? I put that FIXME in, because they are reader monads in ASpec and we don't have a Haskell reader monad on master. The upside would be free properties (no state change), the downside would be that we'd need to tweak their Refine proofs, because we'd be changing their definition compared to master.

spec/haskell/src/SEL4/Object/Structures.lhs
Xaphiosis: Yes, that's better, thank you

spec/haskell/src/SEL4/Object/Notification.lhs
Xaphiosis: Maybe I'm confused because I'm looking at the diff, but this might work:
`but kept because this version is written as a case statement`

spec/haskell/src/SEL4/Object/Interrupt.lhs
Xaphiosis: Yeah especially with two main branches around, it's probably not the time to be thinking about this. Maybe if we're alive after MCS takes over or something. You're right to drop it from the FIXME pile.

spec/haskell/src/SEL4/Model/StateData/ARM.lhs
Xaphiosis: Haha, it's a TODO then! Fair enough.
Complete bikeshed aside: this PR made me think of how we have "error" in the haskell for things that aren't errors. I've been wondering if it would ever be useful, perhaps as a method assisting identifying assumptions, to wrap `error` in something more indicative of the purpose here, like `machine_op` or whatever. It's neither urgent nor important, but I'd like to hear your thoughts.

spec/haskell/src/SEL4/Kernel/VSpace/X64.lhs
Xaphiosis: I think keeping n-flavoured soup to n flavours is my preference, and if this is a machine op, then `Unimplemented - machine op` is good especially if we've seen it before. If it's not a machine op, then `Unimplemented`. There's quite a bit of variety, in ye olde days it was `Unimplemented callback`, and sometimes there's one dash, sometimes two. I think your latter works, given we probably should style on AARCH64 which never actually had callbacks.

spec/haskell/src/SEL4/Kernel/VSpace/AARCH64.hs
Xaphiosis: Oh wow, I had not realised we don't have the Haskell reader monad on master! That explains the desire to drop the FIXME. Let's see what Michael says then.

michaelmcinerney: Yeah it would make sense to make those two reader monads on `rt`. It would be very straightforward for `rt`. We don't have Refine proofs for AARCH64 yet, so we could make the definitions read monads now without having to redo things, but it would be a minor change in any case.

spec/haskell/src/SEL4/Object/Notification.lhs
lsf37: Happy to take that

spec/haskell/src/SEL4/Model/StateData/ARM.lhs
lsf37: That's an excellent idea! We've used `error` so far mostly because we were too lazy to add something new, but a wrapper would be a good marker, wouldn't need a string, and could even be implemented by something else later if we wanted to. The hardest bit will be the name :-)

spec/haskell/src/SEL4/Kernel/VSpace/AARCH64.hs
lsf37: Ok, cool, let's do it. If you have time, that would be a good thing to do before the next merge, because then it will conveniently conflict (where you or @corlewis can then choose the new rt version) and not just silently drop the FIXME.

spec/haskell/src/SEL4/Model/StateData/ARM.lhs
lsf37: I'll go with `isabelleMachineOp` for now to indicate where it is defined. Happy to receive other suggestions, though.

lsf37: And there are immediately instances where it does not quite fit, because they are not machine ops. I'll go with just `isabelleOp` for now.

lsf37:
I have added two separate commits, one for using `isabelleOp` for everything that is defined in Isabelle and making unimplemented boot code comments consistent, and one for removing 2 old FPU machine ops that I found during that operation + the pageColourBits definition that is unused.

Please have another brief look over these before I merge.

lsf37:
(rebased)

PR #1056, opened 2026-09-10
clean up all warnings in Refine

Xaphiosis:
Refine is warning-free possibly for the first time in history, hopefully
we can keep it that way.

Adjusted:
- use `del:` to remove safe elimination rules prior to using them as
  destruction rules (prevents warning from blast), or alternatively
  use `supply *[rule del]`
- after interpreting a locale inside another and re-binding a name, move
  any attribute declarations for re-bound theorems to after the locale,
  otherwise it isn't clear which theorem we are trying to tag
- vcpuBits_def removed from objBits_defs (clash with bit_simps)
- bit_simps, projectKOs and archObjSize_def were already simp
- declares which already happened earlier
- obj_at_simps was only used on 32-bit arches, and had duplicates
- raw_tcb_cte_cases_simps 0 case was a duplicate
- clearMemory was already crunch_ignore
- use no_valid_cap_syn to prevent spec_validE vs valid_cap_syn ambiguity
- adjust some `no_0_*[elim!]` lemmas to be proper `[dest!]` lemmas
- clear crunches that proved nothing, combine overlapping crunches

### DIFF
--- a/proof/invariant-abstract/Invariants_AI.thy
+++ b/proof/invariant-abstract/Invariants_AI.thy
@@ -394,6 +394,11 @@ abbreviation
 where
   "s \<turnstile> c \<equiv> valid_cap c s"
 
+(* include in order to prevent syntax ambiguity warning with spec_validE *)
+bundle no_valid_cap_syn begin
+no_notation valid_cap_syn ("_ \<turnstile> _" [60, 60] 61)
+end
+
 definition
   "valid_caps cs s \<equiv> \<forall>slot cap. cs slot = Some cap \<longrightarrow> valid_cap cap s"
 
--- a/proof/refine/ARM/ArchInvsLemmas_H.thy
+++ b/proof/refine/ARM/ArchInvsLemmas_H.thy
@@ -111,7 +111,6 @@ clear_named_theorems Arch_assms (* accumulate assumptions for Invariants_H_cte_a
 
 (* FIXME arch-split: for proofs which require exact offsets lining up instead of cteSizeBits *)
 lemma raw_tcb_cte_cases_simps:
-  "tcb_cte_cases 0  = Some (tcbCTable, tcbCTable_update)"
   "tcb_cte_cases 16 = Some (tcbVTable, tcbVTable_update)"
   "tcb_cte_cases 32 = Some (tcbReply, tcbReply_update)"
   "tcb_cte_cases 48 = Some (tcbCaller, tcbCaller_update)"
@@ -252,7 +251,7 @@ begin
 
 lemma pspace_in_kernel_mappings_update' [iff]:
   "pspace_in_kernel_mappings' (f s) = pspace_in_kernel_mappings' s"
-  by (simp add: pspace_in_kernel_mappings'_def)
+  by simp
 
 lemma valid_asid_table_update' [iff]:
   "valid_asid_table' t (f s) = valid_asid_table' t s"
@@ -308,7 +307,7 @@ lemma tcb_hyp_refs_of'_simps[simp]:
 
 lemma refs_of_a'_simps[simp]:
   "refs_of_a' ako = {}"
-  by (auto simp: refs_of_a'_def)
+  by auto
 
 lemma hyp_refs_of_hyp_live':
   "hyp_refs_of' ko \<noteq> {} \<Longrightarrow> hyp_live' ko"
@@ -429,7 +428,7 @@ end (* typ_at_props' *)
 context Arch begin arch_global_naming
 
 lemmas bit_simps' = pteBits_def pdeBits_def asidHighBits_def asid_low_bits_def word_size_bits_def
-                    asid_high_bits_def bit_simps
+                    asid_high_bits_def
 
 lemma objBitsT_simps:
   "objBitsT EndpointT = epSizeBits"
--- a/proof/refine/ARM/ArchVSpace_R.thy
+++ b/proof/refine/ARM/ArchVSpace_R.thy
@@ -892,7 +892,7 @@ lemma deleteASID_corres:
       apply (drule Some_to_the)
       apply (wpsimp wp: getASID_wp)+
    apply (clarsimp simp: valid_arch_state_def valid_asid_table_def
-                  dest!: invs_arch_state)
+                   del: invs_arch_state dest!: invs_arch_state)
    apply blast
   apply (clarsimp simp: valid_arch_state'_def valid_asid_table'_def)
   done
--- a/proof/refine/ARM_HYP/ArchInvsLemmas_H.thy
+++ b/proof/refine/ARM_HYP/ArchInvsLemmas_H.thy
@@ -30,7 +30,7 @@ lemmas wordRadix_def' = wordRadix_def[simplified]
 lemma wordSizeCase_simp[simp]: "wordSizeCase a b = a"
   by (simp add: wordSizeCase_def wordBits_def word_size)
 
-lemmas objBits_defs = tcbBlockSizeBits_def epSizeBits_def ntfnSizeBits_def cteSizeBits_def vcpuBits_def
+lemmas objBits_defs = tcbBlockSizeBits_def epSizeBits_def ntfnSizeBits_def cteSizeBits_def
 lemmas untypedBits_defs = minUntypedSizeBits_def maxUntypedSizeBits_def
 lemmas objBits_simps = objBits_def objBitsKO_def word_size_def archObjSize_def
 lemmas objBits_simps' = objBits_simps objBits_defs
@@ -108,7 +108,6 @@ clear_named_theorems Arch_assms (* accumulate assumptions for Invariants_H_cte_a
 
 (* FIXME arch-split: for proofs which require exact offsets lining up instead of cteSizeBits *)
 lemma raw_tcb_cte_cases_simps:
-  "tcb_cte_cases 0  = Some (tcbCTable, tcbCTable_update)"
   "tcb_cte_cases 16 = Some (tcbVTable, tcbVTable_update)"
   "tcb_cte_cases 32 = Some (tcbReply, tcbReply_update)"
   "tcb_cte_cases 48 = Some (tcbCaller, tcbCaller_update)"
@@ -249,7 +248,7 @@ begin
 
 lemma pspace_in_kernel_mappings_update' [iff]:
   "pspace_in_kernel_mappings' (f s) = pspace_in_kernel_mappings' s"
-  by (simp add: pspace_in_kernel_mappings'_def)
+  by simp
 
 lemma valid_asid_table_update' [iff]:
   "valid_asid_table' t (f s) = valid_asid_table' t s"
@@ -470,7 +469,7 @@ end (* typ_at_props' *)
 context Arch begin arch_global_naming
 
 lemmas bit_simps' = pteBits_def pdeBits_def asidHighBits_def asid_low_bits_def word_size_bits_def
-                    asid_high_bits_def bit_simps
+                    asid_high_bits_def
 
 lemma objBitsT_simps:
   "objBitsT EndpointT = epSizeBits"
--- a/proof/refine/CNodeInv_R.thy
+++ b/proof/refine/CNodeInv_R.thy
@@ -908,10 +908,15 @@ termination finaliseSlot'
 lemmas finaliseSlot'_simps_ext =
     finaliseSlot'.simps [THEN ext [where f="finaliseSlot' slot exp" for slot exp]]
 
+context includes no_valid_cap_syn begin
+
 lemmas finalise_spec_induct = finaliseSlot'.induct[where P=
     "\<lambda>sl exp s. s \<turnstile> \<lbrace>P sl exp\<rbrace> finaliseSlot' sl exp \<lbrace>Q sl exp\<rbrace>,\<lbrace>E sl exp\<rbrace>" for P Q E]
 
+end
+
 lemma finaliseSlot'_preservation:
+  includes no_valid_cap_syn
   assumes wp:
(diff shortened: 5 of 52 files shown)


proof/refine/ARM/ArchVSpace_R.thy
```diff
@@ -892,7 +892,7 @@ lemma deleteASID_corres:
       apply (drule Some_to_the)
       apply (wpsimp wp: getASID_wp)+
    apply (clarsimp simp: valid_arch_state_def valid_asid_table_def
-                  dest!: invs_arch_state)
+                   del: invs_arch_state dest!: invs_arch_state)
```
corlewis, line 895: This does seem like the best way of handling it but it's frustrating that we need to duplicate rules like this to avoid warnings.

proof/refine/CNodeInv_R.thy
```diff
@@ -908,10 +908,15 @@ termination finaliseSlot'
 lemmas finaliseSlot'_simps_ext =
     finaliseSlot'.simps [THEN ext [where f="finaliseSlot' slot exp" for slot exp]]
 
+context includes no_valid_cap_syn begin
+
 lemmas finalise_spec_induct = finaliseSlot'.induct[where P=
     "\<lambda>sl exp s. s \<turnstile> \<lbrace>P sl exp\<rbrace> finaliseSlot' sl exp \<lbrace>Q sl exp\<rbrace>,\<lbrace>E sl exp\<rbrace>" for P Q E]
```
corlewis, line 914: This is some pretty bad style, thoughts on fixing it while we're making changes here?

proof/refine/Finalise_R.thy
```diff
@@ -2527,6 +2525,7 @@ lemma unbindNotification_corres:
            apply (clarsimp simp: ntfn_relation_def split:Structures_A.ntfn.splits)
           apply (rule setBoundNotification_corres)
          apply (wp gbn_wp' gbn_wp)+
+   supply [rule del] = invs_valid_objs invs_valid_objs' (* suppress elim warning *)
```
corlewis, line 2528: I think I prefer doing `del:` directly in the relevant steps instead of this sort of `supply`. It's fine in cases like this where it's towards the end of a proof, but it doesn't feel like a pattern that we want to encourage. If the proof becomes more complicated and people don't notice that the rules have been removed from `elim` then it could lead to confusing states.

proof/refine/InvariantUpdates_H.thy
```diff
@@ -86,10 +89,13 @@ lemma ifunsafe_update [iff]:
 (* exports from Arch locale version (safe for generic use) *)
 interpretation Arch_p_arch_idle_update_eq' ..
 
-lemmas valid_arch_state_update'[iff] = valid_arch_state_update'
+lemmas valid_arch_state_update' = valid_arch_state_update'
 
 end
 
+(* avoid duplicate declaration warning when re-binding name after interpretation *)
+lemmas (in p_arch_idle_update_eq') [iff] =  valid_arch_state_update'
```
corlewis, line 97: Extra space after `=`

proof/refine/IpcCancel_R.thy
```diff
@@ -582,6 +583,7 @@ lemma (in delete_one_gen) cancelIPC_ReplyCap_corres:
        od)"
   proof -
   show ?thesis
+  supply state_relation_pspace_relation[rule del] (* suppress safe elim warning *)
```
corlewis, line 586: This is a case where it doesn't feel right to me to remove `state_relation_pspace_relation` as an `elim` rule in the whole proof, instead of just in the one step that uses it as a `dest` rule. I guess it's fine since you have it working but I could easily see other steps still wanting to use it as an `elim` rule.

proof/refine/IpcCancel_R.thy
```diff
@@ -870,7 +872,7 @@ proof -
          y \<leftarrow> setEndpoint epptr ep';
          setThreadState Inactive t
       od \<lbrace>\<lambda>rv. invs'\<rbrace>"
-    supply no_0_obj_at'[rule del] (* avoid weak elim rule warning *)
+    supply no_0_obj_at'[rule del]
```
corlewis, line 875: Out of curiosity, is this still needed? If it is, why remove the comment now?

corlewis (approved):
Very nice, thank you for doing this!

It does feel a bit odd to me about how we have to delete `elim` rules when using them as `del` rules but that's a fight with Isabelle that probably isn't worth having.

proof/invariant-abstract/Invariants_AI.thy
```diff
@@ -394,6 +394,11 @@ abbreviation
 where
   "s \<turnstile> c \<equiv> valid_cap c s"
 
+(* include in order to prevent syntax ambiguity warning with spec_validE *)
+bundle no_valid_cap_syn begin
+no_notation valid_cap_syn ("_ \<turnstile> _" [60, 60] 61)
+end
```
lsf37, line 400: Looking at it again, we should probably consider the new-ish support for the `no` keyword for syntax bundles. This means, we'd have a positive syntax bundle `valid_cap_syn`, and could then use `unbundle no valid_cap_syn` and `includes no valid_cap_syn` to remove it. That provides flexibility for both directions.

The definitions would be

```suggestion
abbreviation valid_cap_syn :: "'z::state_ext state \<Rightarrow> cap \<Rightarrow> bool"  where
"valid_cap_syn s c \<equiv> valid_cap c s"

(* use with "includes no valid_cap_syn" in order to prevent syntax ambiguity warning with spec_validE *)
open_bundle valid_cap_syn begin
notation valid_cap_syn ("_ \<turnstile> _" [60, 60] 61)
end
```

(`open_bundle` is the same as `bundle x <block>` followed directly by `unbundle x`)

It's not entirely great for this one, because the abbreviation cannot be moved into the bundle and when its syntax is removed, people will see `valid_cap_syn` in the output. But that is also the case with the current implementation in this PR, more a general annoyance of the abbreviation/syntax mechanism.

proof/refine/ARM/ArchInvsLemmas_H.thy
```diff
@@ -111,7 +111,6 @@ clear_named_theorems Arch_assms (* accumulate assumptions for Invariants_H_cte_a
 
 (* FIXME arch-split: for proofs which require exact offsets lining up instead of cteSizeBits *)
 lemma raw_tcb_cte_cases_simps:
-  "tcb_cte_cases 0  = Some (tcbCTable, tcbCTable_update)"
   "tcb_cte_cases 16 = Some (tcbVTable, tcbVTable_update)"
```
lsf37, line 114: It's a bit strange that this lemma set ends up being incomplete. I can see a future self wanting to complete the set. Can you add a comment that and where the 0 case is already covered?

proof/refine/ARM_HYP/ArchInvsLemmas_H.thy
```diff
@@ -108,7 +108,6 @@ clear_named_theorems Arch_assms (* accumulate assumptions for Invariants_H_cte_a
 
 (* FIXME arch-split: for proofs which require exact offsets lining up instead of cteSizeBits *)
 lemma raw_tcb_cte_cases_simps:
-  "tcb_cte_cases 0  = Some (tcbCTable, tcbCTable_update)"
```
lsf37, line 111: (same as for ARM)

proof/refine/CNodeInv_R.thy
```diff
@@ -5839,6 +5843,8 @@ crunch capSwapForDelete
 context CNodeInv_R begin
 
 lemma finaliseSlot_abort_cases':
+  includes no_valid_cap_syn
+  shows
   "s \<turnstile> \<lbrace>\<top>\<rbrace>
```
lsf37, line 5848: If we're adding a `shows`, we do need to indent the rest of the lemma statement by 2 under the `shows` or start the entire statement after the `shows`.

proof/refine/CNodeInv_R.thy
```diff
@@ -6563,6 +6571,8 @@ lemmas finalise_induct3 = finaliseSlot'.induct[where P=
     "\<lambda>sl exp s. P sl (finaliseSlot' sl exp) s" for P]
 
 lemma finaliseSlot_rvk_prog:
+  includes no_valid_cap_syn
+  shows
   "s \<turnstile> \<lbrace>\<lambda>s. revoke_progress_ord m (option_map capToRPO \<circ> cteCaps_of s)\<rbrace>
```
lsf37, line 6576: shows statement indent

proof/refine/CNodeInv_R.thy
```diff
@@ -8740,6 +8751,8 @@ crunch capSwapForDelete
   (wp: withoutPreemption_lift)
 
 lemma preemptionPoint_IRQInactive_spec:
+  includes no_valid_cap_syn
+  shows
   "s \<turnstile> \<lbrace>valid_irq_states'\<rbrace>
```
lsf37, line 8756: shows statement indent

proof/refine/CNodeInv_R.thy
```diff
@@ -8753,6 +8766,8 @@ crunch finaliseCap
    simp: crunch_simps o_def)
 
 lemma finaliseSlot_IRQInactive':
+  includes no_valid_cap_syn
+  shows
   "s \<turnstile> \<lbrace>valid_irq_states'\<rbrace>
```
lsf37, line 8771: shows statement indent

proof/refine/CNodeInv_R.thy
```diff
@@ -8813,6 +8828,8 @@ lemma cteDelete_irq_states':
   done
 
 lemma cteRevoke_IRQInactive':
+  includes no_valid_cap_syn
+  shows
   "s \<turnstile> \<lbrace>valid_irq_states'\<rbrace> cteRevoke x
   \<lbrace>\<lambda>_. \<top>\<rbrace>, \<lbrace>\<lambda>rv s. intStateIRQTable (ksInterruptState s) rv \<noteq> irqstate.IRQInactive\<rbrace>"
```
lsf37, line 8834: shows statement indent

proof/refine/CSpace_R.thy
```diff
@@ -5180,6 +5180,7 @@ lemma locateSlot_cap_to'[wp]:
 (* FIXME: ambiguous notation between spec_validE (this) and validE; introduce bundles for both,
    unbundle by default, and use `includes no` contexts to disambiguate *)
 lemma rab_cap_to'':
+  includes no_valid_cap_syn
   assumes P: "\<And>cap. isCNodeCap cap \<longrightarrow> P cap"
   shows
   "s \<turnstile> \<lbrace>\<lambda>s. isCNodeCap cap \<longrightarrow> (\<forall>r\<in>cte_refs' cap (irq_node' s). ex_cte_cap_wp_to' P r s)\<rbrace>
```
lsf37, line 5186: It looks like people have already done this adding of `shows` without indent here before (maybe that is where you picked it up). If we're changing this lemma, we should probably fix that up.

lsf37 (approved):
Nice!

Not much we can do about the name duplication in the del/dest/elim pattern without rewriting the parsers of fastforce and auto, so I think this is the best we can achieve for now.

proof/invariant-abstract/Invariants_AI.thy
Xaphiosis: I'm not sure I'm up to date on the `no` tech. On the surface level, doesn't your example result in no actual syntax being given to `valid_cap_syn` outside of contexts where the bundle is included? i.e. if "open_bundle is the same as bundle x <block> followed directly by unbundle x", then the syntax will be off by default, no?

lsf37: Because of the implicit `unbundle`, the syntax will be on by default exactly like before.

proof/refine/ARM/ArchVSpace_R.thy
Xaphiosis: In an ideal world we could patch upstream in some way, but of course contributions are not welcome. Maybe we should consider trying it on our fork one of these days when there's more time?

proof/refine/CSpace_R.thy
Xaphiosis: Yeah, I was blending in with the existing master works to create a smaller diff. Will adjust.

proof/invariant-abstract/Invariants_AI.thy
Xaphiosis: Will do. This is kind of annoying, because the abbreviation is now only for swapping the arguments, otherwise it could've been a syntax bundle for valid_cap itself. Ho hum.

proof/refine/ARM/ArchInvsLemmas_H.thy
Xaphiosis: will add `(* 0 case already covered by tcb_cte_cases_simps *)` on all arches

proof/refine/IpcCancel_R.thy
Xaphiosis: Not needed, will remove, thanks for catching it

proof/refine/CNodeInv_R.thy
Xaphiosis: does this look more reasonable?
```
lemmas finalise_spec_induct =
  finaliseSlot'.induct[
    where P="λsl exp s. s ⊢ ⦃P sl exp⦄ finaliseSlot' sl exp ⦃Q sl exp⦄,⦃E sl exp⦄" for P Q E]
```

Xaphiosis:
Hopefully I've addressed all the PR comments. Will still need to squash once you're ok with the changes.

proof/refine/CNodeInv_R.thy
corlewis: A lot more, thank you!

corlewis:
Changes look good to me

lsf37 (approved):
Changes look good, thank you!

PR #1057, opened 2026-09-12
cparser: check for zero-length array

### LINKED ISSUE #303: missing error message for 0-length array
See discussion on seL4/seL4#443 -- the parser should produce an error message for this case, not just fail silently.

lsf37:
Zero-length arrays were already previously not supported, they just failed with an obscure internal error message at a later point. Instead check for these cases explicitly in the front-end and produce a readable error message.

In particular:

- check for zero-length array and reject
- check for empty initialiser list and reject (would produce zero-length array)
- improve existing array error messages to use Feedback.Error instead of raising an exception.
- avoid duplicate error message with bogus source position (see separate commit message for details)
- (in separate commit): add test cases for all array length errors

Best to review per commit. The new check and new tests are simple, avoiding the duplicate error message was a bit trickier.

Fixes https://github.com/seL4/l4v/issues/303


tools/c-parser/StrictC.grm
```diff
@@ -1449,6 +1449,10 @@ initializer
       (wrap(InitE rexpression, eleft rexpression, eright rexpression))
 | LCURLY initializer_list RCURLY
       (wrap(InitList initializer_list, LCURLYleft, RCURLYright))
+| LCURLY RCURLY
+      (errorStr'(LCURLYleft, RCURLYright,
+                 "Empty initializer lists are not supported");
+       wrap(InitList [], LCURLYleft, RCURLYright))
```
midnightveil, line 1455: So just want to confirm: are you explicitly saying that

```c
int arr[5] = {};
```

and

```c
struct xyz abc = {};
```

Are unsupported?

---

EDIT: although, these do seem like they aren't part of "standard" C until C23: https://cppreference.com/c/language/initialization, so I must be too used to the GCC extensions.

lsf37: Yes, they are definitely unsupported. They've been unsupported before, they just were leading to more obscure error messages.

tools/c-parser/program_analysis.ML
```diff
@@ -1462,28 +1462,30 @@ in
...
               | SOME (InitList ilist) => let
                   val sz = complit.find_ilist_size ecenv ilist
-                  val _ = if sz = 0 then
-                            raise Fail "Array given empty initializer"
-                          else ()
                 in
-                  Array(elty, SOME (expr_int (IntInf.fromInt sz)))
+                  if sz = 0 then
+                    err ("Array "^str^" has empty initializer; \
+                         \zero-length arrays are not supported")
```
Xaphiosis, line 1486: There's some awkwardness mixing initialiser and initializer. The grammar and this error use the latter, and new errors you introduced use the former.

tools/c-parser/testfiles/array_size_errors.thy
```diff
@@ -0,0 +1,42 @@
...
+
+external_file "errors/zero_length_array.c"
+external_file "errors/empty_initializer.c"
+external_file "errors/unsized_local_array.c"
+external_file "errors/nonconst_array_init.c"
+
+ML \<open>
+fun expect_error file expected thy =
+  case Exn.capture (IsarInstall.interactive_install file) thy of
+    Exn.Res _ => error ("Expected C parser error in " ^ file)
```
Xaphiosis, line 21: hang on, if the other cases are examine the error message and re-raise, then what's this case? The message doesn't seem to illuminate it for me.

tools/c-parser/testfiles/errors/nonconst_array_init.c
```diff
@@ -0,0 +1,15 @@
...
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+/* Array size must be static and must be a constant expression */
+
+int g(void);
+
+int f(void)
+{
+  int a[] = g();
```
Xaphiosis, line 13: is this ok? g returns an int, and this is assigning it to an int[] ... I'm not familiar with that version, I'd expect a {} somewhere

Xaphiosis:
Commit suggestion:
"already previously not supported" -> basically actually simply drop previously? :)

tools/c-parser/testfiles/errors/nonconst_array_init.c
lsf37: It's not ok, you're right that this would fail anyway in type checking, it just happens to fail on the new check first. I should change this to fail only because of the length check, not also on other things.

tools/c-parser/testfiles/array_size_errors.thy
lsf37: This case is "no exception happened, normal result was returned". In the negative tests below that's a failure, because an exception should happen, so it uses `error` and indicates that it was expecting an exception to happen. Maybe it should be `"Expected .. in " ^ file ^ ", but no error happened."`

tools/c-parser/program_analysis.ML
lsf37: right, I'm disagreeing with Michael on those. I shall adjust to what is there.

tools/c-parser/testfiles/array_size_errors.thy
Xaphiosis: In your asm PR you have:
```
  case Exn.capture (install fname) @{theory} of
    Exn.Res _ => error ("expected " ^ fname ^ " to be rejected")
```
which made sense to me. I think the double negative is hard to parse quickly ("Error: no error"), but now that I've managed to get some sleep it makes more sense. I might also frame it as `"C parser did not produce an error for " ^ file` if that's better.

lsf37: Right, I forgot that I already had a better solution for that. More sleep is indeed useful. I'd be happy to just use

```
error ("expected " ^ file ^ " to be rejected")
```

again.

tools/c-parser/testfiles/errors/nonconst_array_init.c
lsf37: Added a slightly more complicated version that now compiles when it's not an initialiser. This means the only reason it is rejected now is the non-constant initialiser position.

tools/c-parser/program_analysis.ML
lsf37: I found two more and fixed those as well.

lsf37:
Rebased, review feedback addressed, and commit message adjusted. Should be ready to merge.

PR #1058, opened 2026-09-13
simpl: pretty printing for const proc names 

### LINKED ISSUE #710: SIMPL: don't print `_'proc`
In `CALL` and `PROC `, the pretty printer inserts the internal `_'proc` in the output. It should not.

lsf37:
Procedure names that were constants were missed by the existing print translation and were printed for instance as `Call f_'proc`. Add the missing Const case and also apply abbreviation contraction to formal parameter names, which are then printed via `PROC` for typed mangled C parser names.

This prints C-parser generated procedure calls correctly now, with output that is equivalent legal input. This also means that bare `Call f_'proc` terms now (correctly) print as `PROC \<acute>f()`.

To remove the `PROC` syntax (leaving only `CALL` and bare `Call f_'proc`), use print mode `NoProc`.

This commit affects printing only, there are no changes to semantics.

In a separate commit (for easier upstreaming): add test cases that make sure the output is legal, equivalent input and that `_'proc` is no longer printed.

Fixes #710


tools/c-parser/Simpl/hoare_syntax.ML
```diff
@@ -1457,33 +1457,44 @@ fun gen_fcall_tr' ctxt init p return result c =
...
+
+      fun dest_string (Const (@{const_syntax Nil}, _)) = []
+        | dest_string (Const (@{const_syntax Cons}, _) $ c $ cs) = dest_chr c :: dest_string cs
+        | dest_string _ = raise Match;
+      in
+        (case try dest_string p of
+          SOME name => Syntax.const (String.implode name)
+        | NONE => antiquote_mult_tr' ctxt (K true) antiquoteCur antiquoteOld p)
+      end;
+in
```
Xaphiosis, line 1485: Uhh I'm lost as to where this came from. I'm guessing this came from HOL directly? Maybe mention the file it came from?

Xaphiosis (approved):
Magic. Thanks for fixing this.

In commit: `no change to semantics` -> changes

tools/c-parser/Simpl/hoare_syntax.ML
lsf37: This is just moved (factored out) from 4 lines below :-)

lsf37: Basically, `pname_tr'` (print translation for pname) had a default case where it did all this string stuff. The new `Const` case is looking at the name of the `Const`: if it ends in `_proc'` it cuts of the suffix, and if it does not end in `_proc'` it does the default case to preserve whatever it was doing before for `Const`. (Hence the name `pname_default_tr'`, for "default case of the pname print translation").

Xaphiosis: Ah, I got confused by `(* from HOL strings to ML strings *)` and then figured there should be functionality like this in HOL already, and then wondered if that's where this came from.

lsf37: It does look pretty confusing in the diff

PR #1059, opened 2026-09-13
c-parser: explicitly check for inline asm statement errors

lsf37:
Before this, inline asm statement errors, e.g. for unsupported types or syntax on input/output operands would lead to obscure exceptions/type errors further down in statement translation in Isabelle.

Instead, explicitly check for type, width, and lvalue where appropriate, and produce a readable error message that also is displayed int the standalone parser.

For unsupported inline asm statement forms, warn only for now. The translation is unchanged and produces and `asm_semantics_ok_to_ignore` constant, which then later will fail in SimplExport, but only in architectures where SimplExport exists.

This is a bit unsatisfactory, but producing an error would break for instance ARM_HYP. I propose now that the standalone parser produces warnings for it, that we fix these and promote the warning to an error later (maybe via config option) when all of them are eliminated in seL4.

(Apparently there has always been a CLI option `--unhandled_asm` that we could have used to find these. Have added a hint towards that in the change log).


lsf37:
Great, glorious x64 has problematic inline asm. I'll have a look tomorrow, but the principles should be reviewable already.

lsf37:
It's writing to a `void *`, which I had forbidden, because it would have lead to a type error in Isabelle and why would you do that in the first place.

It turns out we set up hooks to ignore all inline asm translation in x64, so the type error didn't happen there. Staring at this, I think I could actually allow pointers on the lhs. The C semantics doesn't know anyway what comes out of inline asm (just that something comes out), and the binary verification side treads pointers as word_t already. I'll try that tomorrow.

midnightveil:
> It's writing to a void *, which I had forbidden, because it would have lead to a type error in Isabelle and why would you do that in the first place.

I don't think that function is actually used as far as I can tell. (`get_current_esp`). I doubt we should be writing to the stack manually anyway...

lsf37:
Good point, let me check that. If it's unused it would be better to remove it.

lsf37:
Ok, with that, the standalone parser works and one omission in my change was unmasked (the types in the asm statements need to be normalised first before they can be properly compared, otherwise typedefs are still symbols and not the type they point to). I've fixed that and squashed it in directly.

tools/c-parser/RELEASES.md
```diff
@@ -186,3 +186,15 @@
   the generation of modifies proofs. Default behaviour is unchanged.
 - Allow the character ' in Isabelle identifiers in FNSPEC, so that for instance
   the name StrictC'_f_spec can be used as a lemma identifier.
+
+
+## 1.23 (upcoming)
+
+- Explicitly check operand types of inline `asm` statements (outputs must be
+  unsigned machine-word-size lvalues, inputs must be scalars of at least
+  machine-word width).
```
Xaphiosis, line 195: I'm confused by the discussion on this PR. Is the final result that writing to `void*` is permitted? Is that included in the above comment?

midnightveil: It's not permitted.

tools/c-parser/testfiles/asm_ok.c
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright 2026, Proofcraft Pty Ltd
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+/* asm statements that are accepted */
```
Xaphiosis, line 7: extra newline before the typedef, as this is referring to the whole file?

tools/c-parser/testfiles/asm_ok.c
```diff
@@ -0,0 +1,33 @@
...
+            irq_t irq)
+{
+  word_t v;
+  asm volatile("mrs %0, tpidr_el1" : "=r"(v));
+  asm volatile("msr tpidr_el1, %0" : : "r"(x));
+  asm volatile("msr tpidr_el1, %0" : : "r"(u));            /* narrower: cast */
+  asm volatile("msr tpidr_el1, %0" : : "r"(h));            /* narrower: cast */
+  asm volatile("msr tpidr_el1, %0" : : "r"(i));            /* signed: cast */
+  asm volatile("msr tpidr_el1, %0" : : "r"(p));            /* pointer */
+  asm volatile("msr tpidr_el1, %0" : : "r"(p->b));         /* array decays */
```
Xaphiosis, line 24: decays?

tools/c-parser/testfiles/asm_ok.c
Xaphiosis: Also, in this file it doesn't show that you can have two inputs. Not sure if that's relevant, but someone might use this file as a reference for everything which is permitted one day

tools/c-parser/testfiles/asm_errors_ptr_output.c
```diff
@@ -0,0 +1,13 @@
...
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+/* Error: "=r" output operand of pointer type */
+unsigned long *f(void)
+{
+  unsigned long *p;
+  asm volatile("mrs %0, tpidr_el1" : "=r"(p));
+  return p;
```
Xaphiosis, line 12: re-iterating confusion about outcome of `void*` discussion

tools/c-parser/testfiles/asm_errors.thy
```diff
@@ -0,0 +1,60 @@
...
+      else error ("unexpected error for " ^ fname ^ ": " ^ msg)
+  | Exn.Exn exn => Exn.reraise exn
+
+val output_msg = "asm output operand"
+val input_msg = "asm input operand"
+\<close>
+
+ML \<open>
+expect_error "asm_errors_signed_output.c" (output_msg ^ " \"=r\"(v) has type int. Only unsigned");
+expect_error "asm_errors_narrow_output.c" (output_msg ^ " \"=r\"(v) has type unsigned_short. Its width (16)");
```
Xaphiosis, line 39: these two are kinda cryptic; are they substrings of a larger error?

tools/c-parser/RELEASES.md
Xaphiosis: Mmm. But then the docs might need a slight amendment, because technically a `void*` is an lvalue of word size, at least as far as I know. It's not suitable for us due to type reasons.

Xaphiosis:
Commit message thoughts:
`and produces and asm_semantics_ok_to_ignore` -> and produces *an*
`which then later will fail` -> not sure, `which will then later fail` maybe?
`would break for instance ARM_HYP` -> ok to stay, but Englishes

tools/c-parser/RELEASES.md
lsf37: hm, me with trying to keep things short all the time. My original version of this text had `outputs must be lvalues that are unsigned integers of machine-word size, ..`. Would that work? 

On the pure C side, it'd be perfectly fine to write to a pointer type, it's just us who are restricting it further, because it's hairy to handle later.

tools/c-parser/testfiles/asm_ok.c
lsf37: Good point, I'll add some.

tools/c-parser/testfiles/asm_ok.c
lsf37: the array decays to a pointer. I should be less afraid of long lines :-)

tools/c-parser/testfiles/asm_errors.thy
lsf37: yes, they are just substrings. I was too lazy to type them out and convinced myself that this is good actually, because this makes them less sensitive to change in the message. But indeed a bit cryptic. Happy to expand it to the full message.

tools/c-parser/RELEASES.md
Xaphiosis: I think I'd like your old version more. You could also add `(except void*)`, i.e. something like:
```
- Explicitly check operand types of inline `asm` statements: outputs must be
  unsigned machine-word-size lvalues (except void*), while inputs must be scalars of at least
  machine-word width.
```

tools/c-parser/testfiles/asm_errors_ptr_output.c
Xaphiosis: I think this is resolved now pending comment on docs.

tools/c-parser/testfiles/asm_errors.thy
Xaphiosis: No need, I think this is fine, was just checking since they felt a bit odd.

tools/c-parser/RELEASES.md
lsf37: I'll try one more round, because outputs are only allowed to be actual unsigned integers, i.e. no pointers at all (including *void, but also others). Inputs are more relaxed and can be pointers and other scalars.

lsf37:
Rebased, fixed up, and a new iteration for the explanation in RELEASES.md. @Xaphiosis if you're happy with that, I think this is ready to merge.

PR #1060, opened 2026-09-15
lib/monads: apply wp_pre as first step of wps

### LINKED ISSUE #1019: The `wps` method should include a `wp_pre` step
Currently, `wp`, `wpsimp` etc automatically perform `wp_pre` as the first step, whereas `wps` does not.

We should make these consistent and apply `wp_pre` in `wps` as well.

bernardladenthin:
Hi! Please check:

Fixes #1019 (raised in the review of #1011).

`wps_tac` now runs `wp_pre` first, like `wp` and `wpsimp`. The step is a no-op when the precondition is already schematic; crunch applies its own pre rules before `wps_tac`, so it is unaffected. A regression test in `WPEx.thy` fails without the change.

One behavioural difference: on a goal with a concrete precondition, `wps` now leaves the `P s ==> ?P' s` side goal that `wp` also leaves. Inside a `(wp | wps)+` loop this side goal can end up in front of remaining sibling goals and stop the loop. Across the 541 `wps` uses in the repo this happened in exactly two proofs, both in AARCH64 `ArchUntyped_AI.thy`; they now close the side goal inside the loop (`(wp mapM_x_wp' | wps | simp)+`).

Tested with `./run_tests` on l4v c18eef3c5 (seL4 28b8f4c, Isabelle2025-2, `trustworthysystems/l4v` image), all passing:
- ARM: AInvs Access InfoFlow Refine CRefine DRefine Bisim LibTest
- AARCH64: AInvs Access InfoFlow Refine CRefine LibTest
- RISCV64: AInvs Access InfoFlow Refine CRefine Bisim LibTest
- X64: AInvs Refine CRefine Bisim LibTest
- ARM_HYP: AInvs Refine CRefine LibTest

Not touched: the 83 existing `wp_pre, wps` / `hoare_pre, wps` sites are now redundant but still work; happy to clean them up here or in a follow-up.


lsf37:
We cannot take AI generated pull requests, sorry.

bernardladenthin:
Its fine! Ty for your reply. Maybe it helps as example / poc ;)

lsf37:
> Its fine! Ty for your reply. Maybe it helps as example / poc ;)

If you're happy with that, I might pick up the few lines with the actual fixes later, tweak things a bit and mention you on the PR for that. 

Thanks for reaching out!

bernardladenthin:
Hey! You dont need to mention me if problematic :) I came across this with some local experiments and I'm just happy if this helps somehow :) Thanks a lot for all the effort! Greetings, Bernard

lsf37:
No worries. We do like to give credit where credit is due.

PR #1061, opened 2026-09-15
Proof updates for AArch64 XN right fixes 

lsf37:
- Sync proofs with seL4 change https://github.com/seL4/seL4/pull/1749 and always set NX when "Read" is not set to make AArch64 semantics the same as AArch32.

- In a separate commit, make it visible in the infoflow proofs that NX confers Read authority to user ops by revealing executable non-read pages in the automaton to user ops. Then adjust proofs accordingly to show that the fix in seL4 actually fixes the inflow leak.

Best to review by commit. The first one (functional correctness) is simple, the second one (infoflow scope extension) is slightly larger.

Test with: https://github.com/seL4/seL4/pull/1749


proof/crefine/AARCH64/Arch_C.thy
```diff
@@ -1756,11 +1756,34 @@ lemma cap_frame_cap_lift_asid_upd_idem:
...
 
+(* FIXME: move to Word_Lib *)
+lemma one_or_from_bool[simp]:
+  "1 || from_bool v = 1"
+  by (simp add: from_bool_def split: bool.split)
+
+(* FIXME: move to Word_Lib *)
+lemma from_bool_or_one[simp]:
+  "from_bool v || 1 = 1"
+  by (simp add: from_bool_def split: bool.split)
```
lsf37: I guess I should actually do these

lsf37: Let me briefly clean that up before you start reviewing.

proof/invariant-abstract/AARCH64/ArchADT_AI.thy
```diff
@@ -123,6 +123,11 @@ definition ptable_rights :: "obj_ref \<Rightarrow> 'z::state_ext state \<Rightar
    case_option {} (snd o snd o snd)
       (get_page_info (aobjs_of s) (get_vspace_of_thread (kheap s) (arch_state s) tcb) addr)"
 
+definition ptable_exec :: "obj_ref \<Rightarrow> 'z::state_ext state \<Rightarrow> machine_word \<Rightarrow> bool" where
+  "ptable_exec tcb s \<equiv> \<lambda>addr.
+   case_option False ((\<lambda>attrs. Execute \<in> attrs) o fst o snd o snd)
+      (get_page_info (aobjs_of s) (get_vspace_of_thread (kheap s) (arch_state s) tcb) addr)"
```
lsf37, line 129: I know the style here is funky, but it's consistent with the definitions above and in the other architectures, so I left it this way. We might want to clean this up more generally, though.

proof/crefine/AARCH64/Arch_C.thy
lsf37: Done.

lsf37:
(Set this to draft for now, because we're discussing a different change to the C code which would allow execute-only pages when the cap has Read, which would allow fuller use of the hardware mechanism)

lsf37:
This is now good to go for review.

Xaphiosis:
Some confusion in commit messages. We have:
* NX (first commit)
* xn (second commit title line)
* XN (second commit)
Are any of these actually different entities, or should they all be the same? (my hazy memories have it as XN, execute-never).

Will try look at this later today, or if it spills, tomorrow.

proof/infoflow/Noninterference.thy
```diff
@@ -418,6 +416,14 @@ lemma dmo_user_memory_update_globals_equiv_scheduler:
   apply (clarsimp simp: globals_equiv_scheduler_def split: option.splits)
   done
 
+lemma dmo_user_memory_update_globals_equiv_scheduler:
```
ryybrr, line 419: Where's this one used anyway?

proof/access-control/AARCH64/ArchAccess.thy
```diff
@@ -12,23 +12,23 @@ context Arch begin arch_global_naming
 
 subsection \<open>Arch-specific transformation of caps into authorities\<close>
 
-definition vspace_cap_rights_to_auth :: "cap_rights \<Rightarrow> auth set" where
-  "vspace_cap_rights_to_auth r \<equiv>
+definition vspace_cap_rights_to_auth :: "cap_rights \<Rightarrow> bool \<Rightarrow> auth set" where
```
ryybrr, line 15: Realising now that this is probably a misnomer given it's used in `pte_ref2`

ryybrr (approved):

spec/abstract/AARCH64/ArchDecode_A.thy
```diff
@@ -70,9 +70,16 @@ definition arch_decode_irq_control_invocation ::
...
+text \<open>
+  On AArch64, read access is not required to execute code, but executing code effectively confers
+  read access to the page contents. Hence, we force execute-never if the frame cap does not have
+  the Read right. The masked @{text rights} may still be less than the cap's rights, so that
+  execute-only mappings are possible with a cap that has the Read right.\<close>
+definition make_user_pte ::
+  "paddr \<Rightarrow> vm_attributes \<Rightarrow> vm_rights \<Rightarrow> vm_rights \<Rightarrow> vmpage_size \<Rightarrow> pte" where
+  "make_user_pte addr attr rights crights vm_size \<equiv>
+     let attrs = attr - {Global} - (if AllowRead \<notin> crights then {Execute} else {})
+     in PagePTE addr (vm_size = ARMSmallPage) attrs rights"
```
corlewis, line 82: Very minor but I'd prefer `cap_rights` instead of `crights`. The comment hints towards this but I didn't make the connection until I saw the explicit name in the Haskell version.

corlewis: And I guess there's the usual question about whether it's better to match incorrect local style, but the `where` should be on a new line.

corlewis:
> Some confusion in commit messages. We have:
> 
>     * NX (first commit)
> 
>     * xn (second commit title line)
> 
>     * XN (second commit)
>       Are any of these actually different entities, or should they all be the same? (my hazy memories have it as XN, execute-never).

As another data point, looks like in infoflow the changes here refer to it with xn.

corlewis (approved):
Apart from the terminology inconsistency pointed out by Raf, looks good to me!

proof/infoflow/Noninterference.thy
corlewis: Looks like it's still used in `ArchNoninterference` on ARM and RISCV64.

spec/abstract/AARCH64/ArchDecode_A.thy
lsf37: cap_rights is unfortunately a constant already, so not available as parameter name. I could do caprights, but looked strange to me.

Indanz:
> Some confusion in commit messages. We have:
> 
>     * NX (first commit)
> 
>     * xn (second commit title line)
> 
>     * XN (second commit)
>       Are any of these actually different entities, or should they all be the same? (my hazy memories have it as XN, execute-never).

They should all be XN on Arm:
- Arm has an execute-never bit (XN).
- x86 has a no-execute bit (NX).
- RISC-V is sane and has an explicit execute bit (X).

lsf37:
> > Some confusion in commit messages. We have:
> > ```
> > * NX (first commit)
> > 
> > * xn (second commit title line)
> > 
> > * XN (second commit)
> >   Are any of these actually different entities, or should they all be the same? (my hazy memories have it as XN, execute-never).
> > ```
> 
> As another data point, looks like in infoflow the changes here refer to it with xn.

The official Arm terminology is XN (execute never). AMD calls it NX (no-execute), and people usually use that. And Intel calls it XD (execute disable), because why not confuse everyone even more.

Will change it to XN in commit messages, since this is all AArch64 specific here.

lsf37:
> They should all be XN on Arm:

Indeed, agreed. Managed to miss your message when I replied above.

proof/access-control/AARCH64/ArchAccess.thy
lsf37: True, it's a bit more general. I think it's used for both. Happy to do renaming in a separate PR if we want that. I'll go ahead and merge now so that the fix is in.

PR #1062, opened 2026-09-15
ARM: reserve hardware ASID 0

lsf37:
This is the proof update for https://github.com/seL4/seL4/pull/1750 and contains a moderate amount of cleanup for ARM/ARM_HYP, which transplants some of the AARCH64 VMID proof setup back to ARM/ARM_HP (for findFreeHWASID and related functions).

Some of the comments mention VMID in anticipation of a similar PR for AARCH64 that will share the same setup.

Test with: https://github.com/seL4/seL4/pull/1750

Proof commits are currently split by main session for review, they should probably be squashed into one when we merge.

What is roughly going on in the proofs:

- there is a new reserved_hw_asid
- find_free_hw_asid skips this reserved hw asid
- set_current_pd was removed and replaced with more appropriate functions on the C side (which function depends on ARM vs ARM_HYP)
- there is a new invariant that next_asid is never the reserved asid
- there is a separate (unforced) proof that find_free_hw_asid never returns the reserved asid
- I aligned our definition of `init` from Haskell (takes all but the last elements of a list) with the Isabelle built-in `butlast` and proved it equivalent. That makes the specs nicer to write and the find_free_hw_asid lemmas easier to prove. I have not removed `init` completely, because some of the existing proofs depend on its current form.


lsf37:
The other architectures have a small tweak in their `refine` proofs in `ArchSchedule_R`, because I removed a stray `corres_assert_assume_r` in `Schedule_R` and added a stronger version to Lib.

lsf37:
This is now ready for review. On the C side it is still not fully settled if there should be isb/dsb when setting the current PD/ASID, but that is trivial to change and would not need a re-review of the verification side.

lib/LemmaBucket.thy
```diff
@@ -528,4 +528,38 @@ lemma Least_sym_eq_trivial[simp]:
   unfolding Least_def
   by (clarsimp cong: conj_cong)
 
+lemma butlast_contract_Cons:
+  "xs \<noteq> [] \<Longrightarrow> x # butlast xs = butlast (x # xs)"
+  by simp
```
Xaphiosis, line 533: bit unexpected, I figured you'd want a rule the other way around

lib/Lib.thy
```diff
@@ -1597,6 +1597,10 @@ lemma comp_upd_simp:
   "(f \<circ> (g (x := y))) = ((f \<circ> g) (x := f y))"
   by (rule fun_upd_comp)
 
+lemma map_option_fun_upd:
+  "(\<lambda>x. map_option g ((f(y := None)) x)) = (map_option g \<circ> f)(y := None)"
+  by (auto simp add: comp_upd_simp comp_def)
```
Xaphiosis, line 1602: It's called `map_option_fun_upd`, but it's only dealing with a specific form of upd (None). Also it folds the function composition, which I can't see from here how it helps.

proof/refine/X64/ArchSchedule_R.thy
```diff
@@ -427,7 +427,7 @@ lemma guarded_switch_to_chooseThread_fragment_corres[Arch_assms]:
       apply (rule corres_assert_assume_l)
       apply (rule corres_assert_assume_r)
       apply (rule switchToThread_corres)
-     apply (wp gts_st_tcb_at)+
+     apply (wpsimp wp: gts_st_tcb_at gts_st_tcb_at')+
```
Xaphiosis, line 430: hang on, what are these doing as part of the `corres_assert_assume_r` commit? Nothing about `gts_st_tcb_at'` appears in this commit or before it.

spec/design/m-skel/ARM_HYP/MachineTypes.thy
```diff
@@ -143,8 +143,35 @@ abbreviation (input) "pt_size_index \<equiv> 9"
 abbreviation (input) "pd_size_index \<equiv> 11"
 abbreviation (input) "pt_slot_vaddr_mask \<equiv> 0x1FF"
 
-(* Machine/Hardware/ARM.lhs - hardware_asid, vmfault_type and vmpage_size *)
-#INCLUDE_HASKELL SEL4/Machine/Hardware/ARM.lhs CONTEXT ARM_HYP ONLY HardwareASID VMFaultType HypFaultType VMPageSize pageBits pageBitsForSize pageForPageBits hcrCommon hcrTWE hcrTWI hcrVCPU hcrNative vgicHCREN sctlrDefault actlrDefault gicVCPUMaxNumLR
+(* Machine/Hardware/ARM.lhs - hardware_asid; manual translation of the HardwareASID newtype,
+   so that the hardware ASID length is available as a type. *)
+type_synonym hw_asid_len = 8
+type_synonym hardware_asid = "hw_asid_len word"
```
Xaphiosis, line 149: This setup makes it so that `HardwareASID` newtype is simplified away to a word 8, right? Before this, did we have to manually deal with things like `HardwareASID_def`, or did the Haskell translator also give us the simps?

proof/invariant-abstract/ARM_HYP/ArchVSpace_AI.thy
```diff
@@ -1967,38 +2002,156 @@ lemma invalidateLocalTLB_ASID_valid_irq_states:
...
+       dmo_valid_machine_state dmo_valid_irq_states invalidateLocalTLB_ASID_irq_masks)
+
+crunch invalidate_asid
+  for machine_state[wp]: "\<lambda>s. P (machine_state s)"
+  and valid_global_objs[wp]: valid_global_objs
+  and equal_kernel_mappings[wp]: equal_kernel_mappings
+  and pspace_in_kernel_window[wp]: pspace_in_kernel_window
+  and cap_refs_in_kernel_window[wp]: cap_refs_in_kernel_window
+  and cur_tcb[wp]: cur_tcb
+  (simp: valid_global_objs_def equal_kernel_mappings_def)
```
Xaphiosis, line 2045: I am surprised that you needed all these new blocks of crunches here. Curious: could you say more about what happened?

proof/invariant-abstract/ARM_HYP/ArchVSpace_AI.thy
```diff
@@ -1967,38 +2002,156 @@ lemma invalidateLocalTLB_ASID_valid_irq_states:
...
+  by (simp add: valid_kernel_mappings_def)
+
+lemma valid_asid_map_hwasid_table_update[simp]:
+  "valid_asid_map (s\<lparr>arch_state := arch_state s \<lparr>arm_hwasid_table := v\<rparr>\<rparr>) = valid_asid_map s"
+  by (simp add: valid_asid_map_def vspace_at_asid_def vs_lookup_def vs_lookup1_def)
+
+lemma valid_next_asid_2_update[simp]:
+  "valid_next_asid_2 (if next_asid = maxBound then hw_asid_min else next_asid + 1)"
+  by (clarsimp simp: valid_next_asid_2_def hw_asid_reserved_def maxBound_word hw_asid_min_def
+                     word_wrap_eq_max_word)
```
Xaphiosis, line 2061: this entire `_update` block makes me think of locales and sublocales, e.g. :
```
AARCH64/ArchBits_AI.thy
16:sublocale p_asid_table_kernel_vspace_update:
20:sublocale p_asid_table_vmid_table_update:
24:sublocale p_asid_table_asid_map_update:
28:sublocale p_asid_table_next_vmid_update:
32:sublocale p_asid_table_us_global_vspace_update:
36:sublocale p_asid_table_current_vcpu_update:
40:sublocale p_asid_table_current_fpu_update:
```
are you doing it like this here because ARM/ARM_HYP lacks that kind of setup?

proof/invariant-abstract/ARM/ArchVSpace_AI.thy
```diff
@@ -145,27 +145,79 @@ lemma pd_at_asid_uniq:
...
+  and cur[wp]: cur_tcb
+  and valid_objs[wp]: valid_objs
+
+lemmas find_free_hw_asid_typ_ats[wp] = abs_typ_at_lifts [OF find_free_hw_asid_typ_at]
+lemmas invalidate_asid_typ_ats[wp] = abs_typ_at_lifts [OF invalidate_asid_typ_at]
+lemmas invalidate_hw_asid_entry_typ_ats[wp] = abs_typ_at_lifts [OF invalidate_hw_asid_entry_typ_at]
+lemmas invalidate_asid_entry_typ_ats[wp] = abs_typ_at_lifts [OF invalidate_asid_entry_typ_at]
+lemmas store_hw_asid_typ_ats[wp] = abs_typ_at_lifts [OF store_hw_asid_typ_at]
+lemmas get_hw_asid_typ_ats[wp] = abs_typ_at_lifts [OF get_hw_asid_typ_at]
+lemmas invalidate_tlb_by_asid_typ_ats[wp] = abs_typ_at_lifts [OF invalidate_tlb_by_asid_typ_at]
```
Xaphiosis: while we are here (and on ARM_HYP): remove spaces after `_at_lifts`?

proof/refine/ARM/ArchVSpace_R.thy
```diff
@@ -373,6 +373,19 @@ lemma invalidateHWASIDEntry_corres:
       apply (wp | clarsimp)+
   done
 
+lemma hwASIDMin_abs[simp]:
```
Xaphiosis: is this naming standard? I'm guessing "abs" stands for abstract instead of absolute value :)

proof/refine/X64/ArchSchedule_R.thy
corlewis: I was assuming that it's because `corres_assert_assume_r` being slightly strengthened somehow led to extra conditions appearing in concrete postconditions. Now that you've made me look a bit harder though, that doesn't really make sense, so I'm back to being confused as well.

proof/refine/TcbAcc_R.thy
```diff
@@ -3648,6 +3648,12 @@ lemma doMachineOp_bind:
  "\<lbrakk> empty_fail a; \<And>x. empty_fail (b x) \<rbrakk> \<Longrightarrow> doMachineOp (a >>= b) = (doMachineOp a >>= (\<lambda>rv. doMachineOp (b rv)))"
   by (blast intro: submonad_bind submonad_doMachineOp)
 
+lemma doMachineOp_bind_dist:
+  "doMachineOp (f >>= g) = (doMachineOp f >>= (\<lambda>x. doMachineOp (g x)))"
+  by (fastforce simp: doMachineOp_def gets_def get_def select_f_def
+                      modify_def put_def return_def bind_def
+               split: prod.splits)
```
Xaphiosis, line 3655: FIXME move

Xaphiosis (approved):
Very quick work! Minor questions and comments only.

On commits:
* there are quite a few, have you considered merging any? perhaps after review is done?
* your PR description is pretty good, but doesn't seem to make it into many commit messages; it would be really good to have something at least for the spec changes, it is after all a rare PR that has the `bug` tag
* `remove now unused setCurrentPDPL2` commit is super confusing: no extra info in the commit message, could be merged with another commit, and another large chunk of `setCurrentPDPL2` happens in the CRefine commit

spec/abstract/ARM/Machine_A.thy
```diff
@@ -44,6 +44,18 @@ type_synonym asid_len           = 17
...
+  Hardware ASID 0 is reserved for the global/empty page directory: @{text set_vm_root} installs
+  the empty page directory under hardware ASID 0 for threads without a valid page directory,
+  without a TLB flush. That is only sound if no real page directory ever runs under hardware
+  ASID 0, otherwise such a thread hits that page directory's TLB entries.
+\<close>
+definition hw_asid_reserved :: hardware_asid where
+  "hw_asid_reserved \<equiv> 0"
+
+definition hw_asid_min :: hardware_asid where
+  "hw_asid_min \<equiv> 1"
```
corlewis, line 57: Maybe not for this PR but it would be nice if these constants could be shared between the design and abstract specs.

proof/invariant-abstract/ARM/ArchVSpace_AI.thy
```diff
@@ -145,27 +145,79 @@ lemma pd_at_asid_uniq:
...
+  done
 
-lemma find_free_hw_asid_valid_arch [wp]:
-  "\<lbrace>valid_arch_state\<rbrace> find_free_hw_asid \<lbrace>\<lambda>_. valid_arch_state\<rbrace>"
-  apply (simp add: find_free_hw_asid_def invalidate_hw_asid_entry_def
-                   invalidate_asid_def do_machine_op_def split_def
+\<comment> \<open>Sanity check: @{const find_free_hw_asid} never hands out the reserved hardware ASID.
+    Not used in any proof, but a useful check on the definitions.\<close>
+lemma find_free_hw_asid_not_reserved:
+  "\<lbrace>valid_next_asid\<rbrace> find_free_hw_asid \<lbrace>\<lambda>hw_asid _. hw_asid \<noteq> hw_asid_reserved\<rbrace>"
```
corlewis: Nice idea to add this lemma (and very useful comment)!

proof/invariant-abstract/ARM/ArchVSpace_AI.thy
```diff
@@ -1602,37 +1637,127 @@ lemma invalidateLocalTLB_ASID_valid_irq_states:
...
+  by (simp add: valid_global_refs_def global_refs_def)
+
+lemma valid_global_refs_next_asid_update[simp]:
+  "valid_global_refs (s\<lparr>arch_state := arch_state s \<lparr>arm_next_asid := v\<rparr>\<rparr>)
+   = valid_global_refs s"
+  by (simp add: valid_global_refs_def global_refs_def)
+
+lemma valid_global_refs_asid_map_and_table_update[simp]:
+  "valid_global_refs (s\<lparr>arch_state := arch_state s \<lparr>arm_asid_map := v', arm_hwasid_table := v\<rparr>\<rparr>)
+   = valid_global_refs s"
```
corlewis: Something feels a bit off if you need this combined lemma. Is it possible to make the two previous lemmas have the same form as `valid_global_refs_asid_map_update`? I think them using the update functions could solve this.

proof/invariant-abstract/ARM/ArchVSpace_AI.thy
```diff
@@ -1651,24 +1776,40 @@ lemma arm_context_switch_invs [wp]:
...
+  apply(erule (1) use_valid[OF _ write_ttbr0_ptr_irq_masks])
+  done
+
+lemma set_global_pd_invs[wp]:
+  "set_global_pd \<lbrace>invs\<rbrace>"
+  apply (simp add: set_global_pd_def)
+  apply (wp dmo_invs)
+  apply safe
+   apply (drule_tac Q="\<lambda>_ m'. underlying_memory m' p = underlying_memory m p"
+          in use_valid)
```
corlewis:
```suggestion
   apply (drule_tac Q="\<lambda>_ m'. underlying_memory m' p = underlying_memory m p"
                 in use_valid)
```

proof/invariant-abstract/ARM/Machine_AI.thy
```diff
@@ -766,6 +766,26 @@ crunch handleSpuriousIRQ_mop
 
 end
 
+text \<open>Lifting rules\<close>
+
+lemma dmo_machine_state_lift:
+  "\<lbrace>P\<rbrace> f \<lbrace>Q\<rbrace> \<Longrightarrow> \<lbrace>\<lambda>s. P (machine_state s)\<rbrace> do_machine_op f \<lbrace>\<lambda>rv s. Q rv (machine_state s)\<rbrace>"
```
corlewis, line 772: Thanks for collecting these and making them consistent between ARM and ARM_HYP. While here though, is it worth doing the same for the other architectures as well?

proof/refine/ARM/ArchVSpace_R.thy
```diff
@@ -373,6 +373,19 @@ lemma invalidateHWASIDEntry_corres:
...
+  "hwASIDMin = hw_asid_min"
+  by (simp add: hwASIDMin_def hw_asid_min_def)
+
+lemma hwASIDReserved_abs[simp]:
+  "hwASIDReserved = hw_asid_reserved"
+  by (simp add: hwASIDReserved_def hw_asid_reserved_def)
+
+lemma armKSNextASID_cross:
+  "(s, s') \<in> state_relation \<Longrightarrow>
+   armKSNextASID (ksArchState s') = arm_next_asid (arch_state s)"
```
corlewis: Very minor style question, do we prefer separating the assumption like this or keeping it in one line when it fits?

proof/refine/ARM/ArchVSpace_R.thy
```diff
@@ -465,10 +470,14 @@ lemma getHWASID_corres:
...
-  "setCurrentPD = set_current_pd"
+lemma writeTTBR0Ptr_to_abs:
+  "writeTTBR0Ptr = write_ttbr0_ptr"
   by (rule ext)
-     (clarsimp simp: set_current_pd_def setCurrentPD_def writeTTBR0Ptr_def fromPAddr_def)
+     (clarsimp simp: write_ttbr0_ptr_def writeTTBR0Ptr_def fromPAddr_def)
+
+lemma empty_fail_writeTTBR0Ptr[simp, intro!]:
+  "empty_fail (writeTTBR0Ptr pd)"
+  by (fastforce simp add: writeTTBR0Ptr_def writeTTBR0_def)
```
corlewis: Could be a crunch? If not, no need for the `add` in the `fastforce`.

proof/refine/ARM_HYP/ArchVSpace_R.thy
```diff
@@ -1135,23 +1126,25 @@ lemma setVMRoot_corres:
              (no_0_obj')
              (set_vm_root t) (setVMRoot t)"
 proof -
+  have P: "corres dc \<top> \<top> set_global_pd setGlobalPD"
+    apply (simp add: set_global_pd_def setGlobalPD_def)
+    apply (rule corres_underlying_split [where P=\<top> and P'=\<top> and r'="(=)"])
```
corlewis:
```suggestion
    apply (rule corres_underlying_split[where P=\<top> and P'=\<top> and r'="(=)"])
```

proof/crefine/ARM/VSpace_C.thy
```diff
@@ -1243,14 +1270,17 @@ lemma findFreeHWASID_ccorres:
...
+            apply (clarsimp simp: maxBound_word
                                   ucast_ucast_add minus_one_norm
                            split: if_split)
             apply (simp add: word_sint_msb_eq uint_up_ucast word_size
                              msb_nth nth_ucast bang_big is_up_def source_size_def
-                             target_size_def)
-            apply uint_arith
-            subgoal by simp
+                             target_size_def hw_asid_min_def Kernel_C.hwASIDMin_def
+                             Kernel_C.hwASIDReserved_def)
```
corlewis: I'm a little bit surprised that the `hwASID` constants are being unfolded in this proof, don't we want to keep it symbolic?

corlewis (approved):
Various minor comments and suggestions, otherwise looks good!

corlewis:
> your PR description is pretty good, but doesn't seem to make it into many commit messages; it would be really good to have something at least for the spec changes, it is after all a rare PR that has the `bug` tag

Agreed! I'd also suggest something in one of the commit messages for the new `valid_next_asid` invariant.

lib/Lib.thy
lsf37: It's only true because of `map_option`, not in general, i.e. `(f(y := None) x) = (f x)(y := None)` is not a term that is type correct for the kinds of `f` that we want to use. The `map_option` allows the update to be pulled out. We could call it `map_option_fun_upd_None`, though, because you are right, it is indeed only for the None case.

The function composition doesn't really matter either way, I just found it nicer to read. In application, the main thing we're interested in is that `(y := None)` is on the outside now, which makes what's on the inside no longer that relevant, because we usually don't have to look at it. Happy to remove it if it looks confusing.

lib/LemmaBucket.thy
lsf37: It's one of those helpers that are mostly single use. As a general rewrite rule it would indeed be better the other way around, but I think there are already rules for that pattern (with an `if` on the other side, which is why I couldn't use them).

proof/refine/X64/ArchSchedule_R.thy
lsf37: I should have updated the commit message more, I was surprised by this as well:

The `corres_assert_assume_r` I added in Lib is a bit stronger than the one I removed in `Schedule_R`, which means that in this proof the precondition for the Haskell side for `getThreadState` is now schematic (as it should be). This means it needs the proper wp rule to make progress. (It was able to simplify out the case via False before, I think).

spec/design/m-skel/ARM_HYP/MachineTypes.thy
lsf37: The setup is mostly there so that we can use `hw_asid_len` in the proof for statements and instantiations (of which I instantly missed one). Because of that we can't use what the Haskell translator outputs automatically for the newtype, so I copy/pasted what it generated before, but with `hw_asid_len`.

In output, we still do see the 8, because type abbreviations expand.

proof/invariant-abstract/ARM_HYP/ArchVSpace_AI.thy
lsf37: They were implicitly lifted/unfolded in the `find_free_hw_asid` proofs in ARM and ARM_HYP before, in a bunch of giant blobs. In AARCH64 we did them separately, which means you can then actually understand the `find_free_hw_asid` proofs. I mostly copy/pasted these from AARCH64 with arch adjustments.

proof/invariant-abstract/ARM_HYP/ArchVSpace_AI.thy
lsf37: Our update locales are based on what stays the same (e.g. kheap before equals kheap after) and are then instantiated with updates that preserve those things (e.g. asid_table_update preserves kheap). These ones here are from the other direction: they are predicates that are preserved when specific things change (mostly next_asid/asid_map, some for asid_table).

They could be done in new locales, but since this is the only place they are needed, I didn't want to add more machinery for it. There are a whole bunch of them, though, so maybe it would be nicer to do that after all. I could give it a go and see what comes out.

proof/refine/TcbAcc_R.thy
lsf37: This is where I moved it, the other doMachineOp lemmas are here

proof/crefine/ARM/VSpace_C.thy
lsf37: They have to match up at some point with the abstract ones, that happens here (there's also hw_asid_min_def in the set). Might be cleaner to pull out into a separate lemma, though. I might give that a shot.

proof/invariant-abstract/ARM/Machine_AI.thy
lsf37: X64 and newer do have these sections already. They don't have all exactly the same lemmas, but they do have the same overall shape.

spec/abstract/ARM/Machine_A.thy
lsf37: Indeed. I thought about it, but they are in awkward places as so often. We really should figure out a more general place for such things. The problem is that there are dependencies between many of these (not here so much), so we probably need multiple places.

proof/invariant-abstract/ARM/ArchVSpace_AI.thy
lsf37: I had that initially, then it didn't apply properly, and then I got annoyed and just used the form in the goal :-)

I'll give it another go.

proof/refine/ARM/ArchVSpace_R.thy
lsf37: We do generally prefer on the same line, will update.

proof/refine/ARM/ArchVSpace_R.thy
lsf37: Standard might be too strong, but we do use it elsewhere, yes (and yes "abs" for "abstract"). Let me see if we have a clearer naming convention somewhere for these.

spec/machine/ARM_HYP/MachineOps.thy
```diff
@@ -177,12 +177,6 @@ definition
   dmb :: "unit machine_monad"
 where "dmb \<equiv> machine_op_lift dmb_impl"
 
-consts'
-  setCurrentPDPL2_impl :: "paddr \<Rightarrow> unit machine_rest_monad"
-definition
-  setCurrentPDPL2 :: "paddr \<Rightarrow> unit machine_monad"
-where "setCurrentPDPL2 pd \<equiv> machine_op_lift (setCurrentPDPL2_impl pd)"
-
```
Indanz, line 185: Should we remove the function from the C code too then?

lsf37: Yes we probably should. Let me check if there is any other config combination that uses it.

lsf37: Yes, we can remove it. I'll push a separate commit for that.

lsf37: Done.

proof/crefine/ARM/VSpace_C.thy
lsf37: That works out and is indeed nicer.

proof/invariant-abstract/ARM/ArchVSpace_AI.thy
lsf37: I rephrased them in the style of the _update lemmas we have where we say what we need to stay equal instead of what changes. That simplifies them a bit.

proof/invariant-abstract/ARM_HYP/ArchVSpace_AI.thy
lsf37: Ok, Corey also comment on this, so it's clearly bothering people. As replied to Corey comment: I rephrased them in the style of the _update lemmas we have where we say what we need to stay equal instead of what changes.

I have not made new locales for them, because they are still only used here and the locale dance is not really worth it, but they are now more regular and nicer.

lib/Lib.thy
lsf37: (Done the _None renaming)

proof/refine/ARM/ArchVSpace_R.thy
lsf37: We use the pattern `hwASIDMin_hw_asid_min` a lot, so I've changed to that. Avoids confusion and there are only 14 instances of `_abs`, so the other convention is clearly more popular.

lsf37:
> * there are quite a few, have you considered merging any? perhaps after review is done?

They were for review (I even mentioned it in the PR description :-). Have squashed them and will push shortly.

> * your PR description is pretty good, but doesn't seem to make it into many commit messages; it would be really good to have something at least for the spec changes, it is after all a rare PR that has the `bug` tag
> * `remove now unused setCurrentPDPL2` commit is super confusing: no extra info in the commit message, could be merged with another commit, and another large chunk of `setCurrentPDPL2` happens in the CRefine commit

I agree with both. Have squashed in the setCurrentPDPL2 commit into two parts (the spec and a proof commit) and added info from the PR description to both the overall spec commit and the proof commit, including Corey's suggest for including the new next_asid invariant.

PR #1065, opened 2026-09-17
aarch64: reserve vmid 0

lsf37:
This is the proof update for https://github.com/seL4/seL4/pull/1752. Similarly to #1062 this change reserves VMID 0 so that VMID 0 can be used for the empty page tables on threads that have no other page table.

This one is on top of #1062 because it shares the same Lib and `init`/`butlast` setup. It is also slightly simpler, because VMID and page table root can be updated atomically together.

(still on draft to check sessions after rebase)

Test with: https://github.com/seL4/seL4/pull/1752


proof/invariant-abstract/AARCH64/ArchVSpace_AI.thy
```diff
@@ -672,6 +673,7 @@ crunch get_vmid, invalidate_asid_entry, invalidate_tlb_by_asid, invalidate_tlb_b
   and valid_global_arch_objs[wp]: valid_global_arch_objs
   (ignore: set_asid_pool simp: valid_global_arch_objs_upd_eq_lift)
 
+
```
Xaphiosis: extra newline

proof/invariant-abstract/AARCH64/ArchVSpace_AI.thy
```diff
@@ -680,7 +682,13 @@ lemma find_free_vmid_vmid_inv[wp]:
   apply wpsimp
   apply (frule is_inv_inj)
   apply (drule findNoneD)
-  apply (drule_tac x="arm_next_vmid (arch_state s)" in bspec, simp add: minBound_word)
+  apply (drule_tac x="arm_next_vmid (arch_state s)" in bspec)
+   apply (simp add: vmid_min_def upto_enum_def maxBound_word unat_minus_one_word image_iff)
+   apply (rule conjI; clarsimp)
+    apply (rule_tac x="unat (arm_next_vmid (arch_state s))" in bexI; clarsimp)
+    apply (rule unat_lt2p[where 'a=8, simplified])
```
Xaphiosis: are you sure you want 8 here and not a _len?

proof/invariant-abstract/AARCH64/ArchVSpace_AI.thy
```diff
@@ -707,7 +715,33 @@ crunch find_free_vmid
...
+crunch invalidate_asid, update_asid_map, invalidate_vmid_entry
+  for valid_next_vmid[wp]: valid_next_vmid
+
+lemma vmid_min_not_reserved[simp]:
+  "vmid_min \<noteq> vmid_reserved"
+  by (simp add: vmid_min_def vmid_reserved_def)
+
+lemma valid_next_vmid_min[simp, intro!]:
+  "valid_next_vmid_2 vmid_min"
+  by (simp add: valid_next_vmid_2_def)
```
Xaphiosis: I'm torn, on the one hand this is clearly vspace related, on the other it feels like it would be better in ArchInvariants_AI or similar near where `valid_next_vmid_2` and `vmid_min` are introduced

proof/invariant-abstract/AARCH64/ArchVSpace_AI.thy
```diff
@@ -707,7 +715,33 @@ crunch find_free_vmid
...
+  apply (clarsimp simp: valid_next_vmid_def vmid_reserved_def vmid_min_def maxBound_word
+                        max_word_wrap)
+  done
+
+lemma find_free_vmid_not_reserved:
+  "\<lbrace>valid_next_vmid\<rbrace> find_free_vmid \<lbrace>\<lambda>vmid _. vmid \<noteq> vmid_reserved\<rbrace>"
+  unfolding find_free_vmid_def valid_next_vmid_def
+  apply wpsimp
+  apply (fastforce dest!: findSomeD in_set_butlastD simp: vmid_reserved_def vmid_min_def)
+  done
```
Xaphiosis: can be two-lined if new

proof/crefine/AARCH64/VSpace_C.thy
```diff
@@ -1200,62 +1228,53 @@ lemma findFreeHWASID_ccorres:
...
+             apply (erule_tac x="armKSNextASID_' (globals s) + word_of_nat (length ys)" in allE)
+             apply (clarsimp simp: valid_arch_state'_def ran_def vmid_reserved_def hwASIDReserved_def)
+             apply (cut_tac x="armKSNextASID_' (globals s)" in uint_lt2p)
+             apply (fastforce simp: uint_of_nat ucast_eq_0 is_up)
+            apply (simp add: mask_def)
+            apply unat_arith
+           apply (rule conseqPre, vcg)
+           apply clarsimp
+          apply simp
+          apply (rule hoare_pre, wp)
```
Xaphiosis: while we are here, can this become just wp?

proof/crefine/AARCH64/VSpace_C.thy
```diff
@@ -1192,6 +1192,34 @@ crunch invalidateVMIDEntry, invalidateASID
   for nextVMID[wp]: "\<lambda>s. P (armKSNextVMID (ksArchState s))"
   (wp: crunch_wps getASID_wp)
 
+lemma vmid_full_wrapped_list:
```
Xaphiosis, line 1195: I'm a bit lost here. The lemma is saying something about if we start from nextVMID and we go to the end, then if nextVMID is not reserved, the RHS and LHS add up... but then what is the "full" doing? Especially since nextVMID can't be reserved, the list covered by the map is never "full"

Xaphiosis (approved):
This is so, so much nicer than the ARM/ARM_HYP versions. We really have improved the tech over the years.

Similar commit comments to the other PR:
* description of what's happening didn't really make it to commit messages; no reference to an issue, what's going on, etc. We'd want something at least for the spec comment, no?
* given that this is all aarch64 and that the changes are much shorter, would it make sense to squash the proof updates into a single commit?

proof/invariant-abstract/AARCH64/ArchVSpace_AI.thy
```diff
@@ -707,7 +715,33 @@ crunch find_free_vmid
...
+lemma find_free_vmid_valid_next_vmid[wp]:
+  "find_free_vmid \<lbrace>valid_next_vmid\<rbrace>"
+  unfolding find_free_vmid_def
+  apply (wpsimp split_del: if_split)
+  apply (clarsimp simp: valid_next_vmid_def vmid_reserved_def vmid_min_def maxBound_word
+                        max_word_wrap)
+  done
+
+lemma find_free_vmid_not_reserved:
+  "\<lbrace>valid_next_vmid\<rbrace> find_free_vmid \<lbrace>\<lambda>vmid _. vmid \<noteq> vmid_reserved\<rbrace>"
```
corlewis, line 731: Could add a comment here in the same way you did for `find_free_hw_asid_not_reserved`. It's not needed but it would make sure that none of us remove this lemma just because we notice that it isn't used anywhere.

corlewis (approved):
Apart from whatever has gone wrong with the proof failures, this looks good to me!

proof/invariant-abstract/AARCH64/ArchVSpace_AI.thy
lsf37: oh, we definitely want hw_asid_len, I just missed it :-) (probably did that one before I had enough and introduced hw_asid_len)

proof/invariant-abstract/AARCH64/ArchVSpace_AI.thy
lsf37: Happy to move it there, I don't think it'll interfere with anything and you're right that it's one of those very basic properties.

proof/crefine/AARCH64/VSpace_C.thy
lsf37: will check, possibly yes

proof/crefine/AARCH64/VSpace_C.thy
lsf37: I was trying to convey that it's talking about the entire loop/full list instead of any of the parts. In particular this one does include vmid_reserved, so it is the full set of values (it's UNIV of the hw_asid_type), not the list that is actually used in the code (which is without vmid_reserved). They do have different wrapping behaviour if nextVMID is vmid_reserved, because of how `[_ .e. _]` works.

Basically, this lemma is the first step in linking up the ASpec/Haskell formulation on lists (lhs) with the explicit counting loop that skips one value on the C side (rhs). It shows that they are equivalent if we don't skip vmid_reserved. The next lemmas then say that they stay equivalent if we remove one specific value.

lsf37:
> Apart from whatever has gone wrong with the proof failures, this looks good to me!

Thanks. Turns out the CI failures are from me forgetting the "Test with" line..

proof/crefine/AARCH64/VSpace_C.thy
lsf37: `wpsimp` works

lsf37:
> * description of what's happening didn't really make it to commit messages; no reference to an issue, what's going on, etc. We'd want something at least for the spec comment, no?
> * given that this is all aarch64 and that the changes are much shorter, would it make sense to squash the proof updates into a single commit?

Agreed, have done both of these now.

PR #1067, opened 2026-09-18
Backport cleanup and improvements from rt

corlewis:
None of this is particularly exciting but backporting these makes the rt branch and master more consistent, which should help in future merges.


proof/invariant-abstract/X64/ArchDetSchedSchedule_AI.thy
```diff
@@ -423,8 +422,8 @@ lemmas DetSchedSchedule_AI_handle_hypervisor_fault_assms = Arch_assms (* extract
 end
 
 global_interpretation DetSchedSchedule_AI_handle_hypervisor_fault?: DetSchedSchedule_AI_handle_hypervisor_fault
-  proof goal_cases
+proof goal_cases
   case 1 show ?case by (unfold_locales; (fact X64.DetSchedSchedule_AI_handle_hypervisor_fault_assms)?)
-  qed
+qed
```
Xaphiosis, line 427: very nice that you're cross-pollinating these to arches that RT isn't touching directly

proof/refine/Invariants_H.thy
```diff
@@ -1530,17 +1530,19 @@ lemmas tcb_cte_cases_neqs_n =
     distinct_rev[THEN iffD2, OF tcb_cte_cases_distinct_n, simplified],
     simplified conj_ac, simplified] (* remove duplicates from 0/1 simplification *)
 
-lemma tcb_cte_cases_simps[simp]:
-  "tcb_cte_cases 0  = Some (tcbCTable, tcbCTable_update)"
```
Xaphiosis, line 1534: potential issue: in the most recent cleanup of warnings from Refine, I removed a case of `raw_tcb_cte_cases_simps` with the comment `0 case already covered by tcb_cte_cases_simps`.
If you are changing `tcb_cte_cases_simps` to no longer have a raw 0 case, then you'd potentially have to update `raw_tcb_cte_cases_simps` on all arches and also remove that comment

Xaphiosis:
Wonderful to see the cleanup sync. I won't approve yet, as I suspect there is an issue at least with the raw tcb cte cases I mentioned above.

proof/invariant-abstract/X64/ArchDetSchedSchedule_AI.thy
corlewis: Thanks, I've been making an effort to at the very least not make the consistency between architectures worse while focusing on the consistency between branches.

proof/refine/Invariants_H.thy
corlewis: Very good point. I do introduce `tcb_cte_cases_simps'` just after this which should be exactly the same as the old `tcb_cte_cases_simps`, but I'll double check that this does actually work out.

corlewis: I can confirm that `tcb_cte_cases_simps'` does still have the raw 0 case. I've updated the comment in a new to-be-squashed commit.

proof/refine/AARCH64/ArchInvsLemmas_H.thy.orig
```diff
@@ -0,0 +1,586 @@
+(*
+ * Copyright 2023, Proofcraft Pty Ltd
+ * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+ *
+ * SPDX-License-Identifier: GPL-2.0-only
+ *)
+
+theory ArchInvsLemmas_H
```
lsf37: ArchInvsLemmas_H.thy.orig looks like an accidental `git add`

proof/refine/ARM/ArchInvsLemmas_H.thy.orig
```diff
@@ -0,0 +1,510 @@
+(*
+ * Copyright 2023, Proofcraft Pty Ltd
+ * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+ *
+ * SPDX-License-Identifier: GPL-2.0-only
+ *)
+
+theory ArchInvsLemmas_H
```
lsf37: also a .orig file here

proof/refine/RISCV64/ArchInvsLemmas_H.thy.orig
```diff
@@ -0,0 +1,502 @@
+(*
+ * Copyright 2023, Proofcraft Pty Ltd
+ * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+ *
+ * SPDX-License-Identifier: GPL-2.0-only
+ *)
+
+theory ArchInvsLemmas_H
```
lsf37: and one more :-)

proof/refine/X64/ArchInvsLemmas_H.thy.orig
```diff
@@ -0,0 +1,559 @@
+(*
+ * Copyright 2023, Proofcraft Pty Ltd
+ * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+ *
+ * SPDX-License-Identifier: GPL-2.0-only
+ *)
+
+theory ArchInvsLemmas_H
```
lsf37: final one, I suspect.

proof/refine/Tcb_R.thy
```diff
@@ -1750,7 +1750,7 @@ lemma invokeTCB_corres:
   apply (fastforce dest: tcb_ko_at_valid_objs_valid_tcb' simp: valid_tcb'_def)
   done
 
-end (* Tcb_R *)
+end (* Tcb_R_2 *)
```
lsf37, line 1753: nice catch!

lsf37 (approved):
Thank you for doing this. Maybe we can have nice things sometimes after all :-)

Since Raf's feedback is addressed, I don't see any further issues. Ready to merge from my side.

michaelmcinerney (approved):

proof/refine/AARCH64/ArchInvsLemmas_H.thy.orig
corlewis: Thanks for spotting, fixed.

Xaphiosis (approved):
Would people be in favour of a `*.rej` and `*.orig` lint check of some kind? I completely missed it this time around.

corlewis:
> Would people be in favour of a `*.rej` and `*.orig` lint check of some kind? I completely missed it this time around.

I'd definitely be in favour, good idea! And you didn't miss it, I accidentally added it in my follow-up commit after your suggestions.

lsf37:
Yes, I'll put that on the list for when I'm back. (Or feel free to add one, it's basically `find` succeeding or failing).

I'm mainly not sure exactly where to put it. Could be in the git check globally for all repos, or it could be a separate run just here in l4v.

lsf37:
If it's in the git check job, then it would be a failure on a required job, though, so not so easy to ignore.

corlewis:
We were talking offline, any reason not to just add them to the `.gitignore`? I can't think of any reason that we'd ever want them to be visible to git.

lsf37:
> We were talking offline, any reason not to just add them to the `.gitignore`? I can't think of any reason that we'd ever want them to be visible to git.

Yes, let's do that. Much simpler and also preventative instead of just a check.

PR #1068, opened 2026-09-19
machine: allow numDomains=256

lsf37:
The Kernel_Config condition on numDomains was slightly too strong. Domains are 0 .. numDomains-1, so numDomains = 2^domainBits (=256) is still allowed.

Amazingly, this does not break any proofs. We only had proof obligations of the form `unat d < numDomains ==> unat d < 256` and similar, which still work with `<=`.

I've run this manually on the AARCH64_zynqmp_verified.cmake config with NUM_DOMAINS=256 to check.

There's a related seL4 PR https://github.com/seL4/seL4/pull/1759 which makes sure the docs are in line with the actual limit.

The question is if we should add this as a test case to CI. We already have too many tests and NUM_DOMAINS is abstracted in the proofs now, using only the limits from Kernel_Config, so I'd be fine with not adding an explicit test. The error was that the limit was too strong, not that the proof was broken.


Xaphiosis (approved):
Glad to see our proofs were robust, even with a bit of a derp in the assumptions. Many thanks for spotting this!

PR #1069, opened 2026-09-19
aarch64 spec+proof: context-switch two more registers

lsf37:
Proof update for the corresponding seL4 change seL4/seL4#1761 that adds context switching for VCPURegCSSELR_EL1 and VCPURegVDISR_EL2, and changes the control register hcrVCPU to trap on access to RAS error registers.

All of this just means a tweak to one constant and two more registers in the proof.

Test with: seL4/seL4#1761


Xaphiosis (approved):
Thanks for quick turnaround on this. Luckily it's all at the hardware assumption boundary layer.

corlewis (approved):

